PR #83, opened 2024-10-09
[Cherry-pick] Enable NX Memory Protections [Rebase & FF]

makubacki:
This PR consists of three commits:

- impl Debug for StateTransition
- Remove old memory access attributes when setting image memory attributes
- Set NX on all newly allocated memory from the GCD
  - This commit also sets the software memory capabilities for every GCD range

It is a caller's responsibility to set the memory attributes for their memory range if they do not want NX set. It is also a caller of set_memory_space_capabilities() to not remove the memory access attributes as capabilities (this may be a future check in set_memory_space_capabilities. It will already fail if the attribute is already set.

Tested by booting to Windows.

Note: This will not cover all memory regions until mu-paging is brought in.


makubacki:
Replaced by PR from local repo branch - https://github.com/pop-project/uefi-dxe-core/pull/108

PR #85, opened 2024-10-09
Bugfix: Fix broken headers in mdbook

Javagedes:
A previous commit to satisfy markdownlint requirements broke the headers in the mdbook. The `#` header shows up in the mdbook as a new section outside of chapters. Changing to `##` ended up breaking the intended formatting of the mdbook as those headers no longer showed up.

| Broken | Intended |
|--------|----------|
| image | image |

### DIFF
--- a/docs/src/SUMMARY.md
+++ b/docs/src/SUMMARY.md
@@ -3,7 +3,7 @@
 [Introduction](introduction.md)
 [Core Concepts](concepts.md)
 
-## Rust Development for UEFI
+# Rust Development for UEFI
 
 - [Best Practices](dev/principles.md)
   - [Abstractions](dev/principles/abstractions.md)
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-workflow.yml
makubacki, line 1: These GitHub ref changes are ultimately coming in through another PR, right?

Javagedes: Yes it was just testing. I'll drop the commit

Javagedes: I'm still hoping we don't need those changes.

makubacki: Me too :)

Javagedes: Dropped!

Javagedes:
close in favor of #104

PR #87, opened 2024-10-09
Makefile: Add build-aarch64, build-x64

Javagedes:
At some point in time, the `build-arch` cargo command evolved into building the dxe_core Qemu efi images when it was supposed to be building all crates with the specific uefi target triple.

This commit adds build-aarch64 and build-x64 commands to the Makefile to build all crates with the specific uefi target triple as expected and matches the format in uefi-core. building the qemu binaries has been renamed to `build-bin`, however this will eventually be dropped when the binaries move to their own location outside the crate.

This is also used to be able to merge cargo-make expectations between repos such that the workflows can be re-used.

This removes the building of the qemu binaries from the pr gates. They will be added back in when we add in downstream platform CI testing

### DIFF
--- a/.github/workflows/CiWorkflow.yml
+++ b/.github/workflows/CiWorkflow.yml
@@ -1,0 +1,81 @@
+# @file CiWorkflow.yml
+#
+# A reusable CI workflow that runs basic CI tests and Rust CI tests.
+#
+##
+# Copyright (c) Microsoft Corporation.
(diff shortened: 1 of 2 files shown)


felipebalbi:
Why is the first commit as part of this PR? Looks unrelated.

Javagedes:
Closing in favor of #99

PR #92, opened 2024-10-11
ci-workflow.yml: Checkout pull request branch

Javagedes:


.github/workflows/ci-workflow.yml
```diff
@@ -16,6 +16,11 @@ jobs:
     steps:
       - name: ✅ Checkout Repository ✅
         uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.pull_request.head.ref }} # The source branch of the pull request
+          repository: ${{ github.event.pull_request.head.repo.full_name }} # The repository of the pull request source branch
```
felipebalbi, line 21: Aren't these already the defaults?

.github/workflows/ci-workflow.yml
```diff
@@ -16,6 +16,11 @@ jobs:
     steps:
       - name: ✅ Checkout Repository ✅
         uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.pull_request.head.ref }} # The source branch of the pull request
+          repository: ${{ github.event.pull_request.head.repo.full_name }} # The repository of the pull request source branch
+          token: ${{ secrets.GITHUB_TOKEN }}
+          fetch-depth: 0 # Fetch all history for all branches
```
felipebalbi, line 23: why is this needed? Won't it just make the workflow slower? Do you really need all the history for all branches and tags? Why isn't the default of `1` enough?

Library/UefiDepexLib/src/lib.rs
```diff
@@ -323,7 +323,7 @@ impl Depex {
         }
     }
 
-    /// inidicates that this is a "schedule on request" depex.
+    /// indicates that this is a "schedule on request" depex.
```
felipebalbi, line 326: technically not related to this PR

Library/UefiEventLib/src/lib.rs
```diff
@@ -1247,7 +1247,7 @@ mod tests {
         //signal an ungrouped event
         SPIN_LOCKED_EVENT_DB.signal_event(ungrouped_events.pop().unwrap()).unwrap();
 
-        //all other events should remain unsignalled
+        //all other events should remain un-signaled
```
felipebalbi, line 1250: technically not related to this PR

Library/UefiProtocolDbLib/src/lib.rs
```diff
@@ -1968,8 +1968,8 @@ mod tests {
 
         let (handle, _) = SPIN_LOCKED_PROTOCOL_DB.install_protocol_interface(None, guid1, interface1).unwrap();
 
-        let returned_iface = SPIN_LOCKED_PROTOCOL_DB.get_interface_for_handle(handle, guid1).unwrap();
-        assert_eq!(interface1, returned_iface);
+        let returned_interface = SPIN_LOCKED_PROTOCOL_DB.get_interface_for_handle(handle, guid1).unwrap();
+        assert_eq!(interface1, returned_interface);
```
felipebalbi, line 1972: technically not related to this PR

README.md
```diff
@@ -66,7 +66,7 @@ The following instructions install Rust.
 1. Construction of a bare-metal "kernel" to dispatch from `DxeIpl`.
    1. Built in a basic build environment for no-std.
    2. Uses a basic output subsystem (likely legacy UART, but maybe VGA if it works in QEMU before GOP starts it).
-   3. Integratable into a UEFI build as a replacement for `DxeMain` with observable debug output.
+   3. Integrateable into a UEFI build as a replacement for `DxeMain` with observable debug output.
```
felipebalbi, line 69: technically not related to this PR

README.md
```diff
@@ -116,7 +116,7 @@ firmware environment.
 - Host (Standard Library) DXE Core Build: `cargo make build-std`
   - Simpler Alias Command: `cargo make build`
 
-#### Running the Host Exectuable
+#### Running the Host Executable
```
felipebalbi, line 119: technically not related to this PR

cspell.yml
```diff
@@ -24,12 +24,22 @@ words:
...
   - "longjmp"
   - "mdbook"
+  - "pread"
+  - "pwrite"
+  - "reloc"
+  - "rfind"
   - "rustc"
   - "structs"
+  - "uncacheable"
+  - "wakeup"
```
felipebalbi, line 43: technically not related to this PR

dxe_core/src/protocols.rs
```diff
@@ -664,10 +664,10 @@ extern "efiapi" fn locate_protocol(
 
     if !registration.is_null() {
         if let Some(handle) = PROTOCOL_DB.next_handle_for_registration(registration) {
-            let iface = PROTOCOL_DB
+            let i_face = PROTOCOL_DB
                 .get_interface_for_handle(handle, unsafe { *protocol })
                 .expect("Protocol should exist on handle if it is returned for registration key.");
-            unsafe { interface.write(iface) };
+            unsafe { interface.write(i_face) };
         } else {
```
felipebalbi, line 671: technically not related to this PR

dxe_core/src/protocols.rs
```diff
@@ -677,7 +677,7 @@ extern "efiapi" fn locate_protocol(
                 unsafe { interface.write(core::ptr::null_mut()) };
                 return err;
             }
-            Ok(iface) => unsafe { interface.write(iface) },
+            Ok(i_face) => unsafe { interface.write(i_face) },
```
felipebalbi, line 680: technically not related to this PR

dxe_core/src/protocols.rs
```diff
@@ -755,7 +755,7 @@ pub fn init_protocol_support(bs: &mut efi::BootServices) {
     PROTOCOL_DB.init_protocol_db();
 
     //This bit of trickery is needed because r_efi definition of (Un)InstallMultipleProtocolInterfaces
-    //is not variadic, due to rust only supporting variadics for "unsafe extern C" and not "efiapi"
+    //is not variadic, due to rust only supporting variadic for "unsafe extern C" and not "efiapi"
```
felipebalbi, line 758: technically not related to this PR

.github/workflows/ci-workflow.yml
Javagedes: As it stands, the setting to allow workflows to execute on PRs coming from a branch is disabled at the org level, so our workaround is to use `pull_request_target` as a trigger instead of `pull_request`. When switching to `pull_request_target`, the default checkout branch is the target branch (i.e. main) instead of the pull request branch. Due to this, we have to override.

.github/workflows/ci-workflow.yml
Javagedes: You are right, 1 would be more appropriate.

Library/UefiDepexLib/src/lib.rs
Javagedes: I have this as a separate commit and plan on doing a rebase & ff. I can make them completely separate pull requests if you like.

.github/workflows/ci-workflow.yml
felipebalbi: that's not true (or shouldn't be). What was disabled (or should have been) is running the workflows on the forks themselves. That is, we want pop-project/uefi-dxe-core to run CI worflows, but felipebalbi/uefi-dxe-core doesn't need to run them.

@jeffglaum, matches your understanding?

Library/UefiDepexLib/src/lib.rs
felipebalbi: I think it would be preferrable. That being said, no strong feelings, really. Just noticed that it didn't fit the PR description.

.github/workflows/ci-workflow.yml
Javagedes: Negative, This setting at the org level prevents pull requests in `pop-project/uefi-dxe-core` whose branch comes from a fork from triggering. See an example here (it just sits and never triggers):

https://github.com/pop-project/uefi-dxe-core/pull/87
image

Your scenario is already covered with github in general. Running a workflow in a fork itself uses all of the forks permissions. This can be seen in the fact that dxe_core cannot build on a fork because it does not have access to pop-project and thus you get this build error in the workflow coming from the fork (because this corosensei repo is in pop-project and is private):

https://github.com/Javagedes/uefi-dxe-core/actions/runs/11261632696/job/31315521367?pr=6
image

felipebalbi: Looks like the default permissions should be reviewed. I'd say we definitely want to run CI on PRs

makubacki: My understanding (we don't normally work on internal/private repos) is this setting simply needs be enabled at the org level so it can be enabled in this repository:
image

As Joey mentioned, by using `pull_request_target` we are running the workflow file on the target branch (i.e. `main`) as opposed to from the PR.

Library/UefiDepexLib/src/lib.rs
makubacki: It doesn't really matter (Rebase & FF is fine). I'm going to resquash changes before the repo is opened for wider development anyway.

.github/workflows/ci-workflow.yml
jeffglaum: Yes, looking into why this option is grayed out at the org level.

PR #95, opened 2024-10-11
dxe_core: support non-static instanced components

Javagedes:
dxe_core local driver execution required the component be `&dyn DxeComponent + 'static` as we were using `Coroutine` from the the corosensei crate, which was a re-definition of the `ScopedCoroutine` type with a `static` lifetime bound. By switching to the `ScopedCoroutine` for executing local drivers, we remove this static requirement and allow for non-static instanced components to be executed.


Javagedes:
Close in favor of #105

PR #105, opened 2024-10-15
dxe_core: support non-static instanced components

Javagedes:
## Description

dxe_core local driver execution required the component be `&dyn DxeComponent + 'static` as we were using `Coroutine` from the the corosensei crate, which was a re-definition of the `ScopedCoroutine` type with a `static` lifetime bound. By switching to the `ScopedCoroutine` for executing local drivers, we remove this static requirement and allow for non-static instanced components to be executed.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Repo CI

## Integration Instructions

Any currently static components can be moved into the `_start` function for a dxe core binary.

### DIFF
--- a/dxe_core/src/image.rs
+++ b/dxe_core/src/image.rs
@@ -34,7 +34,7 @@ use crate::{
 
 use corosensei::{
     stack::{Stack, StackPointer, MIN_STACK_SIZE, STACK_ALIGNMENT},
-    Coroutine, CoroutineResult, Yielder,
+    Coroutine, CoroutineResult, ScopedCoroutine, Yielder,
 };
 
 pub const EFI_IMAGE_SUBSYSTEM_EFI_APPLICATION: u16 = 10;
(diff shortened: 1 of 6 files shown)


dxe_core/src/image.rs
```diff
@@ -730,14 +730,16 @@ extern "efiapi" fn start_image(
     }
 }
 
-pub fn core_start_local_image(component: &'static dyn DxeComponent) -> Result<(), efi::Status> {
+pub fn core_start_local_image(component: &dyn DxeComponent) -> Result<(), efi::Status> {
     let stack = ImageStack::new(ENTRY_POINT_STACK_SIZE);
 
-    let mut coroutine =
-        Coroutine::with_stack(stack, move |_: &Yielder<&dyn DxeComponent, crate::error::Result<()>>, component| {
+    let mut coroutine = ScopedCoroutine::with_stack(
```
joschock, line 736: FWIW, Corosensei upstream looks like they removed this for unsoundness reasons: https://github.com/Amanieu/corosensei/pull/28

Javagedes: Can we rebase your corosensei changes on top of tree? I can look at using the changes they made.

joschock: Yes, rebasing is fine. The main change I made was to create a uefi calling convention/stack that was aligned with the UEFI ABI (which is windows, but no TEB).

joschock: So that shouldn't interact badly with the upstream (I hope).

Javagedes: Sounds good. I'll give that a go tomorrow the . I'm honestly not concerned about unsoundness, as the drivers live longer than dxe_core's start() function's finish, but if there is a more sound way to do it then I'd like to do it.

Javagedes:
dropping in favor of #114 as the updated interface in corosensei still requires them to be static

PR #111, opened 2024-10-16
[Rebase&FF] Crate publishing preparation

Javagedes:
## Description

This PR contains two commits that begin to prepare the repository for publishing crates. The first commit adds necessary metadata and files in each crate (readme, description, license, repository) and adds the version field to each of our local crates. This is necessary because when you publish, any `path = ...` in the dependencies are dropped, which leaves the version to link against for crates.io.

The second change relaxes the crate dependency versions to make the dependency resolver's job easier and dropped the Cargo.lock since this is technically a library crate, not a binary crate. Due to this, however, I've updated the workflows to upload the Cargo.lock that is generated during build, so that we can easily reproduce a given build when necessary.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Verified Q35 continues to boot to UEFI Shell.

## Integration Instructions

N/A


crates/uefi_pecoff/README.md
```diff
@@ -0,0 +1,3 @@
+# UEFI PeCOFF
+
+A PeCOFF parsing implementation for the dxe_core.
```
makubacki: Unless there is a specific reason for this abbreviation, I recommend treating them as two separate formats for the general purpose of documentation.

```suggestion
# UEFI PE/COFF

A PE/COFF parsing implementation for the dxe_core.

```

crates/uefi_protocol_db/Cargo.toml
```diff
@@ -1,7 +1,11 @@
 [package]
 name = "uefi_protocol_db"
-version = "0.1.0"
+version = "0.1.8"
 edition = "2021"
+readme = "README.md"
+description = "An Protocol Database implementation for the dxe_core."
```
makubacki:
```suggestion
description = "A Protocol Database implementation for the dxe_core."
```

crates/uefi_protocol_db/README.md
```diff
@@ -0,0 +1,3 @@
+# UEFI Protocol DB
+
+An Protocol Database implementation for the dxe_core.
```
makubacki:
```suggestion
A Protocol Database implementation for the dxe_core.
```

Javagedes: Fixed. Thanks!

crates/uefi_protocol_db/Cargo.toml
Javagedes: Fixed. THanks!

crates/uefi_pecoff/README.md
Javagedes: Fixed. Thanks!

PR #114, opened 2024-10-17
dxe_core: Improve DxeCore public interface

Javagedes:
## Description

This commit improves the DxeCore public interface in many ways:

1. Splits the core into two distinct parts: pre-init and post-init mainly for controlling the types of configuration you can do (no allocations pre-init and allocations post-init).

2. The interface for the post-init DxeCore is locked down and can only be accessed by calling `initialize` on the pre-init DxeCore.

3. Improves the interface by no longer requiring drivers to be static. This is accomplished by holding the driver in a Box and leaking it, effectively making it static, which is required by corosensei. This is done in a safe way as we consume the drivers, thus no driver can be leaked more than once.

4. Sets the example on how to do configuration for the DXE Core and for the drivers. That is to say that `default` must be implemented, then you can override the default configurations by calling the `with_<config>` methods.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [x] Includes documentation?

## How This Was Tested

Ensured Q35 dxe core examples continue to compile and run to the UEFI Shell.

## Integration Instructions

Follow the examples in the documentation and in the bin/* folders to see how to use the new interface.

```rust
    log::set_logger(&LOGGER).map(|()| log::set_max_level(log::LevelFilter::Trace)).unwrap();
    let adv_logger_component = AdvancedLoggerComponent::<serial_writer::Uart16550>::new(&LOGGER);
    adv_logger_component.init_advanced_logger(physical_hob_list).unwrap();

    Core::default()
        .with_cpu_initializer(uefi_cpu_init::X64CpuInitializer::default())
        .with_section_extractor(section_extractor::CompositeSectionExtractor::default())
        // Add any config knob functions for pre-gcd-init Core
        // .with_some_config(true)
        .initialize(physical_hob_list) // We can make allocations now!
        // Add any config knob functions for post-gcd-init Core
        // .with_some_config(true)
        .with_driver(Box::new(adv_logger_component))
        .with_driver(Box::new(HelloComponent::default()))
        .with_driver(Box::new(HelloComponent::default().with_name("Dxe Core")))
        .with_driver(Box::new(HelloComponent::default().with_name("World")))
        .start()
        .unwrap();
```

### DIFF
--- a/dxe_core/src/lib.rs
+++ b/dxe_core/src/lib.rs
@@ -1,5 +1,32 @@
 //! DXE Core
 //!
+//! A pure rust implementation of the UEFI DXE Core. Please review the getting started documentation at
+//! <https://pop-project.github.io/uefi-dxe-core/> for more information.
+//!
+//! ## Examples
+//!
+//! ``` rust,ignore
+//! #[no_std]
+//! #[no_main]
+//! #[panic_handler]
+//! fn panic(info: &PanicInfo) -> ! {
+//!    log::error!("{}", info);
+//!    loop {}
+//! }
+//! #[cfg_attr(target_os = "uefi", export_name = "efi_main")]
+//! pub extern "efiapi" fn _start(physical_hob_list: *const c_void) -> ! {
+//!
+//!     dxe_core::Core::default()
+//!         .with_cpu_initializer(CpuInit::default())
+//!         .with_section_extractor(SectionExtract::default())
+//!         .initialize(physical_hob_list)
+//!         .with_driver(Box::new(Driver::default()))
+//!         .start()
+//!         .unwrap();
+//!     loop {}
+//! }
+//! ```
+//!
 //! ## License
 //!
 //! Copyright (C) Microsoft Corporation. All rights reserved.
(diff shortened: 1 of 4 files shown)


dxe_core/src/lib.rs
```diff
@@ -36,41 +36,50 @@ pub mod test_support;
...
 #[derive(Default)]
 pub struct Core<CpuInitializer, SectionExtractor>
 where
     CpuInitializer: interface::CpuInitializer + Default,
     SectionExtractor: fw_fs::SectionExtractor + Default + Copy + 'static,
 {
-    pub cpu: CpuInitializer,
-    pub se: SectionExtractor,
+    cpu: CpuInitializer,
+    se: SectionExtractor,
```
joschock: I regret naming it this: consider changing it to `section_extractor`?

makubacki:
@Javagedes, this looks good to me, and I'd prefer to get it in as part of the squash.

Javagedes:
> @Javagedes, this looks good to me, and I'd prefer to get it in as part of the squash.

@makubacki  Sounds good I'm finishing up the documentation and will get it merged in soon.

dxe_core/src/lib.rs
Javagedes: updated to `cpu_initializer` and `section_extractor`

PR #117, opened 2024-10-19
dxe_core: Add EFI_MEMORY_MAP trace logging

os-d:
## Description

This adds a new target for logging: efi_memory_map. When set to debug level for a platform, the EFI_MEMORY_MAP will be pretty printed in the log whenever it is created. Info logging was chosen over trace because trace logging prints the file name before each line, which takes up too much room for what this table needs to be easily read and it is distracting with so many lines.

This commits also changes the MAT logic to use the memory descriptor pretty print logic created in this commit, though it does not tie it to the efi_memory_map logging target or any other, as the MAT is created less often and has fewer members and is useful to be in the log by default.

The column sizes and individual column member spacing was tested and adjusted so it lined up visually, even though the numbers in the format strings appear off.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested with printing enabled/disabled for target and seeing both memory map and MAT and only MAT prints.

## Integration Instructions

In the platform's dxe core binary wrapper, set

```rust
uefi_logger::Format::Standard,
    &[
        ("efi_memory_map", log::LevelFilter::Debug),
    ],
```
to enable printing the EFI_MEMORY_MAP.

### DIFF
--- a/dxe_core/src/allocator.rs
+++ b/dxe_core/src/allocator.rs
@@ -8,6 +8,7 @@
 //!
 use core::{
     ffi::c_void,
+    fmt::Debug,
     mem,
     slice::{self, from_raw_parts_mut},
 };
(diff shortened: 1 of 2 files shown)


dxe_core/src/allocator.rs
```diff
@@ -495,6 +610,8 @@ extern "efiapi" fn get_memory_map(
         }
     }
 
+    log::info!(target: "efi_memory_map", "EFI_MEMORY_MAP: \n{:?}", MemoryDescriptorSlice(&efi_descriptors));
```
os-d: I think I need to wrap this in a log_enabled!() call, since I think Rust would run the impl debug to generate the string and then throw it away when it sees the target isn’t enabled? Or will it, through the magic of macros, not do the computations and just skip past if the target isn’t enabled?

Javagedes: I *think* the display / debug function calls do not happen until the `format_args!` call inside the [`self.format.write()`](https://github.com/pop-project/uefi-core/blob/47bc6b47c77391c1a76d0627899ae6945d425567/uefi_logger/src/serial_logger.rs#L60) of the logger implementation, which happens after the `enabled` evaluation, but I'm not 100% confident in that. You could run it through a debugger and follow the call stack. Or just panic and dump the call stack.

However I'm not to worried about this. The log macros can be completely compiled out of the program by using the `log` crate's `max_level_*` and `release_max_level_*` feature flags [here](https://docs.rs/crate/log/0.4.22/features).

Javagedes:
`Debug` level seems like the right level if the only reason you are not going with `trace` is because of the line information. You're almost going from the lowest to the highest level by jumping from `trace` to `log`.

dxe_core/src/allocator.rs
```diff
@@ -74,6 +75,120 @@ static STATIC_ALLOCATORS: &[&UefiAllocator] = &[
...
+
+    if string_len + attrs.len() > 20 || attrs.is_empty() {
+        writeln!(f, "{:<#20X}", attributes)?;
+        return Ok(());
+    }
+
+    writeln!(f, "{:<20}", attrs.join("|"))
+}
+
+fn memory_type_to_str(f: &mut core::fmt::Formatter<'_>, memory_type: efi::MemoryType) -> core::fmt::Result {
```
Javagedes, line 147: This makes me wonder if, in r_efi, it would be better to create a wrapper class with a `repr(transparent)` instead of doing `type MemoryType = u32`. The [`repr(transparent)`](https://doc.rust-lang.org/nomicon/other-reprs.html#reprtransparent) ensures the layout and ABI of it is the same as the u32. This then will allow us to implement traits (such as Debug / Dispaly in this case) on the type.

``` rust
#[repr(transparent)]
struct MemoryType(u32);

impl Deref for MemoryType {
    type Target = u32;

    fn deref(&self) -> &Self::Target {
        &self.0
    }
}

impl DerefMut for MemoryType {
    fn deref_mut(&mut self) -> &mut Self::Target {
        &mut self.0
    }
}

impl std::fmt::Debug for MemoryType {
    fn fmt(&self, f: &mut Formatter<'_>) -> Result<(), Error> {
        todo!()
    }
}
```

dxe_core/src/allocator.rs
```diff
@@ -74,6 +75,120 @@ static STATIC_ALLOCATORS: &[&UefiAllocator] = &[
...
+        efi::PERSISTENT_MEMORY => "Persistent Memory",
+        _ => "Unknown Memory Type",
+    };
+
+    write!(f, "{:<25}", string)
+}
+
+pub struct MemoryDescriptorSlice<'a>(pub &'a [efi::MemoryDescriptor]);
+
+impl<'a> Debug for MemoryDescriptorSlice<'a> {
```
Javagedes, line 183: Memory descriptor is a real type in r-efi [MemoryDescriptor](https://github.com/r-efi/r-efi/blob/1c1a4062a0298cf1cd6f0ed6b7f9ffbde0f82c73/src/system.rs#L375-L376). Seems like you should just implement Debug there with this logic (the second half). It's a much better debug print then what rust is auto-deriving. if you want to keep this MemoryDescriptorSlice type thats fine, but you could just do `for descriptor in self.0 { <call_debug_function> }`

dxe_core/src/allocator.rs
os-d: See below comment

dxe_core/src/allocator.rs
os-d: I agree. I didn't want to have a philosophical fight there about the purity of having pretty prints or not :). This was a request from @makubacki to help with a debug and so I wanted to get it available in our codebase first and then take it to r-efi, where it may or may not got shot down. If it does get shot down, we can move it there if we ever fork off r-efi.

dxe_core/src/allocator.rs
os-d: Gotcha. I think there is still a balance where you have logging enabled for some targets but not all, but don't want to spend the extra time to compute logs we are going to throw away (for physical Intel platforms there are a lot of entries here to go through and we do it quite a few times near the end of boot). So I'll take a look at the call stack and see (probably easiest is to disable the target and just put a panic in the impl Debug and see if it triggers.

os-d:
> `Debug` level seems like the right level if the only reason you are not going with `trace` is because of the line information. You're almost going from the lowest to the highest level by jumping from `trace` to `log`.

Gotcha, I don't know the different log levels we have here, I'll switch to Debug.

dxe_core/src/allocator.rs
joschock: You can always use a newtype to provide a debug impl for any type from another crate:

```
struct PrettyDescriptor(efi::MemoryDescriptor);

impl core::fmt for PrettyDescriptor {
    fn fmt(&self, f: &mut Formatter<'_>) -> Result<(), Error> {
        todo!()
    }
}

log::info!("Pretty Descriptor: {:x?}", PrettyDescriptor(my_descriptor));
```

That has the benefit of being almost a straight port if it gets upstreamed to the original type.

os-d: Sure, I had done a slice to get the headers, but I do see how getting just the type itself is also useful and propogates to other areas. I can refactor this to still have the slice to print headers, but move the MemoryDescriptor logic to it’s own impl Debug

dxe_core/src/allocator.rs
os-d: I did test this with throwing a panic in the top of the impl Debug and it did not panic when I had the target disabled and did panic when the target was enabled, so looks like we are already smart enough to skip the format string calculations if the target is disabled.

PR #120, opened 2024-10-22
[REBASE&FF] Add uefi_test crate

Javagedes:
## Description

Adds a new crate `uefi_test` This crate utilizes the `linkme` crate to do some linker magic similar to google tests that places all tests into a link_section, and is collated at runtime. This crate comes with a single feature flag, `off`, which updates internal the logic to no longer generate the static test cases or the link sections themselves. This is different than other skipping functionality, which is described below, which simply skips the test at runtime.

**note:** The second commit is necessary due to this raised issue: https://github.com/xd009642/tarpaulin/issues/1642. test executables generated for proc-macro crates are exiting before any tests run on windows-latest runners of github actions. I confirmed through outside testing that this is not something specific to our repo. Due to this, I've update cargo make to skip testing the `uefi_test_macro` crate when on windows and for a CI system. We will still hit coverage targets as the crate will be tested on the linux runner, and it is not skipped locally.

## `uefi_test` crate

1. Macro for globally registering tests with the `uefi_test` attribute
4. `assert` macros that return an error instead of panic because we cannot unwind a panic in no_std
5. A `TestRunnerComponent` that runs all collected tests.

## `uefi_test` macro

`uefi_test` is an attribute macro, e.g. `#[uefi_test]` that works exactly like the `test` attribute, except with the expected function ``fn(&dyn DxeComponentInterface)->core::result::Result<(), 'static str>`` interface. The `uefi_test` attribute includes two additional attributes for configuration.

The first is `should_fail` which matches the same logic as `should_panic`, which comes in two flavors:
`#[should_fail]` and `#[should_fail = "msg"]` where `msg` is the expected error string. The second attribute is `#[skip]`, which is functionaly the same as the `skip` attribute for std tests. This is expected to be used in conjunction with `cfg_attr` for tests that may only run under certain circumstances (such as x64 targets, etc). It should be noted that the `skip` attribute skips at runtime, and the test is still compiled in.

## `TestRunnerComponent`

The `TestRunnerComponent` is a component that can be registered with the pure rust dxe core and once the core executes it, will execute all tests. As mentioned previously, all tests are automatically registered, so there is configuration available to skip tests. The current available configuration is as follows:

* `with_filter(&str)` which will **run** any test that matches this filter (instead of running all by default)
* `debug_mode(bool)` which will turn on logging for tests (by default we disable logging for each test)
* `fail_fast(bool)` which will end the test immediately if a test fails.

## Examples

``` rust
use uefi_test::{uefi_test, TestRunnerComponent, Result};
use uefi_component_interface::DxeComponentInterface;

let component = TestRunnerComponent::default()
  .with_filter("aarch64")
  .debug_mode(true)
  .fail_fast(true);

#[cfg_attr(target_arch = "aarch64", uefi_test)]
fn test_case(_interface: &dyn DxeComponentInterface) -> Result {
  uefi_test::assert_eq!(1, 1);
  Ok(())
}

#[uefi_test]
fn test_case2(_interface: &dyn DxeComponentInterface) -> Result {
  uefi_test::assert_eq!(1, 1);
  Ok(())
}

#[uefi_test]
#[should_fail]
fn failing_test_case(_interface: &dyn DxeComponentInterface) -> Result {
   uefi_test::assert_eq!(1, 2);
   Ok(())
}

#[uefi_test]
#[should_fail = "This test failed"]
fn failing_test_case_with_msg(_interface: &dyn DxeComponentInterface) -> Result {
  uefi_test::assert_eq!(1, 2, "This test failed");
  Ok(())
}

#[uefi_test]
#[skip]
fn skipped_test_case(_interface: &dyn DxeComponentInterface) -> Result {
   todo!()
}

#[uefi_test]
#[cfg_attr(not(target_arch = "x86_64"), skip)]
fn x86_64_only_test_case(_interface: &dyn DxeComponentInterface) -> Result {
  todo!()
}
```
- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [x] Includes documentation?

## How This Was Tested

Validated that all scenarios work as expected.

## Integration Instructions

Take a dependency on `uefi_test` for any crate that wants to have runtime tests applied. In these scenarios, write the test and annotate it with `#[uefi_test]`

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -6,7 +6,9 @@ RUSTC_BOOTSTRAP = 1
 NO_STD_FLAGS = "--profile ${RUSTC_PROFILE} -Zbuild-std=core,compiler_builtins,alloc -Zbuild-std-features=compiler-builtins-mem -Zunstable-options --timings=html"
 STD_FLAGS = "--profile ${RUSTC_PROFILE} --features std"
 TEST_FLAGS = { value = "", condition = { env_not_set = ["TEST_FLAGS"] } }
-COV_FLAGS = { value = "--out html --out xml --exclude-files **/tests/* --exclude-files **/benches/*", condition = { env_not_set = ["COV_FLAGS"] } }
+# TODO: Remove when https://github.com/xd009642/tarpaulin/issues/1642 is resolved.
+COV_FILTER = { value = "-e uefi_test_macro", condition = { platforms = ["windows"], env_true = ["CARGO_MAKE_CI"] } }
+COV_FLAGS = { value = "--workspace --out html --out xml --exclude-files **/tests/* --exclude-files **/benches/*", condition = { env_not_set = ["COV_FLAGS"] } }
 
 [env.development]
 RUSTC_PROFILE = "dev"
@@ -120,7 +122,7 @@ dependencies = ["individual-package-targets"]
 description = "Build and run all tests and calculate coverage."
 clear = true
 command = "cargo"
-args = ["tarpaulin", "@@split(COV_FLAGS, )", "--output-dir", "${CARGO_MAKE_WORKSPACE_WORKING_DIRECTORY}/target", "-e", "uefi_test_macro"]
+args = ["tarpaulin", "@@split(COV_FLAGS, )", "@@split(COV_FILTER, )", "--output-dir", "${CARGO_MAKE_WORKSPACE_WORKING_DIRECTORY}/target"]
 dependencies = ["individual-package-targets"]
 
 [tasks.coverage-filter]


uefi_test_macro/src/lib.rs
```diff
@@ -0,0 +1,96 @@
+// In your proc-macro crate (e.g., `my_macro`)
```
joschock: not sure what this comment means...

uefi_test/src/lib.rs
```diff
@@ -0,0 +1,243 @@
...
+//! - `off`: Will not compile any tests.
+//!
+//! ## Example
+//!
+//! ```rust
+//! use uefi_test::{uefi_test, TestRunnerComponent, Result};
+//! use uefi_component_interface::DxeComponentInterface;
+//!
+//! let component = TestRunnerComponent::default()
+//!   .with_filter("aarch64")
```
joschock: Might be nice to clarify here for the casual reader; it could be ambiguous about whether `with_filter` is inclusive or exclusive of tests (though that is well documented on the actual function).
```suggestion
//!   .with_filter("arrach64") //skip test with "aarch64" in the name
```

uefi_test_macro/src/lib.rs
Javagedes: Boilerplate I forgot do remove. I've removed it and added documentation :)

uefi_test/src/lib.rs
Javagedes: Added some clarity as suggested. Also added more clarity in the function call, that the `name` is the function name and the module path.

crates/uefi_test_macro/README.md
```diff
@@ -0,0 +1,3 @@
+# Uefi Test Macro
```
makubacki: For consistency.

```suggestion
# UEFI Test Macro
```

Javagedes: Updated. Thanks!

PR #122, opened 2024-10-22
[REBASE & FF] allocator: Add ensure_capacity API

os-d:
## Description

This PR contains two commits:

Use Correct Target Triple in README
---
The README was incorrectly specifying that the aarch64 build commands would build the i686-unknown-uefi target triple, so this was updated to aarch64-unknown-uefi.

Add ensure_capacity API
--
This commit adds a new ensure_capacity API to the allocator
    that checks if a given size and alignment are free in the
    given allocator. If not, it calls expand to allocate more
    pages in the pool.

    This API is intended for dxe_core only when it needs to
    pre-allocate memory in memory sensitive operations, such as
    constructing the EFI_MEMORY_MAP or allocating page table
    memory.

    By ensuring the pool allocator has enough memory for its
    allocations, it can ensure that recursive calls to the GCD
    will not be made.

    The get_memory_map code is updated here also, not to use
    the new API as was the original design point, but to
    pre-allocate the merged descriptor vector. Same goal,
    but different implementation.

    This API will be consumed when mu-paging is integrated
    into dxe_core.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested by booting to Windows in Q35 and adding some debug prints to witness the `ensure_capacity()` succeeding. Then I tested an artificial `allocate_pool()` allocating the available memory and witnessing Windows booting after `ensure_capacity()` called `expand()`.

## Integration Instructions

N/A.

### DIFF
--- a/crates/uefi_allocator/src/fixed_size_block_allocator.rs
+++ b/crates/uefi_allocator/src/fixed_size_block_allocator.rs
@@ -26,10 +26,12 @@ use r_efi::efi;
 use uefi_gcd::gcd::SpinLockedGcd;
 
 /// Type for describing errors that this implementation can produce.
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
 pub enum FixedSizeBlockAllocatorError {
     /// Could not satisfy allocation request, and expansion failed.
     OutOfMemory,
+    /// Invalid input parameter
+    InvalidParameter,
 }
 
 /// Minimum expansion size - allocator will request at least this much memory
@@ -37,6 +39,7 @@ pub enum FixedSizeBlockAllocatorError {
 pub const MIN_EXPANSION: usize = 0x100000;
 const ALIGNMENT_BITS: usize = 12;
 const ALIGNMENT: usize = 0x1000;
+const DEFAULT_POOL_ALIGNMENT: usize = 0x8;
 
 const BLOCK_SIZES: &[usize] = &[8, 16, 32, 64, 128, 256, 512, 1024, 2048, 4096];
 
@@ -146,6 +149,57 @@ impl FixedSizeBlockAllocator {
             preferred_range: None,
             page_change_callback,
         }
+    }
+
+    /// Ensures that the allocator has enough capacity to satisfy an allocation of the given size and alignment.
+    /// If the allocator does not have enough capacity, it will request more memory from the GCD.
+    /// If the allocator has enough capacity, it will return Ok(()).
+    /// If the allocator cannot allocate more memory, it will return Err(FixedSizeBlockAllocatorError::OutOfMemory).
+    /// If the size or alignment are invalid, it will return Err(FixedSizeBlockAllocatorError::InvalidParameter).
+    /// This function is useful for pre-allocating memory before it is needed for memory sensitive operations, such
+    /// as creating the EFI_MEMORY_MAP or allocating page table memory before changing attributes.
+    pub fn ensure_capacity(&mut self, size: usize, align: usize) -> Result<(), FixedSizeBlockAllocatorError> {
+        let mut alignment = align;
+
+        // next power of two will return alignment here, if already a power of two, or round up to the next power of 2
+        // if overflow occurs, it will return 0
+        alignment = alignment.next_power_of_two();
+
+        if alignment == 0 {
+            return Err(FixedSizeBlockAllocatorError::InvalidParameter);
+        }
+
+        if alignment < DEFAULT_POOL_ALIGNMENT {
+            alignment = DEFAULT_POOL_ALIGNMENT;
+        }
+
+        // let's satisfy the preconditions of constructing a Layout. If we end up overallocating, that's okay, we
+        // probably will anyway when we expand
+        // we could still fail here if align(size, alignment) would cause an overflow. We'll let from_size_align
+        // determine that for us and return InvalidParameter
+        let layout =
+            Layout::from_size_align(size, alignment).map_err(|_| FixedSizeBlockAllocatorError::InvalidParameter)?;
+
+        for node in AllocatorIterator::new(self.allocators) {
+            let allocator = unsafe { &mut (*node).allocator };
+            match allocator.allocate_first_fit(layout) {
+                Ok(ptr) => {
+                    // we have found enough free space in this heap, deallocate it and return
+                    unsafe { allocator.deallocate(ptr, layout) };
+                    return Ok(());
+                }
+                Err(_) => {
+                    // try next
+                    continue;
+                }
+            }
+        }
+
+        // if we hit here, we haven't found a large enough free region, so we need to expand this allocator's
+        // free mem
+        self.expand(layout)?;
+
+        Ok(())
     }
 
     // Expand the memory available to this allocator by requesting a new contiguous region of memory from the gcd setting
@@ -857,6 +911,20 @@ impl SpinLockedFixedSizeBlockAllocator {
         self.lock().reserve_memory_pages(pages)
     }
 
+    /// Ensures that the allocator has enough capacity to allocate a block of the given size and alignment.
+    /// If the allocator does not have enough capacity, it will attempt to expand the allocator to accommodate the
+    /// requested block.
+    /// Returns an error if the allocator cannot be expanded.
+    /// ## Errors
+    /// - `efi::Status::OUT_OF_RESOURCES`: The allocator cannot be expanded to accommodate the requested block.
+    /// - `efi::Status::INVALID_PARAMETER`: The requested size or alignment is invalid
+    pub fn ensure_capacity(&self, size: usize, align: usize) -> Result<(), efi::Status> {
+        self.lock().ensure_capacity(size, align).map_err(|err| match err {
+            FixedSizeBlockAllocatorError::OutOfMemory => efi::Status::OUT_OF_RESOURCES,
+            FixedSizeBlockAllocatorError::InvalidParameter => efi::Status::INVALID_PARAMETER,
+        })
+    }
+
     /// Returns the allocator handle associated with this allocator.
     pub fn handle(&self) -> efi::Handle {
         self.inner.lock().handle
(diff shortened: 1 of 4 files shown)


crates/uefi_allocator/src/fixed_size_block_allocator.rs
```diff
@@ -148,6 +151,49 @@ impl FixedSizeBlockAllocator {
...
+            return Err(FixedSizeBlockAllocatorError::InvalidParameter);
+        }
+
+        if alignment < DEFAULT_POOL_ALIGNMENT {
+            alignment = DEFAULT_POOL_ALIGNMENT;
+        }
+
+        for node in AllocatorIterator::new(self.allocators) {
+            let allocator = unsafe { &mut (*node).allocator };
+            if align_up_size(size, alignment) <= allocator.free() {
```
joschock: I'm not certain about this, but my recollection is that `allocator.free()` returns the aggregate sum of free space in the given heap, but that doesn't necessarily mean that there is that amount free as a contiguous block. 

So, for example, if you were to do something like
```
let a = some_allocator.allocate(<1MB Layout>);
let b = some_allocator.allocate(<1MB Layout>);
let c = some_allocator.allocate(<1MB Layout>);
some_allocator.deallocate(a.as_ptr());
some_allocator.deallocate(c.as_ptr());
```
I think `some_allocator.free()` would indicate 2MB, but the largest allocation you could do without fallback to GCD would be 1MB, because there's a 1MB allocated chunk in the middle of the free space.

joschock: It might be more straightforward to simply call `allocate`/`deallocate` on the layout in question instead of inferring it from free(). This will ensure that the layout can be accommodated regardless of the fragmentation of the map.

os-d: Ah, yeah you are correct here. I was just coming to the same conclusion on `allocate -> deallocate`. I will make that change, thanks for catching this.

dxe_core/src/allocator.rs
```diff
@@ -493,14 +494,26 @@ fn merge_blocks(
...
-pub fn get_memory_map_descriptors() -> Vec<efi::MemoryDescriptor> {
+pub(crate) fn get_memory_map_descriptors() -> Result<Vec<efi::MemoryDescriptor>, efi::Status> {
     let mut descriptors: Vec<MemorySpaceDescriptor> = Vec::with_capacity(GCD.memory_descriptor_count() + 10);
+
+    // the fold operation will allocate boot services data, which we cannot do because we cannot change the memory map
+    // after getting the descriptors from the GCD, we'll now be invalid if we ended up overflowing a pool and getting
+    // more memory from the GCD. Therefore, we need to call the ensure_capacity API before we get the GCD descriptors
+    // to ensure we don't overflow the boot services data pool. Let's make sure we have a few extra descriptors
+    let capacity = (GCD.memory_descriptor_count() + 10) * mem::size_of::<MemorySpaceDescriptor>();
+    if let Err(status) = ensure_capacity(efi::BOOT_SERVICES_DATA, capacity, 0x8) {
```
joschock: I can think of pathological race conditions that this doesn't address - if, for example, some timer interrupt comes in here after ensure_capacity and allocates a big chunk'o'data from BS, it could blow your map even with this safeguard. I'm not sure it's worth the complexity to make it bullet-proof against that scenario,  but might be worth a warning comment.

dxe_core/src/allocator.rs
```diff
@@ -553,7 +566,7 @@ pub fn get_memory_map_descriptors() -> Vec<efi::MemoryDescriptor> {
                 },
             })
         })
-        .fold(Vec::new(), merge_blocks)
+        .fold(Vec::new(), merge_blocks))
```
joschock: I think you could avoid an allocation on fold here, by pre-allocating the vector above, since the fold is always going to be the same or smaller than the descriptor list:
```rust
let mut descriptors: Vec<MemorySpaceDescriptor> = Vec::with_capacity(GCD.memory_descriptor_count() + 10);
//pre-allocate for the fold below
let mut merged_descriptors: Vec<MemorySpaceDescriptor> = Vec::with_capacity(GCD.memory_descriptor_count() + 10);

...etc...
.fold(merge_descriptors, merge_blocks))
```

But I still think ensure_capacity is a good idea since it allows more space to maybe catch descriptors generated by stuff beyond the scope/control of this routine.

dxe_core/src/allocator.rs
os-d: Yeah, that was what I was hoping to capture with the extra 10 descriptors, some wiggle room, though of course massive allocation will still break that. I can definitely add a comment here.

dxe_core/src/allocator.rs
os-d: I'm open to either method. The main reason for this API is preallocating page table memory. I was going to convert the MAT to use this as well in this PR, because today it has to install an empty table for a runtime allocation to occur, then install the real table over top of it. The reason I decided not to is that the re-allocation on the first install is not deterministic, it depends on how the length of other tables that have been already installed. I could do a large enough number, but then if we didn't have that memory but did have the right amount for what is in the config table, that would be bad. So I left that as is.

The other way we can take this is if we don't think there is enough of a use case outside of mu-paging is to just create a custom allocator wrapper for mu-paging that keeps a preallocated pool to draw from. That was our original design point (and what the C code does) before we had discussed the more general ensure_capacity API. I'm open to going back to that if we like that better

joschock: I think it's fine either way as well - just noting that if you don't pre-alloc the fold vector, it's going to eat into your ensure_capacity() call. 

> The main reason for this API is preallocating page table memory. 

That's coming from page allocations, not pool allocations, right? If that's the case, then `allocate`/`deallocate` flow in `ensure_capacity` will claim memory for the pool, and I don't think that memory will be used to satisfy page allocations (since it's now associated with the pool allocator).

os-d: > I think it's fine either way as well - just noting that if you don't pre-alloc the fold vector, it's going to eat into your ensure_capacity() call.
> 

I’m not quite following. The reason I would make the ensure_capacity() call is because we are going to allocate memory for the fold vector. We overallocate a bit just in case something were to happen interrupt wise, but your suggestion of pre-allocating the fold vector would be equivalent.

Oh, are you seeing the ensure_capacity() as trying to foolproof the memory map from changing in between get_memory_map calls? If so, the bootloader will make its own alloc/free calls and thereby change the memory map. I do have a separate planned change to disallow/silently fail alloc/free during EBS to avoid all the issues we see there.

But in this call all I care about is that the memory map does not change after I fetch the descriptors until I return the map to the caller, right?

> > The main reason for this API is preallocating page table memory
>
> That's coming from page allocations, not pool allocations, right? If that's the case, then `allocate`/`deallocate` flow in `ensure_capacity` will claim memory for the pool, and I don't think that memory will be used to satisfy page allocations (since it's now associated with the pool allocator).

With the ensure_capacity() scheme, we would do pool allocations, for the reason you stated. If we do custom allocator wrapper, we’d allocate via pages, probably. mu-paging itself doesn’t care, I think, we just want XP, which will already be set. But, as I integrate that assumption may get challenged :). We do use pages in the C codebase, but split it up into pools in the page table lib.

joschock: > The reason I would make the ensure_capacity() call is because we are going to allocate memory for the fold vector.

yep, in total agreement. I think the ensure_capacity() is good idea either way because of the "something may happen interrupt wise" wiggle room; but if you pre-alloc the vector, you have more wiggle room (or you could reduce the amount of ensure_capacity()). I just meant that if you don't pre-allocate the vector, your ensure_capacity doesn't give you quite as much room for unforeseen interrupt allocations because the fold vector will occupy much of it - which is fine, and kind of the main point of ensure_capacity(). So in that sense it's just a matter of how much extra capacity you're building in for padding.

I thought of one other thing to consider though, which is that Vec.push() allocates an unspecified amount of memory and does a copy if it exceeds the current capacity, so, in theory, it could be bigger than the ensure_capacity() call and still meet Vec spec. I don't think the current implementation of Vec will do that (I think it just grows by 1). But it would be legal for Rust to change and allocate space 1000 elements on vec.push when current vec capacity is exceeded if it wanted to. For that reason it might be a good idea to pre-alloc the fold vector just to avoid expansion + copy during the fold itself.

From [std::Vec](https://doc.rust-lang.org/std/vec/struct.Vec.html):

[push](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.push) and [insert](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.insert) will (re)allocate if [len](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.len) == [capacity](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.capacity).
...

Vec does not guarantee any particular growth strategy when reallocating when full, nor when [reserve](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.reserve) is called. The current strategy is basic and it may prove desirable to use a non-constant growth factor. Whatever strategy is used will of course guarantee O(1) amortized [push](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.push).

os-d: Gotcha, I agree, I'll pre-alloc the vector. Let's discuss in the Rust meeting today whether we think the ensure_capacity API has merits outside of paging (since here we could pre-alloc any vector size we want, which would be the equivalent of ensure_capacity) or if we think a custom allocator wrapper for paging would be better.

The custom allocator feels a bit cleaner, in that it is wrapped up in a nice package, but it also means we would need to implement additional pool functionality, which I would prefer to keep in one place and just have one implementation of a pool allocator. There might be another path too where an allocator is created just for page table memory, so it is still BootServicesData, but has some well known handle (like our standard ones) that we can operate on like standard allocators. But we would still need an ensure_capacity API, so I don't think that's worth the effort.

crates/uefi_allocator/src/fixed_size_block_allocator.rs
```diff
@@ -148,6 +151,57 @@ impl FixedSizeBlockAllocator {
         }
     }
 
+    /// Ensures that the allocator has enough capacity to satisfy an allocation of the given size and alignment.
```
cfernald, line 154: nit function summary should have a space after it.

https://microsoft.github.io/mu/CodeDevelopment/rust_documentation_conventions/#function-template

os-d: Darn you copilot! Will update.

PR #123, opened 2024-10-23
Clean up unwraps to expect/error handling

berlin-with0ut-return:
## Description
This PR examines all usage of unwrap. The usages that are guaranteed safe by current code are converted to expect for more informative error messages, while the usages that could possibly error are converted to error handling and logging.

- [x] Impacts functionality? (No major functional changes. The only observed differences will be returning or logging an error instead of panicking.)
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests? (Existing test should be sufficient to cover these changes, as they do not result in any functional changes.)
- [ ] Includes documentation? (None needed.)

## How This Was Tested
All integration and unit tests are still passing after the error handling changes.

## Integration Instructions
N/A. No functional changes.

### DIFF
--- a/crates/uefi_event/src/lib.rs
+++ b/crates/uefi_event/src/lib.rs
@@ -165,7 +165,7 @@ impl fmt::Debug for EventNotification {
         f.debug_struct("EventNotification")
             .field("event", &self.event)
             .field("notify_tpl", &self.notify_tpl)
-            .field("notify_function", &(self.notify_function as usize))
+            .field("notify_function", &self.notify_function.map(|f| f as usize))
             .field("notify_context", &self.notify_context)
             .finish()
     }


adv_logger/src/component.rs
```diff
@@ -86,9 +86,16 @@ where
...
+        if let Some(physical_hob_list_info) =
+            unsafe { (physical_hob_list as *const PhaseHandoffInformationTable).as_ref::<'static>() }
+        {
+            hob_list = Hob::Handoff(physical_hob_list_info);
+        } else {
+            log::error!("Could not initialize adv logger due to null hob list.");
+            return Err(EfiError::InvalidParameter);
+        }
 
         for hob in &hob_list {
```
makubacki: I slightly prefer tidying up the logic a bit like this so you don't have a lone let declaration and a more space if-else structure. But feel free to go with what you like.

```suggestion
    pub fn init_advanced_logger(&self, physical_hob_list: *const c_void) -> Result<()> {
        debug_assert!(!physical_hob_list.is_null(), "Could not initialize adv logger due to null hob list.");
        let hob_list_info =
            unsafe { (physical_hob_list as *const PhaseHandoffInformationTable).as_ref() }.ok_or_else(|| {
                log::error!("Could not initialize adv logger due to null hob list.");
                EfiError::InvalidParameter
            })?;
        let hob_list = Hob::Handoff(hob_list_info);

        for hob in &hob_list {
```

adv_logger/src/memory_log.rs
```diff
@@ -251,9 +254,14 @@ impl AdvLoggerMessageEntry {
     /// the provided length. The caller is responsible for ensuring this memory
     /// range is valid.
     ///
-    pub unsafe fn init_from_memory(address: *const c_void, length: u32, log_entry: LogEntry) -> &'static Self {
+    pub unsafe fn init_from_memory(address: *const c_void, length: u32, log_entry: LogEntry) -> Option<&'static Self> {
+        debug_assert!(
+            size_of::<Self>() + log_entry.data.len() <= length as usize,
+            "Advanced logger entry initialized in insufficient buffer!"
```
makubacki:
```suggestion
            "Advanced logger entry initialized in an insufficiently sized buffer!"
```

adv_logger/src/memory_log.rs
```diff
@@ -315,9 +323,12 @@ impl<'a> Iterator for AdvLogIterator<'a> {
...
-            Some(entry)
+            if let Some(entry) = unsafe { entry.as_ref() } {
+                self.offset += entry.aligned_len();
+                Some(entry)
+            } else {
+                None
+            }
         }
     }
 }
```
makubacki: I think you can clean up the logic here using `map()`.

```suggestion
        } else {
            let entry = unsafe { (self.log_info as *const AdvLoggerInfo).byte_add(self.offset) }
                as *const AdvLoggerMessageEntry;
            unsafe { entry.as_ref() }.map(|entry| {
                self.offset += entry.aligned_len();
                entry
            })
        }
    }
}
```

crates/uefi_allocator/src/uefi_allocator.rs
```diff
@@ -263,7 +263,13 @@ impl UefiAllocator {
...
-        self.allocator.deallocate(NonNull::new(allocation_info as *mut u8).unwrap(), (*allocation_info).layout);
+        self.allocator.deallocate(
+            match NonNull::new(allocation_info as *mut u8) {
+                Some(non_null_ptr) => non_null_ptr,
+                None => return efi::Status::INVALID_PARAMETER,
+            },
+            (*allocation_info).layout,
+        );
         efi::Status::SUCCESS
     }
```
makubacki: I personally find this logic to be clearer as opposed to embedding a match in place of the function argument. It clearly shows attempt to acquire a non-null pointer followed by a function call in case of success and `INVALID_PARAMETER` returned on failure. This is subjective though, feel free to use what you like.

```suggestion
        (*allocation_info).signature = 0;
        if let Some(non_null_ptr) = NonNull::new(allocation_info as *mut u8) {
            self.allocator.deallocate(non_null_ptr, (*allocation_info).layout);
        } else {
            return efi::Status::INVALID_PARAMETER;
        }
        efi::Status::SUCCESS
    }
```

crates/uefi_device_path/src/lib.rs
```diff
@@ -74,19 +76,26 @@ pub fn device_path_node_count(device_path: *const efi::protocols::device_path::P
...
+
+        if let Ok(offset) = current_length.try_into() {
+            current_node_ptr = unsafe { current_node_ptr.byte_offset(offset) };
+        } else {
+            return Err(efi::Status::INVALID_PARAMETER);
+        }
     }
-    (node_count, dev_path_size)
+    Ok((node_count, dev_path_size))
 }
```
makubacki: You could consider something like this. I tend to prefer it because it handles the state of `offset` in one place and then makes use of its value in a simple assignment if it is good.


```suggestion

        let offset = current_length.try_into().map_err(|_| efi::Status::INVALID_PARAMETER)?;
        current_node_ptr = unsafe { current_node_ptr.byte_offset(offset) };
    }
    Ok((node_count, dev_path_size))
}
```

crates/uefi_event/src/lib.rs
```diff
@@ -154,20 +154,27 @@ pub struct EventNotification {
...
 impl fmt::Debug for EventNotification {
     #[cfg_attr(feature = "nightly", feature(no_coverage))]
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        f.debug_struct("EventNotification")
-            .field("event", &self.event)
-            .field("notify_tpl", &self.notify_tpl)
-            .field("notify_function", &(self.notify_function as usize))
-            .field("notify_context", &self.notify_context)
-            .finish()
+        if let Some(notify_function) = self.notify_function {
```
makubacki: This is a place where a large portion of the code is the same in the success and failure case and I think it'd be beneficial to handle the `notify_function` logic in the argument location. Otherwise, you're eyes have to scan and compare each block.

crates/uefi_event/src/lib.rs
```diff
@@ -457,7 +464,14 @@ impl EventDb {
...
+            let current_event;
+            if let Some(current) = self.events.get_mut(&event) {
+                current_event = current;
+            } else {
+                debug_assert!(false, "Event {:?} not found.", event);
+                log::error!("Event {:?} not found.", event);
+                continue;
+            }
             if current_event.event_type.is_timer() {
                 if let Some(trigger_time) = current_event.trigger_time {
```
makubacki: Since you're essentially performing an assignment based on the success of that condition, I think it'd be a little clearer to just make the assignment directly and show the value being returned.


```suggestion
        for event in events {
            let current_event = if let Some(current) = self.events.get_mut(&event) {
                current
            } else {
                debug_assert!(false, "Event {:?} not found.", event);
                log::error!("Event {:?} not found.", event);
                continue;
            };
            if current_event.event_type.is_timer() {
                if let Some(trigger_time) = current_event.trigger_time {
```

crates/uefi_event/Cargo.toml
```diff
@@ -15,6 +15,7 @@ path = "src/lib.rs"
 [dependencies]
 r-efi = {workspace=true}
 uefi_tpl_lock = {workspace=true}
+log = { workspace = true }
```
makubacki: Minor, but for consistency can you make the others have the same spacing? That will make it more obvious in the future how new dependencies should be formatted.


```suggestion
r-efi = { workspace = true }
uefi_tpl_lock = { workspace = true }
log = { workspace = true }
```

dxe_core/src/misc_boot_services.rs
```diff
@@ -295,7 +299,11 @@ pub extern "efiapi" fn exit_boot_services(_handle: efi::Handle, map_key: usize)
     };
 
     // Clear non-runtime services from the EFI System Table
-    SYSTEM_TABLE.lock().as_mut().unwrap().clear_boot_time_services();
+    SYSTEM_TABLE
+        .lock()
+        .as_mut()
+        .expect("ST pointer is null, but it must exist and be valid.")
```
makubacki:
```suggestion
        .expect("The System Table pointer is null. This is invalid.")
```

crates/uefi_event/src/lib.rs
berlin-with0ut-return: i was a little stuck on this because the cast to a usize makes it a little harder to handle. does something like this improve on the previous iteration?

```
let notify_function_usize = match self.notify_function {
    Some(func) => func as usize,
    None => 0,
};

f.debug_struct("EventNotification")
    .field("event", &self.event)
    .field("notify_tpl", &self.notify_tpl)
    .field("notify_function", &notify_function_usize)
    .field("notify_context", &self.notify_context)
    .finish()
```

adv_logger/src/memory_log.rs
```diff
@@ -344,9 +353,9 @@ mod tests {
         loop {
             let data = entries.to_be_bytes();
             let entry = LogEntry { level: 0, phase: 0, timestamp: 0, data: &data };
-            let log_entry = log.add_log_entry(entry);
+            let log_entry = log.unwrap().add_log_entry(entry);
```
Javagedes, line 356: these `.unwrap()`s will need to be `expects` with a reason why the log is guaranteed to be valid here (or error handle if can't be guaranteed). There are a few other newly introduced lines that call unwrap().

berlin-with0ut-return: this is in the test mod so i haven't addressed the unwraps added here. i don't think it's guaranteed to be valid but i assume it's okay for tests to panic in error cases, so i can still change them into expects?

Javagedes: Ah. I think i saw you had this same conversation with @os-d . It's completely fine since it is in the tests. I'll close this :)

berlin-with0ut-return:
> Have you ran this on mu_tiano_platforms to ensure we still boot to UEFI shell? Our current CI system does not do that at the moment, so I want to make sure :)

just tested on mu_tiano and it boots to shell normally!

PR #125, opened 2024-10-23
Add label syncer workflow

makubacki:
## Description

Workflow is used to ensure any labels defined for the project are available.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- Run workflow on fork and verify labels are synced as expected

## Integration Instructions

Reuse the LabelSyncer.yml workflow in another project GitHub repo if labels should
be synced there.

### DIFF
--- a/.github/workflows/LabelSyncer.yml
+++ b/.github/workflows/LabelSyncer.yml
@@ -1,0 +1,38 @@
+# This workflow syncs GitHub labels to the integrating repository.
+#
+# The labels are declaratively defined in .github/Labels.yml.
+#
+# Copyright (c) Microsoft Corporation.
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+#
+# For more information, see:
+# https://github.com/EndBug/label-sync
+
+name: Label Sync Workflow
+
+on:
+  workflow_call:
+    secrets:
+      GH_PAT:
+        description: "GitHub Personal Access Token"
+        required: true
+
+jobs:
+  sync:
+    name: Sync
+    runs-on: ubuntu-latest
+
+    permissions:
+      issues: write
+
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          sparse-checkout: .github/Labels.yml
+          token: ${{ secrets.GH_PAT }}
+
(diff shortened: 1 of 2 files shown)


.github/workflows/label-sync.yml
```diff
@@ -0,0 +1,26 @@
...
+    - cron: '0 0,12 * * *'
+  workflow_dispatch:
+
+jobs:
+  sync:
+
+    permissions:
+      issues: write
+
+    uses: ./.github/workflows/LabelSyncer.yml
```
Javagedes, line 25: If you want this to be reusable across workflows you need to point it at `pop-project/uefi-dxe-core/.github/workflows/LabelSyncer.yml@main`

.github/workflows/LabelSyncer.yml
```diff
@@ -0,0 +1,38 @@
...
+    name: Sync
+    runs-on: ubuntu-latest
+
+    permissions:
+      issues: write
+
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          sparse-checkout: .github/Labels.yml
```
Javagedes, line 31: I think this expects .github/Labels to be in whatever repo this is being called from. Not sure if that was your intent or if you want it to be a single source of truth in uefi-dxe-core.

.github/workflows/label-sync.yml
Javagedes: Sorry, This is what would be sitting in each repo, so it's fine how it is here. But in other repos it just needs to point at what I mentioned above.

makubacki: That's the plan. I want to keep the path minimal where possible as it is less to update when the project name inevitably changes.

.github/workflows/LabelSyncer.yml
makubacki: It will. I was planning to add the Labels.yml file to each repo for now (until we have a better CI centralization plan in place). These haven't changed since they were created so I'm not too worried about syncing them.

PR #129, opened 2024-10-25
Add use of interrupt manage in DXE core

cfernald:
## Description

Adds use of the `InterruptManager` trait from the `uefi_interrupt` create for initializing the interrupt and exception handlers. 

- [x] Impacts functionality?
- [ ] Impacts security?
- [x] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on Q35

## Integration Instructions

Platforms must use an implementation of Interrupt Manager appropriate to their architecture when creating the core.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,13 +76,13 @@ should download `cargo-make` and `cargo-tarpaulin`, however it is suggested that
 ### Building uefi-dxe-core Crates
 
 The following commands build all crates with one of our three supported targets: `x86_64-unknown-uefi`,
-`i686-unknown-uefi`, and your host system target triple. The default compilation mode is `development`, but you can
+`aarch64-unknown-uefi`, and your host system target triple. The default compilation mode is `development`, but you can
 easily switch modes with the `-p` flag.
 
-- Development Compilation (i686-unknown-uefi): `cargo make build-aarch64`
+- Development Compilation (aarch64-unknown-uefi): `cargo make build-aarch64`
 - Development Compilation (x86_64-unknown-uefi): `cargo make build-x64`
 - Development Compilation (host system): `cargo make build`
-- Release Compilation (i686-unknown-uefi): `cargo make -p release build-aarch64`
+- Release Compilation (aarch64-unknown-uefi): `cargo make -p release build-aarch64`
 - Release Compilation (x86_64-unknown-uefi): `cargo make -p release build-x64`
 - Release Compilation (host system): `cargo make -p release build`
 
--- a/crates/uefi_allocator/src/fixed_size_block_allocator.rs
+++ b/crates/uefi_allocator/src/fixed_size_block_allocator.rs
@@ -26,10 +26,12 @@ use r_efi::efi;
 use uefi_gcd::gcd::SpinLockedGcd;
 
 /// Type for describing errors that this implementation can produce.
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
 pub enum FixedSizeBlockAllocatorError {
     /// Could not satisfy allocation request, and expansion failed.
     OutOfMemory,
+    /// Invalid input parameter
+    InvalidParameter,
 }
 
 /// Minimum expansion size - allocator will request at least this much memory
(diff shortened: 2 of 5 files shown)


cfernald:
relies on https://github.com/pop-project/uefi-core/pull/10

dxe_core/src/lib.rs
```diff
@@ -93,26 +93,35 @@ pub(crate) static GCD: SpinLockedGcd = SpinLockedGcd::new(Some(events::gcd_map_c
 ///     .initialize(physical_hob_list);
 /// ```
 #[derive(Default)]
-pub struct Core<CpuInitializer, SectionExtractor>
+pub struct Core<CpuInitializer, SectionExtractor, InterruptManager>
 where
     CpuInitializer: interface::CpuInitializer + Default,
     SectionExtractor: fw_fs::SectionExtractor + Default + Copy + 'static,
+    InterruptManager: uefi_interrupt::InterruptManager + Default + Copy + 'static,
```
Javagedes, line 100: Since Interrupt Manager impl selection is based on architecture, it does not really need to be selected by the platform. Instead, realistically, only the configuration of the implementation needs to be selected. I can see this being the case for multiple other traits coming from you and vineel and Oliver. Due to this, I would like to propose that we do type re-defines for X64 and AARCH64 that pre-selects the trait implementation (but not the configuration), so that platforms don't have to worry about it. i.e. something like this:

```rust
type X64Core<SectionExtractor> = Core<CpuInitializer = uefi_cpu_init::X64CpuInitializer, SectionExtractor, InterruptManager = uefi_interrupt::X64InterruptManager>

type Aarch64Core<SectionExtractor> = Core<CpuInitializer = uefi_cpu_init::NullCpuInitializer, SectionExtractor, InterruptManager = uefi_interrupt::Aarch64InterruptManager>
```

Platforms can 100% still use the `.with_interrupt_manager()` to use a interrupt manager with custom configuration, but at least doing the above reduces the amount of integration points a platform needs to worry about.

This makes it such that a platform just has to do this:

```rust
X64Core::default()
    .with_section_extractor(<extractor>)
    .initialize()
    .start()
    .unwrap()
```

But can still do customizations if they want:

```rust
X64Core::default()
    .with_section_extractor(<extractor>)
    .with_interrupt_manager(MyInterruptManager::default().with_config(true).with_other_confg(false))
    .initialize()
    .start()
    .unwrap()
```

Also if a platform needs to be different than the norm, they can just use the `Core` struct and select their own depenencies from scatch. But with this, we can pre-select some trait impls for platforms to reduce churn.

cfernald: I think this is reasonable. Let me play around with it a bit

cfernald: I added this. I tested dropping the Defaults for the architecture defaults and it works fine. Seems like goodness to me. 

Before:
```Rust  
Core::default()
        .with_cpu_initializer(uefi_cpu_init::X64CpuInitializer::default())
        .with_interrupt_manager(uefi_interrupt::InterruptManagerX64::default())
        .with_section_extractor(section_extractor::CompositeSectionExtractor::default())
        // Add any config knob functions for pre-gcd-init Core
        // .with_some_config(true)
        .initialize(physical_hob_list) // We can make allocations now!
        // Add any config knob functions for post-gcd-init Core
        // .with_some_config(true)
        .with_driver(Box::new(adv_logger_component))
        .with_driver(Box::new(HelloComponent::default()))
        .with_driver(Box::new(HelloComponent::default().with_name("Dxe Core")))
        .with_driver(Box::new(HelloComponent::default().with_name("World")))
        .start()
        .unwrap();
````

after:

```Rust
    X64Core::default()
        .with_section_extractor(section_extractor::CompositeSectionExtractor::default())
        // Add any config knob functions for pre-gcd-init Core
        // .with_some_config(true)
        .initialize(physical_hob_list) // We can make allocations now!
        // Add any config knob functions for post-gcd-init Core
        // .with_some_config(true)
        .with_driver(Box::new(adv_logger_component))
        .with_driver(Box::new(HelloComponent::default()))
        .with_driver(Box::new(HelloComponent::default().with_name("Dxe Core")))
        .with_driver(Box::new(HelloComponent::default().with_name("World")))
        .start()
        .unwrap();
```

cfernald: Undoing this change. It is causing some odd issues with `cargo make docs` and i'd rather not hold up the breaking change fix. We can follow up with a second change after figuring out the docs issue.

PR #135, opened 2024-10-28
Add uefi_cpu_init and uefi_interrupt crates

Javagedes:
## Description

With PR #133 adding CPU Architecture specific default trait implementations for some DXE Core traits, they have now become implementation details of the DXE Core rather than platform customizable traits. Due to this, these traits (uefi_cpu_init and uefi_interrupt) now make more sense in the uefi-dxe-core repo rather than the uefi-core repo, which is meant to be a place for reference implementations of platform customizable traits for the DXE core.

The trait continues to provide a point of abstraction for the DXE Core, but is now a point of abstraction for future CPU architecture specific implementations rather than platform specific implementations. A platform, if necessary, can still provide its own implementation of the trait, but the expectation is that the default implementation will be sufficient for most platforms.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Built qemu bins and verified boot to UEFI shell.

## Integration Instructions

The git path in the Cargo.toml file of any rust bin repo needs to switch to this repo rather than uefi-core.


vineelko:
Hey folks, Sorry for my lack of UEFI expertise. I’m not entirely sure what’s technically accurate regarding where the code should reside(platforms vs implementation vs configuration vs layering), but I strongly feel we need guidance on how everything should fit into the overall grand scheme of things, especially **before** we start making changes. Shuffling code between repositories doesn’t scale(I am not referring to this PR)—particularly when dealing with three different repositories (`uefi-dxe-core`, `uefi-core`, `uefi-rust-bins` ) and with varying local development dependencies (cargo make vs cargo alias vs workspace version dependencies etc).

Don't get me wrong Joey; I'm not against to this PR (And in fact I am happy as it eliminates `uefi-core`, so that's one less repo to worry about). 

However, unfortunately, each one of us have (or had) our own viewpoints on where each component should reside and due to this, we had to redo bunch of stuff moving across different repos. For example, my specific changes related to CPU Arch protocol, I have been shuffling them from 
1. `mu-rs`(my bad, not the right place to begin with), - https://dev.azure.com/microsoft/MsUEFI/_git/mu-rs/pullrequest/11614168
2. Then moved to `DxeRust` - https://dev.azure.com/microsoft/MsUEFI/_git/DxeRust/pullrequest/11637315 
3. Then moved to `uefi-core` - because that's where the cpu stuff was initially - https://github.com/pop-project/uefi-core/pull/11
4. Then after some discussions among the cpu dxe folks last week, we agreed on below design, `uefi-core`(cpu arch lib and interrupt manager) + `uefi-dxe-core`(only the public cpu arch protocol). - Working on it now, mostly done.
5. image
6. Just sharing the teams chat to make the point that we are disconnected and not to point fingers.
7. image
8. `uefi-dxe-core`(with this PR) - I am fine moving it to `uefi-dxe-core` as it makes my life easy.  
 
So, the point I’m trying to make is, we should have something like the above screenshot for everything we plan to do moving forward. It would help everyone agree on something concrete, rather than relying on discussions in Teams channels (which are hard to follow and decipher later) nor random 1:1 chat. This is especially important as we are working on multiple repos (sometimes confusing, as couple of them have "core" in them :-)). 

My 2cents, FWIW, I am still biased towards having one mono repo with functionality divided among as many crates as we want. This helps keep the developer velocity, at least until we have a fully flushed out implementation of "Uefi in Rust" and for productizing.

makubacki:
> Hey folks...

This PR is the result of a "Code Organization" meeting today that split off as a subteam as discussed in last week's UEFI Rust Design and Dev Sync. Notes from today's meeting are in the UEFI Rust Notebook for reference.

This is a large project that is moving quickly. We are learning as we go and moving parts affect other parts. The CPU code (or any other) code cannot be created in a vacuum, it must adapt and evolve as the project moves from ADO to GitHub and higher-level code patterns and trends emerge that affect it.

The best long-term decision is going to be reached by experimenting with bold ideas quickly and not being afraid to make substantial changes - this is the time to make those changes. It will only be more difficult later. When I look at the history you shared, I see the code direction becoming more confident and the code interfaces and layering becoming cleaner. That's good. It's also good that everyone involved is learning along the way too.

There is a place for diagrams, strong documentation, etc. And we do invest heavily in that (at this point) where the effort is justified. The diagram you shared, for example, is great for understanding how the code is structured and we **should** ask for more of that in code reviews to make sure all participants are on the same page and hopefully it can end up in the code documentation. At the same time, it's not going to solve larger project decisions about how to organize code. We have to stay nimble with open communication.

For reference, we are planning to open source the code in two weeks and productization is already here. We did use a monolithic repo for a long time (~1.5 years) for the reasons you mentioned. It's time to move on now.

For this topic, the PR is a place for discussion and feedback, it can continue here if you'd like. Feel free to invite Joey and I to the CPU meeting. You can also join the next code organization call if you'd like as well. The main takeaway is that we're going to have change. I think that's a good thing as long as it's productive - we're learning and applying those learnings to improve the code. I've taken a note to clarify and document decisions as quickly as possible moving forward to help keep things coordinated.

vineelko:
Thanks, Michael, for adding the meeting minutes in OneNote. I wasn’t part of the meeting, so I didn’t know this was coming. Please include me in the next code organization call for more context. As I mentioned in my previous comment, I have no objections for this PR.

> The best long-term decision is going to be reached by experimenting with bold ideas quickly and not being afraid to make substantial changes 

> we are planning to open source the code in two weeks and productization is already here

I’m not saying we shouldn’t make these code changes. All I’m saying is that if this was a mono repo, changes like these(either from me or from others) would probably cause less churn, especially since we have limited time to open source it. This is why I shared my own experience. But since the mono repo idea has already been vetted, I probably don’t have much to add other than to adapt to the churn.

> The main takeaway is that we're going to have change. I think that's a good thing as long as it's productive

But right now, working with multiple repos is not productive at least as far as I have experienced it. To share my own workflow when trying to make these cross-repo changes (like ours). Just FYI, most of these aren’t technical challenges but rather pain points I felt were worth sharing.

1. Edit each of the three workspace `cargo.toml` files to manually point to local paths. While working on these, if there are new changes upstream, we have to rebase. This isn’t very straightforward, as we point dependencies to specific commit hashes.
2. Each repo has slightly different build commands (Even though, technically we only need to build `uefi-rust-bins`)
3. It’s challenging to perform a global find-and-replace in the editor. One workaround I found is using vscode workspace files (something like below) to open all repos in one vscode instance.
```json
 {
    "folders": [
        {
            "path": "C:/r/mtrr"
        },
        {
            "path": "C:/r/paging"
        },
        {
            "path": "C:/r/uefi-dxe-core"
        },
        {
            "path": "C:/r/uefi-core"
        },
        {
            "path": "C:/r/qemu_rust_bins"
        },
    ],
}
```
4. Because these are independent repos on disk, Rust Analyzer’s behavior is unpredictable. Rust Analyzer is already not great with `uefi_core::if_x64!` 
5. To build these repos separately for any reason, we have to open multiple terminals. The workaround I found is to use terminal splits (not ideal, but it works).
image
6. Switching branches across all repos require running `git checkout <branchname>` in each terminal split window, and we also have to run `git stash` to discard previous `cargo.lock` files before switching.
7. Today, I was chasing a weird build error which I initially thought was due to differences in the toolchain (`uefi-dxe-core`, `uefi-core` are using 1.80 while `uefi-rust-bins` is using 1.76). Though it turned out be a `x86_64` crate version issue, I was sidetracked by it and spent time on toolchain overrides, thinking there was an issue with my setup. 
8. I haven't got a chance to look at how the GitHub workflows are setup for these repos, so that might be additional overhead.

None of these are showstoppers and will likely improve over time, but right now, they definitely affect developer velocity. Please let me know if I’m doing something incorrectly; I’m happy to adapt if there are better ways others are using.

Javagedes:
Hi @vineelko ! I agree building from multiple repositories is annoying. I too have a similar workflow (`qemu_rust_bins`'s Cargo.toml override to point at a path of `uefi-dxe-core`. I then have `uefi-dxe-core`'s Cargo.toml override to a local path of `uefi-core`. 

I've worked on syncing build commands across our repositories and our github workflows expect these commands to exist. This is true for `uefi-dxe-core` and `uefi-core`, and any other repos should do the same (`uefi_rust_bins` being different as it builds bins):

- `cargo make doc` to build documentation
- `cargo make build` to build all crates with std
- `cargo make build-x64` to build all crates with `x86_64-unknown-uefi` target
- `cargo make build-aarch64` to build all crates with `aarch64-unknown-uefi` target

If there are any repos that are not like this, please let me know and I will fix. let me also take a look at cargo make's [load_script](https://github.com/sagiegurari/cargo-make?tab=readme-ov-file#load-scripts) functionality to see about having a single source of truth for the commands, so that we are always in sync.

the issue with rust-analyzer is not with `if_x64!` macro. All that is doing is adding a cfg block to all lines within the macro. The issue, from a quick glance, kind of depends on the repo you are in. If you are in [pop-project/paging](https://github.com/pop-project/paging), you have no `.vscode/settings.json` file overriding the default rust-analyzer to the target you are currently working with (`x86_64-unknown-uefi`, `aarch64-unknown-uefi`, etc). Unfortunately, the analyzer only works with one target, so you'll only have the analyzer working with one architecture at at time. If you are in [pop-project/uefi-core](https://github.com/pop-project/uefi-core), then the issue is that there was an issue when the repo commits were squashed, and the `setting.json` file should replace the `settings.json` file in the repo, and somehow in the `Makefile.toml`, the task.check_no_std is using `NO_STD_FLAGS` instead of `BUILD_FLAGS`. This is all to say, the rust-analyzer __should__ work properly, but there is torn state that needs to be fixed. So, if you ever see annoyances like that, please point them out because something is broken.

As for the workflows, I've been working on them to be reusable across all our repositories, so once we have the Makefile's sync'd across repos that is no issue.

Cargo.toml
```diff
@@ -16,30 +16,34 @@ edition = "2021"
 rust-version = "1.80.0"
 
 [workspace.dependencies]
-uefi_cpu_init = {git = "https://github.com/pop-project/uefi-core.git", rev = "396bac975b85d11539c018385b920f0a8da3fd93"}
+# pop-project crates
```
os-d, line 19: nit: if this name is going to change and we are open sourcing in two weeks, we may not want to embed this name in the code, but obviously all these links will change, so not hard to do a final scrub

Javagedes:
Also, targeting @vineelko @os-d @kuqin12 who have been doing a lot of this cpu work. The reason that I think moving these two crates into uefi-dxe-core is because these are CPU architectural configurations, not a platform configuration made by the platform developers. all X86_64 platforms will use a specific implementation and all AARCH64 platforms will use the other. These two crates are also DXE specific from my understanding and would not be used in other phases. **Is this a correct assumption? If not, it may make sense for it to stay in uefi-core**.

The intent of uefi-core is to house core phase implementations that are customizable by the platform. As that stands right now, the way I see it, that is really only section extraction and logging.

kuqin12:
I do not disagree with the move in general. But I thought uefi_interrupt should remain in the uefi-core? I think its implementation is not UEFI specific and can be reused elsewhere. At least for AArch64 version, it is a combination of exception handler + gic manager + interrupt dispatcher. I think the same applied to x64 version?

uefi_cpu_init, or even better just called `uefi_cpu` is better moved here as it is very DXE specific and will host the implementation for protocols. Thoughts?

os-d:
Adding @cfernald here

@Javagedes , agreed that platforms will not customize the interrupt handling, but as to Kun's point, he, Vineel, and Chris have been trying to structure the cpu work so that the main cpu stuff can be reused across any environment, not specific to uefi (or DXE). And then there is an interface layer for the UEFI/DXE specific stuff.

Javagedes:
> Adding @cfernald here
> 
> @Javagedes , agreed that platforms will not customize the interrupt handling, but as to Kun's point, he, Vineel, and Chris have been trying to structure the cpu work so that the main cpu stuff can be reused across any environment, not specific to uefi (or DXE). And then there is an interface layer for the UEFI/DXE specific stuff.

I have not been following the work in these crates heavily and thus don't know the final "intentions" (nor do I have an opinion, so keep on moving forward ;) ), but I was under the impression the core reusable functionality was being done elsewhere, like mtrr crate, paging crate, etc, and that the crates in uefi-core were the final "wrappers" that implement the uefi dxe-core requirements like implementing the trait.

It's completely fine if that is not the case, and these two crates are the final "reusable" product - they should probably still be moved and renamed to not have `uefi` in it, just not into the `uefi-dxe-core` crate :) Then we can just consume them in uefi-dxe-core and apply any necessary wrappers there (similar to how [goblin](https://crates.io/crates/goblin) is 90% of the functionality of the uefi_pecoff crate in the dxe-core

cfernald:
> > Adding @cfernald here
> > @Javagedes , agreed that platforms will not customize the interrupt handling, but as to Kun's point, he, Vineel, and Chris have been trying to structure the cpu work so that the main cpu stuff can be reused across any environment, not specific to uefi (or DXE). And then there is an interface layer for the UEFI/DXE specific stuff.
> 
> I have not been following the work in these crates heavily and thus don't know the final "intentions" (nor do I have an opinion, so keep on moving forward ;) ), but I was under the impression the core reusable functionality was being done elsewhere, like mtrr crate, paging crate, etc, and that the crates in uefi-core were the final "wrappers" that implement the uefi dxe-core requirements like implementing the trait.
> 
> It's completely fine if that is not the case, and these two crates are the final "reusable" product - they should probably still be moved and renamed to not have `uefi` in it, just not into the `uefi-dxe-core` crate :) Then we can just consume them in uefi-dxe-core and apply any necessary wrappers there (similar to how [goblin](https://crates.io/crates/goblin) is 90% of the functionality of the uefi_pecoff crate in the dxe-core

Honestly, at this point i'm not sure if it matters where we put this. The mental energy to benefit of trying to find the right repo for this stuff just isn't worth it right now. I would say just co-locate as much as possible for ease-of-use at this stage and once dust settles, we can find the right distribution lines.

That said, yes the exception handlers are UEFI specific in that they expect some UEFI ABIs for convenience, but the idea was to avoid trying it to DXE exactly, as this could be potentially re-used in StMM or PEI. But so long as that model persists in the crate itself, i'm not sure I care if it's moved.

Javagedes:
@cfernald if the expectation is that this code can be re-used in StMM and/or PEI in addition to DXE then the crates can stay in uefi-core.

cfernald:
> @cfernald if the expectation is that this code can be re-used in StMM and/or PEI in addition to DXE then the crates can stay in uefi-core.

I think right now I'm fine with it going wherever it makes life easier, so long as it's dependencies are not DXE specific. Is the intention that uefi-core would contain single implementation, but phase agnostics crates?

PR #141, opened 2024-10-30
Add get_memory_descriptor_for_address(), preserve existing attributes when configuring access attributes on a block.

joschock:
## Description

* Adds `get_memory_descriptor_for_address()` function to GCD
* Updates memory protections code to preserve existing attributes when configuring access attributes on a memory block.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Booted to windows on hardware platform, confirmed issues around caching attributes no longer observed.

## Integration Instructions

N/A

### DIFF
--- a/crates/uefi_gcd/src/gcd.rs
+++ b/crates/uefi_gcd/src/gcd.rs
@@ -572,7 +572,12 @@ impl GCD {
         // we free pages, we need to reset the attrs to 0 so that the pages can be merged with other free pages
         // When mu-paging is brought in, unallocated memory will be unmapped, so this logic will look different
         // Don't check the error here, we still want to free the memory if possible
-        let _ = self.set_memory_space_attributes(base_address, len, 0);
+        let attributes = match self.get_memory_descriptor_for_address(base_address as efi::PhysicalAddress) {
+            Ok(descriptor) => descriptor.attributes & !efi::MEMORY_ACCESS_MASK,
+            Err(_) => 0,
+        };
+
+        let _ = self.set_memory_space_attributes(base_address, len, attributes);
 
         let memory_blocks = self.memory_blocks.as_mut().ok_or(Error::NotFound)?;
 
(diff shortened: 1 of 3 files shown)


crates/uefi_gcd/src/gcd.rs
```diff
@@ -1450,11 +1472,15 @@ impl SpinLockedGcd {
...
+                let attributes = match self.get_memory_descriptor_for_address(*base_address as efi::PhysicalAddress) {
+                    Ok(descriptor) => descriptor.attributes,
+                    Err(_) => 0,
+                };
                 // it is safe to call set_memory_space_attributes without calling set_memory_space_capabilities here
                 // because we set efi::MEMORY_XP as a capability on all memory ranges we add to the GCD. A driver could
                 // call set_memory_space_capabilities to remove the XP capability, but that is something that should
                 // be caught and fixed.
-                match self.set_memory_space_attributes(*base_address, len, efi::MEMORY_XP) {
+                match self.set_memory_space_attributes(*base_address, len, attributes | efi::MEMORY_XP) {
```
os-d, line 1483: I think on either allocate or free we would want the caching attributes set back to a default (WB) otherwise for general use code you could end up with weird caching semantics you aren't expecting. Again, with mu-paging this will be explicit in unmapping/mapping fresh, but I think here we still want to maintain that.

The other option would be piping this set_memory_space_attributes into calling into the cpu arch protocol to actually set attributes. We probably don't need to do that effort since we are moving away from it soon. It might be non-trivial with the locking.

So, maybe we do this for an interim state for the next amount of time until we transition to mu-paging.

joschock: Yeah, I fully expect the semantics of this to change when the paging comes in. I think that's the right place to ensure that it all is fully consistent end-to-end.

joschock: Oh, I had also forgotten - early on when investigating, I had hardcoded replacing WB on the region for this, but it caused issues because some regions and allocators (the mmio ones) don't support WB.

PR #143, opened 2024-11-02
dxe_core: local component depex support

Javagedes:
## Description

This commit updates how the dxe_core and dispatcher treat monolithically compiled components. This commit gives local components a `guid` name and a `depex` and then registers them with the dispatcher alongside components found in firmware volumes; They are evaluated the same as normal components.

When a component is registered with `with_driver`, it is now added to the `DISPATCHER_CONTEXT.pending_drivers`

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Continues to boot to UEFI Shell and passes CI

## Integration Instructions

N/A, locally compiled components still take precedence over components found into firmware volumes if no depex is provided.

### DIFF
--- a/dxe_core/src/dispatcher.rs
+++ b/dxe_core/src/dispatcher.rs
@@ -187,7 +187,6 @@ fn dispatch() -> Result<bool, efi::Status> {
                 Some(ref mut depex) => depex.eval(&PROTOCOL_DB),
                 None => dispatcher.arch_protocols_available,
             };
-            log::info!("Depex Satisfied? {:}", depex_satisfied);
 
             if depex_satisfied {
                 scheduled_driver_candidates.push(candidate)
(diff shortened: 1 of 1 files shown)


crates/uefi_component_interface/src/lib.rs
```diff
@@ -29,4 +33,8 @@ pub trait DxeComponentInterface {
 /// is satisfied for a component to be executed by the DXE Core.
 pub trait DxeComponent {
     fn entry_point(&self, interface: &dyn DxeComponentInterface) -> error::Result<()>;
+    fn guid(&self) -> efi::Guid;
+    fn depex(&self) -> Option<uefi_depex::Depex> {
+        Some(Depex::from(vec![Opcode::True, Opcode::End].as_slice()))
```
Javagedes, line 38: @joschock is uefi_depex::Depex something we want to make publicly available for this purpose? Or would an API of a `Vec<Guid>` (of required protocols) be better, and we convert that to a depex internally ourselves?

joschock: my intuition would be to expose uefi_depex - we've talked about it eventually migrating to a general crate (that, for example, could be used to make depex command line parsing utilities and stuff like that). The depex format is well-defined in the UEFI spec.

Javagedes:
Closing for now. May re-open.

PR #151, opened 2024-11-05
image loader: Fix Freeing Stack

os-d:
## Description

Currently image stacks can't be freed because the memory code cannot apply attributes across blocks, so core_free_pages returns not found. There are multiple blocks because of the stack guard having a different attribute (RP) than the rest of the stack (XP). This is fixed by removing the RP attribute and setting XP before freeing the stack, so that the blocks are coalesced and core_free_pages can free the single block.

There is a larger discussion on attribute handling in the memory subsystem that will be revisted later.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on platforms that were seeing stacks failing to be freed.

## Integration Instructions

N/A.

### DIFF
--- a/dxe_core/src/image.rs
+++ b/dxe_core/src/image.rs
@@ -102,7 +102,22 @@ impl Drop for ImageStack {
             // we added a guard page, so we need to subtract a page from the stack pointer to free everything
             let stack_addr = self.stack as *const u64 as efi::PhysicalAddress - UEFI_PAGE_SIZE as u64;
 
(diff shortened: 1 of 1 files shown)


Javagedes:
Validated I no longer see:

 ```
ERROR - core_free_pages returned error Status(
    0x800000000000000e,
) for image stack at ... for num_pages ...
```

For all components started.

PR #155, opened 2024-11-07
Add support for Security Architectural Protocols [Rebase & FF]

joschock:
## Description

Implements support for the Security Architectural Protocols in the core as specified in the [UEFI Platform Initialization Spec](https://uefi.org/specs/PI/1.8A/V2_DXE_Architectural_Protocols.html#security-architectural-protocols)

Key changes:
- improvements/extensions to device path library to support security calls.
- add support for multiple FV image sections within a single FFS file
- Add security calls for checking FVs
- Add security calls for loading images
- Add security calls for connecting controllers
- Add DxeServices->Trust() routine
- Updated Theory of Operations documentation accordingly.

- [x] Impacts functionality?
- [x] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Added rust unit tests for FV and image checks.  driver_services (connect_controller) and dxe_services (trust()) modules do not yet have testing in place, so did not extend. 

Verified via direct instrumentation in QEMU Q35 that Security Stub handlers for Security and Security2 protocols are called by the core at appropriate times. 

Verified via testing on QEMU Q35 with UEFI secure boot enabled that Security2 Architecture calls hit the DxeImageVerificationLib and successfully prevent an unsigned image from being loaded. 

## Integration Instructions

N/A

### DIFF
--- a/docs/src/dxe_core/dispatcher.md
+++ b/docs/src/dxe_core/dispatcher.md
@@ -5,12 +5,6 @@ dependencies. The Rust DXE Core dispatcher generally aligns with the requirement
 Initialization Spec for the [DXE Dispatcher](https://uefi.org/specs/PI/1.8A/V2_DXE_Dispatcher.html), with the exception
 of APRIORI file support.
 
-```admonish bug title="Security WIP"
-The security implementation for the dispatcher and Firmware Volume processing in the Rust DXE Core is under active
-development and is not yet fully defined. This documentation will be updated with details on the security properties of
-the core once that development completes.
-```
-
 ## Dispatcher Initialization
 
 The dispatcher relies on the Rust DXE Core [Event](events.md) and [Protocol](protocol_database.md) services in order to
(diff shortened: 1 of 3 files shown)


joschock:
@Javagedes , if this goes in before #143 , then that will need to be refactored somewhat FYI.

Javagedes:
> @Javagedes , if this goes in before #143 , then that will need to be refactored somewhat FYI.

That's fine. Not sure it's gonna go in anyway.

Cargo.toml
```diff
@@ -48,7 +48,7 @@ lazy_static = { version = "^1", features = ["spin_no_std"] }
 linked_list_allocator = "^0.10"
 log = { version = "^0.4", default-features = false, features = ["release_max_level_warn"]}
 mockall = "0.13.0"
-mu_pi = {git = "https://github.com/microsoft/mu_rust_pi.git", tag = "v5.0.1"}
+mu_pi = {git = "https://github.com/microsoft/mu_rust_pi.git", rev = "7e8042cb42565fd4d01268a3fedc89d0c878d3e3"}
```
os-d: Should a new tag be released?

dxe_core/src/dispatcher.rs
```diff
@@ -198,17 +200,42 @@ fn dispatch() -> Result<bool, efi::Status> {
...
+
+        if let Some(image_handle) = driver.image_handle {
+            match driver.security_status {
+                efi::Status::SUCCESS => {
+                    dispatch_attempted = true;
+                    // Note: ignore error result of core_start_image here - an image returning an error code is expected in some
+                    // cases, and a debug output for that is already implemented in core_start_image.
+                    let _status = core_start_image(image_handle);
+                }
+                efi::Status::SECURITY_VIOLATION => {
```
os-d, line 223: If we get a security violation, could we possibly end up never finishing dispatching drivers if it always returns security violation? Or do we expect to call Trust() or is there another place that the driver will get removed from the pending_drivers list? I haven't finished reviewing the PR yet, so perhaps my question will be answered.

Cargo.toml
joschock: Yes, pending: https://github.com/microsoft/mu_rust_pi/pull/33 to update the versions, I'm planning to do a release update of mu_rust_pi. It'll need a PR in uefi-core as well, I think because of the tangled web of cross-crate dependencies.

dxe_core/src/dispatcher.rs
joschock: If all the drivers the dispatcher encounters are either not depex-satisified or SECURITY_VIOLATION, then dispatch_attempted would be false and the dispatch loop would exit.

Cargo.toml
joschock: updated in latest push.

PR #158, opened 2024-11-11
fv.rs: Fix FV Protocol ReadFile() behavior

makubacki:
## Description

The `EFI_FIRMWARE_VOLUME2_PROTOCOL.ReadFile()` definition in the PI Specification states:

> "Pointer to a pointer to a buffer in which the file contents are returned, not including the file header. See 'Description' below for more details on the use of the Buffer parameter."

However, the implementation currently returns the header followed by the contents, in turn also returning the header + content size instead of just the file content size.

This change updates `read_file()` to match the PI Specification described behavior and match the C code behavior.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- Dumped file buffer read using Fv2Protocol.ReadFile() on a physical machine.

## Integration Instructions

N/A

### DIFF
--- a/dxe_core/src/fv.rs
+++ b/dxe_core/src/fv.rs
@@ -328,7 +328,7 @@ extern "efiapi" fn fv_read_file(
         found_type.write(file.file_type_raw());
         file_attributes.write(file.fv_attributes());
         //TODO: Authentication status is not yet supported.
-        buffer_size.write(file.data().len());
+        buffer_size.write(file.content().len());
     }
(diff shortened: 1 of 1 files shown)


makubacki:
Fyi: Issue filed to track adding unit tests for this code https://github.com/pop-project/uefi-dxe-core/issues/159.

PR #161, opened 2024-11-13
Add UEFI Performance Lib

magravel:
## Description

- Add protocol and function to log performance. 
- Publish the FPDT Table.

### Here is a screenshot report:
image

### Here a report:
remove the .txt extension before.
[report_rs.html.txt](https://github.com/user-attachments/files/18015150/report_rs.html.txt)

### Known issue
1. Need to fix ExitBootServices not reoprted (https://github.com/pop-project/uefi-dxe-core/issues/186)
2. Getting the name of the module. Code has moved and need more work to add this feature back. (https://github.com/pop-project/uefi-dxe-core/issues/187)
3. Getting module name for ComponentName2 Protocol. (https://github.com/pop-project/uefi-dxe-core/issues/192)
4. Getting the module name via FSS UI section. (https://github.com/pop-project/uefi-dxe-core/issues/193)
5. Verify unused performance records. https://github.com/pop-project/uefi-dxe-core/issues/195
6. Finish Module DB end perf record. https://github.com/pop-project/uefi-dxe-core/issues/194

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This was tested by running in QEMU (Shell and can boot to OS) and on real hardware (intel device).

## Integration Instructions

N/A

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,13 +22,13 @@ uefi_debugger = { git = "https://github.com/pop-project/uefi-core.git", rev = "v
 section_extractor = { git = "https://github.com/pop-project/uefi-core.git", rev = "v7.0.2" }
 corosensei = { git = "https://github.com/pop-project/uefi-corosensei.git", rev = "d27241e17a5f3c0703fe9871e6f68cd838c05c6d", default-features = false }
 
-uefi_test_macro = { path = "crates/uefi_test_macro", version = "5.0.0" }
-uefi_test = { path = "uefi_test", version = "5.0.0" }
-uefi_collections = { path = "crates/uefi_collections", version = "5.0.0", default-features = false }
-uefi_depex = { path = "crates/uefi_depex", version = "5.0.0" }
-uefi_device_path = { path = "crates/uefi_device_path", version = "5.0.0" }
-uefi_component_interface = { path = "crates/uefi_component_interface", version = "5.0.0" }
-uefi_performance = { path = "crates/uefi_performance", version = "5.0.0" }
+uefi_test_macro = { path = "crates/uefi_test_macro", version = "6.1.0" }
+uefi_test = { path = "uefi_test", version = "6.1.0" }
+uefi_collections = { path = "crates/uefi_collections", version = "6.1.0", default-features = false }
+uefi_depex = { path = "crates/uefi_depex", version = "6.1.0" }
+uefi_device_path = { path = "crates/uefi_device_path", version = "6.1.0" }
+uefi_component_interface = { path = "crates/uefi_component_interface", version = "6.1.0" }
+uefi_performance = { path = "crates/uefi_performance", version = "6.1.0" }
 
 adv_logger = { path = "adv_logger" }
 sample_components = { path = "sample_components" }


crates/uefi_performance/src/_debug.rs
```diff
@@ -0,0 +1,61 @@
...
+            writeln!(f)?;
+        }
+        write!(f, "]")?;
+        Ok(())
+    }
+}
+
+pub struct DbgGuid<'a>(pub &'a efi::Guid);
+
+impl Debug for DbgGuid<'_> {
```
joschock: for consistency, consider replacing this with the `guid_fmt!` macro that is used elsewhere in the code.

crates/uefi_performance/src/_utils.rs
```diff
@@ -0,0 +1,46 @@
...
+}
+
+pub fn c_char_ptr_from_str(str: &str) -> *const c_char {
+    let mut s = String::from(str);
+    s.push(0 as char);
+    s.as_ptr() as *const c_char
+}
+
+#[macro_export]
+macro_rules! __func__ {
```
joschock: Consider replacing this with `function!()` macro used elsewhere in the core.

crates/uefi_performance/src/_utils.rs
```diff
@@ -0,0 +1,46 @@
...
+    () => {{
+        fn __func__() {}
+        core::any::type_name_of_val(&__func__).strip_suffix("::__func__").unwrap()
+    }};
+}
+
+pub fn spin_lock<'a, T: ?Sized, B: BootServices>(mutex: &'a TplMutex<'a, T, B>) -> TplMutexGuard<'a, T, B> {
```
joschock: I'm not sure this is needed. Since UEFI is inherently single threaded (but interruptible), I think if mutex.try_lock() fails because another context is holding the lock, that context must be at a lower TPL (or it would not have been interrupted by this context), and that means if this context spins waiting for the lock, the lower context will never get control to release it. So, consider just removing this function and calling lock() directly on the mutex.

crates/uefi_performance/src/lib.rs
```diff
@@ -0,0 +1,748 @@
...
+    let _status = unsafe {
+        BOOT_SERVICES.install_configuration_table_unchecked(
+            &guid::EDKII_FPDT_EXTENDED_FIRMWARE_PERFORMANCE_GUID,
+            fbpt.fbpt_address() as *mut c_void,
+        )
+    };
+}
+
+extern "efiapi" fn update_boot_performance_table(_event: efi::Event, _: &()) {
+    // todo
```
joschock: Is there a task filed on this one?

crates/uefi_performance/src/protocols/_status_code_runtime.rs
```diff
@@ -0,0 +1,28 @@
+use r_efi::efi;
```
joschock: This protocol defn. is part of PI spec (https://uefi.org/specs/PI/1.8A/V3_Report_Status_Code_Router.html#report-status-code-handler-protocol) so should probably move to mu_rust_pi

crates/uefi_performance/src/_debug.rs
magravel: Could be nice to improve in the future, we need to add uuid crate to the project to be able to use the macro.

crates/uefi_performance/src/_utils.rs
magravel: I'll remove it, I once had a double lock, but this wasn't a fix. The double lock was elsewhere.

crates/uefi_performance/src/lib.rs
magravel: Yes

Cargo.toml
```diff
@@ -15,8 +15,13 @@ license = "BSD-2-Clause-Patent"
 edition = "2021"
 rust-version = "1.80.0"
 
+
+[patch."https://github.com/pop-project/uefi-sdk"]
```
joschock: need to pull this out before committing

crates/uefi_performance/src/_utils.rs
```diff
@@ -0,0 +1,27 @@
...
+
+    let mut str = String::new();
+    loop {
+        let c = unsafe { ptr::read(c_ptr) };
+        if c == 0 {
+            break;
+        }
+        str.push(c as u8 as char);
+        c_ptr = unsafe { c_ptr.add(1) };
+    }
```
joschock: consider using `core::ffi::Cstr` to convert?

crates/uefi_performance/src/lib.rs
```diff
@@ -0,0 +1,749 @@
...
+
+    let (pei_records, pei_load_image_count) = extract_pei_performance_records(hob_list)?;
+    LOAD_IMAGE_COUNT.store(pei_load_image_count, Ordering::Relaxed);
+    log::info!("{} PEI Records found.", pei_records.iter().count());
+    FBPT.lock().set_records(pei_records);
+
+    // Install the protocol interfaces for DXE performance library instance.
+    static PERFORMANCE_MEASUREMENT_INTERFACE: EdkiiPerformanceMeasurementInterface =
+        EdkiiPerformanceMeasurementInterface { create_performance_measurement };
+    unsafe {
```
joschock: can you use the safe version of install_protocol_interface?

crates/uefi_performance/src/lib.rs
```diff
@@ -0,0 +1,749 @@
...
+
+    let cpu_count = Arch::cpu_count();
+    let timestamp = match ticker {
+        0 => (cpu_count as f64 / Arch::cpu_count_frequency() as f64 * 1_000_000_000_f64) as u64,
+        1 => 0,
+        ticker => (ticker as f64 / Arch::cpu_count_frequency() as f64 * 1_000_000_000_f64) as u64,
+    };
+
+    match perf_id {
+        // TODO This path has not been tested (not used for now).
```
joschock: Is this comment still accurate given the integration testing in qemu and hardware? I think they generate these records...

crates/uefi_performance/src/lib.rs
```diff
@@ -0,0 +1,749 @@
...
+        PerfId::MODULE_DB_END => {
+            if let Ok((module_name, guid)) =
+                get_module_info_from_handle(&BOOT_SERVICES, caller_identifier as *mut c_void)
+            {
+                let module_name = module_name.unwrap_or(String::from("unknown name"));
+                let record = GuidQwordStringEventRecord::new(perf_id, 0, timestamp, guid, address as u64, &module_name);
+                _ = &FBPT.lock().add_record(record);
+            }
+            // TODO something weird if address is not 0 need example to continue development.
```
joschock: is there a bug filed on this one?

crates/uefi_performance/src/lib.rs
```diff
@@ -0,0 +1,749 @@
...
+            return efi::Status::INVALID_PARAMETER;
+        }
+    };
+
+    efi::Status::SUCCESS
+}
+
+extern "efiapi" fn report_fpdt_record_buffer(_event: efi::Event, _ctx: &()) {
+    let fbpt = &mut FBPT.lock();
+    fbpt.report_table(&BOOT_SERVICES, &RUNTIME_SERVICES).unwrap();
```
joschock: consider `expect` instead of unwrap?

crates/uefi_performance/src/_status_code_runtime.rs
```diff
@@ -0,0 +1,84 @@
...
+        let header = EfiStatusCodeData { header_size: header_size as u16, size: data_size as u16, r#type: data_type };
+
+        let mut data_buffer = Vec::from(unsafe { any_as_u8_slice(&header) });
+        data_buffer.extend(unsafe { any_as_u8_slice(&data) });
+
+        let data_ptr: *mut EfiStatusCodeData = data_buffer.as_mut_ptr() as *mut EfiStatusCodeData;
+
+        let caller_id = caller_id.unwrap_or(&mu_rust_helpers::guid::CALLER_ID);
+
+        // debugln!(DEBUG_INFO, "[RustStatusCodeRuntime] caller_id: {}", guid::guid_fmt!(caller_id));
```
joschock: recommend remove this comment (and corresponding cspell entry), since we use `log` crate now

dxe_core/src/lib.rs
```diff
@@ -104,15 +104,23 @@ mod tpl_lock;
...
+    fw_fs,
+    hob::HobList,
+    protocols::{bds, status_code},
+    status_code::{EFI_PROGRESS_CODE, EFI_SOFTWARE_DXE_CORE, EFI_SW_DXE_CORE_PC_HANDOFF_TO_NEXT},
+};
+use r_efi::efi;
 use uefi_component_interface::DxeComponent;
 use uefi_sdk::error::{self, Result};
 
+pub const DXE_CORE_GUID: efi::Guid =
```
joschock: this can come from uefi_sdk - it has a definition now: https://github.com/pop-project/uefi-sdk/blob/04766214b55146d14b953182e0b5d29bab55d3e1/uefi_sdk/src/guid.rs#L33

crates/uefi_performance/src/guid.rs
```diff
@@ -0,0 +1,19 @@
...
+pub const EVENT_READY_TO_BOOT_GUID: efi::Guid =
+    efi::Guid::from_fields(0x7CE88FB3, 0x4BD7, 0x4679, 0x87, 0xA8, &[0xA8, 0xD8, 0xDE, 0xE5, 0x0D, 0x2B]);
+
+pub const PERFORMANCE_PROTOCOL_GUID: efi::Guid =
+    efi::Guid::from_fields(0x76b6bdfa, 0x2acd, 0x4462, 0x9E, 0x3F, &[0xcb, 0x58, 0xC9, 0x69, 0xd9, 0x37]);
+
+pub const PERFORMANCE_EX_PROTOCOL_GUID: efi::Guid =
+    efi::Guid::from_fields(0x1ea81bec, 0xf01a, 0x4d98, 0xa2, 0x1, &[0x4a, 0x61, 0xce, 0x2f, 0xc0, 0x22]);
+
+pub const SMM_COMMUNICATION_PROTOCOL_GUID: efi::Guid =
```
os-d: These guids should probably go in uefi-sdk so that others can use them. Also, some of these are already defined in r-efi, right? Ready to boot at least: https://github.com/r-efi/r-efi/blob/1c1a4062a0298cf1cd6f0ed6b7f9ffbde0f82c73/src/system.rs#L267).

crates/uefi_performance/src/_status_code_runtime.rs
```diff
@@ -0,0 +1,84 @@
+extern crate alloc;
+
+use core::{mem, ops::Deref, slice};
+
```
os-d, line 11: Looks like we are duplicating functionality. Is there a way we can move this to a central place? Is the `_` prefix for the filename indicating this will be dropped later when this is open source? https://github.com/microsoft/mu_plus/blob/dev/202405/MsWheaPkg/Crates/MuTelemetryHelperLib/src/status_code_runtime.rs

crates/uefi_performance/src/lib.rs
```diff
@@ -0,0 +1,749 @@
...
+    Ok(())
+}
+
+pub fn extract_pei_performance_records(hob_list: &HobList) -> Result<(PerformanceRecordBuffer, u32), efi::Status> {
+    let mut pei_records = PerformanceRecordBuffer::new();
+    let mut pei_load_image_count = 0;
+
+    for hob in hob_list.iter() {
+        let guid_hob = match *hob {
+            Hob::GuidHob(hob, _) if hob.name == EDKII_FPDT_EXTENDED_FIRMWARE_PERFORMANCE_GUID => hob,
```
os-d: @cfernald for reference, a usage of HOBs outside of strictly core stuff, for how it plays into your scheme.

crates/uefi_performance/src/performance_record.rs
```diff
@@ -0,0 +1,245 @@
...
+
+use alloc::vec::Vec;
+use core::{fmt::Debug, mem, ops::Deref};
+
+use r_efi::efi;
+use scroll::{self, Pread, Pwrite};
+
+use crate::_debug::DbgMemory;
+
+pub const HEADER_SIZE: usize = mem::size_of::<u16>() // Type
```
os-d: nit: since this is public, I would give it a more specific name to avoid clashing later (yes, it is scoped behind this mod, but it is a common pattern to `use` an entire mod)

crates/uefi_performance/src/performance_table.rs
```diff
@@ -0,0 +1,246 @@
...
+        allocation::{AllocType, MemoryType},
+        BootServices,
+    },
+    runtime_services::RuntimeServices,
+};
+
+use crate::performance_record::{self, PerformanceRecord, PerformanceRecordBuffer};
+
+const PUBLISHED_FBPT_EXTRA_SPACE: usize = 0x10_000;
+const PAGE_SIZE_4K: usize = 0x1000;
```
os-d: This is defined in uefi_sdk now: https://github.com/pop-project/uefi-sdk/blob/04766214b55146d14b953182e0b5d29bab55d3e1/uefi_sdk/src/base.rs#L14

crates/uefi_performance/src/lib.rs
magravel: Need to double check all those on real hardware

crates/uefi_performance/src/lib.rs
magravel: Need to add one. This code was hard to translate from c and need a real use case to be codded.

crates/uefi_performance/src/_status_code_runtime.rs
magravel: I put _ in front of every module that I felt like should be somewhere else.

crates/uefi_performance/src/_debug.rs
makubacki, line 1: Can you please add copyright headers to each new file?

crates/uefi_performance/src/guid.rs
makubacki: Yeah, this needs to be cleaned up before completion.

crates/uefi_performance/src/guid.rs
```diff
@@ -0,0 +1,23 @@
...
+//!
+//! Copyright (C) Microsoft Corporation. All rights reserved.
+//!
+//! SPDX-License-Identifier: BSD-2-Clause-Patent
+//!
+
+use r_efi::efi;
+
+pub const EDKII_FPDT_EXTENDED_FIRMWARE_PERFORMANCE_GUID: efi::Guid =
+    efi::Guid::from_fields(0x3b387bfd, 0x7abc, 0x4cf2, 0xa0, 0xca, &[0xb6, 0xa1, 0x6c, 0x1b, 0x1b, 0x25]);
```
zurcher: Consider using `guid!` macro

Cargo.toml
```diff
@@ -28,6 +28,7 @@ uefi_collections = { path = "crates/uefi_collections", version = "5.0.0", defaul
 uefi_depex = { path = "crates/uefi_depex", version = "5.0.0" }
 uefi_device_path = { path = "crates/uefi_device_path", version = "5.0.0" }
 uefi_component_interface = { path = "crates/uefi_component_interface", version = "5.0.0" }
+uefi_performance = { path = "crates/uefi_performance", version = "0.1.0" }
```
makubacki: We're keeping all crates in the repo at the same version.

```suggestion
uefi_performance = { path = "crates/uefi_performance", version = "5.0.0" }
```

crates/uefi_performance/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[package]
+name = "uefi_performance"
+version = "0.1.0"
```
makubacki:
```suggestion
version = "5.0.0"
```

crates/uefi_performance/src/lib.rs
```diff
@@ -0,0 +1,757 @@
+//! ## License
```
makubacki: Can you please add an overview of the library here to capture in documentation?

crates/uefi_performance/src/_status_code_runtime.rs
makubacki, line 1: Was there a particular reason for temporarily adding the `StatusCodeRuntimeProtocol` trait here versus mu_rust_pi since the protocol definition is already there? Was it just time?

crates/uefi_performance/src/performance_table.rs
```diff
@@ -0,0 +1,253 @@
...
+const PUBLISHED_FBPT_EXTRA_SPACE: usize = 0x10_000;
+
+#[derive(Debug, Clone, Pwrite)]
+#[repr(C)]
+pub struct PerformanceTableHeader {
+    pub signature: u32,
+    pub length: u32,
+}
+
+pub struct FBPT {
```
makubacki, line 39: I think having an acronym as the name is okay, but please add a brief doc comment to explain it for others in the future. Will also appear on hover in the IDE.

crates/uefi_performance/src/_status_code_runtime.rs
magravel: We could open a task for that, we need to add some dependency to mu_rust_helper to add this abstraction in.

PR #164, opened 2024-11-14
Add calls to static debugger routines in DxeCore

cfernald:
## Description

Adds calls to the static routines in uefi_debugger of initializing the debugger and notifying the debugger of image loads

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on Q35

## Integration Instructions

N/A


dxe_core/src/lib.rs
```diff
@@ -132,6 +132,8 @@ where
     pub fn initialize(mut self, physical_hob_list: *const c_void) -> CorePostInit {
         self.cpu_initializer.initialize();
         self.interrupt_manager.initialize().expect("Failed to initialize interrupt manager!");
+        uefi_debugger::initialize(&mut self.interrupt_manager);
```
vineelko, line 180: Do we call `uefi_debugger::set_debugger(&DEBUGGER);` in the rust bins repo?

cfernald: Yeah, the crate will define the static instance of the debugger and call set_debugger. It is more or less the example shown in the uefi_debugger crate doc: https://github.com/pop-project/uefi-core/blob/2b04f109ce711dbd3d92ea0e96a8a23aea2a4af3/uefi_debugger/src/lib.rs#L33

cfernald:
Will fix the crate version in an PR later today, where I will rev the version.

PR #165, opened 2024-11-15
uefi-dxe-core: CPU Arch Protocol Implementation

vineelko:
## Description

Initial set of Cpu Arch Protocol changes in `uefi-dxe-core` repo. GCD integration is ongoing.

- [x] Impacts functionality?
- [ ] Impacts security?
- [x] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- Unit tested and QEMU boots as normal.

## Integration Instructions

The changes span across 3 different repos, `uefi-core`, `uefi-dxe-core`, and
`qemu_rust_bins`. This will be a breaking change and requires a version upgrade
to refer to other changes. Currently the cargo.toml refers to local paths,
evaluating the best approach to merge these changes.

https://github.com/pop-project/uefi-core/pull/23
https://github.com/pop-project/uefi-dxe-core/pull/165
https://microsoft.visualstudio.com/MsUEFI/_git/qemu_rust_bins/pullrequest/11815622

### DIFF
--- a/dxe_core/src/cpu_arch_protocol.rs
+++ b/dxe_core/src/cpu_arch_protocol.rs
@@ -14,7 +14,7 @@ use mu_pi::protocols::cpu_arch::{CpuFlushType, CpuInitType, InterruptHandler, Pr
 pub struct EfiCpuArchProtocolImpl<'a> {
     protocol: Protocol,
 
-    // Private fields
+    // Crate accessible fields
     pub(crate) cpu_init: &'a mut dyn EfiCpuInit,
     pub(crate) interrupt_manager: &'a mut dyn InterruptManager,
 }
@@ -22,7 +22,7 @@ pub struct EfiCpuArchProtocolImpl<'a> {
 // Helper function to convert a raw mutable pointer to a mutable reference.
 fn get_impl_ref<'a>(this: *const Protocol) -> &'a EfiCpuArchProtocolImpl<'a> {
     if this.is_null() {
-        panic!("Null pointer passed to get_as_ref()");
+        panic!("Null pointer passed to get_impl_ref()");
     }
 
     unsafe { &*(this as *const EfiCpuArchProtocolImpl<'a>) }
@@ -30,7 +30,7 @@ fn get_impl_ref<'a>(this: *const Protocol) -> &'a EfiCpuArchProtocolImpl<'a> {
 
 fn get_impl_ref_mut<'a>(this: *mut Protocol) -> &'a mut EfiCpuArchProtocolImpl<'a> {
     if this.is_null() {
-        panic!("Null pointer passed to get_as_ref()");
+        panic!("Null pointer passed to get_impl_ref_mut()");
     }
 
     unsafe { &mut *(this as *mut EfiCpuArchProtocolImpl<'a>) }
@@ -337,6 +337,8 @@ mod tests {
         assert_eq!(timer_value, 0);
         assert_eq!(timer_period, 0);
     }
+
+    // TODO: Following tests will be enabled once the GCD integration is done.
     // #[test]
     // fn test_set_memory_attributes() {
     //     let mut cpu_init = MockMockEfiCpuPaging::new();


makubacki:
@vineelko, for this:

> "The changes span across 3 different repos, uefi-core, uefi-dxe-core, and
qemu_rust_bins. This will be a breaking change and requires a version upgrade
to refer to other changes. Currently the cargo.toml refers to local paths,
evaluating the best approach to merge these changes."

Normally, you would start from the initial dependency (`uefi-core`) and update versions from there. So, `uefi-core` merges w/ version update and release, `uefi-dxe-core` references that version merges w/ version update and release, `qemu_rust_bins` references both of those versions and does the same. Are you saying that it is more complicated than that in this case?

These don't need to be local paths. Your PR branch in `uefi-core` itself provides a git commit hash that can be referenced in the `rev` dependency in this repo's Cargo.toml to resolve the dependency:

``` toml
uefi_cpu_init = {git = "https://github.com/pop-project/uefi-core.git", rev = "c0188ad05c1c08c45c219b9fd2afcf87e534fa8d"}
uefi_core = {git = "https://github.com/pop-project/uefi-core.git", rev = "c0188ad05c1c08c45c219b9fd2afcf87e534fa8d"}
uefi_interrupt = {git = "https://github.com/pop-project/uefi-core.git", rev = "c0188ad05c1c08c45c219b9fd2afcf87e534fa8d"}
uefi_logger = {git = "https://github.com/pop-project/uefi-core.git", rev = "c0188ad05c1c08c45c219b9fd2afcf87e534fa8d"}
section_extractor = {git = "https://github.com/pop-project/uefi-core.git", rev = "c0188ad05c1c08c45c219b9fd2afcf87e534fa8d"}
serial_writer = {git = "https://github.com/pop-project/uefi-core.git", rev = "c0188ad05c1c08c45c219b9fd2afcf87e534fa8d"}
```

dxe_core/src/cpu_arch_protocol.rs
```diff
@@ -0,0 +1,530 @@
...
+/// Architecture independent public C EFI CPU Architectural Protocol definition.
+use crate::protocols::PROTOCOL_DB;
+use alloc::boxed::Box;
+use core::ffi::c_void;
+use r_efi::efi;
+use uefi_cpu_init::{EfiCpuFlushType, EfiCpuInit, EfiCpuInitType, EfiPhysicalAddress};
+use uefi_interrupt::{efi_system_context::EfiSystemContext, InterruptManager};
+use uefi_sdk::error::EfiError;
+
+pub const EFI_CPU_ARCH_PROTOCOL_GUID: efi::Guid =
```
makubacki: Are you intentionally not using the definition in [mu_rust_pi](https://github.com/microsoft/mu_rust_pi/blob/22a4459bd1fd7e9e7719956d73bce0bdd16e0daf/src/protocols/cpu_arch.rs#L22)?

vineelko: Thanks for pointing that out, Michael, let me reconcile with it.

docs/src/driver/interface.md
makubacki, line 1: It looks like the line endings were updated in this file with the edits. Was that intended?

dxe_core/src/cpu_arch_protocol.rs
```diff
@@ -0,0 +1,355 @@
...
+
+    // Private fields
+    pub(crate) cpu_init: &'a mut dyn EfiCpuInit,
+    pub(crate) interrupt_manager: &'a mut dyn InterruptManager,
+}
+
+// Helper function to convert a raw mutable pointer to a mutable reference.
+fn get_impl_ref<'a>(this: *const Protocol) -> &'a EfiCpuArchProtocolImpl<'a> {
+    if this.is_null() {
+        panic!("Null pointer passed to get_as_ref()");
```
makubacki:
```suggestion
        panic!("Null pointer passed to get_impl_ref()");
```

dxe_core/src/cpu_arch_protocol.rs
```diff
@@ -0,0 +1,355 @@
...
+    }
+
+    unsafe { &*(this as *const EfiCpuArchProtocolImpl<'a>) }
+}
+
+fn get_impl_ref_mut<'a>(this: *mut Protocol) -> &'a mut EfiCpuArchProtocolImpl<'a> {
+    if this.is_null() {
+        panic!("Null pointer passed to get_as_ref()");
```
makubacki:
```suggestion
        panic!("Null pointer passed to get_impl_ref_mut()");
```

dxe_core/src/cpu_arch_protocol.rs
```diff
@@ -0,0 +1,355 @@
...
+use uefi_interrupt::InterruptManager;
+use uefi_sdk::error::EfiError;
+
+use mu_pi::protocols::cpu_arch::{CpuFlushType, CpuInitType, InterruptHandler, Protocol, PROTOCOL_GUID};
+
+#[repr(C)]
+pub struct EfiCpuArchProtocolImpl<'a> {
+    protocol: Protocol,
+
+    // Private fields
```
makubacki: Nit, but I think this is a little clearer if the comment is there.

```suggestion
    // Crate accessible fields
```

dxe_core/src/cpu_arch_protocol.rs
```diff
@@ -0,0 +1,355 @@
...
+
+    result.map(|_| efi::Status::SUCCESS).unwrap_or_else(|err| err)
+}
+
+extern "efiapi" fn register_interrupt_handler(
+    this: *const Protocol,
+    interrupt_type: isize,
+    interrupt_handler: InterruptHandler,
+) -> efi::Status {
+    // TODO: Fix interrupt manager to use mu_pi::protocols::cpu_arch::InterruptHandler
```
makubacki, line 95: How long is this work expected to take?

dxe_core/src/cpu_arch_protocol.rs
```diff
@@ -0,0 +1,355 @@
...
+                timer_index,
+                &mut timer_value as *mut _,
+                &mut timer_period as *mut _,
+            )
+        };
+        assert_eq!(status, efi::Status::SUCCESS);
+        assert_eq!(timer_value, 0);
+        assert_eq!(timer_period, 0);
+    }
+    // #[test]
```
makubacki, line 342: If you're holding this for the GCD change, can you please add a TODO? We have a task to track those so it would explain the commented code in case someone else is cleaning up dead code.

dxe_core/src/cpu_arch_protocol.rs
vineelko: Will fixed it in the next iteration.

dxe_core/src/cpu_arch_protocol.rs
vineelko: Will fixed it in the next iteration.

dxe_core/src/cpu_arch_protocol.rs
vineelko: Will fixed it in the next iteration.

dxe_core/src/cpu_arch_protocol.rs
vineelko: I would let @cfernald comment on that.

docs/src/driver/interface.md
vineelko: Some of the doc files are CRLF and few of them are LF ended. 
image
image

These files are set to CRLF.

dxe_core/src/cpu_arch_protocol.rs
vineelko: Yes Michael, Will add a todo.

docs/src/driver/interface.md
makubacki: As mentioned offline, we're specifying code line endings as CRLF via rustfmt. For documentation, I don't think it really matters for now. We can format all of the files consistently when history is cleaned up later. The squash removes the noise from things like this.

makubacki: But please leave them as CRLF in this PR.

dxe_core/src/cpu_arch_protocol.rs
cfernald: We can probably start the cpu crate refactor here whenever now that the dust is settling a bit.

cfernald: Though we need to update that definition to not use the *void first, I can look into that

vineelko: Yaa the reason for the comment is, this interface is based on mu_rust_pi definitions (as per Michael's original comment on the PR) Now, to forward this call to Interrupt Manager the `register_exception_handler(..)` should be based on mu_rust_pi definitions, right? Sure, we can start the cpu crate refactor once the uefi-rust-bins PR is also merged so that Oliver will not be blocked for GCD integrations(will have less rebases).

PR #168, opened 2024-11-19
Prevent memory map changes following ExitBootServices

berlin-with0ut-return:
## Description
According to UEFI spec, the memory map (represented by a key) should not change after exit_boot_services is called. To preserve this invariant, we add a guard against additional allocations/frees after EBS within the GCD.

- [X] Impacts functionality?
- [ ] Impacts security? (possibly?)
- [x] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested
All tests in the uefi_dxe_core repo still pass.
Additionally, I added a "forced" EBS call in the HelloWorld driver and tested allocating/freeing there following EBS. Both calls were caught by the new GCD functions (verified using logs).

Booting to Windows in UefiRust still functions as expected, so this should not break any functionality.

## Integration Instructions
Can be merged without issue.

### DIFF
--- a/dxe_core/src/gcd/spin_locked_gcd.rs
+++ b/dxe_core/src/gcd/spin_locked_gcd.rs
@@ -1469,6 +1469,7 @@ impl SpinLockedGcd {
     /// This call potentially invalidates all allocations made by any allocator on top of this GCD.
     /// Caller is responsible for ensuring that no such allocations exist.
     ///
+    #[cfg(test)]
     pub unsafe fn reset(&self) {
         let (mut mem, mut io) = (self.memory.lock(), self.io.lock());
         mem.maximum_address = 0;
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,9 +45,7 @@ goblin = { version = "^0.9", default-features = false, features = [
 ] }
 lazy_static = { version = "^1", features = ["spin_no_std"] }
 linked_list_allocator = "^0.10"
-log = { version = "^0.4", default-features = false, features = [
-    "release_max_level_warn",
-] }
+log = { version = "^0.4", default-features = false }
 mockall = "0.13.0"
 mu_pi = { git = "https://github.com/microsoft/mu_rust_pi.git", tag = "v5.1.0" }
 mu_rust_helpers = { git = "https://github.com/microsoft/mu_rust_helpers.git", tag = "v1.2.0" }
--- a/dxe_core/src/allocator/uefi_allocator.rs
+++ b/dxe_core/src/allocator/uefi_allocator.rs
@@ -58,6 +58,7 @@ impl UefiAllocator {
     /// Indicates whether the given pointer falls within a memory region managed by this allocator.
     ///
     /// See [`SpinLockedFixedSizeBlockAllocator::contains`]
+    #[allow(dead_code)]
     pub fn contains(&self, ptr: NonNull<u8>) -> bool {
         self.allocator.contains(ptr)
     }
@@ -199,6 +200,7 @@ impl UefiAllocator {
     }
 
     /// Returns the preferred memory range, if any.
+    #[allow(dead_code)]
     pub fn preferred_range(&self) -> Option<Range<efi::PhysicalAddress>> {
         self.allocator.preferred_range()
     }
--- a/dxe_core/src/event_db.rs
+++ b/dxe_core/src/event_db.rs
@@ -378,6 +378,7 @@ impl EventDb {
         Ok(self.events.get(&id).ok_or(efi::Status::INVALID_PARAMETER)?.event_type)
     }
 
+    #[allow(dead_code)]
     fn get_notification_data(&mut self, event: efi::Event) -> Result<EventNotification, efi::Status> {
         let id = event as usize;
         if let Some(found_event) = self.events.get(&id) {
@@ -597,10 +598,7 @@ impl SpinLockedEventDb {
     }
 
     /// Indicates whether the given event is in the signaled state
-    ///
-    /// ## Errors
-    ///
-    /// Returns r_efi:efi::Status::INVALID_PARAMETER if incorrect parameters are given.
+    #[allow(dead_code)]
     pub fn is_signaled(&self, event: efi::Event) -> bool {
         self.lock().is_signaled(event)
     }
@@ -610,6 +608,7 @@ impl SpinLockedEventDb {
     /// ## Errors
     ///
     /// Returns r_efi:efi::Status::INVALID_PARAMETER if incorrect parameters are given.
+    #[allow(dead_code)]
     pub fn clear_signal(&self, event: efi::Event) -> Result<(), efi::Status> {
         self.lock().clear_signal(event)
     }
@@ -644,6 +643,7 @@ impl SpinLockedEventDb {
     /// ## Errors
     ///
     /// Returns r_efi:efi::Status::INVALID_PARAMETER if incorrect parameters are given.
+    #[allow(dead_code)]
     pub fn get_notification_data(&self, event: efi::Event) -> Result<EventNotification, efi::Status> {
         self.lock().get_notification_data(event)
     }
--- a/dxe_core/src/gcd/io_block.rs
+++ b/dxe_core/src/gcd/io_block.rs
@@ -222,6 +222,8 @@ impl IoBlock {
     pub fn len(&self) -> usize {
         self.as_ref().length as usize
     }
+
+    #[allow(dead_code)]
     pub fn is_empty(&self) -> bool {
         self.len() == 0
     }
--- a/dxe_core/src/gcd/memory_block.rs
+++ b/dxe_core/src/gcd/memory_block.rs
@@ -311,6 +311,7 @@ impl MemoryBlock {
         self.as_ref().length as usize
     }
 
+    #[allow(dead_code)]
     pub fn is_empty(&self) -> bool {
         self.len() == 0
     }
--- a/dxe_core/src/protocol_db.rs
+++ b/dxe_core/src/protocol_db.rs
@@ -570,6 +570,7 @@ impl SpinLockedProtocolDb {
     ///
     /// This call completely resets the protocol database and is intended mostly for use in test.
     ///
+    #[cfg(test)]
     pub unsafe fn reset(&self) {
         let mut inner = self.inner.lock();
         inner.handles.clear();


dxe_core/src/allocator.rs
```diff
@@ -352,13 +352,15 @@ pub fn core_allocate_pool(pool_type: efi::MemoryType, size: usize) -> Result<*mu
 }
 
 extern "efiapi" fn free_pool(buffer: *mut c_void) -> efi::Status {
+    log::info!("Free pool called");
```
Javagedes: Either this is extra logging from your test that needs to be removed, or this needs to be a trace statement, and probably needs to log the address of the buffer being freed. Same with core_free_pool

Makefile.toml
```diff
@@ -104,7 +104,7 @@ args = ["check", "--target", "x86_64-unknown-uefi", "@@split(NO_STD_FLAGS, )", "
 description = "Checks rust code for std build errors with results."
 private = true
 command = "cargo"
-args = ["check", "@@split(STD_FLAGS, )", "@@split(CARGO_MAKE_TASK_ARGS, )"]
+args = ["check", "@@split(STD_FLAGS, )", "@@split(CARGO_MAKE_TASK_ARGS, )", "--tests"]
```
Javagedes: This should probably be a standalone PR. I'm sure it required extra changes is tests outside this PR.

crates/uefi_gcd/src/gcd.rs
```diff
@@ -161,20 +161,156 @@ pub fn get_capabilities(gcd_mem_type: dxe_services::GcdMemoryType, attributes: u
     capabilities
```
os-d: As a general note, this should be added to our "things that need to happen before you transition to DxeRust in the mdbook". @Javagedes for reference

crates/uefi_gcd/src/gcd.rs
```diff
@@ -161,20 +161,156 @@ pub fn get_capabilities(gcd_mem_type: dxe_services::GcdMemoryType, attributes: u
...
     memory_blocks: Option<Rbt<'static, MemoryBlock>>,
+    allocate_memory_space_fn: GcdAllocateFn,
+    free_memory_space_fn: GcdFreeFn,
+}
+
+/// This service allocates nonexistent memory, reserved memory, system memory, or memory-mapped I/O resources from the global coherency domain of the processor.
+///
+/// # Documentation
+/// UEFI Platform Initialization Specification, Release 1.8, Section II-7.2.4.2
+fn allocate_memory_space_before_ebs(
```
os-d: nit: because this follows a PI spec function, I would leave the name the same `allocate_memory_space`. In general I would probably keep the existing names the same and then the after_ebs ones call Null or something to be a bit more generic (maybe we would have other allocation blackout windows).

crates/uefi_gcd/src/gcd.rs
```diff
@@ -161,20 +161,156 @@ pub fn get_capabilities(gcd_mem_type: dxe_services::GcdMemoryType, attributes: u
...
+fn allocate_memory_space_after_ebs(
+    _gcd: &mut GCD,
+    _allocate_type: AllocateType,
+    _memory_type: dxe_services::GcdMemoryType,
+    _alignment: usize,
+    _len: usize,
+    _image_handle: efi::Handle,
+    _device_handle: Option<efi::Handle>,
+) -> Result<usize, Error> {
+    log::warn!("GCD not allowed to allocate after EBS has started!");
```
os-d:
```suggestion
    log::error!("GCD not allowed to allocate after EBS has started!");
    debug_assert!(false);

```
This one I would add an assert because we are explicitly failing; any caller is going to fail and maybe not gracefully, so we definitely want to catch this.

crates/uefi_gcd/src/gcd.rs
```diff
@@ -161,20 +161,156 @@ pub fn get_capabilities(gcd_mem_type: dxe_services::GcdMemoryType, attributes: u
...
+    }
+}
+
+fn free_memory_space_worker_after_ebs(
+    _gcd: &mut GCD,
+    _base_address: usize,
+    _len: usize,
+    _transition: MemoryStateTransition,
+) -> Result<(), Error> {
+    log::warn!("GCD not allowed to free after EBS has started!");
```
os-d:
```suggestion
    log::error!("GCD not allowed to free after EBS has started!");

```

Depending on how your investigation goes, we may want to add a debug_assert here for platforms to catch and cleanup (but wouldn't want long term, so maybe we don't care).

crates/uefi_gcd/src/gcd.rs
```diff
@@ -161,20 +161,156 @@ pub fn get_capabilities(gcd_mem_type: dxe_services::GcdMemoryType, attributes: u
...
+            maximum_address: 1 << processor_address_bits,
+            allocate_memory_space_fn: allocate_memory_space_before_ebs,
+            free_memory_space_fn: free_memory_space_worker_before_ebs,
+        }
+    }
+
+    pub fn signal_ebs_start(&mut self) {
+        self.allocate_memory_space_fn = allocate_memory_space_after_ebs;
+        self.free_memory_space_fn = free_memory_space_worker_after_ebs;
+        log::info!("Reassigning GCD memory fx pointers after EBS.")
```
os-d:
```suggestion
        log::info!("Disallowing alloc/free during ExitBootServices.")

```

crates/uefi_gcd/src/gcd.rs
```diff
@@ -213,7 +349,8 @@ impl GCD {
 
         self.add_memory_space(memory_type, base_address, len, capabilities)?;
 
-        self.allocate_memory_space(
+        (self.allocate_memory_space_fn)(
```
os-d: Can this just call `allocate_memory_space`? This is called once when that is the real function, so we can avoid the indirect call (unless that function isn't accessible at this layer, I didn't check)

dxe_core/src/misc_boot_services.rs
```diff
@@ -263,9 +266,11 @@ pub extern "efiapi" fn exit_boot_services(_handle: efi::Handle, map_key: usize)
     };
 
     // Terminate the memory map
+    // According to UEFI spec, in case of an incomplete or failed EBS call we must restore boot service functionality
```
os-d:
```suggestion
    // According to UEFI spec, in case of an incomplete or failed EBS call we must restore boot services memory allocation functionality

```

dxe_core/src/misc_boot_services.rs
```diff
@@ -263,9 +266,11 @@ pub extern "efiapi" fn exit_boot_services(_handle: efi::Handle, map_key: usize)
     };
 
     // Terminate the memory map
+    // According to UEFI spec, in case of an incomplete or failed EBS call we must restore boot service functionality
     let status = terminate_memory_map(map_key);
     if status.is_error() {
         EVENT_DB.signal_group(EBS_FAILED_GUID);
+        GCD.restore_free_allocate_fn();
```
os-d:
```suggestion
        GCD.restore_free_allocate_fn();
        EVENT_DB.signal_group(EBS_FAILED_GUID);
        
```
I would restore memory services first, if a listener for EBS_FAILED_GUID wants to alloc/free, that is now fine

crates/uefi_gcd/src/gcd.rs
```diff
@@ -1382,6 +1448,14 @@ impl SpinLockedGcd {
         }
     }
 
+    pub fn signal_ebs_start(&self) {
+        self.memory.lock().signal_ebs_start();
+    }
+
+    pub fn restore_free_allocate_fn(&self) {
```
os-d: nit: I would make this the same naming convention as the signal function: `signal_ebs_failed` or something similar. That is opaque to ExitBootServices and the GCD can do whatever it needs to do.

Makefile.toml
berlin-with0ut-return: yeah sorry, this shouldn't be in this PR at all.

crates/uefi_gcd/src/gcd.rs
berlin-with0ut-return: could you clarify what you mean by "the after_ebs ones call Null"?

os-d: Something like renaming them `allocate_memory_space_null`

crates/uefi_gcd/src/gcd.rs
berlin-with0ut-return: what is this comment referring to?

os-d: Oh, sorry, I mean that have a list going of things a platform needs to do to transition to the Rust core and one of the things is not allocate or free in [pre] exit boot services handlers.

crates/uefi_gcd/src/gcd.rs
makubacki, line 1: As part of this PR, can you add the "theory of operation" level overview to the documentation for this?

makubacki: Btw, it doesn't need to be very detailed, just behavior changes per their signal, expected platform visible results based on debug/release builds.

makubacki, line 1: I also think this should be marked as breaking. It doesn't really matter right now while we're internal, but this does introduce an assert in a code path without one. To my understanding this has also only been tested on QEMU and not physical platforms which have more drivers and vendor drivers of varying quality.

crates/uefi_gcd/src/gcd.rs
Javagedes: Yes I agree this would good section to add to the mdbook in addition to the memory_management section that was already added by @berlin-with0ut-return. I think a section regarding differences between our dxe_core and the edk2 dxe_core would make sense either at the start or the end of the `Integrating the Rust DXE Core` section of the mdbook. If you could add that section to this PR also, and just lightly explain the change (and probably point back to the memory management section you created) that would be awesome!

docs/src/dxe_core/operation.md
```diff
@@ -2,3 +2,5 @@
 
 The documentation in the following sections document the high-level architectural design, code
 flow, and theory of operation for the different critical components that make up the DXE Core.
+
+- [Memory Management](memory_management.md)
```
Javagedes: This does not need to be added here. The expectation is that they will just continue on to the rest of the book, which contains all of the high-level architectural design docs.

dxe_core/src/gcd/spin_locked_gcd.rs
```diff
@@ -1382,6 +1454,14 @@ impl SpinLockedGcd {
         }
     }
 
+    pub fn signal_ebs_start(&self) {
```
Javagedes: This is nit-picky, but I personally feel like we should not mention exit boot services in the GCD. rather the GCD just provides a way to swap the function pointers, and then the "exit_boot_services" fn just calls those more generic functions.

e.g. rename `signal_ebs_start` to something along the lines of `lock_memory_space` and `unlock_memory_space` and have the ebs fn call those functions.

@os-d is this acceptable? do you have any opinions on this that you would like it to stay as is?

os-d: I agree with that.

makubacki: I agree. This is part of the encapsulation discussed in the meeting and not leaking/coupling implementation details in the interface.

docs/src/dxe_core/memory_management.md
```diff
@@ -0,0 +1,10 @@
+# Memory Management
+
+## Exit Boot Services Handlers
+
+Platforms should ensure that when handling an `EXIT_BOOT_SERVICES` signal (and `PRE_EXIT_BOOT_SERVICES_SIGNAL`),
+they do not change the memory map. This means allocating and freeing are disallowed once `exit_boot_services` begins.
```
makubacki: Nit but more precise to the specification.

```suggestion
they do not change the memory map. This means allocating and freeing are disallowed once `EFI_BOOT_SERVICES.ExitBootServices()` (`exit_boot_services()`) is invoked.
```

PR #169, opened 2024-11-19
[REBASE&FF] Move multiple workspace crates into dxe_core

Javagedes:
## Description

Moves the following workspace crates into `dxe_core`

- uefi_allocator
- uefi_gcd
- uefi_event
- uefi_protocol_db
- uefi_tpl_mutex
- uefi_pecoff

The only major (but still minor) change with all of this is that the dependency on uefi_protocol_db had to be removed from uefi_depex. Previously uefi_depex required `UefiProtocolDb`. Now it just requires a slice of protocol guids to evaluate against.

Additionally, all doc examples had to be removed as doc examples can only be used on public facing structs and functions. By mobing them into dxe_core, that api layer is removed and docs will fail to compile due to trying to access private members.

Each commit is completely independent and can be compiled. Please review each commits comment for additional info.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

N/A

## Integration Instructions

N/A


Javagedes:
Looks like there are some more test shenanigans going on. I need to work on a few things outside this project, so I will get back to this. if anyone has time and can look into it, that would be great.

Cargo.toml
```diff
@@ -28,7 +28,6 @@ uefi_test = { path = "uefi_test", version = "2.0.2" }
 uefi_collections = { path = "crates/uefi_collections", version = "2.0.2", default-features = false }
 uefi_depex = { path = "crates/uefi_depex", version = "2.0.2" }
 uefi_device_path = { path = "crates/uefi_device_path", version = "2.0.2" }
-uefi_pecoff = { path = "crates/uefi_pecoff", version = "2.0.2" }
```
os-d, line 34: This one I could see leaving as independent crate. I haven't looked at it much, would it be valid in other environments, like StMM or non-UEFI environments? (I assume it is UEFI specific, but maybe not DXE specific?). But I don't have any objection to bringing it into the core if that is what folks want to do.

makubacki: I suggest we bring it in now for consistency in the repo and simplicity of maintenance and then break it out if/when the time comes.

Javagedes:
> LGTM. I think a follow on piece will be combining the GCD and allocator (and maybe memory_attributes_table) into a memory_manager module. That can be follow on work, doesn't need to come with this.

@os-d This is actually happening as a part of my work of removing statics for easier testing.

The design is not fully fleshed out as I'm having some issues with a design that appeases our needs and mockalls ability to create mocks, but generically there will be a memory_services module added that replaces gcd and allocator, and can be used by both boot services and runtime services.

dxe_core/src/gcd.rs
```diff
@@ -6,16 +6,21 @@
 //!
 //! SPDX-License-Identifier: BSD-2-Clause-Patent
 //!
+mod io_block;
+mod memory_block;
+mod spin_locked_gcd;
```
cfernald, line 11: Not a blocking comment, but do we have consensus on using mod.rs vs the root file and subfiles in the directory?

Javagedes: @cfernald you should never use `mod.rs` This is an old standard with legacy support. The correct layout is

```
|- my_module/
|      |- file1.rs
|      |- file2.rs
|- my_module.rs
```

If anywhere is using the following design, it needs to be updated to the above design.
```
|- my_module/
|      |- mod.rs
|      |- file1.rs
|      |- file2.rs
```

Rust's design choice to switch the "standard" to the first example, is to remove the unending amount of "mod.rs" files that exist in larger projects.

Cargo.toml
Javagedes: I have no preference and can drop the commit if needed. I think the final decision was that "goblin" is the layer that is re-usable across phases, and everything in `uefi_pecoff` was more-so dxe-specific. But I would leave it to comments from @joschock @makubacki

dxe_core/src/gcd.rs
cfernald: Good to know.. now don't mind these upcoming PRs. 😅

makubacki: I think we should follow the improvements made to simplify module paths away from mod.rs when possible - https://doc.rust-lang.org/nightly/edition-guide/rust-2018/path-changes.html.

Did you have more in mind?

makubacki: To clarify my comment, I responded while the other comments were posted.

PR #175, opened 2024-11-22
Use uefi-sdk common definitions

makubacki:
## Description

Replace common definitions and macros with those recently moved to uefi-sdk.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- Unit tests
- QEMU boot

## Integration Instructions

- N/A - Local integration of `uefi-sdk` changes in this repo.

### DIFF
--- a/dxe_core/src/image.rs
+++ b/dxe_core/src/image.rs
@@ -349,11 +349,12 @@ fn apply_image_memory_protections(pe_info: &UefiPeInfo, private_info: &PrivateIm
                 capabilities |= desc.capabilities;
             }
             Err(status) => {
-                debug_assert!(
-                    false,
+                log::error!(
                     "Failed to find GCD desc for image section {:#X} with Status {:#X?}",
-                    section_base_addr, status
+                    section_base_addr,
+                    status
                 );
+                debug_assert!(false);
                 continue;
             }
         }
@@ -369,11 +370,12 @@ fn apply_image_memory_protections(pe_info: &UefiPeInfo, private_info: &PrivateIm
             if let Ok(virtual_size) = align_up(section.virtual_size as u64, pe_info.section_alignment as u64) {
                 virtual_size
             } else {
-                debug_assert!(
-                    false,
+                log::error!(
                     "Failed to align up section size {:#X} with alignment {:#X}",
-                    section.virtual_size, pe_info.section_alignment
+                    section.virtual_size,
+                    pe_info.section_alignment
                 );
+                debug_assert!(false);
                 continue;
             };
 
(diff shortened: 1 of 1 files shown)


makubacki:
Note: I need to push a `uefi-sdk` ver update to `uefi-core` for CI to pass here.

https://github.com/pop-project/uefi-core/pull/28

dxe_core/src/image.rs
```diff
@@ -365,7 +365,18 @@ fn apply_image_memory_protections(pe_info: &UefiPeInfo, private_info: &PrivateIm
...
-        let aligned_virtual_size = align_up(section.virtual_size as u64, pe_info.section_alignment as u64);
+        let aligned_virtual_size =
+            if let Ok(virtual_size) = align_up(section.virtual_size as u64, pe_info.section_alignment as u64) {
+                virtual_size
+            } else {
+                log::error!(
+                    "Failed to align up section size {:#X} with alignment {:#X}",
+                    section.virtual_size,
+                    pe_info.section_alignment
+                );
```
os-d, line 377: Maybe throw a debug_assert in here? This could be fatal if this is a code section and we can't set RO+X on it. Although I see line 357 above has the same issue. I may want to change this so that if we can't apply memory protections we don't load the image, again, it will fail to execute since memory is XP by default.

That is a bit larger of a change, so we can file an issue for that and I can fix it, if you want

dxe_core/src/image.rs
```diff
@@ -411,7 +422,17 @@ fn remove_image_memory_protections(pe_info: &UefiPeInfo, private_info: &PrivateI
...
+                let aligned_virtual_size =
+                    if let Ok(virtual_size) = align_up(section.virtual_size as u64, pe_info.section_alignment as u64) {
+                        virtual_size
+                    } else {
+                        log::error!(
+                            "Failed to align up section size {:#X} with alignment {:#X}",
+                            section.virtual_size,
+                            pe_info.section_alignment
+                        );
+                        continue;
```
os-d, line 437: I would debug_assert! here. Not setting this back would lead to failures coalescing the GCD, which can cause issues for bootloaders, because we would still coalesce the EFI_MEMORY_MAP (all attrs get dropped, this is something to revisit, which I'll file an issue for). For now, we should catch this in development.

makubacki: Updated to debug_assert!().

dxe_core/src/image.rs
makubacki: I added debug_assert!() to both cases for now and filed this issue to track the longer term change - https://github.com/pop-project/uefi-dxe-core/issues/176.

makubacki:
> Note: I need to push a `uefi-sdk` ver update to `uefi-core` for CI to pass here.

Done

dxe_core/src/image.rs
```diff
@@ -349,10 +349,10 @@ fn apply_image_memory_protections(pe_info: &UefiPeInfo, private_info: &PrivateIm
                 capabilities |= desc.capabilities;
             }
             Err(status) => {
-                log::error!(
```
Javagedes: Don't we want to continue to log here for release mode builds?

makubacki: I think that makes sense. Updated to log the error and then debug_assert!().

PR #181, opened 2024-11-26
Add allocation stat tracking, Memory Type info, and fix a number of allocation subsystem bugs. [Rebase & FF]

joschock:
## Description

Adds a number of enhancements to the allocation subystem:
* Add support for tracking allocation stats (allocate/free call counts, number of currently consumed pages, reserved space)
* Add support the Memory Type Info configuration table. 
* Implement `Display` for GCD to dump GCD in a table format

Fixes a number of bugs in the allocation subsystem:
* Fix an issue where the incorrect non-owning allocator could free a page. This could cause a page in another allocators preferred range to be freed without ownership, resulting in pages claimed in other allocators "buckets"
* Fix an issue where inconsistent allocation strategies were used in different parts of the allocator, leading to inconsistent allocation patterns and messing up memory buckets.
* Allow directed address page allocation to be allocated in GCD types other than SystemMemory (e.g MMIO).

Refactored the allocation subsystem initialization:
* Added a "memory_type" field to the FixedSizeBlockAllocator so that it can know the memory type without lookup by handle.
* Refactored `page_change_callback` to supply a mutable reference to the allocator responsible for the page change; this allows access to allocator state in the callback.
* Marked a few useful but unused routines with "allow dead code"
* Moved allocation hob processing into allocator::init_memory_support() routine so that initial allocations can be properly accounted for in stat tracking. 
* Reworked allocation hob processing logic to better handle/log common scenarios where allocation hobs are not processed - such as duplicate ranges reported by both Memory Allocation Hob and corresponding Memory Allocation Module, or FV hob reported for a region not backed by MMIO.
* Adjusted Logging for memory buckets.

This is a fairly sizable PR, recommend review on a commit-by-commit basis. All individual commits build and pass unit tests.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Passes existing and new tests; manual evaluation of memory allocation processing and memory type info table on Q35 plus testing on real hardware prototypes. 

## Integration Instructions

N/A

### DIFF
--- a/dxe_core/src/allocator.rs
+++ b/dxe_core/src/allocator.rs
@@ -18,7 +18,7 @@ use core::{
 
 extern crate alloc;
 use alloc::{collections::BTreeMap, vec::Vec};
-use mu_rust_helpers::{function, guid::guid};
+use mu_rust_helpers::function;
 
 use crate::{
     gcd::{self, AllocateType as AllocationStrategy},
@@ -35,14 +35,13 @@ use mu_pi::{
 };
 use r_efi::{efi, system::TPL_HIGH_LEVEL};
 use uefi_allocator::UefiAllocator;
-use uuid::uuid;
 
 //FixedSizeBlockAllocator is passed as a reference to the callbacks on page allocations
 pub use fixed_size_block_allocator::FixedSizeBlockAllocator;
 
 use uefi_sdk::{
     base::{UEFI_PAGE_MASK, UEFI_PAGE_SIZE},
-    uefi_size_to_pages,
+    guid, uefi_size_to_pages,
 };
 
 // Allocation Strategy when not specified by caller.
@@ -784,7 +783,7 @@ pub fn install_memory_type_info_table(system_table: &mut EfiSystemTable) -> Resu
     let memory_table_mut = unsafe { (MEMORY_TYPE_INFO_TABLE.as_mut_ptr() as *mut c_void).as_mut().unwrap() };
 
     misc_boot_services::core_install_configuration_table(
-        guid!("4c19049f-4137-4dd3-9c10-8b97a83ffdfa"),
+        guid::MEMORY_TYPE_INFORMATION,
         Some(memory_table_mut),
         system_table,
     )
@@ -832,7 +831,7 @@ fn process_hob_allocations(hob_list: &HobList) {
                     uefi_size_to_pages!(desc.memory_length as usize),
                     &mut address as *mut efi::PhysicalAddress)
                     .inspect_err(|err|{
-                        if *err == efi::Status::NOT_FOUND && desc.name != guid!("00000000-0000-0000-0000-000000000000") {
+                        if *err == efi::Status::NOT_FOUND && desc.name != guid::ZERO {
                             //Guided Memory Allocation Hobs are typically MemoryAllocationModule or MemoryAllocationStack HOBs
                             //which have corresponding non-guided allocation HOBs associated with them; they are rejected as
                             //duplicates if we attempt to log them. Only log trace messages for these.


dxe_core/src/allocator/fixed_size_block_allocator.rs
```diff
@@ -503,20 +508,19 @@ impl FixedSizeBlockAllocator {
         // any general allocations are serviced; that way all "owned" memory is in prime position before any "unowned"
         // memory.
         //
-        let preferred_block = self.allocate_pages(AllocationStrategy::TopDown(None), pages)?;
+        let preferred_block = self.allocate_pages(AllocationStrategy::BottomUp(None), pages)?;
```
os-d: Why does this and the others become BottomUp?

dxe_core/src/allocator/fixed_size_block_allocator.rs
```diff
@@ -503,20 +508,19 @@ impl FixedSizeBlockAllocator {
...
-        )?;
-
         // this will fail if called more than once, but check at start of function should guarantee that doesn't happen.
         self.preferred_range =
             Some(preferred_block_address..preferred_block_address + (pages * ALIGNMENT) as efi::PhysicalAddress);
 
+        // update reserved stat here, since allocate_pages was not yet aware of preferred range to properly track.
+        self.stats.reserved_used += pages * ALIGNMENT;
+        unsafe {
+            self.free_pages(preferred_block_address as usize, pages).unwrap();
```
os-d, line 553: In this case, `allocate_pages` would increase `claimed_pages` and not `reserved_used`, but when we call `free_pages`, it will decrement `claimed_pages` and not `reserved_used`, right?

dxe_core/src/allocator.rs
```diff
@@ -711,6 +712,26 @@ pub(crate) fn ensure_capacity(memory_type: efi::MemoryType, size: usize, align:
     Ok(())
 }
 
+static mut MEMORY_TYPE_INFO_TABLE: [EFiMemoryTypeInformation; 17] = [
```
os-d, line 728: Should this be initially seeded from the HOB, if available? https://github.com/microsoft/mu_basecore/blob/206b001a564fdda9f7f1d055f67211ae1575e28d/MdeModulePkg/Core/Dxe/Gcd/Gcd.c#L2336-L2344

This would capture allocations from pre-DXE, right? Or is that already being handled through the page_change_callback getting invoked when allocation HOBs are added?

dxe_core/src/allocator/fixed_size_block_allocator.rs
joschock: Because the main expand was switched to BottomUp(). So we were reserving pages using TopDown(None) for the buckets, but then allocating pages from the Gcd using BottomUp(None) - since the strategies were different the pages didn't actually come from the reserved bucket; making the buckets essentially useless.

I might even go further and declare something like 
`const UNIVERSAL_ALLOCATION_STRATEGY: AllocationStrategy = AllocationStrategy::BottomUp(None);` 
and use that everywhere just to make sure that it's consistent across the implementaition.

dxe_core/src/allocator/fixed_size_block_allocator.rs
joschock: actually, I expect the opposite. Because `preferred_range` is now set, `free_pages` should observe that the page is in the reserved range and decrement `reserved_used` instead of `claimed_pages`. So at the end of the reserved call, I expect that 
`reserved_size` and `claimed_pages` match the bucket size and `reserved_used` should be zero. The `test_allocation_stats` test I added should enforce this.

dxe_core/src/allocator.rs
joschock: Hmm.  Good point. I think the actual right thing to do is send the Allocation HOBs through the allocators as part of allocator init, rather allocating straight from the GCD. I'll update and test.

dxe_core/src/allocator/fixed_size_block_allocator.rs
os-d: Sorry, I typed that the wrong way, I did mean that `reserved_used` would be decremented and `claimed_pages` would not be. So wouldn't we underflow `reserved_used` and I guess it is okay that `claimed_pages` is not decremented, because those pages are "claimed".

From this:

```rust
self.stats.reserved_used -= pages * ALIGNMENT;
```

dxe_core/src/allocator/fixed_size_block_allocator.rs
joschock: Updated in latest push.

dxe_core/src/allocator.rs
joschock: Initial implementation and unit tests in latest push.

dxe_core/src/allocator.rs
```diff
@@ -777,6 +785,120 @@ pub fn install_memory_type_info_table(system_table: &mut EfiSystemTable) -> Resu
...
+                // both associated with it an pointing to the same memory. Check to see if the region has already
+                // been reserved in the GCD. If so, skip attempting to allocate it again.
+                if let Ok(existing_desc) = GCD.get_memory_descriptor_for_address(desc.memory_base_address) {
+                    if existing_desc.base_address == desc.memory_base_address
+                        && existing_desc.length == desc.memory_length
+                        && existing_desc.memory_type == dxe_services::GcdMemoryType::SystemMemory
+                        && existing_desc.image_handle != INVALID_HANDLE
+                    {
+                        log::info!(
+                            "Skipping duplicate Memory Allocation/Module HOB at {:#x?} of length {:#x?}. Region already present in GCD.",
```
os-d: Maybe add a "this is expected" statement here?

dxe_core/src/allocator.rs
```diff
@@ -777,6 +785,120 @@ pub fn install_memory_type_info_table(system_table: &mut EfiSystemTable) -> Resu
...
+                            "Skipping FV HOB at {:#x?} of length {:#x?}. Containing region is not MMIO.",
+                            base_address,
+                            length,
+                        );
+                        continue;
+                    }
+                }
+
+                //The 4K granularity rule does not apply to FV hobs, so allocate_pages cannot be used.
+                //This means they must be direct-allocated in the GCD, and no stats will be tracked for them.
```
os-d, line 905: Is it worthwhile to have a `report_memory_stats` for this case? I don't have a preference, presumably this would be the only consumer of that function.

dxe_core/src/allocator.rs
```diff
@@ -1034,6 +1040,52 @@ mod tests {
...
+
+            let allocators = ALLOCATORS.lock();
+
+            //Verify that the memory allocation hobs resulted in claimed pages in the allocator.
+            for memory_type in [
+                efi::RESERVED_MEMORY_TYPE,
+                efi::LOADER_CODE,
+                efi::LOADER_DATA,
+                efi::BOOT_SERVICES_CODE,
+                //efi::BOOT_SERVICES_DATA,
```
os-d: I assume this is commented out because in setting up memory services, many more `BOOT_SERVICES_DATA` allocations are made?

dxe_core/src/allocator.rs
```diff
@@ -698,6 +726,182 @@ pub(crate) fn ensure_capacity(memory_type: efi::MemoryType, size: usize, align:
...
+            Hob::MemoryAllocation(hob::MemoryAllocation { header: _, alloc_descriptor: desc })
+            | Hob::MemoryAllocationModule(hob::MemoryAllocationModule {
+                header: _,
+                alloc_descriptor: desc,
+                module_name: _,
+                entry_point: _,
+            }) => {
+                log::trace!("[{}] Processing Memory Allocation HOB:\n{:#x?}\n\n", function!(), hob);
+                //Use allocate_pages here to record these allocations and keep the allocator stats up to date.
+                //Note: PI spec 1.8 III-5.4.1.1 stipulates that memory allocations must page-granularity,
```
os-d:
```suggestion
                //Note: PI spec 1.8 III-5.4.1.1 stipulates that memory allocations must have page-granularity,
```

dxe_core/src/allocator/fixed_size_block_allocator.rs
os-d: Nvm, I see where you adjust for that with reserved_used += pages * ALIGNMENT right before free_pages.

dxe_core/src/allocator.rs
joschock: I refactored this a bit, downgraded the message to trace, and added a block comment for the curious.

dxe_core/src/allocator.rs
joschock: Yeah - all the examples I've observed so far (sample size of two: q35 and an aarch64 hardware prototype) don't report the backing MMIO descriptor anyway, so the attempt allocate in the GCD fails, so this may not matter most of the time. The main reason this HOB is reported is for later core/dispatcher processing - if the FV has been decompressed into RAM via PEI, for example, then there will be a separate memory allocation HOB that covers that region which will protected it in the GCD. The only reason we added processing of the FV hob here was for alignment with the EDK2 core. 

In addition, the main reason to track stats is for helping to tune the buckets for RAM; and the buckets are page-level granularity. So small non-page granularity MMIO allocations like this don't really fit into that paradigm. Do you think we should try and track stats for the MMIO somehow? Does MMIO allocation instability cause issues with OS resume often?

dxe_core/src/allocator.rs
joschock: oversight on my part; had commented it out pending a bug fix, and forgot to put it back in. fixed in latest.

dxe_core/src/allocator.rs
joschock: Thanks, fixed in latest.

joschock:
CI is failing due to x86_64 0.14.13 not working with stable toolchain. https://github.com/pop-project/uefi-sdk/pull/6 with the fix.

dxe_core/src/lib.rs
```diff
@@ -231,17 +231,18 @@ where
...
-        // After this point Rust Heap usage is permitted (since GCD is initialized).
+        // After this point Rust Heap usage is permitted (since GCD is initialized with a single known-free region).
         // Relocate the hobs from the input list pointer into a Vec.
         let mut hob_list = HobList::default();
         hob_list.discover_hobs(physical_hob_list);
 
         log::trace!("HOB list discovered is:");
         log::trace!("{:#x?}", hob_list);
 
-        gcd::add_hob_resource_descriptors_to_gcd(&hob_list, free_memory_start, free_memory_size);
```
os-d, line 257: Looks like this function doesn't need free_memory_start or free_memory_size at all? The trace logs could just move to the gcd function.

dxe_core/src/gcd/spin_locked_gcd.rs
```diff
@@ -879,6 +879,51 @@ impl GCD {
...
+                        let mem_type_str_idx =
+                            usize::min(descriptor.memory_type as usize, Self::GCD_MEMORY_TYPE_NAMES.len() - 1);
+                        writeln!(
+                            f,
+                            "{}  {:016x?}-{:016x?} {:016x?} {:016x?} {:016x?} {:016x?}",
+                            GCD::GCD_MEMORY_TYPE_NAMES[mem_type_str_idx],
+                            descriptor.base_address,
+                            descriptor.base_address + descriptor.length - 1,
+                            descriptor.capabilities,
+                            descriptor.attributes,
```
os-d, line 916: It would be nice for capabilities and attributes if we could have a string of them instead of hex, just makes it easier. Copilot is good at writing this :), e.g.: https://github.com/pop-project/uefi-dxe-core/blob/5388fa8994f5d0be3827f3c1dd1c33f81852e3a5/dxe_core/src/allocator.rs#L84-L201, which probably could be moved to uefi-sdk and consumed here, too, since the attrs/capabilities are the same. But, also don't need to hold up this PR for it, just makes it easier to quickly consume in logs.

joschock: I'm not opposed, but this PR is already gigantic; so I will defer.

dxe_core/src/allocator.rs
os-d: I chatted with the team: non-runtime MMIO should be fine to not be in the bucket because it doesn't get reported to the OS in the EFI_MEMORY_MAP: https://github.com/microsoft/mu_basecore/blob/f2286518d7f41df21c94e2d8ec10ebeaa429a675/MdeModulePkg/Core/Dxe/Mem/Page.c#L2094-L2095.

_However_, the Rust core seems to be reporting even non-runtime MMIO in the EFI_MEMORY_MAP: https://github.com/pop-project/uefi-dxe-core/blob/5388fa8994f5d0be3827f3c1dd1c33f81852e3a5/dxe_core/src/allocator.rs#L519-L531.

So I think we need to fix that, unless I'm missing where we are filtering for runtime.

joschock: I'll address filtering MMIO for runtime in a separate PR.

dxe_core/src/lib.rs
joschock: will fix this; waiting on integration of uefi-core/uefi-sdk with x86_64 dependency fix so I can compile/test

joschock: resolved in latest push.

joschock:
> CI is failing due to x86_64 0.14.13 not working with stable toolchain. [pop-project/uefi-sdk#6](https://github.com/pop-project/uefi-sdk/pull/6) with the fix.

fixed.

dxe_core/src/allocator.rs
```diff
@@ -698,6 +726,205 @@ pub(crate) fn ensure_capacity(memory_type: efi::MemoryType, size: usize, align:
...
+        }
+    }
+}
+
+pub fn install_memory_type_info_table(system_table: &mut EfiSystemTable) -> Result<(), efi::Status> {
+    //MEMORY_TYPE_INFO_TABLE is static mut, so we know the pointer is good.
+    let memory_table_mut = unsafe { (MEMORY_TYPE_INFO_TABLE.as_mut_ptr() as *mut c_void).as_mut().unwrap() };
+
+    misc_boot_services::core_install_configuration_table(
+        guid!("4c19049f-4137-4dd3-9c10-8b97a83ffdfa"),
```
makubacki: Should this GUID be defined in uefi-sdk? That would give platform drivers an opportunity to use it as well such as when building the HOB.

dxe_core/src/allocator.rs
```diff
@@ -698,6 +726,205 @@ pub(crate) fn ensure_capacity(memory_type: efi::MemoryType, size: usize, align:
...
+                }
+
+                let mut address = desc.memory_base_address;
+                let _ = core_allocate_pages(
+                    efi::ALLOCATE_ADDRESS,
+                    desc.memory_type,
+                    uefi_size_to_pages!(desc.memory_length as usize),
+                    &mut address as *mut efi::PhysicalAddress)
+                    .inspect_err(|err|{
+                        if *err == efi::Status::NOT_FOUND && desc.name != guid!("00000000-0000-0000-0000-000000000000") {
```
makubacki: Might make sense to add the zero GUID to uefi-sdk as well.

dxe_core/src/test_support.rs
```diff
@@ -62,10 +80,239 @@ pub(crate) unsafe fn init_test_protocol_db() {
...
+    //
+    // Base:         offset 0                   ************
+    // HobList:      offset base+0              HOBS
+    // Empty:        offset base+HobListSize    N/A
+    // SystemMemory  offset base+0xE0000        SystemMemory (resource_descriptor1)
+    // Reserved      offset base+0xF0000        Untested SystemMemory (resource_descriptor2)
+    // FreeMemory    offset base+0x100000       FreeMemory (phit)
+    // End           offset base+0x200000       ************
+    //
+    // THe test HOB list will also include resource descriptor hobs that describe MMIO/IO as follows:
```
makubacki:
```suggestion
    // The test HOB list will also include resource descriptor hobs that describe MMIO/IO as follows:
```

dxe_core/src/allocator.rs
joschock: fixed

dxe_core/src/allocator.rs
joschock: fixed

dxe_core/src/test_support.rs
joschock: fixed

PR #200, opened 2024-12-18
[REBASE & FF] Use the paging Crate

os-d:
## Description

This PR includes the changes required to move DXE Core to the paging crate. Note, it depends on PRs into most of the other Rust repos we have, so Cargo.toml is not updated with new versions yet, that will come as the other PRs are merged. Tests and documentation will be updated in this PR before it is ready to check in. It also needs to be tested on real hardware.

Rust-y reviews are welcome (why didn't you use this simple function or rewrite it this way?)

- [x] Impacts functionality?
- [x] Impacts security?
- [x] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested by booting Q35 to Windows with the new MpDxe and no CpuDxe.

## Integration Instructions

x86 platforms should drop CpuDxe and only include MpDxe (assuming they want MP services).

### DIFF
--- a/dxe_core/src/gcd/spin_locked_gcd.rs
+++ b/dxe_core/src/gcd/spin_locked_gcd.rs
@@ -26,7 +26,67 @@ use crate::{
     tpl_lock,
 };
 use paging::{page_allocator::PageAllocator, MemoryAttributes, PageTable, PtError, PtResult};
-use uefi_cpu::paging::create_cpu_paging;
+
+// temporary until aarch64 paging comes in. this cfg_if block will be removed
+// and just the else use statement will be used. cfg_if can be removed from
+// Cargo.toml then
+cfg_if::cfg_if! {
+    if #[cfg(all(target_os = "uefi", target_arch = "x86_64"))] {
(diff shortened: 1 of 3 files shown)


dxe_core/src/gcd/spin_locked_gcd.rs
```diff
@@ -208,6 +322,129 @@ impl GCD {
...
+                }
+                Ok(_) => return Ok(()),
+                // any other error is unexpected
+                Err(_) => Err(efi::Status::INVALID_PARAMETER)?,
+            }
+        }
+
+        if needed_pages > 0 {
+            match self.allocate_memory_space(
+                AllocateType::TopDown(None),
```
joschock: Consider using a central `CONST` definition set to `AllocateType::TopDown(None)` so that if/when it needs to be consistent in multiple places, that's self-documented by use of the const.

os-d: Thanks, I had written this before your default allocation strategy, I’ll switch to that (except the one case where we want BottomUp)

PR #203, opened 2024-12-23
Adding CPU arch support for AArch64

kuqin12:
## Description

This change adds the CPU Arch protocol and hardware interrupt protocol for AArch64.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This change was tested on QEMU SBSA platform. With a CpuDxe driver modified to only support memory attribute setting, the system can boot to UEFI shell.

## Integration Instructions

`InterruptBases` needs to be initiated before handing off the core init.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,9 +17,9 @@ rust-version = "1.80.0"
 
 [workspace.dependencies]
 uefi_sdk = { git = "https://github.com/pop-project/uefi-sdk", rev = "v0.1.5" }
-uefi_cpu = { git = "https://github.com/pop-project/uefi-core.git", rev = "v8.0.1" }
-uefi_debugger = { git = "https://github.com/pop-project/uefi-core.git", rev = "v8.0.1" }
-section_extractor = { git = "https://github.com/pop-project/uefi-core.git", rev = "v8.0.1" }
+uefi_cpu = { git = "https://github.com/pop-project/uefi-core.git", rev = "v9.0.0" }
+uefi_debugger = { git = "https://github.com/pop-project/uefi-core.git", rev = "v9.0.0" }
+section_extractor = { git = "https://github.com/pop-project/uefi-core.git", rev = "v9.0.0" }
 corosensei = { git = "https://github.com/pop-project/uefi-corosensei.git", rev = "d27241e17a5f3c0703fe9871e6f68cd838c05c6d", default-features = false }
 
 uefi_test_macro = { path = "crates/uefi_test_macro", version = "6.1.0" }


dxe_core/src/hw_interrupt_protocol.rs
```diff
@@ -0,0 +1,412 @@
...
+pub enum HardwareInterrupt2TriggerType {
+    // HardwareInterrupt2TriggerTypeLevelLow = 0, // Not used
+    HardwareInterrupt2TriggerTypeLevelHigh = 1,
+    // HardwareInterrupt2TriggerTypeEdgeFalling = 2, // Not used
+    HardwareInterrupt2TriggerTypeEdgeRising = 3,
+}
+
+// { 0x2890B3EA, 0x053D, 0x1643, { 0xAD, 0x0C, 0xD6, 0x48, 0x08, 0xDA, 0x3F, 0xF1 } }
+pub const EFI_HARDWARE_INTERRUPT_PROTOCOL_GUID: efi::Guid =
+    efi::Guid::from_fields(0x2890B3EA, 0x053D, 0x1643, 0xAD, 0x0C, &[0xD6, 0x48, 0x08, 0xDA, 0x3F, 0xF1]);
```
os-d: We should define this in uefi-sdk so that other Rust components can use it

dxe_core/src/hw_interrupt_protocol.rs
```diff
@@ -0,0 +1,412 @@
...
+
+type HardwareInterruptRegister =
+    extern "efiapi" fn(*mut EfiHardwareInterruptProtocol, u64, HwInterruptHandler) -> efi::Status;
+type HardwareInterruptEnable = extern "efiapi" fn(*mut EfiHardwareInterruptProtocol, u64) -> efi::Status;
+type HardwareInterruptDisable = extern "efiapi" fn(*mut EfiHardwareInterruptProtocol, u64) -> efi::Status;
+type HardwareInterruptGetState = extern "efiapi" fn(*mut EfiHardwareInterruptProtocol, u64, *mut bool) -> efi::Status;
+type HardwareInterruptEnd = extern "efiapi" fn(*mut EfiHardwareInterruptProtocol, u64) -> efi::Status;
+
+/// C struct for the Advanced Logger protocol.
```
os-d: no it isn't :)

dxe_core/src/hw_interrupt_protocol.rs
```diff
@@ -0,0 +1,412 @@
...
+    if protocol.is_null() {
+        return efi::Status::INVALID_PARAMETER;
+    }
+
+    unsafe { &mut *this }.hw_interrupt_handler.aarch64_int.lock().end_of_interrupt(interrupt_source)
+}
+
+//   { 0x32898322, 0x2d1a, 0x474a, { 0xba, 0xaa, 0xf3, 0xf7, 0xcf, 0x56, 0x94, 0x70 } }
+pub const EFI_HARDWARE_INTERRUPT2_PROTOCOL_GUID: efi::Guid =
+    efi::Guid::from_fields(0x32898322, 0x2d1a, 0x474a, 0xba, 0xaa, &[0xf3, 0xf7, 0xcf, 0x56, 0x94, 0x70]);
```
os-d: same

dxe_core/src/hw_interrupt_protocol.rs
```diff
@@ -0,0 +1,412 @@
...
+type HardwareInterruptGetStateV2 =
+    extern "efiapi" fn(*mut EfiHardwareInterruptV2Protocol, u64, *mut bool) -> efi::Status;
+type HardwareInterruptEndV2 = extern "efiapi" fn(*mut EfiHardwareInterruptV2Protocol, u64) -> efi::Status;
+
+type HardwareInterruptGetTriggerTypeV2 =
+    extern "efiapi" fn(*mut EfiHardwareInterruptV2Protocol, u64, *mut HardwareInterrupt2TriggerType) -> efi::Status;
+type HardwareInterruptSetTriggerTypeV2 =
+    extern "efiapi" fn(*mut EfiHardwareInterruptV2Protocol, u64, HardwareInterrupt2TriggerType) -> efi::Status;
+
+/// C struct for the Advanced Logger protocol.
```
os-d: same :)

dxe_core/src/hw_interrupt_protocol.rs
```diff
@@ -0,0 +1,412 @@
...
+        return efi::Status::INVALID_PARAMETER;
+    }
+
+    let level = unsafe { &mut *this }.hw_interrupt_handler.aarch64_int.lock().get_trigger_type(interrupt_source);
+
+    // I know this looks odd, but this is how ArmGicV3 in EDK2 does it...
+    let t_type = match level {
+        Trigger::Level => HardwareInterrupt2TriggerType::HardwareInterrupt2TriggerTypeLevelHigh,
+        Trigger::Edge => HardwareInterrupt2TriggerType::HardwareInterrupt2TriggerTypeEdgeRising,
+    };
```
os-d: maybe for this and set, doing impl From back and forth?

dxe_core/src/hw_interrupt_protocol.rs
```diff
@@ -0,0 +1,412 @@
...
+    }
+}
+
+struct HwInterruptProtocolHandler {
+    handlers: TplMutex<Vec<Option<HwInterruptHandler>>>,
+    aarch64_int: TplMutex<AArch64InterruptInitializer>,
+}
+
+impl InterruptHandler for HwInterruptProtocolHandler {
+    fn handle_interrupt(&'static self, exception_type: usize, _context: &mut ExceptionContext) {
```
os-d:
```suggestion
    fn handle_interrupt(&'static self, exception_type: usize, context: &mut ExceptionContext) {

```

Looks like context is actually used below

kuqin12: Updated/

dxe_core/src/hw_interrupt_protocol.rs
kuqin12: Thanks, good suggestion. Updated.

dxe_core/src/hw_interrupt_protocol.rs
```diff
@@ -0,0 +1,426 @@
...
+    enable_interrupt_source: HardwareInterruptEnable,
+    disable_interrupt_source: HardwareInterruptDisable,
+    get_interrupt_source_state: HardwareInterruptGetState,
+    end_of_interrupt: HardwareInterruptEnd,
+
+    // Internal rust access only! Does not exist in C definition.
+    hw_interrupt_handler: &'a mut HwInterruptProtocolHandler,
+}
+
+impl<'a> EfiHardwareInterruptProtocol<'a> {
```
Javagedes, line 48: nitpick suggestion:

you can move all of the extern "efiapi" functions inside the `impl<'a> EfiHardwareInterruptProtocol<'a>`, which allows you to get rid of the `_v1` and `_v2` of the function names

And then the `new` looks like this:
``` rust
Self {
  register_interrupt_source: Self::register_interrupt_source,
  enable_interrupt_source: Self::enable_interrupt_source,
  disable_interrupt_source: Self::disable_interrupt_source,
  get_interrupt_source_state: Self::get_interrupt_source_state,
  end_of_interrupt: Self::end_of_interrupt,
  hw_interrupt_handler,
}
```

dxe_core/src/hw_interrupt_protocol.rs
```diff
@@ -0,0 +1,426 @@
...
+use r_efi::efi;
+use uefi_cpu::interrupts::aarch64::gic_manager::{gic_initialize, get_max_interrupt_number, AArch64InterruptInitializer};
+use uefi_cpu::interrupts::{ExceptionContext, InterruptBases, InterruptHandler, InterruptManager};
+
+use arm_gic::gicv3::{GicV3, Trigger};
+
+pub type HwInterruptHandler = extern "efiapi" fn(u64, &mut ExceptionContext);
+
+#[repr(C)]
+pub enum HardwareInterrupt2TriggerType {
```
Javagedes, line 20: you may want `#[non_exhaustive]` here. This makes adding additional trigger types non-(compilation) breaking because it forces any match statements against this enum to include a catch-all ` _ => ...`

dxe_core/src/hw_interrupt_protocol.rs
```diff
@@ -0,0 +1,426 @@
...
+pub extern "efiapi" fn enable_interrupt_source_v1(
+    this: *mut EfiHardwareInterruptProtocol,
+    interrupt_source: u64,
+) -> efi::Status {
+    let protocol = this as *const c_void;
+    if protocol.is_null() {
+        return efi::Status::INVALID_PARAMETER;
+    }
+
+    unsafe { &mut *this }.hw_interrupt_handler.aarch64_int.lock().enable_interrupt_source(interrupt_source)
```
Javagedes: Any `unsafe { ... }` is suppose to have a `// SAFETY: ...` comment about _why_ this is now safe instead of being bubbled upwards as an `unsafe fn`. e.g. technically when you do an `unsafe { ... }` instead of marking the higher-level function as `unsafe` you are typically guaranteeing that something in the code that you control is ensuring this code always safe.

Realistically, you may even want to mark this entire function as unsafe. That is because `this` is a pointer that is passed to us by C code, so we do not have compile time guarantees that `this` is truly a pointer to a `EfiHardwareInterruptProtocol` object. If this was not an extern call, we technically have a compile time guarantee that the pointer is what we say it is, and the only safety check we would have to do is that the pointer is not null.

`unsafe extern "efiapi" fn ...`

dxe_core/src/hw_interrupt_protocol.rs
```diff
@@ -0,0 +1,426 @@
...
+    }
+
+    unsafe { &mut *this }.hw_interrupt_handler.register_interrupt_source(interrupt_source as usize, handler)
+}
+
+pub extern "efiapi" fn enable_interrupt_source_v1(
+    this: *mut EfiHardwareInterruptProtocol,
+    interrupt_source: u64,
+) -> efi::Status {
+    let protocol = this as *const c_void;
```
Javagedes: This does not seem necessary. I think you can just do `if this.is_null() { ... }`

dxe_core/src/hw_interrupt_protocol.rs
Javagedes, line 1: Does this protocol code belong in the dxe_core at all? It seems like something that should be outside of the core.

dxe_core/src/hw_interrupt_protocol.rs
kuqin12: Fixed

dxe_core/src/hw_interrupt_protocol.rs
kuqin12: Fixed

dxe_core/src/hw_interrupt_protocol.rs
kuqin12: Updated

dxe_core/src/hw_interrupt_protocol.rs
kuqin12: Thanks, updated! This looks much better.

dxe_core/src/hw_interrupt_protocol.rs
kuqin12: Update for uefi-sdk is here: https://github.com/pop-project/uefi-sdk/pull/15

dxe_core/src/hw_interrupt_protocol.rs
kuqin12: This is part of the existing CPU DXE logic for AArch64. We would need the GIC to be ready before handing the system off to other components.

I think we can make it into a rust component like adv logger at some point later. But that will need at least the interrupt manager and TplMutex exposed to this component, which is another good topic just by itself.

dxe_core/src/hw_interrupt_protocol.rs
kuqin12: Thanks, updated

dxe_core/src/hw_interrupt_protocol.rs
kuqin12: Agreed, updated to unsafe function. Rust code does not call this function directly anyway.

dxe_core/src/hw_interrupt_protocol.rs
kuqin12: Good call. Updated.

dxe_core/src/hw_interrupt_protocol.rs
Javagedes: on the TPL Mutex side of things, there is an open source version of it that you can use:

https://github.com/microsoft/mu_rust_helpers/tree/main/tpl_mutex

There is a bit of a difference between the two - mainly that the core-specific version works as a regular Mutex until boot services is initialized, where as boot services must be initialized to be able to use the open source one (which should be no problem since protocols are typnot registered until after boot services table has been completely set up)

dxe_core/src/gcd/spin_locked_gcd.rs
```diff
@@ -552,7 +492,7 @@ impl GCD {
         // We need to explicitly allocate the root page below 4GB for x86 MP Services.
         // See comment in PagingAllocator.allocate_pages
         match self.allocate_memory_space(
-            AllocateType::BottomUp(Some(SIZE_4GB - 1)),
```
kuqin12: Is this the right way to proceed?

dxe_core/src/hw_interrupt_protocol.rs
kuqin12: Updated.

dxe_core/src/gcd/spin_locked_gcd.rs
```diff
@@ -26,67 +26,7 @@ use crate::{
...
-        fn create_cpu_paging<A: PageAllocator + 'static>(
-            _page_allocator: A,
-        ) -> Result<Box<dyn PageTable<ALLOCATOR = A>>, efi::Status> {
-            Ok(Box::new(EfiCpuPagingNull { _allocator: core::marker::PhantomData }))
-        }
-    } else {
-        use uefi_cpu::paging::create_cpu_paging;
-    }
-}
+use uefi_cpu::paging::create_cpu_paging;
```
kuqin12, line 29: Needs https://github.com/pop-project/uefi-core/pull/51 to complete first.

kuqin12: Done.

dxe_core/src/gcd/spin_locked_gcd.rs
kuqin12: Has below 4GB, if so, allocate it.

Or try 4GB - 1 and then DEFAULT_ALLOCATION_STRATEGY as fallback.

kuqin12: Updated.

PR #205, opened 2025-01-02
[REBASE & FF] Update Memory Attribute Protocol to Split Attributes Across Descriptors

os-d:
## Description

This consists of the following commits:

Add Prints for Attribute Setting Failures
--
This commit adds prints and debug_asserts for failing to set attributes in the GCD. Prints and debug_asserts exist at other levels (such as in the page table manipulation code) but not at this level. This improves debugging page faults.

Return Correct Result From GCD Functions
--
There was no semicolon after the map change callback functions, meaning that we would return the wrong result if the callback succeeded. In practice, this would rarely matter, but if the code is shifted so that other code should execute after the callback, it will not do so (as was the case in AllocateMemorySpace, fixed previously).

Use Memory Attribute Protocol From r-efi
--
The addition of the Memory Attribute Protocol to r-efi has been completed, so the temporary hosting of the definitions in uefi-dxe-core can be dropped. There is no functional change here.

Update set_memory_space_attributes to Split Attributes Across Descriptors
--
Currently, set_memory_space_attributes does not allow for attributes to be set across multiple descriptors. However, the PI spec allows for this through this API and the CPU arch protocol and memory attribute protocol allow for this and are thin wrappers over the set_memory_space_attributes function.

This fixes an issue with Windows booting on a failure path: when Windows sees that previous boots have shutdown unexpectedly, it will boot to WinRE. When doing so, it unloads winload and sets the entire image space to clear attributes. This fails in the Rust core, because the winload image sections have different attributes set, meaning that they have different GCD descriptors and so set_memory_space_attributes disallows setting the attributes.

This commit simply checks at the top level set_memory_space_attributes code to see if the range splits multiple descriptors and calls the inner function once for each descriptor (or part of descriptor) in the range.

- [x] Impacts functionality?
- [x] Impacts security?
- [x] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on Q35 by seeing the Windows failure path hit a page fault and then succeeding with this change.

## Integration Instructions

This updates the required version of r-efi to ^5.2.0 to pull in the memory attribute protocol definition. Consumers will need to ensure they build with at least that version. That is the only reason this is marked as a breaking change.

### DIFF
--- a/dxe_core/src/gcd/spin_locked_gcd.rs
+++ b/dxe_core/src/gcd/spin_locked_gcd.rs
@@ -2487,13 +2487,30 @@ impl SpinLockedGcd {
     /// # Documentation
     /// UEFI Platform Initialization Specification, Release 1.8, Section II-7.2.4.6
     pub fn set_memory_space_attributes(&self, base_address: usize, len: usize, attributes: u64) -> Result<(), Error> {
-        let result = self.memory.lock().set_memory_space_attributes(base_address, len, attributes);
-        if result.is_ok() {
-            if let Some(callback) = self.memory_change_callback {
-                callback(MapChangeType::SetMemoryAttributes)
-            }
+        // this API allows for setting attributes across multiple descriptors in the GCD (assuming the capabilities
+        // allow it). The lower level set_memory_space_attributes will only operate on a single entry in the GCD/page
+        // table, so at this level we need to check to see if the range spans multiple entries and if so, we need to
+        // split the range and call set_memory_space_attributes for each entry.
+
+        let mut current_base = base_address as u64;
+        let range_end = (base_address + len) as u64;
+        while current_base < range_end {
+            let descriptor = self.get_memory_descriptor_for_address(current_base as efi::PhysicalAddress)?;
+            let descriptor_end = descriptor.base_address + descriptor.length;
+
+            // it is still legal to split a descriptor and only set the attributes on part of it
+            let next_base = u64::min(descriptor_end, range_end);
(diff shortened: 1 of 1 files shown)


dxe_core/src/gcd/spin_locked_gcd.rs
```diff
@@ -2468,13 +2487,30 @@ impl SpinLockedGcd {
...
+
+        let mut current_base = base_address as u64;
+        let range_end = (base_address + len) as u64;
+        while current_base < range_end {
+            let descriptor = self.get_memory_descriptor_for_address(current_base as efi::PhysicalAddress)?;
+            let descriptor_end = descriptor.base_address + descriptor.length;
+
+            // it is still legal to split a descriptor and only set the attributes on part of it
+            let next_base = u64::min(descriptor_end, range_end);
+            let next_len = next_base - current_base;
```
cfernald: Shouldn't this be current_len?

os-d: You mean as a variable name? Yeah, it can be named that. Copilot wrote this part without my asking it to :)

cfernald: yeah the variable name, since it is the length of what is being called `current_base` and not `next_base`

Javagedes: "Copilot wrote this" is going to be my only excuse from now on. 😄

PR #206, opened 2025-01-03
Move debugger initialization to after allocator

cfernald:
## Description

Moves the debugger to later after allocator initialization in preparation for coming debugger crate changes. 

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on Q35 with local debugger changes

## Integration Instructions

N/A

### DIFF
--- a/dxe_core/src/lib.rs
+++ b/dxe_core/src/lib.rs
@@ -271,7 +271,6 @@ where
         PROTOCOL_DB.init_protocol_db();
         // Initialize full allocation support.
         allocator::init_memory_support(&hob_list);
-
         // we have to relocate HOBs after memory services are initialized as we are going to allocate memory and
         // the initial free memory may not be enough to contain the HOB list. We need to relocate the HOBs because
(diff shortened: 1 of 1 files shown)


dxe_core/src/lib.rs
```diff
@@ -269,6 +271,9 @@ where
         PROTOCOL_DB.init_protocol_db();
         // Initialize full allocation support.
         allocator::init_memory_support(&hob_list);
+        // Initialize the debugger if it is enabled.
+        uefi_debugger::initialize(&mut self.interrupt_manager);
```
os-d: If you are using memory in the initialization, we need to do this after the HOB list relocation. On Q35, the PEI HOB list is in unmapped memory and that probably will always be true, but not necessarily. If anything allocates between init_memory_support and relocate_hobs, it could allocate over the HOB list.

PR #207, opened 2025-01-08
The GCD currently defines its own error codes and needs to align with rest of the reusable crates.

HappyCoding-2025:
https://github.com/pop-project/uefi-dxe-core/issues/174
The GCD currently defines its own error codes. However, it constantly needs to convert to and from efi::Status. I believe that this is a legacy of the GCD being one of the first components written.

SInce the scope of changes seemed bigger than expected,  the mappnig error  need to be fixed_size_block_allocator.rs and will take of this in next commit.

Follow  the rest of dxe_core behavior and use efi::Status instead of custom error codes. Custom error codes should be used for non-UEFI specific reusable crates.

<_Include a description of the change and why this change was made._>

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

<_Describe the test(s) that were run to verify the changes._>

## Integration Instructions

<_Describe how these changes should be integrated. Use N/A if nothing is required._>

### DIFF
--- a/dxe_core/src/dxe_services.rs
+++ b/dxe_core/src/dxe_services.rs
@@ -35,7 +35,7 @@ extern "efiapi" fn add_memory_space(
 
     match result {
         Ok(_) => efi::Status::SUCCESS,
-        Err(err) => efi::Status::from(err),
+        Err(err) => err,
     }
 }
 
@@ -84,7 +84,7 @@ extern "efiapi" fn allocate_memory_space(
             unsafe { base_address.write(allocated_addr as u64) };
             efi::Status::SUCCESS
         }
-        Err(err) => efi::Status::from(err),
+        Err(err) => err,
     }
 }
 
@@ -93,7 +93,7 @@ extern "efiapi" fn free_memory_space(base_address: efi::PhysicalAddress, length:
 
     match result {
         Ok(_) => efi::Status::SUCCESS,
-        Err(err) => efi::Status::from(err),
+        Err(err) => err,
     }
 }
 
@@ -101,7 +101,7 @@ extern "efiapi" fn remove_memory_space(base_address: efi::PhysicalAddress, lengt
     let result = GCD.remove_memory_space(base_address as usize, length as usize);
     match result {
         Ok(_) => efi::Status::SUCCESS,
-        Err(err) => efi::Status::from(err),
+        Err(err) => err,
     }
 }
 
@@ -181,7 +181,7 @@ extern "efiapi" fn get_memory_space_map(
     let result = GCD.get_memory_descriptors(&mut descriptors);
 
     if let Err(err) = result {
-        return efi::Status::from(err);
+        return err;
     }
 
     //caller is supposed to free the handle buffer using free pool, so we need to allocate it using allocate pool.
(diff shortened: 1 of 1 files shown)


Javagedes:
I believe it would be better to use `EfiError` from `uefi_sdk` rather than the `usize` from `r-efi` That is the error interface we are leaning into for anything pure rust and I think this qualifies. `EfiError` provides an easy conversion into an equivalent `usize` error when working with any of the boot services functions:

https://github.com/pop-project/uefi-sdk/blob/fe654fbd5ffddc33fb75fff7f6316ce0ce42f66e/uefi_sdk/src/error.rs#L54

@os-d Do you concur? This will mainily affect you.

os-d:
> I believe it would be better to use `EfiError` from `uefi_sdk` rather than the `usize` from `r-efi` That is the error interface we are leaning into for anything pure rust and I think this qualifies. `EfiError` provides an easy conversion into an equivalent `usize` error when working with any of the boot services functions:
> 
> https://github.com/pop-project/uefi-sdk/blob/fe654fbd5ffddc33fb75fff7f6316ce0ce42f66e/uefi_sdk/src/error.rs#L54
> 
> @os-d Do you concur? This will mainily affect you.

@Javagedes I agree. My ask in the issue was just to remove this custom gcd::Error type and conform to what the rest of the code is using inside of uefi-dxe-core. I am agnostic to what that is, just that we are consistent.

HappyCoding-2025:
Thanks @Javagedes  @os-d  - The entire dxe-core is using r_efi::efi;
Please clarify if the entire dxe-core needs to change? The original discussion was only aligning gcd to rest of dxe-core

For E.g - 
dxe_services.rs
memory_log.rs
lib.rs
component.rs
extended.rs
uefi_allocator.rs
io_block.rs

Javagedes:
Hi! Sorry for the confusion. using `r_efi` in general is completely fine. We want to use it. What I meant was we want to use `uefi_sdk::error::EfiError` instead of `r_efi::efi::Status` in almost all cases.

The one scenario where this is not true is in any EDKII callable code, annotated by `extern "efiapi" fn ...` in the rust code. These functions are the external hook points that C written drivers can interact with the rust-written dxe-core. The interface for these functions are spec defined [here](https://uefi.org/specs/UEFI/2.10/07_Services_Boot_Services.html) and require a return value of `EFI_STATUS` (in `C`) which is `efi::Status` in `rust`.

There are some areas within our codebase that still use `efi::Status` when it should probably use `uefi_sdk::error::EfiError`, however I don't think updating that needs to be a part of this particular PR.

HappyCoding-2025:
thank you @Javagedes

HappyCoding-2025:
@Javagedes  - Thanks for clariyfing this

HappyCoding-2025:
Created a new Pull request and copied all the comments. Looks like git rebase to main and git pull and push is causing other commits showing up as well.
New Pull Requet - https://github.com/pop-project/uefi-dxe-core/pull/208

Javagedes:
Closing in favor of #215

PR #208, opened 2025-01-09
Personal/rkartheek/sup error2

HappyCoding-2025:
This is same as Pull request https://github.com/pop-project/uefi-dxe-core/pull/207
https://github.com/pop-project/uefi-dxe-core/pull/207  with rebase and push shows other commits too. Hence creating a new one.

The GCD currently defines its own error codes. However, it constantly needs to convert to and from efi::Status. I believe that this is a legacy of the GCD being one of the first components written.

SInce the scope of changes seemed bigger than expected, the mappnig error need to be fixed_size_block_allocator.rs and will take of this in next commit.

Follow the rest of dxe_core behavior and use efi::Status instead of custom error codes. Custom error codes should be used for non-UEFI specific reusable crates.

How This Was Tested

<Describe the test(s) that were run to verify the chang

### DIFF
--- a/dxe_core/src/allocator/fixed_size_block_allocator.rs
+++ b/dxe_core/src/allocator/fixed_size_block_allocator.rs
@@ -435,21 +435,14 @@ impl FixedSizeBlockAllocator {
 
         // Page allocations and pool allocations are disjoint; page allocations are allocated directly from the GCD and are
         // freed straight back to GCD. As such, a tracking allocator structure is not required.
-        let start_address = self
-            .gcd
-            .allocate_memory_space(
-                allocation_strategy,
-                memory_type,
-                UEFI_PAGE_SHIFT,
-                pages * ALIGNMENT,
-                self.handle,
-                None,
-            )
-            .map_err(|err| match err {
-                efi::Status::INVALID_PARAMETER => efi::Status::INVALID_PARAMETER,
-                efi::Status::NOT_FOUND => efi::Status::NOT_FOUND,
-                _ => efi::Status::OUT_OF_RESOURCES,
-            })?;
+        let start_address = self.gcd.allocate_memory_space(
+            allocation_strategy,
+            memory_type,
+            UEFI_PAGE_SHIFT,
+            pages * ALIGNMENT,
+            self.handle,
+            None,
+        )?;
 
         let allocation = slice_from_raw_parts_mut(start_address as *mut u8, pages * ALIGNMENT);
         let allocation = NonNull::new(allocation).ok_or(efi::Status::OUT_OF_RESOURCES)?;
@@ -476,28 +469,18 @@ impl FixedSizeBlockAllocator {
             return Err(efi::Status::INVALID_PARAMETER);
         }
 
-        let descriptor =
-            self.gcd.get_memory_descriptor_for_address(address as efi::PhysicalAddress).map_err(|err| match err {
-                efi::Status::NOT_FOUND => efi::Status::NOT_FOUND,
-                _ => efi::Status::INVALID_PARAMETER,
-            })?;
+        let descriptor = self.gcd.get_memory_descriptor_for_address(address as efi::PhysicalAddress)?;
 
         if descriptor.image_handle != self.handle {
             Err(efi::Status::NOT_FOUND)?;
         }
 
         if self.preferred_range.as_ref().is_some_and(|range| range.contains(&(address as efi::PhysicalAddress))) {
-            self.gcd.free_memory_space_preserving_ownership(address, pages * ALIGNMENT).map_err(|err| match err {
-                efi::Status::NOT_FOUND => efi::Status::NOT_FOUND,
-                _ => efi::Status::INVALID_PARAMETER,
-            })?;
+            self.gcd.free_memory_space_preserving_ownership(address, pages * ALIGNMENT)?;
             self.stats.reserved_used -= pages * ALIGNMENT;
             // don't update claimed_pages stats here, because they are never actually "released".
         } else {
-            self.gcd.free_memory_space(address, pages * ALIGNMENT).map_err(|err| match err {
-                efi::Status::NOT_FOUND => efi::Status::NOT_FOUND,
-                _ => efi::Status::INVALID_PARAMETER,
-            })?;
+            self.gcd.free_memory_space(address, pages * ALIGNMENT)?;
             self.stats.claimed_pages -= pages;
         }
 


HappyCoding-2025:
Capturing comment given earlier

[Javagedes](https://github.com/Javagedes) commented [18 hours ago](https://github.com/pop-project/uefi-dxe-core/pull/207#issuecomment-2578888892)
I believe it would be better to use EfiError from uefi_sdk rather than the usize from r-efi That is the error interface we are leaning into for anything pure rust and I think this qualifies. EfiError provides an easy conversion into an equivalent usize error when working with any of the boot services functions:

https://github.com/pop-project/uefi-sdk/blob/fe654fbd5ffddc33fb75fff7f6316ce0ce42f66e/uefi_sdk/src/error.rs#L54

@os-d Do you concur? This will mainily affect you.

@[os-d](https://github.com/os-d)
os-d commented [18 hours ago](https://github.com/pop-project/uefi-dxe-core/pull/207#issuecomment-2578897118)
I believe it would be better to use EfiError from uefi_sdk rather than the usize from r-efi That is the error interface we are leaning into for anything pure rust and I think this qualifies. EfiError provides an easy conversion into an equivalent usize error when working with any of the boot services functions:

https://github.com/pop-project/uefi-sdk/blob/fe654fbd5ffddc33fb75fff7f6316ce0ce42f66e/uefi_sdk/src/error.rs#L54

@os-d Do you concur? This will mainily affect you.

@Javagedes I agree. My ask in the issue was just to remove this custom gcd::Error type and conform to what the rest of the code is using inside of uefi-dxe-core. I am agnostic to what that is, just that we are consistent.

@kuqin12
[Adding CPU arch support for AArch64 (](https://github.com/pop-project/uefi-dxe-core/pull/207/commits/4bd63a12a38e29afead5a4ff907347e81e4c0361)https://github.com/pop-project/uefi-dxe-core/pull/203[)](https://github.com/pop-project/uefi-dxe-core/pull/207/commits/4bd63a12a38e29afead5a4ff907347e81e4c0361) 
4bd63a1
@HappyCoding-2025
Author
[HappyCoding-2025](https://github.com/HappyCoding-2025) commented [15 hours ago](https://github.com/pop-project/uefi-dxe-core/pull/207#issuecomment-2579017566)
Thanks @Javagedes @os-d - The entire dxe-core is using r_efi::efi;
Please clarify if the entire dxe-core needs to change? The original discussion was only aligning gcd to rest of dxe-core

For E.g -
dxe_services.rs
memory_log.rs
lib.rs
component.rs
extended.rs
uefi_allocator.rs
io_block.rs

@[Javagedes](https://github.com/Javagedes)
Javagedes commented [15 hours ago](https://github.com/pop-project/uefi-dxe-core/pull/207#issuecomment-2579045449)
Hi! Sorry for the confusion. using r_efi in general is completely fine. We want to use it. What I meant was we want to use uefi_sdk::error::EfiError instead of r_efi::efi::Status in almost all cases.

The one scenario where this is not true is in any EDKII callable code, annotated by extern "efiapi" fn ... in the rust code. These functions are the external hook points that C written drivers can interact with the rust-written dxe-core. The interface for these functions are spec defined [here](https://uefi.org/specs/UEFI/2.10/07_Services_Boot_Services.html) and require a return value of EFI_STATUS (in C) which is efi::Status in rust.

There are some areas within our codebase that still use efi::Status when it should probably use uefi_sdk::error::EfiError, however I don't think updating that needs to be a part of this particular PR.

@[HappyCoding-2025](https://github.com/HappyCoding-2025)
Author
HappyCoding-2025 commented [11 hours ago](https://github.com/pop-project/uefi-dxe-core/pull/207#issuecomment-2579292168)
thank you @Javagedes

@HappyCoding-2025
Author
[HappyCoding-2025](https://github.com/HappyCoding-2025) commented [11 hours ago](https://github.com/pop-project/uefi-dxe-core/pull/207#issuecomment-2579292725)
@Javagedes - Thanks for clariyfing this

dxe_core/src/allocator/fixed_size_block_allocator.rs
```diff
@@ -449,8 +446,8 @@ impl FixedSizeBlockAllocator {
                 None,
             )
             .map_err(|err| match err {
-                gcd::Error::InvalidParameter => efi::Status::INVALID_PARAMETER,
-                gcd::Error::NotFound => efi::Status::NOT_FOUND,
+                efi::Status::INVALID_PARAMETER => efi::Status::INVALID_PARAMETER,
+                efi::Status::NOT_FOUND => efi::Status::NOT_FOUND,
```
os-d: Looks like this map_err can entirely go away

dxe_core/src/allocator/fixed_size_block_allocator.rs
```diff
@@ -481,7 +478,7 @@ impl FixedSizeBlockAllocator {
 
         let descriptor =
             self.gcd.get_memory_descriptor_for_address(address as efi::PhysicalAddress).map_err(|err| match err {
-                gcd::Error::NotFound => efi::Status::NOT_FOUND,
+                efi::Status::NOT_FOUND => efi::Status::NOT_FOUND,
```
os-d: Similarly, all these can go away that map the same value to the same value.

dxe_core/src/gcd/spin_locked_gcd.rs
```diff
@@ -446,11 +386,11 @@ impl GCD {
         base_address: usize,
         len: usize,
         capabilities: u64,
-    ) -> Result<usize, Error> {
-        ensure!(self.maximum_address != 0, Error::NotInitialized);
+    ) -> Result<usize, efi::Status> {
```
os-d, line 389: Looks like you did not move to Joey's suggestion of `EfiError` instead of `efi::Status`. Can you update all the non-efiapi functions to return `EfiError`?

dxe_core/src/allocator/fixed_size_block_allocator.rs
HappyCoding-2025: Thanks Oliver, addressed and taken care .

dxe_core/src/allocator/fixed_size_block_allocator.rs
HappyCoding-2025: Fixed now.

dxe_core/src/gcd/spin_locked_gcd.rs
HappyCoding-2025: Hi Oliver, Entire Dxe-core is using efi::Status only.  Joey commented on this saying efi::Status is fine.

From Joey -
Hi! Sorry for the confusion. using r_efi in general is completely fine. We want to use it. What I meant was we want to use uefi_sdk::error::EfiError instead of r_efi::efi::Status in almost all cases.

The one scenario where this is not true is in any EDKII callable code, annotated by extern "efiapi" fn ... in the rust code. These functions are the external hook points that C written drivers can interact with the rust-written dxe-core. The interface for these functions are spec defined [here](https://uefi.org/specs/UEFI/2.10/07_Services_Boot_Services.html) and require a return value of EFI_STATUS (in C) which is efi::Status in rust.

There are some areas within our codebase that still use efi::Status when it should probably use uefi_sdk::error::EfiError, however I don't think updating that needs to be a part of this particular PR.

makubacki:
@HappyCoding-2025, we like to keep all comments and changes related to a change in a single pull request. In the future, you can rebase and force push your PR branch to the same PR.

dxe_core/src/allocator/fixed_size_block_allocator.rs
makubacki, line 1: Please update the PR title to something meaningful.

Javagedes:
Closing in favor of #215

PR #209, opened 2025-01-10
Changes for internal crate publishing

makubacki:
## Description

Minor updates to ReleaseWorkflow.yml to handle publishing to an internal feed.

Add the feed info in .cargo/config.toml.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Published crates from uefi-sdk.

- Worfklow Run: [Publish Release](https://github.com/pop-project/uefi-sdk/actions/runs/12700169589)
- Feed with published crates: [UefiRust](https://dev.azure.com/microsoft/MsUEFI/_artifacts/feed/UefiRust)

## Integration Instructions

- Repos using the workflow should be publishing to the `UefiRust` feed for now.

### DIFF
--- a/.github/workflows/ReleaseWorkflow.yml
+++ b/.github/workflows/ReleaseWorkflow.yml
@@ -34,12 +34,12 @@ jobs:
 
       - name: 🛠️ Download Rust Tools 🛠️
         uses: pop-project/uefi-dxe-core/.github/actions/rust-tool-cache@main
-      
+
       - name: Add Credentials # Remove once repositories are public
         run: |
           git config --global url."https://${{ secrets.GH_PAT }}@github.com".insteadOf https://github.com
         shell: bash
-      
+
       - name: Get Current Draft Release
         id: draft_release
         uses: actions/github-script@v7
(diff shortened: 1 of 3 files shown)


.github/workflows/ReleaseWorkflow.yml
```diff
@@ -68,25 +68,33 @@ jobs:
...
+
+      - name: Setup Cred Provider
+        run: |
+          mkdir -p ~/.cargo
+          cat <<EOF >> ~/.cargo/config.toml
+          [registry]
+          global-credential-providers = ["cargo:token"]
+          EOF
+
+      - name: Login to Crates.io
```
Javagedes: Do we want to make the feed name that we login to a parameter? Or just revert this commit when we go public?

Once we go public we just need to login with the token, but not to a particular feed I think?

makubacki: It's probably whatever the two of us prefer since we'll do the work. It should just be the token with no feed for crates.io. I'm okay with simply dropping this versus the parameter, but I'm also fine to parameterize it if you'd prefer that.

Javagedes: Thinking about it more, the current way is probably best. If we make it a parameter then we have to update all repos when we go public. With this way, we just revert the change and it cascades to all repos automatically.

PR #211, opened 2025-01-10
rust-tool-cache: Gracefully handle rust-toolchain.toml issues

makubacki:
## Description

Print obvious errors if either the file is missing or the `[tools]` section is not found.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- rust-toolchain.toml not present
- [tools] section missing in rust-toolchain.toml
- [tools] section present in rust-toolchain.toml

## Integration Instructions

- N/A

### DIFF
--- a/.github/actions/rust-tool-cache/action.yml
+++ b/.github/actions/rust-tool-cache/action.yml
@@ -22,16 +22,28 @@ runs:
           ~/.cargo/bin/
           ~/.rustup/toolchains/
         key: ${{ runner.os }}-rust-tools-${{ hashFiles('**/rust-toolchain.toml' )}}
-    
+
     - name: Install cargo-binstall
-      uses: cargo-bins/cargo-binstall@v1.10.17      
+      uses: cargo-bins/cargo-binstall@v1.10.17
 
     # Read any tools from rust-toolchain.toml file and installs them
     - name: Install Rust Tools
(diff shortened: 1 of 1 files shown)


Javagedes:
Depending on how generic you want to be, I feel like these are technically not errors. For us, in these repos, they are errors as we always need at least cargo make and cargo tarpaulin. I agree a nice warning message is very useful. But I'm wondering if they should just exit gracefully with a success rather than fail. 

Overall, I'm fine with this being an error, but its food for thought.

**edit**: At the same time, the point of this workflow is to cache these tools, so an error makes sense if the file (or tools) do not exist, but the workflow is being used.

makubacki:
> Depending on how generic you want to be, I feel like these are technically not errors. For us, in these repos, they are errors as we always need at least cargo make and cargo tarpaulin. I agree a nice warning message is very useful. But I'm wondering if they should just exit gracefully with a success rather than fail.
> 
> Overall, I'm fine with this being an error, but its food for thought.
> 
> **edit**: At the same time, the point of this workflow is to cache these tools, so an error makes sense if the file (or tools) do not exist, but the workflow is being used.

I went down the line of thinking as your edit update.

PR #212, opened 2025-01-10
dxe_core: events: handle all event_notify events

### LINKED ISSUE #199: [Bug]: TPL restore is iterating on a fixed vector of events

### Current Behavior

The current implementation of the restore_tpl in events.rs will gather a vector of events and iterate through them, and dispatch necessary notification as it sees fit.

However, while dispatching the notification, it is possible that new events will occur and queue up new events, which needs to get dispatched when TPL is descending. The current flow does not account for newly added events.

### Expected Behavior

The current flow needs to account for newly added events while descending the TPL.

### Steps To Reproduce

Just do code review.

### Build Environment

```markdown
- OS(s): Windows
- Tool Chain(s): LLVM
- Targets Impacted: Debug & Release
```

### Version Information

```text
Top of main.
```

### Urgency

Medium

Someone else needs to fix it

Maintainer feedback requested

### Anything else?

_No response_

Javagedes:
## Description

When a TPL is restored, any notify events are processed, which may result in a notify callback being called. This callback may result in a notify events being added to the event queue. This commit will continuously process all events in the event queue until the queue is empty.

Additionally, this commit moves the `EVENT_NOTIFIES_IN_PROGRESS` static inside of the `restore_tpl` function, as it is only used in that function.

closes #199 

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Verified continued boot to the UEFI shell for QemuQ35

## Integration Instructions

N/A

### DIFF
--- a/dxe_core/src/events.rs
+++ b/dxe_core/src/events.rs
@@ -18,7 +18,7 @@ use mu_pi::protocols::timer;
 use uefi_cpu::interrupts;
 
 use crate::{
-    event_db::{EventNotification, SpinLockedEventDb, TimerDelay},
+    event_db::{SpinLockedEventDb, TimerDelay},
     gcd,
     protocols::PROTOCOL_DB,
 };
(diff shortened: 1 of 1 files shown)


Javagedes:
Talked to @joschock offline. It looks like we can get away with just not calling `collect()` on the `event_notification_iter()` call, as that directly references the EVENT_DB when pulling out events. To get away with this, we will have to extend the time in which `EVENT_NOTIFIES_IN_PROGRESS` is true to be the entire processing time rather than just long enough to call `event_notification_iter().collect()`. 

I will make the update, and run SCTs to ensure no corner cases get re-exposed with this change.

Changing this to draft in the meantime.

Javagedes:
Updated to v2. Still need to run SCTs before marking this as ready for review.

dxe_core/src/events.rs
```diff
@@ -238,37 +235,30 @@ pub extern "efiapi" fn restore_tpl(new_tpl: efi::Tpl) {
         // To avoid infinite recursion, this logic uses EVENT_NOTIFIES_IN_PROGRESS to ensure that only one instance of
         // restore_tpl is accessing the locked EVENT_DB. restore_tpl calls that occur while the event notification iter is
         // in use will get back an empty vector of event notifications and will simply restore the TPL and exit.
```
joschock, line 237: This part of the comment is no longer accurate.

joschock: I think there is a subtle difference relative to the C code (in the original implementation as well as with this change). I think in the C code, TPL at a higher-level will pre-empt notifications running at lower level, and if it schedules an event notify at a higher TPL than the one that was interrupted, that will also run (on the restore_tpl from the higher level) before control is returned to the original TPL. With this implementation, the new notification will be inserted at the front of the event notification queue, and the event_notification_iter() will return it next, but it won't be executed until the original event notification runs to completion. In other words, a higher TPL event notification callback will not pre-empt a lower TPL event notification in progress, whereas in the C code, it would. The spec has no detailed requirements in terms of these orderings, and the SCT event tests passed with this delta, so I think it is okay. I'll leave it to your judgement whether this difference merits a big comment or not :).

joschock: I guess it didn't show what I highlighted: `will get back an empty vector of event notifications` is no longer accurate.

dxe_core/src/events.rs
```diff
@@ -238,37 +235,30 @@ pub extern "efiapi" fn restore_tpl(new_tpl: efi::Tpl) {
...
-        let events =
-            match EVENT_NOTIFIES_IN_PROGRESS.compare_exchange(false, true, Ordering::Acquire, Ordering::Relaxed) {
-                Ok(_) => {
-                    let events = EVENT_DB.event_notification_iter(new_tpl).collect();
-                    EVENT_NOTIFIES_IN_PROGRESS.store(false, Ordering::Release);
-                    events
+        static EVENT_NOTIFIES_IN_PROGRESS: AtomicBool = AtomicBool::new(false);
+        if EVENT_NOTIFIES_IN_PROGRESS.compare_exchange(false, true, Ordering::Acquire, Ordering::Relaxed).is_ok() {
+            // The notify function may register new event notifications, which will be processed in this same iter so
+            // long as we do not collect the events before iterating over them.
```
makubacki, line 241: It would be nice to have a unit test that verifies this.

kouchekiniad:
@Javagedes do you have any objections to me rebasing/taking this out of draft? To prevent notify_wait events from breaking more than they are already, I believe I need this change to go in before #415

Javagedes:
@kouchekiniad Sure. Your choice. Feel free to rebase it into your PR, or if you like, I can open it back up.

Javagedes:
This is being merged as a part of #415

PR #215, opened 2025-01-15
gcd: Accommodating EfiError

HappyCoding-2025:
The GCD currently defines its own error codes. However, it constantly needs to convert to EfiError. I believe that this is a legacy of the GCD being one of the first components written.

Follow the rest of dxe_core behavior and use EfiError instead of custom error codes. Custom error codes should be used for non-UEFI specific reusable crates.

How This Was Tested

Coverage was good

Cargo make test also was good.

### DIFF
--- a/dxe_core/src/allocator/fixed_size_block_allocator.rs
+++ b/dxe_core/src/allocator/fixed_size_block_allocator.rs
@@ -12,10 +12,10 @@
 
 extern crate alloc;
 use super::{AllocationStrategy, DEFAULT_ALLOCATION_STRATEGY};
-use crate::{
-    gcd::{self, SpinLockedGcd},
-    tpl_lock,
-};
+
+use crate::{gcd::SpinLockedGcd, tpl_lock};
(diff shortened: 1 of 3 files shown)


dxe_core/src/allocator/fixed_size_block_allocator.rs
```diff
@@ -11,11 +11,14 @@
...
+use uefi_sdk::error::EfiError;
 use super::{AllocationStrategy, DEFAULT_ALLOCATION_STRATEGY};
 use crate::{
-    gcd::{self, SpinLockedGcd},
-    tpl_lock,
-};
+        gcd::{self, SpinLockedGcd},
+        tpl_lock,
+    };
+/* use crate::{gcd::SpinLockedGcd, tpl_lock}; */
```
Javagedes: delete comment.

makubacki:
> I'm going to fix/clean up the PR branch before submission. This has too many irrelevant changes right now.

Done.

PR #219, opened 2025-01-17
More perf fixes: macros

berlin-with0ut-return:
## Description
Perf macros are currently incorrectly scoped and have some syntax errors. These have been fixed. In addition, improve macro ergonomics through re-export of `function!` from `mu_rust_helpers`.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested
I instrumented these macros into dispatcher code for testing. They now compile and provide the correct output when instrumented.

## Integration Instructions
N/A.

### DIFF
--- a/crates/uefi_performance/src/lib.rs
+++ b/crates/uefi_performance/src/lib.rs
@@ -52,10 +52,9 @@ use performance_table::FBPT;
 
 use r_efi::system::EVENT_GROUP_READY_TO_BOOT;
 
-use mu_rust_helpers::{
-    function,
-    perf_timer::{Arch, ArchFunctionality},
-};
+pub use mu_rust_helpers::function;
+use mu_rust_helpers::perf_timer::{Arch, ArchFunctionality};
+
 use uefi_sdk::{
     boot_services::{event::EventType, tpl::Tpl, BootServices, StandardBootServices},
     guid,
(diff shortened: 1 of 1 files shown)


crates/uefi_performance/src/lib.rs
```diff
@@ -589,10 +589,11 @@ pub fn perf_event(event_string: &str, caller_id: &efi::Guid) {
     );
 }
 
+// SHERRY: the prepended mod names are necessary (but ugly) due to the way Rust scopes macros
 #[macro_export]
 macro_rules! perf_event_signal_begin {
-    (&event_guid:expr) => {
-        _perf_event_signal_begin($event_guid, function()!)
+    ($event_guid:expr, $caller_id:expr) => {
+        uefi_performance::_perf_event_signal_begin($event_guid, mu_rust_helpers::function!(), $caller_id)
```
Javagedes: The ergonomics are poor for this macro. It is going to require that `uefi_performance` and `mu_rust_helpers` are in scope, in the file that is using this macro. If not, it leads to some confusing errors for the user (see this PR where I fixed a similar issue in mu_rust_helpers: https://github.com/microsoft/mu_rust_helpers/commit/d44a8b9f95934ce893a88b154f9087035d8ad77f, https://github.com/microsoft/mu_rust_helpers/pull/37)

My suggestion is you replace `uefi_performance::` with `$crate::`, and do a public re-export of `mu_rust_helpers::function!()` so that you can also do `$crate::function!()`

Javagedes: Same with the other macros you are fixing.

Javagedes: remember, macros are drop in compile time replacements, that basically replace the macro with whatever the macro generates. [cargo-expand](https://crates.io/crates/cargo-expand) is a great tool to see what gets generated.

PR #220, opened 2025-01-17
GCD logic update to accommodate AArch64 usage

kuqin12:
## Description

This change contains a few minor bug fixes:
1. The get memory map call will include reserved memory regions from the GCD.
2. The AllocateAddress type into page allocation will not be allowed for random type.
3. Follow up with a set memory attribute call after the gcd add a memory space region for the cache attributes.
4. All the memory map manipulation will need to contain the cache attributes.
5. Updated the set capability transition check to inspect that the new capabilities need to cover all existing attributes.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This change is tested on QEMU SBSA and booted to Windows.

## Integration Instructions

N/A

### DIFF
--- a/dxe_core/src/gcd/memory_block.rs
+++ b/dxe_core/src/gcd/memory_block.rs
@@ -252,7 +252,14 @@ impl MemoryBlock {
                 if let Some(device_handle) = device_handle {
                     md.device_handle = device_handle;
                 }
+                // Whenever memory is allocated, we need to set it as XP and preserve the cache
+                // attributes. This is done at this level because in early initialization, the GCD and
+                // page table are not initialized and various checks will fail when done from the
+                // top down. Those components will be synchronized when initialization of them
+                // completes and so this addition becomes a no-op later, but is required at the
+                // beginning.
                 md.attributes = (md.attributes & efi::CACHE_ATTRIBUTE_MASK) | efi::MEMORY_XP;
+
                 *self = Self::Allocated(*md);
                 Ok(())
             }


dxe_core/src/gcd.rs
```diff
@@ -211,6 +209,16 @@ pub fn add_hob_resource_descriptors_to_gcd(hob_list: &HobList) {
                     .expect("Failed to add memory space to GCD");
                 }
             }
+            if let Hob::ResourceDescriptor(res_desc) = hob {
+                let memory_attributes = (MemoryAttributes::from_bits_truncate(res_desc.attributes) & MemoryAttributes::CacheAttributesMask).bits() as u64;
+                match GCD.set_memory_space_attributes(res_desc.physical_start as usize, res_desc.resource_length as usize, memory_attributes) {
+                    Err(Error::NotInitialized) => {
+                        log::warn!("GCD not initialized, skipping memory attribute setting");
+                    }
+                    _ => { panic!("Failed to set memory attributes"); }
```
Javagedes: Is panicking the right thing to do here? maybe a debug_assert or something but a panic under all circumstances seems wrong to me.

Javagedes: Also, regardless of if you stick with panic, or switch to a log message or a debug assert, I think you need more information than just "Failed to set memory attributes". 

Probably log (or in the panic message) the hob, or the attributes, or the addresss and length, or something along those lines.

dxe_core/src/gcd.rs
```diff
@@ -211,6 +209,16 @@ pub fn add_hob_resource_descriptors_to_gcd(hob_list: &HobList) {
                     .expect("Failed to add memory space to GCD");
                 }
             }
+            if let Hob::ResourceDescriptor(res_desc) = hob {
+                let memory_attributes = (MemoryAttributes::from_bits_truncate(res_desc.attributes) & MemoryAttributes::CacheAttributesMask).bits() as u64;
+                match GCD.set_memory_space_attributes(res_desc.physical_start as usize, res_desc.resource_length as usize, memory_attributes) {
+                    Err(Error::NotInitialized) => {
+                        log::warn!("GCD not initialized, skipping memory attribute setting");
+                    }
```
os-d:
```suggestion
                    Err(Error::NotInitialized) => (),
```

This is the expected case

dxe_core/src/gcd.rs
os-d: Panic is probably the right thing to do here, because it means two things:
- We are going to fail to set cache attributes on that range which will lead to undefined behavior (probably a crash somewhere weird)
- The GCD is borked from the very beginning if we can't apply attributes to it. That should be caught and fixed as who knows what else will cascade from that

This function is called once at GCD init, so we only have one circumstance to think about

dxe_core/src/gcd.rs
```diff
@@ -211,6 +209,16 @@ pub fn add_hob_resource_descriptors_to_gcd(hob_list: &HobList) {
...
             }
+            if let Hob::ResourceDescriptor(res_desc) = hob {
+                let memory_attributes = (MemoryAttributes::from_bits_truncate(res_desc.attributes) & MemoryAttributes::CacheAttributesMask).bits() as u64;
+                match GCD.set_memory_space_attributes(res_desc.physical_start as usize, res_desc.resource_length as usize, memory_attributes) {
+                    Err(Error::NotInitialized) => {
+                        log::warn!("GCD not initialized, skipping memory attribute setting");
+                    }
+                    _ => { panic!("Failed to set memory attributes"); }
+                }
+                log::error!("GCD after set mem: {}", GCD);
```
os-d:
```suggestion
```

Or make it a trace with a target that can be filtered out

dxe_core/src/gcd/spin_locked_gcd.rs
```diff
@@ -627,8 +628,12 @@ impl GCD {
                     "Mapping memory region {:#x?} of length {:#x?} with attributes {:#x?}",
                     desc.base_address,
                     desc.length,
-                    desc.attributes
-                );
+                    (MemoryAttributes::from_bits_truncate(desc.attributes) & MemoryAttributes::CacheAttributesMask)
+                        | MemoryAttributes::ExecuteProtect);
+
+                if desc.attributes & efi::MEMORY_RP != 0 {
```
os-d, line 591:
```suggestion
                // we have large ranges of memory that have been allocated but then freed
                // this is what the memory bin code does. We should only map what memory
                // is actually being used. On future allocations of this region, the pages will
                // get mapped again.
                if desc.attributes & efi::MEMORY_RP != 0 {

```

dxe_core/src/gcd/spin_locked_gcd.rs
```diff
@@ -2466,7 +2477,15 @@ impl SpinLockedGcd {
...
             let current_len = next_base - current_base;
-            self.memory.lock().set_memory_space_attributes(current_base as usize, current_len as usize, attributes)?;
+            log::error!("Setting attributes for {:#X} - {:#X} with attributes {:#X}, len {:#X}, range_end {:#X}", current_base, next_base, attributes, len, range_end);
+            match self.memory.lock().set_memory_space_attributes(current_base as usize, current_len as usize, attributes) {
+                Err(Error::NotInitialized) => {
+                    log::warn!("Set attr returned not initialized, skipping memory attribute setting");
+                    res = Err(Error::NotInitialized);
+                }
+                Ok(())  => {},
+                _ => { panic!("Failed to set memory attributes"); }
```
os-d:
```suggestion
            match self.memory.lock().set_memory_space_attributes(current_base as usize, current_len as usize, attributes) {
                Err(Error::NotInitialized) => {
                    // before the page table is installed, we expect to get a return of NotInitialized. This means the GCD
                    // has been updated with the attributes, but the page table is NotInitialized yet. In init_paging, the
                    // page table will be updated with the current state of the GCD. The code that calls into this expects
                    // NotInitialized to be returned, so we must catch that error and report it. However, we also need to
                    // make sure any attribute updates across descriptors update the full range and not error out here.
                    res = Err(Error::NotInitialized);
                }
                Ok(())  => {},
                _ => { 
                // some nicer log here
                debug_assert!(false);
                }

```

dxe_core/src/gcd.rs
kuqin12: Will add more information in the print.

kuqin12: Updated in the latest push.

kuqin12:
Tested locally again on Q35 and SBSA. Still bootable :)

dxe_core/src/gcd/memory_block.rs
```diff
@@ -252,6 +252,7 @@ impl MemoryBlock {
                 if let Some(device_handle) = device_handle {
                     md.device_handle = device_handle;
                 }
+                md.attributes = (md.attributes & efi::CACHE_ATTRIBUTE_MASK) | efi::MEMORY_XP;
```
os-d, line 261:
```suggestion
                // Whenever memory is allocated, we need to set it as XP and preserve the cache
                // attributes. This is done at this level because in early initialization, the GCD and
                // page table are not initialized and various checks will fail when done from the
                // top down. Those components will be synchronized when initialization of them
                // completes and so this addition becomes a no-op later, but is required at the
                // beginning.
                md.attributes = (md.attributes & efi::CACHE_ATTRIBUTE_MASK) | efi::MEMORY_XP;

```

PR #224, opened 2025-01-22
dxe_core: Use Type-State pattern

Javagedes:
Convert the dxe_core Core struct to use the Type-State pattern to prevent allocations for the short period of time before the global allocator is fully initialized instead of having two completely separate structs for the two different states.

The Type-State pattern is a way to gate functionality to certain states by using the type system and phantom data to implement functions for certain states and prevent them, at compile time, from being used in other states.

## Description

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

CI passes and the platform builds with no integration steps

## Integration Instructions

N/A

### DIFF
--- a/dxe_core/src/lib.rs
+++ b/dxe_core/src/lib.rs
@@ -161,11 +161,12 @@ pub struct NoAlloc;
 /// to the `Alloc` phase. Each phase provides a subset of methods that are available to the struct, allowing
 /// for a more controlled configuration and execution process.
 ///
-/// During the `NoAlloc` phase, the struct provides methods to configure the dxe core's cpu related
-/// functionality, the section extractor. During this time, no allocations are available.
+/// During the `NoAlloc` phase, the struct provides methods to configure the DXE core environment
+/// prior to allocation capability such as CPU functionality and section extraction. During this time,
+/// no allocations are available.
 ///
 /// Once the [initialize](Core::initialize) method is called, the struct transitions to the `Alloc` phase,
-/// which provides methods for adding configuration and components with the dxe core, and eventually starting the
+/// which provides methods for adding configuration and components with the DXE core, and eventually starting the
 /// dispatching process and eventual handoff to the BDS phase.
 ///
 /// ## Examples


dxe_core/src/lib.rs
```diff
@@ -146,14 +146,27 @@ macro_rules! error {
...
-/// initializing the system and applying any configuration from the `with_*` function calls. During this phase, no
-/// allocations are available. Allocations are only available once [initialize](Core::initialize) is called.
+/// This struct is the entry point for the DXE Core, which is a two phase system. The current phase is denoted by the
+/// current struct representing the generic parameter "MemoryState". Creating a [Core] object will initialize the
+/// struct in the `NoAlloc` phase. Calling the [initialize](Core::initialize) method will transition the struct
+/// to the `Alloc` phase. Each phase provides a subset of methods that are available to the struct, allowing
+/// for a more controlled configuration and execution process.
+///
+/// During the `NoAlloc` phase, the struct provides methods to configure the dxe core's cpu related
+/// functionality, the section extractor. During this time, no allocations are available.
```
makubacki:
```suggestion
/// During the `NoAlloc` phase, the struct provides methods to configure the DXE core environment
/// prior to allocation capability such as CPU functionality and section extraction. During this time,
/// no allocations are available.
```

makubacki: Minor tweak to loosen what may in the pre-alloc code over time.

dxe_core/src/lib.rs
```diff
@@ -146,14 +146,27 @@ macro_rules! error {
...
 ///
-/// The return type from [initialize](Core::initialize) is a [CorePostInit] object, which signals the completion of
-/// the first phase of the DXE Core and that allocations are available. See [CorePostInit] for more information.
+/// Once the [initialize](Core::initialize) method is called, the struct transitions to the `Alloc` phase,
+/// which provides methods for adding configuration and components with the dxe core, and eventually starting the
```
makubacki: Nit for consistency.

```suggestion
/// which provides methods for adding configuration and components with the DXE core, and eventually starting the
```

dxe_core/src/lib.rs
```diff
@@ -146,14 +146,28 @@ macro_rules! error {
 
 pub(crate) static GCD: SpinLockedGcd = SpinLockedGcd::new(Some(events::gcd_map_change));
 
+#[doc(hidden)]
+/// A zero-sized type to gate allocation functions in the [Core].
+pub struct Alloc;
```
cfernald, line 151: Should these be enums to make them uninhabitable? i.e.

```
pub enum Alloc {}
pub enum NoAlloc {}
```

Javagedes: I don't think it matters too much. Since the attributes of the `Core` are private, and we do not provide a creator function like `new`, the only way to generate this object is by using `::default()` which can only create the `NoAlloc` version.

If you don't think hiding the types in the documentation, and want to ensure that no one can use these types whatsoever, even by accident, I'd be more than happy to update them to enums :)

PR #231, opened 2025-01-23
Prevent line ending normalization

Javagedes:
## Description

Unsets the text attribute on all files to prevent their line endings
from being normalized.

See https://git-scm.com/docs/gitattributes for background.

This was previously accomplished solely by instructing users to set
core.autocrlf to false. However, that does not translate to pipelines
and setting in a pipeline requires the system level git configuration
to be modified prior to checkout. Git configs cannot be checked into
to a repo and automatically used for security reasons.

This is a simple way to check the change into the repo so it applies
for all users including pipelines to prevent line ending conversion.

Ref: https://github.com/microsoft/mu_devops/pull/333

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

N/A

## Integration Instructions

N/A


Javagedes:
@makubacki does this logic make sense? The line endings were messed up without touching this configuration. I added the configuration after the checkout, and they were still messed up. I think having this setting made before checkout should fix it. Do you agree?

makubacki:
> @makubacki does this logic make sense? The line endings were messed up without touching this configuration. I added the configuration after the checkout, and they were still messed up. I think having this setting made before checkout should fix it. Do you agree?

To simplify things locally vs agents, I did this for other repos: https://github.com/microsoft/mu_devops/pull/333

Have you tried that?

Javagedes:
> > @makubacki does this logic make sense? The line endings were messed up without touching this configuration. I added the configuration after the checkout, and they were still messed up. I think having this setting made before checkout should fix it. Do you agree?
> 
> To simplify things locally vs agents, I did this for other repos: [microsoft/mu_devops#333](https://github.com/microsoft/mu_devops/pull/333)
> 
> Have you tried that?

This looks good. I'll do this instead.

PR #232, opened 2025-01-24
Fix bug with `data_size` in `PerformanceRecord` types

magravel:
Remove the data_size from PerformanceRecord trait and use the number of bytes written by `scroll::gwrite` as the size. This was causing bugs due to `mem::size_of_val(self)` including padding. 

(Further explanation: we do not want any padding to be written to FPDT, so it does not make sense to count these bytes in the `data_size`. The equivalent representation in C, and the spec-defined behavior of the FPDT, assumes that fields are packed. Thus, invariably, the size should always be equivalent to the size of all the individual fields written.)

## Description

- [X] Impacts functionality? (fixes bug in FPDT)
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested
Unit tests pass. I also generated an FPDT with this, and we are able to read valid entries from the written FPDT.

## Integration Instructions

N/A


Javagedes:
nit: please reduce the length of the commit title and put more detailed information from the title in the commit summary (PR title / PR summary)

PR #234, opened 2025-01-24
[REBASE&FF] Update to new component interface

Javagedes:
## Description

This pull request updates the component interface from that defined in `uefi_component_interface` to that in `uefi_sdk::component` and deletes `uefi_component_interface`. This pull request makes the necessary changes to `uefi_test` and `adv_logger` to switch them to the new interface. This commit also provides documentation changes related to the new interface.

- [x] Impacts functionality?
- [ ] Impacts security?
- [x] Breaking change?
- [ ] Includes tests?
- [x] Includes documentation?

## How This Was Tested

Confirmed qemu_Q35 continues to boot to shell

### Component Developers

Any driver writers previously using the old component interface by implementing the `Component` trait on a struct (` fn entry_point(&self, interface: &dyn DxeComponentInterface) -> error::Result<()>`) have many more options and should review the documentation in `docs` and in `uefi_sdk`.

The easiest transition will be to switch from `fn entry_point(&self, interface: &dyn DxeComponentInterface) -> error::Result<()>` to 

`fn entry_point(bs: uefi_sdk::boot_services::StandardBootServices) -> uefi_sdk::error::Result<()> { ... }`, however there may be better options available. Any private configuration that was inside the component struct must now be registered with the core using the `.with_config` method. As an example, if your struct was previously `MyComponent { size: usize }`, you should now switch to `struct Size(usize);` and add it via `.with_config(Size(32))`. Components supporting private configuration will be added in the future.

### Platform Developers

Platform developers must update their platform file by switching from `.initialize(physical_hob_list)` to `.init_memory(physical_hob_list)` and remove drivers using the old interface (or update them if they have already integrated the changes) as this is not a backwards compatible change. Components provided via `uefi-dxe-core` including `adv_logger`, `uefi_test`, and `sample_components` have all been updated accordingly, so please read their documentation on how to properly update.


adv_logger/src/component.rs
```diff
@@ -62,155 +63,98 @@ where
...
+        None => {
+            log::error!("Advanced Logger not initialized before component entry!");
+            return Err(EfiError::NotStarted);
+        }
+    };
 
-        let res = component.init_advanced_logger(hob_list);
-        assert_eq!(res, Ok(()));
+    let address = log_info as *const AdvLoggerInfo as efi::PhysicalAddress;
+    let protocol = AdvancedLoggerProtocol::new(adv_log_write::<S>, address, *logger);
```
joschock: why do you need both protocol and protocol2?

adv_logger/src/lib.rs
```diff
@@ -51,9 +49,13 @@
 //! SPDX-License-Identifier: BSD-2-Clause-Patent
 //!
 #![no_std]
+#![feature(let_chains)]
```
joschock: Might want to avoid unstable if we can....

adv_logger/src/logger.rs
```diff
@@ -76,6 +80,22 @@ where
     }
 }
 
+// This should never be used. This is to satisfy the `Default` trait requirement for the `Config` parameter type.
+impl<'a, S> Default for &'static AdvancedLogger<'a, S>
+where
+    S: SerialIO + Send,
+{
+    fn default() -> Self {
+        Box::leak(Box::new(AdvancedLogger {
```
joschock: if this is really never used, consider panic?

docs/src/driver/interface.md
```diff
@@ -1,65 +1,69 @@
...
-This chapter will walk you through writing UEFI components in rust that will be compiled into the
-pure rust DXE Core image.
+In rust, a component is simply a trait implementation. So long as a struct implements [Component](todo/docs.rs) and
+[IntoComponent](todo/docs.rs), it can be consumed and executed by the pure rust DXE core. [uefi_sdk](todo/docs.rs)
+currently provides one implementation for `Component`, which is [FunctionComponent](todo/docs.rs). This type cannot be
+instantiated manually, but a blanket implementation of the `IntoComponent` trait allows for any function whose's
+parameters support dependency injection to be converted into a `FunctionComponent`.
 
-## NOTE1
+Due to this, developing a component is as simple as writing a function whose parameters are apart of the below list of
```
joschock:
```suggestion
Due to this, developing a component is as simple as writing a function whose parameters are a part of the below list of

```

docs/src/driver/interface.md
```diff
@@ -1,65 +1,69 @@
...
+//! | Option\<P\>                  | An Option, where P implements `Param`. Allows components to run even when the underlying parameter is unavailable. See the [params] module for more info.  |
+//! | (P1, P2, ...)                | A Tuple where each entry implements `Param`. Useful when you need more parameters than the current parameter limit. See the [params] module for more info. |
+//! | Config\<T\>                  | An immutable config value that will only be available once the underlying data has been locked. See The [params] module for more info.                     |
+//! | ConfigMut\<T\>               | A mutable config value that will only be available while the underlying data is unlocked. See the [params] module for more info.                           |
+//! | &HobList                     | An immutable reference to a list of Handoff Blocks (HOBs) passed to the DXE Core.                                                                          |
+//! | StandardBootServices         | Rust implementation of Boot Services                                                                                                                       |
```
joschock: Is StandardBootServices a concrete type?

docs/src/dxe_core/component_model.md
```diff
@@ -0,0 +1,181 @@
...
+
+`Config<T>` and `ConfigMut<T>` are special, as they both reference the same underlying data. As you might expect
+`ConfigMut<T>` allows you to mutate configuration. To prevent drivers that rely on `Config<T>` from running with `T`
+that has not been finalized, The below rules have been implemented for the interaction between the two.
+
+1. The underlying value `T` has the concept of being "locked" or "unlocked"
+1. `Config<T>` is only "retrievable" when the underlying value `T` is "locked"
+1. `ConfigMut<T>` is only "retrievable" when the underlying value `T` is "unlocked"
+1. All configuration values are "locked" by default, and cannot be "unlocked" by anyone or anything
+1. Once the underlying value `T` has been "locked", it cannot be "locked".
```
joschock:
```suggestion
1. Once the underlying value `T` has been "locked", it cannot be "unlocked".
```

docs/src/dxe_core/component_model.md
```diff
@@ -0,0 +1,181 @@
...
+[Executing Components](#executing-components) section.
+
+### Registering Components
+
+Registering a component is a fairly simple process. The component is turned into a `Box<dyn Component>`
+(using `into_component()`) so that multiple implementations of `Component` can be supported. The cost of this
+abstraction and v-table lookup is fairly minimal due to the low number of components as compared to other parts of the
+DXE Core. From there, the `Component::initialize` method is called. This is different for each component
+implementation, however generically, this method is used to initialize, register, and validate all `Param` usage that
+the specific component uses. This registration.
```
joschock: incomplete sentence

docs/src/dxe_core/component_model.md
```diff
@@ -0,0 +1,181 @@
...
+
+For initialization, it is used to initialize the param and store any immutable state. As an example the `Config<T>` and
```
joschock: it's not clear what `it` refers to in this sentence

docs/src/dxe_core/component_model.md
```diff
@@ -0,0 +1,181 @@
...
+`ConfigMut<T>` params use this to request and store the global registration id of the underlying `T` value so that when
+attempting to fetch the value, it does not have to first query the global id before requesting the type using the id.
+Each param will be different.
+
+For registration, it is used to register any data accesses the parameter has with the component's metadata. The
+scheduler can use the access set of each component for scheduling parallel execution of components, though this is not
+yet implemented.
+
+For validation, it is used to validate no params in the same component conflict with eachother. This is done via the
+access set mentioned above. As an example, a component that uses `ConfigMut<u32>` and `Config<u32>` is invalid. You
```
joschock:
```suggestion
access set mentioned above. As an example, a component that uses `ConfigMut<T>` and `Config<T>` is invalid. You
```

docs/src/dxe_core/component_model.md
```diff
@@ -0,0 +1,181 @@
...
+cannot have mutable and immutable access to the same data.
+
+Once the component and it's params are fully registered, the component is simply added to a Vec of other components
```
joschock:
```suggestion
Once the component and its params are fully registered, the component is simply added to a Vec of other components
```
Also - the added to a vec bit seems like an implementation detail

docs/src/dxe_core/component_model.md
```diff
@@ -0,0 +1,181 @@
...
+for future execution.
+
+### Executing Components
+
+Executing a component is also a fairly simple process. There is an outer process handled by the DXE Core (controllowing
```
joschock:
```suggestion
Executing a component is also a fairly simple process. There is an outer process handled by the DXE Core (controlling
```

docs/src/dxe_core/component_model.md
```diff
@@ -0,0 +1,181 @@
...
+_when_ to execute a component), and an inner process handled by the `Component` trait implementation on _how_ to
+execute a component.
+
+We will first talk about the logic behind _when_ to execute a component. This processes is straight forward. The
+dispatcher attempts to run all components as is. It does this in a similar fashion as the EDKII dispatcher. That is
```
joschock:
```suggestion
dispatcher attempts to run all components as-is. It does this in a similar fashion as the EDKII dispatcher. That is
```

docs/src/dxe_core/component_model.md
```diff
@@ -0,0 +1,181 @@
...
+to say it will loop through all components, attempting to execute them, either no components are left, or no change has
```
joschock:
```suggestion
to say it will loop through all components, attempting to execute them until either no components are left, or no change has
```

docs/src/dxe_core/component_model.md
```diff
@@ -0,0 +1,181 @@
...
+    D -- Yes --> A
+    D -- No --> E{Env Changed?}
+    E -- Yes --> A
+    E -- No --> Exit
+```
+
+Next, we will talk about the logic behind _how_ to execute a component. The process for executing a component will vary
+between component implementations, however generically the processes is to validate that each component's params are
+available. If all are available, we then execute the component. If all params are not available, we exit with a return
+value indicating as such. it is important to note that params self validate, so it is important that each
```
joschock:
```suggestion
value indicating as such. it is important to note that params self-validate, so it is important that each
```

docs/src/dxe_core/component_model.md
```diff
@@ -0,0 +1,181 @@
...
+    E -- Yes --> B;
+    E -- No --> Y[Run Component];
+    C -- No --> Z[Exit];
+```
+
+## Quick Reference
+
+### Traits
+
+- `Component`: The trait every type of component must implement.
```
joschock:
```suggestion
- `Component`: The trait every component must implement.
```

docs/src/dxe_core/component_model.md
```diff
@@ -0,0 +1,181 @@
...
+
+### Default Param implementations
+
+See [Component Interface](../driver/interface.md) for the list of default `Param` implementations.
+
+### Pro/Con
+
+- **Non-breaking interface**: `Param` implementations can be added at anytime without directly breaking existing
+interfaces. This means extending functionality and adding additional abstractions and functionality can be done at
+anytime without breaking existing drivers. Additionally, new implementations of `Component` can be added that could be
```
joschock:
```suggestion
any time without breaking existing drivers. Additionally, new implementations of `Component` can be added that could be
```

docs/src/dxe_core/component_model.md
```diff
@@ -0,0 +1,181 @@
...
+### Default Component Implementations
+
+- `FunctionComponent`: A function whose parameters all implement `Param`. A blanket `IntoComponent` implementation
+  allows for converting this type of function into a `FunctionComponent`.
+
+### Default Param implementations
+
+See [Component Interface](../driver/interface.md) for the list of default `Param` implementations.
+
+### Pro/Con
```
joschock: Seems like all Pros ;)

```suggestion
### Advantages
```

docs/src/dxe_core/component_model.md
```diff
@@ -0,0 +1,181 @@
...
+
+- **Non-breaking interface**: `Param` implementations can be added at anytime without directly breaking existing
+interfaces. This means extending functionality and adding additional abstractions and functionality can be done at
+anytime without breaking existing drivers. Additionally, new implementations of `Component` can be added that could be
+completely different that existing implementations such as `FunctionComponent`. They may not even use dependency
+injection!
+
+- **Configuration**: A single location exists for all configuration for the platform. This prevents configuration from
+being duplicated across multiple drivers as with Static PCDs. This also allows for configuration to be of any rust
+supported type, unlike the limited configuration of EDKII. Finally, any component is able to hook at configuration
```
joschock:
```suggestion
supported type, unlike the limited configuration of EDKII. Finally, any component is able to hook any configuration
```

docs/src/dxe_core/component_model.md
joschock: If I understand correctly, you wouldn't want to use `Config<u32>`. You'd want to do something like 
```rust
struct MyConfig(u32)
```
and use `Config<MyConfig>`, right? If not, under what circumstances would `Config<u32>` make sense?

dxe_core/src/lib.rs
```diff
@@ -430,24 +427,83 @@ where
...
+                        true
+                    }
+                    Ok(false) => {
+                        log::info!("Dispatching Component: {:?} ... Skipped", component.metadata().name());
+                        false
+                    }
+                    Err(err) => {
+                        log::error!("Dispatching Component: {:?} ... Failed: {:?}", component.metadata().name(), err);
+                        debug_assert!(false);
+                        true // Component dispatched, even if it did fail.
```
joschock:
```suggestion
                        true // Component dispatched, even if it did fail, so remove from self.components to avoid re-dispatch.

```

adv_logger/src/logger.rs
Javagedes: Good Idea. Thanks!

docs/src/driver/interface.md
Javagedes: Yes, it is a concrete type. Unfortunately, the `BootServices` trait from uefi_sdk is not object-safe due to the generic parameters, so it cannot be converted into a `&dyn BootServices` or `Box<dyn BootServices>`. Due to this, I implemented `Param` on [StandardBootServices](https://github.com/pop-project/uefi-sdk/blob/9a871ca998e639165416dbe3c1d7fe016ac22da6/boot_services/src/boot_services.rs#L42).

Also, it is a wrapper the boot services pointer, so we create a temporary instance of it for each component that is run and uses it. That is why we can use `StandardBootServices` instead of `&StandardBootServices`

docs/src/dxe_core/component_model.md
Javagedes: Yes, you are exactly correct. I would say `Config<u32>` will never make sense. I just used that as an example for simplicities sake, but I'd be happy to Update.

In a future release, I will be adding another Component implementation that supports private configuration, so the use case of `Config<T>` and `ConfigMut<T>` will be primarily for generic configuration that would cascade to multiple components.

I will update to use something more appropriate than a base type like `u32`

docs/src/dxe_core/component_model.md
Javagedes: I need to check the cspell step in our CI workflow, this should have been caught...

adv_logger/src/component.rs
Javagedes: Due to the interface of [install_protocol_interface](https://github.com/pop-project/uefi-sdk/blob/9a871ca998e639165416dbe3c1d7fe016ac22da6/boot_services/src/boot_services.rs#L269) being slightly weird to support the ability to install a protocol with a nullptr as an interface.

@magravel is there a better way to handle this in this use case? Any suggestions would be appreciated :)

adv_logger/src/lib.rs
Javagedes: sure no problem. It's only used in one location and simple to remove :)

docs/src/driver/interface.md
Javagedes: Fixed in upcoming rebase. Thanks!

docs/src/dxe_core/component_model.md
Javagedes: Fixed in upcoming rebase. Thanks!

docs/src/dxe_core/component_model.md
Javagedes: Removed `This registration`. in an upcoming rebase. Good catch. Thanks!

docs/src/dxe_core/component_model.md
Javagedes: In an upcoming rebase, I added clarity to these three parts (initialization, registration, validation) and remove the unclear "it"

docs/src/dxe_core/component_model.md
Javagedes: Updated to `T` as per your suggestion

docs/src/dxe_core/component_model.md
Javagedes: Updated. Thanks!

docs/src/dxe_core/component_model.md
Javagedes: Updated, thank you!

docs/src/dxe_core/component_model.md
Javagedes: Updated. Thanks!

docs/src/dxe_core/component_model.md
Javagedes: Updated!

docs/src/dxe_core/component_model.md
Javagedes: Removed the specific container type, just mentioning that it will be stored for dispatch. Thanks!

docs/src/dxe_core/component_model.md
Javagedes: Updated! Thank you.

docs/src/dxe_core/component_model.md
Javagedes: Updated. Thank you!

docs/src/dxe_core/component_model.md
Javagedes: Updated! Thank you :)

docs/src/dxe_core/component_model.md
Javagedes: Updated. Thanks :) I really suck at writing!

dxe_core/src/lib.rs
Javagedes: Added. Thanks!

adv_logger/src/logger.rs
Javagedes: Added a panic stating to use the `new` constructor. Thanks again.

adv_logger/src/lib.rs
Javagedes: Removed the unstable feature and the one use-case in the code!

makubacki:
Please leave up at least until Monday. Came up later in my day and I'd like to look through it.

Javagedes:
> Please leave up at least until Monday. Came up later in my day and I'd like to look through it.

I'm in no rush to get this in and hope to get good feedback from multiple parties, so no rush on a review. Take your time :)

docs/src/dxe_core/component_model.md
```diff
@@ -0,0 +1,180 @@
...
+example the `Config<T>` and `ConfigMut<T>` params use this to request and store the global registration id of the
+underlying `T` value so that when attempting to fetch the value, it does not have to first query the global id before
+requesting the type using the id.
+
+In terms of registration, `Component::initialize` is used to register any data accesses the parameter has with the
+component's metadata. The scheduler can use the access set of each component for scheduling parallel execution of
+components, though this is not yet implemented.
+
+In terms of validation, `Component::initialize` is used to validate no params in the same component conflict with
+eachother. This is done via the access set mentioned above. As an example, a component that uses `ConfigMut<T>` and
```
os-d:
```suggestion
each other. This is done via the access set mentioned above. As an example, a component that uses `ConfigMut<T>` and
```

docs/src/dxe_core/component_model.md
```diff
@@ -0,0 +1,180 @@
...
+`Config<T>` is invalid. You cannot have mutable and immutable access to the same data.
+
+Once the component and it's params are fully registered, the component is stored for dispatch.
```
os-d:
```suggestion
Once the component and its params are fully registered, the component is stored for dispatch.
```

docs/src/dxe_core/component_model.md
```diff
@@ -0,0 +1,180 @@
...
+Executing a component is also a fairly simple process. There is an outer process handled by the DXE Core (controlling
+_when_ to execute a component), and an inner process handled by the `Component` trait implementation on _how_ to
+execute a component.
+
+We will first talk about the logic behind _when_ to execute a component. This processes is straight forward. The
+dispatcher attempts to run all components as-is. It does this in a similar fashion as the EDKII dispatcher. That is
+to say it will loop through all components, attempting to execute them until either no components are left, or no
+change has happened since the previous loop. Where as EDKII would normally exit dispatch, the component dispatcher
+instead will lock all configuration values and restart the process. This is important because it allows any component
+that relies on `Config<T>` to execute, even if a driver that needed `ConfigMut<T>` never locked the underlying value.
```
os-d, line 101: How does this work in a callback case? e.g. a component has ConfigMut access on value X. It executes and registers a callback for event Y and exits. The dispatcher now locks X and dispatches component 2, which has Config access on value X and creates event Y (or a timer tick, whatever), when the first component comes up in the callback, it will not have mutable access to X any longer, right?

Is this just considered a misconfiguration and you need to order your dependencies better? Or do we have some guarantees here?

Javagedes: An eventing system does not currently exist for monolithically compiled components. I should clarify that you can still use the eventing system provided by the UEFI spec through `StandardBootServices` param, but whatever event callback you register is going to be EFI API method, not a dependency injectable component.

I think the design points and questions you are asking will be answered when we have the use case and want to create an observer / event callback system for monolithically compiled systems. A light "think" about it and I would lean towards any callbacks would not be allowed to use `ConfigMut` (enforceable, not just "spec"), but that would be up for discussion.

`Config` / `ConfigMut` is a different concept than internal state in terms of components. So if anything, a callback should be changing internal state of the component, not a configuration value public to the entire system. Internal state does not exist yet, but I have a working example of `StructComponents` that I need to clean up, and that type of component will be allowed to have internal, private, state.

The `Config` and `ConfigMut` logic is a rusty alternative to PCDs that allow for platform level, public, configuration of drivers and functionality. The execution logic regarding `Config` and `ConfigMut` is purely to ensure that no driver is accidently executed while a particular Config is not in its final state

i.e.

Component A and Component B want to update Config<i32> in some manner while Component C does not change anything, but uses Config<i32> for execution.

Component C will run differently depending on order [`<insert A,B,C permutations>`]. So this logic ensures that C will always run last and any other components that "just use `Config<i32>` for execution" all execute with the same value.

docs/src/dxe_core/component_model.md
Javagedes: Fixed in upcoming rebase. Thanks!

docs/src/dxe_core/component_model.md
Javagedes: Fixed in upcoming rebase. Thanks!

adv_logger/src/component.rs
```diff
@@ -62,155 +63,98 @@ where
...
-    ) -> efi::Status {
-        // SAFETY: We have no choice but to trust the caller on the buffer size. convert
-        //         to a reference for internal safety.
-        let data = unsafe { core::slice::from_raw_parts(buffer, num_bytes) };
-        let error_level = error_level as u32;
-
-        // SAFETY: We must trust the C code was a responsible steward of this buffer.
-        unsafe { (*this).adv_logger }.log_write(error_level, data);
-        efi::Status::SUCCESS
+    fn protocol_guid(&self) -> &'static efi::Guid {
```
cfernald: Not to this PR obviously, but it seems this shouldn't require a `&self` and dropping it would make things like the install_protocol not require the double reference as it could operate on the static `P` instead of an instance.

adv_logger/src/logger.rs
```diff
@@ -21,7 +21,7 @@ pub struct AdvancedLogger<'a, S>
 where
     S: SerialIO + Send,
 {
-    hardware_port: S,
+    hardware_port: Option<S>,
```
cfernald: Does not having this cause a compile error? It seems that with the default panic there should be no scenario where it's not set, right?

dxe_core/src/lib.rs
```diff
@@ -430,24 +427,83 @@ where
...
+    fn dispatch_components(&mut self) {
+        loop {
+            let len = self.components.len();
+            self.components.retain_mut(|component| {
+                // Ok(true): Dispatchable and dispatched returning success
+                // Ok(false): Not dispatchable at this time.
+                // Err(e): Dispatchable and dispatched returning failure
+                !match component.run(&mut self.storage) {
+                    Ok(true) => {
+                        log::info!("Dispatching Component: {:?} ... Success", component.metadata().name());
```
cfernald: Shouldn't we print `Dispatching Component` before calling `run()`? or perhaps change the wording to be clear this gets printed after execution.

adv_logger/src/component.rs
```diff
@@ -62,155 +63,98 @@ where
...
+    fn deref(&self) -> &Self::Target {
+        &Self::GUID
+    }
+}
 
-        // deliberate leak
-        let interface = Box::into_raw(Box::new(protocol));
-        let interface = interface as *mut c_void;
+/// EFI API to write to the advanced logger through the advanced logger protocol.
+extern "efiapi" fn adv_log_write<S: SerialIO + Send + 'static>(
```
makubacki: It seems more of these should be marked as unsafe as reminders to callers about safety assumptions. Don't need to change anything here.

adv_logger/src/component.rs
makubacki: Can you give a bit more info? Does [`install_protocol_marker()`](https://github.com/OpenDevicePartnership/uefi-sdk/blob/9a871ca998e639165416dbe3c1d7fe016ac22da6/boot_services/src/boot_services.rs#L301) help?

docs/src/driver/interface.md
```diff
@@ -1,65 +1,69 @@
-# Component Interface
+# Monolithically Compiled Components
 
-**TODO** This section is work-in-progress.
+There is no standard entry point for a monolithically compiled rust component that is dispatched by the rust DXE core.
+Where EDKII Dxe Core dispatcher expects a well defined entry point of
```
makubacki: I think the first statement is important enough to let that get attention in its own paragraph and then continue.

```suggestion
There is no standard entry point for a monolithically compiled rust component that is dispatched by the rust DXE core.

While the EDKII Dxe Core dispatcher expects a well defined entry point of
```

docs/src/driver/interface.md
```diff
@@ -1,65 +1,69 @@
...
+`EFI_HANDLE ImageHandle, EFI_SYSTEM_TABLE *SystemTable`, the Pure rust DXE Core uses [Dependency Injection](https://wikipedia.org/wiki/Dependency_injection)
+to allow a component to define an interface that specifies all dependencies needed to properly execute. Due to this,
+dependency expressions are no longer necessary, as the function interface is the dependency expression. What this
+means is that instead of evaluating a dependency expression to determine if a driver can be executed, it instead
+attempts to fetch all requested parameters defined in the function interface. If all are successfully fetched, then the
+component is executed, if not, it will not be dispatched, and another attempt will be made in the next iteration.
 
-This chapter will walk you through writing UEFI components in rust that will be compiled into the
-pure rust DXE Core image.
+In rust, a component is simply a trait implementation. So long as a struct implements [Component](todo/docs.rs) and
```
makubacki:
```suggestion
In the Rust DXE Core, a component is simply a trait implementation. So long as a struct implements [Component](todo/docs.rs) and
```

docs/src/driver/interface.md
```diff
@@ -1,65 +1,69 @@
...
+[IntoComponent](todo/docs.rs), it can be consumed and executed by the pure rust DXE core. [uefi_sdk](todo/docs.rs)
+currently provides one implementation for `Component`, which is [FunctionComponent](todo/docs.rs). This type cannot be
+instantiated manually, but a blanket implementation of the `IntoComponent` trait allows for any function whose's
+parameters support dependency injection to be converted into a `FunctionComponent`.
```
makubacki:
```suggestion
currently provides one implementation for `Component`, which is [FunctionComponent](todo/docs.rs). This type cannot be
instantiated manually, but a blanket implementation of the `IntoComponent` trait allows any function whose 
parameters support dependency injection to be converted into a `FunctionComponent`.
```

docs/src/dxe_core/component_model.md
```diff
@@ -0,0 +1,180 @@
...
+`Param` types does not become a breaking change. This is because the [Component](todo/docs.rs) trait interface consumes
+the `Storage` object instead each piece of data individually. When adding a new `Param` type, instead of updating the
+`Component` trait interface with this new data, we instead update the `Storage` object, adding the new data to that
+instead. This causes the "breaking change" to be internal to the DXE Core, instead of to every driver written.
+
+Due to this, however, `Storage` is the bridge between `Component`s and the rest of the DXE Core. During DXE Core
+initialization, `Storage` must also be updated to hold certain important pieces of data. As an example, once the boot
+services table has been fully initialized, it must also be registered with `Storage`, otherwise no component that
+wishes to use the boot services table will ever be executed. Another example is that `Storage` directly contains the
+[HobList] object, and allows for components to have am immutable reference to this. This particular part of the model
```
makubacki, line 28: At some point, we might want to consider abstracting components from the HOB list. I'd like to get components to the point they have as little coupling as possible to the PI Specification. These can become our biggest obstacle to breaking away from the PI Spec since they are authored outside our control. I don't think we need to act right away.

docs/src/dxe_core/component_model.md
```diff
@@ -0,0 +1,180 @@
...
+`Config<T>` and `ConfigMut<T>` are special, as they both reference the same underlying data. As you might expect
+`ConfigMut<T>` allows you to mutate configuration. To prevent drivers that rely on `Config<T>` from running with `T`
+that has not been finalized, The below rules have been implemented for the interaction between the two.
+
+1. The underlying value `T` has the concept of being "locked" or "unlocked"
+1. `Config<T>` is only "retrievable" when the underlying value `T` is "locked"
+1. `ConfigMut<T>` is only "retrievable" when the underlying value `T` is "unlocked"
+1. All configuration values are "locked" by default, and cannot be "unlocked" by anyone or anything
+1. Once the underlying value `T` has been "locked", it cannot be "unlocked".
+1. Configuration is automatically "unlocked" prior to component execution if a registered component is detected to
```
makubacki: This item:

```
Configuration is automatically "unlocked" prior to component execution if a registered component is detected to
use `ConfigMut<T>`
```

Seems somewhat contradictory to:

```
All configuration values are "locked" by default, and cannot be "unlocked" by anyone or anything
```

Can you account for that behavior in each description a little more precisely?

docs/src/dxe_core/component_model.md
```diff
@@ -0,0 +1,180 @@
...
+
+Registering a component is a fairly simple process. The component is turned into a `Box<dyn Component>`
+(using `into_component()`) so that multiple implementations of `Component` can be supported. The cost of this
+abstraction and v-table lookup is fairly minimal due to the low number of components as compared to other parts of the
+DXE Core. From there, the `Component::initialize` method is called. This is different for each component
+implementation, however generically, this method is used to initialize, register, and validate all `Param` usage that
+the specific component uses.
+
+In terms of initialization, `Component::initialize` is used to initialize the param and store any immutable state. As an
+example the `Config<T>` and `ConfigMut<T>` params use this to request and store the global registration id of the
```
makubacki:
```suggestion
example, the `Config<T>` and `ConfigMut<T>` params use this to request and store the global registration id of the
```

docs/src/dxe_core/component_model.md
```diff
@@ -0,0 +1,180 @@
...
+    B -- No --> D
+    C --> D{More Components?}
+    D -- Yes --> A
+    D -- No --> E{Env Changed?}
+    E -- Yes --> A
+    E -- No --> Exit
+```
+
+Next, we will talk about the logic behind _how_ to execute a component. The process for executing a component will vary
+between component implementations, however generically the processes is to validate that each component's params are
```
makubacki:
```suggestion
between component implementations, however generically the process is to validate that each component's params are
```

dxe_core/src/lib.rs
```diff
@@ -430,24 +427,83 @@ where
...
+                        true
+                    }
+                    Ok(false) => {
+                        log::info!("Dispatching Component: {:?} ... Skipped", component.metadata().name());
```
makubacki: Is there additional info printed about why a component was skipped?

makubacki: Didn't get to `display_components_not_dispatched()` yet. That's what I was looking for.

adv_logger/src/component.rs
Javagedes: I agree. I'm not sure the reasoning behind having this `&self`, but I do know that by removing `&self`, the object is nolonger `object-safe`, which basically means the trait cannot be used as a trait object such as `&dyn Protocl` or `Box<dyn Protocol>`.

Not sure if that would be an issue or not. @magravel or @zurcher would probably have more context as they implemented / used this before this.

adv_logger/src/logger.rs
Javagedes: This does not cause an issue, however that is a good point, I originally turned this into `Option<S>` to satisfy the default implementation. By now panicing in that per John's suggestion, I can probably change it back to normal.

dxe_core/src/lib.rs
Javagedes: I was doing this to try and make it "pretty", but you are right, this is kind of dumb. All of the logging from the component shows up before saying "Dispatching component ...", so that can definitely be confusing.

I will update this.

adv_logger/src/component.rs
Javagedes: yeah the efiapi functions are kind of a weird spot because pretty much all of them are going to be unsafe because we almost always accept pointers that we cannot validate via compiler.

docs/src/dxe_core/component_model.md
Javagedes: I can remove it right now if you like. We have no use case for it anymore. I originally had a use case for adv_logger when I was initializing adv_logger through the callback functionality that we removed. But since we removed the different callbacks, adv_logger is using the pointer phit pointer directly.

I just didnt remove it because I had already implemented it. But if you want it gone to further break these components from the UEFI spec, I'll happily do so.

makubacki: I think so. My concern is that it becomes a crutch to config that should be more formally defined to the component and increasing our dependencies on specs or aspects of those specs we might want to move further away from over time.

adv_logger/src/logger.rs
Javagedes: I was able to revert this back to `S`. Thanks! It'll show up in my next force push. Closing now though.

dxe_core/src/lib.rs
Javagedes: Updated to the following:

```rust
...
log::info!("DISPATCH_ATTEMPT BEGIN: Id = [{:?}]", component.metadata().name());
...
log::info!("DISPATCH_ATTEMPT END: Id = [{:?}] Status = [Success]", component.metadata().name());
...
log::info!("DISPATCH_ATTEMPT END: Id = [{:?}] Status [Skipped]", component.metadata().name());
...
log::error!("DISPATCH_ATTEMPT END: Id = [{:?}] Status = [Failed] Error = [{:?}]", component.metadata().name(), err);
```

If you have any nit like suggestions for the log message, please reopen :) Thanks for the suggestion, this was a needed change.

docs/src/driver/interface.md
Javagedes: Updated to add a newline as requested! Will see it in the next force push.

docs/src/driver/interface.md
Javagedes: Updated. Thanks! Will show up in the next push.

docs/src/dxe_core/component_model.md
Javagedes: Sounds good. Removed the documentation regarding it in uefi-dxe-core. I will put up a PR in uefi-sdk to actually remove it (And the logic in dxe_core/src/lib.rs that adds the hob list to `Storage`.

docs/src/dxe_core/component_model.md
Javagedes: Updated. Thanks

docs/src/dxe_core/component_model.md
Javagedes: Updated. Thanks!

docs/src/dxe_core/component_model.md
Javagedes: Sure thing. 

Updated to the below. It reads more clear to me:

``` markdown
1. All configuration values are "locked" by default.
1. Only the dispatcher is able to "unlock" a component, which occurs prior to any component execution, if a registered
   component is detected to use `ConfigMut<T>`.
```

docs/src/driver/interface.md
Javagedes: Updated. Thanks!

adv_logger/src/component.rs
Javagedes: We do not want to install a null interface, I was just saying that the function interface for install_protocol_interface separates the Protocol object from the protocol instance object so that the interface can be null if we want it to.

The main issue is I implemented Protocol on the protocol interface object, so to pass both, I had to instantiate it twice. I've since created a new object and implemented Protocol on that instead. Now we only implement the protocol_interface once, and a zero sized struct for the Protocol.

adv_logger/src/component.rs
Javagedes: I talked to Mathieu, we are going to look at improving this interface. Like you said, not for this PR, so I'm going to close this comment :)

PR #236, opened 2025-01-28
Add basic instrumentation and fix minor bugs in perf code

berlin-with0ut-return:
## Description
- Adds instrumentation to measure new performance types
- Fix memory layout issues by adding #repr(C) to FPDT records 
- Fix string issues with new FFI string conversion implementation

- [X]  Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested
I generated a FPDT using this code and it gives expected results with a few caveats (also mentioned in the TODOs). Note that these results were generated with a previous build of uefi-dxe-core, before the perf changes. 
Caveats:
- `get_module_info` still needs to be debugged and fully implemented, so many of the entries have a 0 guid/no module name as of now
- `init_dispatcher` should be instrumented, but doesn't work with the current `BOOT_SERVICES` initialization order
- we may need to instrument other repos like `uefi_sdk`. further discussion on what data we actually want is probably warranted.
- we need to take cross-module measurements from the C code to the Rust code (PEI->DXE and DXE->BDS) 
- I have put all perf measurements behind a guard. This is to avoid any errors caused by `init_performance_lib` currently being commented out (all instrumentation will panic if boot services for perf is not initialized).

## Integration Instructions
N/A.

### DIFF
--- a/crates/uefi_performance/src/lib.rs
+++ b/crates/uefi_performance/src/lib.rs
@@ -740,7 +740,6 @@ pub fn _perf_callback_end(trigger_guid: &efi::Guid, fun_name: &str, caller_id: &
 macro_rules! perf_function_begin {
     ($caller_id:expr) => {
         if cfg!(feature = "instrument_performance") {
-            log::info!("dkddkdk");
             $crate::_perf_function_begin($crate::function!(), $caller_id)
         }
     };


crates/uefi_performance/src/performance_record/extended.rs
```diff
@@ -73,6 +74,7 @@ impl Debug for GuidEventRecord {
     }
 }
 
+#[repr(C)]
```
magravel, line 77: `#[repr(C)]` is probably unnecessary since we use pwrite to write the struct in memory.

crates/uefi_performance/src/lib.rs
```diff
@@ -753,3 +716,7 @@ pub fn perf_start_ex(
 pub fn perf_end_ex(handle: efi::Handle, token: *const c_char, module: *const c_char, timestamp: u64, identifier: u32) {
     end_perf_measurement(handle, token, module, timestamp, identifier);
 }
+
+// TODOs: cross-module measurements
```
Javagedes: misc TODOs probably need removed

dxe_core/src/dispatcher.rs
```diff
@@ -22,6 +22,11 @@ use tpl_lock::TplMutex;
 use uefi_depex::{AssociatedDependency, Depex, Opcode};
 use uefi_device_path::concat_device_path_to_boxed_slice;
 
+#[cfg(feature = "instrument_performance")]
```
Javagedes: I think the better way to handle this macro, is what the log crate does. Basically, your macros will always exist, but if the instrument_performance feature is turned off, the macro compiles to nothing. It makes this much more usable. 

https://docs.rs/log/latest/src/log/macros.rs.html#35

https://docs.rs/log/latest/src/log/lib.rs.html#1534

Javagedes: I know this macro existed before this PR, but I think its worth changing in this PR to improve usability.

crates/uefi_performance/src/performance_record/extended.rs
```diff
@@ -216,6 +219,7 @@ impl Debug for DualGuidStringEventRecord<'_> {
...
```
Javagedes, line 222: Maybe be worth throwing on `Debug` derives. if you think you'll ever want to debug the contents

crates/uefi_performance/src/lib.rs
```diff
@@ -149,6 +151,10 @@ extern "efiapi" fn create_performance_measurement(
     identifier: u32,
     attribute: PerfAttribute,
 ) -> efi::Status {
+    if !PERF_ENABLED {
```
Javagedes: Sorry, I meant that we update the 7 or so macros at this location:

https://github.com/OpenDevicePartnership/uefi-dxe-core/blob/35589989a51220cd7c8df46b4dfa6ffd8d083e15/crates/uefi_performance/src/lib.rs#L477

to be similar to this:

```rust
macro_rules! perf_event_signal_begin {
    ($event_guid:expr, $caller_id:expr) => {
        if cfg!(feature = "instrument_performance") {
            $crate::_perf_event_signal_begin($event_guid, $crate::function!(), $caller_id)
        }
    };
}
```

This effectively causes the function call to disappear when the feature is disabled, but exist when it is enabled

Javagedes: This is what happens if we do it the above way

## Code in file
``` rust
let hello = 5;
perf_event_signal_begin!(...)
let goodbye = 6;
```

## Code after expansion if "instrument_performance" is enabled:

```rust
let hello = 5;
uefi_performance::_perf_event_signal_begin(...)
let goodbye = 6;
```

## Code after expansion if "instrument_performance" is disabled:

```rust
let hello = 5;
let goodbye = 6;
```

Javagedes: It basically does the same exact thing as your original implementation, but the user does not have to be the one that adds the `#[cfg(feature = "instrument_performance")]`

berlin-with0ut-return: gotcha. i think the other `if` condition in `create_performance_measurement` might be necessary too because some of the non-macro and external functions also call `create_performance_measurement`, so i can do a combination of both (otherwise it crashes because of not initializing perf)

crates/uefi_performance/Cargo.toml
```diff
@@ -21,3 +21,6 @@ uuid = { workspace = true }
 mu_pi = { workspace = true }
 mu_rust_helpers = { workspace = true }
 uefi_sdk = { workspace = true }
+
+[features]
+instrument_performance = []
```
Javagedes: nit: newline

dxe_core/src/image.rs
```diff
@@ -822,6 +824,8 @@ pub fn core_load_image(
     file_path: *mut efi::protocols::device_path::Protocol,
     image: Option<&[u8]>,
 ) -> Result<(efi::Handle, efi::Status), efi::Status> {
+    perf_load_image_begin(core::ptr::null_mut());
```
Javagedes: Is there any reason why these can't be macros and do the same thing in regards to configuring it out?

berlin-with0ut-return: no, they definitely can be macros. i don't remember why they aren't.

PR #237, opened 2025-01-28
rust_vs_edk2: Add lack of Traditional SMM support

makubacki:
## Description

Provide more explicit detail about the module types supported and why some are not supported.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [x] Includes documentation?

## How This Was Tested

CI

## Integration Instructions

N/A

### DIFF
--- a/docs/src/integrate/rust_vs_edk2.md
+++ b/docs/src/integrate/rust_vs_edk2.md
@@ -1,6 +1,38 @@
 # Differences from EDK II
 
 The Rust DXE core has several functional and implementation differences from the edk2 DXE core.
+
 Platforms should ensure the following specifications are met when transitioning over to the Rust DXE core:
+
+- Traditional SMM is not supported.
+
+  - This specifically means that the following SMM module types that require cooperation between the SMM and DXE
+    dispatchers are not supported:
+
+    - `EFI_FV_FILETYPE_SMM` (`0xA`)
+    - `EFI_FV_FILETYPE_SMM_CORE` (`0xD`)
+
+  - Further, combined DXE modules will not be dispatched. These include:
+
+    - `EFI_FV_FILETYPE_COMBINED_PEIM_DRIVER` (`0x8`)
+    - `EFI_FV_FILETYPE_COMBINED_SMM_DXE` (`0xC`)
+
+  - DXE drivers and Firmware volumes **will** be dispatched:
+
+    - `EFI_FV_FILETYPE_DRIVER` (`0x7`)
+    - `EFI_FV_FILETYPE_FIRMWARE_VOLUME_IMAGE` (`0xB`)
+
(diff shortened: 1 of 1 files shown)


docs/src/integrate/rust_vs_edk2.md
```diff
@@ -1,6 +1,38 @@
...
+    - `EFI_FV_FILETYPE_DRIVER` (`0x7`)
+    - `EFI_FV_FILETYPE_FIRMWARE_VOLUME_IMAGE` (`0xB`)
+
+  - Because Traditional SMM is not supported, events such as the `gEfiEventDxeDispatchGuid` that was used in the C DXE
+    Core to signal the end of a DXE dispatch round so SMM drivers with DXE dependency expressions could be reevaluated
+    will not be signaled.
+
+  - Dependency expressions such as `EFI_SECTION_SMM_DEPEX` will not be evaluated on firmware volumes.
+
+  - Reason: Traditional SMM is not supported to prevent coupling between the DXE and MM environments. The is error
```
vincent-j-zimmer: "The is error" maybe should be "This is error"

docs/src/integrate/rust_vs_edk2.md
```diff
@@ -1,6 +1,38 @@
...
+
+    - `EFI_FV_FILETYPE_COMBINED_PEIM_DRIVER` (`0x8`)
+    - `EFI_FV_FILETYPE_COMBINED_SMM_DXE` (`0xC`)
+
+  - DXE drivers and Firmware volumes **will** be dispatched:
+
+    - `EFI_FV_FILETYPE_DRIVER` (`0x7`)
+    - `EFI_FV_FILETYPE_FIRMWARE_VOLUME_IMAGE` (`0xB`)
+
+  - Because Traditional SMM is not supported, events such as the `gEfiEventDxeDispatchGuid` that was used in the C DXE
```
os-d, line 25: Is this the only event? If not, can we list the other events that we aren't supporting specifically for lack of traditional SMM support?

makubacki: This is the only event that immediately comes to mind that is dependent being driven from the DXE Core to support Traditional SMM. There are input signals from other sources like SMM Ready to Boot, End of DXE, etc. but those are from BDS. Technically, other external dependencies like the MM Access and MM Communicate protocols were produced from separate drivers so they're not relevant here.

There is some other logic around loading SMM images at fixed addresses that is shared but I don't think it is needed to mention when SMM image loading as a whole is disabled and that's also not a feature I've commonly seen used.

docs/src/integrate/rust_vs_edk2.md
makubacki: Fixed.

PR #238, opened 2025-01-30
extract_pei_performance_records improvements.

magravel:
## Description

Use data from the GuidHob instead of doing the pointer arithmetic myself. Doing the pointer arithmetic was also not correct after the HobList move but now the data will be valid.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Ran on QEMU

## Integration Instructions

N/A

### DIFF
--- a/adv_logger/src/component.rs
+++ b/adv_logger/src/component.rs
@@ -10,7 +10,7 @@
 //! SPDX-License-Identifier: BSD-2-Clause-Patent
 //!
 use alloc::boxed::Box;
-use core::{ffi::c_void, marker::PhantomData, ptr};
+use core::{ffi::c_void, ptr};
 use mu_pi::hob::{Hob, PhaseHandoffInformationTable};
 use r_efi::efi;
 use uefi_sdk::{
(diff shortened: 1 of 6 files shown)


crates/uefi_performance/src/lib.rs
```diff
@@ -122,20 +123,19 @@ fn extract_pei_performance_records(hob_list: &HobList) -> Result<(PerformanceRec
     let mut pei_load_image_count = 0;
 
     for hob in hob_list.iter() {
-        let guid_hob = match *hob {
-            Hob::GuidHob(hob, _) if hob.name == guid::EDKII_FPDT_EXTENDED_FIRMWARE_PERFORMANCE => hob,
+        match hob {
+            Hob::GuidHob(hob, data) if hob.name == guid::EDKII_FPDT_EXTENDED_FIRMWARE_PERFORMANCE => {
+                let mut offset = 0;
+                let [size_of_all_entries, load_image_count, _hob_is_full] =
```
makubacki, line 133: I understand that this comes from [`FPDT_PEI_EXT_PERF_HEADER`](https://github.com/tianocore/edk2/blob/master/MdeModulePkg/Include/Guid/ExtendedFirmwarePerformance.h#L250) which must essentially maintain binary compatibility of the structure layout across phases (PEI and DXE), but it would be nice to not have the fields defined here completely independently of how they are laid out in the struct. At a minimum, they could reuse the struct field order instead of independently redefining that when destructuring into the variables.

Could you deserialize the Rust defined struct instead? If there is a major disadvantage to that this isn't that bad either.

magravel: I think adding a struct would make sense if GuidHob was one struct and not a tuple to I could make something that implement from GuidHob into my struct, but making one to parse the data would result in the same thing but with extra step. Do you think we should include the data part inside GuidHob?

PR #240, opened 2025-01-31
Fix bug in device path for FV handles

berlin-with0ut-return:
## Description
During perf testing, we discovered a bug where the `file_path` in the image read by `load_image` was always `TYPE_END`; this caused the FPDT to be unable to fill in GUIDs for loaded images. 

This PR fixes this bug by appending a filename node following the FV node that provides a guid. It also fixes some errors in the way `core_load_image` was referencing the parent handle.

- [X] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested
This generates a valid FPDT with correct GUIDs and modules.

## Integration Instructions
This PR works in conjunction with #236, which fixes some other perf bugs and adds instrumentation to certain functions in the DXE core.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,7 +10,7 @@ members = [
 ]
 
 [workspace.package]
-version = "7.1.2"
+version = "7.1.3"
 publish = ["UefiRust"]
 repository = "https://github.com/OpenDevicePartnership/uefi-dxe-core"
 license = "BSD-2-Clause-Patent"


dxe_core/src/dispatcher.rs
```diff
@@ -346,11 +346,54 @@ fn add_fv_handles(new_handles: Vec<efi::Handle>) -> Result<(), efi::Status> {
...
+                        filename_nodes_buf.extend_from_slice(unsafe {
+                            core::slice::from_raw_parts(
+                                &filename_end_node as *const _ as *const u8,
+                                core::mem::size_of::<efi::protocols::device_path::Protocol>(),
+                            )
+                        });
+
+                        let boxed_device_path = filename_nodes_buf.into_boxed_slice();
+                        let filename_device_path =
+                            Box::leak(boxed_device_path) as *mut _ as *const efi::protocols::device_path::Protocol;
```
joschock: This will leak the box memory every time it's called without freeing it, I think. Since `concat_device_path_to_boxed_slice` will make a copy of the data, this box doesn't need to live forever. I suggest you do:
```suggestion
                        let filename_device_path = boxed_device_path.as_ptr() as *const efi::protocols::device_path::Protocol;

```
This will keep the box from being leaked, and will drop the memory when it goes out of scope.

dxe_core/src/dispatcher.rs
```diff
@@ -346,11 +346,54 @@ fn add_fv_handles(new_handles: Vec<efi::Handle>) -> Result<(), efi::Status> {
...
+                        let full_device_path_for_file = full_path_bytes
+                            .map(|full_path| Box::into_raw(full_path) as *mut efi::protocols::device_path::Protocol)
+                            .unwrap_or(fv_device_path);
+
                         dispatcher.pending_drivers.push(PendingDriver {
                             file_name,
                             firmware_volume_handle: handle,
                             pe32: pe32_section,
-                            device_path: fv_device_path,
+                            device_path: full_device_path_for_file, // copy + append file node
```
joschock: nit: comment is not exactly accurate since copy+append has already been done above.

dxe_core/src/image.rs
```diff
@@ -623,36 +623,38 @@ fn core_load_pe_image(
 fn get_buffer_by_file_path(
```
joschock, line 623: can you update the Return value comment?

dxe_core/src/image.rs
```diff
@@ -623,36 +623,38 @@ fn core_load_pe_image(
...
 
-    if let Ok(buffer) = get_file_buffer_from_sfs(file_path) {
-        return Ok((buffer, false, 0));
+    if let Ok((buffer, device_handle)) = get_file_buffer_from_sfs(file_path) {
+        return Ok((buffer, false, device_handle, 0));
     }
 
     if !boot_policy {
-        if let Ok(buffer) =
+        if let Ok((buffer, device_path)) =
```
joschock: nit: I think this var name is a typo and should be device_handle (since that's what the other func is returning).
```suggestion
        if let Ok((buffer, device_handle)) =

```

dxe_core/src/image.rs
```diff
@@ -623,36 +623,38 @@ fn core_load_pe_image(
...
             get_file_buffer_from_load_protocol(efi::protocols::load_file2::PROTOCOL_GUID, false, file_path)
         {
-            return Ok((buffer, false, 0));
+            return Ok((buffer, false, device_path, 0));
         }
     }
 
-    if let Ok(buffer) =
+    if let Ok((buffer, device_path)) =
```
joschock:
```suggestion
    if let Ok((buffer, device_handle)) =

```

dxe_core/src/image.rs
```diff
@@ -835,14 +837,26 @@ pub fn core_load_image(
...
         .map_err(|_| efi::Status::INVALID_PARAMETER)?;
 
-    let (image_to_load, from_fv, authentication_status) = match image {
-        Some(image) => (image.to_vec(), false, 0),
-        None => get_buffer_by_file_path(boot_policy, device_path)?,
+    let (image_to_load, from_fv, device_handle, authentication_status) = match image {
+        Some(image) => {
+            // If the buffer is specified, then the device_handle and device_path are set to the input file_path
+            // Check if this is coming from a FV device (if so, then core_locate_device_path(FV) will return OK)
+            if let Ok((_device_path, handle)) =
```
joschock: nit: suggest renaming this to fv_handle for clarity
```suggestion
            if let Ok((_device_path, fv_handle)) =

```

dxe_core/src/image.rs
```diff
@@ -835,14 +837,26 @@ pub fn core_load_image(
...
+                core_locate_device_path(mu_pi::protocols::firmware_volume::PROTOCOL_GUID, file_path)
+            {
+                (image.to_vec(), false, handle, 0)
+            } else {
+                // Error: this means that file_path is supposed to be a device path, but the device path isn't installed on any handle
```
joschock: might not be an actual error in all cases.
```suggestion
                // This means that file_path is supposed to be a device path, but the device path isn't installed on any handle

```

dxe_core/src/image.rs
```diff
@@ -852,18 +866,24 @@ pub fn core_load_image(
...
+    image_info.device_handle = device_handle;
+
+    if device_handle == protocol_db::INVALID_HANDLE {
+        image_info.file_path = file_path;
+    } else if !file_path.is_null() {
+        // Get the device path for the parent device
+        if let Ok(device_path) =
+            PROTOCOL_DB.get_interface_for_handle(device_handle, efi::protocols::device_path::PROTOCOL_GUID)
+        {
+            let (_, device_path_size) =
```
joschock, line 880: recommend an informative comment along the lines of "Strip the parent device path prefix from the full device path to leave only the file node" or something like that to describe what the following device path arithmetic is getting up to.

PR #244, opened 2025-02-07
Handle device path for DB_END cases

berlin-with0ut-return:
## Description
In the case of a driver-binding-related performance end measurement, we want the device path in addition to the module name. This handles that case.

- [X] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested
I generated an FPDT from this PR and it generates a module name and device path string as follows:
image

## Integration Instructions
N/A

### DIFF
--- a/crates/uefi_device_path/src/lib.rs
+++ b/crates/uefi_device_path/src/lib.rs
@@ -368,12 +368,10 @@ impl Iterator for DevicePathWalker {
     }
 }
 
-// cSpell marks certain words here as misspelled, even though they are technically spelled correctly
 fn protocol_to_subtype_str(protocol: efi::protocols::device_path::Protocol) -> &'static str {
     match protocol.r#type {
         r_efi::protocols::device_path::TYPE_HARDWARE => match protocol.sub_type {
             Hardware::SUBTYPE_PCI => "Pci",
-            // cspell:ignore someWord anotherWord
             Hardware::SUBTYPE_PCCARD => "PcCard",
             Hardware::SUBTYPE_MMAP => "MemMap",
             Hardware::SUBTYPE_VENDOR => "Vendor",


crates/uefi_device_path/src/lib.rs
```diff
@@ -343,6 +345,91 @@ impl Iterator for DevicePathWalker {
...
+/// # Safety
+///
+/// The caller must ensure that:
+/// - `device_path` points to a valid UEFI device path structure in memory
+/// - The device path is properly terminated with an end node
+/// - The device path nodes contain valid length fields that correctly represent their size
+/// - The memory referenced by the device path remains valid for the duration of this function call
+/// - Each node's length field accurately reflects the size of its data section
+///
+pub unsafe fn device_path_data_to_string(
```
Javagedes: I don't have the context around this, so its an actual question - Would this be better as a method of `DevicePathWalker`? Or maybe a `impl Into<String> for DevicePathWalker`? 

The big benefit I see is you move the `unsafe` verification to the `DevicePathWalker::new` function (which already exists and has the same requirements) instead of this function. And people can just do things like

`let path_str: String = DevicePathWalker::new(ptr).into()` or `let path_str: String = DevicePathWalker::new(ptr).to_string()` depending on the route you take.

crates/uefi_device_path/src/lib.rs
```diff
@@ -343,6 +368,45 @@ impl Iterator for DevicePathWalker {
     }
 }
 
+// cSpell marks certain words here as misspelled, even though they are technically spelled correctly
```
makubacki: Let's get [this PR](https://github.com/OpenDevicePartnership/uefi-dxe-core/pull/261) in first and then you can remove the cspell logic in this file.

crates/uefi_device_path/src/lib.rs
```diff
@@ -314,6 +316,29 @@ pub struct DevicePathWalker {
     next_node: Option<*const efi::protocols::device_path::Protocol>,
 }
 
+impl From<DevicePathWalker> for String {
```
makubacki, line 319: Would it be possible to add test coverage?

image

crates/uefi_device_path/src/lib.rs
```diff
@@ -343,6 +368,45 @@ impl Iterator for DevicePathWalker {
     }
 }
 
+// cSpell marks certain words here as misspelled, even though they are technically spelled correctly
+// cSpell:disable
+fn protocol_to_subtype_str(protocol: efi::protocols::device_path::Protocol) -> &'static str {
```
makubacki, line 371: Similarly, it would be great to have test coverage for this function as well.

PR #247, opened 2025-02-10
Update GCD Init - Set system memory regions to RP, and correctly handle PHIT free-memory region.

joschock:
## Description

This change updates the initialization of the GCD to set System Memory regions to RP by default (to ensure that all "free" memory is consistently RP) and resolves some inconsistencies with the treatment of the special "Free Memory" range that is reported by the PHIT hob for initial GCD construction. 

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Boots AARCH64 platform with paging enabled. Previously these inconsistencies in the memory map were causing errors that prevented boot.

## Integration Instructions

N/A

### DIFF
--- a/dxe_core/src/image.rs
+++ b/dxe_core/src/image.rs
@@ -813,10 +813,10 @@ fn authenticate_image(
 
 /// Loads the image specified by the device path (not yet supported) or slice.
 /// * parent_image_handle - the handle of the image that is loading this one.
-/// * device_path - optional device path describing where to load the image from.
+/// * file_path - optional device path describing where to load the image from.
 /// * image - optional slice containing the image data.
 ///
-/// One of `device_path` or `image` must be specified.
+/// One of `file_path` or `image` must be specified.
 /// returns the image handle of the freshly loaded image.
 pub fn core_load_image(
     boot_policy: bool,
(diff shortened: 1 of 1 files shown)


dxe_core/src/gcd.rs
```diff
@@ -213,22 +227,21 @@ pub fn add_hob_resource_descriptors_to_gcd(hob_list: &HobList) {
...
-                            "GCD failed to set memory attributes {:#X} for base: {:#X}, length: {:#X}",
-                            memory_attributes, res_desc.v1.physical_start, res_desc.v1.resource_length
-                        );
+                if let Some(attributes) = memory_attributes {
+                    match GCD.set_memory_space_attributes(
+                        split_range.start as usize,
+                        split_range.end.saturating_sub(split_range.start) as usize,
+                        attributes,
+                    ) {
+                        Err(EfiError::NotReady) => (),
```
os-d, line 239: nit: I would add a comment here that this is the expected case because the page table is not initialized yet and so will return not ready, but the GCD will be updated, which will in turn be used to update the page table. Same with the other case below.

joschock: updated comment.

PR #249, opened 2025-02-10
Fix advanced logger messages being optimized out

cfernald:
## Description

Because advanced logger entries were made without volatile access and a reference was returned, the compiler decided that it could optimize out the memory accesses. As a result, all the advanced logger entries after the rust core were empty, despite the header being updated.

Additionally, this fixes a structure issue and adds a static for debug purposes. 

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on Q35 & unit tests

## Integration Instructions

N/A

### DIFF
--- a/adv_logger/src/logger.rs
+++ b/adv_logger/src/logger.rs
@@ -17,6 +17,7 @@ use spin::Once;
 use uefi_sdk::{log::Format, serial::SerialIO};
 
 // Exists for the debugger to find the log buffer.
+#[used]
 static mut DBG_ADV_LOG_BUFFER: u64 = 0;
 
 /// The logger for memory/hardware port logging.


adv_logger/src/memory_log.rs
```diff
@@ -48,7 +48,7 @@ pub struct AdvLoggerInfo {
     /// Current Version
     version: u16,
     /// Reserved for future
-    reserved1: [u8; 3],
+    reserved1: [u16; 3],
```
makubacki, line 51: This looks like a separate bug, right?

adv_logger/src/logger.rs
```diff
@@ -16,6 +16,9 @@ use r_efi::efi;
 use spin::Once;
 use uefi_sdk::{log::Format, serial::SerialIO};
 
+// Exists for the debugger to find the log buffer.
+static mut DBG_ADV_LOG_BUFFER: u64 = 0;
```
Javagedes, line 21: I would mark this with `#[used]` to ensure the compiler never complies it out. If the compiler gets too smart, it will see we write it once and never use it again :)

https://rust-lang.github.io/rfcs/2386-used.html

cfernald:
> I'm fine with the debug u64, but is it add much beyond using the print to identify the address? Or are you going to update uefiext or something to be able to hook into this?

yeah, i have a change for uefiext to auto-find it similar to how it does with the C core.

adv_logger/src/memory_log.rs
cfernald: Yes, abut actually didnt break anything because it was lucky(?). The compiled structure is the same with and without this change.

In general, I think we are making some packing assumptions throughout the rust code that doesn't hold true, but definitely here. The rust compiler was 32-bit aligning the next structure already.  I'm going to see if I can make this stricter without upsetting the compiler.

Javagedes: You probably want #[repr(packed)]. But here is some pertinent information:

https://doc.rust-lang.org/reference/type-layout.html#the-transparent-representation

cfernald: Looked at this again. The structure is identical between this change and even if compiled with the `repr(c, packed)` the problem is that as soon as I add packed, even though the actual alignment doesn't change, It will no longer let me convert the fields into pointers for the atomic operations due it believing there may be alignment issues. I'm going forward with this fix as its still correct, but we are relying on some implicit compiler behavior that makes me nervous.

Javagedes: Maybe just change it to #[repr(packed(32))] (or 64) to (maybe?) fix the pointer issue, but ensure the compiler does not change its behavior later

os-d: You can also use the unaligned pointer reads and writes, right? Which is more annoying, but if it is just in one or two places it is not so bad. But I guess I don't know how that intersects with the atomic operations

PR #250, opened 2025-02-11
Design, Implementation, Testing, and Infrastructure Strategy for Firmware Volume Unit Testing and Code Coverage

HappyCoding-2025:
## Description

This is for supporting the Firmware Volume Unit Test Cases and Coverage.

Please note that Since all the Error Types need to be covered, the data structures are kept common for all valid and invalid cases.

The current coverage % is increased to 82.56%. The original with out changes were 24%.
This also exposed some bugs underlying too which I am planning to raise bugs for.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

cargo make coverage and checking the coverage report.

## Integration Instructions

N/A

### DIFF
--- a/adv_logger/src/component.rs
+++ b/adv_logger/src/component.rs
@@ -10,7 +10,7 @@
 //! SPDX-License-Identifier: BSD-2-Clause-Patent
 //!
 use alloc::boxed::Box;
-use core::{ffi::c_void, marker::PhantomData, ptr};
+use core::{ffi::c_void, ptr};
 use mu_pi::hob::{Hob, PhaseHandoffInformationTable};
 use r_efi::efi;
 use uefi_sdk::{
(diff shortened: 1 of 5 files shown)


makubacki:
@HappyCoding-2025, please notify the reviewers here when CI is passing.

HappyCoding-2025:
@Javagedes  - Thanks a lot for helping!


@makubacki  - ALl the checks have passed @Javagedes  @joschock

dxe_core/src/fv.rs
```diff
@@ -744,3 +744,726 @@ pub fn init_fv_support(hob_list: &hob::HobList, extractor: Box<dyn SectionExtrac
...
+                },
+                get_info: fv_get_info,
+                set_info: fv_set_info,
+            });
+
+            let fv_ptr = fv_interface.as_mut() as *mut mu_pi::protocols::firmware_volume::Protocol as *mut c_void;
+
+            let private_data = PrivateFvData { _interface: fv_interface, physical_address: base_address };
+            // save the protocol structure we're about to install in the private data.
+            PRIVATE_FV_DATA.lock().fv_information.insert(fv_ptr, PrivateDataItem::FvData(private_data));
```
joschock, line 898: PRIVATE_FV_DATA is global to all test instances. `with_global_lock()` should prevent them from running concurrently, but it does means that PRIVATE_FV_DATA may already have stuff in it from a prior test when a given test runs. Nothing in the test seems to depend on PRIVATE_FV_DATA state being deterministic at the start, so this extra data should baiscally be a don't-care, but might be good to either directly re-initialize PRIVATE_FV_DATA state to known values at start of test, or add a comment that PRIVATE_FV_DATA state may not be empty as an aid to future test developers

To phrase it another way, I suspect that an `assert!(PRIVATE_DATA.lock().fv_information.is_empty())` would fail for some tests if it was added to the beginning of each test.

HappyCoding-2025:
@joschock   - Thanks for the review and Good Point on the Private Global Data containing stale values.
I have now added a reset/clear to get this data to a known good state and also added assert!(PRIVATE_FV_DATA.lock().fv_information.is_empty()) and also added a comment for future developers.
With new code, the Private data is clean for every test function.

HappyCoding-2025:
@Javagedes  @makubacki  - addressed @joschock comments and new code is uploaded.

PR #255, opened 2025-02-12
Create integration test for advanced logger

cfernald:
## Description

Creates a simple integration test for advanced logger to check that messages logged through the log crate and the protocol both end up in the memory log.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on Q35.

## Integration Instructions

N/A

### DIFF
--- a/.github/workflows/ReleaseWorkflow.yml
+++ b/.github/workflows/ReleaseWorkflow.yml
@@ -66,7 +66,7 @@ jobs:
         run: cargo login "${{ secrets.CRATES_IO_TOKEN }}" --registry UefiRust
 
       - name: Cargo Release Dry Run
-        run: cargo release ${{ steps.release_tag.outputs.tag }} --workspace --registry UefiRust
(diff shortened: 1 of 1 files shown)


adv_logger/src/integration_test.rs
```diff
@@ -0,0 +1,78 @@
+//! Integrations tests for Advanced Logger.
```
makubacki: nit

```suggestion
//! Integration tests for Advanced Logger.
```

adv_logger/src/integration_test.rs
```diff
@@ -0,0 +1,78 @@
+//! Integrations tests for Advanced Logger.
+//!
```
makubacki, line 6: I think it'd be helpful to explain a bit more of the overall purpose and usage of this code here.

PR #258, opened 2025-02-13
Only notify BeforeExitBootServices once and don't lock memory until after

cfernald:
## Description

The recent Advanced Logger fix exposed another issue. During the BeforeExitBootServices callback, the C based Advanced Logger File Logger will flush the logs a final time during this callback. Because of the existing early GCD lock this causes an assertion failure. However, there is not strong alternative for such a scenario so this change will only lock the GCD after calling BeforeExitBootServices. 

Additionaly this change makes BeforeExitBootServices only be invoked once which is the behavior specified in the [UEFI Specification](https://uefi.org/specs/UEFI/2.10/07_Services_Boot_Services.html#efi-boot-services-exitbootservices). This provides the additional benefit of prevent events that do not register themselves from hitting the assert on a subsequent call from BeforeExitBootServices.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested boot to Windows on Q35

## Integration Instructions

N/A

### DIFF
--- a/dxe_core/src/misc_boot_services.rs
+++ b/dxe_core/src/misc_boot_services.rs
@@ -251,9 +251,6 @@ pub extern "efiapi" fn exit_boot_services(_handle: efi::Handle, map_key: usize) 
         EXIT_BOOT_SERVICES_CALLED.store(true, Ordering::SeqCst);
     }
 
-    // Lock the memory space to prevent edits to the memory map after this point.
-    GCD.lock_memory_space();
-
     // Disable the timer
     match PROTOCOL_DB.locate_protocol(protocols::timer::PROTOCOL_GUID) {
         Ok(timer_arch_ptr) => {
@@ -263,6 +260,9 @@ pub extern "efiapi" fn exit_boot_services(_handle: efi::Handle, map_key: usize) 
         }
         Err(err) => log::error!("Unable to locate timer arch: {:?}", err),
     };
+
+    // Lock the memory space to prevent edits to the memory map after this point.
+    GCD.lock_memory_space();
 
     // Terminate the memory map
     // According to UEFI spec, in case of an incomplete or failed EBS call we must restore boot services memory allocation functionality


dxe_core/src/misc_boot_services.rs
```diff
@@ -240,15 +240,18 @@ extern "efiapi" fn watchdog_arch_available(event: efi::Event, _context: *mut c_v
...
-    }
 
-    // Signal the event group before exit boot services
-    EVENT_DB.signal_group(efi::EVENT_GROUP_BEFORE_EXIT_BOOT_SERVICES);
+        // Signal the event group before exit boot services
+        EVENT_DB.signal_group(efi::EVENT_GROUP_BEFORE_EXIT_BOOT_SERVICES);
+
+        // Lock down the GCD so no further changes can be made, even if this EBS
+        // call fails.
+        GCD.lock_memory_space();
```
os-d: I think this needs to be outside of the conditional, because the memory services get unlocked if terminate_memory_map fails (as UEFI spec requires memory services are still available after failed EBS). So with the goal of preventing EBS handlers from changing the memory map uncatched, we would need to move this.

dxe_core/src/misc_boot_services.rs
```diff
@@ -240,15 +240,18 @@ extern "efiapi" fn watchdog_arch_available(event: efi::Event, _context: *mut c_v
 
 pub extern "efiapi" fn exit_boot_services(_handle: efi::Handle, map_key: usize) -> efi::Status {
     log::info!("EBS initiated.");
-    GCD.lock_memory_space();
-    // Pre-exit boot services is only signaled once
-    if !PRE_EXIT_BOOT_SERVICES_SIGNAL.load(Ordering::SeqCst) {
+    // Pre-exit boot services and before exit boot services are only signaled once
+    if !EXIT_BOOT_SERVICES_CALLED.load(Ordering::SeqCst) {
```
Javagedes, line 245: For now, I think I would prefer you put the static inside this function directly, to limit its scope.

Javagedes: This might be outside the scope of this PR. I missed that this was an existing static that you are reworking.

cfernald: i can make the change easy enough.

dxe_core/src/misc_boot_services.rs
```diff
@@ -239,16 +238,21 @@ extern "efiapi" fn watchdog_arch_available(event: efi::Event, _context: *mut c_v
...
+        // Signal the event group before exit boot services
+        EVENT_DB.signal_group(efi::EVENT_GROUP_BEFORE_EXIT_BOOT_SERVICES);
+
+        EXIT_BOOT_SERVICES_CALLED.store(true, Ordering::SeqCst);
     }
 
-    // Signal the event group before exit boot services
-    EVENT_DB.signal_group(efi::EVENT_GROUP_BEFORE_EXIT_BOOT_SERVICES);
+    // Lock the memory space to prevent edits to the memory map after this point.
+    GCD.lock_memory_space();
```
joschock: hmm. it occurs to me that maybe this should be moved after the timer is shut down?

makubacki: I think that makes sense, treating it like an explicitly signaled set of platform-controlled code called at that point.

os-d: I agree.

joschock: just for clarity, I mean the GCD lock should move after we shutdown the timer callbacks- I didn't mean the signal itself should move (but maybe that too? I haven't thought hard about that). We don't want to lock down the GCD, and then have a timer hit after it is locked since it is legal to allocate memory in a timer callback.

makubacki: Sorry I made that more confusing than it needed to be, we're talking about the same thing. I meant to draw attention to the fact that in the end, all of these - the EBS-related callbacks, the call to shutdown the timer (code in somewhere like 8254TimerDxe or HpetTimerDxe...), and timer callbacks are going to platform-controlled code that can modify the memory map therefore it makes sense to lock the GCD after all that opportunity.

PR #259, opened 2025-02-14
Align EFI_MEMORY_ATTRIBUTES_PROTOCOL to UEFI Spec

os-d:
## Description

The memory attributes protocol implementation in uefi-dxe-core was taking some liberties with its implementation w.r.t. the UEFI spec, namely it was returning some error codes not supported by the spec and more importantly, it had a bug in that it relied on lower layers of code to split the request per GCD descriptor, but the protocol code took attributes from the first descriptor of the range and attempted to apply that to the entire range (as these APIs are different than the GCD APIs that are used under the hood and so need some massaging).

This commit rectifies the bug and aligns the protocol API to the spec.

This fixes a bug seen when bootmgr attempts to unload winload while attempting automatic recovery, leading to a crash due to inconsistent GCD attributes across a range described by one GCD descriptor.

- [x] Impacts functionality?
- [x] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested during self map changes and discovered this bug. Windows boots successfully through recovery flow with this change.

## Integration Instructions

N/A.

### DIFF
--- a/dxe_core/src/memory_attributes_protocol.rs
+++ b/dxe_core/src/memory_attributes_protocol.rs
@@ -33,6 +33,15 @@ extern "efiapi" fn get_memory_attributes(
     // TODO: This should really go to the page table, not GCD, even though GCD is the source of truth...page table actually is
     match dxe_services::core_get_memory_space_descriptor(base_address) {
         Ok(descriptor) => {
+            if descriptor.base_address != base_address || descriptor.length != length {
+                log::error!(
+                    "{} Inconsistent attributes for: base_address {:#x} length {:#x}",
+                    function!(),
+                    base_address,
+                    length
+                );
+                return efi::Status::NO_MAPPING;
+            }
             unsafe { *attributes = descriptor.attributes & efi::MEMORY_ACCESS_MASK };
             efi::Status::SUCCESS
         }
@@ -43,7 +52,7 @@ extern "efiapi" fn get_memory_attributes(
                 status,
                 function!()
             );
-            status
+            efi::Status::NO_MAPPING
         }
     }
 }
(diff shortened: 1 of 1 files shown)


dxe_core/src/memory_attributes_protocol.rs
```diff
@@ -66,26 +75,41 @@ extern "efiapi" fn set_memory_attributes(
...
+
+        // it is still legal to split a descriptor and only set the attributes on part of it
+        let next_base = u64::min(descriptor_end, range_end);
+        let current_len = next_base - current_base;
+
+        // this API only adds new attributes that are set, it ignores all 0 attributes. So, we need to get the memory
+        // descriptor first and then set the new attributes as the GCD API takes into account all attributes set or unset.
+        let new_attributes = descriptor.attributes | attributes;
+
+        res = match dxe_services::core_set_memory_space_attributes(current_base, current_len, new_attributes) {
```
joschock: A failure here will continue to iterate, and a subsequent successful set on the next range will set `res` to success, right? 

So `res` here is only really reporting the result of the last chunk modified, and results of other blocks would be ignored?

dxe_core/src/memory_attributes_protocol.rs
```diff
@@ -106,26 +130,41 @@ extern "efiapi" fn clear_memory_attributes(
...
+
+        // it is still legal to split a descriptor and only set the attributes on part of it
+        let next_base = u64::min(descriptor_end, range_end);
+        let current_len = next_base - current_base;
+
+        // this API only adds clears attributes that are set to 1, it ignores all 0 attributes. So, we need to get the memory
+        // descriptor first and then set the new attributes as the GCD API takes into account all attributes set or unset.
+        let new_attributes = descriptor.attributes & !attributes;
+
+        res = match dxe_services::core_set_memory_space_attributes(current_base, current_len, new_attributes) {
```
joschock: same as previous comment.

dxe_core/src/memory_attributes_protocol.rs
os-d: Thanks, I will fix this and below to properly return an error. I do think it is a bit debatable what to do in the error case, whether to continue iterating over all blocks and get as close as you can to what the caller asked for or what I would probably do is reset the attributes back to what they were before we failed in the middle of the call. Right now we will of course just have some blocks changed and some not.

Resetting the old attributes gets a little tricky because we don't know how many blocks there are, we could use a vec, but allocating that could of course change the GCD and so when we go back to reset the old values we are now applying the memory attributes to the wrong region. That is not perhaps hugely likely because we have the pool allocator that would be backing the vec allocation, but if that triggered an expand the GCD could merge pool allocator owned blocks with each other (if I am remembering correctly on the image handle being set).

At any rate, if we fail in this protocol the system is dead anyway because only the bootloader uses this protocol and only to set up memory protections for the next image it is loading. So I think I may leave this as is for now (I'll make the change to return a failure if any failure occurred) rather than run a risk of applying the wrong attributes, if I am reasoning about that correctly. This way some of the intended memory regions have the necessary attributes and the bootloader knows about the failure. All it could really try to do (which I don't think it does) is allocate some new memory and try again.

joschock: Yes, I think failure in the middle is an unexpected case. So I'm good with a simple partial apply and fail out approach. 

Spec doesn't really stipulate any guarantees about what happens on failure with respect to the map. If we did something more complicated, I'd be in favor trying the rest of the blocks and returning some kind of "WARN_PARTIAL_UPDATE" status (that doesn't exist in the spec yet either); but agree that's really not necessary and should not be done without a spec update anyway.

(also, eh... spec needs to be updated: `Requested attributes cannot be applied due to *luck* of system resources.` - is probably a truthful statement in many firmware implementations, but not probably what was meant).

PR #260, opened 2025-02-14
Fix DXE Core Loaded Image Protocol

os-d:
## Description

DXE Core's loaded image protocol today is constructed using PrivateImageData::new(). However, this function expects that we are about to load this image, so it allocates an image buffer for the image to be loaded into and sets some fields with these assumptions.

DXE Core is unique in creating a loaded image protocol as it is already loaded, so PrivateImageData::new() allocates a large chunk of wasted memory for a buffer that never gets used and incorrectly lists where DXE core is located.

This also creates a problem in PrmLoaderDxe when it searches through loaded image protocols looking for PRM modules and crashes trying to read the invalid DXE core base address in the loaded image protocol.

With this change, that crash is fixed and the wasted memory is not allocated.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Discovered during self map changes. Tested on Q35 and observed this fixing the aforementioned crash in PrmLoaderDxe.

## Integration Instructions

N/A.

### DIFF
--- a/dxe_core/src/image.rs
+++ b/dxe_core/src/image.rs
@@ -219,6 +219,32 @@ impl PrivateImageData {
         Ok(image_data)
     }
 
+    fn new_with_existing_allocation(
+        image_info: efi::protocols::loaded_image::Protocol,
+        image_buffer: *mut [u8],
+        entry_point: efi::ImageEntryPoint,
+        pe_info: &UefiPeInfo,
+        image_base_page: efi::PhysicalAddress,
+        image_num_pages: usize,
+    ) -> Self {
+        PrivateImageData {
+            image_buffer,
+            image_info: Box::new(image_info),
+            hii_resource_section: None,
+            hii_resource_section_base: None,
+            hii_resource_section_num_pages: None,
+            entry_point,
+            started: true,
+            exit_data: None,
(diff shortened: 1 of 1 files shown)


dxe_core/src/image.rs
```diff
@@ -500,9 +500,29 @@ fn install_dxe_core_image(hob_list: &HobList) {
         .expect("Failed to parse PE info for DXE Core")
     };
 
-    let mut private_image_data =
-        PrivateImageData::new(image_info, &pe_info).expect("Failed to create PrivateImageData for dxe_core");
-    private_image_data.entry_point = entry_point;
+    // we do not use PrivateImageData::new() here because it
```
joschock, line 529: I'd recommend adding a new API to PrivateImageData (like `new_with_existing_allocation()` or something) and moving this into there rather than doing direct structure instantiation here.  Otherwise, structure details of PrivateImageData leak into this function; and this function has to change whenever a field is added or modified in PrivateImageData.

os-d: Sure, I can do that. This will be the only usage of that API but I think that's fine.

PR #262, opened 2025-02-19
Update paging to 1.0.2

joschock:
## Description

Update paging crate to 1.0.2

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Verified on AARCH64 platform with local build.

## Integration Instructions

N/A

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -51,7 +51,7 @@ log = { version = "^0.4", default-features = false }
 mockall = "0.13.0"
 mu_pi = { version = "^5.2" }
 mu_rust_helpers = { version = "2" }
-paging = { version = "1.0.1" }
+paging = { version = "^1.0.2" }
 rand = "^0.8"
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -51,7 +51,7 @@ log = { version = "^0.4", default-features = false }
 mockall = "0.13.0"
 mu_pi = { version = "^5.2" }
 mu_rust_helpers = { version = "2" }
-paging = { version = "1.0.1" }
+paging = { version = "1.0.2" }
```
Javagedes: probably needs a `^` (`^1.0.2`) so that we are not locking paging version to 1.0.2, and instead anything between 1.0.2 and 2 (not including 2)

joschock: Sure - updated. It was locked previously, so I followed that, but happy to let it float.

PR #271, opened 2025-02-20
Add SMM Perf Records

magravel:
## Description

Add smm performance record to the FPDT in an event running at ready to boot.

Some abstraction for communication protocol has been added here. Will get move eventually to a common place when we will be satisfied with those.

- [x]  Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- Tested in QEMU
Add some random perf measurements in smm to be able to validate that we can parse them in dxe core

- Tested on real hardware
Can boot to windows.

## Integration Instructions

N/A

### DIFF
--- a/.github/actions/rust-tool-cache/action.yml
+++ b/.github/actions/rust-tool-cache/action.yml
@@ -23,12 +23,18 @@ runs:
           ~/.rustup/toolchains/
         key: ${{ runner.os }}-rust-tools-${{ hashFiles('**/rust-toolchain.toml' )}}
 
-    # Installs toolchain specified in the rust-toolchain.toml file
+    # Installs the toolchain specified in the rust-toolchain.toml file if rustup >= 1.28.0
+    #
+    # if rustup < 1.28.0, this command will fail, but the cargo will automatically install the toolchain in the next
+    # step. This is because rustup less than 1.28.0 automatically installs the toolchain if it is not found where as
+    # This functionality was removed in 1.28.0 in favor of being able to call rustup toolchain install without the
+    # need to specify the toolchain name.
     - name: Install toolchains
       shell: bash
       run: |
         rustup toolchain install
       if: steps.tool-cache.outputs.cache-hit != 'true'
+      continue-on-error: true
 
     - name: Install cargo-binstall
       uses: cargo-bins/cargo-binstall@v1.10.17


crates/uefi_performance/src/lib.rs
```diff
@@ -392,10 +400,116 @@ extern "efiapi" fn report_fpdt_record_buffer(_event: efi::Event, _ctx: &()) {
...
+    let mut boot_records_data_retrieved = 0;
+    log::info!("[12345], loop");
+    loop {
+        let smm_boot_record_communicate = unsafe {
+            communication.communicate(
+                EFI_FIRMWARE_PERFORMANCE_GUID,
+                SmmBootRecordCommunicate {
+                    function: SmmBootRecordCommunicate::SMM_FPDT_FUNCTION_GET_BOOT_RECORD_DATA_BY_OFFSET,
+                    boot_record_offset: boot_records_data_retrieved,
+                    boot_record_data: (smm_communication_memory_region.physical_start as *const u8).byte_add(128),
```
makubacki: Mathieu and I discussed offline. With the way the MM Supervisor validates pointers, we have the following situation:

1. As a general rule, comm buffer data is always untrusted and pointer must always be validated.
2. The MM Supervisor double buffers comm buffer data to prevent TOCTOU attacks. This is not true of all MM environments.
3. The MM Supervisor MM pointer validation routine does not allow pointers outside of MMRAM unless the memory is explicitly unblocked. Right now, only comm buffers really are for general usage and they're intended to be reused throughout boot.
4. The MM Supervisor marks the comm buffer in the MM page table with `SP` attribute which fails the user comm mode buffer check during pointer validation. This means a pointer in the comm buffer (which must be validated) cannot point back into the comm buffer.

For these reasons, the following changes were made when adding Standalone MM perf support in these commits to avoid pointers entirely and if the incoming perf data pointer is `null`, add the data into the comm buffer directly after the headers. The Rust code here should do the same.

- [MdeModulePkg: Add Standalone MM Core performance support](https://github.com/microsoft/mu_basecore/commit/d1cd8c8632d157cf310be483dca28ddb9408a79a)
- [MdeModulePkg: Add Standalone MM perf library support](https://github.com/microsoft/mu_basecore/commit/069e16704e28332138841dc991255c800c48f858)

makubacki:
Thanks for publishing the PR. I'll review on Tuesday.

PR #273, opened 2025-02-27
Makefile.toml: Add a 'cargo make all' task for PR readiness

vineelko:
## Description

Makefile.toml: Add a `cargo make all` task for PR readiness. This makes it slightly easier to run all tasks for PR readiness in one command. It also includes cspell task, so a onetime installation of cspell is needed via `npm install -g cspell@latest`.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

N/A

## Integration Instructions

N/A


makubacki:
Once this is done in all the repos, a few steps in [uefi-dxe-core/.github/workflows/CiWorkflow.yml](https://github.com/OpenDevicePartnership/uefi-dxe-core/blob/main/.github/workflows/CiWorkflow.yml) should be updated to also call the corresponding `cargo make` command. For example, the cargo clippy step. That way, we have 1:1 results with CI on those and keep the definitions local to each repo in a single place.

vineelko:
Sure Michael, will do it.

PR #276, opened 2025-03-06
Drop Preallocation of Page Table Pages

os-d:
## Description

With the introduction of the self map in the paging crate, it is no longer required to preallocate page table pages. This commit removes the preallocation and instead allocates 512 pages to the page table pool on demand.

Note: this depends on the paging PR and uefi-core PR to go in first and be released, this will be updated in response.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested booting Q35 and SBSA to Windows. Physical platform testing is happening in parallel with PR review.

## Integration Instructions

N/A.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,10 +18,10 @@ edition = "2021"
 rust-version = "1.83.0"
 
 [workspace.dependencies]
-section_extractor = { version = "^9.1.1" }
+section_extractor = { version = "10" }
 uefi_sdk = { version = "2" }
-uefi_cpu = { version = "^9.1.1" }
-uefi_debugger = { version = "^9.1.1" }
+uefi_cpu = { version = "10" }
+uefi_debugger = { version = "10" }
 uefi_corosensei = { version = "0.1", default-features = false }
 
 adv_logger = { path = "adv_logger", version = "8" }
@@ -51,7 +51,7 @@ log = { version = "^0.4", default-features = false }
 mockall = "0.13.0"
 mu_pi = { version = "^5.2" }
 mu_rust_helpers = { version = "2" }
-paging = { version = "^1.0.2" }
+paging = { version = "2" }
 rand = "^0.8"
 r-efi = { version = "^5.2.0", default-features = false }
 scroll = { version = "^0.12", default-features = false, features = ["derive"] }
--- a/dxe_core/src/gcd/spin_locked_gcd.rs
+++ b/dxe_core/src/gcd/spin_locked_gcd.rs
@@ -23,7 +23,7 @@ use uefi_sdk::{
 
 use crate::{
     allocator::DEFAULT_ALLOCATION_STRATEGY, ensure, error, events::EVENT_DB, protocol_db, protocol_db::INVALID_HANDLE,
-    tpl_lock,
+    tpl_lock, GCD,
 };
 use paging::{page_allocator::PageAllocator, MemoryAttributes, PageTable, PtError, PtResult};
 use uefi_cpu::paging::create_cpu_paging;
@@ -42,6 +42,8 @@ pub const MEMORY_BLOCK_SLICE_SIZE: usize = MEMORY_BLOCK_SLICE_LEN * node_size::<
 
 const IO_BLOCK_SLICE_LEN: usize = 4096;
 const IO_BLOCK_SLICE_SIZE: usize = IO_BLOCK_SLICE_LEN * node_size::<IoBlock>();
+
+const PAGE_POOL_CAPACITY: usize = 512;
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 enum InternalError {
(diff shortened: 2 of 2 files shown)


dxe_core/src/gcd/spin_locked_gcd.rs
```diff
@@ -2262,6 +1755,114 @@ impl SpinLockedGcd {
...
             ),
             memory_change_callback,
+            page_table: tpl_lock::TplMutex::new(efi::TPL_HIGH_LEVEL, None, "GcdPageTableLock"),
+        }
+    }
+
+    fn set_paging_attributes(&self, base_address: usize, len: usize, attributes: u64) -> Result<(), EfiError> {
+        // we preallocate pages first, even if we won't create new mappings here, because we have to avoid borrowing
+        // self mutably twice and the logic is considerably simpler to preallocate first. In one case where we don't
+        // have a full page pool, we will cause an extra allocation here, but we would have done that next time we
```
joschock: stale comment?

os-d: Yes, thanks, I will go through and update all the comments on this one. I did on the other PRs, but not this one

dxe_core/src/gcd/spin_locked_gcd.rs
```diff
@@ -173,27 +173,14 @@ type GcdAllocateFn = fn(
...
 #[derive(Debug)]
 struct PagingAllocator {
     page_pool: Vec<efi::PhysicalAddress>,
-    root_page: Option<efi::PhysicalAddress>,
 }
 
 impl PagingAllocator {
     fn new() -> Self {
-        Self { page_pool: Vec::with_capacity(PAGE_POOL_MIN_CAPACITY), root_page: None }
+        Self { page_pool: Vec::with_capacity(512) }
```
joschock: Recommend retaining a const here (but renaming) as an aid to developer, because this `with_capacity` needs to match or exceed the
```
  // allocate 512 pages at a time
                    let len = 512;
```
below, or you will trip an FsbLock re-entrancy. By tying them to the same const, you reduce the possibility that one would be changed without the other under future maintenance.

os-d: Agreed, I actually just made that change locally as I was re-reviewing this myself

dxe_core/src/gcd/spin_locked_gcd.rs
os-d: I fixed this and other comments, but didn't save the file when I pushed. Will update shortly

dxe_core/src/gcd/spin_locked_gcd.rs
```diff
@@ -2607,6 +2388,14 @@ impl Display for SpinLockedGcd {
...
 }
 
+impl core::fmt::Debug for SpinLockedGcd {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        writeln!(f, "{:?}", self.memory.lock())?;
+        writeln!(f, "{:?}", self.io.lock())?;
+        Ok(())
+    }
+}
+
```
joschock: Consider a non-panicking version to allow for debug freedom when GCD lock is held. TplMutex already impls a lock-aware debug:

```suggestion
impl core::fmt::Debug for SpinLockedGcd {
    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
        writeln!(f, "{:?}", self.memory)?;
        writeln!(f, "{:?}", self.io)?;
        Ok(())
    }
}
```
if you want to avoid the TplMutext decorations you could consider using a similar construct using try_lock().

os-d: Agreed. I copied the display impl for SpinLockedGcd, I'll change that to try_lock as well

Cargo.toml
```diff
@@ -18,10 +18,10 @@ edition = "2021"
 rust-version = "1.83.0"
 
 [workspace.dependencies]
-section_extractor = { version = "^9.1.1" }
+section_extractor = { version = "10.0.0" }
```
Javagedes: This needs to be 10 (See larger comment. This is just so you cannot merge it)

os-d:
> All of these need to be "10" / "2" or you will be upgrading these again when you release 10.0.1 or something.

Ah, I misunderstood your offline comment and that makes more sense. If you specify exactly, you get that version. If you cut off earlier, you get until that point. Will fix.

dxe_core/src/gcd/spin_locked_gcd.rs
```diff
@@ -173,27 +175,14 @@ type GcdAllocateFn = fn(
...
-// track all the pages in the page pool, Vec will attempt to expand leading to a re-entrant access to the global allocator
-// which will deadlock. This sets the pool to an initial size that should be large enough to track all pages for any
-// reasonable allocation in the global allocator to avoid this problem.
-// Increased PAGE_POOL_MIN_CAPACITY from 1024 * 32 to 1024 * 512 to avoid re-entrant access issues caused by
-// resizing the vector
-//
-const PAGE_POOL_MIN_CAPACITY: usize = 1024 * 512;
-
 #[derive(Debug)]
 struct PagingAllocator {
```
Javagedes, line 179: I'd rather get this PR in now, so maybe I should raise an issue for this instead, but we may want to consider having this struct hold a reference to a SpinLockedGcd, rather than it's methods access the static version (GCD) directly. It'll make testing easier down the road and probably any type of refactoring down the road when we try to make testing easier in general.

os-d: This was the approach I started with and I ran into...some issue. I don't remember what now. And this was refactored since the original implementation so I'm not sure if the same issue will still occur. Something about either doubly mutable references or because this struct is not owned by the GCD code (it is moved to the paging code, then borrow occasionally).

I also want to get this in now :). Let's file the issue and I will take a look after this PR gets merged.

os-d: https://github.com/OpenDevicePartnership/uefi-dxe-core/issues/298

PR #284, opened 2025-03-10
Add SMM Perf Records

magravel:
## Description

Add smm performance record to the FPDT in an event running at ready to boot.

Some abstraction for communication protocol has been added here. Will get move eventually to a common place when we will be satisfied with those.

- [x]  Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- Tested in QEMU
Add some random perf measurements in smm to be able to validate that we can parse them in dxe core

- Tested on real hardware
Can boot to windows.

## Integration Instructions

N/A

# Note:
Some comments are in this pr: #271

### DIFF
--- a/docs/src/driver/interface.md
+++ b/docs/src/driver/interface.md
@@ -58,13 +58,20 @@ Please reference the documentation in `uefi_sdk::component` for more information
 
 ```rust
 use uefi_sdk::{
-    boot_services::StandardBootServices,
+    boot_services::BootServices,
     component::params::{Config, ConfigMut},
     error::Result,
 };
 
-fn validate_random_data_driver(bs: StandardBootServices, data: Config<&[u8]>, expected_crc32: Config<u32>) -> Result<()> {
-    assert_eq!(bs.calculate_crc_32(*data), Ok(*expected_crc32))
+// Note: This uncommented code is valid for demonstraction, but the function will not be registered with the core where
+//       actual implementations must be specified such as the wrapper function shown in the example below.
+//
+// The wrapper function:
+// fn actual_driver(bs: StandardBootServices, data: Config<f32>, expected_crc32: Config<u32>) -> Result<()> {
+//     validate_random_data_driver(bs, data, expected_crc32)
+// }
+fn validate_random_data_driver<T: Default>(bs: impl BootServices, data: Config<T>, expected_crc32: Config<u32>) -> Result<()> {
+    assert_eq!(bs.calculate_crc_32(&*data), Ok(*expected_crc32));
     Ok(())
 }
 
@@ -114,8 +121,8 @@ struct MyComponent2 {
 }
 
 impl MyComponent2 {
-    fn entry_point(self, public_config: ConfigMut<u32>) -> uefi_sdk::error::Result<()> {
-        public_config += self.private_config;
+    fn entry_point(self, mut public_config: ConfigMut<u32>) -> uefi_sdk::error::Result<()> {
+        *public_config += self.private_config;
         Ok(())
     }
 }


crates/uefi_performance/src/lib.rs
```diff
@@ -14,16 +14,18 @@
 extern crate alloc;
 
 pub mod _debug;
+mod _smm;
```
Javagedes, line 17: Does this module need to be `_` if it is not public? typically we only make modules `_<something>` if it is required by public, but technically a part of the private API (and marked with a `#[doc(hidden)]`

crates/uefi_performance/src/lib.rs
```diff
@@ -76,11 +83,10 @@ pub const PERF_ENABLED: bool = cfg!(feature = "instrument_performance");
 
 pub fn init_performance_lib(
     hob_list: &HobList,
-    efi_boot_services: &efi::BootServices,
-    efi_runtime_services: &efi::RuntimeServices,
+    efi_system_table: &'static efi::SystemTable,
 ) -> Result<(), efi::Status> {
-    BOOT_SERVICES.initialize(efi_boot_services);
-    RUNTIME_SERVICES.initialize(efi_runtime_services);
+    BOOT_SERVICES.initialize(unsafe { efi_system_table.boot_services.as_ref() }.unwrap());
```
Javagedes: Need some error handling instead of the `unwrap()`,

crates/uefi_performance/src/lib.rs
```diff
@@ -392,10 +398,97 @@ extern "efiapi" fn report_fpdt_record_buffer(_event: efi::Event, _ctx: &()) {
...
+    log::info!("Report FPDT Table: {:#?}", fbpt);
 }
 
-extern "efiapi" fn update_boot_performance_table(_event: efi::Event, _: &()) {
-    // TODO: There is a task in the backlog for this.
+extern "efiapi" fn fetch_and_add_smm_performance_records(_event: efi::Event, system_table: &efi::SystemTable) {
+    // Make sure that this event only run once.
+    static HAS_RAN_ONCE: AtomicBool = AtomicBool::new(false);
+    let Ok(false) = HAS_RAN_ONCE.compare_exchange(false, true, Ordering::Relaxed, Ordering::Relaxed) else {
+        return;
```
Javagedes, line 430: Should probably print an error message here

crates/uefi_performance/src/lib.rs
```diff
@@ -392,10 +398,97 @@ extern "efiapi" fn report_fpdt_record_buffer(_event: efi::Event, _ctx: &()) {
...
+    };
+
+    let configuration_tables =
+        unsafe { slice::from_raw_parts(system_table.configuration_table, system_table.number_of_table_entries) };
+
+    let Some(smm_comm_region_table) = configuration_tables
+        .iter()
+        .find(|config_table| config_table.vendor_guid == EDKII_PI_SMM_COMMUNICATION_REGION_TABLE_GUID)
+        .and_then(|config_table| unsafe { (config_table.vendor_table as *const SmmCommunicationRegionTable).as_ref() })
```
Javagedes: Need to document safety here, and what this code is doing such that safety does not need to be bubbled up to the function defintion.

crates/uefi_performance/src/_smm.rs
```diff
@@ -0,0 +1,295 @@
...
+
+impl Deref for CommunicateProtocol {
+    type Target = efi::Guid;
+
+    fn deref(&self) -> &Self::Target {
+        self.protocol_guid()
+    }
+}
+
+unsafe impl Protocol for CommunicateProtocol {
```
Javagedes, line 92: I was under the impression @makubacki was implementing SMM communicate as a [Service](https://github.com/OpenDevicePartnership/uefi-sdk/blob/main/uefi_sdk/src/component/service.rs) that gets backported with the `#[protocol = "..."]` macro attribute associated with `IntoService`.

Is that no longer the case? 

(Question targeted at @makubacki and @magravel )

crates/uefi_performance/src/_smm.rs
```diff
@@ -0,0 +1,295 @@
...
+
+/// This trait should be implemented on type that represent a communicate data.
+///
+/// [`TryIntoCtx`] is used to define how to write the struct as the data in communicate buffer.
+///
+/// [`TryFromCtx`] is used to define how to read the struct as the data in communicate buffer.
+///
+/// # Safety
+/// Make sure you write and read the struct in the expected format defined by the guid.
+pub unsafe trait CommunicateData:
```
Javagedes, line 108: Not a huge fan of this trait. and the const GUID. We've already seen problems associated with this design patter regarding protocols. 

It feels unnecessary. Developers writing some type of well-defined structure for Commnication over this layer should just know that the requirement is to write a guid at the beginning of there IntoCtx implementation.

crates/uefi_performance/src/lib.rs
```diff
@@ -392,10 +398,97 @@ extern "efiapi" fn report_fpdt_record_buffer(_event: efi::Event, _ctx: &()) {
...
+    };
+
+    let Some(smm_communication_memory_region) =
+        smm_comm_region_table.iter().find(|a| a.r#type == efi::CONVENTIONAL_MEMORY)
+    else {
+        log::error!("Could not find an available memory region to communication with smm.");
+        return;
+    };
+
+    let Ok(communication) = (unsafe { BOOT_SERVICES.locate_protocol(&CommunicateProtocol, None) }) else {
```
Javagedes, line 462: Need `unsafe` documentation

crates/uefi_performance/src/lib.rs
```diff
@@ -392,10 +398,97 @@ extern "efiapi" fn report_fpdt_record_buffer(_event: efi::Event, _ctx: &()) {
...
+        log::error!("Could not locate communicate protocol interface.");
+        return;
+    };
+
+    // Ask smm for the total size of the perf records.
+    let boot_record_size = match unsafe {
```
Javagedes, line 468: need `unsafe` documentation.

crates/uefi_performance/src/lib.rs
```diff
@@ -392,10 +398,97 @@ extern "efiapi" fn report_fpdt_record_buffer(_event: efi::Event, _ctx: &()) {
...
+        Err(status) => {
+            log::error!("Error while trying to communicate with communicate protocol with error code: {:?}", status);
+            return;
+        }
+    };
+
+    let mut smm_boot_records_data = Vec::with_capacity(boot_record_size);
+
+    while smm_boot_records_data.len() < boot_record_size {
+        match unsafe {
```
Javagedes, line 492: need unsafe documentation.

dxe_core/src/lib.rs
```diff
@@ -403,9 +405,10 @@ where
         // reading the time stamp counter in the way done in this code and results in a divide by zero exception.
         // Other cpu models crash in various other ways. It will be resolved, but is removed now to unblock other
         // development
-        // _ = uefi_performance::init_performance_lib(&hob_list, unsafe { boot_services_ptr.as_ref().unwrap() }, unsafe {
-        //     runtime_services_ptr.as_ref().unwrap()
-        // });
+        _ = uefi_performance::init_performance_lib(
+            &hob_list,
+            unsafe { system_table_ptr.as_ref() }.expect("System Table not initialized!"),
```
Javagedes, line 408: need unsafe documentation.

crates/uefi_performance/src/lib.rs
```diff
@@ -477,7 +570,7 @@ fn get_module_info_from_handle(
     // Method 3 - Get the name string from FFS UI Section.
     // TODO: https://github.com/OpenDevicePartnership/uefi-dxe-core/issues/193
 
-    Ok((None, guid))
+    Ok((Some(String::from("TODO Get name from UefiPeInfo")), guid))
```
Javagedes: Not sure I agree with this change.

crates/uefi_performance/src/lib.rs
magravel: Yes, in the case of this crate module with a prefix "_" is a marker for me to mark the module as temporary and that most of the code there should live elsewhere. But yes, you are right, usually this is to state that a module is not public. Maybe we could find a better notation for these kinds of use cases.

crates/uefi_performance/src/lib.rs
magravel: The even run exactly 3 times in QEMU, I assumed this event running more than once was normal, so I am not sure if printing a warn or error in the case that it is expected to run more that once appropriate. Do you know if this should ideally be called only once?

crates/uefi_performance/src/_smm.rs
magravel: I would have to look into merging my smm with the other existing smm because I have some interesting changes in this one, but when I came to do it, I noticed that the change could be a task in itself because I saw in mu_rust_pi Communication, Communication2 and Communication3. I think my abstraction only applies for Communication, and would be interesting to create a task that will add the improvement for all of them at once.

crates/uefi_performance/src/lib.rs
magravel: Yes, that was a line of code used for debugging.

crates/uefi_performance/src/_smm.rs
magravel: The reason I wanted a guid is to not force a full implementation of IntoCtx and FromCtx when sometimes the deref Pwrite / Pread could be used. Also, you need to write the size in the header which sometime you don't know at compile time, so it would also force the user to write that and do the check what is written in the buffer is the same check that the header value. I see the issue with forcing a const guid, I could make it not a const, with maybe a fn in the trait. I think this abstraction could take more thinking since the plan is to move it to mu_rust_pi in the future.

crates/uefi_performance/src/_smm.rs
makubacki: Let's get this in and then when communicate comes, the changes here can be reconciled.

Javagedes:
@makubacki are we no longer planning on producing the communicate as a `Service<...>` that also backports as protocol?

Javagedes:
Also is there anyway this can be another component like adv logger instead of being directly integrated into the core?

makubacki:
> @makubacki are we no longer planning on producing the communicate as a `Service<...>` that also backports as protocol?

We are, but Mathieu's temporary code here will be updated when that service is added so platforms can get performance tracing support in the meantime.

makubacki:
> Also is there anyway this can be another component like adv logger instead of being directly integrated into the core?

I've given this feedback in the past too. @magravel, this is something I'd expect you to follow up on, does that sound okay?

magravel:
> > Also is there anyway this can be another component like adv logger instead of being directly integrated into the core?
> 
> I've given this feedback in the past too. @magravel, this is something I'd expect you to follow up on, does that sound okay?

So instead of being always enabled in the core, we would add it in rust_bin as a component like this `.with_component(performance_component)`? Do we want to add this in this PR or the next one that will clean up and finalize performance records.

makubacki:
> > > Also is there anyway this can be another component like adv logger instead of being directly integrated into the core?
> > 
> > 
> > I've given this feedback in the past too. @magravel, this is something I'd expect you to follow up on, does that sound okay?
> 
> So instead of being always enabled in the core, we would add it in rust_bin as a component like this `.with_component(performance_component)`? Do we want to add this in this PR or the next one that will clean up and finalize performance records.

I am fine with it being in a future PR. I filed a tracking issue https://github.com/OpenDevicePartnership/uefi-dxe-core/issues/319.

crates/uefi_performance/src/lib.rs
```diff
@@ -76,15 +83,33 @@ pub const PERF_ENABLED: bool = cfg!(feature = "instrument_performance");
 
 pub fn init_performance_lib(
     hob_list: &HobList,
-    efi_boot_services: &efi::BootServices,
-    efi_runtime_services: &efi::RuntimeServices,
+    efi_system_table: &'static efi::SystemTable,
 ) -> Result<(), efi::Status> {
-    BOOT_SERVICES.initialize(efi_boot_services);
-    RUNTIME_SERVICES.initialize(efi_runtime_services);
+    // SAFETY: This is safe because `boot_services` and `runtime services` are valid pointer to theire respective struct inside the systeme table.
```
makubacki:
```suggestion
    // SAFETY: This is safe because `boot_services` and `runtime services` are valid pointer to theire respective struct inside the system table.
```

crates/uefi_performance/src/lib.rs
```diff
@@ -394,8 +423,105 @@ extern "efiapi" fn report_fpdt_record_buffer(_event: efi::Event, _ctx: &()) {
     }
 }
 
-extern "efiapi" fn update_boot_performance_table(_event: efi::Event, _: &()) {
-    // TODO: There is a task in the backlog for this.
+extern "efiapi" fn fetch_and_add_smm_performance_records(_event: efi::Event, system_table: &efi::SystemTable) {
+    // Make sure that this event only run once.
+    static HAS_RAN_ONCE: AtomicBool = AtomicBool::new(false);
```
makubacki:
```suggestion
    static HAS_RUN_ONCE: AtomicBool = AtomicBool::new(false);
```

crates/uefi_performance/src/lib.rs
```diff
@@ -394,8 +423,105 @@ extern "efiapi" fn report_fpdt_record_buffer(_event: efi::Event, _ctx: &()) {
...
+    };
+
+    let configuration_tables =
+        unsafe { slice::from_raw_parts(system_table.configuration_table, system_table.number_of_table_entries) };
+
+    let Some(smm_comm_region_table) = configuration_tables
+        .iter()
+        .find(|config_table| config_table.vendor_guid == EDKII_PI_SMM_COMMUNICATION_REGION_TABLE_GUID)
+        .and_then(|config_table| {
+            // SAFETY: The cast of vendor_table to `SmmCommunicationRegionTable` is safe
```
makubacki: I'm not sure I'd consider this safe as much as reasonable/valid.

```suggestion
            // SAFETY: The cast of vendor_table to `SmmCommunicationRegionTable` is valid
```

crates/uefi_performance/src/lib.rs
```diff
@@ -394,8 +423,105 @@ extern "efiapi" fn report_fpdt_record_buffer(_event: efi::Event, _ctx: &()) {
...
+    };
+
+    let Some(smm_communication_memory_region) =
+        smm_comm_region_table.iter().find(|a| a.r#type == efi::CONVENTIONAL_MEMORY)
+    else {
+        log::error!("Could not find an available memory region to communication with smm.");
+        return;
+    };
+
+    // SAFETY: This is save because the reference returned by locate_protocol is nevery mutated after installation.
```
makubacki:
```suggestion
    // SAFETY: This is safe because the reference returned by locate_protocol is never mutated after installation.
```

crates/uefi_performance/src/lib.rs
```diff
@@ -394,8 +423,105 @@ extern "efiapi" fn report_fpdt_record_buffer(_event: efi::Event, _ctx: &()) {
...
+    let Ok(communication) = (unsafe { BOOT_SERVICES.locate_protocol(&CommunicateProtocol, None) }) else {
+        log::error!("Could not locate communicate protocol interface.");
+        return;
+    };
+
+    // SAFETY: Is safe to use because the memroy region commes for a thrusted source and can be considered valid.
```
makubacki:
```suggestion
    // SAFETY: Is safe to use because the memory region comes for a trusted source and can be considered valid.
```

makubacki:
@magravel, did you test this on an ARM64 platform like QEMU SBSA?

crates/uefi_performance/src/_smm.rs
```diff
@@ -0,0 +1,295 @@
+use core::{marker::PhantomPinned, ops::Deref, ptr, slice};
```
makubacki: Can you add a comment to the top of the file that explains it is temporary and what purpose it is serving during the temporary period of time?

crates/uefi_performance/src/_smm.rs
```diff
@@ -0,0 +1,295 @@
...
+
+unsafe impl Protocol for CommunicateProtocol {
+    type Interface = CommunicateProtocolInterface;
+
+    fn protocol_guid(&self) -> &'static efi::Guid {
+        &EFI_SMM_COMMUNICATION_PROTOCOL_GUID
+    }
+}
+
+/// This trait should be implemented on type that represent a communicate data.
```
makubacki:
```suggestion
/// This trait should be implemented on type that represents communicate data.
```

crates/uefi_performance/src/_smm.rs
```diff
@@ -0,0 +1,295 @@
...
+            Ok(comm_buffer.pread_with::<T>(data_offset, scroll::NATIVE).unwrap())
+        }
+    }
+}
+
+// Communicate protocol data layout to communicate with smm.
+#[derive(Debug, Default, Pwrite, Pread)]
+#[repr(C)]
+struct SmmBootRecordCommunicateMemoryLayout {
+    function: u64,
```
makubacki: This is a `UINTN`. Was there a reason not to use `usize` here?

crates/uefi_performance/src/_smm.rs
```diff
@@ -0,0 +1,295 @@
...
+    return_status: u64,
```
makubacki: Similarly, status code is `usize`.

```suggestion
    return_status: usize,
```

crates/uefi_performance/src/_smm.rs
```diff
@@ -0,0 +1,295 @@
...
+    boot_record_size: u64,
```
makubacki: `UINTN` should be `usize`.

```suggestion
    boot_record_size: usize,
```

crates/uefi_performance/src/_smm.rs
```diff
@@ -0,0 +1,295 @@
...
+    boot_record_data: u64,
```
makubacki:
```suggestion
    boot_record_data: usize,
```

crates/uefi_performance/src/_smm.rs
```diff
@@ -0,0 +1,295 @@
...
+    boot_record_offset: u64,
```
makubacki:
```suggestion
    boot_record_offset: usize,
```

crates/uefi_performance/src/_smm.rs
```diff
@@ -0,0 +1,295 @@
...
+}
+
+impl SmmBootRecordCommunicateMemoryLayout {
+    pub const EFI_FIRMWARE_PERFORMANCE_GUID: efi::Guid =
+        efi::Guid::from_fields(0xc095791a, 0x3001, 0x47b2, 0x80, 0xc9, &[0xea, 0xc7, 0x31, 0x9f, 0x2f, 0xa4]);
+    pub const SMM_FPDT_FUNCTION_GET_BOOT_RECORD_SIZE: u64 = 1;
+    pub const SMM_FPDT_FUNCTION_GET_BOOT_RECORD_DATA_BY_OFFSET: u64 = 3;
```
makubacki: It seems a little unintuitive that the MMI handler function IDs are in the communicate memory layout structure. They're not really related to memory layout they're just values used to resolve what action to take in the MMI handler.

crates/uefi_performance/src/lib.rs
```diff
@@ -394,8 +423,105 @@ extern "efiapi" fn report_fpdt_record_buffer(_event: efi::Event, _ctx: &()) {
...
+            // and the expected value of this configuration is a `SmmCommunicationRegionTable`.
+            unsafe { (config_table.vendor_table as *const SmmCommunicationRegionTable).as_ref() }
+        })
+    else {
+        log::error!("Could not find any smm communication region table.");
+        return;
+    };
+
+    let Some(smm_communication_memory_region) =
+        smm_comm_region_table.iter().find(|a| a.r#type == efi::CONVENTIONAL_MEMORY)
```
makubacki: If it is not already being checked, some other properties of the comm region found should be verified like that physical start address and number of pages being valid.

crates/uefi_performance/src/_smm.rs
```diff
@@ -0,0 +1,295 @@
...
+            Self {
+                boot_record_size: boot_record_communicate.boot_record_size as usize,
+                return_status: efi::Status::from_usize(boot_record_communicate.return_status as usize),
+            },
+            offset,
+        ))
+    }
+}
+
+// Communicate protocol data to ask smm to retrun a BUFFER_SIZE about of byte at an offset.
```
makubacki:
```suggestion
// Communicate protocol data to ask smm to return a BUFFER_SIZE about of byte at an offset.
```

crates/uefi_performance/src/_smm.rs
magravel: It was because of Pwrite, I wanted to use the derive macro.

makubacki: We should stick to the proper types for now to prevent portability issues from sneaking up in the future. This was a problem in PEI code making 32-bit assumptions that became a problem in the transition to 64-bit PEI.

magravel:
> @magravel, did you test this on an ARM64 platform like QEMU SBSA?

No, I'll test it.

PR #290, opened 2025-03-12
CI: dry run publishing to a registry

Javagedes:
## Description

Changes to a crate's Cargo.toml package
information can result in a crate not being able
to publish to a registry (such as a description
missing). This PR adds a new step to our CI that
attempts to publish the crate to the registry
before the PR is merged. This will ensure the
break will be caught before it is merged in,
instead of during the release process.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This PRs workflow

## Integration Instructions

N/A

### DIFF
--- a/.github/workflows/CiWorkflow.yml
+++ b/.github/workflows/CiWorkflow.yml
@@ -102,7 +102,7 @@ jobs:
           RUSTC_BOOTSTRAP: 1
       
       - name: Crate Release Dry Run
-        run: cargo release --no-tag --workspace --no-push --registry UefiRust
+        run: cargo release --no-tag --workspace --no-push --allow-branch HEAD --registry UefiRust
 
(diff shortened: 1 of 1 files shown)


makubacki:
I thought something like this was already in place but maybe I'm thinking about local operations done when first publishing or another repo.

Javagedes:
> I thought something like this was already in place but maybe I'm thinking about local operations done when first publishing or another repo.

We have logic to catch build build and documentation failures, which would result in the crate failing to upload, but we just hit an issue in `uefi_sdk` where we failed to publish because the crate package information was missing a description, which also causes us to fail to release. So this should catch that.

PR #300, opened 2025-03-14
dxe_core: Make Dev-dependencies set via path

Javagedes:
## Description

This is a solution specified via https://github.com/crate-ci/cargo-release/issues/829 for broken ordering when attempting to publish multiple crates in the repository.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

N/A

## Integration Instructions

N/A

### DIFF
--- a/dxe_core/Cargo.toml
+++ b/dxe_core/Cargo.toml
@@ -47,6 +47,10 @@ uefi_depex = { workspace = true}
 uefi_performance = { workspace = true }
 
 [dev-dependencies]
+# To avoid circular dependencies, cargo-release skips dev dependencies when evaluating the release order for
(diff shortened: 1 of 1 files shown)


dxe_core/Cargo.toml
```diff
@@ -47,9 +47,9 @@ uefi_depex = { workspace = true}
 uefi_performance = { workspace = true }
 
 [dev-dependencies]
```
makubacki, line 49: Might want to add a comment to the issue so someone understands why this was done (and/or doesn't inadvertently revert it) without using blame.

Javagedes: Completed. Will need a re-review to merge in. Thanks!

PR #301, opened 2025-03-15
dxe_core: Replace ahash with Xorshift64* based hasher

### LINKED ISSUE #296: [Feature]: Remove usage of `ahash` crate
### Feature Overview

`ahash` eventually brings in `tiny-keecak` which is not cryptographically secure, has not been updated in 5 years, and is flagged in component governance.

```
dxe_core v8.1.0
├── ahash v0.8.11
│   ├── cfg-if v1.0.0
│   ├── const-random v0.1.18
│   │   └── const-random-macro v0.1.16 (proc-macro)
│   │       ├── getrandom v0.2.15
│   │       │   └── cfg-if v1.0.0
│   │       ├── once_cell v1.20.2
│   │       └── tiny-keccak v2.0.2
│   │           └── crunchy v0.2.3
```

### Solution Overview

Replace with an alternative for static hashes.

### Alternatives Considered

_No response_

### Urgency

Medium

No maintainer feedback needed

### Anything else?

_No response_

Javagedes:
## Description

replaces the ahash crate with a custom hasher based off of [Xorshift64*](https://en.wikipedia.org/wiki/Xorshift#xorshift*) for generating a unique number for each handle.

closes #296

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Booted to shell on qemuq35

## Integration Instructions

N/A

### DIFF
--- a/dxe_core/src/protocol_db.rs
+++ b/dxe_core/src/protocol_db.rs
@@ -1922,4 +1922,29 @@ mod tests {
             }
         });
     }
+
+    #[test]
+    fn xorshift64starhasher_test_different_seeds() {
+        let seed1 = 12345;
+        let seed2 = 54321;
+        let mut hasher1 = Xorshift64starHasher::new(seed1);
+        let mut hasher2 = Xorshift64starHasher::new(seed2);
+
+        let num1 = hasher1.next_state();
+        let num2 = hasher2.next_state();
+
+        assert_ne!(num1, num2, "Random numbers should be different for different seeds");
+    }
+
+    #[test]
+    fn xorshift64starhasher_test_same_seed() {
+        let seed = 12345;
+        let mut hasher1 = Xorshift64starHasher::new(seed);
+        let mut hasher2 = Xorshift64starHasher::new(seed);
+
+        let num1 = hasher1.next_state();
+        let num2 = hasher2.next_state();
+
+        assert_eq!(num1, num2, "Random numbers should be the same for the same seed");
+    }
 }


Javagedes:
I may switch this to sha2 instead if blake3 if it is not going to compile easily. Will look at it on monday

joschock:
> I may switch this to sha2 instead of blake3 if it is not going to compile easily. Will look at it on monday

FWIW, hashing the handles is more about enforcing opacity on the API then it is about security (i.e. surfacing bugs based on assumptions of how handles are generated and/or assigned and assumptions about what they point to or don't). We don't make any promises that I'm aware of based on the hash properties of the handles. So - if it's simpler or easier to go with something like sha2 (or even sha1), I don't think the security elements of it matter as much as "fast uniqueness."

Javagedes:
> > I may switch this to sha2 instead of blake3 if it is not going to compile easily. Will look at it on monday
> 
> FWIW, hashing the handles is more about enforcing opacity on the API then it is about security (i.e. surfacing bugs based on assumptions of how handles are generated and/or assigned and assumptions about what they point to or don't). We don't make any promises that I'm aware of based on the hash properties of the handles. So - if it's simpler or easier to go with something like sha2 (or even sha1), I don't think the security elements of it matter as much as "fast uniqueness."

@joschock In that case I just wrote my own custom hasher which is based off of [Xorshift*](https://en.wikipedia.org/wiki/Xorshift#xorshift*) random number generator to reduce dependencies. You may want to re-review as it is completely different now.

makubacki:
@Javagedes, I thought it was easier to just push the unit tests I think make sense.

PR #303, opened 2025-03-18
PagingAllocator: Store GCD Reference in Struct

### LINKED ISSUE #298: [Bug]: Pass Reference to GCD to PageAllocator Struct

### Current Behavior

The PageAllocator struct calls directly into the static GCD.

### Expected Behavior

If possible, this should have a reference to the GCD and access functionality through that.

### Steps To Reproduce

n/a

### Build Environment

```markdown
- OS(s):
- Tool Chain(s):
- Targets Impacted:
```

### Version Information

```text
top of tree
```

### Urgency

Low

I will fix it

No maintainer feedback needed

### Anything else?

_No response_

os-d:
## Description

In order to facilitate easier testing and extensibility, have the PagingAllocator struct accept a GCD reference in the new() function and call allocate operations on it instead of the static mut GCD directly.

Closes #298.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Q35 and SBSA boot to Windows.

## Integration Instructions

N/A.

### DIFF
--- a/dxe_core/src/gcd/spin_locked_gcd.rs
+++ b/dxe_core/src/gcd/spin_locked_gcd.rs
@@ -176,18 +176,18 @@ type GcdFreeFn =
     fn(gcd: &mut GCD, base_address: usize, len: usize, transition: MemoryStateTransition) -> Result<(), EfiError>;
 
 #[derive(Debug)]
-struct PagingAllocator {
+struct PagingAllocator<'a> {
     page_pool: Vec<efi::PhysicalAddress>,
-    gcd: &'static SpinLockedGcd,
+    gcd: &'a SpinLockedGcd,
 }
 
-impl PagingAllocator {
-    fn new(gcd: &'static SpinLockedGcd) -> Self {
+impl<'a> PagingAllocator<'a> {
+    fn new(gcd: &'a SpinLockedGcd) -> Self {
         Self { page_pool: Vec::with_capacity(PAGE_POOL_CAPACITY), gcd }
     }
 }
 
-impl PageAllocator for PagingAllocator {
+impl PageAllocator for PagingAllocator<'_> {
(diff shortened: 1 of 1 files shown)


dxe_core/src/gcd/spin_locked_gcd.rs
```diff
@@ -178,11 +178,12 @@ type GcdFreeFn =
 #[derive(Debug)]
 struct PagingAllocator {
     page_pool: Vec<efi::PhysicalAddress>,
+    gcd: &'static SpinLockedGcd,
```
Javagedes: We should not directly be using static here. This requires (particularly in tests) that we still instantiate a static SpinLockedGcd to do tests (but its still better because we don't have to use _the_ global static `GCD`).

My suggestion is to make this a generic lifetime `'a` and have something higher up the stack set the generic `'a` lifetime to `'static`. 

In a perfect would we would hold off on all `'static` claims until the actual instantiation of the global static `GCD`, but unfortunately the `create_cpu_paging` function only returns a `'static` lifetime page table, so I think we will need to limit how far we can take the "generic" up the static until that is refactored.

I think this commit I did is a decent starting point. But we can also just hold off if you prefer:

https://github.com/OpenDevicePartnership/uefi-dxe-core/commit/4b089a0ed96690eb99c975d771a8b715987dbedf

os-d: That's fine with me, I almost went the generic lifetime direction and then decided I didn't understand that well enough :). I am in the middle of actually generating some unit tests for the PagingAllocator, figured I would put my money where my mouth is, let me push that commit to this branch, then do you want to adjust the original commit with your change and push this branch?

Javagedes: pushed

PR #304, opened 2025-03-19
Cleanup internal use of efi::Status in favor of Result<_, EfiError>

cfernald:
## Description

Changes the internal use of the C-style `efi::Status` for a rust style `Result<_, EfiError>`. This allows for better pattern matching and internal consistency with the rest of the codebase.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on QEMU

## Integration Instructions

N/A


cfernald:
> This doesn't need to be part of this PR (and can't be), but now that we are centralizing on EfiError, if we haven't already done a pretty print for it, we should

@os-d It has a derive debug.

https://github.com/OpenDevicePartnership/uefi-sdk/blob/3c3f8b15f7648506f554b5452bfbe53ef325c073/uefi_sdk/src/error.rs#L17

os-d:
> > This doesn't need to be part of this PR (and can't be), but now that we are centralizing on EfiError, if we haven't already done a pretty print for it, we should
> 
> @os-d It has a derive debug.
> 
> https://github.com/OpenDevicePartnership/uefi-sdk/blob/3c3f8b15f7648506f554b5452bfbe53ef325c073/uefi_sdk/src/error.rs#L17

Does that print out the text or the hex? It would be good to print the text, also adding a Display.

cfernald:
> > > This doesn't need to be part of this PR (and can't be), but now that we are centralizing on EfiError, if we haven't already done a pretty print for it, we should
> > 
> > 
> > @os-d It has a derive debug.
> > https://github.com/OpenDevicePartnership/uefi-sdk/blob/3c3f8b15f7648506f554b5452bfbe53ef325c073/uefi_sdk/src/error.rs#L17
> 
> Does that print out the text or the hex? It would be good to print the text, also adding a Display.

@os-d deriving debug for an enum should result in format converting it to it's name string. e.g `InvalidParameter`

"This trait can be used with #[derive] if all fields implement Debug. When derived for structs, it will use the name of the struct, then {, then a comma-separated list of each field’s name and Debug value, then }. __For enums, it will use the name of the variant and, if applicable, (, then the Debug values of the fields, then ).__"

from: https://doc.rust-lang.org/std/fmt/trait.Debug.html

PR #309, opened 2025-03-21
[REBASE & FF] Apply Some Paging Updates

os-d:
## Description

This PR consists of 4 commits:

spin_locked_gcd: Mapped reserved regions
--
Map reserved regions similar to MMIO for now so they are mapped as R/W in the DXE page table. Some regions are accessed during DXE and need to be mapped to prevent a page fault.

spin_locked_gcd: Only Query First Page for Attributes
--
When deciding whether to map or remap a region, the GCD today queries the entire region to decide which operation to do and if remapping, which caching attributes to do. However, this is a wasted operation, because if the region is unmapped, we attempt to map and if this fails, it will bubble up a failure to us which is the same as we would do when failing the query. If it is a remap operation, the paging code does a query internally to ensure it can remap and then does the remap if the range is consistent. In the case of failure, remap will fail the same as if our original query failed and if it succeeds, then the range is consistent and we only need the cache attributes from the first entry in the range to persist.

This patch updates to only query the first page and update the logging in case of a paging error. This saves an unnecessary page table walk.

spin_locked_gcd: Unmap Memory on remove_memory_space
--
When a memory space is removed from the GCD, we need to unmap the pages in the page table, because DXE Core expects we will not be using this any longer and we cannot rely on a free operation to have done that, because other memory types such as MMIO and reserved which are not allocated/freed can be removed here.

Platforms may also then try to add the memory space back later and map it, which can fail if the region is already mapped.

This patch unmaps memory spaces when remove_memory_space is called unconditionally, we expect it to fail if the memory region is already mapped, but this is handled gracefully and execution continues.

spin_locked_gcd: Set Paging Attributes Per GCD Descriptor
--
When setting memory attributes, the GCD sets all GCD attributes and then moves to do the paging attributes as a whole. This works when the entire GCD operation succeeded, because that has guaranteed the GCD has a consistent view of that range of memory, however, in the case of failure this can create a torn state where the GCD is partially updated and the page table is not updated.

To avoid this torn state, the page table is updated after every GCD descriptor update.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on Q35, SBSA, and an Intel physical system.

## Integration Instructions

N/A.

### DIFF
--- a/dxe_core/src/gcd/spin_locked_gcd.rs
+++ b/dxe_core/src/gcd/spin_locked_gcd.rs
@@ -2290,7 +2290,8 @@ impl SpinLockedGcd {
         // this API allows for setting attributes across multiple descriptors in the GCD (assuming the capabilities
         // allow it). The lower level set_memory_space_attributes will only operate on a single entry in the GCD/page
         // table, so at this level we need to check to see if the range spans multiple entries and if so, we need to
-        // split the range and call set_memory_space_attributes for each entry.
+        // split the range and call set_memory_space_attributes for each entry. We also need to set the paging
+        // attributes per entry to ensure that we keep the GCD and page table in sync
 
         let mut current_base = base_address as u64;
         let mut res = Ok(());
@@ -2317,14 +2318,39 @@ impl SpinLockedGcd {
                 }
                 Ok(()) => {}
                 _ => {
-                    // some nicer log here
+                    log::error!(
+                        "Failed to set GCD memory attributes for memory region {:#x?} of length {:#x?} with attributes {:#x?}",
+                        current_base,
+                        current_len,
+                        attributes
+                    );
                     debug_assert!(false);
                 }
             }
+
+            match self.set_paging_attributes(current_base as usize, current_len as usize, attributes) {
+                Ok(_) => {}
+                Err(EfiError::NotReady) => {
+                    // before the page table is installed, we expect to get a return of NotReady. This means the GCD
+                    // has been updated with the attributes, but the page table is not installed yet. In init_paging, the
+                    // page table will be updated with the current state of the GCD. The code that calls into this expects
+                    // NotReady to be returned, so we must catch that error and report it. However, we also need to
+                    // make sure any attribute updates across descriptors update the full range and not error out here.
+                    res = Err(EfiError::NotReady);
+                }
+                _ => {
+                    log::error!(
+                        "Failed to set page table memory attributes for memory region {:#x?} of length {:#x?} with attributes {:#x?}",
+                        current_base,
+                        current_len,
+                        attributes
+                    );
+                    debug_assert!(false);
+                }
+            }
+
             current_base = next_base;
         }
-
-        self.set_paging_attributes(base_address, len, attributes)?;
 
         // if we made it out of the loop, we set the attributes correctly and should call the memory change callback,
         // if there is one


dxe_core/src/gcd/spin_locked_gcd.rs
```diff
@@ -2280,15 +2333,40 @@ impl SpinLockedGcd {
...
+                    log::error!(
+                        "Failed to set page table memory attributes for memory region {:#x?} of length {:#x?} with attributes {:#x?}",
+                        current_base,
+                        current_len,
+                        attributes
+                    );
                     debug_assert!(false);
                 }
             }
+
```
os-d, line 2351: I will remove this

os-d:
This will be updated to pull in the new uefi-core and paging crates once those PRs complete

dxe_core/src/lib.rs
```diff
@@ -339,7 +361,7 @@ where
         let relocated_c_hob_list = hob_list_slice.to_vec().into_boxed_slice();
 
         // Initialize the debugger if it is enabled.
-        uefi_debugger::initialize(&mut self.interrupt_manager);
+        uefi_debugger::initialize(&mut self.interrupt_manager, Some(env_cmd));
```
Javagedes: This interface needs to be re-evaluated. It is not extensible to adding additional custom commands. I think the interface should be something along the lines of this:



``` rust
struct Cmd {
    cmd: &'static str,
    runner: fn(&mut core::str::SplitWhitespace<'_>, &mut Vec<String>) -> core::result::Result<(), ()>
}

pub fn initialize<I>(im: &mut dyn InterruptManager, cmds: I) -> Result<usize>
where
    I: IntoIterator<Item = Cmd>,
    I::IntoIter: ExactSizeIterator,
{
    todo!()
}
```

And then whenever you get a command you don't recognize, just iterate through the registered commands and check if the cmd matches tuple.0, and if so, execute tuple.1.

This allows the interface to look like this:

```
uefi_debugger::initialize(&mut self.interrupt_manager, [])
uefi_debugger::initialize(&mut self.interrupt_manager, [cmd1])
uefi_debugger::initialize(&mut self.interrupt_manager, [cmd1, cmd2])
```

dxe_core/src/lib.rs
```diff
@@ -148,6 +148,28 @@ pub struct Alloc;
 /// A zero-sized type to gate non-allocation functions in the [Core].
 pub struct NoAlloc;
 
+/// Environment specific debugger monitor command handler. When the debugger is sent a monitor cmd beginning with "env",
+/// it will forward the rest of the tokens to this functions for handling. Currently, the only supported action is
+/// to dump the GCD. We write each line to a separate element in the Vec because the GCD protocol only allows for
+/// 0x1000 bytes to be written at a time. And the GCD is often larger than that, so the debugger will need to split
+/// this up into multiple writes.
+fn env_cmd(cmd: &mut core::str::SplitWhitespace<'_>, response: &mut Vec<String>) -> core::result::Result<(), ()> {
```
Javagedes: It's starting to seem like we need to move debugger usage (including the static initialization) into the core. This is now adding code into the core to support functionality not associated with the core, so that we can leak internal data.

This would take a bit of rework of the debugger to store the selected communication object as a dyn trait object so that it does not need to be statically compiled, but I think it would be best to move all debugger logic into the core and expose to configuration through the core.

dxe_core/src/gcd.rs
```diff
@@ -267,6 +268,10 @@ fn remove_range_overlap<T: PartialOrd + Copy>(a: &Range<T>, b: &Range<T>) -> [Op
     }
 }
 
+pub(crate) fn write_gcd(response: &mut Vec<String>) {
+    spin_locked_gcd::write_gcd(response);
```
Javagedes: Why have this wrapper? Just add a `to_string()` method or something function directly to the SpinLockedGcd struct, and in your code below just call `GCD.to_string()` (or whatever more appropriate name you can think of).

dxe_core/src/lib.rs
Javagedes: I see the PR for this change has not been merged into uefi-core yet. I made a comment there too.

cfernald: I think the idea here is that additional commands the responsibility of the callback to decode. That is the string that the debugger would originally receive is `env foo bar`, the debugger will tokenize this can parse the first value "env" which is an indicator to call the callback to do the rest. It's then up to that callback to recognize "foo" or any other value as a any number of supported commands.

Javagedes: My method allows platforms to implement their own commands without losing the capabilities produced by the car that cannot be replicated due to being private internal data. It also allows us to split up commands into more manageable subcommands rather than one single, huge match statement as we (or platforms) use this interface to implement their own debugger capabilities.

Plus its one less subcommand they have to do. They can just do `foo bar` or any other command they come up with.

Plus there is no allocations in what I provided so it should be no issue.

dxe_core/src/gcd.rs
os-d: I originally did not have the wrapper but this required making GCD public and spinlockedgcd.memory public, so I figured the wrapper was better

dxe_core/src/lib.rs
os-d: I do not think the debugger should be moved into the core. It is purposefully outside of the core as the logic is reusable across environments, so moving it to dxe core would hurt that extensibikity (which is planned to be used). Dxe core is already using the debugger as is, this just extends the functionality. It is the same with other crates we pull in. I’m fine to rework the interface, but I don’t think we should sacrifice the reusability of the debugger.

dxe_core/src/lib.rs
os-d: I don’t have a preference either way, I think the struct approach is fine, a bit much for the two people (Chris and I) who are using the debugger :).

That being said, I’m not sure it actually could be used by a platform to extend, you would have to work out an interface to talk to components and pipe the config through. And that feels much farther away (since we have one component), so I’m not sure how much it makes sense to build it out for a use case we don’t have yet.

But as I said, I think this approach is okay, too, and does allow for more flexibility (I’m just not sure flexibility we’ll ever use).

dxe_core/src/lib.rs
Javagedes: I don't think the code for the debugger should live in the core. I agree it is re-usable and we need to make sure it stays that way.

But as you said, the debugger is already becoming deeply integrated throughout different parts of the dxe core. So I think the runtime management of the debugger should live in the core (e.g. the platform will no longer be responsible 
for instantiating or initializing the global debugger. Instead, we provide toggles from the core's interface for doing whatever configuration is necessary.

Basically I want to move these things outside of the platform and into the core: [q35_dxe_core](https://dev.azure.com/microsoft/MsUEFI/_git/qemu_rust_bins?path=/bin/q35_dxe_core.rs&version=GBmain&line=48&lineEnd=49&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents)

```rust 
static DEBUGGER: uefi_debugger::UefiDebugger<uefi_sdk::serial::Uart16550> =
    uefi_debugger::UefiDebugger::new(uefi_sdk::serial::Uart16550::new(uefi_sdk::serial::Interface::Io(0x3F8)))
        .with_default_config(false, true, 0)
        .with_debugger_logging();
        
 // And inside main

uefi_debugger::set_debugger(&DEBUGGER);
```

os-d: Oh, I see what you are saying. I think that is fine and makes sense to me, thought the platform does need to be able to configure whether to enable the debugger, etc., so that would have to be piped through. But as this gets extended to StMM or other cores it may be cumbersome to do in the platform (though again the platform needs some config control here). But I’ll let @cfernald speak to this

dxe_core/src/lib.rs
Javagedes: Along the lines of my above statement, I was thinking we could take runtime management of the debugger and provide an interface for a platform to register a debugger commands to do interesting things with their code, drivers or something.

```rust
Core::default()
    .with_debugger([]) // Turns on the debugger but with no extra callbacks
    .with_debugger([cmd1]) // Turns on the debugger and registers some extra platform callback
```

If I'm completely out of sorts and there would never be a platform need to this, then maybe we are over engineering this. But I'm still not a huge fan of only being able to register a single extra command like this and I think there needs to be something a bit better. 

Sure we only added one extra command now but there could eventually be more. Maybe we make a command to work with the adv logger through the debugger / dumping contents. Same with drivers that provide services. Or maybe we want to dump the protocol db at runtime, or see event db contents or something else.

Just seems like being able to provide an interface for making custom debugger commands be as simple as writing a function and registering it would be incredibly useful to the normal everyday dev. As in hey I'm a dev and I hit this weird issue. Let me quickly write a function that adds a command and dumps what I'm looking for. I then set a breakpoint, run my platform until I hit the breakpoint, then run my new debugger command and dump the thing I care about, or manipulate it in a certain way depending on the command.

dxe_core/src/lib.rs
cfernald: I'm not opposed to this. Because of it's odd use of early initialization, needing to be static (global or leaked, for interrupt management) and requiring generics i'm not entirely sure what a good way to do it would be without being a bit hacky. This is the reason that static made sense at the time, especially since it can be used before memory allocations in some cases and needs to be 'static. @Javagedes , let me know if you have thoughts on this.

That said, the interface woudl still look largely the same, expect maybe the callbacks for env specifics could be a initialization based association instead of a registered callback.

cfernald: I'll be honest that I also doubt the re-usability model of things like the debugger though, because we rely so heavily on our SDK and CPU implementations, i'd be curious to see if we think things like this can be reasonably used outside of this "kernel", whether its being deployed for uefi or otherwise.

dxe_core/src/lib.rs
cfernald: Yeah, im not opposed to it, but also not sure i want to encourage too much expansion here because I'd want to avoid too much code in the core explicitly for use by the debugger. I'm fine with it either way.

dxe_core/src/lib.rs
Javagedes: I think the static would live in the core, but instead of it being generic over an SerialIO, we make it hold a dyn trait object - e.g. Option<dyn SerialO> and set the IO immediately on Dxe Start.

There probably is something I'm not thinking about that makes this harder than I thought, though.

os-d: Sure, but in our other kernels that may come up (StMM, secure partitions, and PEI) it could be reused

dxe_core/src/lib.rs
Javagedes: Yeah that is a good point.

dxe_core/src/lib.rs
cfernald: @Javagedes, is the DxeCore actually static? Isn't is a stack variable in the platform code?

@os-d , I guess i'm envisioning a world where all use cases have the same generic "kernel" to bring together standard kernel functionality, but that is to be seen i suppose.

Javagedes: No, the DxeCore is not static, however it does technically live for the necessary lifetime. I more meant we would just have a static in the core (like the GCD), though thinking about it, just having it be attribute of the DxeCore struct would probably suffice, though that would probably require extensive changes depending on how much `&'static Debugger`'s you are passing around functions interfaces.

cfernald: The 'static is unfortunately a hard requirement. This is because it gets passed into the interrupt manager and the interrupt manager requires this because when handling an interrupt all normal lifetimes are lost and you are starting from scratch (and initializing only from global data) and so anything that does any such handling must be static.

It could be a static in the file, but since it's generic it gets complicated. I think what you would have to do it box::leak it, but then we lose the ability to start the debugger before allocations.

Javagedes: I don't think that is too big of an issue. Just follow what the log crate does. In a no_std environment you just pass it a static logger (so for us a static WriteIO), and if you have alloc available, you pass it an instance and we leak it ourselves.

cfernald: @Javagedes, are you suggesting that as a new approach or confirming the current one? Currently the debugger implementation is more or less based on the logger where static routines are used to access a secret static dyn reference.

Javagedes: This PR issue can be closed, and we should add an actual issue to investigate moving the DEBUGGER static (not the source code) into the uefi-dxe-core.

Regarding your question, behind the scenes, the log crate has a single static that is a dyn Log that is originally set to a do nothing logger. Then there are two ways to set that logger - the first is to directly provide a static lifetime logger, which it will directly set to that static dyn log object. The second is to pass in a logger, which will be boxed and leaked and then set to the static dyn log object.

I was just suggestion we could do something similar, but regarding the SerialIO implementation, not the debugger as a whole.

dxe_core/src/lib.rs
os-d: Discussed offline, investigating moving to ext.

dxe_core/src/gcd/spin_locked_gcd.rs
os-d: eh, now I like the extra line there.

dxe_core/src/gcd/spin_locked_gcd.rs
```diff
@@ -1760,7 +1765,15 @@ impl SpinLockedGcd {
...
-            match page_table.query_memory_region(base_address as u64, len as u64) {
+
+            // we assume that the page table and GCD are in sync. If not, we will debug_assert and return an error here
+            // as such, we only need to query the first page of this region to get the attributes. This will tell us
+            // whether the region is mapped or not and if so, what cache attributes to persist.
+            // If the first page is unmapped, we will call map_memory_region to map it. If it finds a page that is
+            // mapped inside of there, it will fail and we will debug_assert and return an error.
+            // If the first page is mapped, we will call remap_memory_region to remap it. It queries the range already
+            // so will catch the rest of the range and return an error if it is inconsistently mapped.
+            match page_table.query_memory_region(base_address as u64, UEFI_PAGE_SIZE as u64) {
```
joschock, line 1776: do we need a check on base_address alignment to UEFI_PAGE_SIZE here?

dxe_core/src/gcd/spin_locked_gcd.rs
```diff
@@ -2278,15 +2318,40 @@ impl SpinLockedGcd {
...
+                        attributes
+                    );
+                    debug_assert!(false);
+                }
+            }
+
+            match self.set_paging_attributes(current_base as usize, current_len as usize, attributes) {
+                Ok(_) => {}
+                Err(EfiError::NotReady) => {
+                    // before the page table is installed, we expect to get a return of NotInitialized. This means the GCD
```
joschock:
```suggestion
                    // before the page table is installed, we expect to get a return of `NotReady`. This means the GCD
```

joschock: and below in same comment.

dxe_core/src/gcd/spin_locked_gcd.rs
os-d: I don't think so, the paging crate validates that: https://github.com/OpenDevicePartnership/paging/blob/20a52c0570f9405b218302642939edeb337da442/src/x64/paging.rs#L679-L702 and we would fail and hit the debug_assert/return invalid parameter, so still failing at this level.

This is also a private function only called in one spot, being passed GCD memory descriptor base_addresses (which had better be page aligned...), so I think letting the GCD validation and paging validation happen makes sense, we'll still fail here and not at a random spot.

PR #312, opened 2025-03-24
Add cargo-deny [Rebase & FF]

makubacki:
## Description

Adds a new cargo tool to the dev cycle:
https://github.com/EmbarkStudios/cargo-deny

Can be installed with `cargo install cargo-deny --version "^0.17"`.
(latest `cargo-deny` build (0.18) requires Rust 1.85+)

Adds a `deny.toml` config file to set up configuration for the repo.

Main items in the config file:

- Allowed licenses (SPDX identifiers):
  - Apache-2.0,
  - BSD-2-Clause-Patent
  - MIT

- Licenses enforced to the above list for all dependencies with the following exceptions that will be followed up on:
  - `managed` -> Using `0BSD` license
  - `unicode-ident` -> Using `Unicode-3.0` license

- Rust security advisories against all dependencies enforcee with the following current exception:
  - `RUSTSEC-2024-0436` (https://rustsec.org/advisories/RUSTSEC-2024-0436.html)

- `tiny-keccak` crate explicitly added to the deny list since we have already identified it is old, unmaintained, and not secure via component governance.

- Redudant crate check enforced for all crates in the graph wiith the following exception:
  - `bitflags`. Most crates have moved to the newer version, but `x86` is still using the older version causing the duplication. Following up with the `rust-x86` project to accept an outstanding dependabot PR that will update their version.

The exact cargo-deny command used in the repo can also be run with `cargo make deny`.

The CI workflow file will now run `cargo-deny` with default settings which runs the `check` command.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- `cargo deny check`
- `cargo make deny`

## Integration Instructions

- N/A - Affects local repo build and CI

### DIFF
--- a/docs/src/SUMMARY.md
+++ b/docs/src/SUMMARY.md
@@ -9,6 +9,7 @@
   - [Abstractions](dev/principles/abstractions.md)
   - [Code Reuse](dev/principles/reuse.md)
   - [Configuration in Code](dev/principles/config.md)
+  - [Dependency Management](dev/principles/dependency-management.md)
   - [Error Handling](dev/principles/error-handling.md)
 - [Testing](dev/testing.md)
   - [Unit Testing](dev/testing/unit.md)
--- a/docs/src/dev/principles/dependency-management.md
+++ b/docs/src/dev/principles/dependency-management.md
@@ -1,0 +1,27 @@
+# Dependency Management
+
+The Rust DXE Core is designed to be a monolithic binary, meaning that diverse sets of Rust functionality and the core
+are compiled together. This allows for more checks to directly be performed against the overall dependency graph that
+composes that Rust DXE Core but also leads to a relatively larger number of dependencies in that graph. This document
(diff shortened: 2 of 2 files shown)


makubacki:
I've tested everything except some adjustments that might be needed for the workflow. Doing that on my fork now.

makubacki:
> I've tested everything except some adjustments that might be needed for the workflow. Doing that on my fork now.

Done. Main change made was to use the cargo tool cache mechanism to install cargo deny instead of the action. Has several advantages including eliminating a new action dependency making CI more predictable, simpler to understand CI workflow, faster, and explicit tracking of the tool in rust-toolchain.toml with the version currently being used.

makubacki:
That should be the final push. Added a doc file and updated a commit message most recently.

PR #315, opened 2025-03-24
[REBASE&FF] Add const initialization support for Rbt and Bst

Javagedes:
## Description

Adds const initialization support for the Rbt and Bst, which allows the GCD to no longer need to wrap them in `Option`, which leads to a lot of boilerplate in the GCD that is able to be removed.

This does not change any logic in the GCD itself.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Continued boot to UEFI Shell.

## Integration Instructions

N/A

### DIFF
--- a/dxe_core/src/gcd/spin_locked_gcd.rs
+++ b/dxe_core/src/gcd/spin_locked_gcd.rs
@@ -276,7 +276,7 @@ impl PageAllocator for PagingAllocator<'_> {
 //The Global Coherency Domain (GCD) Services are used to manage the memory resources visible to the boot processor.
 struct GCD {
     maximum_address: usize,
-    memory_blocks: Option<Rbt<'static, MemoryBlock>>,
+    memory_blocks: Rbt<'static, MemoryBlock>,
     allocate_memory_space_fn: GcdAllocateFn,
     free_memory_space_fn: GcdFreeFn,
 }
@@ -296,7 +296,7 @@ impl GCD {
     pub(crate) const fn new(processor_address_bits: u32) -> Self {
         assert!(processor_address_bits > 0);
         Self {
-            memory_blocks: None,
+            memory_blocks: Rbt::new(),
             maximum_address: 1 << processor_address_bits,
             allocate_memory_space_fn: Self::allocate_memory_space_internal,
             free_memory_space_fn: Self::free_memory_space_worker,
@@ -343,11 +343,10 @@ impl GCD {
             ..Default::default()
         });
 
-        let mut memory_blocks =
-            Rbt::new(slice::from_raw_parts_mut::<'static>(base_address as *mut u8, MEMORY_BLOCK_SLICE_SIZE));
-        memory_blocks.add(unallocated_memory_space).map_err(|_| EfiError::OutOfResources)?;
-        self.memory_blocks.replace(memory_blocks);
+        self.memory_blocks
+            .resize(slice::from_raw_parts_mut::<'static>(base_address as *mut u8, MEMORY_BLOCK_SLICE_SIZE));
 
+        self.memory_blocks.add(unallocated_memory_space).map_err(|_| EfiError::OutOfResources)?;
         let idx = self.add_memory_space(memory_type, base_address, len, capabilities)?;
 
         //initialize attributes on the first block to WB + XP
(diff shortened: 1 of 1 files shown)


dxe_core/src/gcd/spin_locked_gcd.rs
```diff
@@ -343,11 +343,10 @@ impl GCD {
             ..Default::default()
         });
 
-        let mut memory_blocks =
-            Rbt::new(slice::from_raw_parts_mut::<'static>(base_address as *mut u8, MEMORY_BLOCK_SLICE_SIZE));
-        memory_blocks.add(unallocated_memory_space).map_err(|_| EfiError::OutOfResources)?;
-        self.memory_blocks.replace(memory_blocks);
+        self.memory_blocks
+            .resize(slice::from_raw_parts_mut::<'static>(base_address as *mut u8, MEMORY_BLOCK_SLICE_SIZE));
```
joschock, line 347: is this .resize guaranteed not to require an allocation out of the GCD itself? I don't think that the GCD is functional until the first block is added. I'll check the implementation of .resize when I get there.

joschock: If I understand the `resize` implementations, they carefully avoid access to non-present prior buffer or doing things that would cause an implicit heap allocation. I think it's worth a comment on `node::Storage::resize()` that it must be careful to avoid any heap allocations if `self.capacity()` is zero; will put a comment on there as well.

crates/uefi_collections/src/node.rs
```diff
@@ -148,6 +167,79 @@ where
...
+            slice::from_raw_parts_mut::<'a, Node<D>>(
+                slice as *mut [u8] as *mut Node<D>,
+                slice.len() / mem::size_of::<Node<D>>(),
+            )
+        };
+
+        assert!(buffer.len() >= self.len());
+
+        // When current capacity is 0, we just need to copy the data and build the available list
+        if self.capacity() == 0 {
```
joschock, line 194: I think a comment here to note that we must be cautious to avoid any heap allocations in the self.capacity() == 0 case, since there may be no allocator present in that scenario.

I think the whole implementation as it presently sits doesn't do any heap allocations, but I think it would be good to explicitly note the need to avoid them.

Javagedes: The uefi_collections data structure(s) are generic no-alloc data structures backed by a byte buffer passed to the struct during initialization. It is intended to be generic and usable for things other than the GCD. Due to this, I don't think the requested comment makes sense here.

Looking at this PR from a GCD perspective, I do agree with your sentiments. If we ever want to add functionality to make the starting size of the GCD smaller than 4K entries, and grow it as needed, we will need to call `resize()` at least one count before max. However this PR does not do that. It just basically replaces the Option.swap() with Rbt::resize() and allows us to no longer have to unwrap the Option whenever we do something with the gcd.

Looking at this PR from purely a `uefi_collections` crate perspective, all I've done is add the ability to replace the current (smaller) backing buffer with a new (larger) backing buffer, while keeping the existing tree in-tact (which is harder than just a clone because I have to update all pointers to the new backing buffer). Where that new backing buffer comes from is outside of the scope of the Rbt / Bst.

dxe_core/src/gcd/spin_locked_gcd.rs
Javagedes: These collections do not do any allocations ever. There is one exception which is hidden behind an `alloc` feature, which just does a DFS to return a copy of the contents as an ordered Vec.

resize just copies the contents from the old buffer into the new buffer and updates the pointers to point at the new buffer. I don't think it is necessary to put a comment regarding allocations because this crate is `no_alloc` in general and does not have any other crate dependencies, so implicit heap allocations cannot sneak in. 

I can add some additional documentation to the crate level documentation that specifically says this is a no_std and no_alloc RBT, BST, SS implementation if you would like.

Javagedes: Added additional documentation to the crate level documentation stating that this is a no_std and no_alloc crate.

crates/uefi_collections/src/node.rs
joschock: Ok - yeah, I'm convinced. I wasn't thinking about this from the POV that this was already a no_alloc crate. I'm fine with no extra comment here.

PR #320, opened 2025-03-27
Initial Memory Manager TOPs documentation

joschock:
## Description

Add Memory Manager Theory of Operations documentation.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [x] Includes documentation?

## How This Was Tested

Confirmed pages render as expected in mdbook; checked links manually.

## Integration Instructions

N/A


docs/src/dxe_core/memory_management.md
```diff
@@ -1,5 +1,258 @@
 # Memory Management
 
+This portion of the core is responsible for producing the capabilies described in [Section 7.2](https://uefi.org/specs/UEFI/2.10_A/07_Services_Boot_Services.html#memory-allocation-services)
```
os-d:
```suggestion
This portion of the core is responsible for producing the capabilities described in [Section 7.2](https://uefi.org/specs/UEFI/2.10_A/07_Services_Boot_Services.html#memory-allocation-services)
```

docs/src/dxe_core/memory_management.md
```diff
@@ -1,5 +1,258 @@
...
+large enough initial allocation such that all memory requests of that type can be satisfied during boot without a
+further call to the GCD for more memory, then all the memory of that type will be reported in a single contiguous block
+to the OS that is stable from boot-to-boot. This facility is important for enabling certain use cases (such as
+hibernate) where the OS assumes a stable boot-to-boot memory map.
+
+### UefiAllocator Operations
+
+The UefiAllocator supports the following operations:
+
+* Creating a new allocator for arbitrary memory types. A subset of well-known allocators are are provided by the core
```
os-d:
```suggestion
* Creating a new allocator for arbitrary memory types. A subset of well-known allocators are provided by the core
```

docs/src/dxe_core/memory_management.md
```diff
@@ -1,5 +1,258 @@
...
+to support UEFI spec standard memory types, but the spec also allows for arbitrary OEM-defined memory types. If a caller
+makes an allocation request to a previously unused OEM-defined memory type, a new allocator instance is dynamically
+instantiated to track memory for the new memory type.
+* Retrieving the `EfiMemoryType` associated with the allocator. All allocations done with this allocator instance will
+be of this type.
+* Reserving pages for the allocator. This is used to see the allocator with an initial [bucket](memory_management.md#allocation-buckets)
```
os-d:
```suggestion
* Reserving pages for the allocator. This is used to seed the allocator with an initial [bucket](memory_management.md#allocation-buckets)
```

docs/src/dxe_core/memory_management.md
```diff
@@ -1,5 +1,258 @@
...
+of memory.
+* Ensuring that the allocator has capacity for satisfying a request (or set of requests) of a given size. This allows a
```
os-d, line 109: I think we actually don't use the `ensure_capacity` API any longer, perhaps we should get rid of it? I agree with the description of priming the allocator if needed, but I wonder if we will come up with use cases for this.

docs/src/dxe_core/memory_management.md
```diff
@@ -1,5 +1,258 @@
...
+
+The GCD supports the following operations:
+
+* Adding, Removing, and Allocating and Freeing regions within the address space. The semantics for these operations
+largely follow the Platform Initialization spec [APIs](https://uefi.org/specs/PI/1.9/V2_Services_DXE_Services.html#gcd-memory-resources)
+for manipulating address spaces.
+* Configuring the capabilities and attributes of the memory space. The GCD uses CPU memory management hardware to
+enforce these attributes where supported. See [Paging](cpu.md#paging-implementation) for details on how this hardware is
+configured.
+* Retriving the current address space map as a list of descriptors containing details about each memory region.
```
os-d:
```suggestion
* Retrieving the current address space map as a list of descriptors containing details about each memory region.
```

docs/src/dxe_core/memory_management.md
```diff
@@ -1,5 +1,258 @@
...
+  space
+  single_large_block["Single Large Block (Characteristics = A)"]:3
+```
+
+## Concurrency
+
+UefiAllocator and GCD operations require taking a lock on the associated data structure to prevent concurrent
+modifications to internal allocation tracking structures by operations taking place at different TPL levels (for
+example, allocating memory in an event callback that interrupts a lower TPL). This is accomplished by using a
+[`TplMutex`](synchronization.md#tplmutex) that switches to the highest TPL level (uninterrptible) before executing the
```
os-d:
```suggestion
[`TplMutex`](synchronization.md#tplmutex) that switches to the highest TPL level (uninterruptible) before executing the
```

docs/src/dxe_core/memory_management.md
```diff
@@ -1,5 +1,258 @@
...
+
+## Rust `Allocator` and `GlobalAlloc` Implementations
+
+In addition to producing the memory allocation APIs required by the UEFI spec, the memory allocation subsystem also
+produces implementations of the [`Allocator`](https://doc.rust-lang.org/std/alloc/trait.Allocator.html) and
+[`GloballAlloc`](https://doc.rust-lang.org/std/alloc/trait.GlobalAlloc.html) traits.
+
+These implementations are used within the core for two purposes:
+
+1. The `GlobalAlloc` implementation allows one of the `UefiAllocator` instances to be desginated as the Rust Global
```
os-d:
```suggestion
1. The `GlobalAlloc` implementation allows one of the `UefiAllocator` instances to be designated as the Rust Global
```

docs/src/dxe_core/memory_management.md
joschock: weird - ran this through cspell and fixed a bunch of stuff already; strange that it was fine with this. Will resolve, thanks for catching.

docs/src/dxe_core/memory_management.md
joschock: We could pull it out, really the main use case would be to avoid thrashing GCD, but if you are concerned about that, go tune the buckets. For now, I'll leave the docs in until we decide to actually remove the capability.

Issue #324, opened 2025-03-27
[Bug]: Panic- Failed to parse PE info for DXE Core: BadSignature(0)


HappyCoding-2025:
### Is there an existing issue for this?

- [x] I have searched existing issues

### Current Behavior

Panics during lib module coverage.

---- allocator::tests::lib_module_coverage_through_allocator_module stdout ----
Coverage for initializing Cpu!
Coverage for initializing Interrupt Manager!
Found MemoryAllocationModule with module_name: Guid { time_low: [47, 50, 201, 35], time_mid: [242, 42], time_hi_and_version: [106, 71], clk_seq_hi_res: 188, clk_seq_low: 76, node: [38, 188, 136, 38, 108, 113] }
thread 'allocator::tests::lib_module_coverage_through_allocator_module' panicked at dxe_core\src\image.rs:533:10:
Failed to parse PE info for DXE Core: BadSignature(0)
stack backtrace:
   0:     0x7ff720fac521 - std::backtrace_rs::backtrace::dbghelp64::trace
                               at /rustc/9fc6b43126469e3858e2fe86cafb4f0fd5068869\library/std\src\..\..\backtrace\src\backtrace\dbghelp64.rs:91
   1:     0x7ff720fac521 - std::backtrace_rs::backtrace::trace_unsynchronized
                               at /rustc/9fc6b43126469e3858e2fe86cafb4f0fd5068869\library/std\src\..\..\backtrace\src\backtrace\mod.rs:66
   2:     0x7ff720fac521 - std::sys::backtrace::_print_fmt
                               at /rustc/9fc6b43126469e3858e2fe86cafb4f0fd5068869\library/std\src\sys\backtrace.rs:66
   3:     0x7ff720fac521 - std::sys::backtrace::impl$0::print::impl$0::fmt
                               at /rustc/9fc6b43126469e3858e2fe86cafb4f0fd5068869\library/std\src\sys\backtrace.rs:39
   4:     0x7ff720fca91a - core::fmt::rt::Argument::fmt
                               at /rustc/9fc6b43126469e3858e2fe86cafb4f0fd5068869\library/core\src\fmt\rt.rs:177
   5:     0x7ff720fca91a - core::fmt::write
                               at /rustc/9fc6b43126469e3858e2fe86cafb4f0fd5068869\library/core\src\fmt\mod.rs:1189
[...]

Issue #322, opened 2025-03-27
[Bug]: Panic - panicked at dxe_core\src\lib.rs:511:39: initial dispatch failed.: NotFound


HappyCoding-2025:
### Is there an existing issue for this?

- [x] I have searched existing issues

### Current Behavior

Panics -
Lib.rs module initiatlises the entire dxe-core . It calls all internal functions.  Passing a simple hoblist crashes below.

---- allocator::tests::lib_module_coverage_through_allocator_module stdout ----
Coverage for initializing Cpu!
Coverage for initializing Interrupt Manager!
thread 'allocator::tests::lib_module_coverage_through_allocator_module' panicked at dxe_core\src\lib.rs:511:39:
initial dispatch failed.: NotFound
stack backtrace:
   0:     0x7ff6c3a7b7a1 - std::backtrace_rs::backtrace::dbghelp64::trace
                               at /rustc/9fc6b43126469e3858e2fe86cafb4f0fd5068869\library/std\src\..\..\backtrace\src\backtrace\dbghelp64.rs:91
   1:     0x7ff6c3a7b7a1 - std::backtrace_rs::backtrace::trace_unsynchronized
                               at /rustc/9fc6b43126469e3858e2fe86cafb4f0fd5068869\library/std\src\..\..\backtrace\src\backtrace\mod.rs:66
   2:     0x7ff6c3a7b7a1 - std::sys::backtrace::_print_fmt
                               at /rustc/9fc6b43126469e3858e2fe86cafb4f0fd5068869\library/std\src\sys\backtrace.rs:66
   3:     0x7ff6c3a7b7a1 - std::sys::backtrace::impl$0::print::impl$0::fmt
                               at /rustc/9fc6b43126469e3858e2fe86cafb4f0fd5068869\library/std\src\sys\backtrace.rs:39
   4:     0x7ff6c3a99b9a - core::fmt::rt::Argument::fmt
                               at /rustc/9fc6b43126469e3858e2fe86cafb4f0fd5068869\library/core\src\fmt\rt.rs:177
   5:     0x7ff6c3a99b9a - core::fmt::write
                               at /rustc/9fc6b43126469e3858e2fe86cafb4f0fd5068869\library/core\src\fmt\mod.rs:1189
   6:     0x7ff6c3a77927 - std::io::Write::write_fmt<alloc::vec::Vec<u8,alloc::alloc::Global> >
[...]

PR #326, opened 2025-03-27
 Code Coverage for Lib module and fix logic and error handling in other modules

HappyCoding-2025:
Support for Code Coverage for lib module and coverage is coming to around 85% with these changes.

Since the lib module inits the entire dxe-core,  it touches almost all the code!! and encountered more than several issues.

Supported adding a new hoblist, the existing hoblist build does not contain the memoryallocation module required for dxe-core.
Added a compact hoblist function to support these.

Error handling in other modules for handling panics.

Fixed the lib module system for  boot_services_ptr by safe casting and moving logic up as it was causing undefined behaviour during coverage causing it holding locks causing other coverage modules to crash. Now this is fixed.

THe following issues were raised and fixed as part of this PR -

https://github.com/OpenDevicePartnership/uefi-dxe-core/issues/251
[Feature]: Develop Unit Tests for Lib.Rs module. Currently it is 0% 

Added support for Hob 
https://github.com/OpenDevicePartnership/uefi-dxe-core/issues/323
Did not find MemoryAllocationModule Hob for DxeCore. Use uefi_sdk::guid::DXE_CORE as FFS GUID. stack

https://github.com/OpenDevicePartnership/uefi-dxe-core/issues/325
TPL lock issue panic - lock held and unclean/undefined behaviour causing other modules to crash

https://github.com/OpenDevicePartnership/uefi-dxe-core/issues/324
Failed to parse PE info for DXE Core: BadSignature(0) #324

https://github.com/OpenDevicePartnership/uefi-dxe-core/issues/322
Panic - panicked at dxe_core\src\lib.rs:511:39: initial dispatch failed.: NotFound #322

https://github.com/OpenDevicePartnership/uefi-dxe-core/issues/321
[Bug]: <title>Panic - panicked at dxe_core\src\gcd\spin_locked_gcd.rs:191

Tested with code coverage

### DIFF
--- a/dxe_core/src/lib.rs
+++ b/dxe_core/src/lib.rs
@@ -11,6 +11,7 @@
 //! use uefi_cpu::interrupts::InterruptBases;
 //! use uefi_cpu::interrupts::ExceptionType;
 //! use uefi_cpu::interrupts::HandlerType;
+//! use uefi_sdk::error::EfiError;
 //! # fn example_component() -> uefi_sdk::error::Result<()> { Ok(()) }
 //! # #[derive(Default, Clone, Copy)]
 //! # struct CpuInitExample;


dxe_core/src/allocator.rs
```diff
@@ -991,6 +991,13 @@ pub fn init_memory_support(hob_list: &HobList) {
     }
 }
 
+#[cfg(test)]
+pub fn allocator_lock_reset() {
+    unsafe {
+        ALLOCATORS.lock().reset();
```
joschock: Is this to handle the scenario where a test fails and panics while holding the lock? This should not be needed in a passing case as the lock should always be released in normal scenarios.

dxe_core/src/allocator.rs
```diff
@@ -1025,6 +1037,105 @@ mod tests {
...
+                PROTOCOL_DB.reset();
+                ALLOCATORS.lock().reset();
+            }
+            f();
+        })
+        .unwrap();
+    }
+
+    #[test]
+    fn lib_module_coverage_through_allocator_module() {
```
joschock, line 1048: Can this be moved back to the appropriate module (lib.rs)?

dxe_core/src/gcd/spin_locked_gcd.rs
```diff
@@ -1914,7 +1914,13 @@ impl SpinLockedGcd {
         log::info!("Initializing paging for the GCD");
 
         let page_allocator = PagingAllocator::new(&GCD);
-        let mut page_table = create_cpu_paging(page_allocator).expect("Failed to create CPU page table");
+        let mut page_table = match create_cpu_paging(page_allocator) {
+            Ok(pt) => pt,
+            Err(e) => {
+                log::error!("Failed to create CPU page table: {:?}", e);
+                return;
```
joschock: I think this failure should panic.

dxe_core/src/image.rs
```diff
@@ -493,7 +493,14 @@ fn install_dxe_core_image(hob_list: &HobList) {
     let dxe_core_hob = hob_list
         .iter()
         .find_map(|x| match x {
-            Hob::MemoryAllocationModule(module) if module.module_name == guid::DXE_CORE => Some(module),
+            Hob::MemoryAllocationModule(module) => {
+                println!("Found MemoryAllocationModule with module_name: {:?}", module.module_name);
```
joschock: println is not appropriate here; if needed, use log:: if this print is desired.

dxe_core/src/image.rs
```diff
@@ -519,11 +526,16 @@ fn install_dxe_core_image(hob_list: &HobList) {
...
+        match UefiPeInfo::parse(core::slice::from_raw_parts(
             dxe_core_hob.alloc_descriptor.memory_base_address as *const u8,
             dxe_core_hob.alloc_descriptor.memory_length as usize,
-        ))
-        .expect("Failed to parse PE info for DXE Core")
+        )) {
+            Ok(info) => info,
+            Err(err) => {
+                log::error!("Failed to parse PE info for DXE Core: {:?}", err);
+                return;
```
joschock: I think this error should be a panic.

dxe_core/src/lib.rs
```diff
@@ -508,8 +505,9 @@ where
         log::info!("Finished Dispatching Local Drivers");
         self.display_components_not_dispatched();
 
-        dispatcher::core_dispatcher().expect("initial dispatch failed.");
-
+        if let Err(err) = dispatcher::core_dispatcher() {
+            log::error!("Initial dispatch failed: {:?}", err);
```
joschock: I think this should panic on fail.

dxe_core/src/allocator.rs
HappyCoding-2025: THanks John. @Javagedes  had requested to make this pub for other modules to use in future. Currently this function is not used in the code. 

The scenario that you asked about -
Other tests panic for a poisoned lock - is the code from lib.rs  -       let boot_services_ptr = 
this is with the fix - .
sys_table.as_mut().unwrap().boot_services_mut() as *mut efi::BootServices;
            // Use `boot_services_ptr` safely here

            tpl_lock::init_boot_services(boot_services_ptr); --> This enitre existing which was already existing was causing undefined behaviour.

I had to move this code up and also init_boot_services to null pointer when test is done to get this working.
adding    /* Reset boot Services. Fixes Panic */ in code coverage after the test cases also helped along with moving the code 
        init_boot_services(core::ptr::null_mut());

dxe_core/src/allocator.rs
HappyCoding-2025: @joschock  -  This module when called from lib.rs panics with dxe_core\src\tpl_lock.rs:74:43:
Re-entrant locks for "AllocatorMapLock" not permitted.
I spent almost a 4-5 days and night to fix this but could not. It looks like one test is also multithreading and holding locks 
I tried over 100 ways to fix this and could not.
The plan is to push these changes and let @Javagedes  fix Re-entrant locks for "AllocatorMapLock" not permitted.
 complex issue.  :)

dxe_core/src/gcd/spin_locked_gcd.rs
HappyCoding-2025: Sure, I can check why a simple hoblist passing with proper GCD reset, GCD init, PRotocol DB reset and init, system init would still fail to create CPU page table.

dxe_core/src/image.rs
HappyCoding-2025: Sure, will remove the println

dxe_core/src/image.rs
HappyCoding-2025: For Code Coverage,  the image.rs also expects a valid image to be passed with valid signature and all. Do you think it is required to create an image for code coverage purpose ? That seemed a little too much too to me. Pls advise.

dxe_core/src/lib.rs
HappyCoding-2025: From Coverage point of point, this was panicing. If the dispatching is already running from another context since coverage runs parallely which it seemed like, how do you handle this case. It is already called in the context of global locks. Pls advise

dxe_core/src/gcd/spin_locked_gcd.rs
joschock: you might need to come up with a way to mock the paging support, since I don't expect that the test on the host will be able to enable hardware paging properly.

dxe_core/src/image.rs
joschock: you can just copy a pre-built version of the DXE core image in the test resources directory and load it from a file in the test into a memory buffer and pass that? I don't think you have to build it by hand in code.

dxe_core/src/lib.rs
joschock: Well, the with global lock closure should prevent parallel execution of multiple dispatchers, so if that's the suspected problem here, then there is some issue with the global lock closure that needs to be investigated.

dxe_core/src/gcd/spin_locked_gcd.rs
HappyCoding-2025: Thanks John. I like this idea of mocking paging support. Looks like mocking infrastructure is required for several test cases where host will not be able to enable some of the hardware functionalities. 
I will fix this.

joschock: I think there is a null version of the paging crate; you might be able to use that when compiling for test; that might get you to the point where this works.

HappyCoding-2025: Thanks John !

dxe_core/src/lib.rs
HappyCoding-2025: Thanks John. Pls advise if there Is  any null version for core_dispatcher() too to use for tests ? Else I will investigate this.

joschock: No, there isn't a mock dispatcher to patch in here; would have to figure out why it is failing.

dxe_core/src/gcd/spin_locked_gcd.rs
HappyCoding-2025: @joschock  - I checked there is no stub implementation[Null version for paging crate similar to  uefi_cpu::paging::create_cpu_paging;]  Pls advise if I can add a new complete crate for this - like - uefi_cpu::paging::create_cpu_paging_stub ?

joschock: I think by default in a test build, you should be pulling in the "null" version of create_cpu_paging: https://github.com/OpenDevicePartnership/uefi-core/blob/67516235ca9a54e348145220974b00cc4a9514b7/uefi_cpu/src/paging.rs#L21

Which looks like:
```rust
pub fn create_cpu_null_paging<A: PageAllocator + 'static>(
    _page_allocator: A,
) -> Result<Box<dyn PageTable<ALLOCATOR = A>>, efi::Status> {
    Err(efi::Status::UNSUPPORTED)
}
```
I think you could consider changing that to a non-error status. Or, you could change the above match statement so that unsupported is not considered an error in test, but will panic otherwise (or if another error is returned).
```rust
let mut page_table = match create_cpu_paging(page_allocator) {
            Ok(pt) => pt,
            Err(err) => {
                log::error!("Failed to create CPU page table: {:?}", e);
                if cfg!(test) && err == efi::Status::UNSUPPORTED {
                  return; 
                } 
                panic!("Failed to create CPU page table: {:?}", e);
            }
}

Comment (HappyCoding-2025) [reply dxe_core/src/gcd/spin_locked_gcd.rs]: Thanks John. Trying.

Comment (HappyCoding-2025) [reply dxe_core/src/gcd/spin_locked_gcd.rs]: This is working. Pushed changes in PR>

File: dxe_core/src/image.rs (hunk 5 above)
Comment (HappyCoding-2025) [reply dxe_core/src/image.rs]: Added a new create to use the dxe core efi and fill up the HOB. it is working now. Resolved.

File: dxe_core/src/image.rs (hunk 4 above)
Comment (HappyCoding-2025) [reply dxe_core/src/image.rs]: Resolved.

**Comment by HappyCoding-2025** [comment]
@joschock  - thanks for your comments. I have resolved them. Pls review

@makubacki  @Javagedes

File: dxe_core/src/allocator.rs [hunk 7]
```diff
@@ -1115,25 +1110,29 @@ mod tests {
...
+                    .with_interrupt_manager(InterruptManagerForCoverage)
+                    .with_section_extractor(SectionExtractorForCoverage)
+                    .with_interrupt_bases(InterruptBasesForCoverage)
+                    .init_memory(physical_hob_list) // We can make allocations now!
+                    .with_config(sample_components::Name("Coverage "))
+                    .with_component(sample_components::log_hello)
+                    .start()
+                    .expect("Initing System Core failed ");
+
+                /* Reset boot Services. Fixes Panic */
```
joschock, line 1132: For this, can you add a bit more of a description of what is fixed? I'm not sure what the panic is (maybe TPL?), but would be good to have something like "Reset boot services to ensure that xyz doesn't happen causing a panic in this test scenario"

dxe_core/src/lib.rs
```diff
@@ -512,8 +513,13 @@ where
         log::info!("Finished Dispatching Local Drivers");
         self.display_components_not_dispatched();
 
-        dispatcher::core_dispatcher().expect("initial dispatch failed.");
-
+        if let Err(err) = dispatcher::core_dispatcher() {
+            if cfg!(test) && err == EfiError::NotFound {
```
makubacki, line 517: I'd prefer to keep test logic out of the main functional path of code. This already returns a result; couldn't it just return an error? I wouldn't expect function execution to continue anyway (even in the test scenario) in the case the dispatcher returns not found.

dxe_core/src/image.rs
```diff
@@ -493,7 +493,13 @@ fn install_dxe_core_image(hob_list: &HobList) {
...
         .iter()
         .find_map(|x| match x {
-            Hob::MemoryAllocationModule(module) if module.module_name == guid::DXE_CORE => Some(module),
+            Hob::MemoryAllocationModule(module) => {
+                if module.module_name == uefi_sdk::guid::DXE_CORE {
+                    Some(module)
+                } else {
+                    None
+                }
+            }
```
makubacki, line 502: Is this necessary?

dxe_core/src/allocator.rs
```diff
@@ -1025,6 +1032,109 @@ mod tests {
...
+                PROTOCOL_DB.reset();
+                ALLOCATORS.lock().reset();
+            }
+            f();
+        })
+        .unwrap();
+    }
+
+    #[test]
+    fn lib_module_coverage_through_allocator_module() {
```
makubacki, line 1048: This is a lot of touch points for a unit test. It is likely to be broken by code outside of the allocator. This seems more like an integration test that should be moved somewhere else.

dxe_core/src/gcd/spin_locked_gcd.rs
```diff
@@ -1914,7 +1914,16 @@ impl SpinLockedGcd {
         log::info!("Initializing paging for the GCD");
 
         let page_allocator = PagingAllocator::new(&GCD);
-        let mut page_table = create_cpu_paging(page_allocator).expect("Failed to create CPU page table");
+        let mut page_table = match create_cpu_paging(page_allocator) {
+            Ok(pt) => pt,
+            Err(e) => {
+                log::error!("Failed to create CPU page table: {:?}", e);
+                if cfg!(test) && e == efi::Status::UNSUPPORTED {
```
makubacki, line 1921: This is another place where I think the code needs to be written to avoid test expectations being embedded in the main control flow.

dxe_core/src/lib.rs
joschock: I think this is in response to my comment where I wanted to keep the panic on fail here that the original "expect" was generating.

dxe_core/src/gcd/spin_locked_gcd.rs
HappyCoding-2025: @makubacki  - Right, I agree that test expectations should be kept separate from main control flow. How do you take care of this since lot of functions are not running on real hardware and in UEFI context for any architecture and these are expected to fail during coverage.

dxe_core/src/allocator.rs
HappyCoding-2025: Sure John, I will add more comments here.

dxe_core/src/allocator.rs
HappyCoding-2025: @makubacki  - Not really, lib.rs code init itself seems lacking some of the resets which are necessary. During my testing , i did see that lib.rs init when called with out the resets were already in a stale state. gcd init is called from lib.rs with out a gcd reset, similarly for protocol_DB.  Even if these are called from other file crate, these seemed still very much neccessary with out which they crashed.
lib.rs is the most complex part as it seemed to initialize the entire dxe-core and seemed to touch every bit of code
Regarding the  ALLOCATORS.lock().reset(); yes, it seemed like the allocator lock was already held in the context of lib init function already and it errored constantly about reentrant locks when this logic was in lib.rs module test.

dxe_core/src/image.rs
HappyCoding-2025: @makubacki  - RIght, this might not be necessary after John's comments are addressed.  I am filling up a hob with DX_CORE module in the context of test. Will try retaining the original one.

dxe_core/src/lib.rs
HappyCoding-2025: @makubacki  - Sure, how do you want to handle this? Pls advise . I can return error but coverage will not go on for other functions below in the code sequence and if a middle function returns an error.
Some of the function during code coverage wont work with out a hardware right, pls advise.

makubacki: Mark this function as skip for tarpaulin and acknowledge we will not test this function.

dxe_core/src/gcd/spin_locked_gcd.rs
makubacki: Guidance here for the moment is to remove the new if statement with `cfg!()` and always `panic!()`. Handle the panic with its message in the unit test. Coverage will stop here since we need to support `create_cpu_paging()` working in host-based unit tests. File an issue for that and work on it as a focused item.

dxe_core/src/lib.rs
makubacki: Always panic (with `except()`) for now. Ensure that unit tests cover function called here to accumulate coverage of individual functionality.

dxe_core/src/allocator.rs
makubacki: Remove this function for now. File an issue with all details and push this code to a commit in a branch linked to from the issue. We will need to rethink how to test these subsystems.

dxe_core/src/image.rs
makubacki: We discussed offline to remove.

dxe_core/src/test_support.rs
```diff
@@ -316,3 +320,147 @@ pub(crate) fn build_test_hob_list(mem_size: u64) -> *const c_void {
     }
     mem.as_ptr() as *const c_void
 }
+
```
makubacki, line 323: Please add comments to summarize what this functions does and why a developer might consider using it in their tests.

dxe_core/src/test_support.rs
```diff
@@ -316,3 +320,147 @@ pub(crate) fn build_test_hob_list(mem_size: u64) -> *const c_void {
...
+            core::ptr::copy(&allocation_hob_template, cursor as *mut hob::MemoryAllocationModule, 1);
+            cursor = cursor.offset(allocation_hob_template.header.length as isize);
+        }
+
+        core::ptr::copy(&end, cursor as *mut header::Hob, 1);
+    }
+    mem.as_ptr() as *const c_void
+}
+
+pub(crate) fn fill_file_buffer_in_memory_allocation_module(hob_list: &HobList) -> Result<(), &'static str> {
```
makubacki, line 434: Please add comments to summarize what this functions does and why a developer might consider using it in their tests.

HappyCoding-2025:
@makubacki  @joschock  @Javagedes  - Based on the input from all of you

1. the lib.rs coverage is more like entire system testing than unit testing. I will close this PR as per the discussion. 
2. Only part that will be ported is the Hob list build as part of another PR. Will send the PR for review shortly.

PR #328, opened 2025-03-31
interface.md: Fix code examples

makubacki:
## Description

Update the code examples so the code is valid for compilation.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [x] Includes documentation?

## How This Was Tested

- Temporarily add .md file to tests

## Integration Instructions

- N/A

### DIFF
--- a/docs/src/driver/interface.md
+++ b/docs/src/driver/interface.md
@@ -58,13 +58,20 @@ Please reference the documentation in `uefi_sdk::component` for more information
 
 ```rust
 use uefi_sdk::{
-    boot_services::StandardBootServices,
+    boot_services::BootServices,
     component::params::{Config, ConfigMut},
     error::Result,
 };
 
-fn validate_random_data_driver(bs: StandardBootServices, data: Config<&[u8]>, expected_crc32: Config<u32>) -> Result<()> {
-    assert_eq!(bs.calculate_crc_32(*data), Ok(*expected_crc32))
+// Note: This uncommented code is valid for demonstraction, but the function will not be registered with the core where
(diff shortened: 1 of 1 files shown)


docs/src/driver/interface.md
```diff
@@ -58,13 +58,13 @@ Please reference the documentation in `uefi_sdk::component` for more information
...
 use uefi_sdk::{
-    boot_services::StandardBootServices,
+    boot_services::BootServices,
     component::params::{Config, ConfigMut},
     error::Result,
 };
 
-fn validate_random_data_driver(bs: StandardBootServices, data: Config<&[u8]>, expected_crc32: Config<u32>) -> Result<()> {
-    assert_eq!(bs.calculate_crc_32(*data), Ok(*expected_crc32))
+fn validate_random_data_driver<T: Default>(bs: impl BootServices, data: Config<T>, expected_crc32: Config<u32>) -> Result<()> {
```
Javagedes, line 73: This does compile as is, and is what you want to write for testing purposes, but it is not able to be registered with the core as a driver. To do that, you have to specify the actual implementations. So what gets registered with the core with actually be a wrapper function.

``` rust
// The wrapper function (Config<f32> could be any type)
fn actual_driver(bs: StandardBootServices, data: Config<f32>, expected_crc32: Config<u32>) -> Result<()> {
    validate_random_data_driver(bs, data, expected_crc32)
}
```

May want to add this as a note or something

makubacki: Done. Thanks for confirming.

PR #330, opened 2025-04-01
[REBASE & FF] Implement Memory Manager Service for DXE Core

cfernald:
## Description

Implements the `MemoryManager` API for the core and publishes as a Service. This introduces an empty static structure that is intended both for internal used through references to the `CoreMemoryManager` as well as externall through dependency injection for `Service<dyn MemoryManager>`.

This initial change is just to introduce the service and does not yet attempt to consume internally or cleanup existing API use cases. These will come in a later PR.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- Tested on Q35 with included integrations tests. 
- All existing tests pass

## Integration Instructions

N/A

### DIFF
--- a/core/dxe_core/Cargo.toml
+++ b/core/dxe_core/Cargo.toml
@@ -48,6 +48,7 @@ uefi_debugger = { workspace = true }
 uefi_device_path = { workspace = true }
 uefi_depex = { workspace = true}
 uefi_performance = { workspace = true }
+uefi_test = { workspace = true }
 
 [dev-dependencies]
 # To avoid circular dependencies, cargo-release skips dev dependencies when evaluating the release order for
--- a/core/dxe_core/src/lib.rs
+++ b/core/dxe_core/src/lib.rs
@@ -52,6 +52,7 @@ mod hw_interrupt_protocol;
 mod image;
 mod memory_attributes_protocol;
 mod memory_attributes_table;
+mod memory_manager;
 mod misc_boot_services;
 mod pecoff;
 mod protocol_db;
@@ -68,6 +69,7 @@ use core::{ffi::c_void, ptr, str::FromStr};
 
 use alloc::{boxed::Box, vec::Vec};
 use gcd::SpinLockedGcd;
+use memory_manager::CoreMemoryManager;
 use mu_pi::{
     fw_fs,
     hob::{get_c_hob_list_size, HobList},
@@ -240,6 +242,7 @@ where
 
         self.storage.add_service(cpu);
         self.storage.add_service(interrupt_manager);
+        self.storage.add_service(CoreMemoryManager);
 
         Core {
             physical_hob_list,
--- a/core/dxe_core/src/memory_manager.rs
+++ b/core/dxe_core/src/memory_manager.rs
@@ -1,0 +1,284 @@
+use alloc::boxed::Box;
+use r_efi::efi;
+use uefi_sdk::{
+    base::{UEFI_PAGE_MASK, UEFI_PAGE_SIZE},
+    component::service::{
+        memory::{
+            AccessType, AllocationOptions, CachingType, MemoryError, MemoryManager, PageAllocation,
+            PageAllocationStrategy,
+        },
+        IntoService, Service,
+    },
+    efi_types::EfiMemoryType,
+    error::EfiError,
+    uefi_pages_to_size,
+};
+use uefi_test::{u_assert, u_assert_eq, uefi_test};
+
+use crate::{
+    allocator::{core_allocate_pages, core_free_pages},
+    dxe_services,
+};
+
+/// Structure for wrapper rust allocator APIs.
+#[derive(IntoService)]
+#[service(dyn MemoryManager)]
+pub(crate) struct CoreMemoryManager;
+
+impl MemoryManager for CoreMemoryManager {
+    fn allocate_pages(&self, page_count: usize, options: AllocationOptions) -> Result<PageAllocation, MemoryError> {
+        allow_allocations_for_type(options.memory_type())?;
+        let mut address: efi::PhysicalAddress = 0;
+        let alignment = options.alignment();
+
+        if !alignment.is_power_of_two() || alignment & UEFI_PAGE_MASK != 0 {
+            return Err(MemoryError::InvalidAlignment);
+        }
+
+        let alloc_type = match options.strategy() {
+            PageAllocationStrategy::Any => efi::ALLOCATE_ANY_PAGES,
+            PageAllocationStrategy::Address(requested_address) => {
+                if requested_address % alignment != 0 {
+                    return Err(MemoryError::UnalignedAddress);
+                }
+
+                address = requested_address as efi::PhysicalAddress;
+                efi::ALLOCATE_ADDRESS
+            }
+        };
+
+        let result =
+            core_allocate_pages(alloc_type, options.memory_type().into(), page_count, &mut address, Some(alignment));
+
+        match result {
+            Ok(_) => {
+                let allocation = unsafe {
+                    PageAllocation::new(address as usize, page_count, &CoreMemoryManager)
+                        .map_err(|_| MemoryError::InternalError)?
+                };
+                Ok(allocation)
+            }
+            Err(EfiError::OutOfResources) => Err(MemoryError::NoAvailableMemory),
+            Err(_) => Err(MemoryError::InternalError),
+        }
+    }
+
+    unsafe fn free_pages(&self, address: usize, page_count: usize) -> Result<(), MemoryError> {
+        let result = core_free_pages(address as efi::PhysicalAddress, page_count);
+        match result {
+            Ok(_) => Ok(()),
+            Err(EfiError::NotFound) => Err(MemoryError::InvalidAddress),
+            Err(_) => Err(MemoryError::InternalError),
+        }
+    }
+
+    fn get_allocator(&self, memory_type: EfiMemoryType) -> Result<&'static dyn core::alloc::Allocator, MemoryError> {
+        // TODO: Because the allocator has to live for a undefined amount of time
+        // to allow for freeing the memory, we can only use the static allocators.
+        // This should be changed in the future.
+        let allocator = match memory_type {
+            EfiMemoryType::LoaderCode => &crate::allocator::EFI_LOADER_CODE_ALLOCATOR,
+            EfiMemoryType::BootServicesCode => &crate::allocator::EFI_BOOT_SERVICES_CODE_ALLOCATOR,
+            EfiMemoryType::BootServicesData => &crate::allocator::EFI_BOOT_SERVICES_DATA_ALLOCATOR,
+            EfiMemoryType::RuntimeServicesCode => &crate::allocator::EFI_RUNTIME_SERVICES_CODE_ALLOCATOR,
+            EfiMemoryType::RuntimeServicesData => &crate::allocator::EFI_RUNTIME_SERVICES_DATA_ALLOCATOR,
+            _ => {
+                return Err(MemoryError::UnsupportedMemoryType);
+            }
+        };
+        Ok(allocator as &'static dyn core::alloc::Allocator)
+    }
+
+    unsafe fn set_page_attributes(
+        &self,
+        address: usize,
+        page_count: usize,
+        access: AccessType,
+        caching: Option<CachingType>,
+    ) -> Result<(), MemoryError> {
+        if page_count == 0 {
+            return Err(MemoryError::InvalidPageCount);
+        }
+
+        if address % UEFI_PAGE_SIZE != 0 {
+            return Err(MemoryError::UnalignedAddress);
+        }
+
+        let access_attributes = match access {
+            AccessType::NoAccess => efi::MEMORY_RP,
+            AccessType::ReadOnly => efi::MEMORY_RO | efi::MEMORY_XP,
+            AccessType::ReadWrite => efi::MEMORY_XP,
+            AccessType::ReadExecute => efi::MEMORY_RO,
+            AccessType::ReadWriteExecute => return Err(MemoryError::UnsupportedAttributes),
+        };
+
+        let cache_attributes = match caching {
+            Some(CachingType::Uncached) => Some(efi::MEMORY_UC),
+            Some(CachingType::WriteBack) => Some(efi::MEMORY_WB),
+            Some(CachingType::WriteCombining) => Some(efi::MEMORY_WC),
+            Some(CachingType::WriteThrough) => Some(efi::MEMORY_WT),
+            Some(CachingType::WriteProtect) => return Err(MemoryError::UnsupportedAttributes),
+            None => None,
+        };
+
+        let mut current_base: u64 = address as u64;
+        let range_end: u64 = (address + uefi_pages_to_size!(page_count)) as u64;
+        while current_base < range_end {
+            let descriptor =
+                match crate::dxe_services::core_get_memory_space_descriptor(current_base as efi::PhysicalAddress) {
+                    Ok(descriptor) => descriptor,
+                    Err(e) => {
+                        log::error!("Memory descriptor fetching failed with error {:#x?} for {:#x}", e, current_base,);
+                        return Err(MemoryError::InvalidAddress);
+                    }
+                };
+            let descriptor_end = descriptor.base_address + descriptor.length;
+
+            // it is still legal to split a descriptor and only set the attributes on part of it
+            let next_base = u64::min(descriptor_end, range_end);
+            let current_len = next_base - current_base;
+
+            // Always clear all access attributes and set the requested ones.
+            let mut new_attributes = descriptor.attributes & !efi::MEMORY_ACCESS_MASK;
+            new_attributes |= access_attributes;
+
+            // If no cache attributes were requested, leave them unchanged.
+            if let Some(cache_attributes) = cache_attributes {
+                new_attributes &= !efi::CACHE_ATTRIBUTE_MASK;
+                new_attributes |= cache_attributes;
+            }
+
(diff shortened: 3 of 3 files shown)


dxe_core/src/allocator.rs
```diff
@@ -1001,6 +1007,79 @@ pub fn install_memory_services(bs: &mut efi::BootServices) {
     bs.get_memory_map = get_memory_map;
 }
 
+/// Structure for wrapper rust allocator APIs.
+#[derive(IntoService)]
+#[service(dyn MemoryManager)]
+pub struct CoreMemoryManager;
+
+impl MemoryManager for CoreMemoryManager {
```
os-d: Right now these wrap the rustified PI spec APIs, which is fine, but we now have 4 or 5 functions called allocate_pages. Is it possible to simply remove the core_allocate_* ones and replace them with these? There should be few enough consumers today to make the change easy enough.

I'm hesitant to add another wrapper, it would be really nice if this was just the same allocation services we use in the core (unless we think there is a reason to distinguish). Then C ABI wrapper is separate and may have some separate handling to ensure the PI spec interface is met.

dxe_core/src/allocator.rs
```diff
@@ -1001,6 +1007,79 @@ pub fn install_memory_services(bs: &mut efi::BootServices) {
...
+    fn free_pages(&self, address: usize, page_count: usize) -> Result<(), MemoryError> {
+        let result = core_free_pages(address as efi::PhysicalAddress, page_count);
+        match result {
+            Ok(_) => Ok(()),
+            Err(EfiError::NotFound) => Err(MemoryError::InvalidAddress),
+            Err(_) => Err(MemoryError::InternalError),
+        }
+    }
+
+    fn get_allocator(&self, memory_type: AllocationType) -> Result<&'static dyn core::alloc::Allocator, MemoryError> {
```
os-d: I again am wondering if there is a cleaner way to do this rather than passing the allocator reference directly, but will defer to others.

dxe_core/src/allocator.rs
cfernald: Completely agree. This is just the bare minimum to prove out the interface. The final change should more thoroughly integrate this interface as the primary use case, and probably remove crate access to the internal allocator functions.

dxe_core/src/allocator.rs
cfernald: I'm not sure we can without really limiting ourselves. The problem is the the allocator interface is directly provided to a lot of standard structures such as Box or BTree. We could create a wrapper object that keeps the allocator private and can create Box's on behalf of the caller, but that would require us create an explicit wrapper function for every allocator use case, which I don't think scales well, but would make it so they have to use the std types. In the end they can always just leak the memory using box though and access the raw pointer if they want to so we cannot prevent unsafe behavior, we can just try to put guard rails up to prevent well-intentioned mistakes.

dxe_core/src/allocator.rs
cfernald: I have decided to first implement the interface and produce it to unlock component, then follow up with isolation and refactor for memory code.

dxe_core/src/memory_manager.rs
```diff
@@ -0,0 +1,279 @@
...
+    dxe_services,
+};
+
+/// Constants for cache attributes that may be changed.
+const SET_CACHE_MASK: u64 = efi::MEMORY_UC | efi::MEMORY_WB | efi::MEMORY_WC | efi::MEMORY_WT;
+
+/// Structure for wrapper rust allocator APIs.
+#[derive(IntoService)]
+#[service(dyn MemoryManager)]
+pub struct CoreMemoryManager;
```
Javagedes: nit: pub(crate)

dxe_core/src/allocator.rs
```diff
@@ -59,7 +59,7 @@ const PRIVATE_ALLOCATOR_TRACKING_GUID: efi::Guid =
 // to a different allocator. This allocator does not need to be public since all dynamic allocations will implicitly
 // allocate from it.
 #[cfg_attr(target_os = "uefi", global_allocator)]
-static EFI_BOOT_SERVICES_DATA_ALLOCATOR: UefiAllocator = UefiAllocator::new(
+pub static EFI_BOOT_SERVICES_DATA_ALLOCATOR: UefiAllocator = UefiAllocator::new(
```
Javagedes: nit: pub(crate)

dxe_core/src/allocator/fixed_size_block_allocator.rs
```diff
@@ -50,6 +50,16 @@ fn list_index(layout: &Layout) -> Option<usize> {
     BLOCK_SIZES.iter().position(|&s| s >= required_block_size)
 }
 
+/// Converts the given alignment to a shift value.
+fn alignment_page_to_shift(alignment: usize) -> Result<usize, EfiError> {
```
Javagedes: Please add tests for this func

dxe_core/src/lib.rs
```diff
@@ -64,6 +65,8 @@ mod runtime;
 mod systemtables;
 mod tpl_lock;
 
+pub(crate) use memory_manager::CoreMemoryManager;
```
Javagedes: `CoreMemoryManager` is only used in this file, so `pub(crate)` is unecessary. I'd just do `use memory_manager::CoreMemoryManager` or where you add it to storage below, just add `memory_manager::CoreMemoryManager` (I think this is my preference since it's only used once)

dxe_core/src/memory_manager.rs
```diff
@@ -0,0 +1,279 @@
+use alloc::boxed::Box;
```
Javagedes: Do we think we will have many more `Core` provided services (that are defined in the core?). I'm curious your thoughts on possibly making a services module or something of the sort, and putting core provided services there. As it stands though, it's only yours, so it may not be useful.

dxe_core/src/allocator.rs
```diff
@@ -431,6 +431,7 @@ pub fn core_allocate_pages(
     memory_type: efi::MemoryType,
     pages: usize,
     memory: *mut efi::PhysicalAddress,
+    alignment: Option<usize>,
```
Javagedes: I'm still iffy on the interface of `None` being a way to override the default value (in this case of 4K), but I think with alignment purposes its acceptable. Even though we unwrap the alignment in the first couple of lines, would you mind adding some docs on this function that just states that the default alignment is 4K

joschock: These functions were originally signatured to match the UEFI spec API semantics, but there is no need for them to remain that way - if we are going to move away from that, we could go even further - here and in uefi_allocator, you could consider just going straight to rust semantics and requiring https://doc.rust-lang.org/beta/core/alloc/struct.Layout.html. You could provide Layout constructor functions or macros that would check invariants (e.g. align is a power of two greater than base page size, for example).

I think 
```
core_allocate_pages(allocation_type, memory_type, page_alignment!(pages))?;
core_allocate_pages(allocation_type, memory_type, custom_alignment!(pages, my_custom_alignment))?;
```
Where `page_alighment` and `custom_alignment` are macros that return a checked Layout object would be a reasonable way to do this, and more consistent with prior allocator art in the larger rust ecosystem.

dxe_core/src/allocator.rs
```diff
@@ -445,18 +446,19 @@ pub fn core_allocate_pages(
     }
 
     let handle = AllocatorMap::handle_for_memory_type(memory_type)?;
+    let alignment = alignment.unwrap_or(UEFI_PAGE_SIZE);
```
Javagedes: unless this is verified somewhere else, we probably need to make sure `alignment % 2 == 0` or something, right? (could be wrong)

dxe_core/src/memory_manager.rs
cfernald: Actually, what I want to do (not in this PR) is to isolate functional code. I want to move all of the "memory" code into a "memory" module, and the only accessible thing outside of that module is CoreMemoryManager and maybe some other functions. To that end it makes more sense to have memory_manager be a abstraction for that code rather then having a group of abstractions elsewhere.

dxe_core/src/allocator.rs
cfernald: @Javagedes , this is intended to be a minimal first pass on changing some of the interfaces. I want to take a second refactor pass to move all the memory code into a submodule and change all non-memory consumers to use the CoreMemoryManager API for allocations. This should help cleanup and rustify the shared interfaces and we can cleanup the internals as time goes on.

@joschock , I agree that we need to rework these. Though my plan right now was for this to be the minimal change to add the new functionality then come in after with a cleanup PR for clarity.

dxe_core/src/lib.rs
cfernald: This was back when I had dreams to cleanup different allocator interfaces used within the core, but chose to do that as a later PR. Will remove this for now.

dxe_core/src/allocator.rs
cfernald: It will be verified internally by alignment_page_to_shift, I figured no need to check it at every layer, just when it is used.

dxe_core/src/allocator/fixed_size_block_allocator.rs
```diff
@@ -393,8 +406,8 @@ impl FixedSizeBlockAllocator {
             .allocate_memory_space(
                 allocation_strategy,
                 GcdMemoryType::SystemMemory,
-                UEFI_PAGE_SHIFT,
-                pages * ALIGNMENT,
+                align_shift,
+                pages << UEFI_PAGE_SHIFT,
```
os-d: nit: use uefi_pages_to_size!

dxe_core/src/memory_manager.rs
```diff
@@ -0,0 +1,279 @@
...
+            Ok(_) => Ok(()),
+            Err(EfiError::NotFound) => Err(MemoryError::InvalidAddress),
+            Err(_) => Err(MemoryError::InternalError),
+        }
+    }
+
+    fn get_allocator(&self, memory_type: EfiMemoryType) -> Result<&'static dyn core::alloc::Allocator, MemoryError> {
+        // TODO: Because the allocator has to live for a undefined amount of time
+        // to allow for freeing the memory, we can only use the static allocators.
+        // This should be changes in the future.
```
os-d:
```suggestion
        // This should be changed in the future.
```

dxe_core/src/memory_manager.rs
```diff
@@ -0,0 +1,279 @@
...
+
+        let cache_attributes = match caching {
+            Some(CachingType::Uncached) => Some(efi::MEMORY_UC),
+            Some(CachingType::WriteBack) => Some(efi::MEMORY_WB),
+            Some(CachingType::WriteCombining) => Some(efi::MEMORY_WC),
+            Some(CachingType::WriteThrough) => Some(efi::MEMORY_WT),
+            None => None,
+        };
+
+        let mut current_base: u64 = address as u64;
```
os-d: This logic is repeated in enough places I wonder if it is worthwhile / I think it is worthwhile to consolidate this into a function, or perhaps better to have a GCD iterator for a given range that you can do a for each on. Not something you have to change here, but if you don't, can you file an issue to track it?

dxe_core/src/memory_manager.rs
```diff
@@ -0,0 +1,279 @@
...
+        let range_end: u64 = (address + page_count * UEFI_PAGE_SIZE) as u64;
```
os-d: nit: uefi_pages_to_size!

dxe_core/src/memory_manager.rs
```diff
@@ -0,0 +1,279 @@
...
+};
+use uefi_test::{u_assert, u_assert_eq, uefi_test};
+
+use crate::{
+    allocator::{core_allocate_pages, core_free_pages},
+    dxe_services,
+};
+
+/// Constants for cache attributes that may be changed.
+const SET_CACHE_MASK: u64 = efi::MEMORY_UC | efi::MEMORY_WB | efi::MEMORY_WC | efi::MEMORY_WT;
```
os-d: Should we not use the r-efi `CACHE_ATTRIBUTE_MASK`? I understand it has two other options set, but also, do we need to allow UCE and WP? May need to reach out to platform teams to see what silicon code is doing.

dxe_core/src/memory_manager.rs
```diff
@@ -0,0 +1,279 @@
...
+            let next_base = u64::min(descriptor_end, range_end);
+            let current_len = next_base - current_base;
+
+            // Always clear all access attributes and set the requested ones.
+            let mut new_attributes = descriptor.attributes & !efi::MEMORY_ACCESS_MASK;
+            new_attributes |= access_attributes;
+
+            // If no cache attributes were requested, leave them unchanged.
+            if let Some(cache_attributes) = cache_attributes {
+                new_attributes &= !SET_CACHE_MASK;
```
os-d: Same comments on caching

dxe_core/src/memory_manager.rs
```diff
@@ -0,0 +1,279 @@
...
+        if page_count == 0 {
+            return Err(MemoryError::InvalidPageCount);
+        }
+
+        if address % UEFI_PAGE_SIZE != 0 {
+            return Err(MemoryError::UnalignedAddress);
+        }
+
+        let base_address = address as efi::PhysicalAddress;
+        let length = (page_count * UEFI_PAGE_SIZE) as u64;
```
os-d: nit: uefi_pages_to_size

dxe_core/src/allocator.rs
Javagedes: Sounds good. It's being verified, that is all that matters :)

cfernald:
> This looks good. I think it needs to be split into two commits though. One for the alignment changes and one for adding the service.

Will do before merging, I like to avoid force pushes during development though.

dxe_core/src/memory_manager.rs
cfernald: Yeah, this was a conscious choice though i'm open to change it. In my thinking being to leave both MEMORY_UCE & MEMORY_WP unaltered as the rust interface doesnt not allow changing them. That is memory_manager will ignore attributes it cannot set, and if one of those bits is set and the internal code doesn't like the combination coming in from this API then that should fail as we would want. I could make this more explicit though and fail using a `Some` for caching when those other two are set.

os-d: Right, that's my question though, are we going to see potential consumers that need to set those caching bits?

makubacki: If this is what we discussed offline, to capture it here - UCE is not used to my knowledge. WP is relatively common especially for covering the flash MMIO range. There will need to be a similar mechanism for that range in DXE.

cfernald: Dropped this and dropped UC.

dxe_core/src/memory_manager.rs
cfernald: https://github.com/OpenDevicePartnership/uefi-dxe-core/issues/385

dxe_core/src/allocator/fixed_size_block_allocator.rs
makubacki: This function seems awkwardly named but it's not a big deal.

makubacki: Something like `page_shift_from_alignment()` seems clearer to me.

dxe_core/src/memory_manager.rs
```diff
@@ -0,0 +1,284 @@
...
+    let boxed_struct = Box::new_in(42, allocator);
+    u_assert_eq!(*boxed_struct, 42, "Failed to allocate boxed struct.");
+    drop(boxed_struct);
+
+    Ok(())
+}
+
+#[uefi_test]
+fn memory_manager_attributes_test(mm: Service<dyn MemoryManager>) -> uefi_test::Result {
+    // The default attributes for memory should bne read/write.
```
makubacki:
```suggestion
    // The default attributes for memory should be read/write.
```

dxe_core/src/memory_manager.rs
```diff
@@ -0,0 +1,284 @@
...
+    let result = mm.allocate_pages(1, AllocationOptions::new());
+    u_assert!(result.is_ok(), "Failed to allocate single page.");
+    let allocation = result.unwrap();
+    let address = allocation.into_raw_ptr::<u8>() as usize;
+    let result = mm.get_page_attributes(address, 1);
+    u_assert!(result.is_ok(), "Failed to  original page attributes.");
```
makubacki: Incomplete sentence in the assert message.

dxe_core/src/memory_manager.rs
```diff
@@ -0,0 +1,284 @@
...
+    let (access, caching) = result.unwrap();
+    u_assert_eq!(access, AccessType::ReadWrite, "Allocation did not return at Read/Write access.");
```
makubacki:
```suggestion
    u_assert_eq!(access, AccessType::ReadWrite, "Allocation did not return Read/Write access.");
```

dxe_core/src/memory_manager.rs
```diff
@@ -0,0 +1,284 @@
...
+
+    // Test changing the attributes to read only.
+    let result = unsafe { mm.set_page_attributes(address, 1, AccessType::ReadOnly, None) };
+    u_assert!(result.is_ok(), "Failed to set page attributes.");
+    let result = mm.get_page_attributes(address, 1);
+    u_assert!(result.is_ok(), "Failed to get altered page attributes.");
+    let (access, new_caching) = result.unwrap();
+    u_assert_eq!(access, AccessType::ReadOnly, "Allocation did not return at Read/Write access.");
```
makubacki:
```suggestion
    u_assert_eq!(access, AccessType::ReadOnly, "Allocation did not return Read/Write access.");
```

PR #338, opened 2025-04-01
[REBASE & FF] Add Compatibility Mode

os-d:
## Description

This PR adds the following commits (note that memory protections and compatibility mode documentation is actively being written):

Remove ensure_capacity Allocator API
--
This API is unused and was added to support the paging crate, but it ended up never being needed.

The functionality is removed in this commit.

Move Memory Protection Enforcement to GCD Layer
--
Part of memory protection management (specifically setting RP on free memory and XP on newly allocated memory) was occurring in the memory block layer instead of the GCD layer. This was inappropriate coupling of the memory protection feature to a more generic layer and caused the feature to unnecessarily and complexly span layers. This was necessary at the time because of a different design, but is no longer required.

This commit moves the memory protections to the GCD layer and out of the memory block layer.

Allocate Initial Memory Blocks with Page Granularity
--
The memory block initialization code was not setting UEFI_PAGE_SHIFT as the required align when calling allocate_memory_space, which could cause it be non-page aligned, which would be fatal. This was unlikely because the PHIT HOB should pass in a page aligned base address, but the memory block code should ensure this.

This commit also updates some tests to use UEFI_PAGE_SHIFT.

Move MAT Publishing Out of Page Change Callback
--
The page change callback hosts the MAT publication code. However, when booting Linux, this causes a locking issue because where the callback is invoked from.

This commits moves the MAT publishing logic out of the locked code to avoid this.

Print EBS Success Message
--
Add a simple info message identifying that EBS succeeded. This is helpful to tell if an OS hangs that EBS succeeded and was not the failure.

Track NX_COMPAT in pe_info Struct
--
This commits adds setting the NX compat bit in the pe_info struct based on the PE/COFF DLL Characteristic. This is used to determine whether to enter compatibility mode.

Add get_memory_ranges_for_memory_type API to Allocator
--
This commit adds an API to the allocator to return a Vec of the ranges owned by the allocator for a given memory type.

The primary use case of this API is for compatibility mode to find all the memory ranges owned by the EFI loader code and data allocators and set them to RWX.

Add Compatibility Mode Support

This patch adds compatibility mode (to be documented soon). This adds a build flag (disabled by default) to enable compatibility mode when launching EFI_APPLICATIONS without the NX_COMPAT flag. The primary use case for this is to launch shim/grub which do not work with memory protections.

When compatibility mode is enabled the following occurs:
- The loaded image that triggered the mode has its memory set to RWX
- All newly allocated memory is allocated RWX
- All existing EFI loader code and data memory is marked RWX
- The memory attributes protocol is uninstalled
- The range 0 - 0xA000 is set to RWX if it is system memory

- [x] Impacts functionality?
- [x] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested by booting Windows and Linux on Q35/SBSA.

## Integration Instructions

Platforms must include the compatibility_mode_allowed feature in their Cargo.toml to enable compatibility mode.

### DIFF
--- a/docs/src/integrate/dxe_core.md
+++ b/docs/src/integrate/dxe_core.md
@@ -226,3 +226,20 @@ pub extern "efiapi" fn _start(physical_hob_list: *const c_void) -> ! {
     loop {}
 }
 ```
+
+## DXE Core Features
+
+### Compatibility Mode
+
+The Rust DXE Core supports Compatibility Mode, the ability to launch bootloaders that do not support memory protections,
+most notably most shipping versions of Linux distros. By default the `compatibility_mode_allowed` feature flag is
+disabled. If a platform wishes to build with it, they must include it in their Cargo.toml as such:
+
+```rust
+[dependencies]
+dxe_core = { version = "9", features = ["compatibility_mode_allowed"] }
+```
+
+With the flag disabled, the Rust DXE Core will not launch any EFI_APPLICATIONs that do not have the NX_COMPAT
+DLL Characteristics set. With the flag enabled, the Rust DXE Core will enable
+[compatibility mode](../dxe_core/memory_management.md).
--- a/dxe_core/src/gcd.rs
+++ b/dxe_core/src/gcd.rs
@@ -267,6 +267,53 @@ fn remove_range_overlap<T: PartialOrd + Copy>(a: &Range<T>, b: &Range<T>) -> [Op
     }
 }
 
+#[cfg(feature = "compatibility_mode_allowed")]
+/// This activates compatibility mode for the GCD.
+/// This will:
+/// - Activate compatibility mode for the GCD lower layers
+/// - Set the memory space attributes for all memory ranges in the loader code and data allocators to be RWX
+/// - Uninstall the memory attributes protocol
+pub(crate) fn activate_compatibility_mode() {
+    GCD.activate_compatibility_mode();
+    // if the allocator doesn't have any memory, then when it is used next it will allocate from the GCD
+    // and the GCD will be in compatibility mode, so we don't care here
+    let mut loader_mem_ranges = crate::allocator::get_memory_ranges_for_memory_type(efi::LOADER_CODE);
+    loader_mem_ranges.extend(crate::allocator::get_memory_ranges_for_memory_type(efi::LOADER_DATA));
+    for range in loader_mem_ranges.iter() {
+        let mut addr = range.start;
+        while addr < range.end {
+            let mut len = uefi_sdk::base::UEFI_PAGE_SIZE;
+            match GCD.get_memory_descriptor_for_address(addr) {
+                Ok(descriptor) => {
+                    let attributes = descriptor.attributes & !efi::MEMORY_XP;
+                    len = match descriptor.base_address + descriptor.length {
+                        end if end > range.end => (range.end - addr) as usize,
+                        _ => descriptor.length as usize,
+                    };
+                    if GCD.set_memory_space_attributes(addr as usize, len, attributes).is_err() {
+                        log::error!(
+                                        "Failed to set memory space attributes for range {:#x?} - {:#x?}, compatibility mode may fail",
+                                        range.start,
+                                        range.end,
+                                    );
+                        debug_assert!(false);
+                    }
+                }
+                _ => {
+                    log::error!(
+                        "Failed to get memory space descriptor for range {:#x?} - {:#x?}, compatibility mode may fail",
+                        range.start,
+                        range.end,
+                    );
+                    debug_assert!(false);
+                }
+            }
+            addr += len as u64;
+        }
+    }
+    crate::memory_attributes_protocol::uninstall_memory_attributes_protocol();
+}
+
 #[cfg(test)]
 mod tests {
     use core::ffi::c_void;
--- a/dxe_core/src/gcd/spin_locked_gcd.rs
+++ b/dxe_core/src/gcd/spin_locked_gcd.rs
@@ -279,6 +279,9 @@ struct GCD {
     memory_blocks: Rbt<'static, MemoryBlock>,
     allocate_memory_space_fn: GcdAllocateFn,
     free_memory_space_fn: GcdFreeFn,
+    /// Default attributes for memory allocations
+    /// This is efi::MEMORY_XP unless we have entered compatibility mode, in which case it is 0, e.g. no protection
+    default_attributes: u64,
 }
 
 impl core::fmt::Debug for GCD {
@@ -300,6 +303,7 @@ impl GCD {
             maximum_address: 1 << processor_address_bits,
             allocate_memory_space_fn: Self::allocate_memory_space_internal,
             free_memory_space_fn: Self::free_memory_space_worker,
+            default_attributes: efi::MEMORY_XP,
         }
     }
 
@@ -1099,6 +1103,15 @@ impl GCD {
         self.memory_blocks.len()
     }
 
+    #[cfg(feature = "compatibility_mode_allowed")]
+    /// This function activates compatibility mode for the GCD, which is just to set the default attributes to 0,
+    /// which will prevent new memory from being allocated as non-executable. This function is purposefully not set
+    /// to be pub(crate) because the only caller of it is SpinLockedGcd.activate_compatibility_mode(). And this should
+    /// not be called except by that function.
+    fn activate_compatibility_mode(&mut self) {
+        self.default_attributes = 0;
+    }
+
     //Note: truncated strings here are expected and are for alignment with EDK2 reference prints.
     const GCD_MEMORY_TYPE_NAMES: [&'static str; 8] = [
         "NonExist ", // EfiGcdMemoryTypeNonExistent
@@ -1727,7 +1740,7 @@ pub enum MapChangeType {
 /// GCD map change callback function type.
 pub type MapChangeCallback = fn(MapChangeType);
 
-/// Implements a spin locked GCD  suitable for use as a static global.
+/// Implements a spin locked GCD suitable for use as a static global.
 pub struct SpinLockedGcd {
     memory: tpl_lock::TplMutex<GCD>,
     io: tpl_lock::TplMutex<IoGCD>,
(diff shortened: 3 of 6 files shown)


dxe_core/src/gcd.rs
```diff
@@ -267,6 +267,53 @@ fn remove_range_overlap<T: PartialOrd + Copy>(a: &Range<T>, b: &Range<T>) -> [Op
...
+    let mut loader_mem_ranges = crate::allocator::get_memory_ranges_for_memory_type(efi::LOADER_CODE);
+    loader_mem_ranges.extend(crate::allocator::get_memory_ranges_for_memory_type(efi::LOADER_DATA));
+    for range in loader_mem_ranges.iter() {
+        let mut addr = range.start;
+        while addr < range.end {
+            let mut len = uefi_sdk::base::UEFI_PAGE_SIZE;
+            match GCD.get_memory_descriptor_for_address(addr) {
+                Ok(descriptor) => {
+                    let attributes = descriptor.attributes & !efi::MEMORY_XP;
+                    len = match descriptor.base_address + descriptor.length {
```
joschock, line 289: Does this ever happen in practice? wouldn't this be a bug in get_memory_ranges_for_memory_type() assuming no parallel modification of LOADER_CODE/LOADER_DATA regions? Maybe convert to an assert.

dxe_core/src/gcd/spin_locked_gcd.rs
```diff
@@ -2442,6 +2458,51 @@ impl SpinLockedGcd {
...
         self.io.lock().io_descriptor_count()
     }
+
+    #[cfg(feature = "compatibility_mode_allowed")]
+    /// This activates compatibility mode for the GCD.
+    /// This will:
+    /// - Map the range 0 - 0xA0000 as RWX if the memory type is SystemMemory.
+    /// - Update the locked GCD to not set efi::MEMORY_XP on newly allocated pages
+    pub(crate) fn activate_compatibility_mode(&self) {
+        const LEGACY_BIOS_WB_ADDRESS: usize = 0xA0000;
```
joschock, line 2468: why not go all the way to 1MB?

dxe_core/src/image.rs
```diff
@@ -649,12 +649,60 @@ fn core_load_pe_image(
...
     }
 
-    // finally, update the GCD attributes for this image so that code sections have RO set and data sections have XP
-    apply_image_memory_protections(&pe_info, &private_info);
+    match pe_info.image_type {
+        EFI_IMAGE_SUBSYSTEM_EFI_APPLICATION if !pe_info.nx_compat => {
+            // we are trying to load an application image that is not NX compatible, likely a bootloader
+            // if we are configured to allow compatibility mode, we need to activate it now. Otherwise, just continue
+            // to load the image
+            //activate_compatibility_mode(&private_info)?;
```
joschock: Unexpected comment?

dxe_core/src/memory_attributes_protocol.rs
```diff
@@ -189,14 +189,52 @@ impl EfiMemoryAttributesProtocolImpl {
     }
 }
 
+static mut MEMORY_ATTRIBUTES_PROTOCOL_HANDLE: Option<efi::Handle> = None;
```
joschock: you could use `AtomicPtr` and `AtomicUsize` to avoid `static mut` here and `unsafe {}` elsewhere.

```rust
static MEMORY_ATTRIBUTES_PROTOCOL_HANDLE: AtomicUsize = AtomicUsize::new(protocol_db::INVALID_HANDLE);
static MEMORY_ATTRIBUTES_PROTOCOL_INTERFACE: AtomicPtr = AtomicPtr::new(core::ptr::null_mut());

...
MEMORY_ATTRIBUTES_PROTOCOL_INTERFACE.store(interface, Ordering::SeqCst);

```

dxe_core/src/gcd.rs
os-d: I think you could get into the situation (unlikely but possible) of:
- Loader[Code|Data] Allocator gets an initial allocation which is 1:1 mapped with GCD
- Allocator needs to expand, so Allocator has two memory ranges, but if the memory is contiguous in GCD, it will be merged and so we end up with not 1:1 mapping here.

So that and other edge cases are what led me to write it this way, but also I never expect the attributes to be anything but WB|XP. So I'm open to converting this to an assert, but this felt a bit safer.

Well, actually, I take back that we'll only see WB|XP. We'd only see that if this is the first bootloader we tried to load. If we loaded another one that was nx compat but it failed (or booted the next in the chain that wasn't NX compat), we would see different GCD descriptors with different memory attributes besides XP. But this is really to preserve the caching attributes and I would be surprised if those changed.

dxe_core/src/gcd/spin_locked_gcd.rs
os-d: This was a team decision made when defining the original compatibility mode in mu_basecore. For legacy BIOS, 0xA0000 is where the video memory started, so we decided that anything (read Linux) that attempted to use legacy BIOS regions would only use up to the video memory region. The goal here being to avoid crashes but also keeping intact as many memory protections as possible.

@makubacki or @apop5 may have more color here.

dxe_core/src/image.rs
os-d: Thanks, that was when I was testing :). Will remove.

dxe_core/src/memory_attributes_protocol.rs
os-d: Thanks for the reminder on those, I'll switch over to that

dxe_core/src/gcd/spin_locked_gcd.rs
joschock: Agree that _typical_ use of A/B seg is video MMIO and E/F is mapped ROM, but, I think you could have probes into C/D seg by drivers checking for ancient oprom structs, for example, or even reads of the E/F seg ROM code for "legacy" reasons. When I've dug into it, Linux usually hits < 1MB probing for ancient stuff like EBDA and MPS structs - and if you are gonna allow that, you might as well open up the whole legacy region IMHO. But honestly, I'm okay supporting none of that :), so if team decision is 0xA0000 in compat mode, I won't push back too hard.

dxe_core/src/gcd.rs
joschock: Even in those cases, why wouldn't the ranges returned by `get_memory_ranges_for_memory_type()` be consistent with the GCD-returned descriptors from `get_memory_descriptor_for_address()`? Maybe I misunderstand something about what get_memroy_ranges_for_memory_type() is supposed to return...

os-d: Well, perhaps I’m misunderstanding how the allocator arranges its allocator nodes, but my udnerstanding is that upon each new expand() in the allocator, it will create an allocator node. get_memory_ranges_for_memory_type() returns a Vec of all the ranges contained within allocator nodes within the allocator for that memory type.

If two expands happened that were in contiguous memory (or if the only allocations between two expands were of the same memory type), the GCD would merge this into one descriptor, but the allocator would still have multiple allocator nodes and therefore the Vec returned would have multiple ranges, so you could have a many to one relationship.

Another way to take this would be to get the allocator handle and search the GCD for all descriptors with that handle and remove XP there. That would be less performant, but also this is compat mode so I don’t really care. But I thought, well, if we already have to dip into the allocator to get the handle, we may as well ask it what memory it owns, since it knows.

dxe_core/src/gcd/spin_locked_gcd.rs
os-d: Yeah, your logic makes sense, but compat mode really has been restricted to only what Linux breaks we will fix (which tbh, I don’t love that’s our definition, but it is what it is). So the more restrictive we can be and get away with it is a net positive. This is what mu_basecore does, so I will leave this here.

dxe_core/src/gcd.rs
joschock: Got it. The allocator might have multiple ranges within a given descriptor. I had it mentally flipped; I didn't see how a range could have multiple descriptors (I don't think it can), but that's not what is handled here.

docs/src/integrate/dxe_core.md
```diff
@@ -226,3 +226,20 @@ pub extern "efiapi" fn _start(physical_hob_list: *const c_void) -> ! {
...
+
+### Compatibility Mode
+
+The Rust DXE Core supports Compatibility Mode, the ability to launch bootloaders that do not support memory protections,
+most notably most shipping versions of Linux distros. By default the `compatibility_mode_allowed` feature flag is
+disabled. If a platform wishes to build with it, they must include it in their Cargo.toml as such:
+
+```rust
+[dependencies]
+dxe_core = { version = "9", features = ["compatibility_mode_allowed"] }
```
Javagedes, line 240: I'm a bit confused with the feature name - Is this enabling compatibility mode, or is this allowing the ability for someone to enable compatibility mode? If the latter, how do you enable compatibility mode?

os-d: It is the latter. My next PR (upon return from vacation) is the documentation for memory protections and compat mode.

As in the C core, compat mode is enabled automatically if an EFI_APPLICATION is encountered without the NX_COMPAT dll characteristic set (i.e. a bootloader). However, unlike the C core, we have a build time feature flag here that says whether this FW is allowed to enter compat mode, it is valid for a platform to say it never wants to compromise its security standing. If the feature flag is disabled (compat mode not allowed) when encountering the non NX_COMPAT image, it will fail to load it gracefully.

dxe_core/src/allocator.rs
```diff
@@ -452,7 +469,20 @@ pub fn core_allocate_pages(
...
         }
         Err(err) => Err(err),
+    };
+
+    // If the memory type is runtime services code or data, we need to install the memory attributes table to reflect
+    // the update. The MAT logic will decide if it is a proper time to install the MAT or not.
+    match memory_type {
+        efi::RUNTIME_SERVICES_CODE | efi::RUNTIME_SERVICES_DATA => {
+            if res.is_ok() {
+                MemoryAttributesTable::install();
```
cfernald, line 479: consider renaming this to `::update()` instead, since as far as I can tell this will only ever update it an not "install" it, correct?

os-d: Well, we can't actually update the table. We regenerate it every time and then install the new config table, dropping the old one

docs/src/integrate/dxe_core.md
os-d: @Javagedes , does this clear up your confusion? (with upcoming documentation)

PR #347, opened 2025-04-10
[RFC]: MM Communicate Service

makubacki:
RFC for adding a MM Communicate Service and the necessary dependencies for the service to work.

Current RFC State: Approved


docs/src/rfc/text/0000-mm-comm.md
```diff
@@ -0,0 +1,331 @@
...
+- `MmConfigurationProvider` - This component is responsible for providing MM configuration information to other
+  components. All other MM related components should be abstracted from MM details by the configuration produced here.
+  - Populates and locks `MmConfigurationConfig` gathering information from implementation-defined sources. This may
+    include pre-existing HOBs and/or HW registers.
+- `SwMmiManager` - This component is responsible for providing the `SwMmiTrigger` service that is used to trigger
+  MMIs.
+  - Consumes Configuration: `MmConfigurationConfig`
+  - Optionally Consumes Service: `PlatformMmControl`
+  - Produces Service: `SwMmiTrigger`
+- `MmCommunicator` - This component is responsible for providing the `MmCommunication` service. It is a service that
```
cfernald: I think you mentioned it, but I don't see it called out here. What is the plat for this to be able to persist to runtime as components are exclusively defined as part of the monolithic core today. Should this be called out the in "Unresolved Questions"?

docs/src/rfc/text/0000-mm-comm.md
```diff
@@ -0,0 +1,331 @@
...
+  can be used by other components to send data to MM handlers.
+  - Consumes Configuration: `MmConfigurationConfig`
+  - Consumes Service: `SwMmiTrigger`
+  - Produces Service: `MmCommunication`
+
+### MM Communication Service
+
+```rust
+pub trait MmCommunication {
+    fn communicate(self, buffer: &[u8]) -> Result<(), Status>;
```
cfernald: Does this not need `MessageGuid`? or is `buffer` the entire C struct header?

docs/src/rfc/text/0000-mm-comm.md
```diff
@@ -0,0 +1,331 @@
...
+- `PlatformMmControl` - This service is responsible for providing any additional platform-specific MM functionality
+  needed to prepare the platform to enter MM and set up MMI support.
+
+  ---
+
+- `MmConfigurationProvider` - This component is responsible for providing MM configuration information to other
+  components. All other MM related components should be abstracted from MM details by the configuration produced here.
+  - Populates and locks `MmConfigurationConfig` gathering information from implementation-defined sources. This may
+    include pre-existing HOBs and/or HW registers.
+- `SwMmiManager` - This component is responsible for providing the `SwMmiTrigger` service that is used to trigger
```
cfernald: Is this in hopes to converge with ARM in the future or is there enough variable on x86 that this layer of abstraction is required? I would think it would just be a matter of what IO port to write to, no?

docs/src/rfc/text/0000-mm-comm.md
makubacki: I brought it up verbally. But you're right that it should be in unresolved questions. I will add it.

docs/src/rfc/text/0000-mm-comm.md
makubacki: Yes. This trigger layer is what I'm expecting to be the abstraction to support ARM. x86 should largely be supported with a single implementation based on ACPI offsets to various cmd/sts registers.

docs/src/rfc/text/0000-mm-comm.md
makubacki: I was considering an earlier design that would accept a type that had that info as metadata but thought it was overly complicated, and I forgot to add the `MessageGuid` back here.

docs/src/rfc/text/0000-mm-comm.md
cfernald: Makes sense. Thanks! Might be good to call this out for the motivation of this abstraction layer.

docs/src/rfc/text/0000-mm-comm.md
os-d: This was also my main question and perhaps warrants another RFC on how all runtime components will be handled (I believe the current theory is implementing a dll loading scheme, but this needs to be explored more, of course). Until that framework is decided, we would need to rely on the C version to still be used and persist into runtime, right?

makubacki: > Until that framework is decided, we would need to rely on the C version to still be used and persist into runtime, right?

Correct

docs/src/rfc/text/0000-mm-comm.md
makubacki: Updated.

makubacki:
@cfernald & @os-d, I've made all planned updates and moved the state to FCP. Please revisit for a final review.

makubacki:
QEMU Q35 Implementation: https://dev.azure.com/microsoft/MsUEFI/_git/qemu_rust_bins/pullrequest/12829836

PR #348, opened 2025-04-10
[RFC]: Memory Management Interface

cfernald:
RFC for adding a Memory Management interface.
Migrated from: https://github.com/OpenDevicePartnership/uefi-dxe-core/issues/340

Current RFC State: FCP

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,10 +18,10 @@ edition = "2021"
 rust-version = "1.83.0"
 
 [workspace.dependencies]
-section_extractor = { version = "10" }
-uefi_sdk = { version = "3" }
-uefi_cpu = { version = "10" }
-uefi_debugger = { version = "10" }
+section_extractor = { version = "12" }
+uefi_sdk = { version = "5" }
+uefi_cpu = { version = "12" }
+uefi_debugger = { version = "12" }
 uefi_corosensei = { version = "0.1", default-features = false }
 
 adv_logger = { path = "adv_logger", version = "9" }
--- a/adv_logger/src/component.rs
+++ b/adv_logger/src/component.rs
@@ -23,7 +23,7 @@ use uefi_sdk::{
 use crate::{
     logger::AdvancedLogger,
     memory_log::{self, AdvLoggerInfo},
-    protocol::{AdvancedLoggerProtocol, AdvancedLoggerProtocolRegister},
+    protocol::AdvancedLoggerProtocol,
 };
 
 /// C struct for the internal Advanced Logger protocol for the component.
@@ -128,8 +128,8 @@ where
         };
 
         let protocol = Box::leak(Box::new(protocol));
-        match bs.install_protocol_interface(None, &AdvancedLoggerProtocolRegister, &mut protocol.protocol) {
-            Err((_, status)) => {
+        match bs.install_protocol_interface(None, &mut protocol.protocol) {
+            Err(status) => {
                 log::error!("Failed to install Advanced Logger protocol! Status = {:#x?}", status);
                 Err(EfiError::ProtocolError)
             }
--- a/adv_logger/src/integration_test.rs
+++ b/adv_logger/src/integration_test.rs
@@ -14,7 +14,7 @@ use r_efi::efi;
 use uefi_sdk::boot_services::{BootServices, StandardBootServices};
 use uefi_test::{u_assert, u_assert_eq, uefi_test};
 
-use crate::{memory_log, protocol};
+use crate::{memory_log, protocol::AdvancedLoggerProtocol};
 
 #[uefi_test]
 fn adv_logger_test(bs: StandardBootServices) -> uefi_test::Result {
@@ -23,7 +23,7 @@ fn adv_logger_test(bs: StandardBootServices) -> uefi_test::Result {
 
     // Get a reference to the advanced logger buffer. The actual transport does
     // not matter so use the NULL implementation as a stand-in.
-    let result = unsafe { bs.locate_protocol(&protocol::AdvancedLoggerProtocolRegister, None) };
+    let result = unsafe { bs.locate_protocol::<AdvancedLoggerProtocol>(None) };
 
     u_assert!(result.is_ok(), "adv_logger_test: Failed to locate the advanced logger protocol.");
     let protocol = result.unwrap();
--- a/adv_logger/src/protocol.rs
+++ b/adv_logger/src/protocol.rs
@@ -8,6 +8,7 @@
 //!
 
 use r_efi::efi;
+use uefi_sdk::protocol::ProtocolInterface;
 
 /// C struct for the Advanced Logger protocol version 2.
 #[repr(C)]
@@ -23,27 +24,12 @@ pub struct AdvancedLoggerProtocol {
     pub(crate) log_info: efi::PhysicalAddress,
 }
 
-/// Structure for registering and locating the Advanced Logger protocol.
-pub struct AdvancedLoggerProtocolRegister;
-
 /// Function definition for writing a log message to the Advanced Logger through
 /// the protocol.
 type AdvancedLoggerWrite = extern "efiapi" fn(*const AdvancedLoggerProtocol, usize, *const u8, usize) -> efi::Status;
 
-unsafe impl uefi_sdk::protocol::Protocol for AdvancedLoggerProtocolRegister {
-    type Interface = AdvancedLoggerProtocol;
-
-    fn protocol_guid(&self) -> &'static efi::Guid {
-        &AdvancedLoggerProtocol::GUID
-    }
-}
-
-impl core::ops::Deref for AdvancedLoggerProtocolRegister {
-    type Target = efi::Guid;
-
-    fn deref(&self) -> &Self::Target {
-        &AdvancedLoggerProtocol::GUID
-    }
+unsafe impl ProtocolInterface for AdvancedLoggerProtocol {
+    const PROTOCOL_GUID: efi::Guid = AdvancedLoggerProtocol::GUID;
 }
 
 impl AdvancedLoggerProtocol {
--- a/crates/uefi_performance/src/_smm.rs
+++ b/crates/uefi_performance/src/_smm.rs
@@ -2,7 +2,7 @@
 //!
 //! This module also contain smm performance communicate structures that define the communicate buffer data that need to be used to fetch perf records from smm.
 
-use core::{debug_assert_eq, marker::PhantomPinned, ops::Deref, ptr, result::Result::Ok, slice};
+use core::{debug_assert_eq, marker::PhantomPinned, ptr, result::Result::Ok, slice};
 
 use r_efi::efi;
 
@@ -10,7 +10,7 @@ use scroll::{
     ctx::{TryFromCtx, TryIntoCtx},
     Endian, Pread, Pwrite,
 };
-use uefi_sdk::{base::UEFI_PAGE_SIZE, protocol::Protocol};
+use uefi_sdk::{base::UEFI_PAGE_SIZE, protocol::ProtocolInterface};
 
 pub const EFI_SMM_COMMUNICATION_PROTOCOL_GUID: efi::Guid =
     efi::Guid::from_fields(0xc68ed8e2, 0x9dc6, 0x4cbd, 0x9d, 0x94, &[0xdb, 0x65, 0xac, 0xc5, 0xc3, 0x32]);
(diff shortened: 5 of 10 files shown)


docs/src/rfc/text/0000-memory-management-interface.md
```diff
@@ -0,0 +1,311 @@
+# RFC: Create Stable Memory Management API/Service
+
+To support scaling both within the core and with the create of Components, stable APIs
```
os-d:
```suggestion
To support scaling both within the core and throughout Components, stable APIs

```

I think is what you mean, originally wording didn't quite make sense.

docs/src/rfc/text/0000-memory-management-interface.md
```diff
@@ -0,0 +1,311 @@
...
+- Where to use UEFI definitions, versus unique definitions in Core APIs
+
+## Prior Art (Existing PI C Implementation)
+
+This design takes inspiration from the [UEFI Memory Allocation APIs](https://uefi.org/specs/UEFI/2.10/07_Services_Boot_Services.html#memory-allocation-services)
+and the UEFI Memory attributes protocol. These are used to establish existing use
+cases but should be improved upon by taking advantage of rust to create a safer and
+more extensible API.
+
+For allocation management, use [Allocator trait](https://doc.rust-lang.org/std/alloc/trait.Allocator.html)
```
os-d:
```suggestion
For allocation management, the [Allocator trait](https://doc.rust-lang.org/std/alloc/trait.Allocator.html)

```

docs/src/rfc/text/0000-memory-management-interface.md
```diff
@@ -0,0 +1,311 @@
...
+can be referenced for prior art on abstracting arbitrary allocations.
+
+## Alternatives
+
+There are a large number of possible ways the specific interface could be implemented,
+but the most significant alternative evaluated is related to the services paradigm
+for core services such as memory management. The alternative to this would be using
+statis routines.
```
os-d:
```suggestion
static routines.

```

docs/src/rfc/text/0000-memory-management-interface.md
```diff
@@ -0,0 +1,311 @@
...
+
+Alternatives to the Service model would be to instead expose functions to the components
+and internally in the core. Generally, one downside of this is that relying on static
+routines can be significantly more difficult to unit test because this requires some
+global locking during the tests and makes creating mocks more difficult depending
+on the implementation. These static routines could be implemented in the core or
+in the SDK wrapping a global dyn reference.
+
+Exposing this from the core would additionally mean that the components must take
+dependency on a specific implementation would only further complicate the testing
```
os-d:
```suggestion
dependency on a specific implementation that would only further complicate the testing

```

docs/src/rfc/text/0000-memory-management-interface.md
```diff
@@ -0,0 +1,311 @@
...
+    memory_manager: &'static dyn MemoryManager,
+}
+
+impl PageAllocation {
+    pub fn into_raw_ptr<T>(self) -> *mut T { ... }
+    pub fn page_count(&self) -> usize { ... }
+    pub fn try_into_box<T>(self, value: T) -> Option<Box<T, PageFree>> { ... }
+    pub fn into_box<T>(self, value: T) -> Box<T, PageFree> { ... }
+    pub fn try_leak_as<T>(self, value: T) -> Option<&'static T> { ... }
+    pub fn leak_as<T>(self, value: T) -> &'static T { ... }
```
os-d: With the failable functions `into_box` or `leak_as` do we even want to offer them? Or only offer the `try_*` versions? I know it is a common Rust pattern to offer both, but for core functionality that has to work for every user, for our service we may want to say, no, you have to do this the safe way and get an option back, we won't let you panic, or fail however. Just a question, I lean towards not allowing the possibly failing options, unless we have a good reason to.

docs/src/rfc/text/0000-memory-management-interface.md
```diff
@@ -0,0 +1,311 @@
...
+The following enums were created to represent the Access and Caching attributes
+of memory exposed to the caller.
+
+```rust
+pub enum AccessType {
+    NoAccess,
+    ReadOnly,
+    ReadWrite,
+    ReadExecute,
+    ReadWriteExecute,
```
os-d: I wonder if we start without allowing this? I would rather the core be in complete control of making RWX pages and that that would only happen with compat mode or the resc desc hob V2.

docs/src/rfc/text/0000-memory-management-interface.md
```diff
@@ -0,0 +1,311 @@
...
+pub enum CachingType {
+    Uncached,
+    WriteCombining,
+    WriteBack,
+    WriteThrough,
+}
+```
+
+These definitions were chosen to more accurate reflect hardware then the definitions
+currently used in the UEFI specification. Notable, there is no current notion of
```
os-d:
```suggestion
currently used in the UEFI specification. Notably, there is no current notion of

```

docs/src/rfc/text/0000-memory-management-interface.md
```diff
@@ -0,0 +1,311 @@
...
+what memory is "capable" of, but only what is actually configured. Additionally,
+There is no concept of "read-protect" as this does not accurate reflect what hardware
+supports. Instead `NoAccess` was chosen as the implementation will treat this as
+unmapped and so it does not make sense to allow this in conjuction with other
```
os-d:
```suggestion
unmapped and so it does not make sense to allow this in conjunction with other

```

docs/src/rfc/text/0000-memory-management-interface.md
```diff
@@ -0,0 +1,311 @@
...
+
+```rust
+pub fn component (memory_manager: Service<dyn MememoryManager>) ->Result<()> {
+    .
+}
+```
+
+### Page Allocations
+
+with the memory manager, the component can allocate typed memory, pages, and otherwise
```
os-d:
```suggestion
With the memory manager, a component can allocate typed memory, pages, and otherwise

```

docs/src/rfc/text/0000-memory-management-interface.md
```diff
@@ -0,0 +1,311 @@
...
+alter/inspect memory state. A common use case of this would be allocate memory pages.
+Allocations can be initialized as several different types; smart pointers, static types,
```
os-d:
```suggestion
Allocations can be initialized as several different types: smart pointers, static types,

```

docs/src/rfc/text/0000-memory-management-interface.md
```diff
@@ -0,0 +1,311 @@
...
+or raw pointers.
+
+```rust
+// Create a boxed type, this will be free'd automatically.
```
os-d:
```suggestion
// Create a boxed type, this will be freed automatically.

```

docs/src/rfc/text/0000-memory-management-interface.md
```diff
@@ -0,0 +1,311 @@
...
+```rust
+let options = AllocationOptions::default().with_memory_type(EfiMemoryType::BootServicesData).with_alignment(0x2000);
+let allocation = memory_manager.allocate_pages(1, options)?;
+```
+
+### Heap Allocations
+
+Another use case of the memory manager is to make heap allocations within a specific memory
+type. This is done by first acquiring the `Allocator` then using the heap allocator to
+create object.
```
os-d: Perhaps worth noting that the default memory type is BootServicesData and any implicit heap allocations will target that, without needing to go through this new process. At least I'm assuming that's how it will be wired up?

docs/src/rfc/text/0000-memory-management-interface.md
```diff
@@ -0,0 +1,311 @@
...
+
+Some operations require changing the attributes of memory, such as access permissions and caching.
+For these operations, the memory manager exposes the `set_page_attributes` and `get_page_attributes`
+routines.
+
+```rust
+// Change a page to be inaccessible, leaving the caching unchanged.
+memory_manager.set_page_attributes(address, 1, AccessType::NoAccess, None)?;
+
+// Check that the access type was changes
```
os-d:
```suggestion
// Check that the access type was changed

```

docs/src/rfc/text/0000-memory-management-interface.md
cfernald: I'm fine just returning a `option` and it's up to the user to decide how they want to handle that. I imagine most users will just unwrap it as they probably allocated to be `pages_from_bytes(size_of::<T>())` anyways, but they have to explicitly decide that on their end. 

In the end, it will be easier to add new functions later then to take away unwanted ones, so I'm fine minimizing the first set.

docs/src/rfc/text/0000-memory-management-interface.md
joschock: Do you mean memory attributes in general? Or RWX in particular? In my view, it's too constraining for the core not to have _some_ mechanism to allow memory region attributes to be modified with some degree of freedom. It's fine to for core to enforce sanity policy (i.e. maybe no-execute on stack pages seems reasonable). 

However, I think there's going to be use cases where it makes sense for a caller to want RWX on a buffer. For example, maybe "John's Toy OS loader" Component wants to `AllocatePages` a buffer to decompress its own blob of executable code containing a microkernel paging service, attribute it RWX, and jump into it before calling EBS and taking over control of MMU from the core. Is that something we want to specifically disallow unless it's written as a PE/COFF and loaded via `load_image`? 

To summarize my position: I think it's fine to require all attribute requests to go through a core API and to reject some of them on the basis of opinionated core policy that we decide. But I don't think I'm in favor of a "no request for RWX shall ever be honored" policy.

docs/src/rfc/text/0000-memory-management-interface.md
os-d: Yeah, I think that pushing the panic into the consuming code is the ideal pattern, it is much more debuggable that way.

makubacki: I agree with minimizing the number of functions and just use `Option`.

docs/src/rfc/text/0000-memory-management-interface.md
os-d: Yeah, so my wondering is that we currently provide (well almost currently provide, the resc desc v2 HOB doesn't support access attributes today, but it will be extended to to also support the special memory region HOB that the C codebase uses) 4 ways to create RWX regions:
- Try to boot an EFI_APPLICATION without the NX_COMPAT DLL Characteristic, in which case we enter compat mode and all new memory allocated is RWX
- Use the resc desc HOB v2 to define a memory region that is RWX (which could be reserved in PEI and allocated by address in DXE, etc)
- Use the memory attribute protocol to mark a region RWX
- Use the CPU arch protocol to mark a region RWX

And from the core perspective we don't want any RWX regions, if possible. So if you were writing John's toy OS loader as a Rust component and wanted to consume this interface, I would say, that is not a supported use case, you do need to have a PE/COFF image that is what UEFI jumps to and then you can do whatever you want (which is already what happens with OS loaders).

So, this is isn't restricting any RWX region, but more saying if we are creating a new interface and moving forward this is how components should interact with memory attributes then I would say, yeah, let's not give it as an option. You should not create components like that (and it is conceivable in the future that in a V2 where we have a userspace/kernelspace model that you couldn't do that, etc.). But, you are definitely not restricted in not doing so, there are many such ways you can craft a region for yourself. In your example, the memory attribute protocol would be the way to go, it is designed for bootloaders to do exactly this (well ideally be more restrictive but they can do whatever they want).

docs/src/rfc/text/0000-memory-management-interface.md
cfernald: Will do, though I will be adding some functions for slices, is this ia useful use-case that can be safely wrapped as well.

docs/src/rfc/text/0000-memory-management-interface.md
cfernald: @os-d, The problem with restricting valid input through the type is that is doesn't work so well if that state is possible at all. That is, if we do `get_attributes` on a page that IS RWX because of compat mode or some other way, we need to be able to return that value, and so unless we have multiple enums then the value has to exist.

I think it is perfectly fine for us to return `MemoryError::UnsupportedAttributes` though in the case a caller tries to set memory RWX.

os-d: Sure, that makes sense. I would advocate for returning UnsupportedAttributes, but that's an implementation detail, so I'll resolve this.

cfernald: In fact, I have this called out in the draft PR of this interface.
```rust
    /// Read-write-execute access. This type of access is generally unsafe and
    /// may not be supported for all operations.
    ReadWriteExecute,
```

joschock: > But, you are definitely not restricted in not doing so, there are many such ways you can craft a region for yourself. In your example, the memory attribute protocol would be the way to go, it is designed for bootloaders to do exactly this (well ideally be more restrictive but they can do whatever they want).

This is okay with me, but it does mean that Component authors who want to do outlandish things like this will basically pull the eject handle and fall back to the traditional UEFI protocols and core interactions for these kind of use cases. If we do end up adding this type of restriction, might be worth a clarifying comment as to the expected alternatives (i.e. fallback to UEFI) that are available if someone finds that their design requires this flexibility.

cfernald:
@OpenDevicePartnership/uefi-contributors - Moving to Final Comment Period.

PR #349, opened 2025-04-10
[RFC] dependency injectable Guided HOB parsing

Javagedes:
RFC for adding a new dependency injectable type `Hob`, which when found
in a registered component, also registers a parser for said HOB. This parser
will be used to parse a guided HOB into the `Hob<T>` type that the
component is requesting. The component will only be executed if the HOB
is found.

Current RFC State: **FCP**

### DIFF
--- a/docs/src/rfc/text/0001-guided-hob-to-hob-param.md
+++ b/docs/src/rfc/text/0001-guided-hob-to-hob-param.md
@@ -1,9 +1,9 @@
 # RFC: `Guided Hob to Hob<T>`
 
 This is a request for comments for a design to allow a platform to register functionality with the core that will parse
-a guided hob into a specific struct and register that struct instance with the Core to be accessible as a `Hob<T>`
-struct. This implementation will remove the need for a Component to parse the hoblist manually before registering
-itself with the core, and instead moves the parsing to the core.
+a guided HOB (Hand-Off Block) into a specific struct and register that struct instance with the Core to be accessible
+as a `Hob<T>` struct. This implementation will remove the need for a Component to parse the HOB list manually before
+registering itself with the core, and instead moves the parsing to the core.
 
 ## Change Log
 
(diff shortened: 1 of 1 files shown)


docs/src/rfc/text/0000-guided-hob-to-config.md
```diff
@@ -0,0 +1,127 @@
...
+
+    #[derive(IntoConfig, Debug, Clone, Copy)]
+    #[hob = "8be4df61-93ca-11d2-aa0d-00e098032b8c"]
+    struct MyOtherHobConfig;
+
+    // In entry point
+    Core::default()
+        .init_memory(physical_hobList)
+        .with_hob_parser::<MyHobConfig>()
+        .with_hob_parser::<MyOtherHobConfig>()
```
cfernald: Will there be default included parsers?

I would think since GUIDed hobs should be globally unique in their structure that instead of requiring the core to enumerate all of the parsers it may be more scalable to have a default list that the platform may add/edit from. It would seems cumbersome for the platforms to root out and include all parsers they may need.

Perhaps this is a good use-case for the bundles you had previously proposed.

Javagedes: Yes I think a default list of parsers with the ability to add more for people writing custom hobs, will be a good addition to this.

Javagedes: and I agree, bundles would help with this, as it would allow a single feature to add multiple components, configs, and config parsers

Javagedes: I have updated my uefi_sdk PR to restructure configuration by creating a new module at `uefi_sdk::component::config` that will allow us to add in spec defined Guid HOB Config structs (and any others we want to standardize) there. Please let me know if there are any hob structs you would like to see added as a part of this RFC.

cfernald: > and I agree, bundles would help with this, as it would allow a single feature to add multiple components, configs, and config parsers

This would also be helpful for components that are tied to GUIDed hobs, so that the platform can just include it's bundle instead of having to remember to add them inidividually. Though maybe that was your point.

cfernald: > I have updated my uefi_sdk PR to restructure configuration by creating a new module at `uefi_sdk::component::config` that will allow us to add in spec defined Guid HOB Config structs (and any others we want to standardize) there. Please let me know if there are any hob structs you would like to see added as a part of this RFC.

These would still have to be manually added by the platform though, right? At least until we had bundles or some similar mechanisms?

Javagedes: Even without the bundle logic (which I don't plan on adding as a part of this rfc), It will be trivial to add a default set of hob config parsers. We would define them in uefi_sdk so that they are globally available, then uefi-dxe-core would just register whatever set or subset of the structs we want.

makubacki:
Capturing offline feedback here. Multiple instances of GUIDed HOBs need to be supported. We decided a `Hob<T>` `Param` type like `Config` but supporting multiple instances where `Deref` gets the first instance and `IntoIter` allows iterations makes sense. `Hob<T>` won't provide a `Default`. `Option<Hob<T>>` would be used instead. With this change, `Config<T>` will no longer support HOB conversion and stay single instance.

os-d:
> Capturing offline feedback here. Multiple instances of GUIDed HOBs need to be supported. We decided a `Hob<T>` `Param` type like `Config` but supporting multiple instances where `Deref` gets the first instance and `IntoIter` allows iterations makes sense. `Hob<T>` won't provide a `Default`. `Option<Hob<T>>` would be used instead. With this change, `Config<T>` will no longer support HOB conversion and stay single instance.

@Javagedes are you updating the RFC to include these changes? If so I will wait to review until then.

Javagedes:
> > Capturing offline feedback here. Multiple instances of GUIDed HOBs need to be supported. We decided a `Hob<T>` `Param` type like `Config` but supporting multiple instances where `Deref` gets the first instance and `IntoIter` allows iterations makes sense. `Hob<T>` won't provide a `Default`. `Option<Hob<T>>` would be used instead. With this change, `Config<T>` will no longer support HOB conversion and stay single instance.
> 
> @Javagedes are you updating the RFC to include these changes? If so I will wait to review until then.

@os-d RFC updated to include this.

Javagedes:
@OpenDevicePartnership/uefi-contributors - Moving to Final Comment Period

makubacki:
Approval for current state of the RFC in FCP

PR #350, opened 2025-04-10
[RFC]: Move CPU Init into Core

Javagedes:
RFC for moving the CPU initilization code into the core and instead exposing minimal configuration knobs as needed for platforms.

Current RFC State: FCP

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -66,6 +66,3 @@ opt-level = 3
 
 [profile.test]
 opt-level = 0
-
-[patch.crates-io]
-uefi_cpu = { path = "C:\\src\\uefi-core\\uefi_cpu"}


docs/src/rfc/text/0000-move-cpu-into-core.md
```diff
@@ -0,0 +1,112 @@
...
+
+Currently, traits generics are used in the Patina Core's struct defintion to support the ability to (1) replace certain
+functionality based off of a platform's requirements and (2) replace cpu architecture specific functionality. As time
+has continued and the core has evolved, we have noted that platforms do not need to customize this functionality; all
+platforms of a certain architecture will always use the same underlying architecture support code. Exposing this to the
+consumer only works to complicate the core initialization and has been deemed unecessary.
+
+This proposal is to remove the architecture specific customization from the public core interface, and automatically
+use the appropriate logic for the given architecture. Configuration knobs can be provided to the core to fine tune
+this logic for a given platform.
```
cfernald: I think we should take this further. This seems reflective of some loosely defined abstraction layers. I would suggest we use the RFC to define those layers for CPU support. e.g. "Code for ISA/Architecture defined functionality should not be overiden, and new ISA Architecture support changes should be made in the core"

docs/src/rfc/text/0000-move-cpu-into-core.md
```diff
@@ -0,0 +1,112 @@
...
+## Prior Art (Existing PI C Implementation)
+
+In the current design, each of the three implementations must be registered with the core using the appropriate
+`.with_*` method, which allows for the registration of a configured initializer
+
+## Alternatives
+
+Switch to a standardized struct instead of trait generics, for initialization.
+
+## Rust Code Design
```
vineelko: I think, maybe adding a "Before RFC" and "After RFC" code blocks will help the reader understand how the code is being changed/enhanced.

docs/src/rfc/text/0000-move-cpu-into-core.md
```diff
@@ -0,0 +1,112 @@
...
+## Goals
+
+1. Remove as many trait generics from the core as possible
+2. Allow configuration for cpu initialization
+
+## Requirements
+
+1. remove `.with_cpu_init` method and `EfiCpuInit` trait from the `Core`'s public interface.
+2. remove `.with_interrupt_manager` method and `InterruptManager` trait from the `Core`'s public interface.
+3. remove `.with_interrupt_bases` method and `InterruptBases` trait from the `Core`'s public interface.
```
cfernald: We would still need to be able to customize the interrupt bases, right? This is SoC specific and not ISA specific IIRC.

docs/src/rfc/text/0000-move-cpu-into-core.md
```diff
@@ -0,0 +1,112 @@
...
+    pub fn init_memory(
+        mut self,
+        physical_hob_list: *const c_void,
+    ) -> Core<SectionExtractor, Alloc> {
+        #[cfg(target_os = "uefi")]
+        self.cpu_init()
+
+        /* Continue as normal */
+
+        #[cfg(all(target_os = "uefi", target_arch = "aarch64"))]
```
vineelko: Hope these won't get littered everywhere!

docs/src/rfc/text/0000-move-cpu-into-core.md
Javagedes: Yes, I expect some configuration to be necessary, either now or in the future. I would like to see a config knob be provided via the core rather than passing in a fully configured interrupt bases initializer object

docs/src/rfc/text/0000-move-cpu-into-core.md
```diff
@@ -0,0 +1,112 @@
...
+        hw_interrupt_protocol::install_hw_interrupt_protocol(&mut InterruptManagerAarch64::default(), &self.interrupt_bases);
+
+        /* Continue as normal */
+    }
+}
+
+```
```
cfernald: Do we want to merge those crates into the core (crate and/or repo) as part of this dependency change as it would no longer be a point of extension and so serves no purpose being separate?

Javagedes: I would not be against that, but I think there might be some pushback from uefi_cpu owners which is mainly @vineelko and maybe a bit of @os-d / @kuqin12

docs/src/rfc/text/0000-move-cpu-into-core.md
Javagedes: Will add now.

docs/src/rfc/text/0000-move-cpu-into-core.md
vineelko: +1. This may or may not be related to this RFC, but to this day, I still find the word 'Core' confusing—especially since we're referring to the 'uefi-dxe-core' repo as 'Core' while 'uefi-core' means something else.

Adding this comment again. Not sure how it was deleted :-)

docs/src/rfc/text/0000-move-cpu-into-core.md
Javagedes: Yes, we should be able to properly design this such that we have two modules for arch specific things - one for x64 and one for aarch64, and just do something like

`#[cfg(all(target_os = "uefi", target_arch = "x64"))]
pub mod x64_thing as thing;
`#[cfg(all(target_os = "uefi", target_arch = "aarch64"))]
pub mod aarch64_thing as thing;

docs/src/rfc/text/0000-move-cpu-into-core.md
vineelko: Consolidation is always good I am up for it. Remember my original implementation of uefi_cpu in uefi-dxe-core?
Back to the future: https://dev.azure.com/microsoft/MsUEFI/_git/DxeRust/pullrequest/11637315 :wink:

docs/src/rfc/text/0000-move-cpu-into-core.md
Javagedes: Added :) Thanks!

docs/src/rfc/text/0000-move-cpu-into-core.md
Javagedes: I agree its pretty confusing. I've been calling it Patina or Patina core now. Its just the struct is still named "Core". I'll update the RFC for clarity

makubacki:
Mentioned offline but (on x86 at least), platforms need the ability to still register custom exception handlers. At some point we also need to abstract a timer service that may depend on the CPU timer.

docs/src/rfc/text/0000-move-cpu-into-core.md
makubacki: We still want generic CPU functionality to be reusable outside the DXE Core (e.g. by PEI Core, MM Core, maybe something like SEA written in Rust). This thread is a little ambiguous in terms of what is referring to exactly, but is that still possible in what is being proposed?

Javagedes: Only if we made the Patina Core re-usable across targets (Dxe, mm, pei, etc) and moved the DXE related things (like setting up the tables, dispatch, etc) into components registered with the generic "Patina Core" (minikernel)

Javagedes:
> Mentioned offline but (on x86 at least), platforms need the ability to still register custom exception handlers. At some point we also need to abstract a timer service that may depend on the CPU timer.

I've updated the requirements to match the expectation of being able to register custom exception handlers.

docs/src/rfc/text/0000-move-cpu-into-core.md
cfernald: This may be a topic for a different RFC then, my thought being rather than making the core dependencies re-usable, make the core re-usable. But we can differ this conversation for such a topic in the future.

Javagedes: This is my end goal of the Patina Core, to be reusable, but I agree - that would need to be another RFC and is a large undertaking.

joschock:
A couple of general comments in defense of the present design point:

1. One of the main things I would be worried about with this RFC is that I want to maintain a hard and clean abstraction on the CPU in the core. By moving this implementation into the core, I'm worried that we dilute that interface, and architectural quirks will slowly work their way into the core as a result of losing the separation this abstraction requires. I'm not saying necessarily that we should have a trait as implemented today. But I do value having a well defined interface separation between the CPU code and the Core. 
2. By extension, a well-defined separation between the core and the CPU support facilitates adding support for new CPU architectures. In some sense, this trait serves as the "implement this for new architectures" spec point. If we subsume this trait into the core, adding a new arch has to deal with the whole core (and any assumptions it has made around existing architectures, since they are now "implementation details") rather than implementing the well-defined interface (or requesting that it be extended for well-defined reasons). 

To be clear, my comments are in defense of having a well-defined interface abstraction and enforcing low coupling between CPU arch support and core. As to the particulars of how that is enforced/implemented, I am less opinionated. If we want to move away from the trait implementation as proposed in this RFC, I don't have a problem with that, as long as we have a good answer for implementation separation and abstraction between the non-CPU-arch dependent part of the core and the CPU-dependent part of the core.

docs/src/rfc/text/0000-move-cpu-into-core.md
makubacki: Yeah. That's fine to propose but not what we're doing now, and its proposal would have to answer open questions about supporting other environments. My request is to (1) not let CPU details leak outside well-defined CPU interfaces and that can still happen with moving around where the code resides as proposed here and (2) acknowledge the constraints of the RFC proposal based on today's state like including this:

> "Only if we made the Patina Core re-usable across targets (Dxe, mm, pei, etc) and moved the DXE related things (like setting up the tables, dispatch, etc) into components registered with the generic "Patina Core" (minikernel)"

Javagedes:
> A couple of general comments in defense of the present design point:

Right now, this RFC is simply to have the core to automatically select the appropriate implementation based off of cpu architecture, rather than allow a platform to register a specific implementation with the core. As it stands, we will keep the cpu logic separated and in the `uefi_core` repo with the appropriate interfaces.

Should the time come where we have made the Patina Core appropriately re-usable across other phases like PEI, mm, etc, we may consider moving that code into this repository. Even then, I would suspect we would keep it as a standalone crate, just in this repository.

docs/src/rfc/text/0000-move-cpu-into-core.md
cfernald: Is the intention to solve that in this RFC? I assume so, as we wont be able to implement this for Arm platforms until an alternative exists.

Javagedes: Yes requirements 5 (which you cannot see in this comment thread, and was added later) is the requirement to provide configuration knob support for platforms

docs/src/rfc/text/0000-move-cpu-into-core.md
```diff
@@ -0,0 +1,190 @@
+# RFC: `Move CPU functionality into Patina Core`
+
+Currently, traits generics are used in the Patina Core's struct defintion to support the ability to (1) replace certain
+functionality based off of a platform's requirements and (2) replace cpu architecture specific functionality. As time
+has continued and the Patina Core has evolved, we have noted that platforms do not need to customize this
+functionality; all platforms of a certain architecture will always use the same underlying architecture support code.
+Exposing this to the consumer only works to complicate the Patina Core initialization and has been deemed unecessary.
```
os-d:
```suggestion
Exposing this to the consumer only works to complicate the Patina Core initialization and has been deemed unnecessary.

```

docs/src/rfc/text/0000-move-cpu-into-core.md
```diff
@@ -0,0 +1,190 @@
...
+## Requirements
+
+1. remove `.with_cpu_init` method and `EfiCpuInit` trait from the `Core`'s public interface.
+2. remove `.with_interrupt_manager` method and `InterruptManager` trait from the `Core`'s public interface.
+3. remove `.with_interrupt_bases` method and `InterruptBases` trait from the `Core`'s public interface.
+4. Automatically select and use the given `EfiCpuInit`, `InterruptManager`, and `InterruptBases` code based off the
+   compilation target architecture.
+5. Provide generic configuration knob support for platforms to fine-tune these initializations where necessary.
+6. Provide a way for a platform (and possibly a component) to register an exception handler.
+7. Continue supporting cpu timer interrupt hook.
```
os-d:
```suggestion
7. Continue supporting the CPU timer interrupt hook.

```

docs/src/rfc/text/0000-move-cpu-into-core.md
Javagedes: closing as we will keep the cpu crates out of this repository for this particular RFC. I think it would be good to do long term thoug.

docs/src/rfc/text/0000-move-cpu-into-core.md
Javagedes: @cfernald  (or others)

Do we suspect other cpu specific configurations to be needed down the road? I'm considering having a single struct with conditionally compiled fields depending on the cpu architecture. Something like below. But if this is the only thing we will ever need then I don't see a need to do something this complex (not that it is that complex).

``` rust
#[derive(Default)]
pub struct ArchConfig {
    #[cfg(all(target_arch = "aarch64", target_os = "uefi"))]
    pub gicd_base,
    #[cfg(all(target_arch = "aarch64", target_os = "uefi"))]
    pub gicr_base,
}

impl <S>Core<S, MemoryState>
where
    S: fw_fs::SectionExtractor + Default + Copy + 'static
{
    fn init_cpu(&self, config: CpuConfig) -> Result<()> {
        #[cfg(all(target_os = "uefi", target_arch = "x86_64"))]
        {
            use uefi_cpu::interrupts::InterruptManager;
            use uefi_cpu::cpu::EfiCpuInit;
            uefi_cpu::cpu::EfiCpuInitX64::default().initialize()?;
            uefi_cpu::interrupts::InterruptManagerX64::default().initialize()?;
        }
        #[cfg(all(target_os = "uefi", target_arch = "aarch64"))]
        {
            use uefi_cpu::interrupts::InterruptManager;
            use uefi_cpu::cpu::EfiCpuInit;
            uefi_cpu::cpu::EfiCpuInitAarch64::default().initialize()?;
            uefi_cpu::interrupts::InterruptManagerAarch64::default().initialize()?;
            // Using the config below)
            let (gicd_base, gicr_base) = config.interrupt_bases;
            uefi_cpu::interrupts::InterruptBasesAarch64::new(gicd_base, gicr_base);
        }
        Ok(())
    }
}

/// For platforms with no overrides, it becomes this
Core::default().init_cpu(CpuConfig::default()).init_memory(...)

/// For platforms with overrides, we could do
Core::default()
    .init_cpu(CpuConfig {
        gicd_base: 0x40060000,
        gicr_base: 0x40080000,
        .. Default::default(),
     ))
    .init_memory(...)
```

Javagedes: Talked offline. We don't expect many CPU specific configurations that will require conditional configuration. Due to this we will refrain form a CpuConfig object and only have the few peices of config.

Javagedes:
Thank you for the reviews @cfernald @vineelko @os-d 

I've updated the RFC to account for recent changes I made to uefi_cpu and addressed some outstanding things. Please feel free to review at your convienience.

docs/src/rfc/text/0000-move-cpu-into-core.md
```diff
@@ -0,0 +1,174 @@
...
+3. Expose `Cpu` trait as a service (`Service<dyn Cpu>`) which has the interface `flush_data_cache`, `init`,
+   `get_timer_value`
+4. Expose `InterruptManager` trait as a service (`Service<dyn InterruptManager>`) which has the interface
+   `register_exception_handler` and `unregister_exception_handler`.
+5. Update cpu_arch protocol and hw_interrupt protocol to use Services instead of references to the trait object.
+
+## Unresolved Questions
+
+1. Do we want to update the `Cpu` or `InterruptManager` trait interfaces?
+2. Do we want to move the `Cpu` or `InterruptManager` traits to another location (uefi-sdk)?
```
vineelko: I think it's better to leave them as-is for now, since moving them to uefi-sdk might add more dependencies between the repos, even though uefi-core is already using uefi-sdk. May be we can revisit this if a future need arises.

Javagedes:
Moving to FCP

PR #351, opened 2025-04-11
Finalize performance

magravel:
## Closed in favor of this PR:
https://github.com/OpenDevicePartnership/patina/pull/405

## Description

- Convert performance lib into a component.
- Add unit tests and refactored the code to be more testable

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Unit test / Run in QEMU / Run on Intel devkit

## Integration Instructions

This Performance component replaced DxeCorePerformanceLib.c. This means that if the device is not built using performance, this component should not be added and if it is built with performance, it should be there. This is because the component publishes the fbpt but when perf is not enabled, FirmwarePerformanceDxe.c creates that table. This is the PCD that tells whether perf is enabled or not: `gEfiMdePkgTokenSpaceGuid.PcdPerformanceLibraryPropertyMask`

### DIFF
--- a/dxe_core/src/lib.rs
+++ b/dxe_core/src/lib.rs
@@ -416,28 +416,16 @@ where
 
         let boot_services_ptr;
         let runtime_services_ptr;
-        let system_table_ptr;
         {
             let mut st = systemtables::SYSTEM_TABLE.lock();
             let st = st.as_mut().expect("System Table is not initialized!");
             boot_services_ptr = st.boot_services_mut() as *mut efi::BootServices;
             runtime_services_ptr = st.runtime_services_mut() as *mut efi::RuntimeServices;
-            system_table_ptr = st.system_table() as *const efi::SystemTable;
         }
 
         tpl_lock::init_boot_services(boot_services_ptr);
 
         memory_attributes_table::init_memory_attributes_table_support();
-
-        // This is currently commented out as it is breaking top of tree booting Q35 as qemu64 does not support
-        // reading the time stamp counter in the way done in this code and results in a divide by zero exception.
-        // Other cpu models crash in various other ways. It will be resolved, but is removed now to unblock other
-        // development
-        _ = uefi_performance::init_performance_lib(
-            &self.hob_list,
-            // SAFETY: `system_table_ptr` is a valid pointer that has been initialized earlier.
-            unsafe { system_table_ptr.as_ref() }.expect("System Table not initialized!"),
-        );
 
         // Add Boot Services and Runtime Services to storage.
         // SAFETY: This is valid because these pointer live thoughout the boot.


makubacki:
@magravel, I will update the SMM code. My understanding is that you can move the runtime status code protocol code to mu_rust_pi if it matches the PI Spec definition https://uefi.org/specs/PI/1.9/V2_DXE_Runtime_Protocols.html#status-code-runtime-protocol.

Anything else you're planning before taking the PR out of draft?

magravel:
> @magravel, I will update the SMM code. My understanding is that you can move the runtime status code protocol code to mu_rust_pi if it matches the PI Spec definition https://uefi.org/specs/PI/1.9/V2_DXE_Runtime_Protocols.html#status-code-runtime-protocol.
> 
> Anything else you're planning before taking the PR out of draft?

Ready to be reviewed now.

magravel:
This PR needs this PR to be merged to build: https://github.com/OpenDevicePartnership/uefi-sdk/pull/81

crates/uefi_performance/src/_smm.rs
```diff
@@ -1,71 +1,61 @@
...
+
+use uefi_sdk::{base::UEFI_PAGE_SIZE, component::hob::FromHob, protocol::ProtocolInterface};
 
 pub const EFI_SMM_COMMUNICATION_PROTOCOL_GUID: efi::Guid =
     efi::Guid::from_fields(0xc68ed8e2, 0x9dc6, 0x4cbd, 0x9d, 0x94, &[0xdb, 0x65, 0xac, 0xc5, 0xc3, 0x32]);
-pub const EDKII_PI_SMM_COMMUNICATION_REGION_TABLE_GUID: efi::Guid =
-    efi::Guid::from_fields(0x4e28ca50, 0xd582, 0x44ac, 0xa1, 0x1f, &[0xe3, 0xd5, 0x65, 0x26, 0xdb, 0x34]);
 
-#[derive(Debug)]
+// GLOBAL_REMOVE_IF_UNREFERENCED EFI_GUID gMmCommonRegionHobGuid = { 0xd4ffc718, 0xfb82, 0x4274, { 0x9a, 0xfc, 0xaa, 0x8b, 0x1e, 0xef, 0x52, 0x93 } };
```
makubacki:
```suggestion
```

crates/uefi_performance/src/_smm.rs
makubacki, line 1: For others, as I've mentioned previously, once the MM services code is merged, I'll update performance to use it and consolidate/remove this SMM code. It is fine for it to be checked in this PR so that can happen as a follow up.

crates/uefi_performance/src/log_perf_measurement.rs
```diff
@@ -0,0 +1,447 @@
+//! Module that define every performance functions used to log performance records.
```
makubacki:
```suggestion
//! Module that defines all performance functions used to log performance records.
```

crates/uefi_performance/src/log_perf_measurement.rs
```diff
@@ -0,0 +1,447 @@
...
+    } else {
+        ptr::null()
+    };
+    // Safety: string parameter is expected to be a valid C string.
+    unsafe {
+        (create_performance_measurement)(handle, None, string, timestamp, 0, identifier, PerfAttribute::PerfEndEntry)
+    };
+}
+
+pub fn perf_image_start_begin(module_handle: efi::Handle, create_performance_measurement: CreateMeasurement) {
```
makubacki: Can you please add doc comments for the public functions in this module?

crates/uefi_performance/src/pei.rs
makubacki, line 1: Consider dropping "PEI" from the way to refer to this code. PEI is one type of pre-DXE environment but technically HOBs can be constructed (and are) in other non-PEI environments to load DXE. Maybe "hob_records.rs" or something like that would be more generic and still apply to what's happening.

makubacki:
+@Javagedes, in particular, to review component usage in [`crates/uefi_performance/src/lib.rs`](https://github.com/OpenDevicePartnership/uefi-dxe-core/pull/351/files#diff-fb548867959fc724a8b91b8fea534cbcaa45e10f016732c89b51efa8e4c65786).

crates/uefi_performance/src/lib.rs
```diff
@@ -2,460 +2,311 @@
...
-    // Install configuration table for performance property.
-    unsafe {
-        BOOT_SERVICES.install_configuration_table(
-            &guid::PERFORMANCE_PROTOCOL,
-            Box::new(PerformanceProperty::new(Arch::perf_frequency(), Arch::cpu_count_start(), Arch::cpu_count_end())),
-        )?
-    };
-    Ok(())
-}
+static LOAD_IMAGE_COUNT: AtomicU32 = AtomicU32::new(0);
```
makubacki: Do these really need to be global statics? This seems a little messy from a code organization perspective.

crates/uefi_performance/src/lib.rs
```diff
@@ -2,460 +2,311 @@
...
-                | PerfId::MODULE_DB_END
-                | PerfId::MODULE_DB_SUPPORT_START
-                | PerfId::MODULE_DB_SUPPORT_END
-                | PerfId::MODULE_DB_STOP_START
-                | PerfId::MODULE_DB_STOP_END
-        )
-    }
+#[derive(Debug, Default)]
+// Config used to endable or disable the performance measuremnt. (Disabled by default)
+pub struct PerformanceMeasurementEnabled(pub bool);
```
makubacki: Is this the only config option you're expecting? I was thinking it might be helpful to have a config where this is an option but it's not entirely devoted to enable/disable of the feature. How is the used versus simply not including the performance component when it should be disabled? Normally we treat perf build as a build time decision.

crates/uefi_performance/src/lib.rs
```diff
@@ -2,460 +2,311 @@
...
-            Err(efi::Status::INVALID_PARAMETER)
-        }
-    }
+/// Performance Lib Component.
+#[derive(IntoComponent)]
 
-    if !PERF_ENABLED {
-        return efi::Status::SUCCESS;
-    }
+pub struct PerformanceLib;
```
makubacki: Is this called "PeformanceLib" because of where the code resides in the C code or is there a Rust reason for the name? I'd consider dropping the "Lib" part as that seems a little confusing to me in relation to a "component". For example, I think this would be considered the "Performance component" not the "PerformanceLib component".

.cargo/config.toml
```diff
@@ -9,4 +9,4 @@ replace-with = "UefiRust"
 
 # This tells cargo to consider the MSV of rust for our crate vs our depencencies
 [resolver]
-incompatible-rust-versions = "fallback"
+incompatible-rust-versions = "fallback"
```
Javagedes: I'm confused as to what changed in this file. It needs a newline though.

crates/uefi_performance/src/log_perf_measurement.rs
```diff
@@ -0,0 +1,447 @@
...
+        module_handle,
+        None,
+        None,
+        0,
+        KnownPerfId::ModuleStart.as_u16(),
+        create_performance_measurement,
+    );
+}
+
+pub fn perf_image_start_end(module_handle: efi::Handle, create_performance_measurement: CreateMeasurement) {
```
Javagedes: If your only reason for exposing `create_performance_measurement: CreateMeasurement` was to enable testing, then I disagree with the decision. We should not sacrifice usability and this function interface like this simply to be able to swap out `create_performance_measurement` for `test_create_performance_measurement` for tests.

My suggestion is to remove the ability to swap out `CreateMeasurement` function. Each of these functions is a single line wrapper around log_perf_measurement. The important thing to test is `log_performance_measurement` which exposes the function. So just test that method as much as possible, using your mock callback, and mark all of these one-line wrappers with `#[cfg(not(tarpaulin_include))]`

If I missed something and it's used for more then enabling testing, we can re-evaluate. But there is a better way to tackle this problem then to expose internals to the caller. Especially when the caller will also select the same thing.

crates/uefi_performance/src/lib.rs
makubacki: Mathieu and I had an offline discussion. The higher-level request is to add a matrix that explains how the Rust perf code integrates with the C code that has not/will not be ported to Rust. Which Rust components are needed? Which Rust config is needed? What C modules/libraries are needed? What C PCD settings are still relevant and why? Then, we can look at how all of that makes sense for the different perf scenarios supported. Any perf scenario support delta from the C code should be called out.

crates/uefi_performance/src/log_perf_measurement.rs
```diff
@@ -0,0 +1,447 @@
+//! Module that defines all performance functions used to log performance records.
```
Javagedes, line 1: We originally had all of these as macros (to mirror `log`) that we could easily conditionally compile them out when we were not doing perf measurements. I know there are some records that must always be available that are measured within the core - But how are you supporting this capability now? When components start caring about perf and want to measure their performance, it needs to be conditionally compile-able. We need a path to do that and it looks like you removed that path.

crates/uefi_performance/src/lib.rs
Javagedes: I've advocated from the start that this needs to basically be similar to `log` and conditionally compileable. Especially when components start caring about performance. I even pushed Sherry to make the measurements macros so that we can compile them, but that functionality has been removed in this PR. I strongly advocate not to remove that functionality and for it to be conditionally compile-able.

I know there are a few core measurements that must always exist, but the vast majority of measurements should be a compile time decision to include orn not. Just like with `log`.

crates/uefi_performance/src/lib.rs
```diff
@@ -493,17 +342,204 @@ extern "efiapi" fn fetch_and_add_smm_performance_records(_event: efi::Event, sys
     }
 
     // Write found perf records in the fbpt table.
+    let mut fbpt = fbpt.lock();
     let mut n = 0;
     for r in Iter::new(&smm_boot_records_data) {
-        FBPT.lock().add_record(r).unwrap();
+        // todo do not crash plz.
```
Javagedes: nit comment remove or edit

crates/uefi_performance/src/lib.rs
```diff
@@ -493,17 +342,204 @@ extern "efiapi" fn fetch_and_add_smm_performance_records(_event: efi::Event, sys
...
+            let record = GuidQwordEventRecord::new(perf_id, 0, timestamp, guid, address as u64);
+            fbpt.lock().add_record(record)?;
+        }
+        KnownPerfId::ModuleDbStopEnd => {
+            let module_handle = caller_identifier as efi::Handle;
+            let Ok(guid) = get_module_guid_from_handle(boot_services, module_handle) else {
+                log::error!("Performance Lib: Could not find the guid for module handle: {:?}", module_handle);
+                return Err(efi::Status::INVALID_PARAMETER);
+            };
+            // TODO: use of commponent 2 protocol, need usecase to test further. https://github.com/OpenDevicePartnership/uefi-dxe-core/issues/192
```
Javagedes, line 481: nit TODO

crates/uefi_performance/src/lib.rs
```diff
@@ -2,460 +2,311 @@
...
+            .unwrap_or_default();
+
+        // Initialize perf data form PEI values.
+        LOAD_IMAGE_COUNT.store(pei_load_image_count, Ordering::Relaxed);
+        fbpt.lock().set_perf_records(pei_perf_records);
+
+        // Install the protocol interfaces for DXE performance library instance.
+        boot_services.as_ref().install_protocol_interface(
+            None,
+            Box::new(EdkiiPerformanceMeasurement { create_performance_measurement }),
```
Javagedes, line 169: This takes a `*const c_void` (self). The paradigm we've been following for these protocols is to store data behind the protocol, and use that data rather than having global statics.

``` rust
struct Protocol {
    prot: EdkiiPerformanceMeasurement
    // State to use
    service: Service<dyn PerfService> // Or something else
}

// unsafe impl of Protocol on this new struct
```

crates/uefi_performance/src/performance_record/extended.rs
```diff
@@ -7,13 +9,12 @@
...
 use core::fmt::Debug;
 
-use mu_rust_helpers::guid::guid_fmt;
 use r_efi::efi;
 use scroll::Pwrite;
 
 use super::PerformanceRecord;
 
-#[repr(C)]
+#[derive(Debug)]
```
Javagedes, line 17: ou lost the "C" repr here. Seems like you would still want it?

crates/uefi_performance/src/performance_record/extended.rs
```diff
@@ -46,35 +47,17 @@ impl PerformanceRecord for GuidEventRecord {
...
+        buff.gwrite_with(self.acpi_id, offset, scroll::NATIVE)?;
+        buff.gwrite_with(self.timestamp, offset, scroll::NATIVE)?;
+        buff.gwrite_with(self.guid.as_bytes().as_slice(), offset, ())?;
+        Ok(())
     }
 }
 
-#[repr(C)]
+#[derive(Debug)]
 pub struct DynamicStringEventRecord<'a> {
```
Javagedes, line 61: Lost repr(C). Seems like we should keep it if this is representing a C struct.

crates/uefi_performance/src/performance_record/extended.rs
```diff
@@ -125,23 +108,19 @@ impl PerformanceRecord for DynamicStringEventRecord<'_> {
...
+        buff.gwrite_with(self.guid.as_bytes().as_slice(), offset, ())?;
+        buff.gwrite_with(self.string.as_bytes(), offset, ())?;
+        buff.gwrite_with(0_u8, offset, scroll::NATIVE)?; // End of the string.
+        Ok(())
     }
 }
 
-#[repr(C)]
+#[derive(Debug)]
 pub struct DualGuidStringEventRecord<'a> {
```
Javagedes, line 124: And here

crates/uefi_performance/src/log_perf_measurement.rs
magravel: Originally, it was because inside the core I use create_performance_measurement directly and outside the core, the use will take the function from the protocol. This change was not made for testing.

crates/uefi_performance/src/performance_record/extended.rs
magravel: Repr C is not useful here because the serialization is made by pwrite. Having the struct repr c does not change anything here.

crates/uefi_performance/src/log_perf_measurement.rs
magravel: Perf component will only be there when perf is enabled, and keep the same behavior as the C code.

crates/uefi_performance/src/lib.rs
magravel: I looked into it and I can't do a void pointer like advance logger or driver binding to hide the state since the protocol does not take self as first argument.

crates/uefi_performance/src/lib.rs
makubacki: @magravel, please include this in the matrix. If you can have that ready for tomorrow's dev sync, let's discuss the scenarios there.

PR #357, opened 2025-04-16
Implement AdvLogger parsing and parser executable

cfernald:
## Description

Implements a full parser for adv logger when in the standard environment and creates an executable that can be used to parse data buffer. This is an initial change to support experimentation with parsing. Further improvements will likely be needed to improve this use case.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Test executable locally.

## Integration Instructions

N/A

### DIFF
--- a/adv_logger/bin/bin.rs
+++ b/adv_logger/bin/bin.rs
@@ -16,11 +16,15 @@ use std::{
 
 #[derive(Parser, Debug)]
 struct Args {
+    /// Path for the input file containing the raw advanced logger buffer.
     input_path: PathBuf,
+    /// Optional path for the output file. If not specified, the output will be printed to stdout.
     #[arg(short, long)]
     output_path: Option<PathBuf>,
+    /// Flag to include entry metadata (level, phase, timestamp) in the output.
     #[arg(short, long, default_value_t = false)]
     entry_metadata: bool,
+    /// Flag to include the header in the output.
     #[arg(long, default_value_t = false)]
     header: bool,
 }


adv_logger/bin/bin.rs
```diff
@@ -0,0 +1,53 @@
...
+use std::env;
+use std::fs::File;
+use std::io::{self, Read};
+use std::path::Path;
+
+use adv_logger::parser::Parser;
+
+fn main() -> io::Result<()> {
+    // Collect the command line arguments
+    let args: Vec<String> = env::args().collect();
```
Javagedes: Do you suspect you will ever want to expand on this to have arguments? If so, I'd just start off with [clap](https://crates.io/crates/clap)

```rust
use clap::Parser;
use std::path::PathBuf;

#[derive(Parser, Debug)]
struct Args {
    input_path: PathBuf,
}

fn main() -> io::Result<()> {
    let args = Args::parse();
    
    let mut File::open(args.input_path)?;
}
```

adv_logger/bin/bin.rs
```diff
@@ -0,0 +1,65 @@
...
+
+use clap::Parser;
+use std::{
+    fs::File,
+    io::{self, Read},
+    path::{Path, PathBuf},
+};
+
+#[derive(Parser, Debug)]
+struct Args {
```
Javagedes, line 18: nit: Add some `///` comments to the `Args` struct and the individual fields. It comes out as documentation when you do `<cmd> --help`

PR #358, opened 2025-04-16
image.rs: Make .rdata sections read only

makubacki:
## Description

If an initialized data section is readable and not writable set the RO attribute on that section.

This is meant to set .rdata sections as RO:

rdata section

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- QEMU boot to EFI shell and Windows

## Integration Instructions

- N/A

### DIFF
--- a/dxe_core/src/image.rs
+++ b/dxe_core/src/image.rs
@@ -41,8 +41,6 @@ pub const EFI_IMAGE_SUBSYSTEM_EFI_RUNTIME_DRIVER: u16 = 12;
 
 pub const ENTRY_POINT_STACK_SIZE: usize = 0x100000;
 
-const INIT_READ_DATA: u32 = section_table::IMAGE_SCN_CNT_INITIALIZED_DATA | section_table::IMAGE_SCN_MEM_READ;
-
 // dummy function used to initialize PrivateImageData.entry_point.
 #[cfg(not(tarpaulin_include))]
 extern "efiapi" fn unimplemented_entry_point(
@@ -367,7 +365,7 @@ fn apply_image_memory_protections(pe_info: &UefiPeInfo, private_info: &PrivateIm
         }
 
         if section.characteristics & section_table::IMAGE_SCN_MEM_WRITE == 0
-            && ((section.characteristics & INIT_READ_DATA) == INIT_READ_DATA)
+            && ((section.characteristics & section_table::IMAGE_SCN_MEM_READ) == section_table::IMAGE_SCN_MEM_READ)
         {
             attributes |= efi::MEMORY_RO;
         }


makubacki:
Context: We discussed .rdata sections recently for SEA validation and it seems like this could be used here. Afaict, this wasn't being set already.

dxe_core/src/image.rs
```diff
@@ -363,6 +364,14 @@ fn apply_image_memory_protections(pe_info: &UefiPeInfo, private_info: &PrivateIm
             attributes = efi::MEMORY_RO;
         }
 
+        if section.characteristics & (pecoff::IMAGE_SCN_CNT_CODE | section_table::IMAGE_SCN_MEM_WRITE) == 0
+            && (section.characteristics & section_table::IMAGE_SCN_CNT_INITIALIZED_DATA
```
cfernald: nit, consider some constants to make this more readable. e.g `const INIT_READ_DATA = (section_table::IMAGE_SCN_CNT_INITIALIZED_DATA | section_table::IMAGE_SCN_MEM_READ)`

dxe_core/src/image.rs
```diff
@@ -363,6 +364,14 @@ fn apply_image_memory_protections(pe_info: &UefiPeInfo, private_info: &PrivateIm
             attributes = efi::MEMORY_RO;
         }
 
+        if section.characteristics & (pecoff::IMAGE_SCN_CNT_CODE | section_table::IMAGE_SCN_MEM_WRITE) == 0
```
os-d: Should we be even broader and say if `IMAGE_SCN_MEM_WRITE` is not present, we set RO? Per PE/COFF spec, that would be appropriate, but I'm not sure if all of our images truly follow that.

makubacki: I was thinking the same thing but was also vaguely concerned about compatibility. As the memory protections owner, I'll let you make the decision. I can test on QEMU again if we drop `IMAGE_SCN_CNT_CODE`.

os-d: I think the right thing to do is to set RO for a section that explicitly tells us it is not writeable. If you want to do a smoke test on QEMU to confirm at least the core drivers follow this expectation, I think it is reasonable to check in, fairly easy to change in the future if we discover a pattern that breaks this and can't be resolved.

dxe_core/src/image.rs
makubacki: Added

dxe_core/src/image.rs
makubacki: Updated

dxe_core/src/image.rs
```diff
@@ -363,6 +366,14 @@ fn apply_image_memory_protections(pe_info: &UefiPeInfo, private_info: &PrivateIm
             attributes = efi::MEMORY_RO;
         }
 
+        if section.characteristics & section_table::IMAGE_SCN_MEM_WRITE == 0
+            && ((section.characteristics & section_table::IMAGE_SCN_CNT_INITIALIZED_DATA
+                | section_table::IMAGE_SCN_MEM_READ)
```
os-d: nit: could use INIT_READ_DATA here, too.

dxe_core/src/image.rs
```diff
@@ -363,6 +366,14 @@ fn apply_image_memory_protections(pe_info: &UefiPeInfo, private_info: &PrivateIm
             attributes = efi::MEMORY_RO;
         }
 
+        if section.characteristics & section_table::IMAGE_SCN_MEM_WRITE == 0
```
os-d, line 367: I was imagining this as a separate if statement (or same if statement, but not checking for `INIT_READ_DATA`. I.e. if any sections does not set the `IMAGE_SCN_MEM_WRITE` characteristic, we do `attributes |= efi::MEMORY_RO`. So we end up with a ternary state:
- Section Code + Write (should be all code): RO
- Section Write: XP
- Section non-Write: RO + XP

What do you think? I believe this covers your case + all other cases. There are a few random sections that we probably won't see that don't set READ or INITIALIZED_DATA and don't set WRITE, but I guess the thought is that if we make the case more general it would be hard for anything to slip through.

makubacki: That sounds fine. I'll try Q35 and SBSA with it.

makubacki: @os-d, what about `IMAGE_SCN_MEM_WRITE` not set AND `IMAGE_SCN_MEM_READ` is set then `attributes |= efi::MEMORY_RO`? Looking at some images, they all have the readable bit in section characteristics if they're actually readable.

image

os-d: Yeah, I was thinking about that, too. That makes sense to me. It is kind of odd, if you had a non-readable section, what are you supposed to do with it? I don't think you could actually unmap it, I think that is just supposed to indicate that the CRT would use it.

makubacki: I agree it would be strange. But I feel setting a read only bit to be dependent on that also make sense. Will update with that.

Issue #355, opened 2025-04-16
[Bug]: Advanced Logger Messages Garbled in Release Mode on Physical Platform


kouchekiniad:
### Is there an existing issue for this?

- [x] I have searched existing issues

### Current Behavior

In the Advanced Logger memory log headers are written without issue, but the messages following the header are garbled. This behavior is only seen in Release mode.

### Expected Behavior

Both Advanced Log headers and messages should be written without issue.

### Steps To Reproduce

1. Run Rust DXE Core on physical platform with the Advanced Logger enabled
2. View the Advanced Logger log. Messages once the Rust DXE Core has started will appear garbled.

### Build Environment

```markdown
- OS(s): Windows
- Tool Chain(s): VS2022
- Targets Impacted: RELEASE
```

### Version Information

```text
v9.2.0
```

### Urgency

Medium

### Are you going to fix this?

I will fix it

### Do you need maintainer feedback?

No maintainer feedback needed

### Anything else?

I have a working fix for this issue, which is to write the message byte-by-byte with `write_volatile`. I'm digging around to see if there's a more efficient way to do this, and to confirm this is an issue with the compiled Rust binary, rather than with CPU caching.

One optimization would be to use `write_volatile` on aligned QWORDs of memory, but I'm not sure if that'd save any significant performance.

*This bug is also reproducible on the debug profile with `debug-assertions = false` set.*

kouchekiniad:
Fix that is potentially slow is on personal/kouchekiniad/adv-logger-volatile-writes

os-d:
@cfernald for reference

kouchekiniad:
Fixed in #361

PR #361, opened 2025-04-17
Use Volatile Copies for Advanced Logger Messages

kouchekiniad:
## Description
With this change, messages written to the Advanced Logger buffer are now written with volatile memory copies, preventing the compiler from optimizing out these memory movements in release builds. Message headers (but not the messages themselves) were already being copied over with volatile copies before this change. This change is a fix for issue #355.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested
Booted to Windows with Advanced Logs functional on release UEFI.

## Integration Instructions
N/A

### DIFF
--- a/adv_logger/src/memory_log.rs
+++ b/adv_logger/src/memory_log.rs
@@ -387,14 +387,14 @@ mod tests {
             let entry = LogEntry { level: 0, phase: 0, timestamp: 0, data: &data };
             let log_entry = log.unwrap().add_log_entry(entry);
             match log_entry {
-                Ok => {}
+                Ok(_) => {}
                 Err(EfiError::OutOfResources) => {
                     assert!(log.unwrap().discarded_size > 0);
                     assert!(entries > 0);
                     break;
                 }
                 Err(status) => {
-                    assert!(false, "Unexpected add_log_entry returned unexpected status {:#x?}.", status)
+                    panic!("Unexpected add_log_entry returned unexpected status {:#x?}.", status)
                 }
             }
             entries += 1;


adv_logger/src/memory_log.rs
```diff
@@ -255,27 +255,62 @@ impl AdvLoggerMessageEntry {
...
             size_of::<Self>() + log_entry.data.len() <= length as usize,
             "Advanced logger entry initialized in an insufficiently sized buffer!"
         );
-
         if size_of::<Self>() + log_entry.data.len() > length as usize {
             return None;
         }
 
+        // Ensure the address and length are aligned.
+        assert!((address as *const u64).is_aligned(), "address must be aligned to 8 bytes");
```
os-d: I don't think we should use `assert` here, but instead `debug_assert` and fail gracefully. In a production platform, we wouldn't want to fail boot just because we couldn't log.

adv_logger/src/memory_log.rs
```diff
@@ -255,27 +255,62 @@ impl AdvLoggerMessageEntry {
...
-            adv_entry,
+        ptr::write_volatile::<Self>(
+            address as *mut Self,
             Self::new(log_entry.phase, log_entry.level, log_entry.timestamp, log_entry.data.len() as u16),
         );
 
-        // write the data.
-        let message = adv_entry.offset(1) as *mut u8;
-        ptr::copy(log_entry.data.as_ptr(), message, log_entry.data.len());
+        const _: () = assert!(
```
os-d, line 283: Same comment here.

adv_logger/src/memory_log.rs
kouchekiniad: Good catch, I'd meant to switch that back before check-in

adv_logger/src/memory_log.rs
kouchekiniad: This is a compile time assert, so there'd be no penalty for running this on release builds

adv_logger/src/memory_log.rs
```diff
@@ -255,27 +255,62 @@ impl AdvLoggerMessageEntry {
...
-
         if size_of::<Self>() + log_entry.data.len() > length as usize {
             return None;
         }
 
+        // Ensure the address and length are aligned.
+        debug_assert!((address as *const u64).is_aligned(), "address must be aligned to 8 bytes");
+        if !(address as *const u64).is_aligned() {
+            return None;
+        }
```
makubacki: This seems cleaner to avoid casting.

```suggestion
        // Ensure the address and length are aligned.
        if address.align_offset(8) != 0 {
            debug_assert!(false, "address must be aligned to 8 bytes");
            return None;
        }
```

adv_logger/src/memory_log.rs
```diff
@@ -255,27 +255,62 @@ impl AdvLoggerMessageEntry {
     /// range is valid.
     ///
     pub unsafe fn init_from_memory(address: *const c_void, length: u32, log_entry: LogEntry) -> Option<&'static Self> {
```
makubacki: This function has enough error handling that I think it should return a `Result`.

adv_logger/src/memory_log.rs
```diff
@@ -255,27 +255,62 @@ impl AdvLoggerMessageEntry {
...
+        // Ensure the entry fits.
         debug_assert!(
```
makubacki: Is there an advantage to executing the same condition twice? From a human readability perspective, it seems tidier to check the condition once in the if statement and then perform any further action based on that condition (including a `debug_assert!()`) inside the if block.

kouchekiniad: Make sense, I'll make the change

adv_logger/src/memory_log.rs
kouchekiniad: Addressed

adv_logger/src/memory_log.rs
kouchekiniad: Addressed

adv_logger/src/memory_log.rs
kouchekiniad: Addressed in my latest changeset

adv_logger/src/memory_log.rs
Javagedes: This is cool. I did not know this about consts, but a panic in a const becomes a compile-time error. You learn something new every day.

PR #363, opened 2025-04-18
dxe_core: Print dxe_core version

makubacki:
## Description

To make it easier to trace and identify the dxe_core library version being used in a debug log, print it out at info level at the beginning of init_memory().

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- Verified correct `dxe_core` library version is printed in serial log

  ```
  INFO - Advanced logger buffer initialized. Address = 0x000000007ebee000
  INFO - DXE Core Library v10.0.0
  INFO - Loaded GDT @ 0x7eb0ad60
  ```

## Integration Instructions

- N/A

### DIFF
--- a/dxe_core/src/lib.rs
+++ b/dxe_core/src/lib.rs
@@ -307,6 +307,8 @@ where
         mut self,
         physical_hob_list: *const c_void,
     ) -> Core<CpuInit, SectionExtractor, InterruptManager, InterruptBases, Alloc> {
+        log::info!("DXE Core Crate v{}", env!("CARGO_PKG_VERSION"));
+
         let _ = self.cpu_init.initialize();
         self.interrupt_manager.initialize().expect("Failed to initialize interrupt manager!");
 


dxe_core/src/lib.rs
```diff
@@ -307,6 +307,8 @@ where
         mut self,
         physical_hob_list: *const c_void,
     ) -> Core<CpuInit, SectionExtractor, InterruptManager, InterruptBases, Alloc> {
+        log::info!("DXE Core Library v{}", env!("CARGO_PKG_VERSION"));
```
os-d: This is maybe a little confusing to call it a Library, when looked at from the platform side. Maybe omitting that or calling it something more generic (not used on the platform side) like Package or Binary

makubacki: We actually have a different version for the binary that I was going to "Binary" for. I'm open to something else here though, but it is the common library crate linked to those binaries.

os-d: Sure, I just thought from a non-Rust UEFI developer point of view, they may not understand that the DXE Core Library is the DXE Core, but I'm fine with whatever you choose.

makubacki: This is showing both prints on Q35:

> INFO - DXE Core Rust Binary v0.3.1
> INFO - DXE Core Rust Library v10.0.0

What about that? With "Rust" mentioned in both.

Javagedes: maybe 

DXE Core Platform Binary v0.3.1
DXE Core Crate v10.0.0

makubacki: I'm fine with that. Will update.

os-d: Seems reasonable to me.

PR #364, opened 2025-04-18
[RFC]: ACPI Services

berlin-with0ut-return:
## Description
RFC for a new `AcpiProvider` service, including all necessary components, services, and structs to support this service.

Current RFC state: **FCP**


docs/src/rfc/text/0000-acpi.md
```diff
@@ -0,0 +1,162 @@
...
+- The ACPI SDT (System Description Table) Protocol provides access to and enables traversal of the full ACPI table hierarchy.
+  - [EFI_ACPI_SDT_PROTOCOL](https://uefi.org/specs/PI/1.8A/V5_ACPI_System_Desc_Table_Protocol.html)
+
+## Goals
+
+Create a idiomatic Rust API for all ACPI functionality, including both ACPI-related protocols.
+
+## Requirements
+
+1. The API should provide all necessary ACPI functionality as a service to both components and within the score
```
os-d:
```suggestion
1. The API should provide all necessary ACPI functionality as a service to components

```

We don't anticipate any ACPI usage from the core, right?

docs/src/rfc/text/0000-acpi.md
```diff
@@ -0,0 +1,162 @@
...
+while preserving compatability with the existing C implementation?
+- Is there value in exposing the ACPI table protocol and ACPI SDT protocol as seperate services?
+
+## Prior Art (Existing PI C Implementation)
+
+The Rust-based ACPI implementation follows the ACPI Table protocol and ACPI SDT protocol
+as described in the UEFI specification. *See [Protocols](#protocols) for more information.*
+
+This RFC chooses to omit most of the ACPI SDT protocol,
+implementing only `GetTable` based on investigation into its practical usage in real platform code.
```
os-d: The protocol will still meet the PI spec defined interfaces (even if they only return unsupported), right? Otherwise drivers discovering the protocol could attempt to execute random memory if they think the structure is the PI spec defined interface.

docs/src/rfc/text/0000-acpi.md
```diff
@@ -0,0 +1,162 @@
...
+    current_handle: usize,
+    acpi_table_protocol: AcpiTableProtocol,
+    acpi_sdt_protocol: AcpiSdtProtocol
+}
+
+```
+
+### ACPI Component
+
+A single `AcpiComponent` initializes the root ACPI tables, registers the ACPI provider service,
```
os-d: nit: I wonder if we have component naming conventions? @Javagedes @makubacki ? We may otherwise end up with a lot of `*Component`s. Which, maybe is what we want, but just wanted to check if we've thought about it. Component in general is a fairly generic term, which sometimes can make it hard to tell what you are talking about (whether a Component or just a component of the system...).

docs/src/rfc/text/0000-acpi.md
```diff
@@ -0,0 +1,162 @@
...
+### ACPI Provider Service
+
+Integrated functionality for installing, uninstalling, and getting a specific ACPI table
+will be provided through the `AcpiProvider` service.
+
+```rust
+pub trait AcpiProvider {
+    fn install_acpi_table(&self, acpi_table: &AcpiTable) -> Result<TableKey, AcpiTableError>;
+    fn uninstall_acpi_table(&self, table_key: TableKey) -> Result<(), AcpiTableError>;
+    fn get_acpi_table(&self, index: usize) -> Result<&AcpiTable, AcpiTableError>;
```
os-d: I agree this matches the PI spec defined interface, but I'm wondering is this a place where we want to update this interface? (keeping the protocol interface the same). It seems a bit lame to continue the pattern of just call me until you get the table you want by chance, is there a better way we can do this where the table desired is passed in? I guess this could be two interfaces: GiveMeThisTable vs GiveMeTableIter, the latter returning an iterator of tables? Just spitballing, but this feels like a C interface that could be updated for the Rust world. I don't have a strong opinion here, just asking whether we feel this interface should be updated.

berlin-with0ut-return: i was also thinking about this when looking at the `get` function. the common use case i see now for C is to iterate over the indices and look for a particular signature (for both custom and system tables). however, this is partially because the C ACPI tables are implemented as a linked list, which makes direct indexing hard. i do not see any reason we have to use a linked list (?) and if we don't, we could possibly `get` tables by signature or table key. (table key is tricker for system tables since it comes from `install_table`). 

either way, the division between `get` and `get_iter` makes sense to me as well.

docs/src/rfc/text/0000-acpi.md
```diff
@@ -0,0 +1,162 @@
+# RFC: `ACPI`
+
+This RFC proposes a Rust-based interface for ACPI functionality in UEFI, replacing the existing C implementations.
```
makubacki: My understanding is that this is essentially providing only the functionality in `EFI_ACPI_TABLE_PROTOCOL` and `EFI_ACPI_SDT_PROTOCOL`. If that's the case, it would be easier to understand the scope of the RFC by explicitly stating that in the intro.

```suggestion
This RFC proposes a Rust-based interface for ACPI table management providing the functionality described in the `EFI_ACPI_TABLE_PROTOCOL` and `EFI_ACPI_SDT_PROTOCOL` services defined respectively in the UEFI Specification and PI Specification.
```

makubacki: The reason being that there is other ACPI manipulation code in UEFI firmware, but this is simple to understand and reference if scoped to these protocols in the spec.

docs/src/rfc/text/0000-acpi.md
```diff
@@ -0,0 +1,162 @@
...
+
+- 2024-04-17: Initial RFC created.
+
+## Motivation
+
+The Advanced Configuration and Power Interface (ACPI)
+provides information about device topology, configuration, and power management.
+This RFC proposes a pure Rust interface for existing ACPI capabilities, replacing the C-based implementations
+while producing protocols that align with the UEFI specification. This provides a simpler, safer Rust-based interface
+while maintaining required ACPI functionality.
```
makubacki: Similar to the comment above, "This RFC proposes a pure Rust interface for existing ACPI capabilities, replacing the C-based implementations while producing protocols that align with the UEFI specification." is a little open ended. Please scope "existing ACPI capabilities" to either a list here or a list elsewhere in the RFC.

docs/src/rfc/text/0000-acpi.md
```diff
@@ -0,0 +1,162 @@
...
+ACPI within UEFI provides a standardized interface for firmware
+to convey hardware configuration and power management information to the operating system.
+This information is organized into a set of structured tables.
+
+For more information on the format and arrangement of these tables,
+see the UEFI specification on the [ACPI Software Programming Model](https://uefi.org/htmlspecs/ACPI_Spec_6_4_html/05_ACPI_Software_Programming_Model/ACPI_Software_Programming_Model.html#acpi-software-programming-model).
+
+### Protocols
+
+The ACPI model exposes two primary protocols for interacting with ACPI data:
```
makubacki:
```suggestion
The UEFI Forum Specifications expose two primary protocols for interacting with ACPI data:
```

docs/src/rfc/text/0000-acpi.md
```diff
@@ -0,0 +1,162 @@
...
+
+- The ACPI Table Protocol installs and uninstalls individual ACPI tables.
+  - [EFI_ACPI_TABLE_PROTOCOL](https://uefi.org/specs/UEFI/2.10/20_Protocols_ACPI_Protocols.html#efi-acpi-table-protocol)
+- The ACPI SDT (System Description Table) Protocol provides access to and enables traversal of the full ACPI table hierarchy.
+  - [EFI_ACPI_SDT_PROTOCOL](https://uefi.org/specs/PI/1.8A/V5_ACPI_System_Desc_Table_Protocol.html)
+
+## Goals
+
+Create a idiomatic Rust API for all ACPI functionality, including both ACPI-related protocols.
```
makubacki: Bound what "all ACPI functionality" means here.

```suggestion
Create a idiomatic Rust API for all ACPI functionality, including both ACPI-related protocols.
```

docs/src/rfc/text/0000-acpi.md
```diff
@@ -0,0 +1,162 @@
...
+## Requirements
+
+1. The API should provide all necessary ACPI functionality as a service to both components and within the score
+2. The API should utilize Rust best practices, particularly memory safety and error handling
+3. The ACPI should produce protocols equivalent to the current C implementations, preserving existing C functionality
+
+## Unresolved Questions
+
+- Which parts of the SDT protocol are necessary to re-implement in Rust?
+  - In platform code, `GetTable` is the only one currently used,
```
makubacki:
```suggestion
  - In platform code, `GetAcpiTable` is the only one currently used,
```

docs/src/rfc/text/0000-acpi.md
makubacki:
```suggestion
implementing only `GetAcpiTable` based on investigation into its practical usage in real platform code.
```

docs/src/rfc/text/0000-acpi.md
Javagedes: I'm open to standardizing, but I do not have an opinion. If we don't want to call it components, then we may just want to rename the entire `Component` trait to be something else, which I'm also open to. I only named it component to match EDKII. We should do it now before we open source, if we are going to do it.

docs/src/rfc/text/0000-acpi.md
makubacki: The ACPI service should likely provide some solution for `RegisterNotify` (currently in `EFI_ACPI_SDT_PROTOCOL`). It doesn't need to be the same interface but a notification mechanism for table installation.

docs/src/rfc/text/0000-acpi.md
makubacki: Something that feels natural to me, is to only have the provider manage the instances of tables installed. Implement an iterator over the provider. Make the table type something that can be filtered cleanly when iterating across tables. Let each `AcpiTable` have an `install` or `uninstall` method OR have these be associated functions on the `AcpiTable` type (e.g. `AcpiTable::install(&table)`.

docs/src/rfc/text/0000-acpi.md
makubacki: More so than just the component name itself, I think we might want to make a convention about where actual components go to make them easier to identify based on the fully qualified type name. We've started that in uefi-sdk but I don't think it's been written down anywhere yet.

docs/src/rfc/text/0000-acpi.md
Javagedes: @makubacki What do you mean by `AcpiTable::install(&table)`? My strong preference is to not use statics, as it makes testing difficult, and to me it looks like you are asking for this associated function to point back to some global static acpi table manager to register and unregister without directly requesting the service. If this is what you are asking, I think my vote is not to do this, however I could be completely misinterpreting your request. Let me know :)

makubacki: I was solely talking about the generic interface for installing an ACPI table. I agree that you'd probably not want to do this in the service model. Instead using the service stored in the instance of the table but `table` here is not necessarily `AcpiTable` and could have its own abstraction that uses the service like a trait that `AcpiTable::install()` uses for the backend of table installation while it does ACPI header validation, etc. That backend could use services, protocols, etc. This is partially trying to not couple "ACPI support" with "our service model" where it doesn't necessarily need to be so more of the generic ACPI code is reusable. I don't have a strong opinion though and even the original proposal can accomplish that separation.

makubacki: Sherry and I spoke offline. In summary, my only real request is to consider keeping generic ACPI functionality separate from being directly in a component where possible. ACPI is generic and code to validate tables, patch code, etc. could have value for people not using our service infrastructure. I do not have preference for how the service is abstracted and the RFC as written with the `AcpiProvider` trait providing an install service for `AcpiTable` is fine with me.

I suggest adding some platform examples to the RFC showing how common tables like MCFG, HPET, MADT, FADT, etc. could be structured with config and component pieces. Just a couple high-level examples at block diagram level. That way, a platform maintainer that knows nothing about components, configs, and services can have a starting point for how to think about transitioning ACPI table production from C code.

docs/src/rfc/text/0000-acpi.md
```diff
@@ -0,0 +1,179 @@
...
+
+The division between ACPI table protocol and ACPI SDT protocol is represented by two traits
+`AcpiTableManager` and `AcpiSdtManager`; which also produce the necessary protocols.
+However, these traits are not exposed to consumers of the ACPI service for simplicity.
+
+### ACPI Tables
+
+All information relating to a specific ACPI table is stored in the `AcpiTableStruct`, defined as follows:
+
+```rust
```
cfernald: It seems we would need safe abstractions to convert this structure into specific tables. Something like..

```rust
let table = AcpiTable(...);
let dmar_table = table.try_into(DmarAcpiTable).unwrap();
```

Where `DmarAcpiTable` would need to implement some `AcpiTable` trait that defined it's `SIGNATURE` and probably need some sort of way to safely derive size and parse for tables that are not fixed length.

docs/src/rfc/text/0000-acpi.md
```diff
@@ -0,0 +1,179 @@
...
+    oem_table_id: [u8; 8],
+    oem_revision: u32,
+    creator_id: u32,
+    creator_revision: u32,
+    table_key: TableKey, // type alias for u32
+    versions: AcpiVersion, // bitflag
+}
+```
+
+A global `AcpiTableProtocolSupport` struct provides metadata, such as pointers to the root ACPI tables.
```
cfernald: Does this need to be global? Can it not be allocated & leaked and kept as a `&'static EfiAcpiTableInstance` in the component?

docs/src/rfc/text/0000-acpi.md
```diff
@@ -0,0 +1,179 @@
...
+The division between ACPI table protocol and ACPI SDT protocol is represented by two traits
+`AcpiTableManager` and `AcpiSdtManager`; which also produce the necessary protocols.
+However, these traits are not exposed to consumers of the ACPI service for simplicity.
+
+### ACPI Tables
+
+All information relating to a specific ACPI table is stored in the `AcpiTableStruct`, defined as follows:
+
+```rust
+struct AcpiTable {
```
cfernald: This needs `#[repr(C, packed)]` right?

docs/src/rfc/text/0000-acpi.md
cfernald: Same probably goes for the service names. I've been going with "*Manager" but it might be good to have some sort of convention to more ideally identify the purpose of different services.

berlin-with0ut-return: as per discussion in the Rust meeting, we resolved on naming things omitting the `Component`, `Config`, etc suffixes.

docs/src/rfc/text/0000-acpi.md
berlin-with0ut-return: i believe the answer is no because the `get` in the ACPI SDT protocol is "stateless" (doesn't take in a self-like parameter)

```
type AcpiTableGet = extern "efiapi" fn(usize, *mut *mut AcpiSdtHeader, *mut u32, *mut usize) -> efi::Status;
```

docs/src/rfc/text/0000-acpi.md
```diff
@@ -0,0 +1,341 @@
...
+#### EDKII Protocols
+
+In addition to holding information relating to ACPI tables, the global static `ACPI_TABLE_INFO`
+will also produce C protocols to support the existing EDKII `EFI_ACPI_TABLE_PROTOCOL` and `EFI_ACPI_SDT_PROTOCOL`.
+
+The produced protocols will essentially be wrappers around their respective Rust functions, as shown below:
+
+```rust
+#[repr(C)]
+#[protocol = "ffe06bdd-6107-46a6-7bb2-5a9c7ec5275c"]
```
Javagedes: This attribute macro was removed to separate services from protocols. You will instead implement this trait: [ProtocolInterface](https://github.com/OpenDevicePartnership/uefi-sdk/blob/864e36473a90a357cd5f6f9de53d1f3de482e22f/uefi_protocol/src/uefi_protocol.rs#L12) so that it can be registered with the StandardBootServices dependency injectable param :)

docs/src/rfc/text/0000-acpi.md
```diff
@@ -0,0 +1,341 @@
...
+        ACPI_TABLE_INFO.install_acpi_table(...);
+    }
+    
+    extern "efiapi" fn uninstall_acpi_table_ext(protocol: *const AcpiTableProtocol, table_key: usize) -> efi::Status {
+        ACPI_TABLE_INFO.uninstall_acpi_table(...);
+    }
+}
+
+#[repr(C)]
+#[protocol = "eb97088e-cfdf-49c6-be4b-d906a5b20e86"]
```
Javagedes: Same with this

berlin-with0ut-return:
moving to FCP

docs/src/rfc/text/0000-acpi.md
```diff
@@ -0,0 +1,419 @@
...
+```rust
+type AcpiNotifyFn = fn(&AcpiTable, AcpiVersion, TableKey) -> Result<(), AcpiTableError>;
+
+pub trait AcpiProvider {
+    type Iter: Iterator<Item = AcpiTable>;
+
+    fn install_acpi_table(&self, acpi_table: &AcpiTable) -> Result<TableKey, AcpiTableError>;
+    fn uninstall_acpi_table(&self, table_key: TableKey) -> Result<(), AcpiTableError>;
+    fn get_acpi_table(&self, index: usize) -> Result<&AcpiTable, AcpiTableError>;
+    fn register_notify(&self, should_register: bool, notify_fn: AcpiNotifyFn) -> Result<(), AcpiTableError>;
```
makubacki: What does `should_register` mean here? Some brief parameter documentation for these functions would be a nice-to-have.

docs/src/rfc/text/0000-acpi.md
```diff
@@ -0,0 +1,419 @@
...
+#### ACPI Support
+
+A `AcpiTableProtocolSupport` struct provides metadata, such as pointers to the root ACPI tables.
+This data is contained in the global `ACPI_TABLE_INFO`, which implements `AcpiProvider`.
+
+```rust
+#[repr(C, packed)]
+#[derive(Debug)]
+struct AcpiTableProtocolSupport {
+    signature: usize,
```
makubacki: Is `signature` meant to be a `usize`?

docs/src/rfc/text/0000-acpi.md
```diff
@@ -0,0 +1,419 @@
...
+
+```rust
+type AcpiNotifyFn = fn(&AcpiTable, AcpiVersion, TableKey) -> Result<(), AcpiTableError>;
+
+pub trait AcpiProvider {
+    type Iter: Iterator<Item = AcpiTable>;
+
+    fn install_acpi_table(&self, acpi_table: &AcpiTable) -> Result<TableKey, AcpiTableError>;
+    fn uninstall_acpi_table(&self, table_key: TableKey) -> Result<(), AcpiTableError>;
+    fn get_acpi_table(&self, index: usize) -> Result<&AcpiTable, AcpiTableError>;
```
makubacki: Can you please elaborate on how a caller would choose what `index` value to pass?

docs/src/rfc/text/0000-acpi.md
```diff
@@ -0,0 +1,419 @@
...
+When installing a table with `install_acpi_table`, a unique identifier `TableKey` will be returned.
+This unique handle can later be used in `uninstall_acpi_table`.
+
+Upon calling `install_acpi_table`, any functions registered with `register_notify` will be called.
+
+`get_acpi_table` finds a specific ACPI table by index,
+either in the firmware-provided tables or custom tables installed by `install_acpi_table`.
+Alternatively, `AcpiProvider` also allows access to all tables through iteration.
+
+### MCFG Example
```
makubacki: This example is exactly what I was looking for. Thanks.

docs/src/rfc/text/0000-acpi.md
```diff
@@ -0,0 +1,419 @@
...
+  let segment_count = /* some platform value */
+  let table_base = /* get from RSDP */
+
+  // ...
+
+  Core::default()
+        // ...
+        .with_component(acpi_init)
+        .with_config(AcpiMcfgInit { segment_count, oem_revision, table_base })
+        .with_component(install_mcfg)
```
makubacki: Maybe clarify here that `install_mcfg` is of type `AcpiSystemTableManager` to easily establish that connection.

docs/src/rfc/text/0000-acpi.md
berlin-with0ut-return: right now in C, you mostly use `get` to iterate over tables and have an `if` for the table that you're looking for (for example, a particular signature/table key/etc). in this way the Rust `get` function is mostly redundant with `iter` (the only thing it provides over `iter` is random access).

i considered making a `get` for a particular table key / signature, but again this functionality is already available with `iter` + `filter`. 

i think it's still useful to have in case you need random access to a previously discovered table, but i'm only 70% sure that we need it.

docs/src/rfc/text/0000-acpi.md
berlin-with0ut-return: it should be a u32, i've fixed this in the latest iteration

docs/src/rfc/text/0000-acpi.md
makubacki: I think it's fine for now, but please clarify how it is used in this RFC API to make it easy for viewers to understand it.

PR #366, opened 2025-04-21
docs: Update Monolithic Component Documentation

Javagedes:
## Description

Update the documentation regarding Monolithic components.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

N/A

## Integration Instructions

N/A

### DIFF
--- a/docs/src/driver/interface.md
+++ b/docs/src/driver/interface.md
@@ -34,8 +34,8 @@ for a complete list, however the below information should be up to date.
 ## Component Execution
 
 Components are executed by validating each individual parameter (See [Params](#component-params) below) in the
-component. If all paramters are validated, then the component is executed and removed from the list of components to
-execute. If any paramter fails to validate, then that parameter is registered as the failed param and the dispatcher
+component. If all parameters are validated, then the component is executed and removed from the list of components to
+execute. If any parameters fails to validate, then that parameter is registered as the failed param and the dispatcher
 will attempt to validate and execute the component in the next iteration. The dispatcher stops executing when no
 components have been dispatched in a single iteration.
 


docs/src/driver/interface.md
```diff
@@ -50,7 +46,104 @@ does matter!
...
+| Option\<P\>                  | An Option, where P implements `Param`. Allows components to run even when the underlying parameter is unavailable.                |
+| (P1, P2, ...)                | A Tuple where each entry implements `Param`. Useful when you need more parameters than the current parameter limit.               |
+| Config\<T\>                  | An immutable config value that will only be available once the underlying data has been locked.                                   |
+| ConfigMut\<T\>               | A mutable config value that will only be available while the underlying data is unlocked.                                         |
+| Service\<T\>                 | A wrapper for producing and consuming services of a particular interface, `T`, that is agnostic to the underlying implementation. |
+<!-- markdownlint-enable -->
+
+### Option\<P\>
+
+Some parameter, discussed below, are not *always* available. When a parameter is not available, the component will
```
os-d:
```suggestion
Some parameters, discussed below, are not *always* available. When a parameter is not available, the component will
```

docs/src/driver/interface.md
```diff
@@ -50,7 +46,104 @@ does matter!
...
+not be executed. There may be a situation where your component may be able to execute even if a particular parameter
+is not available. This is where the `Option<P>` parameter can be used. Instead of *never* executing because a parameter
+is missing, instead it will be executed, but the `Option` will be `None`.
+
+### Config\<T\> / ConfigMut\<T\>
+
+`Config<T>` and `ConfigMut<T>` paremters are available for a platform or a component to produce a generic configuration
```
os-d:
```suggestion
`Config<T>` and `ConfigMut<T>` parameters are available for a platform or a component to produce a generic configuration
```

docs/src/driver/interface.md
```diff
@@ -50,7 +46,104 @@ does matter!
...
+value that can be consumed by any other component. A platform can produce it's own values by using the `Core`'s
```
os-d:
```suggestion
value that can be consumed by any other component. A platform can produce its own values by using the `Core`'s
```

docs/src/driver/interface.md
```diff
@@ -50,7 +46,104 @@ does matter!
...
+This type comes with a `mock(...)` method to make unit testing simple.
+
+### Hob\<T\>
+
+The `Hob<T>` parameter type is used to access a guided hob value, which is automatically parsed from the HOB list
+provided to the Core during initialization. Unlike `Config<T>` and `ConfigMut<T>`, these types are **not** always
+available. This means a component that has this parameter implemented will only be executed if the guided HOB is
+found in the HOB list. Due to how HOBs work, the same guided HOB can be provided multiple times to the platform.
+Due to this, `Hob<T>` implements both `Deref` to access the first found value, or `IntoIterator` to iterate through
+all HOB values.
```
os-d, line 103: Is there any information provided by the core along the lines of: you registered a component that relies on Hob<Q>, but no such HOB was produced by the platform, component won't be executed. I may have said this before, but perhaps something for the platform verification tool @vineelko and @berlin-with0ut-return ?

docs/src/driver/interface.md
```diff
@@ -50,7 +46,104 @@ does matter!
...
+| ConfigMut\<T\>               | A mutable config value that will only be available while the underlying data is unlocked.                                         |
+| Service\<T\>                 | A wrapper for producing and consuming services of a particular interface, `T`, that is agnostic to the underlying implementation. |
+<!-- markdownlint-enable -->
+
+### Option\<P\>
+
+Some parameters are not *always* available. When a parameter is not available, the component will not be executed.
+There may be a situation where your component may be able to execute even if a particular parameter is not available.
+This is where the `Option<P>` parameter can be used. Instead of *never* executing because a parameter is missing,
+instead it will be executed, but the `Option` will be `None`.
```
os-d: Perhaps a bit more clarification here or in the two phase bit below on when `Option<P>` gets executed (presumably this will be in the second phase after all configs are locked it will get executed with None?)

docs/src/driver/interface.md
```diff
@@ -50,7 +46,104 @@ does matter!
...
+    fn write(&self, object: impl Into<Vec<u8>>) -> Result<()> {
+        self.inner.write_bytes(object.into())
+    }
+}
+```
+
+```admonish important
+Each Service type is static and immutable. What this means is that only the &self methods methods are available,
+forcing the implementor to manage their own interior mutability. This means that each each consumer receives their own
+instance and can do what they will with the service, including stashing it when producing their own service or
```
os-d: Does this mean that different structs of the service will be passed around, such that if you had some interior mutable state in your service, it would have different values across consumers, depending on what updates it?

docs/src/driver/interface.md
Javagedes: Yes, we do actually have this. It is not 100% the most clear (see below). We have logging that displays all components not executed and the component that could not be retrieved that caused the component to not be executed. [See logic here](https://github.com/OpenDevicePartnership/uefi-dxe-core/blob/55cd4fdcfe65a3cc2c536a2b47c7b0a450952ffa/dxe_core/src/lib.rs#L514)

It prints a pretty table in the log file similar to this:

| name | failed_param |
|-------|----------------|
|my_component| Hob\<MyHob\> |
|my_other_component| Service\<dyn MyService\> |

We also log any guided HOBs found that do not have a associated parser.

docs/src/driver/interface.md
Javagedes: I will add clarification here. TLDR, it is the same underlying instance that gets passed around to everyone. An update to one will update all instances. We can't pass multiple mutable instances around without some type of locking mechanism, (Which I determined to be outside the responsibility of the `Service` struct) so if a service wants to update `self`, then it must maintain its own interior mutability mechanism.

Javagedes: I've updated to the following. Please close this issue if the description is clear enough, or request more updates :)

> Each service references the same underlying static and immutable type. This means is that only the &self methods are available and forces the implementor to manage their own interior mutability via some sort of locking mechanism. Each component receives their own Service instance (All of which point back to the same underlying implementation), which allows them stash it for their own service implementations or callbacks.

docs/src/driver/interface.md
Javagedes: In most cases, I would say `Option` should not be used on `Config` / `ConfigMut`. I could see an argument for `Option<ConfigMut<T>>` as someone could say "I want this component to run, even if I can't update this particular configuration".

As it stands, however, `Option<P>` is evaluated from the get-go and is not a part of the two-phase system. I think this question brings up a good point about needing to improve the scheduler a bit. I think we may want an RFC to improve the scheduler and probably limit what people can do with these components (i.e. do we want to limit what params can be in the same component? Maybe we don't allow `ConfigMut` and `Service` in the same component because we don't want people relying on a service being available to be able to produce a configuration. Maybe some other option?)

docs/src/driver/interface.md
```diff
@@ -50,7 +46,104 @@ does matter!
     error[E0277]: the trait `function_component::ComponentParamFunction<_>` is not implemented for fn item `<fn_interface>`
 ```
 
-Please reference the documentation in `uefi_sdk::component` for more information regarding these parameters.
+<!-- markdownlint-disable -->
```
makubacki, line 59: nit: Can the sections be in the same order as the rows in the table (e.g. `Hob<T>` first, then `Option<P>`, etc.).

docs/src/driver/interface.md
makubacki: I agree that we need to work out some of the dependencies a bit more. If nothing else, to clearly explain the behavior depending on the parameter type.

In particular, I think `ConfigMut` and `Service` are useful to allow together. We can discuss more but I think we might need to more strongly delineate between "init lock" and "explicit lock" or something like that. I could see `Option<Config<T>>` making sense in the same the "opportunity for mutation" has passed and if someone produced this config use it.

docs/src/driver/interface.md
makubacki: HOB verification definitely has its place in the DXE readiness tool, but I think individual component debug needs to have an independent path from the tool checking the foundational/common DXE requirements.

docs/src/driver/interface.md
Javagedes: yep no problem. Re-ordered.

docs/src/driver/interface.md
os-d: Yep, completely agree. I do remember you presenting that list now, thanks for the reminder!

docs/src/driver/interface.md
os-d: For my understanding, if it is not `Option<Config<T>>` or the mutable version, the only other option for `<P>` is a Service, right? Or am I misunderstanding and there is something else that can be passed as a parameter? From reading this, my understanding was that you can only pass a Config, mutable or not, or a Service, as an input to a component.

But if `Option<P>` is evaluated from the get go, then it couldn't be a service, right, because some component dispatched later could publish a service? Or I may be misunderstanding how services are evaluated :). At any rate, I think some clarification in this section would be helpful, for me at least :)

docs/src/driver/interface.md
os-d: That helps, thanks! And makes sense in terms of service writing.

docs/src/driver/interface.md
makubacki: For "each type" I meant all of the pre-defined `Param` types. It is confusing because each type also brings its own properties to the table like `Config<T>` vs `ConfigMut<T>`. Essentially, I think an additional column of what `Option<T>` means for each row might be necessary.

More generally though, I think a state machine somewhere would be nice and perhaps even a state machine for each individual type (e.g. `ConfigMut<T>`) and entity (e.g. the dispatcher).

Javagedes: The current table is already big, But I will add a new table in the Option section and update the description of Option.

I think we will need another RFC to discuss the appropriate handling of each param when it is wrapped by an Option. As it stands, it is "dumb" and just marks that param type as always available, which will cause components to prematurely run, even if a different component will produce it later.

os-d: I just reread the above section and I had forgotten about `Param`. I am a bit confused on what the difference between `Param` and `Config<T>` is, e.g. is every `Param` not a type of configuration? What kind of things would be passed in that are not a configuration or a service?

Javagedes: `Param` is a trait that describes a concrete type that is allowed to be a parameter in a component. Every struct in that table implements `Param`. The interface defined by the `Param` trait is what the dispatcher uses to (1) validate that a parameter is available, (2) retrieve the particular datum from storage and (3) pass it to the component interface.

Here are a few examples:
1. `Config` -> `validate` checks that the config datum is `locked`.
2. `ConfigMut` -> `validate` checks that the config datum is `unlocked`.
3. `Service` -> `validate` checks that the service has a registered implementation
4. `Hob` -> `validate` checks that a datum has been parsed from the hoblist
5. `Option` -> Always returns true, but the `retrieve` just returns `None` if it is not actually available

As to the other part of the question, we have some other miscellaneous things that currently implement `Param`. 

- The `Storage` struct itself actually implements param, so a component can actually request access to the entire Storage object. This allows you to do many things, one of which is register a `Service`.

- `StandardBootServices` and `StandardRuntimeServices` both implement `Param` right now, which allows you to request those structs through your component interface, but I don't mention them here because the long-term goal is to probably remove them and have them be accessible via the `Service` implementation.

os-d: Thanks that clarifies. So we may possibly add another `Param` type (as `Hob` was just added) but we would expect that a component would only take in one of the 5 you listed (barring another common `Param` added) and possibly `Storage`. But that any component to component interaction would take place through `ConfigMut<T>` or `Service`.

That's what I was trying to understand, whether there was a third option as to what components may pass to one another, but I believe you are saying only `ConfigMut<T>` and `Service` can be used to communicate across components (which is what I expect, corresponding to dynamic PCDs and protocols).

Javagedes: Yes, the nice thing about this design is we can add additional implementations as needed without breaking the world. 

The way I see it is `Service` (yes analogous to protocols) is the main way that components interact with each other. The two `Config` types are not really there to communicate, though. They are simply ways to share configuration across components. There is a bit of `DynamicPcd` thrown in with `ConfigMut`, But the intended use case is for it to be manipulated and eventually locked. Not as a way to communicate by updating values over and over.

I do suspect we will add more ways for components to interact; primary I suspect we will need to add a type of `Event` parameter type. But that will need an RFC to discuss the needed functionality.

Javagedes: Updated docs.

PR #378, opened 2025-04-29
Temporarily disable setting read-only sections

makubacki:
## Description

Commit 30fa804 made sections in a PE/COFF image that are not writable but readable to have the read-only attribute set. While this has successfully booted to Windows on QEMU and no issues have been encountered in UEFI images, it does fail on physical Intel platforms, at least in some cases.

That is being tracked in:
https://github.com/OpenDevicePartnership/uefi-dxe-core/issues/365

To allow physical platforms to release the code is temporarily commented. It is left in the codebase as a reminder that it should be followed up on after the issue is closed.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- QEMU Windows boot
- Intel physical platform Windows boot

## Integration Instructions

N/A

### DIFF
--- a/dxe_core/src/image.rs
+++ b/dxe_core/src/image.rs
@@ -8,7 +8,6 @@
 //!
 use alloc::{boxed::Box, collections::BTreeMap, string::String, vec, vec::Vec};
 use core::{convert::TryInto, ffi::c_void, mem::transmute, slice::from_raw_parts};
-use goblin::pe::section_table;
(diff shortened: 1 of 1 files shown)


makubacki:
I'm open to suggestions for other ways to handle this, but we need to get physical platform releases out without an involved workaround.

PR #379, opened 2025-04-29
Move CPU functionality selection into `Core`

Javagedes:
## Description

This PR is associated with [RFC 0004](https://github.com/OpenDevicePartnership/uefi-dxe-core/blob/main/docs/src/rfc/text/0004-move-cpu-into-core.md) and depends on https://github.com/OpenDevicePartnership/uefi-core/pull/119.

This commit does the following actions:
1. Removes the `.with_cpu_init` and `.with_interrupt_manager` methods, instead selecting and instantiating the types automatically based on architecture and producing them as services (`Service<dyn Cpu>`, `Service<dyn InterruptManager>`)
2. Replaces the `with_interrupt_bases` method with a `GicBases` struct to be used as a `Config` that should be registered via `with_config` on AARCH64 architectures
3. Moves all system table init out of `init_memory` and into `initialize_system_table` that is called by `start()`
4. Adds a `add_default_components()` function that registers components from the core
5. Refactors CPU Arch Protocol and Hardware Interrupt Protocol registration functions into components that are registered and executed by the core.

- [x] Impacts functionality?
- [ ] Impacts security?
- [x] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Boot to shell on SBSA, Q35

## Integration Instructions

Platforms must remove usage of `.with_cpu_init`, `.with_interrupt_manager`, and `.with_interrupt_bases`. If the platform is an AARCH64 platform, platforms must now call `.with_config(GicBases::new(a, b))` post-memory-init.

### DIFF
--- a/dxe_core/src/hw_interrupt_protocol.rs
+++ b/dxe_core/src/hw_interrupt_protocol.rs
@@ -385,13 +385,15 @@ pub(crate) fn install_hw_interrupt_protocol(
     // Produce Interrupt Protocol with the initialized GIC
     let interrupt_protocol = Box::leak(Box::new(EfiHardwareInterruptProtocol::new(hw_int_protocol_handler)));
 
-    boot_services.install_protocol_interface(None, interrupt_protocol)
+    boot_services
+        .install_protocol_interface(None, interrupt_protocol)
         .inspect_err(|_| log::error!("Failed to install HARDWARE_INTERRUPT_PROTOCOL"))?;
 
     // Produce Interrupt Protocol with the initialized GIC
     let interrupt_protocol_v2 = Box::leak(Box::new(EfiHardwareInterruptV2Protocol::new(hw_int_protocol_handler)));
 
-    boot_services.install_protocol_interface(None, interrupt_protocol_v2)
+    boot_services
+        .install_protocol_interface(None, interrupt_protocol_v2)
         .inspect_err(|_| log::error!("Failed to install HARDWARE_INTERRUPT_PROTOCOL_V2"))?;
     log::info!("installed HARDWARE_INTERRUPT_PROTOCOL_V2");
 


dxe_core/src/hw_interrupt_protocol.rs
```diff
@@ -353,58 +365,47 @@ impl HwInterruptProtocolHandler {
...
-        return;
-    } else {
-        log::info!("GICv3 initialized");
+/// This component installs the two hardware interrupt protocols.
+pub(crate) fn install_hw_interrupt_protocol(
+    interrupt_manager: Service<dyn InterruptManager>,
+    gic_bases: Config<GicBases>,
+    bs: StandardBootServices,
+) -> Result<()> {
+    if *gic_bases == GicBases::default() {
```
os-d: is it possible to make this a build time error where you can't compile without providing the config?

dxe_core/src/lib.rs
```diff
@@ -450,8 +351,110 @@ where
...
+        // Note: I had to use the ptr instead of locking the table which event though is static does not seems to return static refs. Need to investigate.
+        unsafe {
+            self.storage.set_boot_services(StandardBootServices::new(&*boot_services_ptr));
+            self.storage.set_runtime_services(StandardRuntimeServices::new(&*runtime_services_ptr));
+        }
+
+        Ok(())
+    }
+
+    fn add_default_components(&mut self) {
```
os-d: nit: default makes this seem like they are expected to be overridden, but we don't expect that. Perhaps naming these core components or something similar?

Javagedes: Good idea. Updated to `add_core_components`

dxe_core/src/hw_interrupt_protocol.rs
Javagedes: No, I cannot think of a way to stop this at compile time, unfortunately.

What I can do is put the assert in the `default` implementation rather than check it here.

dxe_core/src/lib.rs
```diff
@@ -188,47 +191,23 @@ impl<SectionExtractor> Core<SectionExtractor, NoAlloc>
 where
     SectionExtractor: fw_fs::SectionExtractor + Default + Copy + 'static,
 {
-    /// Registers the CPU Init with it's own configuration.
-    pub fn with_cpu_init(mut self, cpu_init: EfiCpu) -> Self {
```
cfernald, line 192: Breaking change

dxe_core/src/hw_interrupt_protocol.rs
```diff
@@ -353,58 +365,40 @@ impl HwInterruptProtocolHandler {
...
-    } else {
-        log::info!("GICv3 initialized");
-    }
-
-    let mut gic_v3 = res.unwrap();
+/// This component installs the two hardware interrupt protocols.
+pub(crate) fn install_hw_interrupt_protocol(
+    interrupt_manager: Service<dyn InterruptManager>,
+    gic_bases: Config<GicBases>,
+    bs: StandardBootServices,
```
vineelko: NIT: Also here, "bs" is not very obvious :-)

dxe_core/src/lib.rs
```diff
@@ -188,47 +191,23 @@ impl<SectionExtractor> Core<SectionExtractor, NoAlloc>
...
-
     /// Initializes the core with the given configuration, including GCD initialization, enabling allocations.
     pub fn init_memory(mut self, physical_hob_list: *const c_void) -> Core<SectionExtractor, Alloc> {
         log::info!("DXE Core Crate v{}", env!("CARGO_PKG_VERSION"));
 
-        let _ = self.cpu_init.initialize();
-        self.interrupt_manager.initialize().expect("Failed to initialize interrupt manager!");
+        let mut cpu = EfiCpu::default();
+        cpu.initialize().expect("Failed to initialize CPU!");
+        let mut im = Interrupts::default();
```
vineelko: NIT: May be "interrupt_manager" is better :-) Because "im" is not very obvious.

Javagedes: Sounds good. Updated to `interrupt_manager` :)

dxe_core/src/hw_interrupt_protocol.rs
Javagedes: Souds good. updated to `boot_services` :)

dxe_core/src/lib.rs
```diff
@@ -106,6 +102,21 @@ macro_rules! error {
 
 pub(crate) static GCD: SpinLockedGcd = SpinLockedGcd::new(Some(events::gcd_map_change));
 
+#[derive(Debug, PartialEq)]
+pub struct GicBases(pub u64, pub u64);
```
cfernald, line 106: Should these by wrapped by aarch64?

dxe_core/src/lib.rs
```diff
@@ -450,8 +357,114 @@ where
...
+            self.storage.set_boot_services(StandardBootServices::new(&*boot_services_ptr));
+            self.storage.set_runtime_services(StandardRuntimeServices::new(&*runtime_services_ptr));
+        }
+
+        Ok(())
+    }
+
+    /// Registers core provided components
+    fn add_core_components(&mut self) {
+        self.insert_component(0, cpu_arch_protocol::install_cpu_arch_protocol.into_component());
```
cfernald, line 453: Why have these part of the core? Should we just break these out?

Javagedes: I'm not against it - what is your suggestion? These are fairly critical as they are registering the protocols with the protocol_db. Are you saying completely pull them out, and have the platform register them?

dxe_core/src/lib.rs
Javagedes: I found it unecessary to add it, as the linker will remove the struct if unused, but I'd be happy to add it if you wish

dxe_core/src/lib.rs
Javagedes: Added label. Good catch.

dxe_core/src/lib.rs
cfernald: No need to block this PR. We can consider in the future.

dxe_core/src/lib.rs
Javagedes: I'm not going to add the conditional compilation in this PR because we do not currently have a `doc` feature for conditional compilation to ensure this struct shows up in public docs, and I think it is outside the scope of this PR to add that. Maybe later if someone gets confused, we can add conditional compilation :)

PR #380, opened 2025-04-29
RFC: Single Patina Repo

makubacki:
## Description

Proposes moving to a single Patina repository that encompasses the main collateral of the project.

Status: Approved

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [x] Includes documentation?

## How This Was Tested

- N/A

## Integration Instructions

- N/A

### DIFF
--- a/docs/src/rfc/text/0006-patina-repo.md
+++ b/docs/src/rfc/text/0006-patina-repo.md
@@ -5,6 +5,8 @@ This RFC proposes a single Patina repository that contains all code and document
 ## Change Log
 
 - 2025-04-29: Initial draft of RFC.
+- 2025-04-30: Added a proposal for existing public repositories to eventually be archived and moved into the Patina
+  repository.
 
 ## Motivation
 


vineelko:
This is such a welcome change Michael, I have already grafted all repos to one codebase(with history preserved) a while back. Will review this RFC later.

image
image

makubacki:
> This is such a welcome change Michael, I have already grafted all repos to one codebase(with history preserved) a while back. Will review this RFC later.

I left the PR in draft state to allow others to push to the PR branch. Feel free to do so if you'd like to add anything. When we're all good with the content, I'll publish the PR. I think that will be a little faster than going through a comment and respond procedure to start.

docs/src/rfc/text/0000-patina-repo.md
```diff
@@ -0,0 +1,146 @@
...
+## Prior Art (Existing PI C Implementation)
+
+This section describes the pre-existing code organization model in place at the time of this RFC.
+
+### Prior Code Organization Model
+
+To allow better separation of maintainership, documentation, release tracking to the GitHub repository, and higher
+cohesion among repo contents, the code was organized as follows:
+
+- DXE Core Specific Crate (`uefi-dxe-core` repository)
```
vineelko: Should we call these crates (plural), since these repos indeed produce multiple crates?

image
Crates with in the repos
image

I can update the branch with these changes.

docs/src/rfc/text/0000-patina-repo.md
```diff
@@ -0,0 +1,146 @@
...
+4. **Documentation** - The Patina repository should contain documentation for all Patina crates and the Patina project.
+   - Crate-specific documentation should generally be maintained in the crate itself.
+   - Crates may also be described in common rep0-level documentation.
+
+## Unresolved Questions
+
+- **Repo Location** - Rename the `uefi-dxe-core` repository to `patina` and move all Patina crates into the repository.
+  - This is suggested to keep issue and pull request history in the repository.
+
+- Exactly how to refactor existing crates into this new organization model. This is considered out of scope for this
```
vineelko: The exact process of refactoring existing crates may be outside the scope of the current RFC, but I believe that flushing out the directory layout would still be beneficial. This will require some proof-of-concept (POC) work to determine what works and what doesn't. As part of analyzing the dependencies and grafting a monorepo, here's are some findings:

### POC Flow:
1. **Collocate all repos as-is into a single Patina monorepo**, as this would eliminate the need to raise multiple chore PRs for version bumps.  
2. **Cargo does not support nested workspace `Cargo.toml` files.** This means we will need to flatten the individual repo workspace `Cargo.toml` files into the Patina workspace `Cargo.toml`, as shown below: Still maintaining the directory layout.
image
3. Further, do we need to flatten the crates to something like below? Something like `patina-dxe-xxxx`/`patina-sdk-xxxx`/ `patina-core-xxxx`. 
image
4. Pipelines should be worked out

I can resume my grafting work to share any further findings before we can make this RFC live.

docs/src/rfc/text/0000-patina-repo.md
makubacki: > Should we call these crates (plural), since these repos indeed produce multiple crates?

Yes.

> I can update the branch with these changes.

Thanks

docs/src/rfc/text/0000-patina-repo.md
makubacki: > The exact process of refactoring existing crates may be outside the scope of the current RFC

That's fine. I was mainly wanting to avoid the bigger change this PR is making getting blocked behind ongoing code changes making its completion a moving target. Some guidelines about directory structure make sense to include.

> Further, do we need to flatten the crates to something like below? Something like patina-dxe-xxxx/patina-sdk-xxxx/ patina-core-xxxx.

I thought about this too and I'm leaning in favor of it to (1) make things obvious that are part of patina when used elsewhere (2) prevent name collisions for on crates.io.

Javagedes: My preference would be to retain some semblance of structure in these repos instead of having all of them in a single folder like is common in other projects. We can also use this to identify the "public" crates people should actually be consuming out of this repo, by not having them in one of the subfolders. This would also help us identify what should actually be a breaking change - e.g. what we would consider breaking changes in crates/*, sdk/*, core/* probably would not be a breaking change, as they are only consumed internally (even if they are a publicized crate) but anything considered a breaking change in `components/*`, `patina`, `patina-test`, or `patina-sdk` would be, as they are meant to be consumed by external entities

```
├── components/
|   ├── adv_logger
|   ├── sample_components
|   ├── mm_communicate
|   └── etc.
├── crates/ (maybe rename?)
|   └── existing crates in uefi-dxe-core
├── sdk/
|   └── existing crates in uefi-sdk (minus umbrella uefi-sdk at the top)
├── core/
|   └── existing crates in uefi-core
├── uefi-dxe-core (patina?)
├── uefi-test (patina-test?)
└── uefi-sdk (patina-sdk?, maybe in components?)
```

docs/src/rfc/text/0000-patina-repo.md
```diff
@@ -0,0 +1,146 @@
...
+## Goals
+
+1. Create a single Patina repository that contains *most*\* code and documentation for the Patina project.
+2. Reduce the amount of overhead for integration between these crates.
+3. Keep Patina issues, pull requests, and documentation in a single location to improve developer ergonomics and
+   make it easier to couple documentation updates with code changes.
+4. Ease platform integration and overrides of Patina crates by having them defined in a single local workspace.
+
+\* An exception to (1) is code that is entirely independent of Patina firmware and can benefit a larger audience. For
+example, the `paging` and `mtrr` crates have no coupling to Patina or firmware and are useful to share with a
```
Javagedes: agreed

docs/src/rfc/text/0000-patina-repo.md
```diff
@@ -0,0 +1,146 @@
...
+broader set of developers to improve the overall quality of that code. These would be maintained by the Patina team but
+considered independent repositories.
+
+## Requirements
+
+1. **Single Repository** - The Patina repository should contain all code and documentation for the Patina project.
+2. **Patina Internal Crates** - A crate in the Patina repository may be marked for "internal" use only. This means
+   that the crate is not intended to be consumed outside of the Patina repository and should not be a dependency in any
+   crate outside the Patina repository. Any crate marked as "internal" must have a name that starts with
+   `patina_internal_`.
```
Javagedes: good thought.

docs/src/rfc/text/0000-patina-repo.md
```diff
@@ -0,0 +1,146 @@
...
+3. **Versioning** - A single version of the Patina repository will be used for all Patina crates and tracked as the
```
Javagedes: agree

docs/src/rfc/text/0000-patina-repo.md
vineelko: Sure, nothing is set in stone until we have a POC to experiment with as a baseline. One of the nice things about Cargo is that we can easily realign the directory structure as long as the crate (more precisely, the package name) remains unchanged. 

That said, I’d like to get some input regarding "internal" usage: when we say "internal," does that imply the crate won't be published to crates.io? If so, aside from naming the crate `patina_internal_xxxxx`, I assume we also mark the crate like this:

```
[package]
publish = false
```

Is this alone sufficient, or do we still prefer to include `_internal_` in the crate name as well? Because marking it as not published means all other patina crates in this mono repo can use it, but other external crates cannot use it.

Javagedes: Unfortunately, all code (so crates) must be published to crates.io. But like other projects (including the ones you've referenced), even though they are published to crates.io, they are "internal" in the sense that they are not expected to be consumed directly by anyone but instead are only used by the root crate. 

Due to the need to publish them, I think including the `_internal_` works fine. I think the main thing is the readme needs to make sure it says this is not intended for public use and all consumption should be through the `patina` (or one of our other public) crate.

makubacki: Yeah, that's exactly what I had in mind - publish to crates.io only used by the root crate. I was also planning to rely on documentation to explain the internal usage. I will update the RFC to clarify exactly how this is handled.

vineelko: Thanks, Joey. That makes sense. Crates that depend on those marked with `publish = false` may fail to publish, so I guess we might have to publish all the code to crates.io. This brings me to a follow-up question:

For example, in the case of the `uefi-sdk` **repo**, we know the public interface is the `uefi_sdk` **crate**. If we don't want external users to access other crates like `uefi_sdk_macro`, `runtime_services`, or `driver_binding`, shouldn't they be made into modules within the `uefi_sdk` crate? That way, we can properly guard their usage.
image

To me they don't feel like individual crates they sound more like modules within the `uefi_sdk` crate.

Javagedes: Yes for uefi-sdk in particular, I'm OK with merging them into a single crate. There is not much code in each of those individual crates. I think that was a design decision by @magravel and/or @joschock so we want to double check with the before we solidify that.

I will say, it is beneficial for core developers to have code compartmentalized by crate for 2 reasons (1) strong decoupling and (2) compile time. By having them in separate crates, making a change to one file does not require a full re-compile, only a re-compile of the single crate (and then linking). Due to that, I think we just need to make a conscious decision on what gets merged and what does not. 

Particularly for uefi-dxe-core, in my opinion those in the crates/* folder should stay as separate crates (in the same repo) as there is a good chance it would increase the compile time by a fair bit.

docs/src/rfc/text/0000-patina-repo.md
os-d: I think this is an opportunity to revisit the naming choice of uefi-dxe-core vs uefi-core (and do we perpetuate the core name) because that has been a source of confusion, as in, wait, you have multiple cores?

Naming a subdirectory patina may be confusing if we are naming the entire repo patina? i.e. uefi-dxe-core is not Patina, right, I thought we were saying Patina is everything under the umbrella? So perhaps keeping some form of UEFI and/or DXE in uefi-dxe-core (maybe dropping DXE as an edk2ism and keeping UEFI).

But I agree that the semantics of naming don't have to be in this RFC, we can debate that in the PR that moves things.

docs/src/rfc/text/0000-patina-repo.md
joschock: The primary thing I like to harp on is "strong decoupling". I'm okay with moving them into modules, but that makes it pretty easy to be like "use mod_foo" everywhere, and without some strong external discipline (e.g. in UEFI this is enforced with CI plugins) after several years of development all boundaries have fallen.

docs/src/rfc/text/0000-patina-repo.md
```diff
@@ -0,0 +1,146 @@
...
+is provided here for reference.
+
+----
+
+The initial focus for UEFI Rust work was to:
+
+1. Create something meaningful that would add real world value to platforms by leveraging Rust's memory safety
+   capabilities.
+2. Create something that could be used in production with a reasonable time frame.
+3. Create something that could be used by the larger firmware community with code paterns, documentation, libraries,
```
joschock:
```suggestion
3. Create something that could be used by the larger firmware community with code patterns, documentation, libraries,

```

makubacki:
@makubacki - Add the public repo crate consolidation plan

docs/src/rfc/text/0000-patina-repo.md
makubacki: Fixed in https://github.com/OpenDevicePartnership/uefi-dxe-core/pull/380/commits/f5052cc470ce45acf5dec0090cb64ce2c4428182

docs/src/rfc/text/0000-patina-repo.md
makubacki: We don't need to put it in the RFC itself, but for Vineel's sake we should settle on a starting point here.

```
├── component/
|   ├── adv-logger
|   ├── sample_components
|   ├── mm_communicate
|   └── etc.
├── core/
|   ├── (crates that support core only functionality - probably internal crates)
|   ├── patina-dxe (dxe core - public crate)
├── lib/
|   └── (simple rust library crates that don't belong elsewhere - a generic fallback option)
├── sdk/
|   ├── (existing crates in uefi-sdk)
|   ├── patina-sdk (today's uefi-sdk)
├── test/
|   ├── (crates that are only used for tests)
|   ├── patina-test
├── util/
|   └── (utility crates)
```

A variation on the above. Some things could go top-level, etc. But I thought these categories would generally house things we care about with `lib` being an acknowledged fall back if a library crate is made that doesn't match a designated category.

makubacki:
> @makubacki - Add the public repo crate consolidation plan

Done

docs/src/rfc/text/0000-patina-repo.md
cfernald: nit: the difference between a lib and a util is unclear, so I'd advise merging those two, but otherwise I agree with this organization

docs/src/rfc/text/0000-patina-repo.md
```diff
@@ -0,0 +1,151 @@
...
+
+### Prior Code Organization Model
+
+To allow better separation of maintainership, documentation, release tracking to the GitHub repository, and higher
+cohesion among repo contents, the code was organized as follows:
+
+- DXE Core Specific Crate (`uefi-dxe-core` repository)
+  - Independent functionality exclusively used by the DXE Core.
+  - Examples: DXE Core, Event infrastructure, GCD, memory allocator
+- Core Specific Crate (`uefi-core` repository)
```
cfernald: might want to remove references to current repo/crate names as it will be meaningless to readers in the future. Happy to push this change if you'd prefer.

docs/src/rfc/text/0000-patina-repo.md
vineelko: Thanks for the inputs on the directory layout. Below is the repo for the POC work, I will push a grafted code in sometime( without any functional changes). My goal is first collocate all the repos into one place and make q35 build and boot. once that is done we can pretty easily restructure the layout(as we will **not be** disturbing the crates/interface boundaries). 

https://microsoft.visualstudio.com/MsUEFI/_git/patina/pullrequest/12844528

joschock: nit: `lib` often has special meaning in crate/rust context. May consider alternatives like `util` or `misc` or something like that?

makubacki: I had the same thought and I considered `misc` too. I'll propose that for now.

```
├── component/
|   ├── adv-logger
|   ├── sample_components
|   ├── mm_communicate
|   └── etc.
├── core/
|   ├── (crates that support core only functionality - probably internal crates)
|   ├── patina-dxe (dxe core - public crate)
├── misc/
|   └── (simple rust library crates that don't belong elsewhere - a generic fallback option)
├── sdk/
|   ├── (existing crates in uefi-sdk)
|   ├── patina-sdk (today's uefi-sdk)
├── test/
|   ├── (crates that are only used for tests)
|   ├── patina-test
├── util/
|   └── (utility crates)
```

> nit: the difference between a lib and a util is unclear, so I'd advise merging those two, but otherwise I agree with this organization

@vineelko, I suggest you consider whether "misc" and "util" make sense based on what you move around. "util" was meant to be more independent "utilities" like support for a specific compression algorithm. "misc" was meant to be something that needs to be shared as a library but doesn't make sense as a standalone utility.

vineelko: Looks like I have a ton of work to do before I can think about the directory layout 😄. But that makes sense, John/Michael. I'll incorporate it.

docs/src/rfc/text/0000-patina-repo.md
makubacki: Dropped them in https://github.com/OpenDevicePartnership/uefi-dxe-core/pull/380/commits/63d58badff19023665fdc3260569f67731bf3f52

docs/src/rfc/text/0000-patina-repo.md
makubacki: To reduce the number of outstanding threads in the PR, I pushed this change in https://github.com/OpenDevicePartnership/uefi-dxe-core/pull/380/commits/48b7e036dd0edce831d34ef93895001443f27687

docs/src/rfc/text/0000-patina-repo.md
makubacki: Resolving this conversation as implementation has started based on the proposal here.

docs/src/rfc/text/0000-patina-repo.md
makubacki: Closing this thread. Reopen if further discussion is needed.

makubacki:
Moving the RFC to FCP. @vineelko, please push any changes needed that have resulted from the prototype.

vineelko:
> Moving the RFC to FCP. @vineelko, please push any changes needed that have resulted from the prototype.

I do not see any more updates to the RFC Michael, it does capture the changes we are going for the monorepo(Single Repository/Patina Internal Crates/Versioning/Documentation).  Though "Patina Internal Crates" and "Documentation" updates to the actual code can come after the monorepo merge.

docs/src/rfc/text/0000-patina-repo.md
vineelko: Resolving it now.

Issue #365, opened 2025-04-18
.rdata Read Only May Cause Windows Not to Boot


makubacki:
### Is there an existing issue for this?

- [x] I have searched existing issues

### Current Behavior

The change to make section without write set that are readable read-only appears to cause some Windows boot scenarios to fail in winload. Still gathering more details about versions and features enabled.

### Expected Behavior

Windows always boots successfully.

### Steps To Reproduce

Seems to repro on Intel physical platforms with Windows 24H2 26100.2894

### Build Environment

```markdown
Windows 24H2 26100.2894
```

### Version Information

```text
Commit: 30fa804935d719bca67aa4466503a9cf0b600158
```

### Urgency

Medium

### Are you going to fix this?

I will fix it

### Do you need maintainer feedback?

No maintainer feedback needed

makubacki:
I was able to boot 26100.1.amd64fre.ge_release.240331-1435_client_enterprise_en-us_vl and the WinVOS build on Q35. I'm assigning to @cfernald to check if he can repro this locally on a device and take a look with the debugger.

cfernald:
Root caused the issue in Windows. The problem is that with this change the `GetMemoryMap` routine returns a very large number of descriptors. On the device I looked at, about **1200 entries**. Most of these entries were something along the lines of

```
00000000`6f156000   00000000`00000000  8000000000000000   EfiRuntimeServicesCode   (0n5 ) 0x1         4 Kb
00000000`6f157000   00000000`00000000  8000000000000000   EfiRuntimeServicesCode   (0n5 ) 0xa         40 Kb
00000000`6f161000   00000000`00000000  8000000000000000   EfiRuntimeServicesCode   (0n5 ) 0x4         16 Kb
00000000`6f165000   00000000`00000000  8000000000000000   EfiRuntimeServicesCode   (0n5 ) 0x3         12 Kb
00000000`6f168000   00000000`00000000  8000000000000000   EfiRuntimeServicesCode   (0n5 ) 0x3         12 Kb
00000000`6f16b000   00000000`00000000  8000000000000000   EfiRuntimeServicesCode   (0n5 ) 0x1         4 Kb
00000000`6f16c000   00000000`00000000  8000000000000000   EfiRuntimeServicesCode   (0n5 ) 0x5         20 Kb
```

[...]

PR #387, opened 2025-05-06
[REBASE & FF] Fix fragmented memory map and reinstate rdata memory protections

cfernald:
## Description

This change resolved issue #365 where the rdata memory protections were causing Windows to fail to boot. This was because it pushed an already framented memory map over some static limits that Windows has in some paths of its execution. This change fixes the fragmentation by masking attributes before the descriptors get folded and reverts the revert of the memory map change.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on Intel platform.

## Integration Instructions

N/A

### DIFF
--- a/dxe_core/src/image.rs
+++ b/dxe_core/src/image.rs
@@ -8,6 +8,7 @@
 //!
 use alloc::{boxed::Box, collections::BTreeMap, string::String, vec, vec::Vec};
 use core::{convert::TryInto, ffi::c_void, mem::transmute, slice::from_raw_parts};
+use goblin::pe::section_table;
 use mu_pi::hob::{Hob, HobList};
 use r_efi::efi;
 use uefi_device_path::{copy_device_path_to_boxed_slice, device_path_node_count, DevicePathWalker};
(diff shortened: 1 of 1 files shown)


dxe_core/src/allocator.rs
```diff
@@ -567,7 +567,7 @@ fn merge_blocks(
     previous_blocks
 }
 
-pub(crate) fn get_memory_map_descriptors() -> Result<Vec<efi::MemoryDescriptor>, EfiError> {
+pub(crate) fn get_memory_map_descriptors(attributes_mask: Option<u64>) -> Result<Vec<efi::MemoryDescriptor>, EfiError> {
```
Javagedes, line 578: nit: Please add some documentation to this function header just to explain what the mask is used for, since its usage is not immediately clear and used about 60 lines further down then the function start

dxe_core/src/allocator.rs
```diff
@@ -622,19 +622,23 @@ pub(crate) fn get_memory_map_descriptors() -> Result<Vec<efi::MemoryDescriptor>,
                 return None; //skip entries not page aligned.
             }
 
-            //TODO: update/mask attributes.
+            // Add the runtime attribute for runtime services code and data as
+            // higher level code will expect this but it is not explicitly tracked.
+            let mut attributes = match memory_type {
+                efi::RUNTIME_SERVICES_CODE | efi::RUNTIME_SERVICES_DATA => descriptor.attributes | efi::MEMORY_RUNTIME,
```
os-d, line 636: edk2 does this for all runtime types, we should probably do the same?

https://github.com/tianocore/edk2/blob/d6101acb083eb0b79e3ca35c18ea87003850e3c4/MdeModulePkg/Core/Dxe/Mem/Page.c#L1998-L2003

https://github.com/tianocore/edk2/blob/d6101acb083eb0b79e3ca35c18ea87003850e3c4/MdeModulePkg/Core/Dxe/Mem/Page.c#L49-L67

os-d: I know you just moved it, but still.

cfernald: The only one we are missing is `EfiPalCode` if I'm reading that correctly, right? We don't really expect `EfiPalCode` to be used.

os-d: Oh you are right. I didn't read the table, assumed the actual runtime types had the runtime bit set, but they don't.

PR #399, opened 2025-05-07
Patina Monorepo: Merge uefi-sdk and uefi-core into uefi-dxe-core

vineelko:
Patina Monorepo: Merge uefi-sdk and uefi-core into uefi-dxe-core

* Import all changes and history from `uefi-sdk` and `uefi-core` into `uefi-dxe-core`  - 05/06, 6 PM PST
* Ensure that all crates derive their `version`, `publish`, `repository`, `license`, and `edition` from the workspace
* Set up the Patina workspace `Cargo.toml`
* Remove the workspace `Cargo.toml` files from the `uefi-sdk` and `uefi-core` repositories
* Consolidate non-functional files into the root directory
* Consolidate the `.cargo/` and `.vscode/` directories into the root
* Reestablish the UefiRust internal artifact repository to satisfy crate dependencies
* Update the directory layout of Patina as per the [[RFC comment](https://github.com/OpenDevicePartnership/uefi-dxe-core/pull/380#discussion_r2070676315)](https://github.com/OpenDevicePartnership/uefi-dxe-core/pull/380#discussion_r2070676315)
* Set the version of Patina to **15.0.0** (`max_version(uefi-sdk, uefi-core, uefi-dxe-core) + 1`)
* `workspace.package.publish` is set to `true`, and all crates inherit this setting. Once the pipelines are fixed, all packages will be published to the UefiRust registry
* `cargo make all` works

```
├── .cargo
├── .github
├── .vscode
├── components
│ ├── adv_logger
│ └── sample_components
├── core
│ ├── dxe_core
│ ├── section_extractor
│ ├── stacktrace
│ ├── uefi_collections
│ ├── uefi_cpu
│ ├── uefi_debugger
│ ├── uefi_depex
│ ├── uefi_device_path
│ └── uefi_performance
├── docs
│ └── src
├── sdk
│ ├── boot_services
│ ├── driver_binding
│ ├── runtime_services
│ ├── tpl_mutex
│ ├── uefi_protocol
│ ├── uefi_sdk
│ ├── uefi_sdk_macro
│ ├── uefi_test
│ └── uefi_test_macro
```

**Note:** The crate interface boundaries/names remain unchanged.

- [x] Impacts functionality?
- [ ] Impacts security?
- [x] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Build Q35 and booted to shell
image

## Integration Instructions

- Pending PRs in uefi-dxe-core/uefi-core/uefi-sdk need to be rebased according to the above directory layout
- Refer [RFC: Single Patina Repo](https://github.com/OpenDevicePartnership/uefi-dxe-core/pull/380) for more details

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,26 +2,9 @@
 resolver = "2"
 
 members = [
-  "components/adv_logger",
-  "components/sample_components",
-  "core/dxe_core",
-  "core/section_extractor",
-  "core/stacktrace",
-  "core/uefi_collections",
-  "core/uefi_cpu",
-  "core/uefi_debugger",
-  "core/uefi_depex",
-  "core/uefi_device_path",
-  "core/uefi_performance",
-  "sdk/boot_services",
-  "sdk/driver_binding",
-  "sdk/runtime_services",
-  "sdk/tpl_mutex",
-  "sdk/uefi_protocol",
-  "sdk/uefi_sdk_macro",
-  "sdk/uefi_sdk",
-  "sdk/uefi_test_macro",
-  "sdk/uefi_test",
+  "components/*",
+  "core/*",
+  "sdk/*",
 ]
 
 [workspace.package]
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -1,3 +1,3 @@
 /target
-book
+*.profraw
```
Javagedes, line 2: need `book` still.

Javagedes:
@vineelko All of the crate versions we own (`uefi_sdk`, `uefi_protocol`, etc) need to stay on their current versions from their previous repository. They technically don't exist at the version you are specifying, which is what is making `cargo release` puke when doing a dry run. When we do our 15.0.0 release, the workflow should automatically bump everything to 15.0.0.

The other option is to comment out this check, as I suspect the workflow (where it actually publishes) will work fine. This is just an issue for the dry-run command. I'm OK with that solution too.

Cargo.toml
```diff
@@ -2,64 +2,258 @@
 resolver = "2"
 
 members = [
-    "crates/*",
-    "dxe_core",
-    "adv_logger",
-    "sample_components",
-    "uefi_test",
+  "./components/adv_logger",
```
Javagedes: The conventions in the rust language documentation is to now have `./` at the start of paths. I think we should follow this convention.

Cargo.toml
```diff
@@ -2,64 +2,258 @@
...
-spin = "^0.9"
-uint = "^0.10"
-uuid = { version = "^1", default-features = false }
-
-syn = "^2"
-linkme = "^0.3.29"
-quote = "^1"
-proc-macro2 = "^1"
+
+# -----------------------------------
```
Javagedes: Are you going to clean up the dependencies in another PR or in this one? I don't have an opinion other than it will need cleaned up.

core/dxe_core/Cargo.toml
```diff
@@ -6,7 +6,6 @@ repository.workspace = true
 license.workspace = true
 edition.workspace = true
 publish.workspace = true
-readme = "../README.md"
```
Javagedes, line 9: This crate still gets published. While we don't need a readme technically, we want a readme in each repository to get uploaded, which is what is shown on the crate's cargo page.

This will be true for any other crate that we removed the readme from.

core/dxe_core/Cargo.toml
```diff
@@ -51,9 +53,9 @@ uefi_performance = { workspace = true }
 # cargo-release, however they are still used when verifying the package is compileable. Any dev-dependency that is
 # local to the workspace should be added via `path` instead of `workspace = true` so that that it does not fail to
 # compile due to the dev-dependency not being published yet.
-sample_components = { path = "../sample_components" }
+sample_components = { workspace = true }
```
Javagedes: This is wrong and needs to be changed back. See the comment directly above.

This is true for any local crate in a [dev-dependency] section. Please review all and update back to local paths.

Cargo publish / cargo release cannot publish when they are workspace dependencies.

makubacki:
@vineelko, this PR must continue to use the BSD-2-Clause-Patent license. This is being tracked in https://github.com/OpenDevicePartnership/uefi-dxe-core/issues/336 and we are awaiting final legal signoff. The license change will be made in the future with that issue tracking it.

Cargo.toml
vineelko: Yaa the current paths indeed starts with `./`

Cargo.toml
vineelko: Left them as a reference/documentation. If something completely weird happens with the build etc we can quickly cross check what was before. I plan to clean them up once we get onboarded with this monorepo change.

Cargo.toml
Javagedes: Apologies I typed this in a rush before heading out the door. I actually meant the exact opposite LOL. That the conventions are not to use "./" and we should follow that :)

Cargo.toml
Javagedes: Sounds good :)

Cargo.toml
vineelko: Got it will fix it.

core/dxe_core/Cargo.toml
vineelko: Will fix it.

core/dxe_core/Cargo.toml
vineelko: Will fix it.

Cargo.toml
```diff
@@ -2,64 +2,260 @@
...
+uefi_sdk = { version = "15.0.0", path = "sdk/uefi_sdk" }
+r-efi = { version = "5.0.0", default-features = false }
+brotli-decompressor = { version = "4.0.0", default-features = false }
+alloc-no-stdlib = { version = "~2.0" }
+
+# -----------------------------------
+# core\stacktrace dependencies
+# -----------------------------------
+cfg-if = { version = "1" }
+# log = { version = "0.4", default-features = false }
```
makubacki: I'd lean toward removing these commented out dependencies since (1) it unnecessarily bloats this file (2) cannot be checked to prevent staleness (3) the same information is obtainable with `cargo tree` for those interested in it.

vineelko: Okay let me fix it.

Cargo.toml
```diff
@@ -2,64 +2,94 @@
 resolver = "2"
 
 members = [
-    "crates/*",
```
Javagedes, line 5: nit: This could realistically be

```toml
members = [
    "components/*",
    "core/*",
    "sdk/*",
]
```

Not saying you have to change it, but it may be better from a maintenance perspective. Looking for thoughts.

Cargo.toml
```diff
@@ -2,64 +2,94 @@
...
-    "dxe_core",
-    "adv_logger",
-    "sample_components",
-    "uefi_test",
+  "components/adv_logger",
+  "components/sample_components",
+  "core/dxe_core",
```
Javagedes: I just saw that we moved the most "core" crate into a sub folder. It may be too late for this (RFC merged and all), but most large repositories have their "main" crate that gets consumed at the top level. Something like

```
├── patina (crate)
├── components/*
├── core/*
├── sdk/*
```

Any reason we went with this format? Just mentioning it and seeing what thoughts are. Does not necessarily need to be changed.

Cargo.toml
vineelko: In the Azure DevOps Patina staging repo, each repo was in a subdirectory. I initially used /*, but the problem was that Cargo expects every first-level subdirectory to contain a Cargo.toml file; otherwise, it will fail. This requirement is not met by the docs folder, so I had to explicitly list the crates. I think we can do that now.

Cargo.toml
vineelko: Interesting, I didn't know that. I think this is still a fair organization of the crates. Let's get this in to unblock others, as that might require more churn/discussion.

Cargo.toml
vineelko: Fixed it. Resolving the comment

makubacki:
Closing PR since changes were force pushed

PR #400, opened 2025-05-07
Disable cargo release dry run - To pass CI for patina monorepo PR

vineelko:
## Description

Disable Cargo Release Dry Run - To Pass CI for Patina Monorepo PR
[https://github.com/OpenDevicePartnership/uefi-dxe-core/pull/399](https://github.com/OpenDevicePartnership/uefi-dxe-core/pull/399)

**Why do we need this(one off exception)?**
With the above Patina PR, we are consolidating code from three different repositories (`uefi-dxe-core`, `uefi-core`, and `uefi-sdk`) into a single versioned repository for the Patina project. As part of this consolidation, we have selected `15.0.0` as the next version for the Patina repository.

All crates now inherit the version from the Patina workspace `Cargo.toml`, and dependencies are declared as shown below:

```toml
# Cargo.toml
# sdk/boot_services dependencies
# -----------------------------------
uefi_protocol = { version = "15.0.0", path = "sdk/uefi_protocol" }
```
Currently, our CI setup for any PR performs a dry run of the release process, which includes checking whether the specified dependency versions are available in the internal registry. Since version `15.0.0` has not yet been published, the dry run fails.

This change disables the dry run in order to allow the CI to pass for the PR mentioned above. Once the PR is merged, we will follow the usual process of generating a draft release, and the `publishrelease` workflow will handle the actual publication of the crates.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

NA

## Integration Instructions

NA

### DIFF
--- a/.github/workflows/ReleaseWorkflow.yml
+++ b/.github/workflows/ReleaseWorkflow.yml
@@ -65,10 +65,11 @@ jobs:
       - name: Login to Registry
         run: cargo login "${{ secrets.CRATES_IO_TOKEN }}" --registry UefiRust
 
-      - name: Cargo Release Dry Run
-        run: cargo release ${{ steps.release_tag.outputs.tag }} --no-tag --workspace --registry UefiRust
-        env:
-          RUSTC_BOOTSTRAP: 1
+      # Commenting below to allow the Patina monorepo PR to skip below dry run for one off release.
+      # - name: Cargo Release Dry Run
+      #   run: cargo release ${{ steps.release_tag.outputs.tag }} --no-tag --workspace --registry UefiRust
+      #   env:
+      #     RUSTC_BOOTSTRAP: 1
 
       - name: Cargo Release
         run: cargo release ${{ steps.release_tag.outputs.tag }} -x --no-tag --no-confirm --workspace --registry UefiRust


.github/workflows/CiWorkflow.yml
```diff
@@ -104,10 +104,11 @@ jobs:
         env:
           RUSTC_BOOTSTRAP: 1
```
Javagedes, line 106: Hi @vineelko,

I was wrong, after some thinking. In addition to this comment out (which is 100% correct), we do actually also need to comment out the same dry run in the ReleaseWorkflow.yml :)

.github/workflows/CiWorkflow.yml
makubacki, line 1: To us closer to what's happening the reason for this is obvious, but can you please elaborate the reason this is done in the bigger picture in the PR description for others?

.github/workflows/CiWorkflow.yml
vineelko: Yeah, makes sense now, this is for when the code is checked in for the first real publish. Got it.

.github/workflows/CiWorkflow.yml
vineelko: Sure Michael, Will add that.

Issue #323, opened 2025-03-27
[Bug]: Panic - Did not find MemoryAllocationModule Hob for DxeCore. Use uefi_sdk::guid::DXE_CORE as FFS GUID. stack backtrace:


HappyCoding-2025:
### Is there an existing issue for this?

- [x] I have searched existing issues

### Current Behavior

Panics during code coverage for lib.rs module  Lib.rs module is unique as it inits the entire dxe-core and hence calls tons of itnernal functions in other modules. 

allocator::tests::lib_module_coverage_through_allocator_module stdout ----
Coverage for initializing Cpu!
Coverage for initializing Interrupt Manager!
thread 'allocator::tests::lib_module_coverage_through_allocator_module' panicked at dxe_core\src\image.rs:499:10:
Did not find MemoryAllocationModule Hob for DxeCore. Use uefi_sdk::guid::DXE_CORE as FFS GUID.
stack backtrace:
   0:     0x7ff6a457c421 - std::backtrace_rs::backtrace::dbghelp64::trace
                               at /rustc/9fc6b43126469e3858e2fe86cafb4f0fd5068869\library/std\src\..\..\backtrace\src\backtrace\dbghelp64.rs:91
   1:     0x7ff6a457c421 - std::backtrace_rs::backtrace::trace_unsynchronized
                               at /rustc/9fc6b43126469e3858e2fe86cafb4f0fd5068869\library/std\src\..\..\backtrace\src\backtrace\mod.rs:66
   2:     0x7ff6a457c421 - std::sys::backtrace::_print_fmt
                               at /rustc/9fc6b43126469e3858e2fe86cafb4f0fd5068869\library/std\src\sys\backtrace.rs:66
   3:     0x7ff6a457c421 - std::sys::backtrace::impl$0::print::impl$0::fmt
                               at /rustc/9fc6b43126469e3858e2fe86cafb4f0fd5068869\library/std\src\sys\backtrace.rs:39
   4:     0x7ff6a459a81a - core::fmt::rt::Argument::fmt
                               at /rustc/9fc6b43126469e3858e2fe86cafb4f0fd5068869\library/core\src\fmt\rt.rs:177
   5:     0x7ff6a459a81a - core::fmt::write
                               at /rustc/9fc6b43126469e3858e2fe86cafb4f0fd5068869\library/core\src\fmt\mod.rs:1189
[...]

PR #402, opened 2025-05-08
Support for Hoblist with dxe core in allocation hob

HappyCoding-2025:
Fixes https://github.com/OpenDevicePartnership/patina/issues/323 
This is already reviewed code and was asked to move this code as part of another PR as per Michael. Hence raising another PR for this.

Also  - this addresses the comments given from  https://github.com/OpenDevicePartnership/uefi-dxe-core/pull/326.

The current hoblist in testsupport crate is not sufficient and does not contain the dxe-core HOB with dxe core efi filled in expected by our initiatlisation. When existing hoblist is used, it panics. 
Hence there was a need to create another hoblist with dxe-core and fill in with dxe EFI.
Based on comments from https://github.com/OpenDevicePartnership/uefi-dxe-core/pull/326.
The comment mentioned on taking only the hoblist part and push to main.
Hence creating this issue for tracking.

### DIFF
--- a/docs/src/rfc/text/0008-consolidate-sdk-crates-into-patina-sdk-modules.md
+++ b/docs/src/rfc/text/0008-consolidate-sdk-crates-into-patina-sdk-modules.md
@@ -1,0 +1,134 @@
+# RFC: Consolidate SDK Crates into `patina_sdk` and `patina_sdk_macro` as Modules
+
+This RFC proposes consolidating the existing crates inside the Patina `sdk`
+directory into the `patina_sdk` and `patina_sdk_macro` crates, organized by
+modules. No functional changes to the modules are intended.
+
+## Change Log
+
+- 2025-05-20: Initial draft of the RFC.
+- 2025-05-22: Final updates during FCP
+  - `patina_test_macro` will be merged into `patina_sdk_macro`, but
+  `patina_sdk_macro` will remain a separate crate from `patina_sdk` because
+  procedural macros must be defined in a separate crate.
+  - Rename `0000-consolidate-sdk-crates-into-patina-sdk-modules` to
+    `0008-consolidate-sdk-crates-into-patina-sdk-modules`
+
+## Motivation
+
+Following the creation of the Patina monorepo, and in line with the [Crates
+Renaming RFC](0007-patina-crates-naming-and-categorization.md), there is now a
+scope to consolidate the crates located in the `sdk` directory. There was
+[consensus](https://github.com/OpenDevicePartnership/patina/pull/408#discussion_r2089478969)
+on merging some of these crates (e.g., runtime services, UEFI protocol) into
+`patina_sdk`, organized via a module hierarchy. This RFC formalizes and
+continues that discussion.
+
+## Technology Background
+
+This RFC does not affect any specific technologies or alter the functionality of
+the modules.
+
+## Goals
+
+1. Reduce the number of standalone crates present in the `sdk` directory.
+2. Reduce the number of standalone crates published to [crates.io](https://crates.io).
+3. Preserve the current crate or module boundaries.
+4. Consolidate the crates into `patina_sdk` and `patina_sdk_macro`.
+5. Leverage Rust's default [module visibility
+   rules](https://doc.rust-lang.org/reference/visibility-and-privacy.html),
+   using `pub(crate)` and similar mechanisms to control visibility both inside
+   and outside the `patina_sdk` crate.
+6. Consolidate the following crates, which are already being re-exported in
+   `patina_sdk` via `pub use`, as modules within the crate:
+   - `patina_boot_services`
+   - `patina_driver_binding`
+   - `patina_runtime_services`
+   - `patina_tpl_mutex`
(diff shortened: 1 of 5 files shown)


HappyCoding-2025:
@makubacki  @Javagedes  @joschock  - All the checks have passed. Please review when available.

patina_dxe_core/src/test_support.rs
```diff
@@ -10,13 +10,17 @@
 //!
 use crate::{protocols::PROTOCOL_DB, GCD};
 use core::ffi::c_void;
+use mu_pi::hob::Hob::MemoryAllocationModule;
```
joschock: if these are just used in the test module below, consider moving them down to the uses section of the test module (around line 326). I think they only need to be up in the main uses module if they are part of the test support implementation that other modules are using.

patina_dxe_core/src/test_support.rs
```diff
@@ -316,3 +320,174 @@ pub(crate) fn build_test_hob_list(mem_size: u64) -> *const c_void {
...
+
+            //PHIT HOB
+            core::ptr::copy(&phit, cursor as *mut hob::PhaseHandoffInformationTable, 1);
+            cursor = cursor.offset(phit.header.length as isize);
+
+            //CPU HOB
+            core::ptr::copy(&cpu, cursor as *mut hob::Cpu, 1);
+            cursor = cursor.offset(cpu.header.length as isize);
+
+            //resource descriptor HOBs - see above comment
```
joschock: which comment does this refer to? It's not clear to me what it is referencing.

patina_dxe_core/src/test_support.rs
```diff
@@ -316,3 +320,174 @@ pub(crate) fn build_test_hob_list(mem_size: u64) -> *const c_void {
...
+mod tests {
+    use super::*;
+    use crate::c_void;
+    use crate::test_support::get_memory;
+    use crate::test_support::header;
+    use crate::test_support::hob;
+    use crate::test_support::BootMode;
+    use patina_sdk::guid;
+
+    /*  Compact Hoblist with DXE core Alloction hob */
```
joschock: Recommend converting function header comments to rust doc comment style: https://doc.rust-lang.org/rust-by-example/meta/doc.html#doc-comments

patina_dxe_core/src/test_support.rs
HappyCoding-2025: Moved to Test Section. Resolved.

patina_dxe_core/src/test_support.rs
HappyCoding-2025: Modified the comment. This is for Resource Descriptor hob. Resolved.

patina_dxe_core/src/test_support.rs
HappyCoding-2025: Thanks John. Resolved all the comments to Rust Style.

patina_dxe_core/src/test_support.rs
```diff
@@ -316,3 +319,175 @@ pub(crate) fn build_test_hob_list(mem_size: u64) -> *const c_void {
...
+            let memory_slice = slice::from_raw_parts_mut(memory_base_address as *mut u8, memory_length as usize);
+            let file_size = file_size as usize; // Convert file_size to usize
+            memory_slice[..file_size].copy_from_slice(&image);
+        }
+
+        Ok(())
+    }
+
+    #[test]
+    fn test_build_test_hob_list_compact() {
```
makubacki, line 496: You could add some asserts to this function to verify `fill_file_buffer_in_memory_allocation_module()` copied the contents as expected and the overall HOB list produced by `build_test_hob_list_compact`.

patina_dxe_core/src/test_support.rs
```diff
@@ -316,3 +319,175 @@ pub(crate) fn build_test_hob_list(mem_size: u64) -> *const c_void {
...
+    use crate::c_void;
+    use crate::test_support::get_memory;
+    use crate::test_support::header;
+    use crate::test_support::hob;
+    use crate::test_support::BootMode;
+    use mu_pi::hob::Hob::MemoryAllocationModule;
+    use patina_sdk::guid;
+
+    // Compact Hoblist with DXE core Alloction hob
+    pub(crate) fn build_test_hob_list_compact(mem_size: u64) -> *const c_void {
```
makubacki, line 335: `build_test_hob_list_compact()` only needs to exist here if used somewhere else (besides the test below) that can make use of it. Can you please update the function comment to explain when this function should be used?

patina_dxe_core/src/test_support.rs
HappyCoding-2025: Resolved. I added assert in fill_file_buffer_in_memory_allocation_module to verify that contents copied are same as that of dxe image.

patina_dxe_core/src/test_support.rs
HappyCoding-2025: Updated. The existing hob list functions do not work as they do not contain dxe core or other memory allocation hobs required by our dxe-core.  The compact one works well with current code.

PR #405, opened 2025-05-08
Make performance lib into a component.

magravel:
## Description

Change the performance library into a component, refactor parts of the code for it to be more testable and add unit tests.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [X] Includes tests?
- [X] Includes documentation?

## How This Was Tested
- Unit tests
- On QEMU Q35
- Intel devkit

## Integration Instructions

This Performance component replaced DxeCorePerformanceLib.c. This means that if the device is not built using performance, this component should not be added and if it is built with performance, it should be there. This is because the component publishes the fbpt but when perf is not enabled, FirmwarePerformanceDxe.c creates that table. This is the PCD that tells whether perf is enabled or not: gEfiMdePkgTokenSpaceGuid.PcdPerformanceLibraryPropertyMask. Note that in rust the value of the pcd is not used and to enable some measurement, config must be used.

### DIFF
--- a/components/patina_performance/src/lib.rs
+++ b/components/patina_performance/src/lib.rs
@@ -29,7 +29,6 @@ use core::{
     mem::MaybeUninit,
     ptr,
     sync::atomic::{AtomicBool, AtomicU32, Ordering},
-    u32,
 };
 use mu_pi::status_code::{EFI_PROGRESS_CODE, EFI_SOFTWARE_DXE_BS_DRIVER};
 
@@ -109,7 +108,7 @@ pub struct EnabledMeasurement(pub &'static [Measurement]);
 
 impl EnabledMeasurement {
     pub fn mask(&self) -> u32 {
-        self.0.into_iter().fold(0, |mask, m| mask | m.as_u32())
+        self.0.iter().fold(0, |mask, m| mask | m.as_u32())
     }
 }
 
@@ -633,7 +632,7 @@ mod test {
     use super::*;
 
     use alloc::rc::Rc;
-    use core::{assert_eq, ptr, u32};
+    use core::{assert_eq, ptr};
 
     use mockall::predicate;
 
--- a/components/patina_performance/src/performance_table.rs
+++ b/components/patina_performance/src/performance_table.rs
@@ -130,7 +130,7 @@ impl FBPT {
                         allocation_nb_page,
                     )
                 },
-                |addr| Ok(addr),
+                Result::Ok,
             )? as *mut u8;
 
         // SAFETY: the allocation at this addres was of size `allocation_size`


magravel:
Some comments are in this PR: https://github.com/OpenDevicePartnership/patina/pull/351

core/uefi_performance/src/lib.rs
```diff
@@ -1,461 +1,294 @@
...
-            if attribute == PerfAttribute::PerfStartEntry {
-                Ok(PerfId::PERF_IN_MODULE_START)
-            } else {
-                Ok(PerfId::PERF_IN_MODULE_END)
-            }
-        } else {
-            Err(efi::Status::INVALID_PARAMETER)
-        }
-    }
+pub struct Performance;
```
berlin-with0ut-return: if this structure really doesn't need any internal state, why isn't it a functional component? alternatively, could we store state in this structure to reduce the amount of `static` and `static mut`? for example storing boot services as a 'static field inside `Performance` and having some of the `extern "efiapi"` functions be impls on `Performance`?

@Javagedes helped me with something similar on ACPI. not sure if it could be applicable here.

core/uefi_performance/src/log_perf_measurement.rs
```diff
@@ -0,0 +1,487 @@
...
+
+use crate::{
+    performance_measurement_protocol::{CreateMeasurement, PerfAttribute},
+    KnownPerfId,
+};
+
+/// Create performance record
+///
+/// `caller_identifier` is either a Handle or a pointer to a caller ID GUID.
+fn log_perf_measurement(
```
berlin-with0ut-return: maybe this is too much work for now, but i think one of the original goals of rewriting things as components was to make things "nicer" for native rust use. for example, returning errors more useful than efi::Status, avoiding dealing with raw C pointers, etc. especially for the functions we're currently using to instrument Rust code

core/uefi_performance/src/lib.rs
```diff
@@ -549,469 +562,317 @@ fn get_module_info_from_handle(
...
 
-#[macro_export]
-macro_rules! perf_in_module_begin {
-    ($measurement_str:expr, $caller_id:expr) => {
-        if $crate::PERF_ENABLED {
-            $crate::_perf_in_module_begin($measurement_str, $caller_id)
-        }
-    };
-}
+        _ = perf_image_start_begin(module_handle, test_create_performance_measurement);
```
berlin-with0ut-return: we should try to use the return values of these Results to propogate up errors (`?`) or at least log something

core/uefi_performance/src/lib.rs
magravel: Do you have a branch I can look at? I know something we hide state at the end of a struct when the protocol uses a pointer to self, but I can't use this technique here.

core/uefi_performance/src/log_perf_measurement.rs
magravel: Changing errors to something else that efi::error would be quite significant and need more thinking, recently the trait `core::error::Error` became stable, and maybe we would want to have a conversation on how to handle custom error with that trait. That would unlock the possibility of using `Box<dyn Error>` as return values. I do not think this would be in the scope of this PR though.

core/uefi_performance/src/lib.rs
magravel: Before the error was ignored earlier, and I decided to move that decision to the caller of the perf logging just in case someone would be interested in looking at it for debug purposes.

core/uefi_performance/src/log_perf_measurement.rs
makubacki: Since this is being reworked as a component, it should follow the guidance in our documentation ([`efi::Status vs. Rust Errors`](https://sturdy-adventure-nv32gqw.pages.github.io/dev/principles/error-handling.html#efistatus-vs-rust-errors)) which is to use custom Rust errors in code like this.

Cargo.toml
```diff
@@ -1,30 +1,26 @@
...
 edition = "2021"
 repository = "https://github.com/OpenDevicePartnership/patina"
-publish = ["patina-fw"]
+publish = true
 
 [workspace.dependencies]
 alloc-no-stdlib = { version = "~2.0" }
 arm-gic = { version = "^0.1.2" }
 bitfield-struct = { version = "0.10" }
-boot_services = { version = "15.0.0", path = "sdk/boot_services", registry = "patina-fw" }
```
joschock: these changes seem to be unrelated? maybe merge issue?

components/performance/src/lib.rs
```diff
@@ -0,0 +1,890 @@
...
+        // SAFETY: This is safe because the state has been init.
+        unsafe { Some((BOOT_SERVICES.assume_init_ref(), FBPT.assume_init_ref())) }
+    } else {
+        None
+    }
+}
+
+/// Performance Component.
+#[derive(IntoComponent)]
+
```
joschock: nit: extra newline?

Cargo.toml
```diff
@@ -1,20 +1,17 @@
...
-]
+members = ["components/*", "core/*", "sdk/*"]
 
 [workspace.package]
 version = "15.1.0"
 license = "BSD-2-Clause-Patent"
 edition = "2021"
 repository = "https://github.com/OpenDevicePartnership/patina"
-publish = ["patina-fw"]
+publish = true
```
makubacki: This should remain at `patina-fw` as we only want to publish to that registry right now.

Cargo.toml
```diff
@@ -66,6 +62,8 @@ x86_64 = { version = "=0.15.2", default-features = false }
 criterion = { version = "^0.5" }
 mockall = { version = "0.13.0" }
 rand = { version = "0.8" }
+sample_components = { version = "15.0.0", path = "components/sample_components" }
+section_extractor = { version = "15.0.0", path = "core/section_extractor" }
```
makubacki: `sample_components` and `section_extractor` were intentionally removed earlier. Are you intentionally adding them here or was this related to rebasing?

components/performance/src/error.rs
```diff
@@ -0,0 +1,48 @@
...
+        debug_assert!($expression, $msg);
+        Err($crate::error::Error::DebugAssert { msg: $msg, file: file!(), line: line!() })
+    }};
+    ($msg:literal) => {
+        error_debug_assert!(false, $msg)
+    };
+}
+
+#[derive(Debug)]
+pub enum Error {
```
makubacki: Since this is a general performance error type, it is anticipated to only reflect the FBPT (which is sub component in the overall perforamnce domain)?

magravel: This Error type if for the whole module, For now, the only custom thing that could happen is related to FBPT, but this could have more.

components/performance/src/lib.rs
```diff
@@ -0,0 +1,889 @@
+//! A library that enables performance analysis of every step of the UEFI boot process.
+//! The Performance component install a protocol that can be used by other libraries or drivers to publish performance reports.
```
makubacki:
```suggestion
//! The Performance component installs a protocol that can be used by other libraries or drivers to publish performance reports.
```

components/performance/Cargo.toml
```diff
@@ -23,6 +19,6 @@ mu_rust_helpers = { workspace = true }
 uefi_sdk = { workspace = true }
 uefi_device_path = { workspace = true }
 
-
-[features]
-instrument_performance = []
+[dev-dependencies]
+uefi_sdk = { workspace = true, features = ["mockall"] }
```
Javagedes: This needs to be a path (not workspace) or cargo release breaks when doing a release. 

Like this:

https://github.com/OpenDevicePartnership/patina/blob/d9999162a6d95816c8826911ddf5247c11211f52/core/dxe_core/Cargo.toml#L53-L60

components/performance/src/_smm.rs
```diff
@@ -1,71 +1,59 @@
...
+
+    fn parse(bytes: &[u8]) -> Self {
+        bytes.pread(0).unwrap()
+    }
 }
 
-impl<'a> Iterator for SmmCommunicationRegionTableIter<'a> {
-    type Item = &'a efi::MemoryDescriptor;
+impl MmCommRegion {
+    pub fn is_supervisor_type(&self) -> bool {
```
Javagedes: nit: This tehcnically needs documentation since it is public, but if @makubacki is going to come in and replace this with his mm communicate service, then it's probably not necessary.

components/performance/src/lib.rs
```diff
@@ -0,0 +1,889 @@
...
+        EFI_PROGRESS_CODE,
+        EFI_SOFTWARE_DXE_BS_DRIVER,
+        0,
+        &mu_rust_helpers::guid::CALLER_ID,
+        efi::Guid::clone(&EDKII_FPDT_EXTENDED_FIRMWARE_PERFORMANCE),
+        fbpt_address,
+    );
+
+    if status.is_err() {
+        log::error!("Performance: Fail to report FBPT status code.");
```
makubacki: Is "FBPT" and "FPDT" being used accurately everywhere in this file?

components/performance/src/lib.rs
```diff
@@ -0,0 +1,889 @@
...
+
+    let Ok(fbpt_address) = fbpt.lock().report_table(
+        performance_table::find_previous_table_address(runtime_services.as_ref()),
+        boot_services.as_ref(),
+    ) else {
+        log::error!("Performance: Fail to report FPDT.");
+        return;
+    };
+
+    const EFI_SOFTWARE: u32 = 0x03000000;
```
makubacki: Can you use the definition from https://github.com/microsoft/mu_rust_pi/blob/1205e8b54abc50231278a0854c2925144e24ccb1/src/status_code.rs#L78?

components/performance/src/lib.rs
```diff
@@ -0,0 +1,889 @@
...
+    const EFI_PROGRESS_CODE: u32 = 0x00000001;
+    const EFI_SOFTWARE_DXE_BS_DRIVER: u32 = EFI_SOFTWARE | 0x00050000;
```
makubacki: Similarly, try to use others like https://github.com/microsoft/mu_rust_pi/blob/1205e8b54abc50231278a0854c2925144e24ccb1/src/status_code.rs#L519.

components/performance/src/_smm.rs
makubacki: Yeah, let's just get this in so it can be replaced.

components/performance/src/lib.rs
```diff
@@ -0,0 +1,889 @@
...
+            fbpt_address as *mut c_void,
+        )
+    };
+    if status.is_err() {
+        log::error!("Performance: Fail to install configuration table for FPDT firmware performance.");
+    }
+}
+
+/// Event callback that add the SMM performance record to the FBPT.
+extern "efiapi" fn fetch_and_add_smm_performance_records<BB, B, F>(
```
makubacki: It'd be more precise to refeer to content here as "MM" instead of "SMM", but it's not a big deal. Feel free to ignore.

components/performance/src/lib.rs
```diff
@@ -0,0 +1,889 @@
...
+                status
+            );
+            return;
+        }
+    };
+
+    let mut smm_boot_records_data = Vec::with_capacity(boot_record_size);
+
+    while smm_boot_records_data.len() < boot_record_size {
+        // SAFETY: Is safe to use because the memroy region commes for a thrusted source and can be considered valid.
```
makubacki:
```suggestion
        // SAFETY: Is safe to use because the memory region commes for a thrusted source and can be considered valid.
```

components/performance/src/performance_table.rs
```diff
@@ -0,0 +1,486 @@
...
+    ) -> Result<usize, Error>;
+}
+
+/// Firmware Basic Boot Performance Table (FBPT)
+#[derive(Debug)]
+pub struct FBPT {
+    /// When the table will be reported, this will be the address where the fbpt table is.
+    fbpt_address: usize,
+    /// First value is the length when the table is not been reported and the second one is when the table is reported.
+    /// Use `length()` or `length_mut()`. Do now use this field directly.
```
makubacki:
```suggestion
    /// Use `length()` or `length_mut()`. Do not use this field directly.
```

components/performance/src/performance_table.rs
```diff
@@ -0,0 +1,486 @@
...
+            .map_or_else(
+                || {
+                    // Allocate at a new address if no address found or if the previous address allocation failed.
+                    boot_services.allocate_pages(
+                        AllocType::MaxAddress(u32::MAX as usize),
+                        MemoryType::RESERVED_MEMORY_TYPE,
+                        allocation_nb_page,
+                    )
+                },
+                Result::Ok,
```
makubacki:
```suggestion
                |addr| Ok(addr),
```

components/performance/src/performance_table.rs
```diff
@@ -0,0 +1,486 @@
...
+        efi::Guid::from_fields(0xc095791a, 0x3001, 0x47b2, 0x80, 0xc9, &[0xea, 0xc7, 0x31, 0x9f, 0x2f, 0xa4]);
+}
+
+impl TryFrom<Vec<u8>> for FirmwarePerformanceVariable {
+    type Error = ();
+
+    fn try_from(value: Vec<u8>) -> Result<Self, Self::Error> {
+        if value.len() == mem::size_of::<Self>() {
+            // SAFETY: This is safe because the value for ADDRESS_VARIABLE_GUID is an address where a FirmwarePerformanceVariable is.
+            Ok(unsafe { ptr::read_unaligned(value.as_ptr() as *const FirmwarePerformanceVariable) })
```
makubacki: Is potential misalignment of `value` a concern?

components/performance/src/performance_table.rs
```diff
@@ -0,0 +1,486 @@
...
+        &mut self,
+        address: Option<usize>,
+        boot_services: &B,
+    ) -> Result<usize, Error> {
+        let fbpt_buffer = self.allocate_table_buffer(address, boot_services)?;
+
+        self.fbpt_address = fbpt_buffer.as_ptr() as usize;
+
+        let mut offset = 0;
+        fbpt_buffer.gwrite(Self::SIGNATURE, &mut offset).unwrap();
```
makubacki: I think we should try to avoid `unwrap()` in this function in favor of error propagation.

components/performance/src/performance_record.rs
```diff
@@ -0,0 +1,353 @@
...
+        Self::new()
+    }
+}
+
+impl Debug for PerformanceRecordBuffer {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        let _is_published = match self {
+            Self::Unpublished(_) => true,
+            Self::Published(_, _) => false,
+        };
```
makubacki: Was this meant to be:

```suggestion
        let _is_published = match self {
            Self::Unpublished(_) => false,
            Self::Published(_, _) => true,
        };
```

components/performance/src/performance_record.rs
```diff
@@ -0,0 +1,353 @@
...
+
+impl Default for PerformanceRecordBuffer {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl Debug for PerformanceRecordBuffer {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        let _is_published = match self {
```
makubacki: It seems like `_is_published` is not actually used.

components/performance/src/error.rs
makubacki: Then I'd drop the FBPT assumptions around the error codes (comments) and this:

```rust
        match self {
            Error::OutOfResources => write!(f, "FBPT buffer full, can't add more performance records."),
            Error::BufferTooSmall => write!(f, "Buffer to small to allocate FBPT table"),
            Error::Efi(efi_error) => write!(f, "{efi_error:?}"),
            Error::DebugAssert { msg, file, line } => write!(f, "Assertion at {file}:{line}: {msg}"),
        }
    }
```

components/performance/src/performance_record.rs
```diff
@@ -0,0 +1,353 @@
...
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.buffer.is_empty() {
+            return None;
+        }
+        let mut offset = 0;
+        let record_type = self.buffer.gread::<u16>(&mut offset).unwrap();
+        let length = self.buffer.gread::<u8>(&mut offset).unwrap();
+        let revision = self.buffer.gread::<u8>(&mut offset).unwrap();
+
+        let data = &self.buffer[offset..length as usize];
```
makubacki: Just looking at the range itself... Can you quickly confirm this is expected versus something like `offset..offset + (length as usize)`?

components/performance/src/lib.rs
magravel: Some FPDT comes from the naming on the C side, but I think everything should be fbpt. Will double-check.

components/performance/src/performance_record.rs
makubacki: We spoke offline and based on the explanation of how the fields are used, this is fine.

components/performance/src/lib.rs
```diff
@@ -0,0 +1,886 @@
...
+    let (boot_services, mm_comm_region, fbpt) = *ctx;
+    let _ = boot_services.as_ref().close_event(event);
+
+    // SAFETY: This is safe because the reference returned by locate_protocol is never mutated after installation.
+    let Ok(communication) = (unsafe { boot_services.as_ref().locate_protocol::<CommunicateProtocol>(None) }) else {
+        log::error!("Performance: Could not locate communicate protocol interface.");
+        return;
+    };
+
+    // SAFETY: Is safe to use because the memory region comes for a trusted source and can be considered valid.
```
joschock: nit
```suggestion
    // SAFETY: Is safe to use because the memory region comes from a trusted source and can be considered valid.
```

components/performance/src/lib.rs
```diff
@@ -0,0 +1,886 @@
...
+                status
+            );
+            return;
+        }
+    };
+
+    let mut smm_boot_records_data = Vec::with_capacity(boot_record_size);
+
+    while smm_boot_records_data.len() < boot_record_size {
+        // SAFETY: Is safe to use because the memory region commes for a thrusted source and can be considered valid.
```
joschock: nit:
```suggestion
        // SAFETY: Is safe to use because the memory region commes from a thrusted source and can be considered valid.
```

components/performance/src/lib.rs
```diff
@@ -0,0 +1,886 @@
...
+        match unsafe {
+            // Ask smm to return us the next bytes in its buffer.
+            communication
+                .communicate(SmmFpdtGetRecordDataByOffset::<1024>::new(smm_boot_records_data.len()), mm_comm_region)
```
joschock: consider making a named `const` here for clarity on what this number is.

components/performance/src/lib.rs
```diff
@@ -0,0 +1,886 @@
...
+    address: usize,
+    identifier: u32,
+    attribute: PerfAttribute,
+) -> efi::Status {
+    let Some((boot_services, fbpt)) = get_static_state() else {
+        // If the state is not initialized, it is because perf in not enabled.
+        return efi::Status::SUCCESS;
+    };
+
+    let string = unsafe { string.as_ref().map(|s| CStr::from_ptr(s).to_str().unwrap().to_string()) };
```
joschock: consider `to_string_lossy()` to avoid panic if caller-provided string isn't valid UTF-8.

components/performance/src/lib.rs
```diff
@@ -0,0 +1,886 @@
...
+            let record = GuidQwordEventRecord::new(perf_id, 0, timestamp, guid, address as u64);
+            fbpt.lock().add_record(record)?;
+        }
+        KnownPerfId::ModuleDbStopEnd => {
+            let module_handle = caller_identifier as efi::Handle;
+            let Ok(guid) = get_module_guid_from_handle(boot_services, module_handle) else {
+                log::error!("Performance Lib: Could not find the guid for module handle: {:?}", module_handle);
+                return Err(EfiError::InvalidParameter.into());
+            };
+            // TODO: use of component 2 protocol, need usecase to test further. https://github.com/OpenDevicePartnership/pantina/issues/192
```
joschock: Issue is closed; should we remove the todo comment? Also `pantina`->`patina` if retained.

components/performance/src/lib.rs
```diff
@@ -0,0 +1,886 @@
...
+            }
+        }
+        None
+    };
+
+    if let Some(loaded_image) = loaded_image_protocol {
+        // SAFETY: File path is a pointer from C that is valid and of type Device Path (efi).
+        if let Some(file_path) = unsafe { loaded_image.file_path.as_ref() } {
+            if file_path.r#type == TYPE_MEDIA && file_path.sub_type == Media::SUBTYPE_PIWG_FIRMWARE_FILE {
+                guid = unsafe { ptr::read(loaded_image.file_path.add(1) as *const efi::Guid) }
```
joschock: I'm not sure what the `.add(1)` is doing? Maybe add a comment?

components/performance/src/performance_record.rs
```diff
@@ -0,0 +1,353 @@
...
+    Unpublished(Vec<u8>),
+    Published(&'static mut [u8], usize),
+}
+
+impl PerformanceRecordBuffer {
+    pub const fn new() -> Self {
+        Self::Unpublished(Vec::new())
+    }
+
+    /// Add a performance records to the buffer.
```
joschock: nit:
```suggestion
    /// Add a performance record to the buffer.
```

components/performance/src/performance_table.rs
magravel: Since every record is written in memory one after another without padding between them, sometime values are not properly aligned. I remember adding read_unaligned after having a crash.

PR #406, opened 2025-05-08
RFC: `Unstable Feature`

magravel:
# RFC: `Unstable Feature`

Phase: FCP

This RFC is to add a way for feature gating new features that we consider unstable. This would include an issue template and a way of feature gating the unstable feature.

Use cases I see that would make sense to have unstable are:

- New feature, this gives time to get additional feedback to apply changes. Having a transition period allows to make API changes without incrementing the major version.
- Working feature that still has some unresolved questions that has a potential to change the API design.
- Proof of concept features.

## Change Log

- 2025-05-08: Initial RFC created.

## Dev flow example

- Add a feature to the cargo.toml named `unstable-<feature-name>
- Use `#[cfg(feature = "unstable-<feature-name>")]` to feature gate the unstable code.
- Create an issue to track the stabilization of the feature.

Example of issue template for unstable feature stabilization:
```yml

# GitHub issue form for tracking unstable features.
#
##
# Copyright (c) Microsoft Corporation.
# SPDX-License-Identifier: BSD-2-Clause-Patent
##
name: 📓 Tracking unstable feature issue.
description: "Track an unstable feature."
title: "[Unstable] Tracking Issue for `<feature gate name>`"

body:
  - type: markdown 
    id: feature-gate
    attributes:
      value: "Feature gate: `#![feature(<feature gate name>)]`"

  - type: markdown
    id: description
    attributes:
      value: A concise description of the unstable feature.

 - type: checkboxes
    attributes:
      label: Type of unstable feature
      description: Select what kind of unstable feature this is.
      options:
      - label: New feature, this gives time to get additional feedback to apply changes. Having a transition period allows to make API changes without incrementing the major version.
      - label: Working feature that still has some unresolved questions that has a potential to change the API design.
      - label: Proof of concept features.
      required: true

  - type: checkboxes
    attributes:
      label: Unresolved Questions
      description: List every unresolved questions that need to be answered for the feature to be stable.
      options:
      - label: <Unresolved Questions>
      required: true
```

## Motivation

The motivation to have this feature is to be able to merge features that are mostly working but have some unresolved questions that could lead to an API change. Marking something unstable allows a user of the unstable feature to not rely heavily on that feature and participate in the stabilization of this one. Another pro of feature gating unstable features is that we wouldn't need to increment the version each time such a feature changes, but only when something considered stable changes. That would result in less version bumping.

## Goals

Making the API clearer by having a way of telling what is unstable and keeping versioning cleaner.

## Requirements

Easy to follow the state of the feature stabilization and enabling / disabling the feature should be easy.

## Unresolved Questions

- None for now.

## Prior Art (Existing PI C Implementation)

Doing a PR marked as breaking change and increment the version.

## Alternatives

- Not feature gating potentially unstable new features.
- Add a doc comment saying that the API could change.
- Having a nightly branch


docs/src/rfc/text/0007-unstable-feature.md
cfernald, line 1: It seems the actual RFC is empty.

magravel: I thought writing the RFC at the end in case it changes, is the RFC the same as the PR description?

Javagedes: [Introduction](https://sturdy-adventure-nv32gqw.pages.github.io/introduction.html#introduction)

Here are the docs for it. But yes, you create the RFC in the PR, and edit the markdown as you go while also updating the "Change Log" portion of the RFC.

Javagedes: nit: should be `0000` until ready to merge. Just in-case someone else comes in and finalizes a RFC before you :)

docs/src/rfc/text/0000-unstable-feature.md
makubacki, line 1: @magravel, can you please add an example dev flow to the RFC for what you are proposing? For example, walk through a developer adding an unstable feature to a crate in the project. What steps would they do? What is the purpose of each step?

docs/src/rfc/text/0000-unstable-feature.md
```diff
@@ -0,0 +1,78 @@
+# RFC: `Unstable Feature`
+
+This RFC is to add a way for feature gating new features that we consider unstable.
```
makubacki: Do you feel like we need more guidance about when something should be considered "unstable"? I'm a little concerned it will be easier to opt for something to be "unstable" than spend more time to create a stronger API up front and we end up with a lot of features and interfaces that no one can really depend on. Do you consider that to be an issue? If so, are there recommendations to better classify when the unstable approach should be used?

magravel: Yes, I think adding a list of requirements for unstability would be a good idea.

docs/src/rfc/text/0000-unstable-feature.md
```diff
@@ -0,0 +1,78 @@
...
+This would include an issue template and a way of feature gating the unstable feature.
+
+## Change Log
+
+- 2025-05-08: Initial RFC created.
+
+## Dev flow example
+
+- Add a feature to the cargo.toml named `unstable-<feature-name>
+- Use `#[cfg(feature = "unstable-<feature-name>")]` to feature gate the unstable code.
```
joschock: I don't know if there is an rust-ecosystem-standard naming convention for this kind of thing, but if there is, lets try and follow it.

magravel: There is no official documentation that covers that, but from what I saw looking at a couple of crates, it seems to have 2 main ways of dealing with that, simply incrementing the major version (often used with smaller crate), or using so sort of unstable feature / comment of some sort for bigger crates. This trend is most likely driven by the usage of unstable in the standard library, which makes every rust developer somewhat familiar with this concept.

docs/src/rfc/text/0000-unstable-feature.md
magravel: Use cases I see that would make sense to have unstable are:

- New feature, this gives time to get additional feedback to apply changes. Having a transition period allows to make API changes without incrementing the major version. *1
- Working feature that still has some unresolved questions that has a potential to change the API design.
- Proof of concept features.

*1: Not incrementing major version for feature stabilization has some benefit. First, not having to constantly bump the version during a stabilization phase, but also shows that something stable change and that the user of patina might be impacted, this makes breaking changes more meaningful and intentional.

docs/src/rfc/text/0000-unstable-feature.md
```diff
@@ -0,0 +1,78 @@
...
+```
+
+## Motivation
+
+The motivation to have this feature is to be able to merge features that are mostly working but have
+some unresolved questions that could lead to an API change. Marking something unstable allows a user
+of the unstable feature to not rely heavily on that feature and participate in the stabilization of
+this one. Another pro of feature gating unstable features is that we wouldn't need to increment the
+version each time such a feature changes, but only when something considered stable changes. That
+would result in less version bumping.
```
cfernald: I'm not sure where this belong, but one downside to a feature-based gating is that it can significantly complicate our testing matrix. Perhaps some thought should be put into how this should be handled and its implications, or even that quality/testing with _any_ unstable features may not be guaranteed.

joschock: >or even that quality/testing with _any_ unstable features may not be guaranteed.

That's the point, IMHO - we feature gate things where the design/implementation is not necessarily "stable". Customers who are very interested/engaged with a particular unstable feature will know the state of play, I think, but the gate serves as a warning to casual use of stuff that might not be fully baked/tested yet but that would benefit from visibility and feedback. As a corollary to that, I think that before the feature gate is removed on a feature that is ready to go stable one of the checkpoints is whether the testing/quality story is complete.

docs/src/rfc/text/0000-unstable-feature.md
makubacki: This seems reasonable to me as a starting point to see how it goes in practice.

docs/src/rfc/text/0000-unstable-feature.md
makubacki: I agree with this conceptually, but I think we'll have to keep an eye out to avoid things perpetually sitting in an "unstable" state because they worked well enough for their initial use case, so the author moved on before getting to a production quality testing story.

docs/src/rfc/text/0000-unstable-feature.md
```diff
@@ -0,0 +1,78 @@
...
+This RFC is to add a way for feature gating new features that we consider unstable.
+This would include an issue template and a way of feature gating the unstable feature.
+
+## Change Log
+
+- 2025-05-08: Initial RFC created.
+
+## Dev flow example
+
+- Add a feature to the cargo.toml named `unstable-<feature-name>
```
makubacki:
```suggestion
- Add a feature to the cargo.toml named `unstable-<feature-name>`
```

docs/src/rfc/text/0000-unstable-feature.md
joschock: heh. being unstable for millennia is the rust way though ;)

docs/src/rfc/text/0000-unstable-feature.md
makubacki: @magravel, can you please include this in the RFC and the proposed RFC GitHub issue template?

makubacki:
@magravel, when you're ready, please mention in the PR description that the RFC is moving to the "Final Comment Phase" so reviewers can perform a final review and give their approval.

PR #407, opened 2025-05-08
Add device path.

magravel:
## Description

This PR adds an abstraction over device path so we can do operations on them natively in rust. DevicePathBuf is an own version of the type and DevicePath is used as a borrow version of it. Similar to Path and PathBuf in std rust. The memory layout in the internal vector is the same as the one in C, so these structs can be easily converted and used from both languages.

This feature is unstable for now, enable feature `unstable-device-path` to use it:
https://github.com/OpenDevicePartnership/patina/issues/547

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [X] Includes tests?
- [X] Includes documentation?

## How This Was Tested

Converted a device path pointer from C and then printed it to see if the layout was read correctly.

## Integration Instructions

N/A

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,7 +4,7 @@ resolver = "2"
 members = ["components/*", "core/*", "sdk/*", "patina_dxe_core"]
 
 [workspace.package]
-version = "4.0.1"
+version = "4.0.2"
 license = "BSD-2-Clause-Patent"
 edition = "2021"
 rust-version = "1.85"
@@ -30,17 +30,17 @@ linkme = { version = "^0.3.29" }
 log = { version = "0.4", default-features = false }
 mu_pi = { version = "5.2.3" }
 mu_rust_helpers = { version = "3.0.1" }
-patina_debugger = { version = "4.0.1", path = "core/patina_debugger", registry = "patina-fw" }
-patina_internal_collections = { version = "4.0.1", path = "core/patina_internal_collections", default-features = false, registry = "patina-fw" }
-patina_internal_cpu = { version = "4.0.1", path = "core/patina_internal_cpu", registry = "patina-fw" }
-patina_internal_depex = { version = "4.0.1", path = "core/patina_internal_depex", registry = "patina-fw" }
-patina_internal_device_path = { version = "4.0.1", path = "core/patina_internal_device_path", registry = "patina-fw" }
+patina_debugger = { version = "4.0.2", path = "core/patina_debugger", registry = "patina-fw" }
+patina_internal_collections = { version = "4.0.2", path = "core/patina_internal_collections", default-features = false, registry = "patina-fw" }
+patina_internal_cpu = { version = "4.0.2", path = "core/patina_internal_cpu", registry = "patina-fw" }
+patina_internal_depex = { version = "4.0.2", path = "core/patina_internal_depex", registry = "patina-fw" }
+patina_internal_device_path = { version = "4.0.2", path = "core/patina_internal_device_path", registry = "patina-fw" }
 patina_mtrr = { version = "1.0.0", registry = "patina-fw" }
 patina_paging = { version = "8", registry = "patina-fw" }
-patina_performance = { version = "4.0.1", path = "components/patina_performance", registry = "patina-fw" }
-patina_sdk = { version = "4.0.1", path = "sdk/patina_sdk", registry = "patina-fw" }
-patina_sdk_macro = { version = "4.0.1", path = "sdk/patina_sdk_macro", registry = "patina-fw" }
-patina_stacktrace = { version = "4.0.1", path = "core/patina_stacktrace", registry = "patina-fw" }
+patina_performance = { version = "4.0.2", path = "components/patina_performance", registry = "patina-fw" }
+patina_sdk = { version = "4.0.2", path = "sdk/patina_sdk", registry = "patina-fw" }
+patina_sdk_macro = { version = "4.0.2", path = "sdk/patina_sdk_macro", registry = "patina-fw" }
+patina_stacktrace = { version = "4.0.2", path = "core/patina_stacktrace", registry = "patina-fw" }
 proc-macro2 = { version = "1" }
 quote = { version = "1" }
 r-efi = { version = "5.0.0", default-features = false }


sdk/patina_sdk/src/uefi_protocol/device_path/device_path_node.rs
```diff
@@ -0,0 +1,248 @@
...
+    fn header(&self) -> Header;
+
+    /// Return true if this device path node has the same type and sub_type/
+    fn is_type(r#type: u8, sub_type: u8) -> bool
+    where
+        Self: Sized;
+
+    /// Write the device path node into that buffer and return the number of byte written.
+    #[allow(clippy::result_unit_err)]
+    fn write_into(self, buffer: &mut [u8]) -> Result<usize, ()>;
```
joschock: maybe a clarifying comment on why unit_err is okay in this case?

sdk/patina_sdk/src/uefi_protocol/device_path/device_path_node.rs
```diff
@@ -0,0 +1,248 @@
...
+    }
+}
+
+impl DevicePathNode for UnknownDevicePathNode<'_> {
+    fn header(&self) -> Header {
+        self.header
+    }
+
+    fn is_type(_type: u8, _sub_type: u8) -> bool {
+        true
```
joschock, line 108: maybe a comment on why this is always true and does not need to check the type/sub_type?

sdk/patina_sdk/src/uefi_protocol/device_path/device_path_node.rs
```diff
@@ -0,0 +1,248 @@
...
+        f.write_fmt(format_args!("Path({}, {},", &self.header.r#type, &self.header.sub_type))?;
+        for b in self.data {
+            f.write_fmt(format_args!(" {:02X}", b))?;
+        }
+        f.write_char(')')
+    }
+}
+
+#[macro_export]
+macro_rules! device_path_node {
```
joschock, line 160: are there doc comments somewhere with examples of these macros? If not, consider adding some here to help understand the macros and usage better.

sdk/patina_sdk/src/uefi_protocol/device_path.rs
```diff
@@ -0,0 +1,532 @@
...
+}
+
+impl DevicePathBuf {
+    /// Create a DevicePathBuf with an empty buffer.
+    fn new_empty() -> Self {
+        Self { buffer: Vec::new() }
+    }
+
+    /// Append a node to the device path.
+    /// This function do not ensure that the device path is valid, the EndEntire node must be manually added to the device path.
```
joschock:
```suggestion
    /// This function does not ensure that the device path is valid, the EndEntire node must be manually added to the device path.
```

sdk/patina_sdk/src/uefi_protocol/device_path.rs
```diff
@@ -0,0 +1,532 @@
...
+/// Only immutable opperation are possible on this type.
+#[derive(Debug)]
+#[repr(transparent)]
+pub struct DevicePath {
+    buffer: [u8],
+}
+
+impl DevicePath {
+    /// Create a &DevicePath for a DevicePathBuf.
+    pub fn from(device_path_buff: &DevicePathBuf) -> &Self {
```
joschock, line 155: nitpick: naming consistency
```suggestion
    pub fn from(device_path_buf: &DevicePathBuf) -> &Self {
```

sdk/patina_sdk/src/uefi_protocol/device_path.rs
```diff
@@ -0,0 +1,532 @@
...
+                break;
+            }
+
+            let new_length = buffer.len() + header.length;
+            offset += header.length;
+            buffer = core::slice::from_raw_parts(buffer.as_ptr(), new_length);
+        }
+
+        let device_path = &*(buffer as *const [u8] as *const DevicePath);
+        device_path.is_device_path_valid()?;
```
joschock: checking my understanding: `is_device_path_valid()` will always succeed here, since above code loops infinitely if no EndEntire is found, right?

sdk/patina_sdk/src/uefi_protocol/device_path.rs
```diff
@@ -0,0 +1,532 @@
...
+                    return Err("Device path has an end entire in the middle, everything after will be ignored.");
+                } else {
+                    return Ok(());
+                }
+            }
+        }
+        Err("Device path need to end with an end entire node.")
+    }
+
+    /// Return the size in byte of the device path.
```
joschock:
```suggestion
    /// Return the size in bytes of the device path.
```

sdk/patina_sdk/src/uefi_protocol/device_path.rs
```diff
@@ -0,0 +1,532 @@
...
+    pub fn size(&self) -> usize {
+        self.buffer.len()
+    }
+
+    /// Return the number of node in the device path.
```
joschock:
```suggestion
    /// Return the number of nodes in the device path.
```

sdk/patina_sdk/src/uefi_protocol/device_path.rs
```diff
@@ -0,0 +1,532 @@
...
+    pub fn node_count(&self) -> usize {
+        self.iter().count()
+    }
+
+    /// Return true if the device path contains more than one instance, otherwise false.
+    pub fn is_multi_instances(&self) -> bool {
```
joschock:
```suggestion
    pub fn is_multi_instance(&self) -> bool {
```

sdk/patina_sdk/src/uefi_protocol/device_path.rs
```diff
@@ -0,0 +1,532 @@
...
+        self.iter().any(|n| EndInstance::is_type(n.header.r#type, n.header.sub_type))
+    }
+
+    /// Return a &DevicePath for the n last node of the device path.
+    /// This operation does not copy memory since the trailing end of a device path is a valid device path.
+    pub fn slice_end(&self, n: usize) -> &DevicePath {
```
joschock, line 209: consider adding a comment that this will panic if n > self.node_count()

sdk/patina_sdk/src/uefi_protocol/device_path.rs
```diff
@@ -0,0 +1,532 @@
...
+        for _ in 0..nb_skip {
+            let header = self.buffer.pread_with::<Header>(idx, scroll::LE).unwrap();
+            idx += header.length;
+        }
+        let end_buffer = &self.buffer[idx..];
+        unsafe { &*(end_buffer as *const [u8] as *const DevicePath) }
+    }
+
+    /// Return true if the device path start with an other device path.
+    pub fn start_with(&self, other: &DevicePath) -> bool {
```
joschock:
```suggestion
    /// Return true if the device path starts with the other device path.
    pub fn starts_with(&self, other: &DevicePath) -> bool {
```

sdk/patina_sdk/src/uefi_protocol/device_path.rs
```diff
@@ -0,0 +1,532 @@
...
+    ///
+    /// # Safety
+    ///
+    /// the buffer pointer must point to valid device path data.
+    pub unsafe fn try_from_ptr<'a>(buffer: *const u8) -> Result<&'a DevicePath, &'static str> {
+        let mut buffer = core::slice::from_raw_parts(buffer, Header::size_of_header());
+        let mut offset = 0;
+        loop {
+            let header =
+                buffer.pread_with::<Header>(offset, scroll::LE).map_err(|_| "Error while trying to read header.")?;
```
Javagedes, line 177: I am against introducing scroll to this crate. It seems like unnecessary code size bloat. 

I'm not seeing any complex usage for scroll. Everything I see implementing Pread can also be a simple cast + Clone of the type, such as `Header`, `Pci`, `PcCard`, `Controller`, `BMC`, `Acpi`

I see slightly complex usage with DevicePathNode / UnknownDevicePathNode, But I'm not sure it warrants the size increase associated with scroll, but they are still simple enough that you could do the casting yourself.

Does anyone else have a strong opinion on this, or is it just me? I feel like it is worth investigating the size bloat from using scroll here.

joschock: Agree if it actually changes the code size materially, it's worth considering whether we are better off without it. But I'll provide some counter points for the sake of deliberation because I do see some advantages to the `scroll` approach. 

1: Explicit serialization as is done with scroll i.e. 'these bytes go to that field via this explicit transformation' is safer and more obvious than overlaying a struct on memory with a cast. Explicit conversions with `scroll` are potentially less performant (maybe? - subject to empirical confirmation). But I'm not sure that we do these kinds of transformations enough or in performance-sensitive places enough for this to matter. 

2: It also allows for the Rust structures to be free of FFI constraints. For example, the present definition of `Header` in this code uses a `length: usize` but if we moved to casting, that struct would need to become `#[repr(C)]` and change to `length: u16` - which would mean `as usize` everywhere else it is used.

sdk/patina_sdk/src/uefi_protocol/device_path.rs
magravel: Yes, I don't think this is useful now.

sdk/patina_sdk/src/uefi_protocol/device_path.rs
magravel: For right now, most type could be written as a pointer write, but some node will have string and buffer, which will be better to be abstract with scroll.

sdk/patina_sdk/src/uefi_protocol/device_path.rs
```diff
@@ -0,0 +1,529 @@
+//! A Device Path is used to define the programmatic path to a device.
+//! The primary purpose of a Device Path is to allow an application, such as an OS loader, to determine the physical device that the interfaces are abstracting.
```
makubacki:
```suggestion
//! A Device Path is used to define the programmatic path to a device.
//!
//! The primary purpose of a Device Path is to allow an application, such as an OS loader, to determine the physical device that the interfaces are abstracting.
//!
//! ## License
//!
//! Copyright (C) Microsoft Corporation. All rights reserved.
//!
//! SPDX-License-Identifier: BSD-2-Clause-Patent
//!
```

sdk/patina_sdk/src/uefi_protocol/device_path.rs
```diff
@@ -0,0 +1,529 @@
...
+    ops::Deref,
+};
+
+use log;
+use scroll::Pread;
+
+use device_path_node::{DevicePathNode, Header, UnknownDevicePathNode};
+use nodes::{DevicePathType, EndEntire, EndInstance};
+
+/// DevicePathBuf is an owned version of device path. This is used to create device paths or when doing mutable operation on them.
```
makubacki:
```suggestion
/// DevicePathBuf is an owned version of device path. This is used to create device paths or when performing mutable operations on them.
```

sdk/patina_sdk/src/uefi_protocol/device_path.rs
```diff
@@ -0,0 +1,529 @@
...
+        // SAFETY: This is safe because DevicePath have the same memory layout as `[u8]`.
+        unsafe { &*(device_path_buff.buffer.as_slice() as *const [u8] as *const Self) }
+    }
+
+    /// Create a &DevicePath from a pointer to a byte buffer.
+    /// This is used to use device path from C.
+    ///
+    /// # Safety
+    ///
+    /// the buffer pointer must point to valid device path data.
```
makubacki:
```suggestion
    /// Create a &DevicePath from a pointer to a byte buffer.
    /// This is used to interface with device paths from C code.
    ///
    /// # Safety
    ///
    /// The buffer pointer must point to valid device path data that remains valid
    /// for the lifetime 'a. The device path must be properly terminated with an
    /// EndEntire node.
```

sdk/patina_sdk/src/uefi_protocol/device_path.rs
```diff
@@ -0,0 +1,529 @@
...
+    pub unsafe fn try_from_ptr<'a>(buffer: *const u8) -> Result<&'a DevicePath, &'static str> {
+        let mut buffer = core::slice::from_raw_parts(buffer, Header::size_of_header());
```
makubacki: It seems worthwhile to just explicitly check if the given raw pointer is null before doing anything else.

```suggestion
    pub unsafe fn try_from_ptr<'a>(buffer: *const u8) -> Result<&'a DevicePath, &'static str> {
        if buffer.is_null() {
            return Err("Null pointer provided");
        }

        let mut buffer = core::slice::from_raw_parts(buffer, Header::size_of_header());
```

sdk/patina_sdk/src/uefi_protocol/device_path.rs
```diff
@@ -0,0 +1,529 @@
...
+
+impl DevicePathBuf {
+    /// Create a DevicePathBuf with an empty buffer.
+    fn new_empty() -> Self {
+        Self { buffer: Vec::new() }
+    }
+
+    /// Append a node to the device path.
+    /// This function does not ensure that the device path is valid, the EndEntire node must be manually added to the device path.
+    pub fn append<T>(&mut self, node: T)
```
makubacki, line 47: I think this should return a `Result` and return an `Err` from the `node.write_into()` call instead of just printing an error message. It could be `Result<(), &'static str>` or something so the caller can access the error message if needed.

sdk/patina_sdk/src/uefi_protocol/device_path.rs
```diff
@@ -0,0 +1,529 @@
...
+    where
+        T: DevicePathNode + Sized,
+    {
+        let writing_start = self.buffer.len();
+        let header = node.header();
+        self.buffer.resize(writing_start + header.length, 0);
```
makubacki: Probably best to check for a potential overflow explicitly before attempting to resize.

```suggestion
        let new_size = writing_start.checked_add(header.length)
            .ok_or("Device path exceeds maximum allowed size")?;
            
        self.buffer.resize(new_size, 0);
```

sdk/patina_sdk/src/uefi_protocol/device_path.rs
```diff
@@ -0,0 +1,529 @@
...
+    }
+
+    /// Append a device path to this device path, the EndEntire node of self will be replaced with an end instance.
+    pub fn append_device_path_instances(&mut self, device_path: &DevicePath) {
+        self.buffer.truncate(self.buffer.len() - EndEntire.header().length);
+        self.append(EndInstance);
+        self.buffer.extend_from_slice(&device_path.buffer[..]);
+    }
+
+    /// Create a device path form an iterator of device path nodes.
```
makubacki:
```suggestion
    /// Create a device path from an iterator of device path nodes.
```

sdk/patina_sdk/src/uefi_protocol/device_path.rs
```diff
@@ -0,0 +1,529 @@
...
+    /// If the iterator does not end with an `EndEntire` node, it will be added to the device path.
+    pub fn from_device_path_node_iter<I, T>(iter: I) -> DevicePathBuf
```
makubacki, line 86: If `append()` is updated to return `Result`, then can return `Result<DevicePathBuf, &'static str>` here and call `device_path.append(n)` in this function instead with `device_path.append(n)?`.

sdk/patina_sdk/src/uefi_protocol/device_path.rs
```diff
@@ -0,0 +1,529 @@
...
+impl PartialEq for DevicePathBuf {
+    fn eq(&self, other: &DevicePathBuf) -> bool {
+        DevicePath::from(self) == DevicePath::from(other)
+    }
+}
+
+impl Eq for DevicePathBuf {}
+
+/// DevicePath is the borrowed version of a [`DevicePathBuf`].
+/// Only immutable opperation are possible on this type.
```
makubacki:
```suggestion
/// Only immutable operations are possible on this type.
```

sdk/patina_sdk/src/uefi_protocol/device_path.rs
```diff
@@ -0,0 +1,529 @@
...
+    pub fn node_count(&self) -> usize {
+        self.iter().count()
+    }
+
+    /// Return true if the device path contains more than one instance, otherwise false.
+    pub fn is_multi_instance(&self) -> bool {
+        self.iter().any(|n| EndInstance::is_type(n.header.r#type, n.header.sub_type))
+    }
+
+    /// Return a &DevicePath for the n last node of the device path.
```
makubacki:
```suggestion
    /// Return a &DevicePath for the n last nodes of the device path.
```

sdk/patina_sdk/src/uefi_protocol/device_path.rs
```diff
@@ -0,0 +1,529 @@
...
+    /// This operation does not copy memory since the trailing end of a device path is a valid device path.
+    pub fn slice_end(&self, n: usize) -> &DevicePath {
+        let count = self.node_count();
+
+        debug_assert!(n > 1, "Device path need to at least have the end node.");
```
makubacki: I assume one is allowed?

```suggestion
        debug_assert!(n >= 1, "Device path needs to have at least the end node.");
```

sdk/patina_sdk/src/uefi_protocol/device_path.rs
```diff
@@ -0,0 +1,529 @@
...
+        debug_assert!(n <= count, "Can't retrun a device path bigger than self.");
```
makubacki:
```suggestion
        debug_assert!(n <= count, "Cannot return a device path bigger than self.");
```

sdk/patina_sdk/src/uefi_protocol/device_path.rs
```diff
@@ -0,0 +1,529 @@
...
+    fn next(&mut self) -> Option<Self::Item> {
+        // No more byte to read in the device path.
+        if self.offset == self.device_path.buffer.len() {
+            return None;
+        }
+
+        let Ok(unknown_device_path) =
+            self.device_path.buffer.gread_with::<UnknownDevicePathNode>(&mut self.offset, scroll::LE)
+        else {
+            debug_assert!(false, "The buffer is corrupted, could not read the device path.");
```
makubacki:
```suggestion
            debug_assert!(false, "The buffer is corrupted, could not read the device path node.");
```

sdk/patina_sdk/src/uefi_protocol/device_path/device_path_node.rs
```diff
@@ -0,0 +1,262 @@
...
+use super::nodes;
+
+/// Common header of device path nodes.
+#[derive(Debug, Clone, Copy, Eq, PartialEq)]
+pub struct Header {
+    /// Type of the device path node.
+    pub r#type: u8,
+    /// Subtype of the device path node.
+    pub sub_type: u8,
+    /// Total length in byte of the device path node, including the header.
```
makubacki:
```suggestion
    /// Total length in bytes of the device path node, including the header.
```

sdk/patina_sdk/src/uefi_protocol/device_path/device_path_node.rs
```diff
@@ -0,0 +1,262 @@
...
+    pub length: usize,
+}
+
+impl Header {
```
makubacki, line 30: I think a `new()` function would make sense.

sdk/patina_sdk/src/uefi_protocol/device_path/device_path_node.rs
```diff
@@ -0,0 +1,262 @@
...
+        ))
+    }
+}
+
+/// Trait that every device path node must implement.
+pub trait DevicePathNode: Debug + Display {
+    /// Return the header of the device path node.
+    fn header(&self) -> Header;
+
+    /// Return true if this device path node has the same type and sub_type/
```
makubacki:
```suggestion
    /// Return true if this device path node has the same type and sub_type.
```

sdk/patina_sdk/src/uefi_protocol/device_path/device_path_node.rs
```diff
@@ -0,0 +1,262 @@
...
+    fn is_type(r#type: u8, sub_type: u8) -> bool
+    where
+        Self: Sized;
+
+    /// Write the device path node into that buffer and return the number of byte written.
```
makubacki:
```suggestion
    /// Write the device path node into the buffer and return the number of bytes written.
```

sdk/patina_sdk/src/uefi_protocol/device_path/device_path_node.rs
```diff
@@ -0,0 +1,262 @@
...
+    fn write_into(self, buffer: &mut [u8]) -> Result<usize, scroll::Error>;
+}
+
+/// UnknownDevicePathNode are device path nodes that have not been cast to a more specific associated type or that are undefined in the spec.
```
makubacki:
```suggestion
/// `UnknownDevicePathNode` represents device path nodes that have not been cast to a more specific associated type
/// or that are undefined in the spec.
```

sdk/patina_sdk/src/uefi_protocol/device_path/device_path_node.rs
```diff
@@ -0,0 +1,262 @@
...
+        f.write_char(')')
+    }
+}
+
+/// This macro is used to defined a device path node struct. And implement device path node traits for it.
+/// To configure the type and subtype of this node, use a attribute `@[DevicePathNode(DevicePathType::Type, SubType::MySubtype)]`
+/// where the Type and subtype are enum path where the value can be expess as u8.a
+///
+/// Some additional trait can be implemented with `@[DevicePathNodeDerive(...)]`
+/// Currently supported traits are: Debug and Display.
```
makubacki:
```suggestion
/// This macro is used to define a device path node struct and implement device path node traits for it.
/// To configure the type and subtype of this node, use an attribute `@[DevicePathNode(DevicePathType::Type, SubType::MySubtype)]`
/// where the Type and subtype are enum paths where the value can be expressed as u8.a
///
/// Some additional traits can be implemented with `@[DevicePathNodeDerive(...)]`
/// Currently supported traits are: Debug and Display.
```

sdk/patina_sdk/src/uefi_protocol/device_path/device_path_node.rs
```diff
@@ -0,0 +1,262 @@
...
+                $(
+                    .field(&self.$field_name)
+                )*
+                    .finish()
+            }
+        }
+    };
+    // Internal matching to add a complite error in case of an unsupported trait in derive.
+    (@Derive; $struct_name:ident, $($field_name:ident),*; $trait:ident) => {
+        compile_error!("A trait is unsupported derive trait.");
```
makubacki:
```suggestion
        compile_error!("An unsupported derive trait was specified.");
```

sdk/patina_sdk/src/uefi_protocol/device_path/device_path_node.rs
```diff
@@ -0,0 +1,262 @@
...
+            fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+                f.debug_tuple(stringify!($struct_name))
+                $(
+                    .field(&self.$field_name)
+                )*
+                    .finish()
+            }
+        }
+    };
+    // Internal matching to add a complite error in case of an unsupported trait in derive.
```
makubacki: Was this meant?

```suggestion
    // Internal matching to add a compilation error in case of an unsupported derive trait.
```

sdk/patina_sdk/src/uefi_protocol/device_path/nodes.rs
```diff
@@ -0,0 +1,369 @@
...
+
+use super::device_path_node::{DevicePathNode, UnknownDevicePathNode};
+
+use crate::device_path_node;
+
+#[derive(Debug, Eq, PartialEq, Clone, Copy)]
+#[repr(u8)]
+pub enum DevicePathType {
+    Hardware = 1,
+    ACPI = 2,
```
makubacki: I'd personally prefer to strictly stick with Pascal case in these variants (so, "`Acpi`" in this case). But will leave it up to you.

sdk/patina_sdk/src/uefi_protocol/device_path.rs
magravel: writing_start in relative to the start of the buffer. Renamed the variable to make this more clear that is it not an address.

sdk/patina_sdk/src/uefi_protocol/device_path.rs
makubacki: We discussed a different approach offline.

sdk/patina_sdk/src/uefi_protocol/device_path.rs
makubacki: Closed in offline discussion with another approach in `append()`.

sdk/patina_sdk/src/uefi_protocol/device_path.rs
magravel: Here, the only way something would fail, is by the unlikelihood of an allocation error. To mitigate the risk of panic when an allocation error occurs in vec, the use of `try_reserve_exact` is used. In that case, I can generate the allocation in a controlled way and prevent the device from crashing by handling the failing use case.

PR #408, opened 2025-05-09
RFC: Categorizing and Renaming Patina Crates for Consistency

vineelko:
## Description

This RFC proposes a classification of Patina crates into public and internal categories, and renames them to follow a consistent naming convention.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

NA

## Integration Instructions

NA


docs/src/rfc/text/0000-patina-crates-naming-and-categorization.md
```diff
@@ -0,0 +1,123 @@
...
+│   ├── section_extractor        │   ├── patina_internal_section_extractor
+│   ├── stacktrace               │   ├── patina_internal_stacktrace
+│   ├── uefi_collections         │   ├── patina_internal_collections
+│   ├── uefi_cpu                 │   ├── patina_internal_cpu
+│   ├── uefi_debugger            │   ├── patina_internal_debugger
+│   ├── uefi_depex               │   ├── patina_internal_depex
+│   ├── uefi_device_path         │   ├── patina_internal_device_path
+│   └── uefi_performance         │   └── patina_internal_performance
+└── sdk                          └── sdk
+    ├── boot_services                ├── patina_internal_boot_services
```
cfernald: I would argue that if there is anything "internal" in the SDK then it shouldn't be in the SDK, right?

makubacki: I'd agree.

makubacki: I think there is also an element of whether we are confident enough in the crate as-is for it to be advertised as a public dependency. It's easier to make something public later (rename the crate to drop "internal") than commit to its public interface now.

@vineelko, was that what you had in mind for some of these? Or something else?

docs/src/rfc/text/0000-patina-crates-naming-and-categorization.md
```diff
@@ -0,0 +1,123 @@
...
+The identified public crates from the dependency graph are shown below, along
+with the proposed crate names. Additional input is needed to finalize this
+classification.
+
+![Patina public crates](0000-patina-crates-naming-and-categorization/patina_public_crates.png)
+
+```text
+  Existing                             New
+├── components                   ├── components
+│   ├── adv_logger               │   ├── patina_internal_adv_logger
```
joschock: adv_logger is instantiated by the consumers of the core to instantiate the core (e.g. qemu bins) - so it should be external?

docs/src/rfc/text/0000-patina-crates-naming-and-categorization.md
```diff
@@ -0,0 +1,123 @@
...
+│   └── sample_components        │   └── patina_internal_sample_components
+├── core                         ├── core
+│   ├── dxe_core                 │   ├── patina_dxe_core
+│   ├── section_extractor        │   ├── patina_internal_section_extractor
```
joschock: section_extractor is also instantiated by the consumers of dxe_core to configure what sections are supported.

docs/src/rfc/text/0000-patina-crates-naming-and-categorization.md
```diff
@@ -0,0 +1,123 @@
...
+│   ├── stacktrace               │   ├── patina_internal_stacktrace
```
joschock: I'm using stack trace external to the core in my project's panic handler, so would like it to be external as well.

docs/src/rfc/text/0000-patina-crates-naming-and-categorization.md
```diff
@@ -0,0 +1,123 @@
...
+│   ├── uefi_collections         │   ├── patina_internal_collections
+│   ├── uefi_cpu                 │   ├── patina_internal_cpu
+│   ├── uefi_debugger            │   ├── patina_internal_debugger
```
joschock: debugger is used by qemu bins to allow for breakpoint on panic.

docs/src/rfc/text/0000-patina-crates-naming-and-categorization.md
joschock: they could also be re-exports? i.e. that way you only have to list patina_sdk as your dependency, and then you would do like patina_sdk::tpl_mutex

docs/src/rfc/text/0000-patina-crates-naming-and-categorization.md
cfernald: I think all components should be external, including adv_logger

docs/src/rfc/text/0000-patina-crates-naming-and-categorization.md
```diff
@@ -0,0 +1,123 @@
...
+with the proposed crate names. Additional input is needed to finalize this
+classification.
+
+![Patina public crates](0000-patina-crates-naming-and-categorization/patina_public_crates.png)
+
+```text
+  Existing                             New
+├── components                   ├── components
+│   ├── adv_logger               │   ├── patina_internal_adv_logger
+│   └── sample_components        │   └── patina_internal_sample_components
```
cfernald: @Javagedes, perhaps we should just flatten sample components into the sdk. What do you think?

docs/src/rfc/text/0000-patina-crates-naming-and-categorization.md
cfernald: Agreed. Debugger needs to be external to provide those functions

docs/src/rfc/text/0000-patina-crates-naming-and-categorization.md
cfernald: I think at that point I would just flatten these into some crates serving specific purposes. e.g. `patina_sdk`, `patina_uefi`, `patina_test`. Now the macros i know have some limitations and need to be independent so maybe those are the few that internal & re-export makes sense

docs/src/rfc/text/0000-patina-crates-naming-and-categorization.md
```diff
@@ -0,0 +1,123 @@
...
+    └── uefi_test_macro
+```
+
+### New Crate Organization
+
+The identified public crates from the dependency graph are shown below, along
+with the proposed crate names. Additional input is needed to finalize this
+classification.
+
+![Patina public crates](0000-patina-crates-naming-and-categorization/patina_public_crates.png)
```
cfernald: Might be worth capturing this from a q35/sbsa build, as that may more accurately reflect what is "internal"

docs/src/rfc/text/0000-patina-crates-naming-and-categorization.md
vineelko: image

I don't recall exactly when, but I believe this came up during one of our Rust sync-up meetings(separate from the [RFC: Single Repo](https://github.com/OpenDevicePartnership/patina/pull/380#discussion_r2069146345) conversation). I had proposed the idea of making all of them modules instead of separate crates, since they seemed too small and could be grouped under a single `uefi_sdk` umbrella crate. However, I believe the feedback I received at the time was that users are not expected to use these highlighted crates directly, based on the documentation. Therefore, I thought we could mark them as internal in their names and ensure they are used only by `uefi_sdk`. But if these crates are meant for public consumption, then I’m totally fine with marking them as public (non-internal).

docs/src/rfc/text/0000-patina-crates-naming-and-categorization.md
Javagedes: Agree

docs/src/rfc/text/0000-patina-crates-naming-and-categorization.md
Javagedes: I'm OK with that. As a crate or as a module? If a module, where?

It does not matter much to me, so long as I have a place to point people to. I plan on adding more samples now that we have more uses and as we develop standards. So I hope to have more than just a few simple hello world examples.

docs/src/rfc/text/0000-patina-crates-naming-and-categorization.md
Javagedes: Not 100% associated with this PR, but I plan on putting up an RFC to move uefi_test into the uefi_sdk crate as a `test` module, and `uefi_test_macro` into `uefi_sdk_macro`. If that influences your decisions here regarding uefi_test.

___________________

I'm still curious (And maybe John or Mathieu have answered, and I've missed it) why we don't collapse these extra crates (runtime services, uefi protocol, etc) into uefi_sdk. The `pub use boot_services;` bcomes `pub mod boot_services`. I'm open to being told no. I've just missed the reasoning for not doing it.

docs/src/rfc/text/0000-patina-crates-naming-and-categorization.md
```diff
@@ -0,0 +1,123 @@
...
+
+```text
+  Existing                             New
```
Javagedes: I would like to start this chat generically for feedback for moving the "main" crate (which I will call the `dxe_core`
) crate to the top level of the repository. Something like:

```md
├── dxe_core (crate)
├── core/*
├── sdk/*
├── components/*
```

It's pretty common for large rust repositories to do it this way. Just proposing the question / possibility. I don't care that much

docs/src/rfc/text/0000-patina-crates-naming-and-categorization.md
```diff
@@ -0,0 +1,123 @@
+# RFC: Categorizing and Renaming Patina Crates for Consistency
```
Javagedes: If we are not going to rename `dxe_core` to `patina`, I think at a minimum we need to create a patina crate to reserve the name in crates.io. Maybe make it a re-export of `dxe_core` + `uefi_sdk` + other common ones. Or maybe just throw our docs in it. Or something. But we really should reserve that crate name.

vineelko: "3. In addition to the above guidelines, any inconsistent crate names should be renamed to improve clarity" 

I am open to that change(`dxe_core` to `patina`).

docs/src/rfc/text/0000-patina-crates-naming-and-categorization.md
makubacki: If we have a `/components` directory, can we just add `/components/samples` and put them there?

docs/src/rfc/text/0000-patina-crates-naming-and-categorization.md
makubacki: I'm fine with moving it to the root level since it is a primary integration point.

docs/src/rfc/text/0000-patina-crates-naming-and-categorization.md
makubacki: Technically, someone reserved "patina" in 2014 - https://crates.io/crates/patina. We should reserve decided upon crate names soon with minimal content. The "patina" ADO org (dev.azure.com/patina) was already taken as well.

Javagedes: unfortunate.

docs/src/rfc/text/0000-patina-crates-naming-and-categorization.md
makubacki: I also think it makes sense to collapse them into the sdk crate into an organized module hierarchy.

docs/src/rfc/text/0000-patina-crates-naming-and-categorization.md
makubacki: Yeah. The crates used in the Intel binary have already been mentioned - adv_logger, section_extractor, stacktrace, uefi_debugger.

docs/src/rfc/text/0000-patina-crates-naming-and-categorization.md
joschock: I don't remember any reason why `boot_services` can't be part of `uefi_sdk`. I don't remember having a conversation about that, but my memory is probably faulty. The only reason I can think of that they are separate is historical; `boot_services` predates `uefi_sdk`. If I had a concern it was probably around separation of concerns/coupling/interface definition; but I feel like the `uefi_sdk` crate being separate from the core implementation addresses that.

docs/src/rfc/text/0000-patina-crates-naming-and-categorization.md
cfernald: I was thinking as a module. It makes sense to me to have the samples co-located with the interface definitions. `/components/samples` makes sense to me

docs/src/rfc/text/0000-patina-crates-naming-and-categorization.md
cfernald: I agree this is more clear.

docs/src/rfc/text/0000-patina-crates-naming-and-categorization.md
cfernald: > Technically, someone reserved "patina" in 2014 - https://crates.io/crates/patina. We should reserve decided upon crate names soon with minimal content. The "patina" ADO org (dev.azure.com/patina) was already taken as well.

@makubacki  I'm curious, how do you know it's been reserved? The link just shows 404 for me and searching it yields no info.

Javagedes: Maybe Michael did something? that crate was there last week :)

docs/src/rfc/text/0000-patina-crates-naming-and-categorization.md
```diff
@@ -0,0 +1,123 @@
...
+```text
+  Existing                             New
+├── components                   ├── components
+│   ├── adv_logger               │   ├── patina_internal_adv_logger
+│   └── sample_components        │   └── patina_internal_sample_components
+├── core                         ├── core
+│   ├── dxe_core                 │   ├── patina_dxe_core
+│   ├── section_extractor        │   ├── patina_internal_section_extractor
+│   ├── stacktrace               │   ├── patina_internal_stacktrace
+│   ├── uefi_collections         │   ├── patina_internal_collections
```
os-d: Do we see any value of using this outside of patina? I.e. might a component use these collections or are they only valid for the core to use? @Javagedes ?

Javagedes: The intention was for them to be used externally by component writers, but I don't see a use-case for them anymore.

docs/src/rfc/text/0000-patina-crates-naming-and-categorization.md
makubacki: I did not. But we can jump on the crate name after publishing the mtrr and paging crates. The escalation path to report name squatting is clear, I just don't want it to be used against us if our only "patina" crate is name squatting.

vineelko:
Thanks for the feedback. I’m going to update the RFC with the suggested changes. One question I have is about the naming format of the crate directories (not the crate names themselves): should we use `patina-xxxx` or `patina_xxxx`?

Changelog for 2nd iteration:
1. All crates inside `uefi_sdk` are intended for public consumption and should be marked accordingly.
2. There is consensus on merging some crates with in sdk directory (e.g., runtime services, UEFI protocol) into `uefi_sdk`, organized by module hierarchy.
3. Move `dxe-core` to the root of the repository as `patina-dxe-core.
4. Rename `dxe-core` to `patina` once the name is available on crates.io.
5. All components should be external crates.
6. The following crates should be made external:
   - `section_extractor`
   - `stacktrace`
   - `debugger`
7. Rename `patina_internal_sample_components` to `/components/patina_samples`.

vineelko:
> Thanks for the feedback. I’m going to update the RFC with the suggested changes. One question I have is about the naming format of the crate directories (not the crate names themselves): should we use `patina-xxxx` or `patina_xxxx`?
> 
> Changelog for 2nd iteration:
> 
> 1. All crates inside `uefi_sdk` are intended for public consumption and should be marked accordingly.
> 2. There is consensus on merging some crates with in sdk directory (e.g., runtime services, UEFI protocol) into `uefi_sdk`, organized by module hierarchy.
> 3. Move `dxe-core` to the root of the repository as `patina-dxe-core.
> 4. Rename `dxe-core` to `patina` once the name is available on crates.io.
> 5. All components should be external crates.
> 6. The following crates should be made external:
>    
>    * `section_extractor`
>    * `stacktrace`
>    * `debugger`
> 7. Rename `patina_internal_sample_components` to `/components/patina_samples`.

With the help from copilot and friends. I am settling on `patina_xxxx` for directory names. 

In the Rust community, the **general preference** for crate **directory names** (and crate names too, when applicable) is to use **underscores** (`patina_xxxx`) rather than hyphens (`patina-xxxx`). Here's why:

---

### ✅ **Underscore (`patina_xxxx`) is preferred for crate directory names**

* It aligns with **Rust's naming conventions** for **identifiers** (like variables and module names), which use `snake_case`.
* Using underscores avoids issues with generated module names or file paths in some tooling or macro contexts.

docs/src/rfc/text/0000-patina-crates-naming-and-categorization.md
vineelko: Resolving this issue since we have a general consensus on merging some crates with in sdk directory (e.g., runtime services, UEFI protocol) into `uefi_sdk`, organized by module hierarchy. But as it is outside the scope of this RFC, Below issue is filed to track it separately.

https://github.com/OpenDevicePartnership/patina/issues/418

Added the same to the changelog for iteration 2 of this RFC.

docs/src/rfc/text/0000-patina-crates-naming-and-categorization.md
vineelko: Incorporated the changes to the new layout in iteration 2 of the RFC. Resolving this thread.

docs/src/rfc/text/0000-patina-crates-naming-and-categorization.md
vineelko: Incorporated the changes to the new layout in iteration 2 of the RFC. Resolving this thread.

docs/src/rfc/text/0000-patina-crates-naming-and-categorization.md
vineelko: Incorporated the changes to the new layout in iteration 2 of the RFC. Resolving this thread.

docs/src/rfc/text/0000-patina-crates-naming-and-categorization.md
vineelko: Incorporated the changes to the new layout in iteration 2 of the RFC. Resolving this thread.

docs/src/rfc/text/0000-patina-crates-naming-and-categorization.md
vineelko: Settling on `/components/patina_samples`. Incorporated the changes to the new layout in iteration 2 of the RFC. Resolving this thread.

docs/src/rfc/text/0000-patina-crates-naming-and-categorization.md
vineelko: Incorporated the changes to the new layout in iteration 2 of the RFC. Resolving this thread.

docs/src/rfc/text/0000-patina-crates-naming-and-categorization.md
vineelko: Settling on `patina_core` to the root. Incorporated the changes to the new layout in iteration 2 of the RFC. Resolving this thread.

docs/src/rfc/text/0000-patina-crates-naming-and-categorization.md
vineelko: Settling on `patina_core` to the root. Incorporated the changes to the new layout in iteration 2 of the RFC. Resolving this thread.

docs/src/rfc/text/0000-patina-crates-naming-and-categorization.md
```diff
@@ -0,0 +1,138 @@
...
+### New Crate Organization
+
+The identified public crates from the dependency graph are shown below, along
+with the proposed crate names. Additional input is needed to finalize this
+classification.
+
+![Patina public crates](0000-patina-crates-naming-and-categorization/patina_public_crates.png)
+
+```text
+  Existing                             New
```
makubacki: Decision from today's dev meeting was to drop `patina_` prefix from directory names.

vineelko: Michael, is this also true for the root `dxe_core` directory?

makubacki: My is understanding is, yes. Since this is the "patina" repo I think prefixing directories in it at any level with "patina" was considered redundant. It still makes sense in the crate names as they're referenced as independent entities externally in manifest files.

vineelko: Updated  the same in the latest iteration. Resolving this thread.

vineelko:
Moving the RFC to FCP.

docs/src/rfc/text/0000-patina-crates-naming-and-categorization.md
vineelko: Keeping them internal for now.

vineelko:
As per the latest iteration and yesterday's Rust sync discussion, we decided not to use the `patina_` prefix for directory names, while retaining it for crate names. However, while trying to implement this change in parallel, I noticed that many places (mostly in documentation and `Cargo.toml` files) require extra care to handle this discrepancy.

I also strongly feel that anyone looking at the code might need to spend additional mental effort to differentiate between directory names and crate names (e.g., `dxe_core` vs `patina_dxe_core`, especially when used outside the .rs files).

Would it be worth reconsidering this decision and aligning the directory names with the crate names—that is, including the `patina_` prefix in the directory names as well? This would help eliminate differences and make it cleaner.


Thoughts? @makubacki @joschock @cfernald @Javagedes @os-d 

image

makubacki:
> Would it be worth reconsidering this decision and aligning the directory names with the crate names—that is, including the `patina_` prefix in the directory names as well? This would help eliminate differences and make it cleaner.
> 
> Thoughts? @makubacki @joschock @cfernald @Javagedes @os-d

The agreement to drop the `patina_` prefix was decided as a group in the meeting. I don't have a strong opinion and will defer to others.

joschock:
No strong preference from my side.

vineelko:
Final update during FCP:
  - Revert to using the `patina_` prefix for directory names, matching the crate names to simplify the differences.
  - Rename `0000-patina-crates-naming-and-categorization*` to `0007-patina-crates-naming-and-categorization*`

PR #409, opened 2025-05-09
Switch to a public registry (`patina-fw`) [Rebase & FF]

### LINKED ISSUE #398: [Task]: Explore publicly accessible Patina crate registry
### Overview

Currently crates are published to an internal Azure crate registry. This issue tracks exploring an alternative crate registry location with the same ODP permissions as the source code repositories.

### Alternatives Considered

_No response_

### Urgency

Low

I will implement the feature

No maintainer feedback needed

### Anything else?

_No response_

makubacki:
## Description

Closes #398

---

**Switch to a public registry (`patina-fw`)**

Moves to a temporary Azure Artifacts feed that serves as the Patina crate registry until the project's crates are published to crates.io.

Connect to the crate in a `.cargo/config.toml` with:

```
[registries]
temporary = { index = "sparse+https://pkgs.dev.azure.com/patina-fw/crates/_packaging/temporary/Cargo/index/" }
```

Feed URL:

https://dev.azure.com/patina-fw/crates/_artifacts/feed/temporary

Additional Setup Instructions:

https://dev.azure.com/patina-fw/crates/_artifacts/feed/temporary/connect

---

### Special Note

To clarify terminology the following are involved in the public feed:

- Organization: `patina-fw`
- Project: `crates`
- Feed: `temporary`

Originally, I tried making "crates.io" an upstream of the `temporary` feed so we could use a single feed for our published crates while also caching upstream crates using contributor permission in CI and for local users.

This would only work if the `crates` repository was set to `enterprise` access level, not `public`. In that case, the feed is also not public, which defeats the purpose of this change. A `401` occurs when the feed attempts to pull in the crate version from crates.io. This happens whether using a PAT (with proper permissions) or MSAL with `msrustup` as an authenticator. By simply making the project, `enterprise` both approaches begin working with the exact same token.

Therefore, the PR has been updated to pull all crates on crates.io from crates.io and only allow publishing to the `patina-fw` feed and also set up all crates published to that feed to be fetched from there.

---

### Special Note 2

`cargo release` does not seem to honor the `registry` setting for each dependency in the workspace Cargo.toml file. It was working previously based on the complete replacement of `crates-io` with:

```toml
[source.crates-io]
replace-with = "UefiRust"
```

Due to *Special Note* above, that's not possible with this registry as we can only use it for crates not being pulled from crates.io upstream.

After trying several approaches, the one that seems to work and is the least hacky is to patch each crate that would go to the `crates-io` registry to the local crate with:

```toml
[patch.crates-io]
dxe_core = { path = "core/dxe_core" }
```

Therefore, an update has been made to do that for every crate published in this repo.

An outstanding issue is that the `mtrr` and `paging` crates are not local to this repo but also not on crates.io. Therefore, this does not directly apply to those crates.

There are at least three ways to solve that:

1. Make the `mtrr` and `paging` crates public and publish their crates to crates.io
2. Move the `mtrr` and `paging` crates to the `patina` repo so they become "local paths"
3. Patch the `mtrr` and `paging` crates as git dependencies to their current ODP repo locations
4. Patch the `mtrr` and `paging` crates back to the `patina-fw` registry.

- (1) is ideal and simplest if we're ready to publish those crates
  - Note: `paging` is already taken on crates.io, so it will need to be renamed, which is inevitable anyway.
- (2) is less ideal since it defeats the separation of these crates into their own repos and the separation, we wanted for those crates from the rest of the Patina code.
- (3) is even less ideal because:
  - Private ODP GitHub repos (which these are) would still need to be authenticated in CI which is another layer of complexity to add.
  - The publish step will be referencing the `main` branches of those repos which is different from the other cargo commands operating on the workspace dependencies
  - Because other cargo commands are using the "real" `mtrr` and `paging` crates, this warning will appear with the patches present when running commands like `cargo update:

    ```
    Updating `patina-fw` index
    warning: Patch `mtrr v0.1.1 ([https://github.com/OpenDevicePartnership/mtrr.git?branch=main#ab5369bc)`](https://github.com/OpenDevicePartnership/mtrr.git?branch=main#ab5369bc)%60 was not used in the crate graph.
    Perhaps you misspelled the source URL being patched.
    Possible URLs for `[patch.<URL>]`:
        patina-fw
    warning: Patch `paging v4.0.2 ([https://github.com/OpenDevicePartnership/paging.git?branch=main#218277d8)`](https://github.com/OpenDevicePartnership/paging.git?branch=main#218277d8)%60) was not used in the crate graph.      
    Perhaps you misspelled the source URL being patched.
    Possible URLs for `[patch.<URL>]`:
        patina-fw
    ```

- (4) was chosen to solve this problem immediately with work on (1) in progress as a long-term solution.

---

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [x] Includes documentation?

## How This Was Tested

- `cargo update` locally
- `cargo build` in CI
- `cargo release` (dry run) locally and in CI

## Integration Instructions

- The crate will now be published to a public ADO registry. Connect to it as follows:

```
[registries]
temporary = { index = "sparse+https://pkgs.dev.azure.com/patina-fw/artifacts/_packaging/temporary/Cargo/index/" }
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -66,8 +66,6 @@ x86_64 = { version = "=0.15.2", default-features = false }
 criterion = { version = "^0.5" }
 mockall = { version = "0.13.0" }
 rand = { version = "0.8" }
-sample_components = { version = "15.0.0", path = "components/sample_components", registry = "patina-fw" }
-section_extractor = { version = "15.0.0", path = "core/section_extractor", registry = "patina-fw" }
 uint = { version = "^0.10" }
 winapi = { version = "0.3" }
 


makubacki:
Going to add the PAT (with contributor permission) for the new feed soon.

Cargo.toml
```diff
@@ -33,41 +33,41 @@ lazy_static = { version = "^1" }
...
-uefi_debugger = { version = "15.0.0", path = "core/uefi_debugger" }
-uefi_depex = { version = "15.0.0", path = "core/uefi_depex" }
-uefi_device_path = { version = "15.0.0", path = "core/uefi_device_path" }
-uefi_performance = { version = "15.0.0", path = "core/uefi_performance" }
-uefi_protocol = { version = "15.0.0", path = "sdk/uefi_protocol" }
-uefi_sdk = { version = "15.0.0", path = "sdk/uefi_sdk" }
-uefi_sdk_macro = { version = "15.0.0", path = "sdk/uefi_sdk_macro" }
-uefi_test = { version = "15.0.0", path = "sdk/uefi_test" }
-uefi_test_macro = { version = "15.0.0", path = "sdk/uefi_test_macro" }
+uefi_collections = { version = "15", path = "core/uefi_collections", default-features = false, registry = "patina-fw" }
```
cfernald: Shouldn't we use the full version? so that way we can treat breaking changes between "internal" components as non-breaking since they move in lock-step

core/dxe_core/Cargo.toml
```diff
@@ -55,9 +55,9 @@ uefi_test = { workspace = true }
 # cargo-release, however they are still used when verifying the package is compileable. Any dev-dependency that is
 # local to the workspace should be added via `path` instead of `workspace = true` so that that it does not fail to
 # compile due to the dev-dependency not being published yet.
-sample_components = { path = "../../components/sample_components" }
-section_extractor = { path = "../../core/section_extractor" }
-uefi_collections = { path = "../../core/uefi_collections" }
+sample_components = { workspace = true }
```
vineelko: +@Javagedes

The suggestion for local dev-dependencies is to use the local paths directly as mentioned in the above comment.

sdk/driver_binding/Cargo.toml
```diff
@@ -26,5 +26,5 @@ mockall = { workspace = true, optional = true }
 # cargo-release, however they are still used when verifying the package is compileable. Any dev-dependency that is
 # local to the workspace should be added via `path` instead of `workspace = true` so that that it does not fail to
 # compile due to the dev-dependency not being published yet.
-boot_services = { path="../../sdk/boot_services", features = ["mockall"] }
+boot_services = { workspace = true, features = ["mockall"] }
```
vineelko: Same here.

Cargo.toml
makubacki: Can you please elaborate a bit more? This is to reduce churn in the file for compatible versions published to the registry.

core/dxe_core/Cargo.toml
Javagedes: Yes. Good catch. This will fail to release due to circular dependencies.

Cargo.toml
Javagedes: Yes, we should now hardcode all three versions for any crates that are specifically coming from this repository. (so all the patina crates) We don't accidently want a torn state in a platform where they are using 15.0.0 of uefi_test and 15.0.1 of uefi_sdk.

This should not reduce churn because the platform will still just say "15", and that will auto to 15.0.1 as a whole.

cfernald: From my understanding, one of the benefits of the mono repo was to avoid "breaking" churn for affectively internal changes, but this requires strict matching for these crates. 

That is, if in version 20.0.1 we make a breaking change between uefi_cpu and dxe_core, this would be problematic if for some reason dxe_core = 20.0.0 but uefi_core = 20.0.1. Which could easily happen if the caller wants to use 20.0.0 but doesnt specify a uefi_core version.

Similarly, if we add a functionality in uefi_sdk in 30.0.1 and add support for it in the core in 30.0.1 this may not compile if we are using uefi_sdk 30.0.1 but dxe_core 30.0.0.

makubacki: I see. I think I just disagree then. If these are publicly consumable crates the versions should make sense for the crates not internal repo assumptions.

core/dxe_core/Cargo.toml
makubacki: Does this only happen outside the dry run?

Cargo.toml
makubacki: To get this in, I'll go ahead and use exact versions. But, for public crates, I think we need to treat versions as semantic versions between crates (because that's what they are).

> That is, if in version 20.0.1 we make a breaking change between uefi_cpu and dxe_core, this would be problematic if for some reason dxe_core = 20.0.0 but uefi_core = 20.0.1.

If `dxe_core` and `uefi_core` are both public crates, a breaking change in `uefi_cpu` for its consumer (`dxe_core`) should lead to its major version being updated.

> Similarly, if we add a functionality in uefi_sdk in 30.0.1 and add support for it in the core in 30.0.1 this may not compile if we are using uefi_sdk 30.0.1 but dxe_core 30.0.0.

If the functionality added in `uefi_sdk` 30.0.1 is backward compatible (not rolling the major version) and `dxe_core` is a consumer, it should continue to build.

makubacki: I understand we're trying to use the workspace versioning to our advantage but, imo, versioning becomes more complicated when the version for a crate stops representing what is actually happening in the crate.

makubacki: Updated to exact version.

core/dxe_core/Cargo.toml
makubacki: In any case, the dry run worked. But I'll follow the existing pattern.

PR #411, opened 2025-05-09
[REBASE&FF] Support Services as static references

Javagedes:
## Description

Updates the `IntoService` derive macro to additionally implement `IntoService` for a static reference of a struct, not just a concrete instance. This is to support scenarios where a developer must maintain their own static instance of the concrete type but want to be able to register that concrete instance as a service.

This scenario specifically came up when implementing a service that is also the backing functionality for an EDKII protocol where the protocol did not take a pointer to self as it's first parameter.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Unit Tests

## Integration Instructions

Users may now register static references as services as seen below:

```rust
static MY_STRUCT: MyStruct = MY_STRUCT::new(50);

fn entry_point(mut commands: Commands) -> Result<()> {
    commands.add_service(&MY_STRUCT);
}
```

### DIFF
--- a/sdk/uefi_sdk/src/component/service.rs
+++ b/sdk/uefi_sdk/src/component/service.rs
@@ -330,7 +330,7 @@ mod tests {
         storage.add_service(MyServiceImpl);
 
         #[derive(IntoService)]
-        #[service(Self)]
+        #[service(SomeStruct)]
         struct SomeStruct {
             x: u32,
         }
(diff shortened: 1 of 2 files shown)


makubacki:
Please hold until Monday for me.

PR #414, opened 2025-05-12
Debugger: Invalidate TLB and instruction caches on exit

cfernald:
## Description

The debugger may edit the page table mappings or instructions in the instruction cache. These caches are not inherently coherent with the data cache and so these should be invalidated on returning from the debugger. 

Notably, the x64 memory model does guarantee writes will invalidate the icache so long as they use the same linear address, which will be true since we only operate identity mapped.  

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on QEMU

## Integration Instructions

N/A

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -211,6 +211,13 @@ dependencies = ["individual-package-targets"]
 description = "Builds all rust documentation in the workspace. Example `cargo make doc`"
 env = { RUSTDOCFLAGS = "-D warnings"}
 command = "cargo"
+args = ["doc", "@@split(INDIVIDUAL_PACKAGE_TARGETS, )", "--features", "doc"]
+
+[tasks.doc-open]
+description = "Builds all rust documentation in the workspace and opens the documentation. Example `cargo make doc-open`"
+env = { RUSTDOCFLAGS = "-D warnings"}
+clear = true
+command = "cargo"
 args = ["doc", "@@split(INDIVIDUAL_PACKAGE_TARGETS, )", "--features", "doc", "--open"]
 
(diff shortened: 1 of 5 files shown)


core/uefi_debugger/src/arch/aarch64.rs
```diff
@@ -111,6 +111,16 @@ impl DebuggerArch for Aarch64Arch {
             if unsafe { core::slice::from_raw_parts(elr, instruction_size) } == breakpoint_instruction {
                 exception_info.context.elr += instruction_size as u64;
             }
+
+            // Always clear the ICache and TLB since the debugger may have altered
+            // instructions or the page tables.
+            unsafe {
+                asm!("dsb sy");
```
Javagedes: any reason you have these as individual `asm!` macros? also do we need to worry about any of the special configs like `preserve_flags`, `noreturn`, `nostack`?

I have no experience with the `asm!` macros, so this is more of an actual question :)

cfernald: It probably should just be 1 line. For whatever reason my brain just likes it this way, but I will change to the single asm call. 

As for the options, `preserve_flags` -> we don't care and none of these change flags anyways, `noreturn` -> it does fall through, `nostack` -> we dont care or are using the stack in the asm.

Javagedes: Thanks for the explanation :) I'll close this

PR #415, opened 2025-05-13
Handle All Notify Events and Prevent Double-Queueing of Event Notifications

### LINKED ISSUE #199: [Bug]: TPL restore is iterating on a fixed vector of events

### LINKED ISSUE #374: [Bug]: SignalEvent callbacks erronously stack when TPL is raised

### Current Behavior

When SignalEvent() is called at a TPL::TPL_HIGH_LEVEL, a callback to the notify function is queued but not executed. When SignalEvent() is called again at a TPL::PTL_HIGH_LEVEL, _another_ callback to the notify function is queued, and neither is executed yet.
When the TPL is lowered, the callback is executed twice.

### Expected Behavior

The C-core has the behavior of ignoring additional SignalEvent() calls when the events callback has already been queued, but not yet called. The effect with the same scenario above is that the callback is executed once.

### Steps To Reproduce

Call SignalEvent() at TPL::TPL_HIGH_LEVEL
Call SignalEvent() on the same event at TPL::TPL_HIGH_LEVEL
Lower the TPL, observe the event's callback is called twice

### Build Environment

```markdown
- OS(s): Windows 11
- Tool Chain(s): VS22022
- Targets Impacted: DEBUG
```

### Version Information

```text
Tag: 3.6.1
```

### Urgency

Medium

I will fix it

No maintainer feedback needed

### Anything else?

We're observing issues this on an Intel physical platform with I2cHost.c in MdeModulePkg expecting the C-core's behavior.

kouchekiniad:
## Description

When a TPL is restored, any notify events are processed, which may result in a notify callback being called. This callback may result in a notify events being added to the event queue. This change will continuously process all events in the event queue until the queue is empty. Additionally, this commit moves the EVENT_NOTIFIES_IN_PROGRESS static inside of the restore_tpl function, as it is only used in that function.

Closes #199

Additionally, this change guarantees that event notifications will not be double-queued in order to match the Tianocore implementation of events. Before this change, the implementation of `Ord` on `TaggedEventNotification` prevents an event notification being queued twice if `signal_event` is called one-after-the-other, but allows double-queuing in the scenario where an event is signaled, the pending events iterator is collected, then the event is signaled again before the event's callback is called. This scenario will be run into when an event is signaled, then the callback for an event earlier in the iterator signals the same event.

Closes #374

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Booted to Windows on x64 physical platform. Includes unit test.

## Integration Instructions

N/A

### DIFF
--- a/core/dxe_core/src/events.rs
+++ b/core/dxe_core/src/events.rs
@@ -11,8 +11,6 @@ use core::{
     sync::atomic::{AtomicBool, AtomicU64, AtomicUsize, Ordering},
 };
 
-use alloc::vec;
-
 use r_efi::efi;
 
 use mu_pi::protocols::timer;
@@ -29,7 +27,6 @@ pub static EVENT_DB: SpinLockedEventDb = SpinLockedEventDb::new();
 
 static CURRENT_TPL: AtomicUsize = AtomicUsize::new(efi::TPL_APPLICATION);
 static SYSTEM_TIME: AtomicU64 = AtomicU64::new(0);
-static EVENT_NOTIFIES_IN_PROGRESS: AtomicBool = AtomicBool::new(false);
 
 extern "efiapi" fn create_event(
     event_type: u32,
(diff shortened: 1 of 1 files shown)


kouchekiniad:
To not break notify_wait events, this change needs to go in at the same time as or right after #212

kouchekiniad:
@Javagedes I've cherry-picked #212 on top of my changes, with the intent to _not squash_ the commits before merging to preserve history.

Issue #410, opened 2025-05-09
[Bug]: FSB lock re-entrancy when signaling cache attribute change event group


kouchekiniad:
### Is there an existing issue for this?

- [x] I have searched existing issues

### Current Behavior

`spin_locked_gcd.rs`'s `set_paging_attributes` signals `CACHE_ATTRIBUTE_CHANGE_EVENT_GROUP`, which queues an event notify. This queuing occasionally causes a memory allocation when a TaggedEventNotification is inserted into the pending_notifies BTreeSet. When `set_paging_attributes` is called from the FSBs `allocate_pages`, these memory allocations result in FSB lock re-entrancies.

This issue is being observed on an x64 physical platform.

### Expected Behavior

The FSB lock should not be re-entered.

### Steps To Reproduce

Repeatedly allocate pages without signaling other events until the BTreeSet needs to allocate more memory.

### Build Environment

```markdown
- OS(s): Windows 11
- Tool Chain(s): VS2022
- Targets Impacted: DEBUG and RELEASE
```

### Version Information

```text
Commit: 6cdb95f7cabf0b14f85d2bdf3062de4f4a5f15d6
```

### Urgency

Medium

### Are you going to fix this?

I will fix it

### Do you need maintainer feedback?

No maintainer feedback needed

kouchekiniad:
Addressed by PR #416. Unfortunately there doesn't seem to be a super clean solution, but #416 is the least invasive one I've found.

PR #416, opened 2025-05-14
Prevent Allocaction in `signal_event` for `CACHE_ATTRIBUTE_CHANGE_EVENT_GROUP`

kouchekiniad:
## Description
This change addresses issue #410 by tracking pending events that are part of the `CACHE_ATTRIBUTE_CHANGE_EVENT_GROUP` in a separate vector pre-allocated to fit any TaggedEventNotifications generated by `signal_event`. This prevents any memory allocations from occurring when `CACHE_ATTRIBUTE_CHANGE_EVENT_GROUP` is signaled, which is necessary to prevent a FSB lock re-entrancy on the global allocator which signals this group.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Booted to Windows on x64 physical platform with these changes based on top of #415 and #212.

## Integration Instructions

N/A

### DIFF
--- a/core/dxe_core/src/event_db.rs
+++ b/core/dxe_core/src/event_db.rs
@@ -18,7 +18,7 @@ use alloc::{
 };
 use core::{cmp::Ordering, ffi::c_void, fmt};
 use r_efi::efi;
-use uefi_sdk::error::EfiError;
+use uefi_sdk::{error::EfiError, guid::CACHE_ATTRIBUTE_CHANGE_EVENT_GROUP};
 
 use crate::tpl_lock;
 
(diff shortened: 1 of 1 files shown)


kouchekiniad:
@os-d Unless I'm missing something I'm not seeing the FSB raising/lowering the TPL in such a way that the queued event notify would be executed within the FSB locked region

os-d:
> @os-d Unless I'm missing something I'm not seeing the FSB raising/lowering the TPL in such a way that the queued event notify would be executed within the FSB locked region

Oh, I might have been misunderstanding the issue. The issue is that the FSB code holds the lock while `allocate_pages` is being called, which triggers the event, which attempts to allocate, hitting re-entrancy?

In that case, we have a restriction on any memory service triggering an event, right? This just happens to be the only one now.

kouchekiniad:
@os-d Yep, that's right. I think we have to assume that we can allocate when restoring the TPL or else a lot of things break, this change just prevents allocation for SignalEvent()/SignalGroup() for this event group. You're also right that this would need to be expanded if the GCD/FSB signaled any other event/group in a locked region, so this PR assumes that having to signal the cache attributes changed event is more of a one-off. Generalizing the solution for all events would require switching to a `pending_notifies` data structure that could handle pre-allocation, and pre-allocating space for all existing events, which would have considerably more overhead.

os-d:
Well, we currently also signal the memory map change event from the GCD: https://github.com/OpenDevicePartnership/patina/blob/36ffc7b9c8c83fa7f559a23b3dc8c69f41228348/core/dxe_core/src/gcd/spin_locked_gcd.rs#L2217. Which can be called in exactly the same place with the FSB lock held.

@joschock @makubacki @Javagedes do you have thoughts here? We may have to move the event code to not allocate memory to prevent lock re-entrancy. Or adjust how we do the event signalling in the memory system, though that is not trivial with the locks, mostly in the case where the FSB calls back into the GCD to refill pages.

kouchekiniad:
You're right; I could expand this PR to also keep EVENT_GROUP_MEMORY_MAP_CHANGE in the same pre-allocatable vector, but I do see how that's a bit messy

kouchekiniad:
Abandoning this PR in favor of a broader refactoring of how the FSB interacts with the GCD.

PR #424, opened 2025-05-15
Quick cleanup to readme.md file

rogurr:
## Description

Quick cleanup to readme.md file to include links to other repositories, change to compilation steps, and update formatting.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [X] Includes documentation?

## How This Was Tested

Ran through steps on a personal PC with no access to Microsoft internal servers to confirm everything is accessible through GitHub or public access points.

## Integration Instructions

N/A


README.md
```diff
@@ -50,67 +52,59 @@ docs.rs once we begin uploading to crates.io.
 
 ## First-Time Tool Setup Instructions
 
-The following instructions install Rust.
+1. Follow the steps outlined by [Getting Started - Rust Programming Language (rust-lang.org)](https://www.rust-lang.org/learn/get-started)
+to install, update (if needed), and test cargo/rust.
 
-1. Download and install rust/cargo from [Getting Started - Rust Programming Language (rust-lang.org)](https://www.rust-lang.org/learn/get-started).
-   > `rustup-init` installs the toolchain and utilities.
+2. The [toolchain] section of the [rust-toolchain.toml](https://github.com/OpenDevicePartnership/patina/blob/personal/rogurr/md/rust-toolchain.toml)
```
makubacki: `[toolchain]` needs to have a URL if it's a link.

makubacki: Needs to point to the main branch instead of a personal branch.

```suggestion
2. The [toolchain] section of the [rust-toolchain.toml](https://github.com/OpenDevicePartnership/patina/blob/HEAD/rust-toolchain.toml)
```

README.md
```diff
@@ -50,67 +52,59 @@ docs.rs once we begin uploading to crates.io.
...
-
-3. Install toolchain specified in `rust-toolchain.toml`
-
-   \>`rustup toolchain install`
-
-4. While the specific toolchains and components specified in `[toolchain]` section of the `rust-toolchain.toml` are
-automatically installed with `rustup toolchain install`, the tools in the `[tools]` section, such as `cargo-make`
-are not. At a minimum, you should download `cargo-make` and `cargo-tarpaulin`, however it is suggested that you
-download all tools in the `[tools]` section of the `rust-toolchain.toml`.
+3. The [tools] section of the [rust-toolchain.toml](https://github.com/OpenDevicePartnership/patina/blob/personal/rogurr/md/rust-toolchain.toml)
```
makubacki:
```suggestion
3. The `[tools]` section of the [rust-toolchain.toml](https://github.com/OpenDevicePartnership/patina/blob/personal/rogurr/md/rust-toolchain.toml)
```

README.md
makubacki:
```suggestion
2. The `[toolchain]` section of the [rust-toolchain.toml](https://github.com/OpenDevicePartnership/patina/blob/personal/rogurr/md/rust-toolchain.toml)
```

makubacki: Made a code suggestion I think matches the intention.

README.md
```diff
@@ -50,67 +52,59 @@ docs.rs once we begin uploading to crates.io.
...
+```
+cargo make -p development build-aarch64
+   - or -
+cargo make -p release build-aarch64
+```
 
 ## Test
-
-- `cargo make test`
+Use the test command to invoke a test build and execute all unit tests.
```
makubacki: You'll need an empty line between section headings and content for markdownlint.

README.md
```diff
@@ -50,67 +52,59 @@ docs.rs once we begin uploading to crates.io.
...
-1. This project uses `RUSTC_BOOSTRAP=1` environment variable due to internal requirements
-   1. This puts us in parity with the nightly features that exist on the toolchain targeted
-   2. The `nightly` toolchain may be used in place of this
+- This project uses the "RUSTC_BOOSTRAP=1" environment variable due to internal requirements which puts us in parity with the
+nightly features that exist on the toolchain targeted.  The "nightly" toolchain may be used in place of this.
 
 ## Contributing
-
-- Review Rust Documentation in the `docs` directory.
+- Review Rust Documentation in the [/docs](https://github.com/OpenDevicePartnership/patina/tree/personal/rogurr/md/docs) directory.
```
makubacki: Remove personal branch. Point to intro doc on `HEAD`.

```suggestion
- Review Rust Documentation in the [/docs](https://github.com/OpenDevicePartnership/patina/blob/HEAD/docs/src/introduction.md) directory.
```

README.md
```diff
@@ -50,67 +52,59 @@ docs.rs once we begin uploading to crates.io.
...
-3. Install toolchain specified in `rust-toolchain.toml`
-
-   \>`rustup toolchain install`
-
-4. While the specific toolchains and components specified in `[toolchain]` section of the `rust-toolchain.toml` are
-automatically installed with `rustup toolchain install`, the tools in the `[tools]` section, such as `cargo-make`
-are not. At a minimum, you should download `cargo-make` and `cargo-tarpaulin`, however it is suggested that you
-download all tools in the `[tools]` section of the `rust-toolchain.toml`.
+3. The [tools] section of the [rust-toolchain.toml](https://github.com/OpenDevicePartnership/patina/blob/personal/rogurr/md/rust-toolchain.toml)
+file contains items to support the pipeline and must be installed manually.  A local build does not need them all, but at a minimum, cargo-make
```
Javagedes:
```suggestion
file contains tools to support some commands, such as `cargo make coverage`, and must be installed manually.  A local build does not need them all, but at a minimum, cargo-make
```

README.md
```diff
@@ -50,67 +52,59 @@ docs.rs once we begin uploading to crates.io.
...
+and cargo-tarpaulin should be installed.
+   ```
```
Javagedes: We also suggest cargo binstall <tool> (where cargo-binstall must be installed) because it directly downloads a binary instead of performing local compilation, which is much faster.

https://github.com/cargo-bins/cargo-binstall?tab=readme-ov-file#installation

README.md
```diff
@@ -50,67 +52,59 @@ docs.rs once we begin uploading to crates.io.
...
+cargo make coverage dxe_core
+```
 
 ## Notes
-
-1. This project uses `RUSTC_BOOSTRAP=1` environment variable due to internal requirements
-   1. This puts us in parity with the nightly features that exist on the toolchain targeted
-   2. The `nightly` toolchain may be used in place of this
+- This project uses the "RUSTC_BOOSTRAP=1" environment variable due to internal requirements which puts us in parity with the
+nightly features that exist on the toolchain targeted.  The "nightly" toolchain may be used in place of this.
```
Javagedes: I would maybe add a note that all the cargo make commands set RUSTC_BOOTSTRAP=1 automatically, so if someone is using `cargo make`, they do not need to change their environment

README.md
makubacki: I think it makes sense to give both options here. Recommend `cargo binstall` but also provide the alternative and just explain that it will be slower.

`cargo binstall` is great and what we primarily use but does have the rare corner case like the glibc the precompiled binary was built against being incompatible with your Linux installation. I don't think we need to go into all that detail just add a simple alternative command.

Javagedes: Yes, Providing docs for both is good.

README.md
rogurr: Ah, good catch.

README.md
rogurr: It looks funny to me to have inline words in a text box like this, but if it is the standard we want to use, there are several other places that probably need to be updated.

README.md
rogurr: Yup, your suggestion was the intention.  Thanks.

README.md
rogurr: binstall is currently not part of the [tools] section of the rust-toolchain.toml, should we update that file first before suggesting it?

Also, I tried installing it and get an error stating it needs compiler 1.85 instead of 1.84 as indicated.  Maybe we should open a task to update all places then add it to the readme later?

README.md
rogurr: I think that note was a holdover from original data, do you have an objection to just removing it?  I liked how simplistic this readme was so maybe keeping it would just add confusion?

makubacki: This could be argued either way. But since we use `cargo make` for build and `Makefile.toml` encapsulates this detail, I think we can keep this simpler and say "use cargo make or look at Makefile.toml for more details".

README.md
makubacki: `cargo binstall` is installed differently in CI than it will be locally. But all installation methods are described well in the [cargo-binstall repo readme](https://github.com/cargo-bins/cargo-binstall). Therefore, I think you should just link to that and tell the user to following installation instructions for their platform.

README.md
rogurr: After talking to others using the code, it appears this is a pertinent note to help understand using the nightly features.  I updated the text to indicate the makefile uses the setting, but left it in for clarity.

README.md
rogurr: Documentation updated to point to the GitHub cargo-binstall repo and state it is helpful, but optional to speed up the build.

PR #425, opened 2025-05-16
Reorganized and Renamed Patina Directories and Crates for Consistency

vineelko:
## Description

- Updated the directory and crate organization as per https://github.com/OpenDevicePartnership/patina/pull/408
```text
Before                           After
                                 ├── patina_dxe_core
├── components                   ├── components
│   ├── adv_logger               │   ├── patina_adv_logger
│   └── sample_components        │   └── patina_samples
├── core                         ├── core
│   ├── dxe_core                 │   │
│   ├── section_extractor        │   ├── patina_section_extractor
│   ├── stacktrace               │   ├── patina_stacktrace
│   ├── uefi_collections         │   ├── patina_internal_collections
│   ├── uefi_cpu                 │   ├── patina_internal_cpu
│   ├── uefi_debugger            │   ├── patina_debugger
│   ├── uefi_depex               │   ├── patina_internal_depex
│   ├── uefi_device_path         │   ├── patina_internal_device_path
│   └── uefi_performance         │   └── patina_internal_performance
└── sdk                          └── sdk
    ├── boot_services                ├── patina_boot_services
    ├── driver_binding               ├── patina_driver_binding
    ├── runtime_services             ├── patina_runtime_services
    ├── tpl_mutex                    ├── patina_tpl_mutex
    ├── uefi_protocol                ├── patina_uefi_protocol
    ├── uefi_sdk                     ├── patina_sdk
    ├── uefi_sdk_macro               ├── patina_sdk_macro
    ├── uefi_test                    ├── patina_test
    └── uefi_test_macro              └── patina_test_macro
```
- Ensure that `cargo make all` works
- [ ] Impacts functionality?
- [ ] Impacts security?
- [x] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Booted to UEFI Shell on Q35
image

## Integration Instructions

The functionality hasn't changed, but because the directory and crate names have changed - impacting `use` statements and crate references like `patina_sdk::...` - extra effort is needed to rebase or cherry-pick any ongoing PRs.

### DIFF
--- a/.github/release-draft-config.yml
+++ b/.github/release-draft-config.yml
@@ -11,8 +11,8 @@
 # Copyright (c) Microsoft Corporation.
 # SPDX-License-Identifier: BSD-2-Clause-Patent
 ##
-name-template: 'v$RESOLVED_VERSION'
-tag-template: 'v$RESOLVED_VERSION'
+name-template: 'patina-v$RESOLVED_VERSION'
+tag-template: 'patina-v$RESOLVED_VERSION'
 
 commitish: refs/heads/main
 filter-by-commitish: true


os-d:
I did not see the request to review adding `patina_` back to all the directory names until now and the RFC is approved folks all agreeing, so I don't think this needs to change, but I will say seeing the full list where every single directory is named `patina_` does seem a little silly. But, if no one else has objections, this seems fine to me. I understand aligning with the crate name and having the crate name start with `patina_` makes sense to me.

components/patina_adv_logger/Cargo.toml
```diff
@@ -8,24 +8,19 @@ edition.workspace = true
 publish.workspace = true
 description = "Advanced logger support."
 
-# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
-
-[lib]
-path = "src/lib.rs"
```
os-d, line 14: This is no longer needed?

docs/src/concepts.md
```diff
@@ -44,6 +44,6 @@ types of PCDs with both static and dynamic values.
 
 Due to the monolithic nature of the pure rust DXE Core, this complex system for platform
 configuration is no longer necessary. Instead, we can perform configuration in code, and share
-configuration values between the dxe_core and any driver being compiled as a part of the dxe_core.
+configuration values between the dxe core and any driver being compiled as a part of the dxe core.
```
os-d: nit: Any of these places where we went from `dxe_core` the crate name to `dxe core` the entity, I would suggest

```suggestion
configuration values between the DXE Core and any driver being compiled as a part of the DXE Core.
```

There are a lot of instances throughout the files.

vineelko:
> I did not see the request to review adding `patina_` back to all the directory names until now and the RFC is approved folks all agreeing, so I don't think this needs to change, but I will say seeing the full list where every single directory is named `patina_` does seem a little silly. But, if no one else has objections, this seems fine to me. I understand aligning with the crate name and having the crate name start with `patina_` makes sense to me.

Yes, it might seem redundant, but trust me - when actually implementing the change, having the directory names differ from the crate names made things much worse, especially since both used underscores. My find and replace skills needed an upgrade! Hence, I had to revert that decision.

makubacki:
@vineelko, you're planning to squash this, right?

components/patina_adv_logger/Cargo.toml
vineelko: Yes, these are redundant if `src\lib.rs` is present then cargo will treat it as a library crate.

vineelko:
> @vineelko, you're planning to squash this, right?

Yes Michael 😄

docs/src/concepts.md
vineelko: Let me take a 2nd pass.

Cargo.toml
cfernald, line 1: I think we should reset the version number to 0.0.1 (or at least not 15) while we are doing this since these are new crates names anyways. 

Alternatively, we could so such a reset before full open-source (and pushing to crates-io). Or both even. 

@makubacki, @Javagedes , thoughts?

vineelko: No, I will update the version. Since all these crates are new, 0.0.1 should work but I am torn between 0.0.1 vs 16.0.0.

makubacki: 0.0.1 is fine with me.

docs/src/patina.md
```diff
@@ -288,10 +288,10 @@ Three main types of testing are currently supported.
   state of the module. Only the external interfaces are being tested. Cargo will detect and run these tests with the
   same command as for unit tests. More information about integration tests are available in the
   [cargo book entry](https://doc.rust-lang.org/rust-by-example/testing/integration_testing.html).
-- **On-platform tests** are supported with code in a crate called `uefi_test` that provides a testing framework similar
+- **On-platform tests** are supported with code in a crate called `patina_test` that provides a testing framework similar
   to the typical rust testing framework. The key difference is that instead of tests being collected and executed on
-  the host system, they are instead collected and executed via a component (`uefi_test::TestRunner`) provided by
-  the same crate. The platform must register this component with the `dxe_core`. The dxe_core will then dispatch this
+  the host system, they are instead collected and executed via a component (`patina_test::TestRunner`) provided by
+  the same crate. The platform must register this component with the `dxe core`. The dxe core will then dispatch this
```
makubacki:
```suggestion
  the same crate. The platform must register this component with the DXE Core. The DXE Core will then dispatch this
```

docs/src/patina.md
makubacki, line 1: There is a line in the file (currently 255) that says "`uefi_dxe_core` as a Library Crate". Can you update that to `patina_dxe_core`?

makubacki:
There are a couple of references in documentation that need cleaned up:

- "uefi_event crate" in docs/src/dxe_core/events.md (currently Line 10)
- "uefi_pecoff" in docs/src/dxe_core/images.md (currently Line 58)

makubacki:
I think we should change the `uefi_depex_lib` filter name to `depex` or something similar in this change.

Cargo.toml
Javagedes: The only thing worth considering is whether we want to go open source with 0.0.1 (so save this change until then), or reset now and go open source with whatever we happen to be at when we are ready

makubacki: I was prepared to start at 0.0.1 now and keep the version going into open source. A larger audience is already starting to use this so resetting and keeping the version going under the new crate names seems the cleanest to me to keep things obvious and consistent moving forward.

docs/src/concepts.md
vineelko: Fixed in the latest iteration. Resolving the thread.

Cargo.toml
vineelko: Updated to 0.0.1 in the latest iteration. Resolving the thread.

docs/src/patina.md
vineelko: Fixed in the latest iteration. Resolving the thread.

docs/src/patina.md
vineelko: Fixed in the latest iteration. Resolving the thread.

PR #429, opened 2025-05-16
Make copy of device path in LoadImage for loaded image protocol

cfernald:
## Description

This change makes a copy of the device path in the `LoadImage` function for the loaded image protocol. This is not explicitly state for loaded_image_protocol, but is mentioned for the required by the UEFI specification for `EFI_LOADED_IMAGE_DEVICE_PATH_PROTOCOL`.

This specifically revolves a use after free issue that presented in the shell which happend to overwrite a device path when it was re-allocated this memory while the pointer was still in the loaded image protocol. Whenever this protocol is next inspected it can cause unexpected behavior. Specifically, the `dh` command would cause the device to get stuck in a loop trying to parse the invalid path and eventually assert.

Relevant UEFI Description:

image

This is also done by the C Core:

image

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on Q35

## Integration Instructions

N/A

### DIFF
--- a/patina_dxe_core/src/image.rs
+++ b/patina_dxe_core/src/image.rs
@@ -994,11 +994,12 @@ pub fn core_load_image(
     }
 
     if let Some(path) = fixed_file_path {
-        let copy = Box::into_raw(
-            copy_device_path_to_boxed_slice(path).map_err(|status| EfiError::status_to_result(status).unwrap_err())?,
-        ) as *mut efi::protocols::device_path::Protocol;
-
-        image_info.file_path = copy;
+        if !path.is_null() {
+            image_info.file_path = Box::into_raw(
+                copy_device_path_to_boxed_slice(path)
+                    .map_err(|status| EfiError::status_to_result(status).unwrap_err())?,
+            ) as *mut efi::protocols::device_path::Protocol;
+        }
     }
 
     let mut private_info = core_load_pe_image(image_to_load.as_ref(), image_info)


core/dxe_core/src/image.rs
```diff
@@ -986,12 +987,20 @@ pub fn core_load_image(
...
-            image_info.file_path = file_path as *mut efi::protocols::device_path::Protocol;
+            fixed_file_path = Some(file_path as *mut efi::protocols::device_path::Protocol);
         } else {
-            image_info.file_path = file_path;
+            fixed_file_path = Some(file_path);
         }
     }
 
+    if let Some(path) = fixed_file_path {
+        let copy = Box::into_raw(
```
makubacki: nit: I think it'd be a little easier to read without using `copy`.

core/dxe_core/src/image.rs
```diff
@@ -986,12 +987,20 @@ pub fn core_load_image(
             let device_path_size_minus_end_node: usize =
                 device_path_size.saturating_sub(core::mem::size_of::<efi::protocols::device_path::Protocol>());
             let file_path = unsafe { (file_path as *const u8).add(device_path_size_minus_end_node) };
-            image_info.file_path = file_path as *mut efi::protocols::device_path::Protocol;
+            fixed_file_path = Some(file_path as *mut efi::protocols::device_path::Protocol);
         } else {
-            image_info.file_path = file_path;
+            fixed_file_path = Some(file_path);
         }
     }
```
os-d: Do we need to write null to image_info.file_path in the else case from the outer if statement? i.e. if device_handle != INVALID_HANDLE and file_path.is_null().

cfernald: Switch to only set if its not null. This is consistent to use the default of other scenario (which is null)

PR #433, opened 2025-05-20
[REBASE&FF] uefi_sdk: support Service const instantiation

Javagedes:
## Description

Adds support for const instantiation of a service (in an uninitialized state) to support using services in const contexts without the need of wrapping it in some type of container object.

This is done by using a `OnceCell`, which does come with a performance cost, which was mitigated (to almost exactly the same performance) by updating the `Deref` implementation to hint to the compiler that one path is unreachable (which it is, per the safety notes).

This also comes with a commit to add service benchmarks to benchmark performance of calling a services function, and the performance impact of using a dyn object service vs a concrete service. For anyone curious, these were the performance numbers:

| description | dyn    | concrete |
|-------------|--------|--------|
| Prior Art   | ~3.7ns | ~2.5ns |
| With const  | ~3.6ns | ~2.4ns |

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [x] Includes documentation?

## How This Was Tested

Multiple Tests and new benchmarks

## Integration Instructions

N/A

### DIFF
--- a/patina_dxe_core/src/cpu_arch_protocol.rs
+++ b/patina_dxe_core/src/cpu_arch_protocol.rs
@@ -55,7 +55,7 @@ extern "efiapi" fn flush_data_cache(
     length: u64,
     flush_type: CpuFlushType,
 ) -> efi::Status {
-    let cpu = get_impl_ref(this).cpu;
+    let cpu = &get_impl_ref(this).cpu;
 
     let result = cpu.flush_data_cache(start, length, flush_type);
 
@@ -86,7 +86,7 @@ extern "efiapi" fn get_interrupt_state(this: *const Protocol, state: *mut bool) 
 }
 
 extern "efiapi" fn init(this: *const Protocol, init_type: CpuInitType) -> efi::Status {
-    let cpu = get_impl_ref(this).cpu;
+    let cpu = &get_impl_ref(this).cpu;
 
     let result = cpu.init(init_type);
 
@@ -120,7 +120,7 @@ extern "efiapi" fn get_timer_value(
     timer_value: *mut u64,
     timer_period: *mut u64,
 ) -> efi::Status {
-    let cpu = get_impl_ref(this).cpu;
+    let cpu = &get_impl_ref(this).cpu;
 
     let result = cpu.get_timer_value(timer_index);
 


sdk/patina_sdk/src/component/service.rs
```diff
@@ -188,11 +188,44 @@ pub trait IntoService {
...
+
+    /// Replaces the uninitialized service with the provided, initialized, service.
+    ///
+    /// ## Panics
+    ///
+    /// Panics if the service is already initialized or if the provided service is not initialized.
+    pub fn replace(&self, service: &Service<T>) {
+        #[allow(suspicious_double_ref_op)]
+        // We do in-fact want a copy of the static reference, not a copy of the underlying type.
+        let v = service.value.get().expect("Provided Service was not initialized!").clone();
```
cfernald: Why is a clone necessary on a reference?

sdk/patina_sdk/src/component/service.rs
```diff
@@ -229,32 +262,67 @@ impl<T: ?Sized + 'static> Service<T> {
...
-        self.value.downcast_ref().unwrap_or_else(|| panic!("Config should be of type {}", core::any::type_name::<T>()))
+        if let Some(service) = self.value.get() {
+            if let Some(service) = service.downcast_ref() {
+                service
+            } else {
+                // Using core::hint::unreachable_unchecked() here results in the compiler optimizing away this entire
+                // code path, and will result in UB if the path is reached. This code path truly is unreachable as we
+                // (as patina developers) control all ways of instantiating a Service type.
+                //
+                // The performance impact of using this over unreachable! or panic! is about 25% improved performance
```
cfernald, line 290: 25% of what? This routine?

Javagedes: Yes this routine. In local testing, we are 25% faster at calling a "do nothing" service function, which requires this deref. The benchmarks I added in this PR is how I determined that.

sdk/patina_sdk/src/component/service.rs
Javagedes: Clone gives you a copy of the inner value, getting rid of the reference. I was using that to go from `&&dyn Any` to `&dyn Any`. You were right though, it was not necessary. I just dereferenced it instead, to get the inner value I cared about.

PR #434, opened 2025-05-20
RFC: Standardize crate layout for component crates

Javagedes:
An RFC for defining a standardize crate layout for any crate who produces a component

Current RFC State: **FCP**


docs/src/rfc/text/0000-standardize-component-crate-layout.md
```diff
@@ -0,0 +1,80 @@
...
+## Rust Code Design
+
+The current design requirement is suggested, and looking for feedback and improvements. Once this RFC is accepted,
+documentation will be added to the patina mdbook that lays out the requirements defined here.
+
+The intent is for this RFC is to define certain modules that must exist and be accessable via the root of the library
+crate. It does not restrict the possibility of other modules existing, only that certain modules must exist at the root
+of they crate given certain circumstances as defined below:
+
+1. No public definitions are accessible via the top level lib.rs (or equivalent) module, only public modules.
```
joschock: are re-exports permitted? if so, where do they go?

docs/src/rfc/text/0000-standardize-component-crate-layout.md
```diff
@@ -0,0 +1,80 @@
...
+
+N/A
+
+## Goals
+
+Define a standard layout for a crate that produces a component(s)
+
+## Requirements
+
+A standard layout for a crate that produces a component(s) is defined
```
joschock: this could be expanded a bit to speak to some of the guiding requirements behind some of the design choices below. For example, a possible requirement might be `minimize namespace pollution` which informs some of the guidelines about what is allowed to be exported where. Another might be `formalize component architecture vocabulary` or something to that effect -i.e. we should all use the same terms/filenames/whatever for the same things.

docs/src/rfc/text/0000-standardize-component-crate-layout.md
```diff
@@ -0,0 +1,80 @@
+# RFC: Standardize Component Crate Layout
+
+The purpose of this RFC is to solicit feedback on setting a standardized layout for crates to produce components, so
+that when a platform wishes to consume a component, there is a well-defined layout for where any Component, Service,
+Config, Hob, etc. definitions are located and accesses from.
```
makubacki:
```suggestion
Config, Hob, etc. definitions are located and accessed from.
```

docs/src/rfc/text/0000-standardize-component-crate-layout.md
```diff
@@ -0,0 +1,80 @@
...
+
+## Change Log
+
+- YYYY-MM-DD: Initial RFC created.
+
+## Motivation
+
+With more components being individually developed and packaged into separate crates, it becomes important to
+standardize the layout of these crates to make consumption of their functionality easy for the platform. Forcing a
+platform to understand how each component is laid out is burdomsome and time consuming.
```
makubacki:
```suggestion
platform to understand how each component is laid out is burdensome and time consuming.
```

docs/src/rfc/text/0000-standardize-component-crate-layout.md
makubacki: Yeah, this section is meant to help relate points in the solution to the requirements defined.

docs/src/rfc/text/0000-standardize-component-crate-layout.md
```diff
@@ -0,0 +1,80 @@
...
+N/A
+
+## Rust Code Design
+
+The current design requirement is suggested, and looking for feedback and improvements. Once this RFC is accepted,
+documentation will be added to the patina mdbook that lays out the requirements defined here.
+
+The intent is for this RFC is to define certain modules that must exist and be accessable via the root of the library
+crate. It does not restrict the possibility of other modules existing, only that certain modules must exist at the root
+of they crate given certain circumstances as defined below:
```
makubacki:
```suggestion
of the crate given certain circumstances as defined below:
```

docs/src/rfc/text/0000-standardize-component-crate-layout.md
```diff
@@ -0,0 +1,80 @@
...
+
+1. No public definitions are accessible via the top level lib.rs (or equivalent) module, only public modules.
+2. `component` module: This module must always exist, and contain the publicly importable component(s) for the crate.
+3. `config` module: This module may optionally exist if the component consumes configuration data that is registered
+   with the platform via `.with_config` and this config is not publically accessible via `patina_sdk` or elsewhere.
+4. `service` module: This module may optionally exist if the component produces a service that is not publically
+   via `patina_sdk` or another crate.
+5. `hob` module: This module may optionally exist only if the hob is expected to be consumable by other public
```
makubacki: I think there is a subtle nuance to the HOB definition owner that might affect this, but I don't feel very strongly about it.

If the HOB interface is defined by the component (e.g. a GUID HOB that the component has defined - not pre-existing), then the component crate should own the public interface. If the HOB definition exists independent of the component, then the component is just the "first" to claim the interface, but it does not necessarily belong under it. In that case, the interface should go to a neutral location with the component as "another consumer" of it.

Neutral locations may vary depending on the HOB interface:

- PI Spec defined: `mu_rust_pi`
- Otherwise (for now): `patina_sdk`

docs/src/rfc/text/0000-standardize-component-crate-layout.md
Javagedes: Overall, I think re-exports of types is fine. They would just go where a normal "new" type would be made public to support one of the modules - something like this:

| module | when acceptable |
| - | - |
| component.rs | Types used for private configuration of a struct component. |
| config.rs | Attributes of a config object. |
| error.rs | The actual error type, if re-using something common. |
| hob.rs | Not necessary. Hob.rs is private to the execution of the component. |
| service.rs | If part of the service interface uses this type. |

docs/src/rfc/text/0000-standardize-component-crate-layout.md
Javagedes: Thank you. This is good. I've updated the requirements.

With this, I've decided to slightly expand the standardization this is doing. It is now also standardizing Crate Name, Test Names and Documentation requirements.

docs/src/rfc/text/0000-standardize-component-crate-layout.md
Javagedes: I have updated to this

> 5. `hob` module: This module may optionally exist if a new guided hob type has been created for this component. The
   hob module and associated guided HOB(s) should be made public such that it can be consumed by others if the need
   arise. Any common or spec defined HOBs should be added to the associated crates (such as `mu_rust_pi`, `patina_sdk`, 
   etc.) rather than this crate. HOBs may become a common interface and should thus be moved to the appropriate crate.
   If the HOB type already exists elsewhere, the crate should consume that definition instead of making their own.

Javagedes:
Thank you for the reviews, @joschock @cfernald @makubacki . I've added a few new standardizations to the PR (along with requested changes). Please feel free to re-review.

Javagedes:
Moving to FCP

Javagedes:
Seeking Final feedback before merging (I will merge tomorrow morning).

PR #435, opened 2025-05-20
Move GCD interaction from FixedSizeBlockAllocator to SpinLockedFixedSizeBlockAllocator

kouchekiniad:
## Description

To avoid interacting with the GCD in the FSB's locked regions (which can potentially cause an FSB lock re-entrancy if the GCD allocates memory), this change moves all interaction with the GCD outside of the locked regions of the FSB and into the spin-locked FSB. As such, new memory allocations from the GCD are now either requested by the FSB then allocated from the spin-locked FSB (in the case of alloc/dealloc) or made by the spin-locked FSB, which then informs the FSB of the change.

Main changes include:
- Rather than expanding by allocating memory from the GCD directly, the FSB's `expand` is now passed-in a region of memory to expand itself with which is allocated from the GCD by the spin-locked FSB
  -  Attempted allocations in the FSB will return `FixedSizeBlockAllocatorError::OutOfMemory(additional_mem_requested)` if the FSB is unable to fit the allocation, prompting the spin-locked FSB to allocate additional memory with the GCD and call `expand` on the FSB
    - Note: This has the side effect that the FSB's `pool_allocation_calls` stat would be incremented twice in the case that additional memory needs to be allocated by the spin-locked FSB
- As there is no longer a direct analogue for the `alloc`/`dealloc` functions of `GlobalAlloc` and the `allocate`/`deallocate` functions of `Allocator`, the FSB now defines only one `alloc` and `dealloc` which are shared by both trait implementations
- The bulk of the allocation logic for `allocate_pages` and `free_pages` have been moved into the spin-locked FSB, which then notifies the FSB of any page allocations made/freed with `notify_page_allocation` and `notify_pages_freed`
- `preferred_range` has been renamed to `reserved_range` as the FSB itself makes no preference for putting new allocations into the `preferred_range`, and any preference the GCD would show for allocating memory in that region would be solely predicated by the "`preferred_range`" being allocated earlier in the boot process
- GCD interaction as part of `reserve_memory_region` has been moved to the spin-locked FSB, with the reserved memory region being registered in the FSB with `set_reserved_range(range)`. Rather than calling into `allocate_pages` then `free_pages` to reserve the memory region, `reserve_memory_region` now reserves memory in the GCD by directly calling `gcd.allocate_memory_space` then `gcd.free_memory_space_preserving_ownership` to avoid having to undo the stats side-effects of the `allocate_pages` and `free_pages` functions.
  - Note: This has the effect of not iterating `page_allocation_calls` and `page_free_calls` when calling `reserve_memory_pages`
- To keep things consistent, raw pointers and raw pointers + usize lengths have been largely replaced with `NonNull<u8>` and `NonNull<[u8]>` respectively.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Booted to Windows on x64 physical platform.

## Integration Instructions

N/A

### DIFF
--- a/docs/src/rfc/text/0008-consolidate-sdk-crates-into-patina-sdk-modules.md
+++ b/docs/src/rfc/text/0008-consolidate-sdk-crates-into-patina-sdk-modules.md
@@ -1,0 +1,134 @@
+# RFC: Consolidate SDK Crates into `patina_sdk` and `patina_sdk_macro` as Modules
+
+This RFC proposes consolidating the existing crates inside the Patina `sdk`
+directory into the `patina_sdk` and `patina_sdk_macro` crates, organized by
+modules. No functional changes to the modules are intended.
+
+## Change Log
+
+- 2025-05-20: Initial draft of the RFC.
+- 2025-05-22: Final updates during FCP
+  - `patina_test_macro` will be merged into `patina_sdk_macro`, but
+  `patina_sdk_macro` will remain a separate crate from `patina_sdk` because
+  procedural macros must be defined in a separate crate.
+  - Rename `0000-consolidate-sdk-crates-into-patina-sdk-modules` to
+    `0008-consolidate-sdk-crates-into-patina-sdk-modules`
+
+## Motivation
+
+Following the creation of the Patina monorepo, and in line with the [Crates
+Renaming RFC](0007-patina-crates-naming-and-categorization.md), there is now a
+scope to consolidate the crates located in the `sdk` directory. There was
+[consensus](https://github.com/OpenDevicePartnership/patina/pull/408#discussion_r2089478969)
+on merging some of these crates (e.g., runtime services, UEFI protocol) into
+`patina_sdk`, organized via a module hierarchy. This RFC formalizes and
+continues that discussion.
+
+## Technology Background
+
+This RFC does not affect any specific technologies or alter the functionality of
+the modules.
+
+## Goals
+
+1. Reduce the number of standalone crates present in the `sdk` directory.
+2. Reduce the number of standalone crates published to [crates.io](https://crates.io).
+3. Preserve the current crate or module boundaries.
+4. Consolidate the crates into `patina_sdk` and `patina_sdk_macro`.
+5. Leverage Rust's default [module visibility
+   rules](https://doc.rust-lang.org/reference/visibility-and-privacy.html),
+   using `pub(crate)` and similar mechanisms to control visibility both inside
+   and outside the `patina_sdk` crate.
+6. Consolidate the following crates, which are already being re-exported in
+   `patina_sdk` via `pub use`, as modules within the crate:
+   - `patina_boot_services`
+   - `patina_driver_binding`
+   - `patina_runtime_services`
+   - `patina_tpl_mutex`
+   - `patina_uefi_protocol`
+7. Consolidate the following crates into `patina_sdk_macro`, as modules
+   within the crate:
+   - `patina_test_macro`
+
+## Requirements
+
+1. Preserve the current crate or module boundaries.
+2. Do not introduce any changes to the logic or functionality of the modules,
+   other than updating import paths.
+3. Update all references in the codebase to reflect the new module paths after
+   reorganization.
+
+## Unresolved Questions
+
+## Prior Art: Crate/Module Hierarchy Before and After Consolidation
+
+Below is a comparison of the current and proposed crate/module hierarchies. Some
+of the crates being consolidated were previously re-exported in `patina_sdk` and
+are now integrated directly as modules.
+
+```text
+Before                             After
+C:\r\patina\sdk>                    C:\r\patina\sdk>
+├── patina_boot_services            │
+│   ├── allocation.rs               │
+│   ├── boxed.rs                    │
+│   ├── c_ptr.rs                    │
+│   ├── event.rs                    │
+│   ├── global_allocator.rs         │
+│   ├── lib.rs                      │
+│   ├── protocol_handler.rs         │
+│   └── tpl.rs                      │
+├── patina_driver_binding           │
+│   └── lib.rs                      │
+├── patina_runtime_services         │
+│   ├── lib.rs                      │
+│   └── variable_services.rs        │
+├── patina_sdk                      ├── patina_sdk
+│   ├── base.rs                     │   ├── base.rs
+│   │                               │   ├── boot_services/*
+│   │                               │   ├── boot_services.rs    (was `patina_boot_services/lib.rs`)
+│   ├── component/*                 │   ├── component/*
+│   ├── component.rs                │   ├── component.rs
+│   │                               │   ├── driver_binding/*
+│   │                               │   ├── driver_binding.rs   (was `patina_driver_binding/lib.rs`)
+│   ├── efi_types.rs                │   ├── efi_types.rs
+│   ├── error.rs                    │   ├── error.rs
+│   ├── guid.rs                     │   ├── guid.rs
+│   ├── lib.rs                      │   ├── lib.rs              (`pub use` becomes `pub mod`)
+│   ├── log/*                       │   ├── log/*
+│   ├── log.rs                      │   ├── log.rs              (was reexporting `boot_services/driver_binding/runtime_services/uefi_protocol`)
+│   ├── macros.rs                   │   ├── macros.rs
+│   │                               │   ├── runtime_services/*
+│   │                               │   ├── runtime_services.rs (was `patina_runtime_services/lib.rs`)
+│   ├── serial/*                    │   ├── serial/*
+│   └── serial.rs                   │   ├── serial.rs
+│                                   │   ├── tpl_mutex/*
+│                                   │   ├── tpl_mutex.rs        (was `patina_tpl_mutex/lib.rs`)
+│                                   │   ├── test/*
+│                                   │   ├── test.rs             (was `patina_test/lib.rs`)
+│                                   │   ├── uefi_protocol/*
+│                                   │   └── uefi_protocol.rs    (was `patina_uefi_protocol/lib.rs`)
+├── patina_sdk_macro                └── patina_sdk_macro
+│   ├── component_macro.rs              ├── component_macro.rs
+│   ├── hob_macro.rs                    ├── hob_macro.rs
+│   ├── lib.rs                          ├── lib.rs
+│   └── service_macro.rs                ├── service_macro.rs
+│                                       ├── test_macro/*
+│                                       └── test_macro.rs       (was `patina_test_macro/lib.rs`)
+├── patina_test
+│   ├── __private_api.rs
+│   └── lib.rs
+├── patina_test_macro
+│   └── lib.rs
+├── patina_tpl_mutex
+│   └── lib.rs
+└── patina_uefi_protocol
+    ├── lib.rs
+    └── status_code.rs
+```
+
+## Alternatives
+
+- Keep the current crate organization
+  - This option is not recommended, as it does not achieve the goals outlined in
+    this RFC.
--- a/patina_dxe_core/src/event_db.rs
+++ b/patina_dxe_core/src/event_db.rs
@@ -500,24 +500,6 @@ impl EventDb {
     }
 }
 
-struct EventNotificationIterator {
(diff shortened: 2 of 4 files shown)


kouchekiniad:
Rebased to latest

patina_dxe_core/src/allocator/fixed_size_block_allocator.rs
```diff
@@ -135,95 +158,80 @@ pub type PageChangeCallback = fn(&mut FixedSizeBlockAllocator);
...
-                size,
-                self.handle,
-                None,
-            )
-            .map_err(|_| FixedSizeBlockAllocatorError::OutOfMemory)?;
+    /// Expand the memory available to this allocator with a new contiguous region of memory, setting up a new allocator
+    /// node to manage this range.
+    pub fn expand(&mut self, new_region: NonNull<[u8]>) -> core::result::Result<(), FixedSizeBlockAllocatorError> {
+        // Ensure we're expanding enough to fit a new allocator list node
+        if new_region.len() < MIN_EXPANSION || new_region.len() <= size_of::<AllocatorListNode>() {
```
joschock: I could go either way, but consider leaving out the `new_region.len() < MIN_EXPANSION` condition here. It's not technically a requirement of the allocator, more a policy choice. I think it might be better to just enforce this as a policy choice in the caller than mandate that it must be at least this big in the receiver.

patina_dxe_core/src/allocator/fixed_size_block_allocator.rs
```diff
@@ -234,40 +242,43 @@ impl FixedSizeBlockAllocator {
...
-        //attempt to expand and then allocate again
-        if self.expand(layout).is_err() {
-            return ptr::null_mut();
-        }
-        self.fallback_alloc(layout)
+
+        // Determine how much additional memory is required
+        let additional_mem_required =
+            layout.pad_to_align().size() + Layout::new::<AllocatorListNode>().pad_to_align().size();
+        let additional_mem_required = max(additional_mem_required, MIN_EXPANSION);
```
joschock: again, I could go either way, but consider only handling the `MIN_EXPANSION` aspect as policy choice at the SpinlockedFsbLayer and removing it from the internal FSB layer.

joschock: same goes for the ALIGNMENT aspect. The FSB doesn't really need to know or care about either MIN_EXPANSION or ALIGNMENT as a matter of technical correctness of the FSB itself. Those are both kinda GCD things that could be handled by the outside caller now that we are decoupling things. One could imagine some weird alternate reality where you wanted an FSB with a smaller expansion or tigher/looser alignment requirements, and it would be nice to be able to handle that in the caller.

patina_dxe_core/src/allocator/fixed_size_block_allocator.rs
```diff
@@ -276,64 +287,49 @@ impl FixedSizeBlockAllocator {
...
                     }
                 }
             }
             None => self.fallback_alloc(layout),
+        };
+
+        if let Err(FixedSizeBlockAllocatorError::OutOfMemory(_)) = allocation {
+            // The spin-locked FSB will try again. Decrement the page_allocation_calls so that these two calls
+            // only reflect the one allocation.
+            self.stats.pool_allocation_calls -= 1;
```
joschock: The pool_allocation_call is pretty flexible; I'm okay if it bumps twice for a double call on expand if that makes things simpler. The reserved-page tracking has a much stronger need to be accurate since that drives memory type info implementation, but the allocation call counts are a nice-to-have a could be eliminated entirely without affecting other stuff if needed.

patina_dxe_core/src/allocator/fixed_size_block_allocator.rs
```diff
@@ -572,8 +468,11 @@ impl Display for FixedSizeBlockAllocator {
 
 /// Spin Locked Fixed Size Block Allocator
 ///
-/// A wrapper for [`FixedSizeBlockAllocator`] that provides Sync/Send via means of a spin mutex.
+/// A wrapper for [`FixedSizeBlockAllocator`] that allocated additional memory as needed from a GCD
```
joschock: nit:
```suggestion
/// A wrapper for [`FixedSizeBlockAllocator`] that allocates additional memory as needed from a GCD
```

patina_dxe_core/src/allocator/fixed_size_block_allocator.rs
```diff
@@ -658,7 +627,28 @@ impl SpinLockedFixedSizeBlockAllocator {
     /// This routine will return Err(efi::Status::ALREADY_STARTED) if it is called more than once.
     ///
     pub fn reserve_memory_pages(&self, pages: usize) -> Result<(), EfiError> {
-        self.lock().reserve_memory_pages(pages)
+        if self.lock().reserved_range.is_some() {
+            Err(EfiError::AlreadyStarted)?;
+        }
+
+        let reserved_block_len = pages << UEFI_PAGE_SHIFT;
```
joschock: consider using the pages_to_size macro

patina_dxe_core/src/allocator/fixed_size_block_allocator.rs
kouchekiniad: Sounds good, though I will at least need to check for alignment with AllocatorListNode for the `ptr.write()`.

patina_dxe_core/src/allocator/fixed_size_block_allocator.rs
kouchekiniad: Sounds reasonable to me

patina_dxe_core/src/allocator/fixed_size_block_allocator.rs
kouchekiniad: I'll probably just leave the double-call then since it's still accurate for the inner FSB to say alloc was called twice.

patina_dxe_core/src/allocator/fixed_size_block_allocator.rs
```diff
@@ -33,7 +36,11 @@ use r_efi::efi;
 #[derive(Debug, PartialEq)]
 pub enum FixedSizeBlockAllocatorError {
     /// Could not satisfy allocation request, and expansion failed.
-    OutOfMemory,
+    OutOfMemory(usize),
```
os-d, line 44: nit: update the comment to indicate what the usize in the return is

patina_dxe_core/src/allocator/fixed_size_block_allocator.rs
kouchekiniad: This check has been removed in my latest changeset

patina_dxe_core/src/allocator/fixed_size_block_allocator.rs
kouchekiniad: Addressed in latest changeset.

patina_dxe_core/src/allocator/fixed_size_block_allocator.rs
kouchekiniad: Removed in latest changeset.

patina_dxe_core/src/allocator/fixed_size_block_allocator.rs
kouchekiniad: Fixed.

patina_dxe_core/src/allocator/fixed_size_block_allocator.rs
kouchekiniad: Changed in my latest changeset.

patina_dxe_core/src/allocator/fixed_size_block_allocator.rs
```diff
@@ -1456,7 +1610,7 @@ mod tests {
     }
```
os-d, line 1612: I may have missed it, but did you update one of the tests to specifically test the re-entrancy case that this is fixing?

patina_dxe_core/src/allocator/fixed_size_block_allocator.rs
```diff
@@ -14,26 +14,36 @@ extern crate alloc;
 use super::{AllocationStrategy, DEFAULT_ALLOCATION_STRATEGY};
 
 use crate::{gcd::SpinLockedGcd, tpl_lock};
-use patina_sdk::{base::UEFI_PAGE_SIZE, error::EfiError};
+
+use patina_sdk::{error::EfiError, patina_boot_services::c_ptr::CMutPtr};
```
os-d: Is the CMutPtr actually used? I'm not certain if we want the core to be consuming from patina_boot_services either.

patina_dxe_core/src/allocator/uefi_allocator.rs
```diff
@@ -536,42 +536,42 @@ mod tests {
...
             drop(allocator);
 
             let allocator = unreserved_allocator.allocator.lock();
-            assert!(allocator.preferred_range.is_none());
+            assert!(allocator.reserved_range.is_none());
             drop(allocator);
 
-            println!("preferred range: {:#x?}", preferred_range);
-            //verify that the first 0x100 pages from the reserved allocator are in the preferred_range, and that allocating
+            println!("preferred range: {:#x?}", reserved_range);
```
os-d:
```suggestion
            println!("reserved range: {:#x?}", reserved_range);
```

patina_dxe_core/src/allocator/fixed_size_block_allocator.rs
kouchekiniad: It was being used, but it looks like it's unnecessary. I've removed it in the latest changeset.

patina_dxe_core/src/allocator/uefi_allocator.rs
kouchekiniad: Fixed, and updated the comments in the same file.

patina_dxe_core/src/allocator/fixed_size_block_allocator.rs
kouchekiniad: Added in latest changeset.

patina_dxe_core/src/allocator/fixed_size_block_allocator.rs
kouchekiniad: In my mind the fact that FixedSizeBlockAllocator doesn't have access to a GCD implies there is no possibility of a FSB re-entrancy. I could add a test that takes the GCD lock then executes functionality on the FSB, but I'm not sure what good that'd do.

There is still the possibility of a GCD re-entrancy if an FSB expand() occurs when there's not enough memory for even a pending_notifies BTree node, which would then trigger another expand() which would call into the GCD again. I'm planning on addressing that in another PR as this PR is already quite large. To fix that issue I wouldn't be able to do the same thing I did here as instead of running into a re-entrancy we'd hit an infinite loop. To fix the issue I'm working on two possible solutions:
1. Create a separate no_global_alloc_allocate_memory_space for use by expand() which would be followed up by manually triggering a cache changed event + memory map changed event after the expansion is complete
2. Always keep an extra page of memory available in the FSB as a reserve that'd be able to fit the two extra BTree nodes that would be created

os-d: I meant a test from the SpinLockedFixedSizeBlockAllocator side, since it holds the lock (i.e. a unit test confirming the patch works and we don't regress).

I would tend to go with the second option you list, to keep scratchpad memory available for the FSB itself rather than create new APIs.

kouchekiniad: In order to create a test case that tests the reentrancy I'd need to set the global allocator to a UEFI allocator for a portion of the test. @os-d do you happen to know of any way I'd be able to achieve this? From what I can find `#[global_allocator]` can only be set globally.

os-d: If this is not reasonable to test, we don't need to. It was more if we can, we should.

kouchekiniad: Sounds good, I'll look more into the feasibility of testing this with my next planned PR.

PR #439, opened 2025-05-20
mm_communicator: Remove comm buffer trace message

makubacki:
## Description

Some platforms always have tracing on and this is noisy. Can be viewed in debugger if needed.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- `cargo make all`

## Integration Instructions

- Comm buffer dump no longer present in trace level by default


makubacki:
> The other way to do this would be to have a trace target and let a platform enable/disable based on the target. I don't mind either way.

Too much overhead for this. I think we should also not have Q35 at trace level by default if we're concerned about log verbosity in the default build.

PR #447, opened 2025-05-22
[REBASE & FF] Image Loader: Remove Runtime Image Memory Protections on EBS

os-d:
## Description

This PR consists of three commits:

Dispatcher: Reduce Depex Evaluation Message to Trace
--
One of the noisest logs in the patina boot currently is the log that says a depex guid was evaluated, which in general does not provide useful information to an average build. This commit moves that to a trace log.

GCD: Update Compatibility Mode to Allow For Unaligned Loader Allocations
--
In the case of loader code and data allocations having been made before compatibility mode is activated, the current GCD code does not ensure that these returned ranges are page_aligned (they may be missing the allocator node length), so this updates the logic to apply compatibility mode to page aligned lengths and addresses.

Image Loader: Remove Runtime Image Memory Protections on EBS
--
edk2 follows a pattern where runtime image memory protections are removed on ExitBootServices to allow the SetVirtualAddressMap callback to relocate runtime images. Patina was not following this, which only matters for certain flavors of Linux, it could cause the core to attempt to write to RO memory.

For now, Patina follows the same method as edk2, but investigation is happening to determine if we should have a different method.

- [x] Impacts functionality?
- [x] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested booting TinyLinux on SBSA.

## Integration Instructions

N/A.

### DIFF
--- a/patina_dxe_core/src/image.rs
+++ b/patina_dxe_core/src/image.rs
@@ -15,14 +15,16 @@ use patina_performance::{
     create_performance_measurement, perf_image_start_begin, perf_image_start_end, perf_load_image_begin,
     perf_load_image_end,
 };
-use patina_sdk::base::{align_up, UEFI_PAGE_SIZE};
+use patina_sdk::base::{align_up, DEFAULT_CACHE_ATTR, UEFI_PAGE_SIZE};
 use patina_sdk::error::EfiError;
-use patina_sdk::{guid, uefi_size_to_pages};
+use patina_sdk::{guid, uefi_pages_to_size, uefi_size_to_pages};
 use r_efi::efi;
 
+use crate::dxe_services::core_set_memory_space_attributes;
 use crate::{
     allocator::{core_allocate_pages, core_free_pages},
     dxe_services,
+    events::EVENT_DB,
     filesystems::SimpleFile,
     pecoff::{self, relocation::RelocationBlock, UefiPeInfo},
     protocol_db,
@@ -701,7 +703,7 @@ fn activate_compatibility_mode(private_info: &PrivateImageData) -> Result<(), Ef
     // for this image map all mem RWX preserving cache attributes if we find them
     let stripped_attrs = dxe_services::core_get_memory_space_descriptor(private_info.image_base_page)
         .map(|desc| desc.attributes & efi::CACHE_ATTRIBUTE_MASK)
-        .unwrap_or(0);
+        .unwrap_or(DEFAULT_CACHE_ATTR);
     if dxe_services::core_set_memory_space_attributes(
         private_info.image_base_page,
         patina_sdk::uefi_pages_to_size!(private_info.image_num_pages) as u64,
@@ -726,6 +728,40 @@ fn activate_compatibility_mode(private_info: &PrivateImageData) -> Result<(), Ef
     log::error!("Attempting to load {} that is not NX compatible. Compatibility mode is not allowed in this build, not loading image.",
                 private_info.pe_info.filename.clone().unwrap_or(String::from("Unknown")));
     Err(EfiError::LoadError)
+}
+
+extern "efiapi" fn runtime_image_protection_fixup_ebs(event: efi::Event, _context: *mut c_void) {
+    let mut private_data = PRIVATE_IMAGE_DATA.lock();
+
+    for image in private_data.private_image_data.values_mut() {
+        if image.pe_info.image_type == EFI_IMAGE_SUBSYSTEM_EFI_RUNTIME_DRIVER {
+            let cache_attrs = dxe_services::core_get_memory_space_descriptor(image.image_base_page)
+                .map(|desc| desc.attributes & efi::CACHE_ATTRIBUTE_MASK)
+                .unwrap_or(DEFAULT_CACHE_ATTR);
+
+            match core_set_memory_space_attributes(
+                image.image_base_page,
+                uefi_pages_to_size!(image.image_num_pages) as u64,
+                cache_attrs,
+            ) {
+                Ok(_) => {
+                    // success, keep going
+                }
+                Err(status) => {
+                    log::error!(
+                        "Failed to set GCD attributes for runtime image {:#X?} with Status {:#X?}, may fail to relocate",
+                        image.image_base_page,
+                        status
+                    );
+                    debug_assert!(false);
+                }
+            };
+        }
+    }
+
+    if let Err(status) = EVENT_DB.close_event(event) {
+        log::error!("Failed to close image EBS event with status {:#X?}. This should be okay.", status);
+    }
 }
 
 // Reads an image buffer using simple file system or load file protocols.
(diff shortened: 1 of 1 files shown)


patina_dxe_core/src/gcd.rs
```diff
@@ -282,15 +285,25 @@ pub(crate) fn activate_compatibility_mode() {
...
                         _ => descriptor.length as usize,
                     };
-                    if GCD.set_memory_space_attributes(addr as usize, len, attributes).is_err() {
+
+                    // We need to ensure we are operating on a page aligned addresses and lengths, as the image(s) that
+                    // were allocated here may not be page aligned. We don't share pools across types, so this is safe.
+                    if GCD
+                        .set_memory_space_attributes(
+                            (addr & !(UEFI_PAGE_MASK as u64)) as usize,
+                            align_up(len as u64, UEFI_PAGE_SIZE as u64).unwrap_or(UEFI_PAGE_SIZE as u64) as usize,
```
cfernald: We are aligning the addr down and the len up, but not relative to eachother, so this could cause dropped space at the end, right?

e.g. addr = 0x1800, len = 0x1000 -> set on 0x1000 - 0x2000, which drops the 0x2000 - 0x2800 from the original range.

os-d: Yep, good catch, thanks!

os-d: Addressed in latest.

patina_dxe_core/src/image.rs
```diff
@@ -728,6 +730,41 @@ fn activate_compatibility_mode(private_info: &PrivateImageData) -> Result<(), Ef
...
+    for image in private_data.private_image_data.values_mut() {
+        let image_base = image.image_info.image_base as u64 & !UEFI_PAGE_MASK as u64;
+        if image.pe_info.image_type == EFI_IMAGE_SUBSYSTEM_EFI_RUNTIME_DRIVER {
+            let cache_attrs = dxe_services::core_get_memory_space_descriptor(image_base)
+                .map(|desc| desc.attributes & efi::CACHE_ATTRIBUTE_MASK)
+                .unwrap_or(efi::MEMORY_WB);
+
+            match core_set_memory_space_attributes(
+                image_base,
+                align_up(image.image_info.image_size, UEFI_PAGE_SIZE as u64).unwrap_or(UEFI_PAGE_SIZE as u64),
```
cfernald: I think this has the same alignment issue with the size being relative to the aligned-down address. This feels like something it would be useful to have a function for. 

i.e. `uefi_sdk::base::align_base_offset(base: u64, length: u64, alignment: u64) -> (u64, u64)`

Though don't feel the need to do that as part of this PR.

os-d: Yep, totally valid. I'll add that.

patina_dxe_core/src/image.rs
```diff
@@ -728,6 +730,41 @@ fn activate_compatibility_mode(private_info: &PrivateImageData) -> Result<(), Ef
     Err(EfiError::LoadError)
 }
 
+extern "efiapi" fn runtime_image_protection_fixup_ebs(event: efi::Event, _context: *mut c_void) {
+    let mut private_data = PRIVATE_IMAGE_DATA.lock();
+
+    for image in private_data.private_image_data.values_mut() {
+        let image_base = image.image_info.image_base as u64 & !UEFI_PAGE_MASK as u64;
```
joschock: I think this is guaranteed by load_image - is there any way that an image could have a non-page aligned image_base? If that's never legit, consider an assert instead of silent fix?

patina_dxe_core/src/dispatcher.rs
```diff
@@ -166,7 +166,7 @@ fn dispatch() -> Result<bool, EfiError> {
         let driver_candidates: Vec<_> = dispatcher.pending_drivers.drain(..).collect();
         let mut scheduled_driver_candidates = Vec::new();
         for mut candidate in driver_candidates {
-            log::info!("Evaluting depex for candidate: {:?}", guid_fmt!(candidate.file_name));
+            log::trace!("Evaluting depex for candidate: {:?}", guid_fmt!(candidate.file_name));
```
makubacki: Can you please fix this while updating the line?

```suggestion
            log::trace!("Evaluating depex for candidate: {:?}", guid_fmt!(candidate.file_name));
```

patina_dxe_core/src/gcd.rs
```diff
@@ -282,19 +285,41 @@ pub(crate) fn activate_compatibility_mode() {
...
                     len = match descriptor.base_address + descriptor.length {
-                        end if end > range.end => (range.end - addr) as usize,
-                        _ => descriptor.length as usize,
+                        end if end > range.end => range.end - addr,
+                        _ => descriptor.length,
                     };
-                    if GCD.set_memory_space_attributes(addr as usize, len, attributes).is_err() {
+
+                    // We need to ensure we are operating on a page aligned addresses and lengths, as the image(s) that
+                    // were allocated here may not be page aligned. We don't share pools across types, so this is safe.
```
makubacki:
```suggestion
                    // We need to ensure we are operating on page aligned addresses and lengths, as the image(s) that
                    // were allocated here may not be page aligned. We don't share pools across types, so this is safe.
```

patina_dxe_core/src/image.rs
```diff
@@ -728,6 +730,41 @@ fn activate_compatibility_mode(private_info: &PrivateImageData) -> Result<(), Ef
...
+        if image.pe_info.image_type == EFI_IMAGE_SUBSYSTEM_EFI_RUNTIME_DRIVER {
+            let cache_attrs = dxe_services::core_get_memory_space_descriptor(image_base)
+                .map(|desc| desc.attributes & efi::CACHE_ATTRIBUTE_MASK)
+                .unwrap_or(efi::MEMORY_WB);
```
makubacki: Does this C code also default to WB (I think that's reasonable)? Does other Rust code do this? If so, it might be nice to define in one place in a macro.

os-d: It is actually worse in edk2, it sets it to 0: https://github.com/tianocore/edk2/blob/11f1d28bcb8553af9c689ec121c191a258e6f01a/MdeModulePkg/Core/Dxe/Misc/MemoryProtection.c#L979C6-L979C36, which is the lack of cacheability attributes, so there is an ambiguity on what is actually set. Chatted with Chris and Kun on this, we think WB is a reasonable default for Patina all up, so I will make this a const.

patina_dxe_core/src/image.rs
os-d: Correct, but I'm using the wrong field here, this image_base may not be page aligned, but image_base_page will be. Switching to that.

patina_dxe_core/src/image.rs
os-d: I added the API, switching the other usage but I actually realized when resolving John's commit that I already have the image base page and number of pages in the private image data, so this calculation isn't needed here.

os-d:
@makubacki @joschock @cfernald I have made the requested changes and retested on Windows and Linux on arm64 and x64 and all are booting on qemu. Can you please re-review?

PR #448, opened 2025-05-22
patina_sdk:memory: Implement mock MemoryManager

Javagedes:
## Description

Implements a working MemoryManager implementation for the MemoryManager trait by using the global allocator. This should be used when a mock is needed for testing purposes, that can allocate and deallocate memory as expected.

This commit also adds mockall's automock to the MemoryManager trait, to simplify mocking MemoryManager interface method failures in tests.

Overall this provides two mocks for MemoryManager automock's `MockMemoryManager` and a custom working MemoryManager, `StdMemoryManager`.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [x] Includes documentation?

## How This Was Tested

Compiles

## Integration Instructions

N/A

### DIFF
--- a/sdk/patina_sdk/src/component/service/memory.rs
+++ b/sdk/patina_sdk/src/component/service/memory.rs
@@ -3,6 +3,26 @@
 //! This module contains traits and types for services related memory management
 //! and access for use in services. See [MemoryManager] for the primary interface.
 //!
+//! ## Testing
+//!
+//! This module contains a std implementation of the [MemoryManager] trait called `StdMemoryManager` to enable testing.
+//! It provides a fully working implementation of the [MemoryManager] based off of the std global allocator.
+//!
+//! ```rust
+//! use patina_sdk::component::service::{ Service, memory::*};
+//!
+//! use std::boxed::Box;
+//!
+//! fn test_that_needs_memory_manager() {
+//!     let memory_manager = StdMemoryManager::new();
+//!
+//!     let service: Service<dyn MemoryManager> = Service::mock(Box::new(memory_manager));
+//! }
+//! ```
+//!
+//! As always, if you need a fallable mock for testing your code's ability to handle allocation errors, A `mockall` mock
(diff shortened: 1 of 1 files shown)


berlin-with0ut-return:
does this cover the case where we need to mock the actual return value of MemoryManager? ie

```
mock_memory_manager.expect_allocate_pages().returning(|| PageAllocation { ... } )
```

as of now i don't think i need this use case but isn't the idea of mocking that you can "force" a return value? or is that getting too deep into the mock abstraction

Javagedes:
> does this cover the case where we need to mock the actual return value of MemoryManager? ie
> 
> ```
> mock_memory_manager.expect_allocate_pages().returning(|| PageAllocation { ... } )
> ```
> 
> as of now i don't think i need this use case but isn't the idea of mocking that you can "force" a return value? or is that getting too deep into the mock abstraction

You can do the following:

``` rust
let mut mock_memory_manager = MockMemoryManager::new();
let mut ret = MockMemoryManager::new();
ret.expect_free_pages().returning(|address, page_count| return Err(MemoryError::InternalError));
mock_memory_manager.expect_allocate_pages().returning(|a, b, c| PageAllocation::new(5, 5, Box::leak(Box::new(ret)) );
```

sdk/patina_sdk/src/component/service/memory.rs
```diff
@@ -18,12 +38,16 @@ use core::{alloc::Allocator, ptr::NonNull};
 use alloc::boxed::Box;
 use r_efi::efi;
 
+#[cfg(any(test, feature = "mockall"))]
```
cfernald, line 41: why any? without the mockall feature pulling in the dependency this would fail, right?

Javagedes: mockall is in the `[dev-dependencies]` section, which makes it available when running tests, regardless of if you enable the mockall feature or not.

Issue #418, opened 2025-05-14
[Feature]: Merge some of the crates with in the patina/src/sdk/* into patina/src/sdk/uefi_sdk crate


vineelko:
### Feature Overview

This is in reference to the discussion that happened on below RFC.

RFC: Categorizing and Renaming Patina Crates for Consistency
https://github.com/OpenDevicePartnership/patina/pull/408#discussion_r2080688749

There is consensus on merging some crates within sdk directory (e.g., runtime services, UEFI protocol) into uefi_sdk`, organized by module hierarchy.

### Solution Overview

More details below, 
RFC: Categorizing and Renaming Patina Crates for Consistency
https://github.com/OpenDevicePartnership/patina/pull/408#discussion_r2080688749

### Urgency

Medium

### Are you going to implement the feature request?

I will implement the feature

### Do you need maintainer feedback?

No maintainer feedback needed

PR #437, opened 2025-05-20
RFC: Consolidate SDK Crates into `patina_sdk` and `patina_sdk_macro` as Modules

vineelko:
## Description

This RFC proposes consolidating the existing crates inside the Patina `sdk`
directory into the `patina_sdk` and `patina_sdk_macro` crates, organized by
modules. No functional changes to the modules are intended.

Tracks: #418 

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

NA

## Integration Instructions

NA

PR #449, opened 2025-05-23
RFC Implementation: Consolidate SDK Crates into patina_sdk and patina_sdk_macro as Modules

vineelko:
## Description

This PR implements the RFC(#437) consolidating the existing crates inside the
Patina `sdk` directory into the `patina_sdk` and `patina_sdk_macro` crates,
organized by modules. No functional changes to the modules are intended.

Closes: #418, #437

- [ ] Impacts functionality?
- [ ] Impacts security?
- [x] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested
Validated on Q35/SBSA

## Integration Instructions
The crates inside `sdk` have been consolidated into `patina_sdk` and `patina_sdk_macro`. Any new code should adhere to the module hierarchy below.

patina_sdk


docs/src/dxe_core/protocol_database.md
```diff
@@ -6,7 +6,7 @@ specification. This section assumes basic familiarity with the protocol services
 focuses on how they are implemented in the Rust DXE Core.
 
 The main implementation for protocol services in the Rust DXE Core resides in the protocol database object implemented
-by the patina_uefi_protocol crate. The protocol database implemented by that crate is used to implement the UEFI protocol
+by the `patina_dxe_core` crate. The protocol database implemented by that crate is used to implement the UEFI protocol
 handler services by the main Rust DXE Core protocol.rs module.
```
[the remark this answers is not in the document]
vineelko: _Copilot's review comments behave like review comments from humans. You can add reactions to them, comment on them, resolve them and hide them._

**_Any comments you add to Copilot's review comments will be visible to humans, but they won't be visible to Copilot, and Copilot won't reply._**

My dear Copilot, it's unfortunate that you cannot respond to my comments. 😢

cfernald:
@vineelko , Should we mark RFCs as breaking separate of their implementation? It would cause a major version rev without actually having breaking changes until the associated code PR happens.

vineelko:
> @vineelko , Should we mark RFCs as breaking separate of their implementation? It would cause a major version rev without actually having breaking changes until the associated code PR happens.

My understanding of a breaking change is that it only applies when we actually check in something that breaks existing code. From that perspective, I think it's more appropriate to mark the implementation as breaking, rather than the RFC itself. Bumping the major version just for an RFC without any code changes feels unnecessary in my opinion.

components/patina_adv_logger/src/integration_test.rs
```diff
@@ -10,14 +10,15 @@
 //!
 //! SPDX-License-Identifier: BSD-2-Clause-Patent
 //!
-use patina_sdk::patina_boot_services::{BootServices, StandardBootServices};
-use patina_test::{u_assert, u_assert_eq, uefi_test};
+use patina_sdk::boot_services::{BootServices, StandardBootServices};
+use patina_sdk::test::uefi_test;
```
Javagedes: Apologies, I thought as a part of the crate renaming, `uefi_test` macro was being turned into `patina_test`. 

I know this is a bit outside of the scope of this PR, but can you rename `uefi_test` to `patina_test`? If not, I'd like to get a PR up to do exactly that before we actually do a release with this change, to prevent another almost immediate major version change.

vineelko: No worries Joey will update it as part of this PR.

Javagedes: Thank you very much :)

docs/src/rfc/text/0007-patina-crates-naming-and-categorization.md
```diff
@@ -102,7 +102,7 @@ RFC.
     ├── uefi_protocol
     ├── uefi_sdk
     ├── uefi_sdk_macro
-    ├── uefi_test
+    ├── patina_test
```
Javagedes: This particular section is prior crate organization, so it should stay as uefi_test

sdk/patina_sdk_macro/src/test_macro.rs
```diff
@@ -18,13 +18,9 @@ const KEY_SHOULD_FAIL: &str = "should_fail";
...
 const KEY_SKIP: &str = "skip";
 
-#[proc_macro_attribute]
-pub fn uefi_test(_: proc_macro::TokenStream, item: proc_macro::TokenStream) -> proc_macro::TokenStream {
-    uefi_test2(item.into()).into()
-}
-
-fn uefi_test2(stream: proc_macro2::TokenStream) -> proc_macro2::TokenStream {
-    let mut item = syn::parse2::<ItemFn>(stream).expect("The #[uefi_test] attribute can only be applied to functions");
+pub fn uefi_test2(stream: proc_macro2::TokenStream) -> proc_macro2::TokenStream {
```
Javagedes: rename this to `patina_test2` please. It probably comes with some changes to the unit tests also

PR #453, opened 2025-05-23
Remove TPL lock from uninstall_multiple_protocol_interfaces

joschock:
## Description

This removes the TPL lock from `uninstall_multiple_protocol_interfaces`. This brings the behavior of this function with respect to TPL in line with EDK2 C reference behavior. 

Acquiring the lock in unisntall_multiple_protocol_interfaces can lead to issues with drivers that are coded against the EDK2 behavior. 

An example of such an issue is in the FAT driver (https://github.com/tianocore/edk2/tree/master/FatPkg/EnhancedFatDxe) which attempts to acquire its own lock at TPL_CALLBACK as part of DriverBindingStop() functionality. This will trigger a TPL inversion if `uninstall_multiple-protocol_interfaces` is used to remove a BlockIo instance that is being used by the FAT driver - this will cause the FAT driver to be stopped while the TPL is NOTIFY and when it tries to acquire its own lock at TPL_CALLBACK a TPL inversion occurs. 

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Functional testing with USB hot remove - prior to this change, USB hot remove will cause a TPL assert in FAT driver binding as described above. After this change, TPL assert in FAT driver no longer occurs. 

## Integration Instructions

N/A

### DIFF
--- a/patina_dxe_core/src/protocols.rs
+++ b/patina_dxe_core/src/protocols.rs
@@ -516,10 +516,6 @@ unsafe extern "C" fn install_multiple_protocol_interfaces(handle: *mut efi::Hand
 }
 
 unsafe extern "C" fn uninstall_multiple_protocol_interfaces(handle: efi::Handle, mut args: ...) -> efi::Status {
-    // See note in install_multiple_protocol_interfaces.
-    let tpl_mutex = TplMutex::new(efi::TPL_NOTIFY, (), "atomic_protocol_uninstall");
-    let _tpl_guard = tpl_mutex.lock();
-
     if handle.is_null() {
         return efi::Status::INVALID_PARAMETER;
     }


joschock:
> Looks like we don't try to take the lock on `uninstall_protocol_interface` and I didn't immediately see where any containing code called a TPLMutex, but there is that danger, right, that something could take a TPLMutex in this code in the future? Or was it specifically going to notify that was the problem?

I don't think the lock is needed. The underlying data structures (i.e. the protocol database) will take locks as needed for internal protection. The C core takes a lock on install_protocol_interface - but I think that is to avoid notifying any callbacks until all the interfaces are installed. The C core doesn't take a lock on uninstall_protocol_interface, but that's because there is no `register_uninstall_protocol_interface_notify`. I was overzealously looking for symmetry between uninstall/install, but that was a misunderstanding on my part in the initial implementation.

os-d:
> I don't think the lock is needed. The underlying data structures (i.e. the protocol database) will take locks as needed for internal protection. The C core takes a lock on install_protocol_interface - but I think that is to avoid notifying any callbacks until all the interfaces are installed. The C core doesn't take a lock on uninstall_protocol_interface, but that's because there is no `register_uninstall_protocol_interface_notify`. I was overzealously looking for symmetry between uninstall/install, but that was a misunderstanding on my part in the initial implementation.

I agree the lock isn't needed, what I mean is that the problem was trying to enter TPL_NOTIFY while in TPL_CALLBACK, right? So would this be a problem if anything else in uninstall_protocol_interface took a TPLMutex? i.e. is this something we need to be sure doesn't happen?

joschock:
> > I don't think the lock is needed. The underlying data structures (i.e. the protocol database) will take locks as needed for internal protection. The C core takes a lock on install_protocol_interface - but I think that is to avoid notifying any callbacks until all the interfaces are installed. The C core doesn't take a lock on uninstall_protocol_interface, but that's because there is no `register_uninstall_protocol_interface_notify`. I was overzealously looking for symmetry between uninstall/install, but that was a misunderstanding on my part in the initial implementation.
> 
> I agree the lock isn't needed, what I mean is that the problem was trying to enter TPL_NOTIFY while in TPL_CALLBACK, right? So would this be a problem if anything else in uninstall_protocol_interface took a TPLMutex? i.e. is this something we need to be sure doesn't happen?

The issue is that the TplMutex I removed was raising to TPL_NOTIFY over the whole uninstall process. Whereas an internal lock in the protocol database only raises and restores the TPL locally - so not really visible to external code. 

The upshot is that prior to this fix, any side effects of the uninstall (e.g. DrvierBindingStop() due to the removal of an interface open_by_driver) would run at TPL_NOTIFY. Now those same side effects run at whatever TPL uninstall_multiple_protocol_interfaces() was executed at (typically TPL_CALLBACK or TPL_APPLICATION), which is in line with the C behavior.

os-d:
> > > I don't think the lock is needed. The underlying data structures (i.e. the protocol database) will take locks as needed for internal protection. The C core takes a lock on install_protocol_interface - but I think that is to avoid notifying any callbacks until all the interfaces are installed. The C core doesn't take a lock on uninstall_protocol_interface, but that's because there is no `register_uninstall_protocol_interface_notify`. I was overzealously looking for symmetry between uninstall/install, but that was a misunderstanding on my part in the initial implementation.
> > 
> > 
> > I agree the lock isn't needed, what I mean is that the problem was trying to enter TPL_NOTIFY while in TPL_CALLBACK, right? So would this be a problem if anything else in uninstall_protocol_interface took a TPLMutex? i.e. is this something we need to be sure doesn't happen?
> 
> The issue is that the TplMutex I removed was raising to TPL_NOTIFY over the whole uninstall process. Whereas an internal lock in the protocol database only raises and restores the TPL locally - so not really visible to external code.
> 
> The upshot is that prior to this fix, any side effects of the uninstall (e.g. DrvierBindingStop() due to the removal of an interface open_by_driver) would run at TPL_NOTIFY. Now those same side effects run at whatever TPL uninstall_multiple_protocol_interfaces() was executed at (typically TPL_CALLBACK or TPL_APPLICATION), which is in line with the C behavior.

Gotcha, thanks for the clarification. Makes sense.

PR #462, opened 2025-05-26
Fix issues with Uninstall and Close protocol caused by remove_protocol_usage behavior.

joschock:
## Description

Prior to this change, `remove_protocol_usage` was using retain, which removed all matching usages. This had the effect of prematurely removing some usages which changed the ordering/execution expectations of some calling code, causing certain flows involving DisconnectController to fail.

`remove_protocol_usage` now only removes a single instance, and instances are removed in last-in-first-out order. This matches EDK2 reference behavior for usage management and is also symmetric with `add_protocol_usage`. 

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Functional testing involving USB mass storage hot add/remove. Prior to this change, removing USB storage could cause ASSERTs in FAT and Partition drivers due to behavior described above violating assumptions in those drivers. After this change, USB mass storage hot add/remove now works as expected.

## Integration Instructions
N/A

### DIFF
--- a/.github/workflows/CiWorkflow.yml
+++ b/.github/workflows/CiWorkflow.yml
@@ -104,11 +104,10 @@ jobs:
         env:
           RUSTC_BOOTSTRAP: 1
 
-      # Commenting below to allow the PR# 425 to pass CI.
-      # - name: Crate Release Dry Run
-      #   run: cargo release --no-tag --workspace --no-push --allow-branch '*' --registry patina-fw
-      #   env:
-      #     RUSTC_BOOTSTRAP: 1
+      - name: Crate Release Dry Run
+        run: cargo release --no-tag --workspace --no-push --allow-branch '*' --registry patina-fw
+        env:
+          RUSTC_BOOTSTRAP: 1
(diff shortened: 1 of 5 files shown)


patina_dxe_core/src/protocols.rs
```diff
@@ -425,9 +425,13 @@ extern "efiapi" fn close_protocol(
         }
     };
 
-    match PROTOCOL_DB.remove_protocol_usage(handle, unsafe { *protocol }, Some(agent_handle), controller_handle) {
-        Err(err) => err.into(),
-        Ok(_) => efi::Status::SUCCESS,
+    // Remove all usagages matching the specified controller and agent handles.
```
makubacki:
```suggestion
    // Remove all usages matching the specified controller and agent handles.
```

joschock: Thanks, fixed in latest.

PR #464, opened 2025-05-27
[REBASE&FF] Set minimum rust version to 1.85.0

Javagedes:
## Description

Updates the minimum supported rust version to 1.85.0 as we are also updating goblin to 0.10, which requires 1.85

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

N/A

## Integration Instructions

N/A

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,7 @@ crc32fast = { version = "1.4", default-features = false }
 fallible-streaming-iterator = { version = "0.1.9" }
 fixedbitset = { version = "^0.5", default-features = false }
 gdbstub = { version = "0.7.3", default-features = false }
-goblin = { version = "0.9.3-hotfix.1", default-features = false, registry = "patina-fw" }
+goblin = { version = "0.10", default-features = false }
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -7,6 +7,7 @@ members = ["components/*", "core/*", "sdk/*", "patina_dxe_core"]
 version = "0.1.1"
 license = "BSD-2-Clause-Patent"
 edition = "2021"
```
os-d, line 9: We also discussed this, but should we move to 2024? Seemingly moving to goblin v0.10.0 which did that move didn't require it on our side.

Javagedes: Just looked. This is a bigger update. I will do this separately in another PR.

makubacki:
@Javagedes, are you updating all of our Rust code (including Mu repos) to 1.85?

Javagedes:
Sure I can work on it. I've got a thing or two that takes priority, but I can get it knocked out after those.

PR #472, opened 2025-05-29
Integrate markdownlint-cli into cargo make all to validate documentation

vineelko:
## Description

Integrate markdownlint-cli into cargo make all to validate documentation

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

NA

## Integration Instructions

NA

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -236,6 +236,10 @@ args = ["fmt", "--all"]
 description = "Run cspell for spell checking." # npm install -g cspell@latest
 script = "cspell --quiet  --no-progress --no-summary  --dot --gitignore -e \"{.git/**,.github/**,.vscode/**}\" ."
 
+[tasks.markdownlint]
+description = "Run markdownlint for documentation linting." # npm install -g markdownlint-cli@latest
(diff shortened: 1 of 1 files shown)


vineelko:
> It's not guaranteed that everyone has this tool (markdownlint) installed. You need to have a check to either skip this tool if it's > not installed or throw a well defined error message for the missing tool.


What you said is true. However, I’ll be following up with another PR that updates the `README.md` to include the prerequisite tools and software needed to work with the Patina repository. That will be more [comprehensive and self contained](https://github.com/OpenDevicePartnership/patina/blob/users/vineelko/patina_setup_guide_0528/README.md#prerequisites) and should handle it.

makubacki:
> What you said is true. However, I’ll be following up with another PR that updates the `README.md` to include the prerequisite tools and software needed to work with the Patina repository. That will be more [comprehensive and self contained](https://github.com/OpenDevicePartnership/patina/blob/users/vineelko/patina_setup_guide_0528/README.md#prerequisites) and should handle it.

You should still have an error message. There is no guarantee people are going to read documentation and the error message can simply point them in the right direction.

PR #473, opened 2025-05-29
Patina Setup Guide - Validated on QEMU - Windows/Linux/WSL - X64/AArch64

vineelko:
## Description

This PR improves the **Patina** front page documentation (`README.md`) to make it more self-contained and comprehensive.
It updates the setup instructions and reorganizes the content to better guide first-time users of the `patina`
repository. The goal is to help users easily build and run `patina` with QEMU on **Windows 11 (24H2)**, **Linux (Ubuntu
24.04 LTS)**, and **WSL**, for both **x64** and **AArch64** architectures all without needing to leave the `README.md`.
All configurations have been thoroughly tested and are current as of 05/28. 😄

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [x] Includes documentation?

## How This Was Tested
Windows/X64: image
Windows/AArch64: image
Linux/X64: image
Linux/AArch64: image

## Integration Instructions

NA

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,5 @@
+<!-- markdownlint-disable MD013 --> <!-- MD013: Line length -->
+<!-- markdownlint-disable MD033 --> <!-- MD033: Inline HTML -->
 
 # Patina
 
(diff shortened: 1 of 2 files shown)


README.md
makubacki, line 1: Shouldn't a lot of this content be in patina-qemu? This repo's documentation was focused more on the Patina source code and workspace whereas patina-qemu is focused on running Patina on QEMU.

README.md
```diff
@@ -16,67 +18,108 @@ incremental migration of today's firmware components largely written in C to Rus
...
-`mdbook serve docs` to self host the getting started book.
+**Note:** Add the LLVM bin directory (`C:\LLVM\bin`) and the QEMU bin directory
+(`C:\Program Files\qemu`) to the `PATH` environment variable.
+</details>
 
-You can also generate API documentation for the project using `cargo make doc`. This will eventually be hosted on
-docs.rs once we begin uploading to crates.io. You can have the documentation opened in your browser by running
-`cargo make doc-open`.
+<details>
+<summary><b> 🐧 Linux/WSL - Ubuntu 24.04 LTS - Bash </b></summary>
```
makubacki, line 60: Linux instructions tend to go stale quickly and dependencies break between components over time. I'd prefer to just provide a container as the Linux solution and use the Dockerfile as reference for how the container is built. Then, we can verify it works in CI.

README.md
```diff
@@ -16,67 +18,108 @@ incremental migration of today's firmware components largely written in C to Rus
...
+| [Node](https://nodejs.org/en)                                                                                                         | `winget install --id OpenJS.NodeJS.LTS -e` <ol><li> **Add cspell:** `npm install -g cspell@latest` </li><li> **Add markdown lint cli:** `npm install -g markdownlint-cli` </li></ol>                                                                                                                                                                                                                                                                |
+| [QEMU](https://www.qemu.org/)                                                                                                         | `winget install --id SoftwareFreedomConservancy.QEMU -e`                                                                                                                                                                                                                                                                                                                                                                                            |
```
makubacki, line 50: We need to constrain what version of QEMU is used. Otherwise, it might not work. On Windows, they should just use what is in the ext dep unless they have a specific need to use something else.

README.md
```diff
@@ -16,67 +18,108 @@ incremental migration of today's firmware components largely written in C to Rus
...
+| [Chocolatey](https://chocolatey.org/)                                                                                                 | `winget install --id Chocolatey.Chocolatey -e`                                                                                                                                                                                                                                                                                                                                                                                                      |
+| [Python 3](https://www.python.org/)                                                                                                   | `winget install Python3` <br> **Note:** Disable any app execution alias defined for `python.exe` and `python3.exe` from Windows settings(Apps > Advanced app settings > App execution alias)                                                                                                                                                                                                                                                        |
```
makubacki, line 43: I think we should specify a Python minor version that aligns with what is being tested in CI. People have already run into issues with Python 3.13 like `win32console` being deprecated that we didn't hit on Python 3.12.

README.md
```diff
@@ -16,67 +18,108 @@ incremental migration of today's firmware components largely written in C to Rus
...
+| [VSCode](https://code.visualstudio.com/docs/setup/linux#_debian-and-ubuntu-based-distributions) | `wget https://go.microsoft.com/fwlink/?LinkID=760868 -O code.deb` <br> `sudo apt install ./code.deb`                                                                                                                                                                                                                                                                                                                     |
+| [Rust Analyzer](https://marketplace.visualstudio.com/items?itemName=rust-lang.rust-analyzer)    | `code --install-extension rust-lang.rust-analyzer`                                                                                                                                                                                                                                                                                                                                                                       |
+
+</details>
+
+### Code
```
makubacki, line 74: This repo table is fine to leave here to point people to the other repos.

README.md
```diff
@@ -16,67 +18,108 @@ incremental migration of today's firmware components largely written in C to Rus
...
+The above steps will help you build and test the vanilla code, with dependencies
+fetched from crates.io. For local development, you should modify the relevant
+crates within the patina repository and update the dependencies using
+appropriate path patches.
 
-   ```shell
-   cargo install cargo-make
-   cargo install cargo-tarpaulin
-   ```
+### Debugging
```
makubacki, line 115: This section should also go to patina-qemu.

README.md
```diff
@@ -16,67 +18,108 @@ incremental migration of today's firmware components largely written in C to Rus
...
 
-4. Another optional tool that has proven useful for speeding up the build process is 'cargo-binstall', located on
-[GitHub](https://github.com/cargo-bins/cargo-binstall).  See the readme.md file in that repository for installation and
-usage instructions.
+- [WinDbg + QEMU + Patina UEFI - Debugging Guide](https://github.com/OpenDevicePartnership/patina-qemu/blob/main/Platforms/Docs/Common/windbg-qemu-uefi-debugging.md)
+- [WinDbg + QEMU + Patina UEFI + Windows OS - Debugging Guide](https://github.com/OpenDevicePartnership/patina-qemu/blob/main/Platforms/Docs/Common/windbg-qemu-windows-debugging.md)
 
-## Build
+## Miscellaneous
+
```
makubacki, line 121: If the section is empty, please drop it.

```suggestion
```

README.md
```diff
@@ -115,14 +158,43 @@ cargo make coverage
...
+Before making pull requests at a minimum, run:
+
+```shell
+cargo make all
+```
+
+## Important Notes
+
+This repository is still considered to be in a "beta" stage at this time. Platform testing and integration feedback
+is very welcome.
```
makubacki, line 200: I would either drop the beta note or explain what is meant by that.

README.md
os-d: I agree, this looks like platform level setup, which is important, but should be co-located with the platform in question

README.md
```diff
@@ -16,67 +18,108 @@ incremental migration of today's firmware components largely written in C to Rus
...
+| [VSCode](https://code.visualstudio.com/)                                                                                              | `winget install --id Microsoft.VisualStudioCode -e`                                                                                                                                                                                                                                                                                                                                                                                                 |
+| [Rust Analyzer](https://marketplace.visualstudio.com/items?itemName=rust-lang.rust-analyzer)                                          | `code --install-extension rust-lang.rust-analyzer`                                                                                                                                                                                                                                                                                                                                                                                                  |
```
os-d, line 53: These three I would note are nice to haves, but you don't need any of them to build and run

README.md
```diff
@@ -16,67 +18,108 @@ incremental migration of today's firmware components largely written in C to Rus
...
+| [VSCode](https://code.visualstudio.com/docs/setup/linux#_debian-and-ubuntu-based-distributions) | `wget https://go.microsoft.com/fwlink/?LinkID=760868 -O code.deb` <br> `sudo apt install ./code.deb`                                                                                                                                                                                                                                                                                                                     |
+| [Rust Analyzer](https://marketplace.visualstudio.com/items?itemName=rust-lang.rust-analyzer)    | `code --install-extension rust-lang.rust-analyzer`                                                                                                                                                                                                                                                                                                                                                                       |
```
os-d, line 70: Same for these

README.md
```diff
@@ -16,67 +18,108 @@ incremental migration of today's firmware components largely written in C to Rus
...
...bin/activate` <br><br> **Build Perquisites:** <br>`pip install --upgrade -r pip-requirements.txt` <br><br> **Stuart Setup:** <br>`stuart_setup  -c Platforms/QemuQ35Pkg/PlatformBuild.py` <br> <br>**Stuart Update:** <br>`stuart_update -c Platforms/QemuQ35Pkg/PlatformBuild.py` <br>**Note:** Retry the command if failed with `Filename too long` error <br><br> **Stuart Build and Launch Uefi Shell:** <br>🪟 `stuart_build  -c Platforms/QemuQ35Pkg/PlatformBuild.py --flashrom BLD_*_DXE_CORE_BINARY_PATH="C:\r\patina-dxe-core-qemu\target\x86_64-unknown-uefi"` <br>🐧 `stuart_build  -c Platforms/QemuQ35Pkg/PlatformBuild.py TOOL_CHAIN_TAG=CLANGPDB --flashrom BLD_*_DXE_CORE_BINARY_PATH="$HOME/r/patina-dxe-core-qemu/target/x86_64-unknown-uefi"` <br><br> **Stuart Build and Launch OS(Optional):** <br>🪟 `stuart_build  -c Platforms/QemuQ35Pkg/PlatformBuild.py --flashrom BLD_*_DXE_CORE_BINARY_PATH="C:\r\patina-dxe-core-qemu\target\x86_64-unknown-uefi" PATH_TO_OS="C:\OS\Windows11.qcow2"` <br>🐧 `stuart_build  -c Platforms/QemuQ35Pkg/PlatformBuild.py TOOL_CHAIN_TAG=CLANGPDB --flashrom BLD_*_DXE_CORE_BINARY_PATH="$HOME/r/patina-dxe-core-qemu/target/x86_64-unknown-uefi" PATH_TO_OS="$HOME/OS/Windows11.qcow2"` |
```
os-d, line 92: The average Windows user would probably choose VS2022, I would note that either VS2022 or CLANGPDB works for Windows (with VS2022 vastly the larger tested option, so I would use this as the default). Similarly, the average Linux user is going to use GCC, so I would note either GCC5 or CLANGPDB works for WSL.

Also, I would add that this is first time or if non-patina has been updated, but add steps to do the fw patcher because that is a much better user experience for just working on patina.

cspell.yml
```diff
@@ -207,4 +212,4 @@ words:
   - xxxxxxxxxb
   - xxxzzzzz
   - zbuild
-  - zunstable
+  - zunstable
```
os-d, line 215: nit: newline at EOF

README.md
vineelko: Thought about it. Let me add it as optional

README.md
vineelko: I'm actually on the fence about this. In fact, I first made this change in the `patina-qemu` repo. But then I felt that the first point of contact for anyone would be the `patina` repo, and from there, branching out to the other repos would make more sense. I thought it would give them a quick glance at what we're up to. I can move this to `patina-qemu`.

os-d: We will see for patina, for Mu, the primary entrypoint for folks has always been mu_tiano_platforms (i.e. the patina-qemu equivalent). I think most folks want to be able to immediately try something and then go jump into the code after that.

README.md
vineelko: IMO, I think it's worth tracking the dependencies and keeping them up to date as we make progress. Docker makes the most sense for CI, but for local development, I still feel that having the dependencies explicitly called out gives developers working on Linux a clear understanding of what's needed to build `patina`, rather than black boxing it with Docker/Dockerfile.

I understand it may be difficult to track these dependencies, but as more users start using the platform, any breaking changes will be caught and can be addressed. I've noticed that build times are considerably faster on Linux compared to Windows, and I'm thinking of switching to Linux myself.

README.md
vineelko: Makes sense let me take a look at it.

README.md
vineelko: As I understand it, building for AArch64 on Windows requires CLANGPDB due to the TF-A build. So, for Windows + AArch64, I have CLANGPDB set up. For Windows + x64, we're still using VS2022.

I’ll also include the firmware patcher and workspace script as easier alternatives. 

The intent of this PR is to make it explicit what it takes to build `patina`, not necessarily to prescribe the only way to do so.

README.md
makubacki: It is way easier to track and identify the dependencies with a Dockerfile. The Dockerfile itself records the dependencies and the versions used are captured at the time of the container build. Dev containers are make using a container almost automatic and it works fine with WSL at native Linux speed (since it's running docker in Linux).

README.md
makubacki: We have to co-locate content appropriately. Someone coming here to build for a physical platform should not have to filter through QEMU specifics. Someone in the QEMU repo, should not have to find this repo to get QEMU instructions. This repo should focus on the Patina content in the repo and `patina-qemu` on how to use it with QEMU.

README.md
makubacki: Yeah, we know that there are people that refuse to use VS Code that work in firmware.

README.md
makubacki: This is all really good. It just needs to be consolidated cleanly in patina-qemu and clearly describe the options available with this being the primary UEFI code build path when they need to generate a new UEFI ROM. workspace_setup for completely new users that need help and the binary patching script for those making Rust dev changes.

os-d: > As I understand it, building for AArch64 on Windows requires CLANGPDB due to the TF-A build. So, for Windows + AArch64, I have CLANGPDB set up. For Windows + x64, we're still using VS2022.

My comment was that your X64 instructions say CLANGPDB: `stuart_build  -c Platforms/QemuQ35Pkg/PlatformBuild.py TOOL_CHAIN_TAG=CLANGPDB`

vineelko: It's actually difficult to review the changes directly in the `.md` file because Markdown tables handle multiline content poorly. Could you please review the [rendered version](https://github.com/OpenDevicePartnership/patina/blob/users/vineelko/patina_setup_guide_0528/README.md) instead?

Below is what I have validated.

  | X64 | AArch64
-- | -- | --
Windows | VS2022 | CLANGPDB
Linux | CLANGPDB | CLANGPDB




image

vineelko: > This is all really good. It just needs to be consolidated cleanly in patina-qemu and clearly describe the options available with this being the primary UEFI code build path when they need to generate a new UEFI ROM. workspace_setup for completely new users that need help and the binary patching script for those making Rust dev changes.

Will move the PR to `patina-qemu` and will follow it up with other alternatives as well.

vineelko:
Thanks for the feedback. I'll incorporate it and move it to the `patina-qemu` repo. Closing this PR for now.

README.md
vineelko: Will be resolved in `patina-qemu` PR

README.md
vineelko: Sure, I might need more time to address this comment. For now, I am moving this PR to `patina-qemu` incorporating other comments except this one. Will follow it up later.

PR #476, opened 2025-05-29
Debugger: Implement callback for core monitor commands

cfernald:
## Description

Implements callback support for the core and components to register monitor command handlers using the new `patina_debugger::add_monitor_command` function. This function allows for adding a command that will cause the debugger to invoke the callback, providing the output writer and the tokens following the invoked command. 

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on Q35

## Integration Instructions

N/A

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -2,14 +2,18 @@
 patina-fw = { index = "sparse+https://pkgs.dev.azure.com/patina-fw/artifacts/_packaging/temporary/Cargo/index/" }
 
 [patch.crates-io]
+patina_adv_logger = { path = "components/patina_adv_logger" }
 patina_debugger = { path = "core/patina_debugger" }
 patina_dxe_core = { path = "patina_dxe_core" }
 patina_internal_collections = { path = "core/patina_internal_collections" }
 patina_internal_cpu = { path = "core/patina_internal_cpu" }
 patina_internal_depex = { path = "core/patina_internal_depex" }
 patina_internal_device_path = { path = "core/patina_internal_device_path" }
+patina_mm = { path = "components/patina_mm" }
 patina_mtrr = { registry = "patina-fw", version = "1" }
 patina_paging = { registry = "patina-fw", version = "6" }
+patina_performance = { path = "components/patina_performance" }
+patina_samples = { path = "components/patina_samples" }
 patina_sdk = { path = "sdk/patina_sdk" }
 patina_sdk_macro = { path = "sdk/patina_sdk_macro" }
 patina_section_extractor = { path = "core/patina_section_extractor" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,7 +4,7 @@ resolver = "2"
 members = ["components/*", "core/*", "sdk/*", "patina_dxe_core"]
 
 [workspace.package]
-version = "1.0.1"
+version = "2.0.0"
 license = "BSD-2-Clause-Patent"
 edition = "2021"
 rust-version = "1.85"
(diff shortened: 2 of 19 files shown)


core/patina_debugger/src/system.rs
```diff
@@ -74,6 +91,11 @@ impl Modules {
     }
 }
 
+pub(crate) struct MonitorCallback {
```
Javagedes, line 125: Needs docs

core/patina_debugger/src/lib.rs
```diff
@@ -113,6 +113,8 @@ use patina_sdk::serial::SerialIO;
 ///
 static DEBUGGER: spin::Once<&dyn Debugger> = spin::Once::new();
 
+type MonitorCommandFn = fn(&mut core::str::SplitWhitespace<'_>, &mut dyn core::fmt::Write);
```
Javagedes: I think this type cast needs to be public, since it is part of the public add_monitor_command function interface. Also needs docs.

core/patina_debugger/src/debugger.rs
```diff
@@ -371,6 +371,24 @@ impl<T: SerialIO> Debugger for UefiDebugger<T> {
             }
         }
     }
+
+    fn add_monitor_command(&'static self, command: &'static str, callback: crate::MonitorCommandFn) {
```
Javagedes, line 375: Docs

core/patina_debugger/src/lib.rs
```diff
@@ -190,6 +195,14 @@ pub fn enabled() -> bool {
     }
 }
 
+/// Adds a monitor command to the debugger. This may be called before initialization,
+/// but should not be called before memory allocations are available.
+pub fn add_monitor_command(cmd: &'static str, function: MonitorCommandFn) {
```
Javagedes, line 226: I'd add an example here.

core/patina_debugger/src/dbg_target/monitor.rs
```diff
@@ -65,9 +71,27 @@ impl ext::monitor_cmd::MonitorCmd for UefiTarget {
             Some("arch") => {
                 SystemArch::monitor_cmd(&mut tokens, &mut self.monitor_buffer);
             }
-            _ => {
-                let _ = self.monitor_buffer.write_str("Unknown command. Use 'help' for a list of commands.");
-            }
+            Some(cmd) => match self.system_state.try_lock() {
```
Javagedes, line 74: This section of code seems fairly easy to test. I'd consider making this handler a method of system state and writing  tests for that function.

``` rust
Some(cmd) => self.system_state.custom_cmd(cmd)
```

core/patina_debugger/src/system.rs
os-d: Somebody talked with Sean after the meeting yesterday ;)

patina_dxe_core/src/lib.rs
```diff
@@ -236,6 +236,10 @@ where
         self.hob_list.relocate_hobs();
 
         // Initialize the debugger if it is enabled.
```
os-d: small nit: this comment belongs to the initialize call below this now

core/patina_debugger/src/debugger.rs
cfernald: We dont typically do docs on a trait implementation right since it will inhereit the docs form the trait, right?

core/patina_debugger/src/dbg_target/monitor.rs
cfernald: Just can't let me be lazy 😅

core/patina_debugger/src/debugger.rs
Javagedes: Ah, sorry I missed this was a trait implementation. I agree with your statement. Closing.

PR #480, opened 2025-05-30
Debugger: Add support for partial monitor responses

cfernald:
## Description

Adds the following support to monitor commands
1. Add offset tag option where the call can specify `O[N]` before the message to skip N bytes.
2. Add a tag `#T#` to the end of truncated messages.

This is intended to be used with a change from `UefiExt` to allow long messages to be printed despite the buffer and windbg limitations.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on Q35

## Integration Instructions

N/A

### DIFF
--- a/core/patina_debugger/src/dbg_target/monitor.rs
+++ b/core/patina_debugger/src/dbg_target/monitor.rs
@@ -40,19 +40,18 @@ impl ext::monitor_cmd::MonitorCmd for UefiTarget {
         let mut tokens: SplitWhitespace<'_> = cmd_str.split_whitespace();
         self.monitor_buffer.reset();
 
-        let mut token = tokens.next();
-
         // Check for an offset modifier, and configure the monitor buffer accordingly.
-        if let Some(cmd) = token {
-            if cmd.starts_with("O[") && cmd.ends_with("]") {
-                let offset_str = &cmd[2..cmd.len() - 1];
+        let cmd = match tokens.next() {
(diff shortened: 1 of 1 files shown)


cfernald:
https://github.com/microsoft/mu_feature_debugger/pull/106 is the UefiExt counterpart

core/patina_debugger/src/dbg_target/monitor.rs
```diff
@@ -40,7 +40,19 @@ impl ext::monitor_cmd::MonitorCmd for UefiTarget {
         let mut tokens: SplitWhitespace<'_> = cmd_str.split_whitespace();
         self.monitor_buffer.reset();
 
-        match tokens.next() {
+        let mut token = tokens.next();
+
+        // Check for an offset modifier, and configure the monitor buffer accordingly.
```
Javagedes, line 43: I'd maybe do something like this:

```rust
let token = match token.next() {
    Some(cmd) if cmd.starts_with("O[") && cmd.ends_with("]") => {
        let offset = cmd[2..cmd.len() - 1].parse().unwrap_or(0);
        self.monitor_buffer.set_start_offset(offset);
        tokens.next()
    }
    other => other,
};
```

Javagedes: But this is just stylistic change so if you don't like it or don't care enough to change it, thats fine.

Issue #467, opened 2025-05-27
[Task]: Remaining Perf Work


makubacki:
### Feature Overview

1. Ability to publish the FBPT as a fallback if perf tracing is "disabled".
2. Install the FPDT when perf tracing is "enabled".
3. Add FPDT report status code listeners. Make sure all status code are picked up.

- Note: We do not support S3 in Patina. As a result, we do not support lock box (lock box is more generic than S3 but that was the main use case, and we've rejected it).

- Note: Try to avoid the `EFI_FIRMWARE_PERFORMANCE_VARIABLE_NAME` variable implementation in Rust code. We should not share pointers between modules using UEFI variables.

- Goal: We eliminate all C performance - DxeCorePerformanceLib (already done), DxePerformanceLib (already done), FirmwarePerformanceDataTableDxe.

---

Today:

- `DxeCorePerformanceLib` - Used by the DXE core only to acquire perf records from MM with the communicate service and from PEI using the HOB list. Installs the performance protocol. Installs the `gPerformanceProtocolGuid` EFI configuration table.
- `DxePerformanceLib` - Used by non-core DXE drivers to report performance records. Wraps the performance protocol so they don't have to deal with it.
- `FirmwarePerformanceDataTableDxe` - Installs the FBPT (in case it is not already installed), FPDT, and S3PT.

### Urgency

Medium

### Are you going to implement the feature request?

Someone else needs to implement the feature

### Do you need maintainer feedback?

No maintainer feedback needed

makubacki:
Assigned to @berlin-with0ut-return

Issue #469, opened 2025-05-27
[Feature]: Add UEFI Variable service


makubacki:
### Feature Overview

This tracks abstracting UEFI variable services so components do not need to directly depend on the runtime services table.

### Solution Overview

See above.

### Urgency

Low

### Are you going to implement the feature request?

Someone else needs to implement the feature

### Do you need maintainer feedback?

No maintainer feedback needed

makubacki:
I've temporarily cleared the ownership for this issue as its implementation may depend on a pending RFC.

Issue #468, opened 2025-05-27
[Feature]: Add Status Code Service


makubacki:
### Feature Overview

C code will continue to produce status codes during important operations that Rust features need to pick up. This tracks abstracting status codes with a Rust service.

### Solution Overview

See above.

### Urgency

Low

### Are you going to implement the feature request?

Someone else needs to implement the feature

### Do you need maintainer feedback?

No maintainer feedback needed

PR #481, opened 2025-05-30
RFC: Firmware Performance Component

berlin-with0ut-return:
## Description
This RFC proposes a Rust-based `FirmwarePerformanceDxe` component that replaces the existing C implementation
responsible for installing and managing UEFI firmware performance tables, specifically the FBPT and FPDT.
The new component initializes during DXE dispatch and handles tasks such as
installing the FPDT, publishing the FBPT, recording ExitBootServices timings, and registering the FPDT status code listener.

Current RFC State: **Draft**

tracks issue #467 

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

### DIFF
--- a/docs/src/rfc/text/0000-firmware-performance-dxe.md
+++ b/docs/src/rfc/text/0000-firmware-performance-dxe.md
@@ -52,7 +52,7 @@ The new `FirmwarePerformanceDxe` component will cover remaining performance func
 
 The Rust `FirmwarePerformanceDxe` component closely follows the design of the `FirmwarePerformanceDxe` driver in C.
 
-The primary functionality of the C `FirmwarePerformancedxe` driver is encapsulated in
+The primary functionality of the C `FirmwarePerformanceDxe` driver is encapsulated in
 the `FirmwarePerformanceDxeEntryPoint`, which registers the necessary FPDT status code listener
 and sets up two event callbacks for `EndOfDxe` and `ExitBootServices`.
 


docs/src/rfc/text/0000-firmware-performance-dxe.md
```diff
@@ -0,0 +1,338 @@
...
+For the FBPT, the driver fills in recorded event times and installs the table, regardless of whether performance is enabled.
+
+### Dependencies on C Protocols/Services
+
+At the time of this RFC, performance has several C dependencies.
+Even if the `FirmwarePerformanceDxe` component is implemented in Rust,
+we still have unresolved dependencies on C protocols and services, including:
+
+- Runtime services (for variable storage and retrieval across boot)
+- Status code protocol (for registering a new FPDT listener)
```
Javagedes, line 72: #468.

docs/src/rfc/text/0000-firmware-performance-dxe.md
```diff
@@ -0,0 +1,338 @@
...
+
+For the FBPT, the driver fills in recorded event times and installs the table, regardless of whether performance is enabled.
+
+### Dependencies on C Protocols/Services
+
+At the time of this RFC, performance has several C dependencies.
+Even if the `FirmwarePerformanceDxe` component is implemented in Rust,
+we still have unresolved dependencies on C protocols and services, including:
+
+- Runtime services (for variable storage and retrieval across boot)
```
Javagedes, line 71: #469 for Variable services - so it sounds like we don't need runtime services at all?

docs/src/rfc/text/0000-firmware-performance-dxe.md
```diff
@@ -0,0 +1,338 @@
...
+pub struct FirmwarePerformanceDxe<B: BootServices + 'static, R: RuntimeServices + 'static> {
+    /// Internal storage for the FBPT
+    boot_performance_table: Mutex<BootPerformanceTable>,
+
+    /// Internal storage for the FPDT
+    firmware_performance_table: Mutex<FirmwarePerformanceTable>,
+
+    /// Provides memory allocation services
+    /// Primarily used to install the FPDT
+    memory_manager: OnceCell<Service<dyn MemoryManager>>,
```
Javagedes, line 111: Service init now supports const .

```suggestion
    memory_manager: Service<dyn MemoryManager,
```

However, I don't see a need for this structure to be static initialized though (e.g. `static FIRMWARE_PERFORMANCE_DXE: FirmwarePerformanceDxe = FirmwarePerformanceDxe::new()` because this is not providing a protocol that that needs global static accesses. This component simply registers a few callbacks.

the boot services table events allow you to pass context, and while the C protocol implementation for Report Status Code Handler Protocol does not, we will be re-writing that in rust, so we just need to make sure the `Service<>` interface for the Report Status Code Handler does support context

docs/src/rfc/text/0000-firmware-performance-dxe.md
```diff
@@ -0,0 +1,338 @@
...
+As part of the goal to move to a pure Rust interface for performance,
+we require a Rust-based implementation that installs the FBPT and FPDT,
+deals with events related to the end of DXE and ExitBootServices, and registers FPDT status code listeners.
+At the time of this RFC, such functionality does not exist in the performance component.
+
+## Technology Background
+
+Firmware performance is recorded by two main ACPI tables: the FBPT and FPDT.
+
+The FBPT captures basic boot performance data early in the boot process, and is usually passed to DXE in the HOB.
```
Javagedes, line 26: Can you add information on the firmware performance guided hob layout / consumption

docs/src/rfc/text/0000-firmware-performance-dxe.md
```diff
@@ -0,0 +1,338 @@
...
+
+1. Publishing the FBPT when performance tracing is disabled
+2. Installing the FPDT when performance tracing is enabled
+3. Collecting early boot performance metrics for the FBPT: reset-end and OS loader timestamps
+4. Recording timestamps of `ExitBootServices` for the FPDT and FBPT
+5. Registering FPDT status code listeners
+
+## Unresolved Questions
+
+- How do we avoid the use of global statics, given that many of the C-based APIs provide limited ways to pass around context?
```
Javagedes, line 47: I don't think this is a concern. Two of the three callbacks are boot services events, which allow for passing context, and the third is Report Status Code, which we will be re-writing, so we just need to make sure the `Service` interface for it allows passing context.

docs/src/rfc/text/0000-firmware-performance-dxe.md
```diff
@@ -0,0 +1,338 @@
...
+we use the `context` argument to store necessary context for callbacks.
+
+An example for `fpdt_notify_end_of_dxe` is provided below.
+
+```rust
+struct FpdtContext<R>
+where
+    R: RuntimeServices,
+{
+    firmware_table: *const spin::Mutex<FirmwarePerformanceTable>,
```
Javagedes, line 217: I think we need to investigate a solution other than a bunch of pointers. It's quite un-rust-like and we have the opportunity to do it different, and better. At a minimum, runtime services (The actual runtime services) is already a wrapper around a pointer, and you can clone/copy it to get your own instance (that uses the same pointer), similar to the `Service`

Additionally, if all you need runtime services for is variables, then currently the plan is to make a service for variables, so you won't even need runtime services.

docs/src/rfc/text/0000-firmware-performance-dxe.md
berlin-with0ut-return: yes, we only need variable services really (which are retrieved through `RuntimeServices` from what i understand)

PR #482, opened 2025-05-30
[REBASE&FF] mdbook: Add Component Crate format standards

Javagedes:
## Description

Implements [RFC 9](https://github.com/OpenDevicePartnership/patina/blob/main/docs/src/rfc/text/0009-standardize-component-crate.md) to set standardized requirements for crates that produce components that platforms wish to consume.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [x] Includes documentation?

## How This Was Tested

N/A

## Integration Instructions

N/A

### DIFF
--- a/docs/src/component/interface.md
+++ b/docs/src/component/interface.md
@@ -10,15 +10,15 @@ means is that instead of evaluating a dependency expression to determine if a dr
 attempts to fetch all requested parameters defined in the function interface. If all are successfully fetched, then the
 component is executed, if not, it will not be dispatched, and another attempt will be made in the next iteration.
 
-In the Rust DXE Core, a component is simply a trait implementation. So long as a struct implements [Component](todo/docs.rs)
-and [IntoComponent](todo/docs.rs), it can be consumed and executed by the pure rust DXE core. [patina_sdk](todo/docs.rs)
-currently provides two implementations for `Component`:
+In the Rust DXE Core, a component is simply a trait implementation. So long as a struct implements [IntoComponent][patina_sdk],
+it can be consumed and executed by the pure rust DXE core. [patina_sdk][patina_sdk] currently provides two
+implementations for `Component`:
 
-The first is [FunctionComponent](todo/docs.rs). This type cannot be instantiated manually, but a blanket implementation
-of the `IntoComponent` trait allows any function whose parameters support dependency injection to be converted into a
+1. [FunctionComponent][patina_sdk]: this type cannot be instantiated manually, but a blanket implementation of the
+`IntoComponent` trait allows any function whose parameters support dependency injection to be converted into a
 `FunctionComponent`.
 
-The second is [StructComponent](todo/docs.rs). This type cannot be instantiated manually, but a derive proc-macro of
+2. [StructComponent][patina_sdk]: this type cannot be instantiated manually, but a derive proc-macro of
 `IntoComponent` is provided that will allow any struct or enum to be used as a component. This derive proc-macro
 expects that a `Self::entry_point(self, ...) -> patina_sdk::error::Result<()> { ... }` exists, where the `...` in the
 function definition can be any number of parameters twho support dependency injection as shown below. The function
@@ -28,7 +28,7 @@ See [Samples](https://github.com/OpenDevicePartnership/patina/tree/main/componen
 for examples of basic components using these two methods.
 
 Due to this, developing a component is as simple as writing a function whose parameters are a part of the below list of
-supported parameters (which is subject to change). Always reference the trait's [Type Implementations](todo/docs.rs)
+supported parameters (which is subject to change). Always reference the trait's [Type Implementations][patina_sdk]
 for a complete list, however the below information should be up to date.
 
 ## Component Execution
@@ -44,7 +44,7 @@ components have been dispatched in a single iteration.
 Writing a component is as simple as writing a function whose parameters are a part of the below list of suppported
 types (which is subject ot change). The `Param` trait is the interface that the dispatcher uses to (1) validate that a
 parameter is available, (2) retrieve the datum from storage, and (3) pass it to the component when executing it. Always
-reference the `Param` trait's [Type Implementations](todo/docs.rs) for a complete list of parameters that can be used
+reference the `Param` trait's [Type Implementations][patina_sdk] for a complete list of parameters that can be used
 in the function interface of a component.
 
 ```admonish warning
@@ -95,10 +95,10 @@ This type comes with a `mock(...)` method to make unit testing simple.
 
 ### Hob\<T\>
 
-The `Hob<T>` parameter type is used to access a guided hob value, which is automatically parsed from the HOB list
+The `Hob<T>` parameter type is used to access a GUID HOB value, which is automatically parsed from the HOB list
 provided to the Core during initialization. Unlike `Config<T>` and `ConfigMut<T>`, these types are **not** always
 available. This means a component that has this parameter implemented will only be executed if the guided HOB is
-found in the HOB list. Due to how HOBs work, the same guided HOB can be provided multiple times to the platform.
+found in the HOB list. Due to how HOBs work, the same GUID HOB can be provided multiple times to the platform.
 Due to this, `Hob<T>` implements both `Deref` to access the first found value, or `IntoIterator` to iterate through
 all HOB values.
 
@@ -243,3 +243,9 @@ impl MyComponent2 {
     }
 }
 ```
+
+<!--
+    TODO: Replace usage of patina_sdk with links to specific documentation once patina_sdk is available on
+    crates.io docs.rs.
+-->
+[patina_sdk]: https://github.com/OpenDevicePartnership/patina/tree/main/sdk/patina_sdk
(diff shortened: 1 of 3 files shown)


docs/src/component/requirements.md
```diff
@@ -0,0 +1,103 @@
+# Component Crate Requirements
+
+Cargo crates are used to bundle like-minded functionality together; this concept also extends to Patina components.
+It is common for a "feature" or "bundle" of functionality to utilize multiple components, configurations, Services,
```
makubacki: nit: I didn't see a reason to capitalize "Services" if somethings like "components" is not capitalized.

```suggestion
It is common for a "feature" or "bundle" of functionality to utilize multiple components, configurations, services,
```

docs/src/component/requirements.md
```diff
@@ -0,0 +1,103 @@
...
+etc. to fully define certain functionality while keeping the different areas of concerns segregated. The Patina team
+found that due to this complexity, these features can become confusing / burdomsome for platforms to integrate,
```
makubacki:
```suggestion
found that due to this complexity, these features can become confusing / burdensome for platforms to integrate,
```

docs/src/component/requirements.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Documentation
+
+Patina requires that all public modules, types, traits, etc. must be documented at a minimum. Documentation must meet
+the same requirements as specified in [Documenting](../dev/documenting.md). It is suggested that you set the docs lint
+[missing_docs](https://doc.rust-lang.org/rustdoc/lints.html#missing_docs) to error.
+
+## Naming Standards
+
+### Crate naming
```
makubacki, line 20: I feel like this should touch on "private" crates and the naming convention for them.

docs/src/component/interface.md
```diff
@@ -10,9 +10,9 @@ means is that instead of evaluating a dependency expression to determine if a dr
 attempts to fetch all requested parameters defined in the function interface. If all are successfully fetched, then the
 component is executed, if not, it will not be dispatched, and another attempt will be made in the next iteration.
 
-In the Rust DXE Core, a component is simply a trait implementation. So long as a struct implements [Component](todo/docs.rs)
-and [IntoComponent](todo/docs.rs), it can be consumed and executed by the pure rust DXE core. [patina_sdk](todo/docs.rs)
-currently provides two implementations for `Component`:
+In the Rust DXE Core, a component is simply a trait implementation. So long as a struct [IntoComponent](todo/docs.rs),
```
berlin-with0ut-return: nit: "implements" should probably still be here

docs/src/component/requirements.md
```diff
@@ -0,0 +1,105 @@
...
+   with the platform via `.with_config` and this config is not publically accessible via `patina_sdk` or elsewhere.
+4. `error` module: This module may optionally exist if a `service` module is present and the public Service's interface
+   contains custom errors.
+5. `hob` module: This module may optionally exist if a new guided hob type has been created for this component. The
+   hob module and associated guided HOB(s) should be made public such that it can be consumed by others if the need
+   arise. Any common or spec defined HOBs should be added to the associated crates (such as `mu_rust_pi`, `patina_sdk`,
+   etc.) rather than this crate. HOBs may become a common interface and should thus be moved to the appropriate crate.
+   If the HOB type already exists elsewhere, the crate should consume that definition instead of making their own.
+6. `service` module: This module may optionally exist if the component produces a service that is not publically
+   accessible via `patina_sdk` or another crate.
```
berlin-with0ut-return: the wording here is confusing, at least to me - isn't the goal of produced Services always to be publicly available to other components?

docs/src/component/requirements.md
```diff
@@ -0,0 +1,105 @@
...
+
+The below is a list of requirements for the crate, but it does not prevent additional modules from existing.
+
+1. No public definitions are accessible via the top level lib.rs (or equivalent) module, only public modules.
+2. `component` module: This module must always exist, and contain the publicly importable component(s) for the crate.
+3. `config` module: This module may optionally exist if the component consumes configuration data that is registered
+   with the platform via `.with_config` and this config is not publically accessible via `patina_sdk` or elsewhere.
+4. `error` module: This module may optionally exist if a `service` module is present and the public Service's interface
+   contains custom errors.
+5. `hob` module: This module may optionally exist if a new guided hob type has been created for this component. The
```
berlin-with0ut-return, line 78: nit, but i think we should use GUID hob or GUIDed hob to avoid confusion with the word "guide"

docs/src/component/requirements.md
Javagedes: A crate may exist that does not create new service definitions, only provides their own implementations for services that are defined by another entity. In this scenario, they will not need a `service` module :)

docs/src/component/interface.md
Javagedes: whoops yes. good catch thank you

docs/src/component/interface.md
```diff
@@ -10,9 +10,9 @@ means is that instead of evaluating a dependency expression to determine if a dr
...
+it can be consumed and executed by the pure rust DXE core. [patina_sdk](todo/docs.rs) currently provides two
```
makubacki: I think linking to the `patina_sdk` crate in the meantime might be more useful than the todo.

docs/src/component/requirements.md
Javagedes: Additionally, not all components are expected to produce services. Many components will be expected to consume services to complete certain actions, without needing to produce a service themselves!

docs/src/component/interface.md
Javagedes: Sounds good. Updated them all to the base of the patina_sdk crate and added a todo

docs/src/component/requirements.md
Javagedes: Updated to GUID. Good idea ;)

docs/src/component/requirements.md
Javagedes: Happy to add some clarification to make that more clear, however. Provide any suggestions, @berlin-with0ut-return and I'll add it and close this conversation :)

docs/src/component/interface.md
Javagedes: Resolved

docs/src/component/requirements.md
```diff
@@ -0,0 +1,105 @@
+# Component Crate Requirements
+
+Cargo crates are used to bundle like-minded functionality together; this concept also extends to Patina components.
+It is common for a "feature" or "bundle" of functionality to utilize multiple components, configurations, services,
+etc. to fully define certain functionality while keeping the different areas of concerns segregated. The Patina team
+found that due to this complexity, these features can become confusing / burdensome  for platforms to integrate,
```
cfernald:
```suggestion
found that due to this complexity, these features can become confusing / burdensome for platforms to integrate,
```

docs/src/component/requirements.md
```diff
@@ -0,0 +1,105 @@
...
+because each feature had it's own crate layout which can be difficult to navigate if not documented properly. Due to
+this, the Patina team introduced RFC #9 to partially standardize a patina component crate's layout, providing
```
cfernald:
```suggestion
this, the Patina team introduced [RFC #9](https://github.com/OpenDevicePartnership/patina/blob/main/docs/src/rfc/text/0009-standardize-component-crate.md) to partially standardize a patina component crate's layout, providing
```

docs/src/component/requirements.md
```diff
@@ -0,0 +1,105 @@
...
+[missing_docs](https://doc.rust-lang.org/rustdoc/lints.html#missing_docs) to error.
+
+## Naming Standards
+
+### Crate naming
+
+It is required that any crate published to crates.io (or other registry) that produces one or more components must be
+prefixed with `patina_` to allow them to be easily identifiable and locatable via normal searches. Should you maintain
+not meant for public consumption, it should be prefixed with `patina_internal_`. The one exception are macro crates,
+which should contain a suffix of `_macro` instead (e.g. `patina_my_crate` and `patina_my_crate_macro`).
```
cfernald: Should this be caveated that this only applies to crates from the official patina repos? I'm on the fence as it seems to go without saying, but might be good to be explicit to avoid cruft being published from other places under the patina_* naming scheme.

docs/src/component/requirements.md
cfernald: I would just delete the word "publicly", doesn't add anything IMO and would clarify.

docs/src/component/requirements.md
Javagedes: In my head, I actually had the patina naming as a requirement for external, vendor made patina components. Otherwise it becomes difficult for the "open source community" concept to work as it will be hard to find available components on crates.io. We can change this if we want though. I'm not too worried about it. 

I think the other option is just require that "patina" be in their keywords section of the cargo.toml file for their crate

```toml
[package]
keywords = ["patina"]
```

Javagedes: Updated

PR #487, opened 2025-06-04
Debugger: Renaming and interfaces improvments

cfernald:
## Description

1. Renamed `UefiDebugger` to `PatinaDebugger` for consistency
2. Changes `.with_debugger_logging` to a policy that allows all logging, non-debugger logging, and no logging to better support different transport configurations.
3. Replaces `.with_default_config` with a simpiler `.with_force_enable`.

- [x] Impacts functionality?
- [ ] Impacts security?
- [x] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on Q35 & SBSA

## Integration Instructions

- Change `UefiDebugger` referenced to `PatinaDebugger`
- Change `.with_debugger_logging` with `.with_log_policy(DebugLoggingPolicy::FullLogging)`
- replace `.with_default_config` with `.with_force_enable`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 /target
 *.profraw
 Cargo.lock
+docs/book/
--- a/docs/src/SUMMARY.md
+++ b/docs/src/SUMMARY.md
@@ -53,9 +53,10 @@
   - [Synchronization](dxe_core/synchronization.md)
   - [Testing](dxe_core/testing.md)
 
-# Creating a Rust DXE Driver
+# Creating a Patina Component
 
-- [Component Interface](driver/interface.md)
+- [Component Interface](component/interface.md)
+- [Component Crate Requirements](component/requirements.md)
 
 -----------
 [Contributors](misc/contributors.md)
--- a/docs/src/component/interface.md
+++ b/docs/src/component/interface.md
@@ -10,15 +10,15 @@ means is that instead of evaluating a dependency expression to determine if a dr
 attempts to fetch all requested parameters defined in the function interface. If all are successfully fetched, then the
 component is executed, if not, it will not be dispatched, and another attempt will be made in the next iteration.
 
-In the Rust DXE Core, a component is simply a trait implementation. So long as a struct implements [Component](todo/docs.rs)
-and [IntoComponent](todo/docs.rs), it can be consumed and executed by the pure rust DXE core. [patina_sdk](todo/docs.rs)
-currently provides two implementations for `Component`:
+In the Rust DXE Core, a component is simply a trait implementation. So long as a struct implements [IntoComponent][patina_sdk],
+it can be consumed and executed by the pure rust DXE core. [patina_sdk][patina_sdk] currently provides two
+implementations for `Component`:
 
-The first is [FunctionComponent](todo/docs.rs). This type cannot be instantiated manually, but a blanket implementation
-of the `IntoComponent` trait allows any function whose parameters support dependency injection to be converted into a
+1. [FunctionComponent][patina_sdk]: this type cannot be instantiated manually, but a blanket implementation of the
+`IntoComponent` trait allows any function whose parameters support dependency injection to be converted into a
 `FunctionComponent`.
 
-The second is [StructComponent](todo/docs.rs). This type cannot be instantiated manually, but a derive proc-macro of
+2. [StructComponent][patina_sdk]: this type cannot be instantiated manually, but a derive proc-macro of
 `IntoComponent` is provided that will allow any struct or enum to be used as a component. This derive proc-macro
 expects that a `Self::entry_point(self, ...) -> patina_sdk::error::Result<()> { ... }` exists, where the `...` in the
 function definition can be any number of parameters twho support dependency injection as shown below. The function
@@ -28,7 +28,7 @@ See [Samples](https://github.com/OpenDevicePartnership/patina/tree/main/componen
 for examples of basic components using these two methods.
 
 Due to this, developing a component is as simple as writing a function whose parameters are a part of the below list of
-supported parameters (which is subject to change). Always reference the trait's [Type Implementations](todo/docs.rs)
+supported parameters (which is subject to change). Always reference the trait's [Type Implementations][patina_sdk]
 for a complete list, however the below information should be up to date.
 
 ## Component Execution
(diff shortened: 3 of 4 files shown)


core/patina_debugger/src/debugger.rs
```diff
@@ -116,28 +116,26 @@ impl<T: SerialIO> UefiDebugger<T> {
...
     }
 
-    /// Customizes the default configuration of the debugger.
+    /// Forces the debugger to be enabled, regardless of later configuration. This
+    /// is used for development purposes and is not intended for production or
+    /// standard use. If `False` is provided, this routine will not change the configuration.
     ///
-    /// To be used with a new debugger invocation, this routine allows the caller
-    /// to customize the static debugger creation with specific configuration.
+    /// This will also forcible enable the initial breakpoint with no timeout. This
```
os-d:
```suggestion
    /// This will also forcibly enable the initial breakpoint with no timeout. This
```

core/patina_debugger/src/lib.rs
```diff
@@ -167,8 +167,22 @@ enum DebugError {
...
+    /// The debugger will suspend logging while broken in, but will not change the
+    /// logging state outside of the debugger. This may cause instability if the
+    /// debugger and logging share a transport.
+    SuspendLogging,
+    /// The debugger will disable all logging after a connection is made. This is
+    /// the safest option if the debugger and logging share a transport.
+    DisableLogging,
+    /// The debugger will not suspend logging while broken in. This should only
+    /// be used if the debugger and logging transport are separate.
+    DebuggerLogging,
```
os-d: nit: naming is a little confusing here, this makes it sound like you are enabling logging for the debugger itself, maybe calling this something like `KeepLogging` or something better

core/patina_debugger/src/lib.rs
```diff
@@ -181,7 +195,9 @@ pub fn initialize(interrupt_manager: &mut dyn InterruptManager) {
     }
 }
 
-/// Invokes a debug break instruction.
+/// Invokes a debug break instruction. Callers should ensure that the debugger
+/// is enabled before invoking this routine using the [enabled] routine. If this
+/// routine is invoked when the debugger is not enabled, it will cause an exception.
```
os-d:
```suggestion
/// routine is invoked when the debugger is not enabled, it will cause an unhandled exception.
```
It will always cause an exception :)

core/patina_debugger/src/debugger.rs
```diff
@@ -179,11 +177,15 @@ impl<T: SerialIO> UefiDebugger<T> {
...
-        }
+        // Suspend or disable logging. If suspended, logging will resume when the struct is dropped.
+        let _log_suspend = match self.log_policy {
+            DebuggerLoggingPolicy::SuspendLogging => Some(LoggingSuspender::suspend()),
+            DebuggerLoggingPolicy::DisableLogging => {
+                log::set_max_level(log::LevelFilter::Off);
+                None
+            }
+            DebuggerLoggingPolicy::DebuggerLogging => None,
+        };
```
vineelko:
```suggestion
        // Suspend or disable logging. If suspended, logging will resume when the struct is dropped.
        match self.log_policy {
            DebuggerLoggingPolicy::SuspendLogging => {
                LoggingSuspender::suspend();
            }
            DebuggerLoggingPolicy::DisableLogging => {
                log::set_max_level(log::LevelFilter::Off);
            }
            DebuggerLoggingPolicy::DebuggerLogging => (),
        };
```

core/patina_debugger/src/lib.rs
cfernald: Well you are in a sense, this will allow log messages from within the debugger itself, same as `.with_debugger_logging` previously. I will improve the comment to make this clear.

core/patina_debugger/src/debugger.rs
cfernald: The `_log_suspend` is actually needed. I made it to ensure that we only suspend logging for a given scope, so it uses the `Drop` trait to restore logging when we exit this scope. If I dont store it at the appropriate scope, then it will be restored immediately.

core/patina_debugger/src/lib.rs
cfernald: I suppose so 😅

core/patina_debugger/src/debugger.rs
cfernald: I can clean this up a bit though, the option is probably not needed.

core/patina_debugger/src/debugger.rs
```diff
@@ -179,10 +177,18 @@ impl<T: SerialIO> UefiDebugger<T> {
...
         };
 
-        // Suspend logging. This will resume logging when the struct is dropped.
+        // Suspend or disable logging. If suspended, logging will resume when the struct is dropped.
         let _log_suspend;
-        if !self.debugger_log {
-            _log_suspend = LoggingSuspender::suspend();
+        match self.log_policy {
+            DebuggerLoggingPolicy::SuspendLogging => {
+                _log_suspend = LoggingSuspender::suspend();
```
vineelko, line 184: Wonder why is `_log_suspend` still needed here?

cfernald: To establish the proper scope. log suspender will resume when dropped and so we want it to be at the outer scope instead of the inner scope.

vineelko: It feels a bit like a workaround and might be hard to follow as-is. Would it be possible to move this logic into the `suspend` function instead?

```rust
pub fn suspend(log_policy: &DebuggerLoggingPolicy) -> Self {
    let level;
    match log_policy {
        DebuggerLoggingPolicy::SuspendLogging => {
            level = log::max_level();
            log::set_max_level(log::LevelFilter::Off);
        }
        DebuggerLoggingPolicy::DisableLogging => {
            level = log::LevelFilter::Off;
            log::set_max_level(log::LevelFilter::Off);
        }
        DebuggerLoggingPolicy::FullLogging => {
            level = log::max_level();
        }
    }

    LoggingSuspender { level }
}
```
And the above code in `debugger.rs` gets replaced with below to maintain the scope rules.
```rust
LoggingSuspender::suspend(&self.log_policy);
```

PR #488, opened 2025-06-04
DxeCore: Add Debug Image Info Table

### LINKED ISSUE #177: [Bug]: Missing Configuration Tables

### Current Behavior

The EDK2 C core produces the following configuration tables and installs them in the SystemTable->ConfigurationTable list, but uefi-dxe-core does not:

gEfiDxeServicesTableGuid       = { 0x05AD34BA, 0x6F02, 0x4214, { 0x95, 0x2E, 0x4D, 0xA0, 0x39, 0x8E, 0x2B, 0xB9 }}
gEfiMemoryTypeInformationGuid  = { 0x4C19049F, 0x4137, 0x4DD3, { 0x9C, 0x10, 0x8B, 0x97, 0xA8, 0x3F, 0xFD, 0xFA }}
gEfiDebugImageInfoTableGuid    = { 0x49152E77, 0x1ADA, 0x4764, { 0xB7, 0xA2, 0x7A, 0xFE, 0xFE, 0xD9, 0x5E, 0x8B }}

### Expected Behavior

These tables should be constructed by the uefi-dxe-core for feature parity with the edk2 C core.

### Steps To Reproduce

1. Dump the SystemTable->Configuration Table at Ready To Boot In the EDK2 Core and the Rust Core
2. Compare system table contents, note the discrepancies between them.

### Build Environment

```markdown
- OS(s): N/A
- Tool Chain(s): N/A
- Targets Impacted: All
```

### Version Information

```text

```

### Urgency

Medium

I will fix it

None

### Anything else?

_No response_

os-d:
## Description

This adds the UEFI spec defined debug image info table to patina and uses it to enable debugger support for finding loaded images.

Many of these definitions are intended to be upstreamed to r-efi, but that is a slow process and will be done in parallel.

Note: These structures are *very* C based structures and don't play well into how Rust would like structures to be formed.

Closes #177 .

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested by updating uefiext to support !findall for Rust.

## Integration Instructions

N/A.

### DIFF
--- a/patina_dxe_core/src/config_tables.rs
+++ b/patina_dxe_core/src/config_tables.rs
@@ -26,10 +26,12 @@ extern "efiapi" fn install_configuration_table(table_guid: *mut efi::Guid, table
     }
 
     let table_guid = unsafe { *table_guid };
-    let table = unsafe { table.as_mut() };
 
     let mut st_guard = SYSTEM_TABLE.lock();
-    let st = st_guard.as_mut().expect("System table support not initialized");
+    let st = match st_guard.as_mut() {
+        Some(st) => st,
+        None => return efi::Status::NOT_FOUND,
+    };
 
     match core_install_configuration_table(table_guid, table, st) {
         Err(err) => err.into(),
@@ -39,7 +41,7 @@ extern "efiapi" fn install_configuration_table(table_guid: *mut efi::Guid, table
 
 pub fn core_install_configuration_table(
     vendor_guid: efi::Guid,
-    vendor_table: Option<&mut c_void>,
+    vendor_table: *mut c_void,
     efi_system_table: &mut EfiSystemTable,
 ) -> Result<(), EfiError> {
     let system_table = efi_system_table.as_mut();
@@ -63,8 +65,8 @@ pub fn core_install_configuration_table(
             // a configuration table list is already present.
             let mut current_table = cfg_table.to_vec();
             let existing_entry = current_table.iter_mut().find(|x| x.vendor_guid == vendor_guid);
-            if let Some(vendor_table) = vendor_table {
-                //vendor_table is some; we are adding or modifying an entry.
+            if !vendor_table.is_null() {
+                // vendor_table is not null; we are adding or modifying an entry.
                 if let Some(entry) = existing_entry {
                     //entry exists, modify it.
                     entry.vendor_table = vendor_table;
@@ -88,12 +90,12 @@ pub fn core_install_configuration_table(
             current_table
         }
         None => {
-            //config table list doesn't exist.
-            if let Some(table) = vendor_table {
+            // config table list doesn't exist.
+            if !vendor_table.is_null() {
                 // table is some, meaning we should create the list and add this as the new entry.
-                vec![efi::ConfigurationTable { vendor_guid, vendor_table: table }]
+                vec![efi::ConfigurationTable { vendor_guid, vendor_table }]
             } else {
-                //table is none, but can't delete a table entry in a list that doesn't exist.
+                // table is null, but can't delete a table entry in a list that doesn't exist.
                 //since the list doesn't exist, we can leave the (null) pointer in the st alone.
                 return Err(EfiError::NotFound);
             }
--- a/patina_dxe_core/src/allocator.rs
+++ b/patina_dxe_core/src/allocator.rs
@@ -775,11 +775,13 @@ fn page_change_callback(allocator: &mut FixedSizeBlockAllocator) {
 }
 
 pub fn install_memory_type_info_table(system_table: &mut EfiSystemTable) -> Result<(), EfiError> {
-    //MEMORY_TYPE_INFO_TABLE is static mut, so we know the pointer is good.
+    // SAFETY: This is safe because we are initializing the table with a static array
     #[allow(static_mut_refs)]
-    let memory_table_mut = unsafe { (MEMORY_TYPE_INFO_TABLE.as_mut_ptr() as *mut c_void).as_mut().unwrap() };
-
-    config_tables::core_install_configuration_table(guid::MEMORY_TYPE_INFORMATION, Some(memory_table_mut), system_table)
+    config_tables::core_install_configuration_table(
+        guid::MEMORY_TYPE_INFORMATION,
+        unsafe { MEMORY_TYPE_INFO_TABLE.as_mut_ptr() as *mut c_void },
+        system_table,
+    )
 }
 
 fn process_hob_allocations(hob_list: &HobList) {
--- a/patina_dxe_core/src/config_tables/debug_image_info_table.rs
+++ b/patina_dxe_core/src/config_tables/debug_image_info_table.rs
@@ -121,14 +121,10 @@ pub(crate) fn initialize_debug_image_info_table(system_table: &mut EfiSystemTabl
     });
 
     let table_ptr = Box::into_raw(debug_image_info_table_header) as *mut c_void;
-    // SAFETY: This is safe because we just allocated the table and we are going to use it immediately
-    unsafe {
-        if core_install_configuration_table(EFI_DEBUG_IMAGE_INFO_TABLE_GUID, table_ptr.as_mut(), system_table).is_err()
-        {
-            log::error!("Failed to install configuration table for EFI_DEBUG_IMAGE_INFO_TABLE_GUID");
-            return;
-        };
-    }
+    if core_install_configuration_table(EFI_DEBUG_IMAGE_INFO_TABLE_GUID, table_ptr, system_table).is_err() {
+        log::error!("Failed to install configuration table for EFI_DEBUG_IMAGE_INFO_TABLE_GUID");
+        return;
+    };
 
     // SAFETY: This is safe because we just allocated the table and we are going to use it immediately
     let table = Box::new(DebugImageInfoTableMetadata {
--- a/patina_dxe_core/src/config_tables/memory_attributes_table.rs
+++ b/patina_dxe_core/src/config_tables/memory_attributes_table.rs
@@ -128,17 +128,11 @@ pub fn core_install_memory_attributes_table() {
                     };
                     MEMORY_ATTRIBUTES_TABLE.store(empty_ptr, Ordering::Relaxed);
 
-                    // it is unsafe to get a mutable reference to the MAT here, but we know that we have a valid ptr
-                    unsafe {
-                        if let Err(status) =
-                            core_install_configuration_table(efi::MEMORY_ATTRIBUTES_TABLE_GUID, empty_ptr.as_mut(), st)
-                        {
-                            log::error!(
-                                "Failed to create a null MAT table with status {:#X?}, cannot create MAT",
-                                status
-                            );
-                            return;
-                        }
+                    if let Err(status) =
+                        core_install_configuration_table(efi::MEMORY_ATTRIBUTES_TABLE_GUID, empty_ptr, st)
+                    {
+                        log::error!("Failed to create a null MAT table with status {:#X?}, cannot create MAT", status);
+                        return;
                     }
                 }
             }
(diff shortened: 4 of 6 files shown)


patina_dxe_core/src/debug_image_info_table.rs
```diff
@@ -0,0 +1,360 @@
...
+    protocol_db,
+    systemtables::EfiSystemTable,
+    GCD,
+};
+
+use mu_pi::dxe_services::GcdMemoryType;
+
+use r_efi::efi;
+
+// to be sent upstream to r_efi
```
cfernald: mind filing an issue for this, to make sure it gets cleaned up

patina_dxe_core/src/debug_image_info_table.rs
```diff
@@ -0,0 +1,360 @@
...
+    update_status: u32, // This is made not pub to force volatile access to the field, per UEFI spec
+    pub table_size: u32,
+    pub efi_debug_image_info_table: *const EfiDebugImageInfo,
+}
+
+/// The update status field in the DebugImageInfoTableHeader is used to indicate the status of the table and per
+/// UEFI spec, it should be accessed using volatile reads and writes to ensure that the debugger can read it.
+/// The only way to guarantee this in Rust is to force volatile reads and writes; the member cannot be made volatile
+impl DebugImageInfoTableHeader {
+    pub unsafe fn get_update_status(&self) -> u32 {
```
cfernald: since this is Pub, function comments (even if simple)

patina_dxe_core/src/debug_image_info_table.rs
```diff
@@ -0,0 +1,360 @@
...
+#[allow(unused)]
+pub struct EfiSystemTablePointer {
+    pub signature: u64,
+    pub efi_system_table_base: efi::PhysicalAddress,
+    pub crc32: u32,
+}
+
+/// UEFI spec defined constants for the update status field in the DebugImageInfoTableHeader
+pub const EFI_DEBUG_IMAGE_INFO_UPDATE_IN_PROGRESS: u32 = 0x1;
+pub const EFI_DEBUG_IMAGE_INFO_TABLE_MODIFIED: u32 = 0x2;
```
cfernald: Opinionated: I like putting consts like this in the impl for that associated structure so that they are coupled with their used-struct.

```rust
impl DebugImageInfoTableHeader {
    pub const UPDATE_IN_PROGRESS: u32 = 0x1;
    pub const TABLE_MODIFIED: u32 = 0x2;
...
```

patina_dxe_core/src/debug_image_info_table.rs
```diff
@@ -0,0 +1,360 @@
...
+static METADATA_TABLE: AtomicPtr<DebugImageInfoTableMetadata> = AtomicPtr::new(core::ptr::null_mut());
+
+const ALIGNMENT_SHIFT_4MB: usize = 22;
+const IMAGE_INFO_TABLE_SIZE: usize = 128; // initial size of the table
+
+static DBG_SYSTEM_TABLE_POINTER_ADDRESS: AtomicU64 = AtomicU64::new(0);
+
+// this should be an associated function, but r_efi doesn't like implementation, so we host this here
+// this returns a mutable reference to the entry at the given index in the table
+unsafe fn get_entry_mut(debug_image_info_header: &mut DebugImageInfoTableHeader, index: u32) -> &mut EfiDebugImageInfo {
```
joschock: consider something like https://doc.rust-lang.org/core/ops/trait.Index.html?

patina_dxe_core/src/debug_image_info_table.rs
cfernald, line 1: Consider moving this under the image module considering that is the only place it gets called. We need more structure to dxe_core IMO

patina_dxe_core/src/debug_image_info_table.rs
```diff
@@ -0,0 +1,360 @@
...
+}
+
+/// This function is called upon image load to create a new entry in the EFI_DEBUG_IMAGE_INFO_TABLE_GUID table.
+pub(crate) fn core_new_debug_image_info_entry(
+    image_info_type: u32,
+    loaded_image_protocol_instance: *const efi::protocols::loaded_image::Protocol,
+    image_handle: efi::Handle,
+) {
+    // This is a very funny check for null because it is working around an LLVM bug where checking is_null() or variations
+    // of that on a load of an atomic pointer causes improper code generation and LLVM to crash. So, this check is a workaround
```
joschock: maybe a link to the issue number on LLVM side so that we can know when it is safe to remove the funniness it once it is fixed?

patina_dxe_core/src/debug_image_info_table.rs
os-d: When the definition is upstreamed to r-efi, can I do a trait impl on a trait I don't own? No, right?

patina_dxe_core/src/debug_image_info_table.rs
```diff
@@ -0,0 +1,360 @@
...
+    // SAFETY: This is safe because we ensure that the index is within bounds of the table size in the caller
+    unsafe {
+        let table_ptr = debug_image_info_header.efi_debug_image_info_table as *mut EfiDebugImageInfo;
+        &mut *table_ptr.add(index as usize)
+    }
+}
+
+/// Initializes the EFI_DEBUG_IMAGE_INFO_TABLE_GUID configuration table in the UEFI system table with an empty table.
+pub(crate) fn initialize_debug_image_info_table(system_table: &mut EfiSystemTable) {
+    let initial_table =
```
cfernald: Could this be a `[EfiDebugImageInfo; IMAGE_INFO_TABLE_SIZE]` and by extension use the Box interfaces instead of pool_allocator? Might be able to avoid some unsafe later too.

patina_dxe_core/src/debug_image_info_table.rs
```diff
@@ -0,0 +1,360 @@
...
+    });
+
+    let table_ptr = Box::into_raw(debug_image_info_table_header) as *mut c_void;
+    // SAFETY: This is safe because we just allocated the table and we are going to use it immediately
+    unsafe {
+        match core_install_configuration_table(EFI_DEBUG_IMAGE_INFO_TABLE_GUID, table_ptr.as_mut(), system_table) {
+            Ok(_) => {
+                // Successfully installed the configuration table
+            }
+            Err(_) => {
```
cfernald: `if core_install_configuration_table(EFI_DEBUG_IMAGE_INFO_TABLE_GUID, table_ptr.as_mut(), system_table).is_err {...}` might be more succinct.

patina_dxe_core/src/debug_image_info_table.rs
```diff
@@ -0,0 +1,360 @@
...
+
+        // Update the table pointer and size, using the old size, as table_size is the number of entries
+        metadata_table.table.efi_debug_image_info_table = new_table;
+        metadata_table.table.table_size = old_table_size;
+    }
+
+    // size here is last_index + 1
+    // SAFETY: This is safe because we are accessing the table header and we ensure that it is initialized
+    let debug_image_info = unsafe { get_entry_mut(metadata_table.table, metadata_table.table.table_size) };
+    let address = match core_allocate_pool(efi::BOOT_SERVICES_DATA, size_of::<EfiDebugImageInfoNormal>()) {
```
cfernald: Use box & leak to avoid unsafe.

patina_dxe_core/src/debug_image_info_table.rs
```diff
@@ -0,0 +1,360 @@
...
+    };
+}
+
+/// This function is called on image unload to remove an entry from the EFI_DEBUG_IMAGE_INFO_TABLE_GUID table.
+pub(crate) fn core_remove_debug_image_info_entry(image_handle: efi::Handle) {
+    // This is a very funny check for null because it is working around an LLVM bug where checking is_null() or variations
+    // of that on a load of an atomic pointer causes improper code generation and LLVM to crash. So, this check is a workaround
+    // to check if the pointer is in the first page of memory, which is a valid check for null in this case, as we mark
+    // that entire page as invalid.
+    let metadata_table = METADATA_TABLE.load(Ordering::SeqCst);
```
cfernald: Can you just use an `Option` to avoid having to have a null pointer check at all? And for that matter avoid having to malloc something that could easily just be a global struct in-itself.

patina_dxe_core/src/debug_image_info_table.rs
os-d: I agree. I can move it.

patina_dxe_core/src/debug_image_info_table.rs
os-d: Sure, I'm preparing the PR now, but filed the issue: https://github.com/OpenDevicePartnership/patina/issues/490.

patina_dxe_core/src/debug_image_info_table.rs
os-d: I'm good with that, r-efi may not be. That may smell too close to the dreaded implementation...but I'll try it.

patina_dxe_core/src/debug_image_info_table.rs
os-d: Yeah, that's a good point. Ol' C brain working. Let me try that, should work just fine.

patina_dxe_core/src/debug_image_info_table.rs
os-d: I think I had originally done an option and that requires a static mut, right, which we are trying to avoid, so in the past I've been steered towards doing the atomics, e.g.: https://github.com/OpenDevicePartnership/patina/pull/338#discussion_r2023717356. I don't have a preference, but will leave it as is to match other patterns unless we want to change that pattern.

patina_dxe_core/src/debug_image_info_table.rs
joschock: Ah, true. You could new-type it tho.

joschock: but at that point it's an implementation detail, and I don't have strong feelings about that. resolving.

os-d: Yeah, I prefer to avoid the type wrapper unless it provides more value. But, that being said, Chris made the suggestion of making this an array, in which case this wouldn't be needed. Taking a look at that option now.

patina_dxe_core/src/debug_image_info_table.rs
os-d: Thinking about this more, I think it makes more sense to have a configuration tables module and move all of the miscellaneous configuration tables there. This isn't really image loading, but specifically the configuration tables (where it is called out in the uefi spec).

patina_dxe_core/src/debug_image_info_table.rs
os-d: I decided to leave as is

patina_dxe_core/src/config_tables.rs
```diff
@@ -0,0 +1,125 @@
...
+extern "efiapi" fn install_configuration_table(table_guid: *mut efi::Guid, table: *mut c_void) -> efi::Status {
+    if table_guid.is_null() {
+        return efi::Status::INVALID_PARAMETER;
+    }
+
+    let table_guid = unsafe { *table_guid };
+    let table = unsafe { table.as_mut() };
+
+    let mut st_guard = SYSTEM_TABLE.lock();
+    let st = st_guard.as_mut().expect("System table support not initialized");
```
cfernald: should this instead return an error?

patina_dxe_core/src/config_tables.rs
```diff
@@ -0,0 +1,125 @@
...
+    systemtables::{EfiSystemTable, SYSTEM_TABLE},
+};
+
+extern "efiapi" fn install_configuration_table(table_guid: *mut efi::Guid, table: *mut c_void) -> efi::Status {
+    if table_guid.is_null() {
+        return efi::Status::INVALID_PARAMETER;
+    }
+
+    let table_guid = unsafe { *table_guid };
+    let table = unsafe { table.as_mut() };
```
cfernald: what is the value of turning it into a reference since the type and size is unknown. It will be a pointer from end-to-end and not be dereferenced, right?

patina_dxe_core/src/config_tables.rs
os-d: Not my code :). Just copied here, but updated.

patina_dxe_core/src/config_tables.rs
os-d: Again, didn't write this, but agreed, this is an annoying property. I will change this to take a ptr.

PR #494, opened 2025-06-05
Handle special interrupt IDs for hardware interrupt handler

kuqin12:
## Description

Current implementation did not take into account the special interrupt IDs (1021, 1022, and 1023). Thus when such interrupts occur, the `raw_value` will be used as an index into the handler vector, causing a panic.

This change is to handle the special interrupt IDs by checking if the `raw_value` is a special ID, otherwise it needs to be within the range of the handler vector.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This was tested on physcial hardware and booted to Windows.

## Integration Instructions

N/A

### DIFF
--- a/patina_dxe_core/src/allocator.rs
+++ b/patina_dxe_core/src/allocator.rs
@@ -1004,6 +1004,12 @@ pub fn install_memory_services(bs: &mut efi::BootServices) {
     bs.get_memory_map = get_memory_map;
 }
 
+// Resets the ALLOCATOR map to empty and resets the static allocators for test purposes.
+#[cfg(test)]
+pub(crate) unsafe fn reset_allocators() {
+    ALLOCATORS.lock().reset();
+}
+
 #[cfg(test)]
 mod tests {
 
--- a/patina_dxe_core/src/memory_attributes_table.rs
+++ b/patina_dxe_core/src/memory_attributes_table.rs
@@ -273,6 +273,7 @@ mod tests {
 
             unsafe {
                 test_support::init_test_gcd(None);
+                test_support::reset_allocators();
                 init_system_table();
             }
             f();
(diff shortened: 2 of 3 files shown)


kuqin12:
Technically I should pull in the updated crate for arm-gic, which has the definition for those magic numbers. But those releases contain breaking changes, such as initializing the gic with number of cores, etc, which needs more caution to proceed than a simple bug fix.

I will follow up with a sperate PR to update arm-gic once this change is checked-in to fix the system behavior.

patina_dxe_core/src/hw_interrupt_protocol.rs
```diff
@@ -313,13 +313,25 @@ impl InterruptHandler for HwInterruptProtocolHandler {
     fn handle_interrupt(&'static self, exception_type: usize, context: &mut ExceptionContext) {
         let int_id = GicV3::get_and_acknowledge_interrupt();
         if int_id.is_none() {
```
Javagedes, line 315: I'd suggest something like this:

```rust
let raw_value: u32 = match int_id {
    Some(1021) | Some(1022) | Some(1023) | None => {
        return; // The special interrupt does not need to be acknowledged
    }
    Some(v) if v as usize >= self.handles.lock().len() => {
        log::error!("Invalid interrupt source: 0x{:x}", v);
        return;
    }
    Some(v) => v,
};
```

Javagedes: Also, why do you specify raw_value as a u32? You never use it as such. You only cast it toa usize or a u64

kuqin12: This is because the int_id is not a Some(number), but a Some(IntId). I think we will need a int_id.into(), which is u32. So the suggestion as is will not work. Any suggestions to make it like your style?

Javagedes: `let raw_value: u32 = match int_id.map(|id| id.into::<u32>()) { ... }` should work. It will convert it from a `Option<IntId>` to an `Option<u32>` before performing the `match`

Javagedes: and you could maybe swap the u32's for usize or u64 or something. that is up to you.

kuqin12: hmm, `end_interrupt` later on will need the IntId to work... So I guess pull out the IntId from some and check the raw_value needs to be separate?

Javagedes: I think you can just do `GicV3::end_interrupt(raw_value);` or `GicV3::end_interrupt(raw_value.into());` if you kept the raw_value type as a u32. The Into/From trait is implemented to convert between IntId and u32.

Javagedes: If that does not work and more changes are needed to make the match statement work, then it's probably best to keep your PR the way it is. No need to add additional complexity just to use a match statement.

kuqin12: I was going to reply. It did not work. Probably missing some punctuations here and there ;(

I will get back to this when I update the arm-gic crate.

PR #497, opened 2025-06-05
Replace cargo test with nextest for better test runner UI and results

vineelko:
## Description

- Tests are still executed as `cargo make test` no change here!
- Makefile test task now uses [nexte.st](https://nexte.st/) which offers a much-improved test runner experience
- Provides more insightful test results
  - Including the number of tests executed
  - Test execution time
  - Improved execution speed
  - Call stack on failures
- Makefile test task is improved to directly accept any additional parameters like providing individual test names or `--nocapture` arguments etc.
   - ex: `cargo make test boot_services::test::test_open_protocol`
---
 
- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

rust-optimized

## Integration Instructions

NA

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -113,8 +113,7 @@ run_task = [{ name = ["check_no_std", "check_std"], parallel = true }]
 description = "Builds all rust tests in the workspace. Example `cargo make test`"
 clear = true
 command = "cargo"
-args = ["test", "@@split(INDIVIDUAL_PACKAGE_TARGETS, )", "@@split(TEST_FLAGS, )"]
-dependencies = ["individual-package-targets"]
+args = ["nextest", "run", "@@split(TEST_FLAGS, )", "${@}"]
 
 [tasks.coverage]
 description = "Build and run all tests and calculate coverage."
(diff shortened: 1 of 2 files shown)


Javagedes:
Personally, I'm not a huge fan of adding another tool someone must download locally just to have a better UI.

I would prefer to make this an optional tool rather than a mandatory one (so let it have its own task).

If you want to use it in CI, I have no problem with that.

Also you'll need to update docs.

Javagedes:
Additionally, our cargo make test wrapper just runs _only_ cargo test under the hood (only difference is we set RUSTC_BOOTSTRAP for the user)

So I think this can just be one of those things that if a developer wants to use it, they can. I don't think we should force it on them.

cfernald:
While I do like the look/perf of this, I see Joey's point around another tool but don't personally have a strong opinion on this one. I would at least like to have this option, but will recuse myself from the "required" nature conversation.

makubacki:
I agree that this should not be required. I think it'd be fine as a separate make task with docs updated to explain how to use it.

vineelko:
Thanks for the feedback. I understand that there is some interest in making this an alternative task to `cargo make test` or as an option in CI. However, I’d prefer to refrain from doing that, as it introduces yet another way to run tests, potentially causing unnecessary confusion for users.

I am closing this PR.

Issue #492, opened 2025-06-04
[Feature]: Consider Downstream Build Options for Patina Consumers


os-d:
### Feature Overview

edk2 has a unique build infrastructure that defines all toolchain options for consumers of it, which, while creating some pain points, does also unify the build landscape and have fewer odd build configurations come out of the woodwork. Patina does not have any such concept and may be at risk of random platform build options causing unexpected behavior/output.

### Solution Overview

Consider how Patina should evangelize build options, this may be:
- nothing, this is like most other build systems you are on your own
- Write up in the md book what recommendations are (runs risk of platforms not noticing)
- Construct something for platforms to consume to define build options
- Something else

At the very least, we should document the decision made here.

### Urgency

Medium

### Are you going to implement the feature request?

Someone else needs to implement the feature

### Do you need maintainer feedback?

No maintainer feedback needed

### Anything else?

@makubacki may be interesting to discuss in a future Rust meeting.

Javagedes:
One option is to write a cli tool that sets all the build flags for someone when building. That way they don't require having to copy our cargo make config file, or their own custom command overrides, which can be annoying.

```
> cargo install patina
> patina build
> patina test
> patina ...
```

makubacki:
@os-d, can you please provide more detail about the ask? What exactly do you mean by "build options" here?

os-d:
@makubacki , we were having some offline discussions on something that fills an equivalent function as tools_def.template in edk2, i.e. we are somewhat unique as a project that requires other projects to build us not in our repo (most projects get built within the repo they originated in and so use makefiles, etc. there).

If we don't have a standardized set of build options that a platform uses to build with, we are likely to see spurious errors and build failures. I don't know in the Rust world what the best way to do this is, whether build.rs files (which I'm told can be included by a consumer), the tool Joey is suggesting, or something else, but the overall goal is to create a uniform build environment for platforms to minimize noise from inconsistent compiler/linker options.

Javagedes:
I continue to think my suggestion of having a very small (NOT BUILDTOOLS) wrapper around cargo is a good option. This is also basically what @vineelko suggested a while ago with "xtask".

We could also use the main `patina` crate as that command line tool.

[...]

PR #503, opened 2025-06-11
RFC: Switch from cargo-make to xtask

vineelko:
## Description

This RFC outlines the limitations of using `cargo-make` setup and discusses how
adopting [`xtask`](https://github.com/matklad/cargo-xtask) can provide a more
flexible and maintainable build and tooling infrastructure.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Proofread (but not authored) by Copilot. Also have a draft implementation [here](https://github.com/OpenDevicePartnership/patina/compare/main...users/vineelko/xtask_0610)

https://github.com/user-attachments/assets/98989220-7ad6-4564-b6ab-f5cad86cb612

## Integration Instructions

NA

### DIFF
--- a/docs/src/rfc/text/0000-switch-from-cargo-make-to-xtask.md
+++ b/docs/src/rfc/text/0000-switch-from-cargo-make-to-xtask.md
@@ -1,0 +1,377 @@
+# RFC: Switch from cargo-make to xtask
+
+This RFC outlines the limitations of using `cargo-make` setup and discusses how
+adopting [`xtask`](https://github.com/matklad/cargo-xtask) can provide a more
+flexible and maintainable build and tooling infrastructure.
+
+## Change Log
+
+- 2025-06-11: Initial draft of the RFC.
+
+## Motivation
+
+Patina and its associated repositories have been using `cargo-make` as the build
+task runner since the beginning. While its TOML-based syntax is suitable for
+simple tasks, more complex workflows often require using its custom scripting
+language, `duckscript`. These include scenarios like parsing arguments as
+package arguments, passing values through environment variables, or the patch
+build task from other patina repo. Even basic actions, such as listing available
+tasks with descriptions (`cargo make --list-all-steps`), are not
+straightforward, and the output is not particularly intuitive. `cargo-make`
+supports task dependencies, but its flexibility is limited, and writing
+non-trivial workflows in `Makefile.toml` can become tedious over time.
+
+`cargo make all` is our current recommended usage, but it does not manage the
+installation of all prerequisite tools, such as `cargo-make` itself,
+`cargo-tarpaulin`, `cargo-deny`, and `cspell` which must be installed manually.
+
+The need for a better task runner became apparent while working on the following
+PR [#497](https://github.com/OpenDevicePartnership/patina/pull/497), where we
+aimed to introduce a more actionable testing tool—`nextest`. The current setup
+relies on `cargo test`, whose output is subpar compared to what `nextest`
+offers. However, since `nextest` is an external tool that must be installed
+manually, the PR understandably received some pushback. I think not using the
+right tool holds us back. It makes it harder to improve the developer
+experience, and more importantly, we miss the chance to make test results more
+actionable. To prove the point, using `nextest` we were able to uncover faulty
+test cases that had been incorrectly passing. These were later identified and
+fixed in [#491](https://github.com/OpenDevicePartnership/patina/pull/491) and
+[#495](https://github.com/OpenDevicePartnership/patina/pull/495). Setting up of
+any such additional prerequisite tools should be automated and included as the
+first command a user runs—ideally as a one-time setup step.
+
+While `cargo-make` covers most of our current needs, setting up environment
+variables and crafting the correct command-line arguments for tasks has become
+increasingly cumbersome. On top of that, Even something as simple as rerunning a
+failed test by name is difficult due to the current rigid conventions imposed in
+the `Makefile.toml`.
+
+```cmd
+C:\r\patina>cargo make test -p patina_dxe_core gcd::tests::test_full_gcd_init
+...
+[cargo-make] INFO - Execute Command: "cargo" "test" "-p" "-p;patina_dxe_core;gcd::tests::test_full_gcd_init"
+error: invalid character `;` in package name: `;patina_dxe_core;gcd`, the first character must be a Unicode XID start character (most letters or `_`)
+[cargo-make] ERROR - Error while executing command, exit code: 101
+
+C:\r\patina>cargo make test patina_dxe_core gcd::tests::test_full_gcd_init
+...
+[cargo-make] INFO - Execute Command: "cargo" "test" "-p" "patina_dxe_core;gcd::tests::test_full_gcd_init"
+error: invalid character `;` in package name: `patina_dxe_core;gcd`, characters must be Unicode XID characters (numbers, `-`, `_`, or most letters)
+[cargo-make] ERROR - Error while executing command, exit code: 101
+```
+
+The primary motivation for choosing `xtask` is to establish a clear and
+actionable workflow for building, testing, and running Patina, while also
+future-proofing our setup. This not only improves the onboarding experience but
+also maintains flexibility for advanced users who need to customize execution
+through additional command-line arguments or extend workflows without rigid
+configuration constraints.
+
+   > "There should be one and preferably only one obvious way to do it." -Zen of Python
+
+## Who is using `xtask`?
+
+An `xtask` based build setup is used by many well-known Rust projects,
+including:
+
+- [cargo](https://github.com/rust-lang/cargo/blob/master/crates/xtask-build-man/src/main.rs)
+- [rust-analyzer](https://github.com/rust-lang/rust-analyzer/tree/master/xtask)
+- [uefi-rs](https://github.com/rust-osdev/uefi-rs/tree/main/xtask)
+- [openvmm](https://github.com/microsoft/openvmm/tree/main/xtask)
+- [ratatui](https://github.com/ratatui/ratatui/tree/main/xtask)
+- [IronRDP](https://github.com/Devolutions/IronRDP/tree/master/xtask)
+- [microbit](https://github.com/nrf-rs/microbit/tree/main/xtask)
+- many others across the Rust ecosystem.
+
+## Technology Background
+
+Even though `cargo-make` is a capable build task runner, `xtask` presents a
+significantly more flexible and maintainable alternative. Before diving into how
+we can leverage it, it’s important to clear up some common misconceptions about
+what `xtask` actually is:
+
+1. **Not an External Tool**: Unlike `cargo-make`, `xtask` is not a third-party
+   tool or crate that we install from crates.io. This means it introduces
+   **zero additional dependencies** upfront.
+
+2. **Convention**: `xtask` is simply a **convention** for organizing a custom
+   binary crate inside the Rust workspace. The structure typically looks like
+   this:
+
+   ```text
+   patina/
+   ├── .cargo/config.toml     # Defines alias: xtask = "run --package xtask --"
+   ├── xtask/                 # A binary crate containing build CLI logic
+   │   └── src/main.rs
+   ├── patina-specific-crates/
+   └── Cargo.toml             # Includes xtask in [workspace] members
+   ```
+
+3. **Standard Rust Code**: The `xtask` crate is just another Rust binary that
+   uses standard library features, particularly `std::process::Command`, to
+   orchestrate build operations by spawning OS-level processes.
+
+4. **Cargo Alias Integration**: To make running `xtask` feel native, it's
+   recommended to define a Cargo alias like so:
+
+   ```toml
+   [alias]
+   xtask = "run --package xtask --"
+   ```
+
+   This enables commands like `cargo xtask ...`, where everything after `xtask`
+   is passed as arguments to the `xtask` binary.
+
+5. **Execution**: When we run `cargo xtask ...`, it effectively invokes
+   `target/debug/xtask.exe ...`.
+
+6. **Naming**: There’s nothing special about the name `xtask`. We could name
+   the crate `patina` or even `make` though the latter is discouraged to avoid
+   confusion. That said, `xtask` has become the **conventionally accepted name**
+   used by many Rust projects.
+
+7. **Fully Customizable in Rust**: Since the build orchestration logic lives in
+   real Rust code, it's type-safe, testable, and easier to evolve over time
+   unlike TOML-based configurations or shell scripts.
+
+8. **No Custom Tooling Sections Needed**: We don’t need `[tool]` sections in
+   TOML or separate shell scripts to install tools. Instead, we can simply run
+   `cargo xtask setup`, where `setup` is a custom task we author to handle
+   installations.
+
+9. **Unified Local and CI Workflows**: With `xtask`, We can streamline the local
+   and CI workflows, ensuring they behave consistently with minimal duplication
+   or special cases.
+
+10. **Startup and Performance**: Since `xtask` is purpose-built for the
+    specific repository, it avoids the overhead of generic task runners like
+    `cargo-make`. This specialization makes it faster to start up and execute
+    tasks, as it's compiled along with the workspace and tailored to just what
+    the repo needs, no plugin resolution, runtime scripting, or extra setup.
+
+## Goals
+
+1. Ensure that new users can set up the prerequisite tools and environment for
+   Patina with minimal effort and a consistent workflow across platforms and
+   architectures.
+2. Being able to author build tasks more reliably and make them easy to extend
+   over time.
+3. Eliminate stale `Makefile.toml` tasks.
+4. Future-proof Patina’s build and tooling infrastructure.
+5. Each Patina repo has slightly different build requirements, so each repo can
+   adopt the conventions used in this one. However, trying to completely reuse
+   the `xtask` across repos can be over-engineering and make it hard to evolve.
+   The spirit of `xtask` is in many ways similar to traditional Makefiles -
+   Custom implementation for each repo but with the added benefit of improved
+   reliability and maintainability by authoring them in Rust, not necessarily
+   aiming for reusability.
+
+## Requirements
+
+1. Maintain a one-to-one mapping with the current `cargo-make` `<tasks>` to
+   ensure a smooth transition.
+2. Update documentation accordingly, though we expect only minimal changes to
+   the existing workflow.
+
+## Unresolved Questions
+
+## Current Build Setup
+
+Here are the build tasks currently defined in the `Makefile.toml` at the root of
+the Patina workspace:
+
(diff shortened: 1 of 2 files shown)


docs/src/rfc/text/0000-switch-from-cargo-make-to-xtask.md
```diff
@@ -0,0 +1,377 @@
+# RFC: Switch from cargo-make to xtask
+
+This RFC outlines the limitations of using `cargo-make` setup and discusses how
+adopting [`xtask`](https://github.com/matklad/cargo-xtask) can provide a more
```
Javagedes, line 4: I would mention that this is a pattern, not a tool or crate. Maybe lightly describe the pattern (a one liner) here too or give examples (such as probe-os's implementation

docs/src/rfc/text/0000-switch-from-cargo-make-to-xtask.md
```diff
@@ -0,0 +1,377 @@
...
+language, `duckscript`. These include scenarios like parsing arguments as
+package arguments, passing values through environment variables, or the patch
+build task from other patina repo. Even basic actions, such as listing available
+tasks with descriptions (`cargo make --list-all-steps`), are not
+straightforward, and the output is not particularly intuitive. `cargo-make`
+supports task dependencies, but its flexibility is limited, and writing
+non-trivial workflows in `Makefile.toml` can become tedious over time.
+
+`cargo make all` is our current recommended usage, but it does not manage the
+installation of all prerequisite tools, such as `cargo-make` itself,
```
Javagedes, line 25: This is not quite right. It does auto install cargo / rust tools by default, but we disabled it because installing tools in our repo is unstable and often leads to build errors due to our project not always being on the latest rust version. When a dependency of one of our build tools updates to require, say 1.87, but we are on 1.85, the tool installation fails. It also prevents re-produceable builds because it attempts to install the latest. So if we have PLE come in 5 years later, its going to install the latest version, which may no longer be compatible, which is no Beuno

We will hit the same issue with this custom tool, but I think we can probably get around it with some effort.

docs/src/rfc/text/0000-switch-from-cargo-make-to-xtask.md
```diff
@@ -0,0 +1,377 @@
...
+fixed in [#491](https://github.com/OpenDevicePartnership/patina/pull/491) and
+[#495](https://github.com/OpenDevicePartnership/patina/pull/495). Setting up of
+any such additional prerequisite tools should be automated and included as the
+first command a user runs—ideally as a one-time setup step.
+
+While `cargo-make` covers most of our current needs, setting up environment
+variables and crafting the correct command-line arguments for tasks has become
+increasingly cumbersome. On top of that, Even something as simple as rerunning a
+failed test by name is difficult due to the current rigid conventions imposed in
+the `Makefile.toml`.
```
Javagedes, line 47: While I agree the xtask pattern is better, I don't think this is a valid reason why. We already partially do this with build command where we pass extra arguments directly to the underlying command.

`cargo make build -p MyCrate` -> `cargo build <all our special flags> -p MyCrate`.

 I've just never had anyone mention they want this added to cargo test so I never added it.

docs/src/rfc/text/0000-switch-from-cargo-make-to-xtask.md
```diff
@@ -0,0 +1,377 @@
...
+1. **Not an External Tool**: Unlike `cargo-make`, `xtask` is not a third-party
+   tool or crate that we install from crates.io. This means it introduces
+   **zero additional dependencies** upfront.
+
+2. **Convention**: `xtask` is simply a **convention** for organizing a custom
+   binary crate inside the Rust workspace. The structure typically looks like
+   this:
+
+   ```text
+   patina/
```
Javagedes, line 102: I don't think the typical xtask pattern format will work for our use case. Our code sits in multiple crates and all platforms will want to use our exact tooling for easy builds per #492 

I personally think we want to use the `patina` crate name as the command line tool / wrapper for all cargo commands. Locally for the patina repo, we can use an alias `patina = run --package patina --`, but we need to publish this to a feed so that other crates (probably mainly platform crates, and our patina-qemu-dxe-core repo) can do `cargo install patina` and then run `patina <subcommands>`

docs/src/rfc/text/0000-switch-from-cargo-make-to-xtask.md
```diff
@@ -0,0 +1,377 @@
...
+### Get help for existing tasks
+
+```cmd
+C:\r\patina>cargo xtask help
+Usage: `cargo xtask <task> [options]`
+Tasks are run in the root of the repository.
+
+Tasks:
+all           Run all task before drafting a PR
+build-aarch64 Build the project for aarch64
```
Javagedes, line 353: I think we can keep build as a single command line option, with some flags for build std x64, aarch64

docs/src/rfc/text/0000-switch-from-cargo-make-to-xtask.md
```diff
@@ -0,0 +1,377 @@
...
+6. **Naming**: There’s nothing special about the name `xtask`. We could name
+   the crate `patina` or even `make` though the latter is discouraged to avoid
+   confusion. That said, `xtask` has become the **conventionally accepted name**
+   used by many Rust projects.
+
+7. **Fully Customizable in Rust**: Since the build orchestration logic lives in
+   real Rust code, it's type-safe, testable, and easier to evolve over time
+   unlike TOML-based configurations or shell scripts.
+
+8. **No Custom Tooling Sections Needed**: We don’t need `[tool]` sections in
```
Javagedes, line 137: We still need a [tool] section, or an alternative. We should not control any dependency's versions for a platform, even CI related dependencies. That always backfires. an "alternate" is using external dependency files - e.g. exactly what `stuart_update` does. And I don't know if anyone truly likes `stuart`.

docs/src/rfc/text/0000-switch-from-cargo-make-to-xtask.md
```diff
@@ -0,0 +1,377 @@
...
+   TOML or separate shell scripts to install tools. Instead, we can simply run
+   `cargo xtask setup`, where `setup` is a custom task we author to handle
```
Javagedes, line 139: This is starting to sound a lot like `stuart_update` and i don't know if anyone truly likes `stuart`

joschock:
> Overall I agree that this is good and probably useful.
> 
> But you are very very close to reinventing `stuart`. Especially when you are going to have to start considering platform customizations that they will eventually request.

Agree. One of the most serious critiques of EDK2 is the complexity of its highly customizable build system.  This is not relevant to delta between cargo make and xtask - I could go either way on that question, and I am attracted to xtask if it is more performant and obvious. This is more a comment that powerful tools in the build system tend to tempt complexity. If the motivation is "I found that cargo make was limiting me from this complicated thing I want to do in the build system", then my skepticism increases.

docs/src/rfc/text/0000-switch-from-cargo-make-to-xtask.md
vineelko: I think pattern is the right word. Let me update it in the next iteration.

docs/src/rfc/text/0000-switch-from-cargo-make-to-xtask.md
vineelko: Joey, could you point me to where we disabled that functionality in `cargo-make`?

My main point in that paragraph is that we don't currently have a `setup` task in the `Makefile.toml` to handle installing the required tools with **pinned version numbers**. Additionally, this `setup` task is can be invoked independently.

Also I am with you, I don't want the `setup` task to randomly install the latest tools. Thats a bad idea. Here is the draft [implementation](https://github.com/OpenDevicePartnership/patina/commit/f8635efb58edfcf74b41538ea071d21f4084872c#diff-7b5143e7b8ca52e53a2b8aedfc844b7ec9cd160f28f10841613c2ed661cb0212R206-R238) of installing tools to the exact version numbers

docs/src/rfc/text/0000-switch-from-cargo-make-to-xtask.md
vineelko: Absolutely, we do have some support for that in the build task, and I agree that what we have today is sufficient for our current needs. However, in general, handling slightly more advanced use cases (like I called out in the RFC for testing) can become tedious—especially since it's difficult to manage all the possible combinations of arguments using `cargo-make` variables that are created by dependent tasks. Even if we do that, for instance, things can easily break if we need to pass an additional `-Z` flag.

One thing to note is that `xtask` isn’t doing anything particularly clever here either—but it does offer more flexibility (since the control flow is in code rather than in data) by simply allowing additional arguments to pass through the task command. Of course, we can achieve something similar in `Makefile.toml` using `"${@}"` and by not processing anything on the command line.

docs/src/rfc/text/0000-switch-from-cargo-make-to-xtask.md
vineelko: Yes, we can use the same `xtask` patterns in each of the three repositories (`patina`, `patina-qemu`, `patina-dxe-core-qemu`), but I strongly discourage us from turning this into a shared crate and reusing it across repositories. It’s not that we *can’t* do it, but I believe this is one area where having some loose coupling between the repos benefits us by allowing each to evolve its own needs independently.

That said, I’m not sure the requirements are drastically different right now b/w the repos. All I’m suggesting is that we avoid designing for reusability before this approach is even sufficiently tested. If, in the future, we find real opportunities for reuse, we can always revisit the idea of turning it into a crate and publishing it to crates.io.

Also, I want the tooling to feel familiar to other Rust projects, while still differentiating itself where it makes sense. So, I'm still leaning towards using `cargo xtask ...`. This has two benefits, 
1. **Familiarity**:  This is **just** the same xtask pattern
2.  **Discoverability**: Users can easily discover what this xtask has to offer using `cargo xtask help`

Making it `patina` command can defeat this purpose no matter how well we document it. Atleast this was my experience. 

Below is what I have in mind, A new user can come in,

Go to `C:\r\patina`
do `cargo xtask help`
do `cargo xtask setup`
do `cargo xtask build`
do `cargo xtask test`
do `cargo xtask all...`  <-- before pushing to github

Go to `C:\r\patina-dxe-core-qemu`
do `cargo xtask help`
do `cargo xtask setup`
do `cargo xtask build`
do `cargo xtask test`
do `cargo xtask all...`  <-- before pushing to github

docs/src/rfc/text/0000-switch-from-cargo-make-to-xtask.md
vineelko: Yaa that can be done.

docs/src/rfc/text/0000-switch-from-cargo-make-to-xtask.md
vineelko: Maybe I went a little too far in claiming we don’t need anything. What I actually meant was that the tools (pinned to specific versions) can live within the `setup` task.[As done here](https://github.com/OpenDevicePartnership/patina/commit/f8635efb58edfcf74b41538ea071d21f4084872c#diff-7b5143e7b8ca52e53a2b8aedfc844b7ec9cd160f28f10841613c2ed661cb0212R206-R216). Let me make that clear in the next iteration.

vineelko:
> But you are very very close to reinventing stuart. Especially when you are going to have to start considering platform customizations that they will eventually request.

Interesting! Trust me. I have never thought of stuart in my mind, not even once when coming up with this RFC 😄

But here is my motivation, We need a way to unblock ourselves from any future build/tooling needs for patina(just the current repo). Being able to freely adapt any new tools that could enhance/simplify workflow **without breaking the dev inner loop experience**. Nothing beyond that. I might have to sync with you on what you meant by _platform customizations_

I am not sure what is possible with `stuart` and any current downsides associated with it. Just because I haven't played much with it.

Also, in response to John's comment, `xtask` is being evaluated not to make the existing build system more complex, but quite the opposite—to make it simpler and more intuitive for users. At the same time, it aims to provide greater flexibility (as outlined in the Motivation section).

os-d:
> > Overall I agree that this is good and probably useful.
> > But you are very very close to reinventing `stuart`. Especially when you are going to have to start considering platform customizations that they will eventually request.
> 
> Agree. One of the most serious critiques of EDK2 is the complexity of its highly customizable build system. This is not relevant to delta between cargo make and xtask - I could go either way on that question, and I am attracted to xtask if it is more performant and obvious. This is more a comment that powerful tools in the build system tend to tempt complexity. If the motivation is "I found that cargo make was limiting me from this complicated thing I want to do in the build system", then my skepticism increases.

I completely agree that we should not invent a bespoke build system a la edk2 and then wrapping it later in stuart. This actually has a fundamental difference than stuart though, in that it is a commonly used pattern in the rust community within the standard build system. So I feel less concerned that this is "stuart-like" because other folks use this pattern (including folks who are likely to interact with patina, e.g. openvmm). I think the big problem with stuart was that it was something new to folks and no one likes something new. But if this isn't new, but a commonly used pattern, I don't think it is as serious an issue.

Now, that being said I do agree there will be platform asks to add functionality, but I don't think that changes w.r.t xtask vs cargo make, no matter what platforms will want my_favorite_build_option to be available.

All of this being said, I don't have a large opinion here, I just want to achieve the goals of using intuitive (insofar is possible with build systems), fast, and standard build systems. I think this is a reasonable approach and less arcane than makefiles, but if folks see issues with the approach, I'm okay to stay with the current system and iterate differently.

I think there is the question of how this does or doesn't intersect with the issue I posted on standardizing build options for consumers.

docs/src/rfc/text/0000-switch-from-cargo-make-to-xtask.md
makubacki, line 1: This is a reasonable proposal, but my primary concerns are:

1. I'd like our build configuration to be as simple* as possible (where _simple_ is subjective).
2. I'd like us to spend time working on Rust firmware code and not extending/maintaining custom build code that tends to get more complex over time.

These points have been expressed in various forms so far in the RFC.

_\* I look at **simpler** here as leveraging an upstream build project handles complex build logic. We mostly use declarative data to describe our build using their well-documented infrastructure. We extend with complex logic using the hooks they provide (Python, etc.) when necessary. Otherwise, we risk diverting further and further into more complex build environment code that makes our environment more unique and, in the end, more difficult to maintain._

As a firmware team in the Rust community, I'd like to identify tangible concerns with `cargo-make` first and see if we can simply address them by contributing directly to `cargo-make`. This will allow us to focus our efforts on firmware.

### Cargo Make Handling Complexity

`cargo-make` support languages other than `duckscript` ([Other Programming Languages](https://github.com/sagiegurari/cargo-make?tab=readme-ov-file#other-programming-languages)). We can simply write complex logic in Python if we'd like.

For most logic, the declarative nature of the TOML file condenses logic, reduces potential for error, and simplifies maintenance.

So, we could go down a path of building a build environment in Rust which will lead to library crates being shared, etc. but I'm arguing that `cargo-make` is that tool and our extension point is Python where necessary.

### Installing Pre-Requisite Tools

`cargo-make` can automatically install cargo plugins as described in [Cargo Plugins](https://github.com/sagiegurari/cargo-make?tab=readme-ov-file#installing-dependencies).

We can easily write wrappers to install binaries using `cargo-binstall` ([and have](https://github.com/OpenDevicePartnership/patina/blob/main/.github/actions/rust-tool-cache/action.yml)).

### `cargo-make` as a 3rd-Party Dependency

Replacing `cargo-make` does not eliminate our dependency on third-party plugins. We still have:

- `cargo-deny`
- `cargo-tarpaulin`
- `cargo-release`
- `mdbook`

### `cargo-nextest` as an Improvement for Testing

This is irrespective of `xtask` versus `cargo-make`. While we don't want this to be a **required** tool locally. I don't see why it couldn't always be run in PR gates / CI and then users can repro locally if there is a problem uniquely found by that they need to fix or they just have a preference for using it.

---

The RFC raises good points for conversation, but my current stance is that we should evaluate how to better leverage `cargo-make` (contributing where necessary) as a first step and it's a bit premature to jump to writing our build code in Rust and abandoning the benefits it provides.

vineelko: Thanks for the feedback and for pointing me to the Cargo Plugins documentation section.

However, I don't completely agree with the following aspects:

1. Complex scripts requiring yet another programming language (e.g., Python and its dependencies).
2. Contributing to `cargo-make` when it doesn’t cater to our needs (which I presume should be rare, considering the scripts mentioned above).
3. Writing wrappers or scripts on top of existing tools.

These are the scenarios this RFC is trying to avoid.

That said, I’m fine with leveraging `cargo-make` as it stands today for our needs. At the very least, this RFC captured what `xtask` is, what it’s capable of (both the positives and the negatives), potential alternative to `cargo-make`, and the feedback from the team.

I’ll go ahead and close this RFC for now. We can revisit it if needed.

vineelko:
> > > Overall I agree that this is good and probably useful.
> > > But you are very very close to reinventing `stuart`. Especially when you are going to have to start considering platform customizations that they will eventually request.
> > 
> > 
> > Agree. One of the most serious critiques of EDK2 is the complexity of its highly customizable build system. This is not relevant to delta between cargo make and xtask - I could go either way on that question, and I am attracted to xtask if it is more performant and obvious. This is more a comment that powerful tools in the build system tend to tempt complexity. If the motivation is "I found that cargo make was limiting me from this complicated thing I want to do in the build system", then my skepticism increases.
> 
> I completely agree that we should not invent a bespoke build system a la edk2 and then wrapping it later in stuart. This actually has a fundamental difference than stuart though, in that it is a commonly used pattern in the rust community within the standard build system. So I feel less concerned that this is "stuart-like" because other folks use this pattern (including folks who are likely to interact with patina, e.g. openvmm). I think the big problem with stuart was that it was something new to folks and no one likes something new. But if this isn't new, but a commonly used pattern, I don't think it is as serious an issue.
> 
> Now, that being said I do agree there will be platform asks to add functionality, but I don't think that changes w.r.t xtask vs cargo make, no matter what platforms will want my_favorite_build_option to be available.
> 
> All of this being said, I don't have a large opinion here, I just want to achieve the goals of using intuitive (insofar is possible with build systems), fast, and standard build systems. I think this is a reasonable approach and less arcane than makefiles, but if folks see issues with the approach, I'm okay to stay with the current system and iterate differently.
> 
> I think there is the question of how this does or doesn't intersect with the issue I posted on standardizing build options for consumers.

I did some experiments after chatting with you offline. This RFC does not solve [this issue #492](https://github.com/OpenDevicePartnership/patina/issues/492) and would require a completely different discussion, given the way Cargo currently handles profiles between binary and library crates.

vineelko:
Based on the feedback from the team, we would, by and large, still prefer to use `cargo-make` as our build/task runner.

Hence, I’m closing this RFC. Thanks again for the feedback.

PR #507, opened 2025-06-12
[REBASE & FF] AdvLogger: Add Theory of Operation 

cfernald:
## Description

Adds the initial theory of implementation for advanced logger as well as some minor interfaces improvements.

Not marking breaking as none of the removed public APIs were intended and were unused.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

N/A

## Integration Instructions

N/A

### DIFF
--- a/docs/src/dxe_core/advanced_logger.md
+++ b/docs/src/dxe_core/advanced_logger.md
@@ -1,0 +1,158 @@
+# Advanced Logger
+
+Patina uses the [Log crate](https://docs.rs/log/latest/log/) to facilitate logging
+and the platform code is responsible for configuring its desired log implementation
+prior to executing the Patina core using the standard `log::set_logger` routine.
+
+The platform may choose any implementation of the `Log` trait it wishes; however,
+Patina provides a robust serial and memory logging solution called Advanced Logger, which
+is designed to be compatible with the [Project Mu Advanced Logger](https://microsoft.github.io/mu/dyn/mu_plus/AdvLoggerPkg/Docs/ReadMe/).
+This compatibility means that Patina Advanced Logger can be used in combination
+with Project Mu Advanced Logger to provide a memory log that spans multiple phases
+(e.g. PEI and DXE) and can be parsed with existing tools.
+
+Patina Advanced Logger is implemented in two parts: the `log::Log` implementation called
+[`AdvancedLogger`](https://github.com/OpenDevicePartnership/patina/blob/main/components/patina_adv_logger/src/logger.rs),
+and the component called [`AdvancedLoggerComponent`](https://github.com/OpenDevicePartnership/patina/blob/main/components/patina_adv_logger/src/component.rs).
+The former is responsible for handling log messages and forwarding them to both the serial
+transport and the memory log as appropriate. The latter is responsible for
+implementing the [Advanced Logger Protocol](#advanced-logger-protocol) and configuring
+the logger based on HOB input to Patina.
+
+## Logger
+
+The logger is designed to be statically instantiated as required by the log crate
+and is responsible for handling the log messages forwarded through the logging interfaces
+such as `log::info!()`. Logger initialization acts as the primary interface for platform
+configurations such as serial interfaces, log level limits, message filtering,
+etc.
+
+The `Log` implementation will send all messages to both the configured serial port
+and its memory log to be picked up later by components in DXE or at runtime.
+This splits all messages to be both output to the serial console and stored in
+the memory log, depending on configuration. More messages that be sent to the memory
+log then the serial port depending on the [message filtering](#message-filtering) used.
+
+```mermaid
+flowchart LR
+    patina["Patina"]
+    log["Log Crate"]
+    logger["Advanced Logger"]
+    serial[["SerialIO"]]
+    memory_log[["Memory Log"]]
+    patina -- log::info!('foo') --> log
+    log --> logger
+    logger -- 'foo' --> serial
+    logger -- entry{'foo'} --> memory_log
+```
+
+### Memory Log
+
+Advanced Logger generates a binary format memory log which includes a header,
+followed by entries containing the entry metadata and the log string. Entry metadata
+includes: timestamp, log level, and the boot phase (e.g., DXE). For the exact format,
+see the [memory log implementation](https://github.com/OpenDevicePartnership/patina/blob/main/components/patina_adv_logger/src/memory_log.rs).
+The format is subject to change, and the memory log is intended to be accessed and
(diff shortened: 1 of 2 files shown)


components/patina_adv_logger/src/memory_log.rs
```diff
@@ -131,6 +131,7 @@ impl AdvLoggerInfo {
         }
     }
 
+    #[allow(dead_code)]
```
os-d, line 135: Comments on why we carry the dead code (or if it isn't really dead)? Here are below

components/patina_adv_logger/src/logger.rs
```diff
@@ -37,6 +37,15 @@ impl<'a, S> AdvancedLogger<'a, S>
 where
     S: SerialIO + Send,
 {
+    /// Creates a new AdvancedLogger.
+    ///
+    /// # Arguments
```
Javagedes: nit:

```suggestion
    /// ## Arguments
```

docs/src/dxe_core/advanced_logger.md
```diff
@@ -0,0 +1,156 @@
+# Advanced Logger
+
+Patina uses the [Log crate](https://docs.rs/log/latest/log/) for its logging
+diagnostics, and the platform code is responsible for configuring its desired log implementation
```
Javagedes: orchestration may be a better word here? Not sure. Diagnostics just does not seem "correct"

docs/src/dxe_core/advanced_logger.md
```diff
@@ -0,0 +1,156 @@
...
+## Logger
+
+The logger is designed to be statically instantiated as required by the log crate
+and is responsible for handling the log messages forwarded through the logging interfaces
+such as `log::info!()`. Logger initialization acts as the primary interface for platform
+configurations such as serial interfaces, log level limits, message filtering,
+etc.
+
+The `Log` implementation will send all messages to both the configured serial port
+and its memory log to be picked up later by components in DXE or at runtime.
```
Javagedes, line 31: Is there an option to not set a serial logger and then it will only log to the memory log?

docs/src/dxe_core/advanced_logger.md
```diff
@@ -0,0 +1,156 @@
...
+    logger -- entry{'foo'} --> memory_log
+```
+
+### Memory Log
+
+Advanced Logger generates a binary format memory log which includes a header,
+followed by entries containing the entry metadata and the log string. Entry metadata
+includes: timestamp, log level, and the boot phase (e.g., DXE). For the exact format,
+see the [memory log implementation](https://github.com/OpenDevicePartnership/patina/blob/main/components/patina_adv_logger/src/memory_log.rs).
+The format is subject to change, and the memory log is intended to be accessed and
```
Javagedes, line 55: Any plans / ability to implement the other side of the advanced logger purely in rust so we can get rid of "Project Mu" all together?

docs/src/dxe_core/advanced_logger.md
```diff
@@ -0,0 +1,156 @@
...
+#### Memory Log Initialization
+
+The memory log will typically be allocated and initialized by an earlier phase of
+boot such as PEI or SEC. In this case, Patina will inherit the memory log from this earlier
+phase to provide a continuous log of the entire boot sequence. The location of the
+memory log provided by a previous phase will be provided through the Advanced Logger
+HOB, which simply consists of the physical address of the log. From there, the header,
+which is validated by signature, will describe the size and other state of the memory log.
+
+The memory log may also be initialized within Patina. This has some limitations,
```
Javagedes, line 68: Not needed in this document, but in general is this process documented? If not it should probably be added to the crate top level module documentation.

docs/src/dxe_core/advanced_logger.md
makubacki: I'd just say something like

> "Patina uses the [Log crate](https://docs.rs/log/latest/log/) to facilitate logging
and the platform code is responsible for configuring its desired log implementation..."

docs/src/dxe_core/advanced_logger.md
```diff
@@ -0,0 +1,156 @@
+# Advanced Logger
+
+Patina uses the [Log crate](https://docs.rs/log/latest/log/) for its logging
+diagnostics, and the platform code is responsible for configuring its desired log implementation
+prior to executing the Patina core using the standard `log::set_logger` routine.
+The platform may choose any implementation of the `Log` trait it wishes; however,
```
makubacki:
```suggestion
prior to executing the Patina core using the standard `log::set_logger` routine.

The platform may choose any implementation of the `Log` trait it wishes; however,
```

docs/src/dxe_core/advanced_logger.md
```diff
@@ -0,0 +1,156 @@
...
+```
+
+### Memory Log
+
+Advanced Logger generates a binary format memory log which includes a header,
+followed by entries containing the entry metadata and the log string. Entry metadata
+includes: timestamp, log level, and the boot phase (e.g., DXE). For the exact format,
+see the [memory log implementation](https://github.com/OpenDevicePartnership/patina/blob/main/components/patina_adv_logger/src/memory_log.rs).
+The format is subject to change, and the memory log is intended to be accessed and
+parsed by tools within Project MU and this crate. For details on accessing and
```
makubacki:
```suggestion
parsed by tools within Project Mu and this crate. For details on accessing and
```

makubacki: nit: Update throughout the doc.

docs/src/dxe_core/advanced_logger.md
cfernald: This is mentioned in the filtering section, but will call that ou there.

docs/src/dxe_core/advanced_logger.md
cfernald: Well depends on when we target PEI for Patina/rust, which is a larger question then AdvLogger IMO.

docs/src/dxe_core/advanced_logger.md
cfernald: Its not because isn't not actually well supported at the moment. The problem is that allocations are not actually available this early in the boot flow and so that story needs to be thought through more. A few possible examples. 
1. Allocate the component to allocate the buffer. This leads to lost logs
2. Allow the platform to have a static region. No dropped logs, but more platform magic required. 

When support is added we can document better.

PR #522, opened 2025-06-16
RFC: Firmware File System support crate in Patina.

joschock:
## Description

RFC for moving FFS support from `mu_rust_pi` into Patina and adding modification/generation support for FFS.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [x] Includes documentation?

## How This Was Tested

N/A

## Integration Instructions

N/A

### DIFF
--- a/docs/src/rfc/text/0014-firmware-filesystem-support.md
+++ b/docs/src/rfc/text/0014-firmware-filesystem-support.md
@@ -50,22 +50,6 @@ existing capabilities of the mu_rust_pi implementation are not.
 3. Crate produces an API for generating new FFS volumes and modifying existing volumes and serializing them back into
 PI-spec confirmant representations (new functionality)
 
-## Unresolved Questions
-
-- General approach to generation?
-
-Initial proposed approach is to have two separate sets of structures: one for fast, mostly zero-copy* read access
-against a `&[u8]` slice to be used in constrained environments (like the core dispatcher), and a second "owned" set of
-heap-allocated structures that model FFS data relationships and allow for manipulation of FFS, and have an API serialize
-the resulting FFS data into a `Vec<u8>` spec-compliant representation. The first set of structures is essentially what
-is implemented today in mu_rust_pi. The borrowed set of structures could be used to read and populate an "owned" set of
-structures for manipulation and re-serialization. More details of the initial proposed approach in the "Rust Code
-Design" section below.
-
-*"mostly" zero-copy because `Section` data is heap-allocated since section extraction will require allocations for
-extracted sections, and having all `Section` data be owned instead of having a mix of owned and borrowed data backing
-`Section` greatly simplifies the ownership model.
-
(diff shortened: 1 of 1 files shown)


docs/src/rfc/text/0000-firmware-filesystem-support.md
```diff
@@ -0,0 +1,132 @@
...
+existing capabilities of the mu_rust_pi implementation are not.
+2. Crate produces an API for reading and accessing FFS data (existing functionality)
+3. Crate produces an API for generating new FFS volumes and modifying existing volumes and serializing them back into
+PI-spec confirmant representations (new functionality)
+
+## Unresolved Questions
+
+- Crate location?
+
+Initial proposal is `core/patina_ffs` but could perhaps be an entirely separate crate. There
```
os-d: If the intended use case for new consumers is outside of Patina core(s), I would advocate we don't put it in `core/` as that is intended for things only consumed by a Patina core (e.g. DXE Core, StMM Core, etc.).

docs/src/rfc/text/0000-firmware-filesystem-support.md
```diff
@@ -0,0 +1,132 @@
...
+
+
+## Alternatives
+
+Other alternatives considered:
+1. Do nothing. Retain existing design point and develop FFS generation capability independently. This would reduce the
+cohesion of the design, and limit the availability of the generation capability. In addition, one of the learnings of
+the evolution of Patina was that it makes sense to properly organize and co-locate related code (see: RFC
+0006-patina-repo.md).
+2. Develop FFS generation capability at current location in mu_rust_pi. See motivation section above for the conerns
```
os-d:
```suggestion
2. Develop FFS generation capability at current location in mu_rust_pi. See motivation section above for the concerns

```

docs/src/rfc/text/0000-firmware-filesystem-support.md
joschock: Is `sdk` more appropriate? I don't think it's a `component` in the Patina-sense of the word...

joschock: (maybe it should be a `component`?)

os-d: Unless it is using the Component framework Joey built, I don't think it should go in `component`. `sdk` probably makes more sense since it is targeting multiple environments.

makubacki: `sdk` is the right place imo.

joschock: Settled on `sdk/patina_ffs`. Updated the Rust Code Design section to indicate this, and removed the open question.

joschock:
Updating status to FCP.

PR #527, opened 2025-06-17
patina_sdk: Make patina tests(`#[patina_test]`) an opt-in feature

vineelko:
## Description

 - The Patina SDK test infrastructure relies on the `linkme` crate.
   - linkme: **Safe cross-platform linker shenanigans**
 - One requirement of the crate is that the `distributed_slice` we use to collocate all Patina tests must contain at least one element. This means any external utility (e.g., a platform validation tool) that depends on `patina_sdk` will encounter a linker crash (not just an error) if it does not provide at least one `#[patina_test]` test.
 - The current feature flag `off` is misleading, as opting out of this functionality requires opting in to a feature called `off`
 - Rename the current feature flag (`off`) to `enable_patina_tests`.
 - Invert the logic for processing the feature flag to make it opt-in as enabling this functionality by default can trip the consumers.

Linker Crash when building platform validation tool:
image

(skip cargo release dry run)
---

- [ ] Impacts functionality?
- [ ] Impacts security?
- [x] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Verified by integrating it with platform validation tool and `cargo make all` works

## Integration Instructions

Any crate trying to use this functionality should opt-in to the feature "enable_patina_tests" feature as shown below
`patina_sdk = { ...., features = ["enable_patina_tests",...] }`

### DIFF
--- a/sdk/patina_sdk_macro/src/test_macro.rs
+++ b/sdk/patina_sdk_macro/src/test_macro.rs
@@ -25,7 +25,7 @@ pub fn patina_test2(stream: proc_macro2::TokenStream) -> proc_macro2::TokenStrea
 
     // Wait until we filter out or custom attributes so that we don't confuse the compiler
     // with attributes it does not expect.
-    if cfg!(not(feature = "patina_tests")) {
+    if cfg!(not(feature = "enable_patina_tests")) {
         return handle_feature_off(item);
     }
 
@@ -146,7 +146,7 @@ mod tests {
         };
 
         let expanded = patina_test2(stream);
-        let expected = if cfg!(feature = "patina_tests") {
+        let expected = if cfg!(feature = "enable_patina_tests") {
             quote! {
                 #[patina_sdk::test::linkme::distributed_slice(patina_sdk::test::__private_api::TEST_CASES)]
                 #[linkme(crate = patina_sdk::test::linkme)]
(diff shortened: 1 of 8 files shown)


sdk/patina_sdk_macro/src/test_macro.rs
```diff
@@ -146,20 +146,28 @@ mod tests {
...
-            static __my_test_case_TestCase: patina_sdk::test::__private_api::TestCase = patina_sdk::test::__private_api::TestCase {
-                name: concat!(module_path!(), "::", stringify!(my_test_case)),
-                skip: false,
-                should_fail: false,
-                fail_msg: None,
-                func: |storage| patina_sdk::test::__private_api::FunctionTest::new(my_test_case).run(storage.into()),
-            };
-            fn my_test_case() -> Result {
-                assert!(true);
+        let expected = if cfg!(feature = "patina_tests") {
```
Javagedes: I don't agree with how this is being handled in the tests. In my opinion (open to other opinions) We need to strictly enable the feature flag for tests and always expect that the feature flag be enabled for tests. 

If we don't, then the test is a do nothing test. If we ever accidentally stop enabling the feature flag in CI, then we are quite literally not testing at all.

vineelko: Yaa I think that is a fair point. Let me add that feature flag to tests

vineelko: Fixed in the latest iteration.

makubacki:
> I agree with this PR as a whole,
> 
> However I don't think the name you selected for the feature adds much additional clarity. Can we consider a different name? Maybe link include_patina_tests, compile_patina_tests, enable_patina_tests, or something else? I'm fine if not, if others think it is clear enough as is.

I think `enable_patina_tests` is clearer without being overly descriptive.

PR #539, opened 2025-06-19
[REBASE & FF] AdvLogger: Fix UB from casting header values to atomic.

### LINKED ISSUE #530: [Bug]: Interior mutablity pattern used in `AdvLoggerInfo` has UB

### Current Behavior

The `AdvLoggerInfo` type looks like this:

https://github.com/OpenDevicePartnership/patina/blob/72bc8b4d2a906d206181111986bd17179908be7a/components/patina_adv_logger/src/memory_log.rs#L46-L83

where some of the fields are used with atomics to give interior mutability:

https://github.com/OpenDevicePartnership/patina/blob/72bc8b4d2a906d206181111986bd17179908be7a/components/patina_adv_logger/src/memory_log.rs#L152-L162

https://github.com/OpenDevicePartnership/patina/blob/72bc8b4d2a906d206181111986bd17179908be7a/components/patina_adv_logger/src/memory_log.rs#L175-L180

https://github.com/OpenDevicePartnership/patina/blob/72bc8b4d2a906d206181111986bd17179908be7a/components/patina_adv_logger/src/memory_log.rs#L202-L206

The problem is that the types of these fields must have [interior mutability themselves](https://doc.rust-lang.org/reference/interior-mutability.html) in order for this kind of update to be well defined. While the atomic accesses prevent data races, they still mutate through a shared reference without interior mutability [which is Undefined Behavior](https://doc.rust-lang.org/reference/behavior-considered-undefined.html#:~:text=Moreover%2C%20the%20bytes%20pointed%20to%20by%20a%20shared%20reference%2C%20including%20transitively%20through%20other%20references%20(both%20shared%20and%20mutable)%20and%20Boxes%2C%20are%20immutable%3B%20transitivity%20includes%20those%20references%20stored%20in%20fields%20of%20compound%20types.) in Rust.

This [Rust Playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=512c37c4ce7d20f090da96c0cfd4509a) demonstrates the issue succinctly. Try running the code in Debug, then Release (notice the optimizer has ignored the write to `f.x`) and then finally with Miri (Tools -> Miri) which complains:

```
error: Undefined Behavior: trying to retag from <360> for SharedReadWrite permission at alloc226[0x0], but that tag only grants SharedReadOnly permission for this location
    --> /playground/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/sync/atomic.rs:3635:1
     |
3635 | / atomic_int! {
3636 | |     cfg(target_has_atomic = "32"),
3637 | |     cfg(target_has_atomic_equal_alignment = "32"),
3638 | |     stable(feature = "integer_atomics_stable", since = "1.34.0"),
...    |
3651 | |     u32 AtomicU32
3652 | | }
     | | ^
     | | |
     | |_trying to retag from <360> for SharedReadWrite permission at alloc226[0x0], but that tag only grants SharedReadOnly permission for this location
     |   this error occurs as part of retag at alloc226[0x0..0x4]
     |
     = help: this indicates a potential bug in the program: it performed an invalid operation, but the Stacked Borrows rules it violated are still experimental
     = help: see https://github.com/rust-lang/unsafe-code-guidelines/blob/master/wip/stacked-borrows.md for further information
help: <360> was created by a SharedReadOnly retag at offsets [0x0..0x4]
    --> src/main.rs:9:46
     |
9    |         let x = unsafe { AtomicU32::from_ptr(&self.x as *const u32 as *mut u32) };
     |                                              ^^^^^^^
     = note: BACKTRACE (of the first span):
     = note: inside `std::sync::atomic::AtomicU32::from_ptr::<'_>` at /playground/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/sync/atomic.rs:2634:26: 2634:38
note: inside `Foo::incr_x_atomically`
    --> src/main.rs:9:26
     |
9    |         let x = unsafe { AtomicU32::from_ptr(&self.x as *const u32 as *mut u32) };
     |                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
note: inside `main`
    --> src/main.rs:16:5
     |
16   |     f.incr_x_atomically();
     |     ^^^^^^^^^^^^^^^^^^^^^
     = note: this error originates in the macro `atomic_int` (in Nightly builds, run with -Z macro-backtrace for more info)

note: some details are omitted, run with `MIRIFLAGS=-Zmiri-backtrace=full` for a verbose backtrace

error: aborting due to 1 previous error
```

### Expected Behavior

I believe the correct solution is to change the type of these fields to be `AtomicU32`/`AtomicU64` which has the same size and layout as the regular integer types but has interior mutability so that these updates are well defined. ([Fixed playground](https://play.rust-lang.org/?version=stable&mode=release&edition=2024&gist=7a973bee8369d984b133bbce89b39e48))

### Steps To Reproduce

n/a

### Build Environment

```markdown
n/a
```

### Version Information

```text
Commit: dc18c7dddfd4ad6eb4b79bacbf96d80d75155bbd
```

### Urgency

Medium

Someone else needs to fix it

Maintainer feedback requested

### Anything else?

_No response_

cfernald:
## Description

The initial implementation tried to stay as accurate to a C definition of the structures as possible and so used u32 in cases where valuers are only atomically updated. This required that those types be converted to atomic (internally mutable) values later which can cause undefined behavior. This change is to make the structure itself contain the atomic versions of the types.

Resolves #530 

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on Q35 with C parser

## Integration Instructions

N/A

### DIFF
--- a/components/patina_adv_logger/src/memory_log.rs
+++ b/components/patina_adv_logger/src/memory_log.rs
@@ -55,9 +55,9 @@ pub struct AdvLoggerInfo {
     /// Reserved for future
     reserved2: u32,
     /// Offset from LoggerInfo to where to store next log entry.
-    log_current_offset: u32,
+    log_current_offset: AtomicU32,
     /// Number of bytes of messages missed
-    discarded_size: u32,
+    discarded_size: AtomicU32,
     /// Size of allocated buffer
     log_buffer_size: u32,
     /// Log in permanent RAM
@@ -73,7 +73,7 @@ pub struct AdvLoggerInfo {
     /// Reserved for future
     reserved3: [bool; 3],
     /// Ticks per second for log timing
-    timer_frequency: u64,
+    timer_frequency: AtomicU64,
     /// Ticks when Time Acquired
     ticks_at_time: u64,
     /// UEFI Time Field
@@ -103,8 +103,8 @@ impl AdvLoggerInfo {
             reserved1: [0, 0, 0],
             log_buffer_offset: size_of::<AdvLoggerInfo>() as u32,
             reserved2: 0,
-            log_current_offset: size_of::<AdvLoggerInfo>() as u32,
-            discarded_size: 0,
+            log_current_offset: AtomicU32::new(size_of::<AdvLoggerInfo>() as u32),
+            discarded_size: AtomicU32::new(0),
             log_buffer_size,
             in_permanent_ram: true,
             at_runtime: false,
@@ -112,7 +112,7 @@ impl AdvLoggerInfo {
             hw_port_initialized: false,
             hw_port_disabled,
             reserved3: [false, false, false],
-            timer_frequency,
+            timer_frequency: AtomicU64::new(timer_frequency),
             ticks_at_time,
             time,
             hw_print_level,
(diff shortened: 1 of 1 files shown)


makubacki:
It's not an issue, but why is this marked "Rebase & FF"? There is only a single commit right now.

PR #549, opened 2025-06-20
patina-sdk: Make align* Functions More User Friendly

os-d:
## Description

Currently the align family of functions in the sdk requires u64s to be passed in, which often requires a pattern of my_usize as u64 getting passed in and the result of this being cast back to usize. This is needless, these functions can easily be generic.

In addition to this, these functions were passing back string errors which were universally being converted to EfiErrors. It was not using to return strings, instead return EfiErrors.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Ran local tests as well as booted Windows on Q35 and SBSA.

## Integration Instructions

N/A. Existing users of the align functions don't need to change, but can change if they wish to make the usage simpler.

### DIFF
--- a/sdk/patina_sdk/src/base.rs
+++ b/sdk/patina_sdk/src/base.rs
@@ -8,7 +8,10 @@
 //!
 //! SPDX-License-Identifier: BSD-2-Clause-Patent
 
+use num_traits;
 use r_efi::efi;
+
+use crate::error::EfiError;
 
 /// EFI memory allocation functions work in units of EFI_PAGEs that are 4KB.
 /// This should in no way be confused with the page size of the processor.
@@ -141,6 +144,21 @@ pub const SIZE_256TB: usize = 0x1000000000000;
 /// Patina uses write back as the default cache attribute for memory allocations.
 pub const DEFAULT_CACHE_ATTR: u64 = efi::MEMORY_WB;
 
+/// Checks if the given value is a power of two.
+/// This function checks if the value `x` is greater than zero and if it is a power of two.
+/// # Parameters
+/// - `x`: The value to check.
+/// # Returns
+/// - `true`: If `x` is a power of two.
+/// - `false`: If `x` is not a power of two.
+#[inline]
+pub fn is_power_of_two<T>(x: T) -> bool
+where
+    T: num_traits::PrimInt,
+{
+    x > T::zero() && (x & (x - T::one())) == T::zero()
+}
+
 /// Aligns the given address down to the nearest boundary specified by align.
 ///
 /// # Parameters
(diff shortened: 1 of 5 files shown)


sdk/patina_sdk/src/base.rs
```diff
@@ -221,18 +256,25 @@ pub const fn align_down(addr: u64, align: u64) -> Result<u64, &'static str> {
...
+pub fn align_up<T>(addr: T, align: T) -> Result<T, EfiError>
+where
+    T: Copy
+        + PartialEq
+        + PartialOrd
+        + core::ops::BitAnd<Output = T>
+        + core::ops::Sub<Output = T>
+        + core::ops::BitOr<Output = T>
+        + core::ops::Add<Output = T>
+        + num_traits::PrimInt,
```
vineelko: Same here

sdk/patina_sdk/src/base.rs
```diff
@@ -177,11 +201,20 @@ pub const DEFAULT_CACHE_ATTR: u64 = efi::MEMORY_WB;
...
-        return Err("`align` must be a power of two");
+pub fn align_down<T>(addr: T, align: T) -> Result<T, EfiError>
+where
+    T: Copy
+        + PartialEq
+        + PartialOrd
+        + core::ops::BitAnd<Output = T>
+        + core::ops::Not<Output = T>
+        + core::ops::Sub<Output = T>
+        + num_traits::PrimInt,
```
vineelko: Since `PrimInt` is already implementing all other bounds, we can remove them here and `T: num_traits::PrimInt` should be sufficient.

```rust
        + Copy
        + PartialEq
        + PartialOrd
        + core::ops::BitAnd<Output = T>
        + core::ops::Sub<Output = T>
        + core::ops::BitOr<Output = T>
        + core::ops::Add<Output = T>
```

sdk/patina_sdk/src/base.rs
```diff
@@ -141,6 +144,27 @@ pub const SIZE_256TB: usize = 0x1000000000000;
...
+#[inline]
+pub fn is_power_of_two<T>(x: T) -> bool
+where
+    T: num_traits::Zero
+        + num_traits::One
+        + Copy
+        + core::ops::BitAnd<Output = T>
+        + core::cmp::PartialOrd
+        + core::ops::Sub<Output = T>
+        + PartialEq,
```
vineelko: Same here.

sdk/patina_sdk/src/base.rs
os-d: Good call, thanks, fixed all instances

sdk/patina_sdk/src/base.rs
vineelko: This was missed.

os-d: Thanks, I also noticed that and fixed it

PR #552, opened 2025-06-23
patina-dxe-core: Add ARM64 64 KB Runtime Allocation Granularity

### LINKED ISSUE #524: [Feature]: Support ARM64 64KB Runtime Allocation Granularity
### Feature Overview

The UEFI spec requires that ARM64 runtime memory allocations are aligned to 64KB to accommodate OSes that use 16KB or 64KB page sizes.

### Solution Overview

Ensure that runtime memory types are allocated at a 64KB granularity for ARM64.

### Alternatives Considered

_No response_

### Urgency

Medium

I will implement the feature

No maintainer feedback needed

### Anything else?

_No response_

os-d:
## Description

This patch adds support and unit tests for ARM64 64 KB runtime page allocation granularity, which is required by the UEFI spec to ensure that 16 KB and 64 KB OSes can boot.

Note, this granularity is not tested at all in edk2 and has been a source of many bugs. This is testable in patina with the with_granularity_modulation fn.

This patch also adds debug_asserts to places where the FSB fails; otherwise we lose what the allocator was doing because Rust core module code will receive the `AllocError` and then call the `alloc_error_handler`, which panics, but we've lost the stack at that point. We see the top level allocation that failed, but not what failed within the allocator, so debugging is useless at that point. With the debug_asserts, we can see what failed in the allocator and debug from there. There is a concern that in production code we will panic on `AllocError`s, but this seems to be a fundamental design of the core Rust code, so I do not think Patina can do anything else.

This patch dropped some unit tests that weren't testing anything meaningful and removed a couple of test asserts that cannot be tested now due to debug_asserting in alloc errors.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested by booting Windows and Linux (4KB and 64KB for ARM64) on Q35 and SBSA and viewing the EFI_MEMORY_MAP and MAT table to confirm all runtime types have 4 KB alignment and length on x86 and 64 KB on ARM64.

## Integration Instructions

This closes #524 and so allows platforms to boot 16 KB and 64 KB Linuxes.

### DIFF
--- a/docs/src/integrate/patina_dxe_core_requirements.md
+++ b/docs/src/integrate/patina_dxe_core_requirements.md
@@ -72,22 +72,27 @@ in FFS listed order.
 > Platforms may also list drivers in FFSes in the order they should be dispatched, though it is recommended to rely on
 > depex statements.
 
-#### 1.3 Driver Section Alignment Must Be a Positive Multiple of 4 KB
+#### 1.3 Driver Section Alignment Must Be a Multiple of 4 KB
 
-Patina relies on using a 4 KB page size and as a result requires that the C based drivers it dispatches have a positive
+Patina relies on using a 4 KB page size and as a result requires that the C based drivers it dispatches have a
 multiple of 4KB as a page size in order to apply image memory protections. The EDK II DXE Core cannot apply image
 memory protections on images without this section alignment requirement, but it will dispatch them, depending on
 configuration.
 
+ARM64 DXE_RUNTIME_DRIVERs must have a multiple of 64 KB image section alignment per UEFI spec requirements.
+This is required to boot operating systems with 16 KB or 64 KB page sizes.
+
 Patina components will have 4 KB section alignment by nature of being compiled into Patina.
 
-The DXE Readiness Tool validates all drivers have a positive multiple of 4 KB section alignment and reports an error if
-not.
+The DXE Readiness Tool validates all drivers have a multiple of 4 KB section alignment and reports an error if
+not. It will also validate that ARM64 DXE_RUNTIME_DRIVERs have a multiple of 64KB section alignment.
 
 > **Guidance:**
-> All C based drivers must be compiled with a linker flag that enforces a positive multiple of 4 KB section alignment.
-> Commonly, 4 KB is used except for ARM64 runtime drivers, which use 64 KB per UEFI spec requirements. For MSVC, this
-> linker flag is `/ALIGN:4096` for GCC/CLANG, the flag is `-z common-page-size=0x1000`.
+> All C based drivers must be compiled with a linker flag that enforces a multiple of 4 KB section alignment.
+> For MSVC, this linker flag is `/ALIGN:0x1000` for GCC/CLANG, the flag is `-z common-page-size=0x1000`. This section
+> allows for multiples of 4 KB or 64 KB, depending on driver, but unless a specific use case dictates greater section
+> alignment, then it is recommended to use 4 KB for everything except for ARM64 DXE_RUNTIME_DRIVERs, which should use
+> 64 KB, e.g. `/ALIGN:0x10000` for MSVC and `-z common-page-size=0x10000` for GCC/CLANG.
 
 ### 2. Hand Off Block (HOB) Requirements
 
@@ -195,15 +200,3 @@ Tracking issue: [#517](https://github.com/OpenDevicePartnership/patina/issues/51
 
 > **Guidance:**
 > Temporarily, LZMA compressed sections that will be decompressed in DXE should use Brotli or TianoCompress.
-
-#### 4.2 ARM64 64KB Runtime Memory Alignment Not Guaranteed
-
-The Patina DXE Core does not currently support allocating ARM64 runtime memory with 64KB granularity as required per the
-UEFI spec to allow booting operating systems that use 16KB or 64KB page sizes.
-
-Windows only uses a 4KB page size and Linux by default uses 4KB.
-
-Tracking issue: [#524](https://github.com/OpenDevicePartnership/patina/issues/524)
-
-> **Guidance:**
-> Ensure any operating system being booted with Patina uses a 4KB page size.
--- a/patina_dxe_core/src/allocator/fixed_size_block_allocator.rs
+++ b/patina_dxe_core/src/allocator/fixed_size_block_allocator.rs
@@ -28,7 +28,7 @@ use core::{
 use linked_list_allocator::{align_down_size, align_up_size};
 use mu_pi::dxe_services::GcdMemoryType;
 use patina_sdk::{
-    base::{UEFI_PAGE_SHIFT, UEFI_PAGE_SIZE},
+    base::{align_up, UEFI_PAGE_SHIFT, UEFI_PAGE_SIZE},
     error::EfiError,
     uefi_pages_to_size, uefi_size_to_pages,
 };
@@ -54,6 +54,9 @@ pub const MIN_EXPANSION: usize = 0x100000;
 const ALIGNMENT: usize = 0x1000;
 
 const BLOCK_SIZES: &[usize] = &[8, 16, 32, 64, 128, 256, 512, 1024, 2048, 4096];
+
+// Compile-time check to ensure the MIN_EXPANSION is a multiple of RUNTIME_PAGE_ALLOCATION_GRANULARITY.
+const _: () = assert!(MIN_EXPANSION % super::RUNTIME_PAGE_ALLOCATION_GRANULARITY == 0);
 
 // Returns the index in the block list for the minimum size block that will
 // satisfy allocation for the given layout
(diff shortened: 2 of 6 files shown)


patina_dxe_core/src/allocator/fixed_size_block_allocator.rs
```diff
@@ -55,6 +56,41 @@ const ALIGNMENT: usize = 0x1000;
 
 const BLOCK_SIZES: &[usize] = &[8, 16, 32, 64, 128, 256, 512, 1024, 2048, 4096];
 
+// Per the UEFI spec, AARCH64 runtime pages need to be allocated on 64KB boundaries in units of 64KB to accommodate
+// OSes that use 16KB or 64KB page sizes. Other architectures use 4KB pages, so we don't have any additional
+// granularity requirements for them.
+cfg_if::cfg_if! {
+    if #[cfg(test)] {
+        // by default use 4 KB, but allow tests to configure this
+        static mut RUNTIME_PAGE_ALLOCATION_GRANULARITY: usize = 0x1000;
```
cfernald: nit: use `patina_sdk::base::SIZE_4KB`

patina_dxe_core/src/allocator/fixed_size_block_allocator.rs
```diff
@@ -55,6 +56,41 @@ const ALIGNMENT: usize = 0x1000;
...
+    else if #[cfg(target_arch = "aarch64")] {
+        const RUNTIME_PAGE_ALLOCATION_GRANULARITY: usize = patina_sdk::base::SIZE_64KB;
+    } else {
+        const RUNTIME_PAGE_ALLOCATION_GRANULARITY: usize = patina_sdk::base::SIZE_4KB;
+
+    }
+}
+
+#[cfg(not(test))]
+macro_rules! get_runtime_page_allocation_granularity {
```
cfernald: Just curious, why a macro instead of a function? Not a problem, just seems a macro is unnecessary

patina_dxe_core/src/allocator/fixed_size_block_allocator.rs
```diff
@@ -755,6 +811,44 @@ impl Display for SpinLockedFixedSizeBlockAllocator {
...
+        _ => UEFI_PAGE_SIZE,
+    };
+
+    // Compute the required alignment as the least common multiple of alignment and granularity
+    let required_alignment = if granularity % alignment == 0 {
+        granularity
+    } else if alignment % granularity == 0 {
+        alignment
+    } else {
+        lcm(alignment, granularity)
```
cfernald: These are both powers of 2, right? So, shouldn't this just be whichever one is greater as it is guaranteed to be a multiple of the other?

docs/src/integrate/patina_dxe_core_requirements.md
```diff
@@ -79,15 +79,20 @@ multiple of 4KB as a page size in order to apply image memory protections. The E
 memory protections on images without this section alignment requirement, but it will dispatch them, depending on
 configuration.
 
+ARM64 DXE_RUNTIME_DRIVERs must have a positive multiple of 64 KB image section alignment per UEFI spec requirements.
+This is required to boot operating systems with 16 KB or 64 KB page sizes.
```
joschock: nit: concision - I'm not aware of any systems anywhere that support "negative" alignments.  
```suggestion
ARM64 DXE_RUNTIME_DRIVERs must have a multiple of 64 KB image section alignment per UEFI spec requirements.
This is required to boot operating systems with 16 KB or 64 KB page sizes.
```

docs/src/integrate/patina_dxe_core_requirements.md
```diff
@@ -79,15 +79,20 @@ multiple of 4KB as a page size in order to apply image memory protections. The E
...
 The DXE Readiness Tool validates all drivers have a positive multiple of 4 KB section alignment and reports an error if
-not.
+not. It will also validate that ARM64 DXE_RUNTIME_DRIVERs have a positive multiple of 64KB section alignment.
 
 > **Guidance:**
 > All C based drivers must be compiled with a linker flag that enforces a positive multiple of 4 KB section alignment.
-> Commonly, 4 KB is used except for ARM64 runtime drivers, which use 64 KB per UEFI spec requirements. For MSVC, this
-> linker flag is `/ALIGN:4096` for GCC/CLANG, the flag is `-z common-page-size=0x1000`.
+> Commonly, 4 KB is used except for ARM64 runtime drivers, which use 64 KB. For MSVC, this
+> linker flag is `/ALIGN:4096` for GCC/CLANG, the flag is `-z common-page-size=0x1000`. This section allows for positive
```
makubacki: To prevent confusion, it might be best to just give the flag name (`/ALIGN`) or add an example to the end of the paragraph showing 64 KB values after "which should use 64 KB".

patina_dxe_core/src/allocator/fixed_size_block_allocator.rs
os-d: I can change this to a function. The only reason it exists is so we don't have to have an unsafe (and then ignore unused unsafe) for the actual builds.

docs/src/integrate/patina_dxe_core_requirements.md
os-d: Yeah, that's fair. I don't know why I was being overly pedantic :)

patina_dxe_core/src/allocator/fixed_size_block_allocator.rs
```diff
@@ -55,6 +56,41 @@ const ALIGNMENT: usize = 0x1000;
 
 const BLOCK_SIZES: &[usize] = &[8, 16, 32, 64, 128, 256, 512, 1024, 2048, 4096];
 
+// Per the UEFI spec, AARCH64 runtime pages need to be allocated on 64KB boundaries in units of 64KB to accommodate
+// OSes that use 16KB or 64KB page sizes. Other architectures use 4KB pages, so we don't have any additional
+// granularity requirements for them.
+cfg_if::cfg_if! {
```
joschock: I haven't thoroughly thought thru the implications, but could you potentially use const generics instead of static mut for this? e.g.
```rust
pub struct FixedSizeBlockAllocator<const ALIGNMENT:usize>
pub struct SpinLockedFixedSizeBlockAllocator<const ALIGNMENT:usize = patina_sdk::base::SIZE_4KB> {
    gcd: &'static SpinLockedGcd,
    handle: efi::Handle,
    inner: tpl_lock::TplMutex<FixedSizeBlockAllocator<ALIGNMENT>>,
}
...

//normal 4K alignment
pub(crate) static EFI_BOOT_SERVICES_DATA_ALLOCATOR: UefiAllocator = UefiAllocator::new(
    &GCD,
    efi::BOOT_SERVICES_DATA,
    protocol_db::EFI_BOOT_SERVICES_DATA_ALLOCATOR_HANDLE,
    page_change_callback,
);

// RUNTIME_PAGE_ALLOCATION_GRANULARTIY is cfg dependent const
pub static EFI_RUNTIME_SERVICES_CODE_ALLOCATOR: UefiAllocator<RUNTIME_PAGE_ALLOCATION_GRANULARITY> = UefiAllocator::new(
    &GCD,
    efi::RUNTIME_SERVICES_CODE,
    protocol_db::EFI_RUNTIME_SERVICES_CODE_ALLOCATOR_HANDLE,
    page_change_callback,
);
```
https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=6495804cf1fa822483a71ee2714ca5ca

joschock: you can also access the const directly:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=8045e5b78842b97d6f1649cade137b9c

os-d: That's an interesting idea, thanks, let me play with that

patina_dxe_core/src/allocator/fixed_size_block_allocator.rs
```diff
@@ -55,6 +55,11 @@ const ALIGNMENT: usize = 0x1000;
 
 const BLOCK_SIZES: &[usize] = &[8, 16, 32, 64, 128, 256, 512, 1024, 2048, 4096];
 
+// Compile-time check to ensure the MIN_EXPANSION is a multiple of RUNTIME_PAGE_ALLOCATION_GRANULARITY. This cannot be
+// done in a test because the RUNTIME_PAGE_ALLOCATION_GRANULARITY is not const, so we can't use a const assert
```
joschock: Is this comment still correct? I think `RUNTIME_PAGE_ALLOCATION_GRANULARITY` is `const` now, right?

patina_dxe_core/src/allocator/fixed_size_block_allocator.rs
```diff
@@ -167,17 +172,41 @@ pub type PageChangeCallback = fn(&mut FixedSizeBlockAllocator);
...
+    /// The memory type that this allocator is managing. This is used to bucketize memory for the EFI_MEMORY_MAP and
+    /// handle any special cases for memory types.
     memory_type: efi::MemoryType,
+
+    /// The heads of the linked lists for each fixed-size block. Each index corresponds to a block size in
+    /// `BLOCK_SIZES`.
     list_heads: [Option<&'static mut BlockListNode>; BLOCK_SIZES.len()],
+
+    /// The linked-list of allocators that this allocator uses to back allocations that are larger than the fixed-size
+    /// blocks.
```
joschock: pedantry:
```suggestion
    /// The linked-list of allocators that this allocator uses to back allocations that are larger than the fixed-size
    /// blocks or if the required fixed-size block list is empty.
```

patina_dxe_core/src/allocator/fixed_size_block_allocator.rs
os-d: After much offline conversation, we decided to move to an immutable struct field.

patina_dxe_core/src/allocator/fixed_size_block_allocator.rs
os-d: Correct and thus it works for test code now. Fixed comment and removed cfg

patina_dxe_core/src/allocator/fixed_size_block_allocator.rs
os-d: In the refactor this is removed entirely.

patina_dxe_core/src/allocator/fixed_size_block_allocator.rs
```diff
@@ -538,14 +573,21 @@ impl SpinLockedFixedSizeBlockAllocator {
     ) -> Result<NonNull<[u8]>, EfiError> {
         // Record this call in the FSB's stats
         self.lock().stats.page_allocation_calls += 1;
+        let granularity = self.lock().page_allocation_granularity;
+
+        // Granularity and alignment both are powers of two, so we can use the max of the two
+        let required_alignment = if granularity >= alignment { granularity } else { alignment };
```
joschock: nit: 
```suggestion
        let required_alignment = cmp::max(granularity, alignment);
```

Issue #269, opened 2025-02-20
[Feature]: Add unit tests for dxe_services


HappyCoding-2025:
### Feature Overview

Current coverage: 25.58%

Add unit tests until coverage exceeds 80%

### Solution Overview

Functions to be covered are listed as sub-tasks

### Urgency

Medium

### Are you going to implement the feature request?

I will implement the feature

### Do you need maintainer feedback?

No maintainer feedback needed

makubacki:
@kat-perez, can you please update the status?

kat-perez:
@makubacki still in progress. planning on sending a PR by next week

makubacki:
@kat-perez, please update the issue status when you can.

makubacki:
I'm closing this parent issue as all subtasks are complete.

kat-perez:
@makubacki I need to re-open this since I didn't finish filling out the rest of the subtasks and the current code coverage is only at 25%, less than the target code coverage goal of 80%. Adding the rest of the subtasks to track the rest of the work

kat-perez:
#639 closes this

PR #557, opened 2025-06-26
gcd: use checked_add to prevent overflow panic in bounds validation

kat-perez:
## Description

Replace unchecked addition with checked_add() in add_memory_space to prevent debug build panics when validating address ranges.

The original `base_address + len <= self.maximum_address` would panic in debug builds due to Rust's overflow checking, even for safe values like 0x8000000000000002 + 0 (which is only halfway to usize::MAX on 64-bit).

Using checked_add().is_some_and() provides the same logical validation while gracefully handling potential overflow by returning EfiError::Unsupported instead of panicking.

- increases unit test coverage for DXE services (#269) from 3.3 to 6.2%

- [ ] Impacts functionality?
- [x] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

`cargo make coverage` 
- New unit tests were created in dxe_services.rs testing add_memory_space on the global static GCD

## Integration Instructions

N/A

### DIFF
--- a/patina_dxe_core/src/dxe_services.rs
+++ b/patina_dxe_core/src/dxe_services.rs
@@ -444,91 +444,57 @@ mod add_memory_space_tests {
 
     fn with_locked_state<F: Fn() + std::panic::RefUnwindSafe>(f: F) {
         test_support::with_global_lock(|| {
-            GCD.init(48, 16);
+            unsafe {
+                crate::test_support::init_test_gcd(None);
+            }
             f();
         })
         .unwrap();
     }
 
     #[test]
-    fn test_add_memory_space_basic_functionality() {
+    fn test_add_memory_space_success() {
         with_locked_state(|| {
-            println!("=== Testing add_memory_space basic functionality ===");
+            let result = add_memory_space(GcdMemoryType::SystemMemory, 0x80000000, 0x1000, efi::MEMORY_WB);
 
-            // Since GCD is a global singleton that might already be initialized,
-            // we can't control its state. Let's just test that our wrapper works.
-
-            let result = add_memory_space(
-                GcdMemoryType::SystemMemory,
-                0x80000000, // Use a high address to avoid conflicts
-                0x1000,
-                efi::MEMORY_WB,
-            );
-
-            println!("add_memory_space returned: {:#x} ({})", result.as_usize(), result.as_usize());
-
-            // Decode the status for debugging
-            let status_value = result.as_usize();
-            let status_name = match status_value {
-                0 => "SUCCESS",
-                0x8000000000000002 => "INVALID_PARAMETER",
-                0x8000000000000003 => "UNSUPPORTED",
-                0x8000000000000006 => "NOT_READY",
-                0x8000000000000009 => "OUT_OF_RESOURCES",
-                0x800000000000000F => "ACCESS_DENIED",
-                _ => "UNKNOWN",
-            };
-            println!("Status: {}", status_name);
-
-            // The test succeeds if we get a valid EFI status code
-            // (We can't guarantee success because we don't control GCD state)
-            let is_valid_status = matches!(result.as_usize(), 0 | 0x8000000000000000..=0x80000000000000FF);
-
-            assert!(is_valid_status, "Should return a valid EFI status code");
-
-            // The function works if it returns any valid status
-            println!("✓ add_memory_space function is working and returns valid status codes");
+            assert_eq!(result, efi::Status::SUCCESS);
         });
     }
 
     #[test]
     fn test_add_memory_space_parameter_validation() {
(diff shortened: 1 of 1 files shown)


patina_dxe_core/src/dxe_services.rs
```diff
@@ -435,3 +435,201 @@ pub fn init_dxe_services(system_table: &mut EfiSystemTable) {
...
+            GCD.init(48, 16);
+            f();
+        })
+        .unwrap();
+    }
+
+    #[test]
+    fn test_add_memory_space_basic_functionality() {
+        with_locked_state(|| {
+            println!("=== Testing add_memory_space basic functionality ===");
```
os-d: I don’t think we should have printlns in tests and let the test framework do its printing of what tests are run and then tests just use asserts and the test framework to indicate status. Printing both slows down test execution and is not scalable to thousands of tests executing to discover status. So I would remove all printlns in this PR. Otherwise this seems reasonable to me.

kat-perez: agreed, will update

patina_dxe_core/src/dxe_services.rs
```diff
@@ -435,3 +435,201 @@ pub fn init_dxe_services(system_table: &mut EfiSystemTable) {
...
+
+            // Since GCD is a global singleton that might already be initialized,
```
joschock: You can use the SpinLockedGcd::reset() function to reset GCD state to initial conditions, FWIW:
https://github.com/OpenDevicePartnership/patina/blob/10dacfc57f12649e9f5151a799a226d16d27461f/patina_dxe_core/src/gcd/spin_locked_gcd.rs#L1911

joschock: also the test wrapper closure is generating a new init state for GCD as well. so I'm not sure I get this comment?

patina_dxe_core/src/dxe_services.rs
```diff
@@ -435,3 +435,201 @@ pub fn init_dxe_services(system_table: &mut EfiSystemTable) {
...
+                GcdMemoryType::SystemMemory,
+                0x80000000, // Use a high address to avoid conflicts
+                0x1000,
+                efi::MEMORY_WB,
+            );
+
+            println!("add_memory_space returned: {:#x} ({})", result.as_usize(), result.as_usize());
+
+            // Decode the status for debugging
+            let status_value = result.as_usize();
```
makubacki: Similar to the comment about removing prints (which will also remove this), I think this kind of logic (determining `status_name`) makes the overall test more complicated without much testing value.

patina_dxe_core/src/dxe_services.rs
```diff
@@ -435,3 +435,201 @@ pub fn init_dxe_services(system_table: &mut EfiSystemTable) {
...
+
+            // This test shows how add_memory_space should be used in practice
+            // and verifies that our implementation follows the expected pattern
+
+            // Example 1: Adding system memory
+            let result1 = add_memory_space(
+                GcdMemoryType::SystemMemory,
+                0x1000000,                       // 16MB - reasonable address
+                0x100000,                        // 1MB - reasonable size
+                efi::MEMORY_WB | efi::MEMORY_XP, // Typical capabilities
```
makubacki: You probably should a comment explaining why you chose XP for SystemMemory here.

patina_dxe_core/src/dxe_services.rs
```diff
@@ -435,3 +435,201 @@ pub fn init_dxe_services(system_table: &mut EfiSystemTable) {
...
+            let result2 = add_memory_space(
+                GcdMemoryType::MemoryMappedIo,
+                0xFE000000,     // Typical MMIO address
+                0x1000,         // Single page
+                efi::MEMORY_UC, // Uncacheable for MMIO
+            );
+            println!("MMIO example: {:#x}", result2.as_usize());
+
+            // Example 3: Adding reserved memory
+            let result3 = add_memory_space(
```
makubacki: I think an example showing a specific reserved with attributes set would be a little more useful if this function is serving as a demonstration.

patina_dxe_core/src/dxe_services.rs
makubacki, line 1: Can you please give the before and after coverage percentage in the PR description?

patina_dxe_core/src/dxe_services.rs
kat-perez: Added in by mistake. Using MEMORY_WB only

patina_dxe_core/src/dxe_services.rs
makubacki: Thanks

PR #566, opened 2025-07-02
dxe_services: add allocate_memory_space unit tests

### LINKED ISSUE #563: dxe_services: create unit tests for allocate_memory_space

kat-perez:
## Description

Add unit tests to DXE services for allocate memory space function.
Increases unit test coverage +10.13%
Closes #563 

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

```
cargo make coverage
```

## Integration Instructions

N/A

### DIFF
--- a/patina_dxe_core/src/dxe_services.rs
+++ b/patina_dxe_core/src/dxe_services.rs
@@ -535,23 +535,6 @@ mod add_memory_space_tests {
             assert_eq!(result, efi::Status::SUCCESS);
         });
     }
-}
-
-#[cfg(test)]
-mod allocate_memory_space_tests {
-    use super::*;
-    use crate::test_support;
-    use dxe_services::GcdMemoryType;
-
-    fn with_locked_state<F: Fn() + std::panic::RefUnwindSafe>(f: F) {
-        test_support::with_global_lock(|| {
-            unsafe {
-                crate::test_support::init_test_gcd(None);
-            }
-            f();
-        })
-        .unwrap();
-    }
 
     #[test]
     fn test_allocate_memory_space_success() {


patina_dxe_core/src/dxe_services.rs
```diff
@@ -538,3 +538,236 @@ mod add_memory_space_tests {
         });
     }
 }
+
+#[cfg(test)]
+mod allocate_memory_space_tests {
```
os-d: I missed this on the earlier PR, I think we should just keep these in the same `mod tests` so we don't have to repeat the boilerplate

patina_dxe_core/src/dxe_services.rs
```diff
@@ -538,3 +538,236 @@ mod add_memory_space_tests {
...
+                dxe_services::GcdAllocateType::MaxAddressSearchTopDown,
+                GcdMemoryType::SystemMemory,
+                12,     // 4KB alignment
+                0x1000, // 4KB length
+                &mut base_address,
+                1 as _,
+                core::ptr::null_mut(),
+            );
+
+            assert!(
```
makubacki: In other places this is asserting on both conditions (AND) here it is an OR. Is that intended?

kat-perez: i meant to have it only assert on success. updating

patina_dxe_core/src/dxe_services.rs
kat-perez: updated, thanks!

PR #568, opened 2025-07-04
Add ACPI service [NEW]

berlin-with0ut-return:
## Description
Adds Rust-based ACPI services as patina_acpi. The functionality provided is mostly similar to the ACPI Table Protocol and ACPI SDT Protocol. This implementation also produces the necessary C-compatible EDKII protocols.

Note that this is a re-creation of another PR for the same service. I have closed the old one due to an excess of comments that are no longer relevant to the current design.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [X] Includes tests?
- [X] Includes documentation?

## How This Was Tested
- All unit tests pass.
- All integration tests pass.
- Boots to UEFI shell on SBSA and Q35.
- Boots to Windows on SBSA and Q35.
- Passes all SCTs.
- `acpiview` in UEFI shell shows no errors or malformed tables.

## Integration Instructions
To use the Rust component instead of the C ACPI implementation, initialize the `AcpiProviderManager`, `AcpiSystemProtocolManager`, and `GenericAcpiManager` components and remove `AcpiTableDxe` from the FDF.

By default the C implementation is still active.

The current implementation will not boot on physical platforms, as it is missing parts of the SDT protocol.

### DIFF
--- a/components/patina_acpi/src/acpi_protocol.rs
+++ b/components/patina_acpi/src/acpi_protocol.rs
@@ -63,10 +63,11 @@ impl AcpiTableProtocol {
         acpi_table_buffer_size: usize,
         table_key: *mut usize,
     ) -> efi::Status {
-        log::info!("table signature from protocol: {:#x}", unsafe {
-            (*(acpi_table_buffer as *const AcpiTableHeader)).signature
-        });
         if acpi_table_buffer.is_null() || acpi_table_buffer_size < 4 {
+            return efi::Status::INVALID_PARAMETER;
+        }
+
+        if table_key.is_null() {
             return efi::Status::INVALID_PARAMETER;
         }
 
@@ -76,7 +77,7 @@ impl AcpiTableProtocol {
 
         // The size of the allocated table buffer must be large enough to store the whole table.
         let tbl_length = unsafe { (*(acpi_table_buffer as *const AcpiTableHeader)).length } as usize;
-        if tbl_length > acpi_table_buffer_size {
+        if tbl_length != acpi_table_buffer_size {
             return efi::Status::INVALID_PARAMETER;
         }
 
--- a/components/patina_acpi/src/error.rs
+++ b/components/patina_acpi/src/error.rs
@@ -78,7 +78,7 @@ impl From<AcpiError> for efi::Status {
             AcpiError::InvalidSignature => efi::Status::INVALID_PARAMETER,
             AcpiError::FadtAlreadyInstalled => efi::Status::ALREADY_STARTED,
             AcpiError::InstallConfigurationTableFailed => efi::Status::UNSUPPORTED,
-            AcpiError::InvalidTableKey => efi::Status::INVALID_PARAMETER,
+            AcpiError::InvalidTableKey => efi::Status::NOT_FOUND,
             AcpiError::InvalidTableIndex => efi::Status::INVALID_PARAMETER,
             AcpiError::InvalidNotifyUnregister => efi::Status::INVALID_PARAMETER,
             AcpiError::FreeFailed => efi::Status::OUT_OF_RESOURCES,


Cargo.toml
```diff
@@ -28,9 +28,11 @@ lazy_static = { version = "^1" }
 linked_list_allocator = { version = "^0.10" }
 linkme = { version = "^0.3.29" }
 log = { version = "0.4", default-features = false }
+memoffset = {version = "0.9.1" }
```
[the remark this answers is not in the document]
os-d: Nope.

components/patina_acpi/src/acpi_table.rs
```diff
@@ -0,0 +1,482 @@
...
+    ///
+    /// - Caller must ensure the provided table, `T`, has a C compatible layout (typically using `#[repr(C)]`).
+    /// - Caller must ensure that the table's first field is [AcpiTableHeader].
+    pub unsafe fn new<T>(table: T, mm: &Service<dyn MemoryManager>) -> Self {
+        let table = Table::new(table);
+
+        // FACS and UEFI tables must always be located in NVS (by spec).
+        let allocator_type = match table.signature() {
+            signature::FACS | signature::UEFI => EfiMemoryType::ACPIMemoryNVS,
+            _ => EfiMemoryType::ACPIReclaimMemory,
```
Javagedes, line 367: From an offline conversation, I think @os-d mentioned we are always doing ACPIMemoryNVS - or did I mis-remember?

components/patina_acpi/src/acpi_table.rs
```diff
@@ -0,0 +1,482 @@
...
+        unsafe { &self.table.as_ref().header }
+    }
+
+    pub fn header_mut(&mut self) -> &mut AcpiTableHeader {
+        // SAFETY: The table is guaranteed to be a valid ACPI table.
+        unsafe { &mut self.table.as_mut().header }
+    }
+
+    /// Returns a raw byte slice over the entire table.
+    /// SAFETY: self.length must accurately reflect the allocated size of the table.
```
Javagedes: If this is a safety requirement, then you should use `## Safety` (Then the requirement underneath) and mark the function itself as unsafe also.

components/patina_acpi/src/acpi_table.rs
```diff
@@ -0,0 +1,482 @@
...
+    pub fn as_bytes(&self) -> &[u8] {
+        unsafe { slice::from_raw_parts(self.table.as_ptr() as *const u8, self.header().length as usize) }
+    }
+
+    /// Returns a mutable byte slice over the entire table.
+    /// (This is primarily useful for computing the checksum.)
+    /// SAFETY: self.length must accurately reflect the allocated size of the table.
```
Javagedes: If this is a safety requirement, then you should use ## Safety (Then the requirement underneath) and mark the function itself as unsafe also.

components/patina_acpi/src/acpi_table.rs
```diff
@@ -0,0 +1,482 @@
...
+    pub fn as_bytes_mut(&mut self) -> &mut [u8] {
+        unsafe { slice::from_raw_parts_mut(self.table.as_ptr() as *mut u8, self.header().length as usize) }
+    }
+
+    pub fn update_checksum(&mut self, offset: usize) -> Result<(), AcpiError> {
```
Javagedes, line 474: nit: missing documentation

components/patina_acpi/src/acpi_table.rs
```diff
@@ -0,0 +1,482 @@
...
+            bytes[offset] = (0u8).wrapping_sub(sum);
+            Ok(())
+        } else {
+            Err(AcpiError::InvalidChecksumOffset)
+        }
+    }
+
+    /// Returns a reference to the entire AcpiTable.
+    ///
+    /// ## Safety
```
Javagedes, line 493: you have a safety comment here but did not mark the function as unsafe

components/patina_acpi/src/component.rs
```diff
@@ -0,0 +1,166 @@
...
+};
+
+/// Initializes the ACPI provider service.
+#[derive(IntoComponent, Default)]
+pub struct AcpiProviderManager {}
+
+/// Hob that contains information about previously installed ACPI tables.
+#[derive(Copy, Clone, FromHob)]
+#[hob = "9f9a9506-5597-4515-bab6-8bcde784ba87"]
+pub struct AcpiMemoryHob {
```
Javagedes: Per the component requirements, this hob should be defined in a `hob` module: https://sturdy-adventure-nv32gqw.pages.github.io/component/requirements.html

components/patina_acpi/src/acpi_table.rs
```diff
@@ -0,0 +1,482 @@
...
+
+impl<T> Table<T> {
+    /// Creates a new table.
+    ///
+    /// ## Safety
+    ///
+    /// - Caller must ensure the provided table, `T`, has a C compatible layout (typically using `#[repr(C)]`).
+    /// - Caller must ensure that the table's first field is [AcpiTableHeader].
+    pub unsafe fn new(table: T) -> Self {
+        Table { inner: ManuallyDrop::new(table) }
```
cfernald: Should probably do some validation here. For example:
1. Check a valid signature. (ASCII string)
2. Check the length is not less then `sizeof::<T>()`
3. Enforce the checksum is correct, or set it?

And return a `Result` to account for bad data.

components/patina_acpi/src/service.rs
```diff
@@ -0,0 +1,131 @@
...
+        self.provider_service.uninstall_acpi_table(table_key)
+    }
+
+    /// Retrieves an ACPI table by its table key.
+    ///
+    /// The `table_key` is the opaque reference returned by `install_acpi_table`.
+    /// The generic type `T` should be the expected type of the table.
+    ///
+    /// The RSDP and XSDT cannot be accessed through `get_acpi_table`.
+    pub fn get_acpi_table<T: 'static>(&self, table_key: TableKey) -> Result<&T, AcpiError> {
```
cfernald: Do we not need a way to lookup table[s] by signature? Or is the expectation they use iter to do this?

components/patina_acpi/src/config.rs
```diff
@@ -0,0 +1,27 @@
+/// Initialization configuration for ACPI provider.
+#[derive(Debug, Clone, Copy, Default)]
+pub struct AcpiProviderInit {
```
Javagedes: Is any of this needed by other components? Or maybe just a subset? I think we should probably move most, or all, of this to be private initialization data for the `AcpiProviderManager` Component. If some config like the version information needs to be public config, then we could have the AcpiProviderManager produce that config value in it's own component - something like this:

``` rust

impl AcpiProviderManager {

    fn entry_point(
        self,
        ...
        mut config: ConfigMut<AcpiProvider>,
    ) -> Result<()> {

        ...
        *config.oem_revision = 40;
    }
}
```

definitely not saying you need to do this. But it could reduce the integration for a platform - e.g. they don't need to add 3 components and a config, just 3 configs.

components/patina_acpi/src/lib.rs
```diff
@@ -0,0 +1,56 @@
...
+//! `AcpiSystemTableProtocolManager` publishes the ACPI Table and ACPI SDT protocols.
+//!
+//! This library also provides a service interface, `AcpiProvider`, which can be consumed by other components to perform ACPI operations.
+//!
+//! ## Examples and Usage
+//!
+//! To initialize the `AcpiProviderManager`, the configuration should be customized with the correct platform values (`oem_id`, etc).
+//! In the platform start routine, provide these configuration values and initialize a new `AcpiProviderManager` instance.
+//!
+//! ```rust,ignore
```
Javagedes, line 14: I would do `no_run`, not `ignore`. With `ignore`, if our interface changes, this will become outdated without us knowing. You can copy what we do in patina_dxe_core to get a compilable version of this. It basically involves making some mock implementations :)

components/patina_acpi/src/acpi_table.rs
```diff
@@ -0,0 +1,482 @@
+//! ACPI Table Definitions.
+//!
+//! Defines standard formats for system ACPI tables.
+//! Supports only ACPI version >= 2.0.
+//! Fields corresponding to ACPI 1.0 are preceded with an underscore (`_`) and are not in use.
```
Javagedes, line 5: nit: Missing license info

components/patina_acpi/src/acpi.rs
```diff
@@ -0,0 +1,1183 @@
+use alloc::collections::btree_map::BTreeMap;
```
Javagedes, line 13: nit: missing module information and license information.

components/patina_acpi/src/acpi_protocol.rs
```diff
@@ -0,0 +1,194 @@
+//! ACPI C Protocol Definitions.
+//!
+//! Wrappers for the C ACPI protocols to call into Rust ACPI implementations.
```
Javagedes, line 3: missing license information

components/patina_acpi/src/config.rs
```diff
@@ -0,0 +1,27 @@
+/// Initialization configuration for ACPI provider.
```
Javagedes: missing module information and license info

components/patina_acpi/src/error.rs
```diff
@@ -0,0 +1,89 @@
+use r_efi::efi;
```
Javagedes, line 12: missing module documentation and license information

components/patina_acpi/src/integration_test.rs
```diff
@@ -0,0 +1,45 @@
+use patina_sdk::component::service::Service;
```
Javagedes: missing module information (mainly licensing)

components/patina_acpi/src/signature.rs
```diff
@@ -0,0 +1,30 @@
+use r_efi::efi;
```
Javagedes, line 14: missing module documentation and license information

components/patina_acpi/src/acpi_protocol.rs
```diff
@@ -0,0 +1,194 @@
...
+        }
+
+        if acpi_table_buffer_size < mem::size_of::<AcpiTableHeader>() {
+            return efi::Status::INVALID_PARAMETER;
+        }
+
+        // SAFETY: acpi_table_buffer is checked non-null and large enough to read an AcpiTableHeader.
+        if let Some(global_mm) = ACPI_TABLE_INFO.memory_manager.get() {
+            let acpi_header = unsafe { *(acpi_table_buffer as *const AcpiTableHeader) };
+            let acpi_table = unsafe { AcpiTable::new(acpi_header, global_mm) };
```
Javagedes: This will not work as-is. Since you are casting it to the AcpiTableHeader, we only allocate enough space for an AcpiTableHeader, which means we will lose all of the actual data of the table.

Instead, you will need to make a `AcpiTable::new_from_ptr` function that utilizes `memory_manager.get_allocator()` to get an `Allocator` Interface for which you can create a `Layout` struct for the exact number of bytes needed (The size value in the header), and use that to call the `allocate` function. You would then copy the bytes from the pointer provided by the C code into this newly allocated pointer, and use that as the inner value of the `AcpiTable`

There may be a better way, the above is just one suggestion. But overall you just need to make sure the entire table is allocated, not just the header

components/patina_acpi/src/service.rs
```diff
@@ -0,0 +1,131 @@
...
+        let acpi_table = self.provider_service.get_acpi_table(table_key)?;
+
+        // There may be ACPI tables whose type is unknown at installation, due to installation from the HOB or a C protocol.
+        // In these cases, the type is is unspecified (AcpiTableHeader instead of a specific table type), so we skip type checking.
+        // In all other cases, verify the type provided by the user is valid.
+        if acpi_table.type_id() != TypeId::of::<AcpiTableHeader>() && acpi_table.type_id() != TypeId::of::<T>() {
```
Javagedes: With my previously mentioned change, we won't be able to do this type_id check for tables that come from the C protocol - or we can and it will just fail if you cast to anything other then AcpiTableHeader.

I would suggest providing a `get_acpi_table_unchecked` which skips validating the typeid.

This unchecked is also useful because a single provided table could be represented by multiple different types. Take this as an example:

struct TableA {
   hdr: AcpiTableHeader,
   value1: u32,
}

struct TableB {
    hdr: AcpiTableHeader,
    value1: u32,
    value2: u64
}

If we create a `TableB`, we could technically treat it as a `TableA` or a `TableB`, but with only this checked value, we cannot ever cast to `TableA`

components/patina_acpi/src/acpi.rs
```diff
@@ -0,0 +1,1183 @@
...
+use core::{
+    cell::OnceCell,
+    ffi::c_void,
+    mem::{self, offset_of},
+    sync::atomic::{AtomicUsize, Ordering},
+};
+
+use crate::{
```
Javagedes: nit: you have `use crate::` multiple times. Should probably reduce this down to 1 large `use crate` statement.

components/patina_acpi/src/service.rs
```diff
@@ -0,0 +1,131 @@
...
+use patina_sdk::component::service::memory::MemoryManager;
+use patina_sdk::component::service::{IntoService, Service};
+use r_efi::efi;
+
+use crate::acpi_table::{AcpiTable, AcpiTableHeader};
+use crate::error::AcpiError;
+
+/// Represents an opaque reference to an installed ACPI table.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord)]
+pub struct TableKey(pub(crate) usize);
```
Javagedes, line 23: This is a pub type, so needs documented

components/patina_acpi/src/service.rs
```diff
@@ -0,0 +1,131 @@
...
+
+/// A notification function that is called when a new ACPI table is installed.
+pub type AcpiNotifyFn = fn(&AcpiTableHeader, u32, usize) -> efi::Status;
```
Javagedes: This is a pub type, so needs documented

components/patina_acpi/src/service.rs
```diff
@@ -0,0 +1,131 @@
...
+use r_efi::efi;
+
+use crate::acpi_table::{AcpiTable, AcpiTableHeader};
+use crate::error::AcpiError;
+
+/// Represents an opaque reference to an installed ACPI table.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord)]
+pub struct TableKey(pub(crate) usize);
+
+/// A notification function that is called when a new ACPI table is installed.
```
cfernald, line 25: Add description of the arguments to the comment. i.e. what is the u32 and usize expected to be?

components/patina_acpi/src/service.rs
```diff
@@ -0,0 +1,131 @@
...
+            .as_ptr(); // get a *const T
+
+        Ok(unsafe { &*raw_table_ptr })
+    }
+
+    /// Registers or unregisters a function which will be called whenever a new ACPI table is installed.
+    ///
+    /// If `should_register` is true, it will register the function.
+    /// Otherwise, it will unregister the function if it exists in the current notify list.
+    pub fn register_notify(&self, should_register: bool, notify_fn: AcpiNotifyFn) -> Result<(), AcpiError> {
```
cfernald: nit: another function is cheap. I would just have a `register_notify` and `unregister_notify`, which can still call the same `self.provider_service.register_notify(should_register, notify_fn)` with the appropriate true/false. I think it's just a more explicit/clear interface.

components/patina_acpi/src/service.rs
```diff
@@ -0,0 +1,131 @@
...
+    ///
+    /// The RSDP and XSDT are not included in the list of iterable ACPI tables.
+    pub fn iter_tables(&self) -> Vec<AcpiTable> {
+        self.provider_service.iter_tables()
+    }
+}
+
+/// The `AcpiTableManager` provides functionality for installing, uninstalling, and accessing ACPI tables.
+/// This struct serves as the API by which internal implementations can provide custom ACPI implementation.
+pub trait AcpiProvider {
```
cfernald: Does this need to be pub, or can it be pub(crate)?

components/patina_acpi/src/acpi.rs
```diff
@@ -0,0 +1,1183 @@
...
+
+        // Read the header to validate the XSDT signature is valid.
+        // SAFETY: `xsdt_address` has been validated to be non-null.
+        let xsdt_header = rsdp.xsdt_address as *const AcpiTableHeader;
+        if (unsafe { *xsdt_header }).signature != signature::XSDT {
+            return Err(AcpiError::InvalidSignature);
+        }
+
+        // SAFETY: We validate that the XSDT is non-null and contains the right signature.
+        let xsdt_ptr = rsdp.xsdt_address as *mut AcpiTableHeader;
```
cfernald: I don't think this needs to be a mut pointer.

components/patina_acpi/src/integration_test.rs
```diff
@@ -0,0 +1,45 @@
...
+use patina_sdk::test::patina_test;
+
+use crate::{
+    acpi_table::{AcpiDsdt, AcpiFacs, AcpiFadt, AcpiTableHeader},
+    service::AcpiTableManager,
+    signature::{self, ACPI_HEADER_LEN},
+};
+
```
cfernald, line 27: I would add an integration test that uses the protocol as well.

components/patina_acpi/src/acpi_table.rs
os-d: You are misremembering, we should always do ACPIReclaimMemory (except for the special cases), so this is correct.

components/patina_acpi/src/acpi_table.rs
```diff
@@ -0,0 +1,482 @@
...
+    /// - Caller must ensure the provided table, `T`, has a C compatible layout (typically using `#[repr(C)]`).
+    /// - Caller must ensure that the table's first field is [AcpiTableHeader].
+    pub unsafe fn new(table: T) -> Self {
+        Table { inner: ManuallyDrop::new(table) }
+    }
+
+    /// Returns the signature of the ACPI table.
+    pub fn signature(&self) -> u32 {
+        // SAFETY: [Self::new] ensures that the first field is a u32.
+        unsafe { self.signature }
```
os-d, line 332: For my understanding, why do these three functions have unsafe blocks?

components/patina_acpi/src/acpi.rs
```diff
@@ -0,0 +1,1183 @@
...
+    fn install_fadt(&self, mut fadt_info: AcpiTable) -> Result<TableKey, AcpiError> {
+        if self.acpi_tables.read().get(&Self::FADT_KEY).is_some() {
+            // FADT already installed. By spec, only one copy of the FADT should ever be installed, and it cannot be replaced.
+            return Err(AcpiError::FadtAlreadyInstalled);
+        }
+
+        // If the FACS is already installed, update the FADT's x_firmware_ctrl field.
+        // If not, it will be updated when the FACS is installed.
+        if let Some(facs) = self.acpi_tables.read().get(&Self::FACS_KEY) {
+            unsafe { fadt_info.as_mut::<AcpiFadt>() }.inner.x_firmware_ctrl = facs.as_ptr() as u64;
```
os-d, line 352: For my understanding, what makes this (and the other instances below) unsafe? as_mut already has an internal unsafe, but not is an unsafe function, is it unsafe to operate on the AcpiTable? Or because the generic info is filled in?

components/patina_acpi/src/acpi.rs
```diff
@@ -0,0 +1,1183 @@
...
+        Ok(())
+    }
+
+    // Performs checksums on shared ACPI tables (the XSDT and RSDP track other tables).
+    pub(crate) fn checksum_common_tables(&self) -> Result<(), AcpiError> {
+        if let Some(rsdp_table) = self.acpi_tables.write().get_mut(&Self::RSDP_KEY) {
+            rsdp_table.update_checksum(offset_of!(AcpiRsdp, extended_checksum))?;
+        }
+
+        if let Some(ref mut xsdt_data) = *self.xsdt_metadata.write() {
```
os-d, line 637: For this we can't use update_checksum because xsdt_metadata is not an AcpiTable, is that correct?

components/patina_acpi/src/acpi.rs
```diff
@@ -0,0 +1,1183 @@
...
+    }
+
+    /// Deletes a table from the list of installed tables and frees its memory.
+    fn delete_table(&self, physical_addr: u64, signature: u32) -> Result<(), AcpiError> {
+        match signature {
+            signature::FADT => {
+                self.acpi_tables.write().remove(&Self::FADT_KEY);
+            }
+            signature::FACS => {
+                self.acpi_tables.write().remove(&Self::FACS_KEY);
```
os-d: nit: Remove after setting x_firmware_ctrl to 0? break before make, not that it matters too much here. Same with DSDT below.

components/patina_acpi/src/acpi.rs
```diff
@@ -0,0 +1,1183 @@
...
+
+        Ok(())
+    }
+
+    /// Removes an address entry from the XSDT when a table is uninstalled.
+    fn remove_table_from_xsdt(&self, table_address: u64) -> Result<(), AcpiError> {
+        if let Some(ref mut xsdt_data) = *self.xsdt_metadata.write() {
+            // Calculate where entries are in the slice.
+            let entries_n_bytes = ACPI_XSDT_ENTRY_SIZE * xsdt_data.n_entries;
+            let entries_bytes = xsdt_data
```
os-d, line 578: Can you just index into the slice since we know the exact size of an XSDT entry? Do we need to do this as bytes instead of slice entries?

components/patina_acpi/src/acpi.rs
```diff
@@ -0,0 +1,1183 @@
...
+                let start_ptr = ACPI_HEADER_LEN + idx * ACPI_XSDT_ENTRY_SIZE; // Find where the target entry starts.
+                let end_ptr = ACPI_HEADER_LEN + xsdt_data.n_entries * ACPI_XSDT_ENTRY_SIZE; // Find where the XSDT ends.
+
+                // Shift all entries after the one being removed to the left.
+                // [.. before .. | target | <- .. after .. ]
+                // becomes [.. before .. | .. after.. ]
+                xsdt_data.slice.copy_within(start_ptr + ACPI_XSDT_ENTRY_SIZE..end_ptr, start_ptr);
+
+                // Decrement entries.
+                xsdt_data.n_entries -= 1;
```
os-d, line 597: 0 the old entry? Okay to not, though we 0'd all the entries when we started, so seems like either we don't do that or we consistently 0.

components/patina_acpi/src/acpi_table.rs
```diff
@@ -0,0 +1,482 @@
...
+#[repr(C)]
+#[derive(Default, Copy, Clone)]
+pub struct AcpiXsdt {
+    pub(crate) header: AcpiTableHeader,
+}
+
+/// Stores implementation-specific data about the XSDT.
+pub(crate) struct AcpiXsdtMetadata {
+    pub(crate) n_entries: usize,
+    pub(crate) max_capacity: usize,
```
os-d, line 194: iirc, when we discussed this, we talked about the metadata also have a reference to the xsdt header (e.g. an &mut AcpiXsdt) so that you can do simple operations on it instead of having to interact with the slice, e.g. get_length just becomes `return self.xsdt_header.length` (or whatever) instead of having to muck with the slice. Same with all of the other associated functions here, because these only operate on the header. What do you think?

components/patina_acpi/src/acpi_table.rs
```diff
@@ -0,0 +1,482 @@
...
+        let table = Table::new(table);
+
+        // FACS and UEFI tables must always be located in NVS (by spec).
+        let allocator_type = match table.signature() {
+            signature::FACS | signature::UEFI => EfiMemoryType::ACPIMemoryNVS,
+            _ => EfiMemoryType::ACPIReclaimMemory,
+        };
+
+        let table =
+            NonNull::from(Box::leak(Box::new_in(table, mm.get_allocator(allocator_type).unwrap()))).cast::<Table>();
```
os-d: Why do we need the NonNull here? As opposed to just the Box? Then you get rid of all the unsafes below to access it?

components/patina_acpi/src/signature.rs
```diff
@@ -0,0 +1,30 @@
...
+pub const FACP: u32 = 0x50434146;
+pub const DSDT: u32 = 0x54534444;
+pub const XSDT: u32 = 0x54445358;
+pub const FADT: u32 = FACP; // For legacy ACPI reasons, the FADT has signature 'FACP'.
+
+pub const ACPI_TABLE_GUID: efi::Guid =
+    efi::Guid::from_fields(0x8868E871, 0xE4F1, 0x11D3, 0xBC, 0x22, &[0x00, 0x80, 0xC7, 0x3C, 0x88, 0x81]);
+
+pub const ACPI_HEADER_LEN: usize = 36;
+pub const MAX_INITIAL_ENTRIES: usize = 32;
```
os-d: Some of these, like this one, look like they can be `pub(crate)`

components/patina_acpi/src/acpi_table.rs
Javagedes: any union access is inherently unsafe, and rust makes you mark it as such, or you get a compile error.

components/patina_acpi/src/acpi.rs
Javagedes: You are referencing the wrong `as_mut`. The unsafe as_mut is in the `AcpiTable` struct (as of right now, acpi_table:412). This is unsafe because we are casting it to type T, which is not guaranteed to be castable, and could cause UB.

One option to stop this from being unsafe is to move the type id to this layer and do validation there, which returns a result if the type id's do not match.

components/patina_acpi/src/acpi_table.rs
os-d: Ah, I missed this was a union, thanks.

components/patina_acpi/src/acpi_table.rs
Javagedes: `NonNull` is what allows us to do the type erasure and allows us to use just `AcpiTable` instead of `AcpiTable<T>`, which then allows us to store the structs in the same data structure. When we do this creation we create a `NonNull<T>`, then cast it to a `NonNull<u8>` and store that as the AcpiTable. Then whenever we want to access the underlying table (not just the table header), we use `as_mut` or `as_ref` which casts `NonNull<u8>` back to `NonNull<T>`

os-d: Ah, I see. That's unfortunate that it forces us into (almost) raw pointers. But, this has gone through many iterations, I won't ask for further changes.

But my speculation is, do we even need the generic at all? If you just have an AcpiTable is a header followed by a slice? Since we never operate on the data in the common code.

Javagedes: A user can request a table and operate on it. If the interface is that we just give you a header, then the user is responsible for turning that into the table they care about. I don't trust the user.

This abstraction around a raw pointer does it for them and allows them to get back the table in a nice, safe, way.

Also we do operate on the data in the common code, for a few of the special tables we maintain - so we need the functionality too.

os-d: Well, in the common code we never operate on an AcpiTable, do we? We operate on an AcpiXsdt, for example, which is a real type. I don't want to go back through the thousands of lines to see if this assertion is true, but even if it weren't we'd either be operating on the header, so doesn't matter, or it is trivial to convert the AcpiTable to an AcpiXsdt.

I see your point about the user being able to just operate on their type and I think that is valuable. I think I would have to see a consumer of the service to better understand that flow. Again, I'm not asking for changes, just trying to understand why we need the NonNull and if there is a better way to do it. Just seems odd to have to convert a Box to a NonNull, but I understand the generic is forcing us to.

components/patina_acpi/src/service.rs
berlin-with0ut-return: it is expected they would use iter, which is similar to how the C protocol functions

components/patina_acpi/src/acpi.rs
berlin-with0ut-return: yea, i tried a few different designs, but it's easier to store the XSDT as bytes in this metadata struct rather than as an AcpiTable due to the way its entries are laid out and the need to store metadata like the current number of entries.

os-d: Okay, I had another question about xsdt_metadata and whether it made sense to be a slice. I'll close this and we can discuss there.

components/patina_acpi/src/acpi_table.rs
berlin-with0ut-return: i found that it was a bit annoying to deal with lifetimes and manage having two pointers to technically the same location in memory. with the helper getters and setters, i don't think it's too complex to represent everything as a slice :)

components/patina_acpi/src/acpi_table.rs
```diff
@@ -0,0 +1,610 @@
...
+    /// - Caller must ensure the pointer refers to a valid ACPI table.
+    /// - Caller must ensure `table_length` is correctly specifies the length of the table, including the header and any trailing data bytes.
+    pub unsafe fn new_from_ptr(
+        header_ptr: *const AcpiTableHeader,
+        mm: &Service<dyn MemoryManager>,
+    ) -> Result<Self, AcpiError> {
+        // SAFETY: If function preconditions are met, the pointer is valid and points to a valid ACPI table header.
+        let table_signature;
+        let table_length;
+        unsafe {
```
Javagedes: If you want it to be one line like what you are doing you can do something along the lines of

```rust
        // SAFETY: If function preconditions are met, the pointer is valid and points to a valid ACPI table header.
        let (table_signature, table_length) = unsafe {
            ((*header_ptr).signature, (*header_ptr).length as usize)
        };
```

I'm not a fan of C style variable declarations without being set. Rust compiler will yell at you if it gets used before set, but I still don't like it :)

components/patina_acpi/src/acpi_table.rs
```diff
@@ -0,0 +1,610 @@
...
+        assert_eq!(header.oem_table_id, *b"test_tes");
+        assert_eq!(header.oem_revision, 0xDEADBEEF);
+        assert_eq!(header.creator_id, 0xCAFEBABE);
+        assert_eq!(header.creator_revision, 0xFEEDFACE);
+
+        // Check that the body bytes are correct.
+        let bytes = unsafe { acpi_table.as_bytes() };
+        let body_offset = mem::size_of::<AcpiTableHeader>();
+        assert_eq!(&bytes[body_offset..body_offset + 3], &[42, 43, 44]);
+    }
```
Javagedes, line 622: I would also add this to the test:

```rust
assert_eq!(unsafe { acpi_table.as_ref::<TestTable>().body }, [42, 43, 44]);
```

components/patina_acpi/src/acpi_table.rs
```diff
@@ -0,0 +1,610 @@
...
+            Err(AcpiError::InvalidChecksumOffset)
+        }
+    }
+
+    /// Returns a reference to the entire AcpiTable.
+    ///
+    /// ## Safety
+    ///
+    /// - Caller must ensure that the provided table format is the same as `T`.
+    pub unsafe fn as_ref<T>(&self) -> &T {
```
Javagedes, line 496: I noticed you removed the type_id check from this function (we talked about how it could not support tables coming form the C protocol.

I think I mentioned it before, but I think the type check ins `as_ref` and `as_mut` is great, and will enable you to remove the `unsafe` from the functions because the typeid comparison ensures the provided table format is the same as `T`.

I just think you need to add `as_ref_unchecked` and `as_mut_unchecked` which are basically these two functions as they are now, which does not do the type id validation, and pushes that safety validation to the user.

components/patina_acpi/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+[dependencies]
+log = { workspace = true }
+memoffset = { workspace = true }
+mockall = { workspace = true, optional = true }
+patina_sdk = { workspace = true, features = ["alloc"] }
+r-efi = { workspace = true }
+scroll = { workspace = true }
+spin = { workspace = true }
+
+[dev-dependencies]
```
Javagedes, line 20: need to add `patina_sdk = { workspace = true, features = ["mockall"] }` to [dev-dependencies] so that the mocked memory manager is available during your tests.

components/patina_acpi/src/service.rs
```diff
@@ -0,0 +1,162 @@
...
+    ///
+    /// The generic type `T` should be the expected type of the table, which should match the type passed in during installation.
+    /// To cast retrieved tables to different types, or to retrieve tables across the C FFI protocol interface,
+    /// use `get_acpi_table_unchecked` for a untyped retrieval.
+    ///
+    /// The RSDP and XSDT cannot be accessed through `get_acpi_table`.
+    pub fn get_acpi_table<T: 'static>(&self, table_key: TableKey) -> Result<&T, AcpiError> {
+        let acpi_table = self.provider_service.get_acpi_table(table_key)?;
+
+        // There may be ACPI tables whose type is unknown at installation, due to installation from the HOB or a C protocol.
```
berlin-with0ut-return, line 85: can we return a copy?

berlin-with0ut-return: immutability

components/patina_acpi/src/acpi_protocol.rs
```diff
@@ -0,0 +1,203 @@
...
+    ///
+    /// # Errors
+    ///
+    /// Returns [`INVALID_PARAMETER`](r_efi::efi::Status::INVALID_PARAMETER) the table buffer is null or too small to contain the ACPI table header.
+    /// Returns [`UNSUPPORTED`](r_efi::efi::Status::UNSUPPORTED) if the system page size is not 64B-aligned (required for FACS in ACPI 2.0+).
+    /// Returns [`UNSUPPORTED`](r_efi::efi::Status::UNSUPPORTED) if boot services cannot install the given table format.
+    /// Returns [`OUT_OF_RESOURCES`](r_efi::efi::Status::OUT_OF_RESOURCES) if allocating memory for the table fails.
+    /// Returns [`ALREADY_STARTED`](r_efi::efi::Status::ALREADY_STARTED) if the FADT already exists and `install` is called on a FADT again.
+    /// Returns [`NOT_STARTED`](r_efi::efi::Status::NOT_STARTED) if memory or boot services are not properly initialized.
+    extern "efiapi" fn install_acpi_table_ext(
```
berlin-with0ut-return, line 60: more checks for known tables

makubacki:
@berlin-with0ut-return to add physical platform roadmap and current limitations.

makubacki:
> As noted offline, I'm good to check this in and then iterate on it, since it boots Q35 and SBSA to Windows. We should file an issue to track the remaining work to get it working on x64 and ARM64 physical platforms.

@berlin-with0ut-return can you reply with this?

> add physical platform roadmap and current limitations.

We need to keep `main` stable and release ready. If there is no impact there, please explain that. The other option is to create an acpi feature branch as described in the [patina release process](https://github.com/OpenDevicePartnership/patina/blob/main/docs/src/rfc/text/0015-patina-release-process.md#branching).

berlin-with0ut-return:
> > As noted offline, I'm good to check this in and then iterate on it, since it boots Q35 and SBSA to Windows. We should file an issue to track the remaining work to get it working on x64 and ARM64 physical platforms.
> 
> @berlin-with0ut-return can you reply with this?

The main tasks left are:
 - implement remaining members of SDT protocol
- update Windows to use 64-bit ACPI fields (this is technically not necessary for boot since we force allocation in a 32-bit range, but needs to be updated to follow latest ACPI spec)
- update `mu_basecore` to use 64-bit ACPI fields (there is only one instance of this that needs to be fixed that I could find)
> 
> > add physical platform roadmap and current limitations.
> 
> We need to keep `main` stable and release ready. If there is no impact there, please explain that. The other option is to create an acpi feature branch as described in the [patina release process](https://github.com/OpenDevicePartnership/patina/blob/main/docs/src/rfc/text/0015-patina-release-process.md#branching).

There is no impact to main since I have chosen not to enable the ACPI component. As long as it is not enabled, we will continue using `AcpiTableDxe` from the C code. For now I suppose it's more of a proof-of-concept. Happy to put it in a feature branch if others think this is the right move

makubacki:
> There is no impact to main since I have chosen not to enable the ACPI component. As long as it is not enabled, we will continue using `AcpiTableDxe` from the C code. For now I suppose it's more of a proof-of-concept. Happy to put it in a feature branch if others think this is the right move

External dependencies like necessary Windows changes should be clearly documented so potential users are aware of what to expect. Imo, the feature branch makes sense if you need to refine the code itself before others should use it. It sounds like that *might* be the case.

os-d:
>  Imo, the feature branch makes sense if you need to refine the code itself before others should use it. It sounds like that _might_ be the case.

Yeah, I wasn't thinking about this in terms of the new branching scheme. I agree a feature branch makes sense until physical platforms can use this. I think it will help track the changes required for that instead of continuing to iterate on this large PR.

berlin-with0ut-return:
i've created a feature branch [feature/acpi](https://github.com/OpenDevicePartnership/patina/tree/feature/acpi). i'll leave this PR up for another week in case anyone else wants to comment, but otherwise further updates will be made there

berlin-with0ut-return:
closing as all comments have been resolved and this feature has been moved to a branch

PR #569, opened 2025-07-08
Add performance support for ARM & non SMM platforms 

liqiqiii:
## Description

Add performance support for ARM platforms.
ARM doesn't have PEI phase so it doesn't have the hob required. Make the Hob optional.
Make SMM communication protocol Option<> too to make it work for non SMM platforms.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested with ARM platform, get performance records correctly.

### DIFF
--- a/components/patina_performance/src/lib.rs
+++ b/components/patina_performance/src/lib.rs
@@ -235,7 +235,7 @@ impl Performance {
                 EventType::NOTIFY_SIGNAL,
                 Tpl::CALLBACK,
                 Some(fetch_and_add_mm_performance_records),
-                Box::new((BB::clone(&boot_services), Some(mm_comm_region), fbpt)),
+                Box::new((BB::clone(&boot_services), mm_comm_region, fbpt)),
                 &EVENT_GROUP_READY_TO_BOOT,
             )?;
         } else {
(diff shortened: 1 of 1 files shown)


components/patina_performance/src/lib.rs
```diff
@@ -297,7 +316,7 @@ extern "efiapi" fn report_fbpt_record_buffer<BB, B, RR, R, F>(
 /// Event callback that add the SMM performance record to the FBPT.
 extern "efiapi" fn fetch_and_add_mm_performance_records<BB, B, F>(
     event: efi::Event,
-    ctx: Box<(BB, MmCommRegion, &TplMutex<'static, F, B>)>,
+    ctx: Box<(BB, Option<MmCommRegion>, &TplMutex<'static, F, B>)>,
```
Javagedes: Since this callback is only registered if the mm_comm_region exists (See L233 of what you added), then it seems like this function can stay as `MmCommRegion` (rather than `Option<MmCommRegion>`, and you don't need to wrap `mm_comm_region` in a `Some` on L238 above.

This makes it so that you don't need to change anything in `fetch_and_add_mm_performance_records`

liqiqiii: Fixed, thanks Joey!

Issue #544, opened 2025-06-19
[Bug]: Uses of `ptr::write_volatile` in `AdvLoggerMessageEntry::init_from_memory` cause UB


wesleywiser:
### Is there an existing issue for this?

- [x] I have searched existing issues

### Current Behavior

`AdvLoggerMessageEntry::init_from_memory` contains a few uses of `ptr::write_volatile` which were upgraded from `ptr:write` (or similar functions) in various PRs because the compiler was undesirably optimizing out the writes. While `ptr::write_volatile` prevents the compiler from optimizing out the writes, the use here is UB because we're mutating memory the Rust compiler thinks is immutable which could cause other issues like the writes not being "visible" unless `ptr::read_volatile` is also used.

Looking at these lines:

https://github.com/OpenDevicePartnership/patina/blob/3bb9821ad250baef139e1f3be5773dde7941b676/components/patina_adv_logger/src/memory_log.rs#L298-L301

and then tracing `address` back to the source:

https://github.com/OpenDevicePartnership/patina/blob/3bb9821ad250baef139e1f3be5773dde7941b676/components/patina_adv_logger/src/memory_log.rs#L278

https://github.com/OpenDevicePartnership/patina/blob/3bb9821ad250baef139e1f3be5773dde7941b676/components/patina_adv_logger/src/memory_log.rs#L185-L186

https://github.com/OpenDevicePartnership/patina/blob/3bb9821ad250baef139e1f3be5773dde7941b676/components/patina_adv_logger/src/memory_log.rs#L146

We see that the place we're writing to is `&self` which is not considered writable by Rust. I tried looking a bit further up the call stack as well and I'm not sure this is the only issue of this kind. Depending on where the memory that backs the `hob_list` comes from in `AdvancedLoggerComponent::init_advanced_logger`, there could be more of these kinds of issues with the trailing log writes.

### Expected Behavior

The best but most invasive solution would be for `AdvLoggerMessageEntry::init_from_memory` to be passed memory that Rust knows is going to be mutated via `&mut self` in the call site. At a quick glance, I can't see whether or not that is viable.

If that isn't possible, using interior mutability via `Cell<T>` may be which should also resolve the issue.

### Steps To Reproduce

n/a

### Build Environment

```markdown
n/a
```

### Version Information

```text
Commit: 3bb9821ad250baef139e1f3be5773dde7941b676
```

### Urgency

Medium

### Are you going to fix this?

Someone else needs to fix it

[...]

PR #570, opened 2025-07-09
AdvLogger: Remove UB and improve parsing safety

cfernald:
## Description

Undefined behavior and safety improvements:
- Removes the undefined behavior by moving the log data into a UnsafeCell.
- More tightly binds buffer parsing by using data slice and `ZeroCopy` to perform data casting.

General Cleanup:
- Introduce the wrapper `AdvancedLog` type to abstract the entire log structure and routines.
- Simplified to use `LogEntry` for all intput and output of entries for `AdvancedLog`
- Allow for readonly version of log for parsing

CLOSES: #544 

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- [x] Tested on QEMU
- [x] Unit tests
- [x] Integration Test
- [x] Manual parsing test

## Integration Instructions

N/A

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -53,6 +53,8 @@ uart_16550 = { version = "^0.3.2" }
 uefi_corosensei = { version = "0.1.3", default-features = false, registry = "patina-fw" }
 uuid = { version = "1.8", default-features = false }
 x86_64 = { version = "=0.15.2", default-features = false }
+zerocopy = { version = "0.8" }
+zerocopy-derive = { version = "0.8" }
 
 # dev dependencies
 criterion = { version = "^0.5" }
--- a/components/patina_adv_logger/src/component.rs
+++ b/components/patina_adv_logger/src/component.rs
@@ -20,11 +20,7 @@ use patina_sdk::{
 };
 use r_efi::efi;
 
-use crate::{
-    logger::AdvancedLogger,
-    memory_log::{self, AdvLoggerInfo},
-    protocol::AdvancedLoggerProtocol,
-};
+use crate::{logger::AdvancedLogger, memory_log, protocol::AdvancedLoggerProtocol};
 
 /// C struct for the internal Advanced Logger protocol for the component.
 #[repr(C)]
@@ -113,15 +109,11 @@ where
     /// Installs the Advanced Logger Protocol for use by non-local components.
     ///
     fn entry_point(self, bs: StandardBootServices) -> Result<()> {
-        let log_info = match self.adv_logger.get_log_info() {
-            Some(log_info) => log_info,
-            None => {
-                log::error!("Advanced logger not initialized before component entry point!");
-                return Err(EfiError::NotStarted);
-            }
+        let Some(address) = self.adv_logger.get_log_address() else {
+            log::error!("Advanced logger not initialized before component entry point!");
+            return Err(EfiError::NotStarted);
         };
 
-        let address = log_info as *const AdvLoggerInfo as efi::PhysicalAddress;
         let protocol = AdvancedLoggerProtocolInternal {
             protocol: AdvancedLoggerProtocol::new(Self::adv_log_write, address),
             adv_logger: self.adv_logger,
@@ -149,6 +141,8 @@ mod tests {
     use mu_pi::hob::{GUID_EXTENSION, GuidHob, header::Hob};
     use patina_sdk::serial::uart::UartNull;
 
+    use crate::memory_log::AdvancedLog;
+
     use super::*;
 
     static TEST_LOGGER: AdvancedLogger<UartNull> =
@@ -160,7 +154,7 @@ mod tests {
         let log_address = log_buff as *const u8 as efi::PhysicalAddress;
 
         // initialize the log so it's valid for the hob list
-        unsafe { AdvLoggerInfo::initialize_memory_log(log_address, LOG_LEN as u32) };
+        unsafe { AdvancedLog::initialize_memory_log(log_address, LOG_LEN as u32) };
 
         const HOB_LEN: usize = size_of::<GuidHob>() + size_of::<efi::PhysicalAddress>();
         let hob_buff = Box::into_raw(Box::new([0_u8; HOB_LEN]));
--- a/components/patina_adv_logger/src/integration_test.rs
+++ b/components/patina_adv_logger/src/integration_test.rs
@@ -47,9 +47,11 @@ fn adv_logger_test(bs: StandardBootServices) -> patina_sdk::test::Result {
     u_assert_eq!(efi_status, efi::Status::SUCCESS, "adv_logger_test: Failed to write to the advanced logger protocol.");
 
     // Check that the strings were added to the log.
-    let log_info = unsafe { memory_log::AdvLoggerInfo::adopt_memory_log(protocol.log_info) };
-    u_assert!(log_info.is_some(), "adv_logger_test: Failed to adopt the memory log.");
-    let log_info = log_info.unwrap();
+    // SAFETY: We know this memory is safe and well structure as we just created it
+    //         using the counterpart functions.
+    let log = unsafe { memory_log::AdvancedLog::adopt_memory_log(protocol.log_info) };
+    u_assert!(log.is_some(), "adv_logger_test: Failed to adopt the memory log.");
+    let log_info = log.unwrap();
     let mut direct_found = false;
     let mut protocol_found = false;
     for entry in log_info.iter() {
--- a/components/patina_adv_logger/src/memory_log.rs
+++ b/components/patina_adv_logger/src/memory_log.rs
@@ -10,13 +10,18 @@
 //! SPDX-License-Identifier: BSD-2-Clause-Patent
 //!
 use core::{
-    ffi::c_void,
+    cell::UnsafeCell,
     mem::size_of,
     ptr, slice,
     sync::atomic::{AtomicU32, AtomicU64, Ordering},
 };
-use patina_sdk::error::{EfiError, Result};
+use patina_sdk::{
+    base::align_up,
+    error::{EfiError, Result},
+};
 use r_efi::efi;
+use zerocopy::{FromBytes, IntoBytes};
+use zerocopy_derive::*;
 
 // { 0x4d60cfb5, 0xf481, 0x4a98, {0x9c, 0x81, 0xbf, 0xf8, 0x64, 0x60, 0xc4, 0x3e }}
 pub const ADV_LOGGER_HOB_GUID: efi::Guid =
@@ -31,7 +36,10 @@ pub const DEBUG_LEVEL_VERBOSE: u32 = 0x00400000;
 // Phase definitions.
 pub const ADVANCED_LOGGER_PHASE_DXE: u16 = 4;
 
-/// A struct for carrying log entry data through this module.
+/// A struct for carrying log entry both as input and output to this module.
+/// This struct contains the key information for the log entry, but excludes the
+/// log entry specifics that are not needed by generic code.
+#[derive(Clone, Copy)]
 pub struct LogEntry<'a> {
     pub phase: u16,
     pub level: u32,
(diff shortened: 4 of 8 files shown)


components/patina_adv_logger/src/memory_log.rs
```diff
@@ -31,19 +33,243 @@ pub const DEBUG_LEVEL_VERBOSE: u32 = 0x00400000;
...
+            LogData::ReadOnly(slice) => slice,
+            LogData::ReadWrite(cell) => unsafe { &*cell.get() },
+        }
+    }
+
+    /// Returns the data as a mutable slice.
+    fn get_mut(&self) -> Result<*mut [u8]> {
+        match self {
+            LogData::ReadOnly(_) => Err(EfiError::AccessDenied),
+            LogData::ReadWrite(cell) => Ok(cell.get()),
```
os-d: Seems a little odd for `get` to return a reference and `get_mut` to return a pointer, I would expect they return the same type, just one is mut. But not a big deal if it is better this way.

components/patina_adv_logger/src/memory_log.rs
```diff
@@ -31,19 +33,243 @@ pub const DEBUG_LEVEL_VERBOSE: u32 = 0x00400000;
...
+}
+
+// SAFETY: The only interior mutability is the UnsafeCell for the data region of
+//         the log. Safety here is checked by the allocation logic in add_log_entry
+//         which relies on atomics to safely allocate portions of the buffer.
+unsafe impl Send for AdvancedLog<'static> {}
+unsafe impl Sync for AdvancedLog<'static> {}
+
+impl AdvancedLog<'static> {
+    pub unsafe fn adopt_memory_log(address: efi::PhysicalAddress) -> Option<Self> {
```
os-d, line 82: nit: func comments

Cargo.toml
```diff
@@ -52,6 +52,8 @@ uart_16550 = { version = "^0.3.2" }
 uefi_corosensei = { version = "0.1.3", default-features = false, registry = "patina-fw" }
 uuid = { version = "1.8", default-features = false }
 x86_64 = { version = "=0.15.2", default-features = false }
+zerocopy = { version = "0.8" }
+zerocopy-derive = { version = "0.8" }
```
Javagedes, line 57: You don't need to directly use `zerocopy-derive`. Just enable the `derive` feature in zero copy.

components/patina_adv_logger/src/memory_log.rs
```diff
@@ -31,19 +33,243 @@ pub const DEBUG_LEVEL_VERBOSE: u32 = 0x00400000;
...
+        }
+
+        // Check that the various pointers are consistent.
+        let log_current = header.log_current_offset.load(Ordering::Relaxed);
+        if log_current > header.log_buffer_size || header.log_buffer_offset < size_of::<AdvLoggerInfo>() as u32 {
+            return Err(EfiError::InvalidParameter);
+        }
+
+        // Only require that the valid portion of the log buffer be present.
+        if header.log_buffer_offset + log_current > log_bytes.len() as u32 {
```
os-d: It's been awhile since I wrote this, but isn't `log_current_offset` an offset from the start of `AdvLoggerInfo` not from the buffer start?

components/patina_adv_logger/src/memory_log.rs
```diff
@@ -31,19 +33,243 @@ pub const DEBUG_LEVEL_VERBOSE: u32 = 0x00400000;
...
+
+    pub fn add_log_entry(&self, log_entry: LogEntry) -> Result<()> {
+        // Adding a log entry consists of two steps:
+        // 1. Atomically allocate space in the log buffer. This must be done before
+        //    writing the log entry to ensure that no other system can write to the
+        //    same space in the log buffer.
+        // 2. Write the log entry to the allocated space in the log buffer.
+        //
+
+        // Get the total size of the long entry with the header, inclusing the
```
os-d:
```suggestion
        // Get the total size of the long entry with the header, including the
```

components/patina_adv_logger/src/memory_log.rs
```diff
@@ -31,19 +33,243 @@ pub const DEBUG_LEVEL_VERBOSE: u32 = 0x00400000;
...
+        // alignment padding for 8 byte alignment.
+        let data_offset = size_of::<AdvLoggerMessageEntry>() as u16;
+        let unaligned_size = data_offset as u32 + log_entry.data.len() as u32;
+        let message_size = (unaligned_size + 7) & !7;
```
os-d: Use `align_up`?

components/patina_adv_logger/src/memory_log.rs
```diff
@@ -31,19 +33,243 @@ pub const DEBUG_LEVEL_VERBOSE: u32 = 0x00400000;
 // Phase definitions.
 pub const ADVANCED_LOGGER_PHASE_DXE: u16 = 4;
 
-/// A struct for carrying log entry data through this module.
+/// A struct for carrying log entry both as input and output to this module.
+/// This struct contains the key information for the log entry, but excludes the
+/// log entry specifics that are not needed by generic code.
```
os-d, line 41: What is the benefit of having this split off? It seems odd to me to have two definitions for this, one that is correct and one that is wrong and we use the wrong one (yes I'm being over the top for emphasis) until we have to use the right one when we write to memory. Seems like that is an area ripe for errors. Why not just use the full structure?

Cargo.toml
Javagedes: @cfernald closing this. I read the zerocopy documentation and it suggests doing it the way you do it, for compile speed :)

components/patina_adv_logger/src/memory_log.rs
```diff
@@ -31,19 +33,243 @@ pub const DEBUG_LEVEL_VERBOSE: u32 = 0x00400000;
...
+        if length < size_of::<AdvLoggerInfo>() as u32 {
+            return None;
+        }
+
+        let header = address as *mut AdvLoggerInfo;
+        if header.is_null() {
+            None
+        } else {
+            // Create an empty log then use adopt for code re-usage.
+            ptr::write(header, AdvLoggerInfo::new(length, false, 0, 0, efi::Time::default(), 0));
```
vineelko: I think we should make sure the address for header is aligned as per the alignment requirements of `AdvLoggerInfo`. If not, we are stepping into UB as ptr::write requirements are not upheld.

```rust
else if header % std::mem::align_of::<AdvLoggerInfo>() != 0 {
    None or some error.
}
```

components/patina_adv_logger/src/memory_log.rs
```diff
@@ -31,19 +33,243 @@ pub const DEBUG_LEVEL_VERBOSE: u32 = 0x00400000;
...
+}
+
+/// This struct represents an advanced logger memory log. It contains the appropriate
+/// pointers and interior mutability to allow for safe access to the log data. This
+/// serves as the idiomatic rust container for the C based structures.
+pub struct AdvancedLog<'a> {
+    /// The header of the memory log.
+    pub(crate) header: &'a AdvLoggerInfo,
+    /// The data portion of the memory log.
+    data: LogData<'a>,
```
vineelko, line 64: Why do we need this extra indirection of `LogData` enum for data? Since `data` is a private field why not have it as `UnsafeCell<'a>`?

components/patina_adv_logger/src/memory_log.rs
```diff
@@ -31,19 +33,243 @@ pub const DEBUG_LEVEL_VERBOSE: u32 = 0x00400000;
...
+
+        let (header_slice, entry_slice) = entry_slice.split_at_mut(size_of::<AdvLoggerMessageEntry>());
+        let (data_slice, remainder_slice) = entry_slice.split_at_mut(log_entry.data.len());
+
+        AdvLoggerMessageEntry::from_log_entry(&log_entry)
+            .write_to(header_slice)
+            .map_err(|_| EfiError::BufferTooSmall)?;
+
+        log_entry.data.write_to(data_slice).map_err(|_| EfiError::BufferTooSmall)?;
+        remainder_slice.fill(0);
```
vineelko, line 235: NIT: Is this necessary everytime?

components/patina_adv_logger/src/component.rs
```diff
@@ -113,15 +113,14 @@ where
     /// Installs the Advanced Logger Protocol for use by non-local components.
     ///
     fn entry_point(self, bs: StandardBootServices) -> Result<()> {
-        let log_info = match self.adv_logger.get_log_info() {
-            Some(log_info) => log_info,
+        let address = match self.adv_logger.get_log_address() {
```
wesleywiser: nit: This is totally stylistic but I thought you might like to know you can use [`let-else`](https://doc.rust-lang.org/rust-by-example/flow_control/let_else.html) to write this more concisely:

```rust
let Some(address) = self.adv_logger.get_log_address() else {
    log::error!("Advanced logger not initialized before component entry point!");
    return Err(EfiError::NotStarted);
};
```

components/patina_adv_logger/src/memory_log.rs
```diff
@@ -31,19 +33,243 @@ pub const DEBUG_LEVEL_VERBOSE: u32 = 0x00400000;
...
+    /// A string slice representing the log message.
+    ReadWrite(&'a UnsafeCell<[u8]>),
+}
+
+impl<'a> LogData<'a> {
+    /// Returns the data as a slice.
+    fn get(&self) -> &'a [u8] {
+        match self {
+            LogData::ReadOnly(slice) => slice,
+            LogData::ReadWrite(cell) => unsafe { &*cell.get() },
```
wesleywiser: Would be good to have a `// SAFETY` comment here explaining why this is justified especially since this is a foundational data structure.

components/patina_adv_logger/src/memory_log.rs
```diff
@@ -31,19 +33,243 @@ pub const DEBUG_LEVEL_VERBOSE: u32 = 0x00400000;
...
+            self.header.discarded_size.fetch_add(message_size, Ordering::Relaxed);
+            return Err(EfiError::OutOfResources);
+        }
+
+        let data_index = (current_offset - self.header.log_buffer_offset) as usize;
+
+        // SAFETY: The space hase been allocated. It should now be safe to write
+        // data so long as it sticks to the range of the allocated entry. Get only
+        // the allocated slice to maintain safety.
+        let entry_slice = unsafe {
```
wesleywiser, line 222: Has this memory been initialized or is this "uninitialized but about to be written to" memory?

cfernald: The memory is not expected to be initialized in any ways. So yeah, "uninitialized but about to be written to"

wesleywiser: Gotcha. Since this memory originates from "outside" the program, it's not entirely clear to me whether the memory is considered initialized by the optimizer or not. This is important because when you call `get_mut` to return a slice, you're telling the compiler you have a valid _and initialized_ value at this point in the program. Obviously, on the hardware level, there's going to be something in this slice and it has to be a `u8`. 

If you wanted to be very pedantic, the most correct thing to do would be to model this as `&[MaybeUninit<u8>]`. Since you're immediately writing onto the memory and the compiler can't know the memory might be uninitialized to begin with, I _think_ what you have is correct.

components/patina_adv_logger/src/memory_log.rs
cfernald: This is intentional. The purpose is that the safety of the writable case is different as it requires that the caller ensures that they only dereference the section that was appropriate "allocated" using the atomic pointers. I think keeping it as a raw pointer until then makes more sense to me.

components/patina_adv_logger/src/memory_log.rs
cfernald: I was on the fence about this, but writing this way means we don't need to many unnecessary "unsafes" when parsing the code as we can just keep it as a reference the whole time. I'm curious to get other's thoughts though as i'm unsure what I prefer.

@os-d , @wesleywiser , thoughts on this enum wrapper vesus just always using UnsafeCell

components/patina_adv_logger/src/memory_log.rs
cfernald: Correct, at this point log_bytes is the _entire_ memory log including the header.

cfernald: Oh wait, sorry misread that. This spun me through a loop for a second. You are right and its a bit inconsistent. Will confirm its being used correctly elsewhere too.

components/patina_adv_logger/src/memory_log.rs
os-d: tbh I've never written with an `UnsafeCell` and only have seen it in the code a few times. Just read through the documentation for it :). I'm not sure I have a preference. Maybe cleaner as `UnsafeCell`?

cfernald: The real concern is about taking a `&[u8]` and putting into a unsafecell where it _could_ then be written to through other function even though it was only ever meant to be read. This enum seemed to prevent that and had the added benefits of less unsafes.

components/patina_adv_logger/src/memory_log.rs
cfernald: There are two reasons. 

The original reason I had LogEntry is that there is a data associated with each log entry that gets passed from the logger layer to the memory log and having a structure for it is cleaner than having them as loose arguments to all the functions. It makes sense for this structure to not be the memory log structure as constructing that outside of the memory log (or even in it) is problematic as it is very difficult to build. I would argue doing it the other way for the input case would not make sense.

Now in this change I also switch to returning this type from the iterator so the memory log structure is contained to this file and not used in the parser. This is more of a philosophical shift. Before this logic _tried_ to have the C structs also implement the functionality for parsing out the log such that the info header was used to parse the log and the entries were used to query and set entries. I've found this to be too limiting to write correctly and I believe only using the C structs and using consistent rusty types outside of this file feels cleaner.

components/patina_adv_logger/src/memory_log.rs
vineelko: Sorry, I should have been clearer in my comment. I think we still need `get()` and `get_mut()` methods to make usage of the unsafe parts easier. However, looking at the code, this is a private field, so it should be accessible only within this module. So I think we don't have to worry about its incorrect usage. Let me know if I am missing something.

components/patina_adv_logger/src/memory_log.rs
cfernald: I's only zeroing the alignment padding. I suppose we could just leave it as whatever was there before but for the rare change of looking at the binary representation, I feel zeroing these few bytes is not going to break bank here.

components/patina_adv_logger/src/memory_log.rs
cfernald: The problem is that because we have a readwrite path and a readonly path, we have tro track whether or not the buffer should be modified, and to do that I originally went down the path of having a boolean that wrapped all the mutable accesses around a `UnsafeCell` like you suggest, but then at that point it was actually less clear and required some unsafe code to do. I think this, while more verbose, is actually safer and more explicit.

components/patina_adv_logger/src/memory_log.rs
cfernald: What is the implication if the compiler doesn't believe it is initialized for vice versa? I would image in all scenarios the compiler will assume these are initialized an contain _some value_. And my code will try to determine if those values are what we expect. Will the compiler need to know more than that here?

cfernald: @wesleywiser, thinking about this more, doesn't the initialized only matter for types that have `Drop` or similar implemented? Does it matter for a `u8`? I'll close this for now, but please re-open if you think there is an issue.

components/patina_adv_logger/src/memory_log.rs
```diff
@@ -31,19 +36,268 @@ pub const DEBUG_LEVEL_VERBOSE: u32 = 0x00400000;
...
+                || (*log_info).version != AdvLoggerInfo::VERSION
+                || (*log_info).log_buffer_offset < size_of::<AdvLoggerInfo>() as u32
+        } {
+            None
+        } else {
+            // SAFETY: The log_info is valid, convert the data for future safety.
+            unsafe {
+                let header = log_info.as_ref()?;
+                let data_size = header.log_buffer_size;
+                let data_start = (log_info as *const u8).byte_add(header.log_buffer_offset as usize) as *mut u8;
```
os-d: Just because this has been the source of some bugs already, want to confirm: is this okay that we are saying:
```rust
let log_info = address as *const AdvLoggerInfo; // const pointer
let data_start = (log_info as *const u8).byte_add(header.log_buffer_offset as usize) as *mut u8; // const as const as mut
 let data = slice::from_raw_parts_mut(data_start, data_size as usize); // using a mut function
Some(Self { header, data: LogData::ReadWrite(UnsafeCell::from_mut(data)) }) // continuing to rely on it being mut
```
I agree that `log_info` is a const pointer, that address should not change. I also agree the data is mutable. Just want to double confirm that it is okay we are generating a mut pointer from an immutable pointer, because that has caused issues. We probably should be using provenance functions here instead, but I won't make you go learn those in order to submit this PR ;)

components/patina_adv_logger/src/memory_log.rs
```diff
@@ -346,39 +445,67 @@ impl AdvLoggerMessageEntry {
 
     /// Returns the aligned length of the entire log entry.
     pub fn aligned_len(&self) -> usize {
-        (self.len() + 7) & !7
+        align_up(self.len(), 8).unwrap()
```
os-d, line 450: I'm okay to say this will never overflow (so align_up will never fail) and the bare unwrap is okay, but we should at least comment it (or use expect). Or, depending on context here return a safe default.

components/patina_adv_logger/src/memory_log.rs
cfernald: This is a good point, will look into this pointer casting to see if it can be improved

components/patina_adv_logger/src/memory_log.rs
```diff
@@ -31,19 +36,268 @@ pub const DEBUG_LEVEL_VERBOSE: u32 = 0x00400000;
...
+            LogData::ReadOnly(_) => Err(EfiError::AccessDenied),
+            LogData::ReadWrite(cell) => Ok(cell.get()),
+        }
+    }
+}
+
+// SAFETY: The only interior mutability is the UnsafeCell for the data region of
+//         the log. Safety here is checked by the allocation logic in add_log_entry
+//         which relies on atomics to safely allocate portions of the buffer.
+unsafe impl Send for AdvancedLog<'static> {}
```
Javagedes, line 70: nit: Is there a reason this is only implemented when AdvancedLog is initialized statically? It seems like this should be:

`unsafe impl<'a> Send for AdvancedLog<'a> {}`

components/patina_adv_logger/src/memory_log.rs
```diff
@@ -31,19 +36,268 @@ pub const DEBUG_LEVEL_VERBOSE: u32 = 0x00400000;
...
+
+/// Wrapper to allow for a read-only or read-write data region for the log.
+enum LogData<'a> {
+    /// A slice of bytes representing the log message.
+    ReadOnly(&'a [u8]),
+    /// A string slice representing the log message.
+    ReadWrite(&'a UnsafeCell<[u8]>),
+}
+
+impl<'a> LogData<'a> {
```
Javagedes: nit: Not a huge fan that the enum and it's impls are defined in-between AdvancedLog and it's impls.

components/patina_adv_logger/src/memory_log.rs
```diff
@@ -31,19 +36,268 @@ pub const DEBUG_LEVEL_VERBOSE: u32 = 0x00400000;
...
+impl AdvancedLog<'static> {
+    /// Initializes a `AdvancedLog` from an existing advanced logger buffer at the
+    /// provided address. The caller must ensure that this memory is accessible.
+    ///
+    /// ### Safety
+    ///
+    /// This function assumes that the provided address points to a valid `AdvLoggerInfo`
+    /// structure and that the memory is properly sized initialized based on the
+    /// information in that structure.
+    pub unsafe fn adopt_memory_log(address: efi::PhysicalAddress) -> Option<Self> {
```
Javagedes, line 82: Not a huge fan of these functions all continuing to use efi::PhysicalAddress. Do we have a reason to? Seems like we could use `NonNull<AdvLoggerInfo>`

components/patina_adv_logger/src/memory_log.rs
```diff
@@ -31,19 +36,268 @@ pub const DEBUG_LEVEL_VERBOSE: u32 = 0x00400000;
...
+    #[allow(dead_code)]
+    pub fn open_log(log_bytes: &'a [u8]) -> Result<Self> {
+        if log_bytes.len() < size_of::<AdvLoggerInfo>() {
+            return Err(EfiError::BufferTooSmall);
+        }
+
+        // SAFETY: We have checked the length of the buffer is at least the size.
+        //         Ideally we use ZeroCopy to do this, but `efi::Time` does not
+        //         support that.
+        let header =
```
Javagedes, line 146: Not really a big deal, but trying to get away form raw pointers when we know the pointer is non null seems like a good thing to do - `NonNull::from(log_bytes).cast::<AdvLoggerInfo>().as_ref()`

components/patina_adv_logger/src/memory_log.rs
```diff
@@ -346,39 +445,69 @@ impl AdvLoggerMessageEntry {
...
+
+            let entry_header = AdvLoggerMessageEntry::ref_from_bytes(header_slice).ok()?;
+            data_index += size_of::<AdvLoggerMessageEntry>();
+
+            if self.offset + size_of::<AdvLoggerMessageEntry>() + entry_header.message_length as usize
+                > self.log.header.log_current_offset.load(Ordering::Relaxed) as usize
+            {
+                None
+            } else {
+                let entry_data = unsafe {
```
Javagedes, line 499: nit: missing safety comment

components/patina_adv_logger/src/component.rs
```diff
@@ -22,7 +22,7 @@ use r_efi::efi;
 
 use crate::{
     logger::AdvancedLogger,
-    memory_log::{self, AdvLoggerInfo},
+    memory_log::{self},
```
Javagedes: I think this can just be `memory_log` right? not 100% sure

components/patina_adv_logger/src/integration_test.rs
```diff
@@ -47,9 +47,9 @@ fn adv_logger_test(bs: StandardBootServices) -> patina_sdk::test::Result {
     u_assert_eq!(efi_status, efi::Status::SUCCESS, "adv_logger_test: Failed to write to the advanced logger protocol.");
 
     // Check that the strings were added to the log.
-    let log_info = unsafe { memory_log::AdvLoggerInfo::adopt_memory_log(protocol.log_info) };
-    u_assert!(log_info.is_some(), "adv_logger_test: Failed to adopt the memory log.");
-    let log_info = log_info.unwrap();
+    let log = unsafe { memory_log::AdvancedLog::adopt_memory_log(protocol.log_info) };
```
Javagedes, line 52: nit: missing safety comment, even if this is a test :)

components/patina_adv_logger/src/component.rs
cfernald: Yeah, I think rust analyzer does this when I do a `::function` when it should be a `.function` and so it pulls in the `self` field.

components/patina_adv_logger/src/memory_log.rs
wesleywiser: No, unfortunately merely constructing a value whose content is uninitialized is instant UB (try running this under Miri in the playground https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=b319563869d8cdbda6b067f323928ae9).

That being said, I think what you have here is strictly more correct than the previous version so I would certainly not block this from being merged 🙂

components/patina_adv_logger/src/memory_log.rs
cfernald: This is because this should only be required when using the active static advlog, and may not be true for all use-cases such as parsing. Basically, scope the unsafe impl to only the expected scenario.

components/patina_adv_logger/src/memory_log.rs
cfernald: While I agree we shouldn't be using efi::PhysicalAddress, i'm not sure if this is proper use of NonNull. From readin the description (https://doc.rust-lang.org/std/ptr/struct.NonNull.html) it seems like the intention of NonNull is really supposed to be as a means of getting Option<NonNull<T>> so that is affectively has the built is "IsNull" check in that the 0 value is reserved for the `None` case. 

From reading that, it seems we shouldn't really be using pointers more then NonNull for cases where there is no expected "Null" input just as a stand is for a pointer, right? 

In any case, here its a bit weird because the type of the memory is not known to the caller, by design. So I suppose it could be a cvoid or *u8, but i'm not convinced either of these buy us much considering the value comes from a hob and would just have to be converted by the caller in this case.

cfernald: Going to resolve this for now as this change is already tracking a ton of differences. I'm open to making this change, but will follow up with a second change if so.

PR #577, opened 2025-07-12
Update to the 2024 Edition [Rebase & FF]

### LINKED ISSUE #572: [Feature]: Update to 2024 edition
### Feature Overview

Update to the Rust 2024 edition to support some of the latest tools and prepare code for moving to later Rust releases.

### Solution Overview

See above.

### Alternatives Considered

_No response_

### Urgency

Medium

I will implement the feature

No maintainer feedback needed

### Anything else?

_No response_

makubacki:
## Description

Closes #572 

---

**error.rs: Fix match ergonomics**

In the 2024 edition of Rust, binding modifiers may only be written
when the default binding method mode is move. This removes the
redundant explicit `ref` binding.

---

**Fix temporary value dropped while still in use errors**

Assign the value of pointers to a temporary variable to extend its
lifetime to make it clear that the temporary value is not dropped
before the borrow is finished.

---

**Add missing unsafe blocks**

The 2024 edition requires unsafe to be used even within unsafe
functions. This is a large set of changes for that.

---

**Update to 2024 edition**

Updates the edition to 2024 with `fmt` changes with that change.

---

**Add missing unsafe blocks to tests**

Updates test to mark unsafe code for the 2024 edition.

---

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- `cargo make all`

## Integration Instructions

Note: Interoperable with the 2021 edition.

### DIFF
--- a/components/patina_adv_logger/src/component.rs
+++ b/components/patina_adv_logger/src/component.rs
@@ -160,21 +160,23 @@ mod tests {
         let log_address = log_buff as *const u8 as efi::PhysicalAddress;
 
         // initialize the log so it's valid for the hob list
-        AdvLoggerInfo::initialize_memory_log(log_address, LOG_LEN as u32);
+        unsafe { AdvLoggerInfo::initialize_memory_log(log_address, LOG_LEN as u32) };
 
         const HOB_LEN: usize = size_of::<GuidHob>() + size_of::<efi::PhysicalAddress>();
         let hob_buff = Box::into_raw(Box::new([0_u8; HOB_LEN]));
         let hob = hob_buff as *mut GuidHob;
-        ptr::write(
-            hob,
-            GuidHob {
-                header: Hob { r#type: GUID_EXTENSION, length: HOB_LEN as u16, reserved: 0 },
-                name: memory_log::ADV_LOGGER_HOB_GUID,
-            },
-        );
+        unsafe {
+            ptr::write(
+                hob,
+                GuidHob {
+                    header: Hob { r#type: GUID_EXTENSION, length: HOB_LEN as u16, reserved: 0 },
+                    name: memory_log::ADV_LOGGER_HOB_GUID,
+                },
+            )
+        };
 
-        let address: *mut efi::PhysicalAddress = hob.add(1) as *mut efi::PhysicalAddress;
-        (*address) = log_address;
+        let address: *mut efi::PhysicalAddress = unsafe { hob.add(1) } as *mut efi::PhysicalAddress;
+        unsafe { (*address) = log_address };
         hob_buff as *const c_void
     }
 
--- a/core/patina_stacktrace/src/error.rs
+++ b/core/patina_stacktrace/src/error.rs
@@ -91,7 +91,7 @@ impl fmt::Display for Error {
                 )
             }
             #[cfg(test)]
-            Error::ModuleLoadFailed(ref module) => {
+            Error::ModuleLoadFailed(module) => {
                 write!(fmt, "Failed to load module: {}", module.as_ref().unwrap_or(&no_module_str))
             }
         }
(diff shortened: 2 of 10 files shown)


makubacki:
@Javagedes, @os-d, @vineelko, normally this change wouldn't be very high priority, but I've also based some changes to the performance component on top of this branch. The performance changes are needed to essentially unblock Patina DXE Core integration on Intel platforms. More details on that are in:

- [[Feature]: Support dynamic configuration of the `patina_performance` component](https://github.com/OpenDevicePartnership/patina/issues/578)

So, I'd appreciate if we can prioritize getting this out of the way and then I'll post those changes.

components/patina_adv_logger/src/component.rs
```diff
@@ -160,21 +160,23 @@ mod tests {
         let log_address = log_buff as *const u8 as efi::PhysicalAddress;
 
         // initialize the log so it's valid for the hob list
-        AdvLoggerInfo::initialize_memory_log(log_address, LOG_LEN as u32);
+        unsafe { AdvLoggerInfo::initialize_memory_log(log_address, LOG_LEN as u32) };
```
Javagedes, line 163: nit: unsafe needs safety comment

components/patina_adv_logger/src/component.rs
```diff
@@ -160,21 +160,23 @@ mod tests {
...
-            hob,
-            GuidHob {
-                header: Hob { r#type: GUID_EXTENSION, length: HOB_LEN as u16, reserved: 0 },
-                name: memory_log::ADV_LOGGER_HOB_GUID,
-            },
-        );
-
-        let address: *mut efi::PhysicalAddress = hob.add(1) as *mut efi::PhysicalAddress;
-        (*address) = log_address;
+        unsafe {
```
Javagedes, line 168: nit: unsafe needs safety comment

components/patina_adv_logger/src/memory_log.rs
```diff
@@ -121,13 +121,14 @@ impl AdvLoggerInfo {
 
     pub unsafe fn adopt_memory_log(address: efi::PhysicalAddress) -> Option<&'static Self> {
         let log_info = address as *mut Self;
-        if (*log_info).signature != Self::SIGNATURE
-            || (*log_info).version != Self::VERSION
-            || (*log_info).log_buffer_offset < size_of::<AdvLoggerInfo>() as u32
+        let log_info_ref = unsafe { log_info.as_ref() }?;
```
Javagedes, line 124: nit: unsafe needs safety comment

makubacki:
> These all technically need safety requirement comments, either saying that the safety is met due to the safety requirements of this function, or due to some other reason. Not sure if we want to do that in this PR or not.

Makes sense. I'm happy to do that too. But I'd want a little more time to look at each usage and write more meaningful comments, so I deferred that as tracked in https://github.com/OpenDevicePartnership/patina/issues/583.

components/patina_adv_logger/src/component.rs
makubacki: Deferred. Tracked in https://github.com/OpenDevicePartnership/patina/issues/583.

components/patina_adv_logger/src/memory_log.rs
makubacki: Deferred. Tracked in https://github.com/OpenDevicePartnership/patina/issues/583.

components/patina_adv_logger/src/component.rs
makubacki: Deferred. Tracked in https://github.com/OpenDevicePartnership/patina/issues/583.

PR #582, opened 2025-07-14
RFC: Patina Package [FCP]

makubacki:
## Description

Proposes a new ODP repo and Patina edk2 style package to host new definitions authored by Patina that need to be used by edk2 style C code.

**Status: FCP**

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [x] Includes documentation?

## How This Was Tested

- N/A

## Integration Instructions

- N/A


docs/src/rfc/text/0000-patina-package.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Change Log
+
+- 2025-07-14: Initial draft of RFC.
+
+## Motivation
+
+As the Patina project grows, it is inevitable that it needs to define content that it is made available to interact
+with edk2 C code today, such as GUID definitions for HOBs that are exclusively defined by Patina components. There is
+not a clear common package to put this content in today - Patina is an independent project of Project Mu and edk2.
```
os-d:
```suggestion
not a clear common package to put this content in today - Patina is a project independent of Project Mu and edk2.

```

The original wording sounded a bit like it was a joint venture between Mu and edk2

docs/src/rfc/text/0000-patina-package.md
```diff
@@ -0,0 +1,67 @@
...
+2. **Proper Scope** - The `patina-edk2` repository should only contain Patina definitions that are intended to be
+   consumed by edk2 C code.
+   - This includes Patina-specific HOBs, GUIDs, and other content that is not intended to be used by Rust code.
+   - The repository should not contain any Rust code or Patina crates.
+3. **Proper Classification** - The Patina repository should be rarely used within Patina work. Patina is not meant to
+   establish a large set of C-specific content that is not already defined in specifications and made available
+   through existing means. However, Patina will define some new content that is not defined in specifications and may
+   be made available for use in edk2 C code. In that case, this repository is the proper place to put a single
+   officially maintained copy of that content.
+4. **Minimal Dependencies** - The repository is not allowed to depend on any content other than `MdePkg`, `MdeModulePkg`,
```
os-d: Unfortunately ArmPkg will probably need to be part of this list, also (as the UefiCpuPkg equivalent) but I'm fine to leave it out until we need it.

docs/src/rfc/text/0000-patina-package.md
```diff
@@ -0,0 +1,67 @@
...
+   and `UefiCpuPkg` from upstream `edk2`. Any other dependency to an `edk2` package must be requested via RFC and
+   Project Mu dependencies are completely disallowed.
+5. **No Dependencies From Patina** - No other Patina repository can depend on content in this repository. No other
+   Patina repository documentation should reference details in this repository. The repository should be able to be
+   removed from the Patina project at any time without impacting any other Patina repository.
+
+## Unresolved Questions
+
+- None
```
os-d: One open question I would have is do we have any guidance or requirements for how content in this repo is generated? i.e. must be generated from cbindgen from a source in a Patina repository (presumably linking to the patina definition)? I'm not necessarily advocating for that, but it is a question to me on how rigorous we want this to be vs as long as hand writing it goes okay, then its fine.

makubacki: It's a good question and I've added it to this section so it's on the record. That seems to imply the content generation is vast enough to justify an automated process to facilitate its generation and I'm really hoping we don't need that much here.

docs/src/rfc/text/0000-patina-package.md
makubacki: I considered that, but its stability, maturity, and conceptual scope is not quite the same as the other packages. These form what was once called the "Green H" of dependencies fundamental to firmware construction. For example, I would have been more hesitant to put "IntelPkg" than "UefiCpuPkg". In reality, I agree we should wait until we need to make that decision.

docs/src/rfc/text/0000-patina-package.md
makubacki: @os-d, I'll let you close this conversation if you're okay with that outcome for now.

docs/src/rfc/text/0000-patina-package.md
joschock: Why is UefiCpuPkg needed?

docs/src/rfc/text/0000-patina-package.md
os-d: Yep, I just wanted to note it, I share the same hope as you, that this package is minimal.

docs/src/rfc/text/0000-patina-package.md
makubacki: It's just proactive to say that's allowed. I'd be fine to drop it though.

joschock: my lean would be to omit it, since including it basically authorizes it to be part of the patina "c" interface. Including it here basically means stuff like GUIDs and structs from UefiCpuPkg are legit to include in e.g. Patina Component APIs.

makubacki: Removed it.

makubacki:
Note: Moved the RFC status to FCP.

PR #584, opened 2025-07-14
[REBASE & FF] Debugger:  Remove the global monitor buffer

cfernald:
## Description

With the newest Windbgx (2507+), the configuration for UEFI has been updated such
that the GDB extension will wait until a monitor command is completed with
an OK packet to signal that all responses have been sent. This means that buffering
the entire response into a single buffer is no longer required and the message
may be sent over multiple GDB packets. So the monitor buffer that is pre-allocated
is no longer required.

Additionally, this change introduces a `MonitorBuffer` structure that will be inserted
on the stack. This temporary stack is used to reduce the number of packets that will
be sent. Format commands in rust will result in a large number of small writes as it
formats the various types of potentially character by character. This new stack buffer
serves as an optimization to reduce those packets.

The second change in this PR is to remove the use of an allocated formatted string for 
the GCD monitor command.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on Q35

## Integration Instructions

N/A

### DIFF
--- a/core/patina_debugger/src/debugger.rs
+++ b/core/patina_debugger/src/debugger.rs
@@ -225,7 +225,19 @@ impl<T: SerialIO> PatinaDebugger<T> {
         while !target.is_resumed() {
             gdb = match gdb {
                 GdbStubStateMachine::Idle(mut gdb) => {
-                    let byte = gdb.borrow_conn().read().unwrap();
+                    let byte = loop {
+                        match gdb.borrow_conn().read() {
+                            Ok(0x0) => {
+                                log::error!(
+                                    "Debugger: Read 0x00 from the transport. This is unexpected and will be ignored."
+                                );
+                                continue;
+                            }
+                            Ok(b) => break b,
+                            Err(_) => return Err(DebugError::TransportFailure),
+                        }
+                    };
+
                     match gdb.incoming_data(&mut target, byte) {
                         Ok(gdb) => gdb,
                         Err(e) => return Err(DebugError::GdbStubError(e)),
--- a/core/patina_debugger/src/lib.rs
+++ b/core/patina_debugger/src/lib.rs
@@ -165,6 +165,8 @@ enum DebugError {
     GdbStubError(gdbstub::stub::GdbStubError<(), patina_sdk::error::EfiError>),
     /// Failure to reboot the system.
     RebootFailure,
+    /// Failure in the transport layer.
+    TransportFailure,
 }
 
 /// Policy for how the debugger will handle logging on the system.


core/patina_debugger/src/dbg_target/monitor.rs
```diff
@@ -35,125 +35,157 @@ Mod commands:
...
-                self.monitor_buffer.set_start_offset(offset);
-                tokens.next()
-            }
-            other => other,
-        };
 
-        match cmd {
+        // Wrap the output in a buffer to reduce the number of packets sent. Without
+        // this formated string may send a packet for each character.
+        let mut out = MonitorBuffer::<'_, 128>::new(out);
```
vineelko: >out [](http://example.com/codeflow?start=16&length=3)

NIT NIT: Maybe calling this `buf` instead of `out` makes it slightly clearer when reading the following code

core/patina_debugger/src/dbg_target/monitor.rs
```diff
@@ -35,125 +35,157 @@ Mod commands:
...
-                    let _ = self
-                        .monitor_buffer
-                        .write_str("ERROR: Failed to acquire system state lock for monitor callbacks!");
+                    let _ = out.write_str("ERROR: Failed to acquire system state lock for monitor callbacks!");
                 }
             },
         }
 
-        self.monitor_buffer.flush_to_console(&mut out);
+        out.flush();
```
vineelko: >out.flush(); [](http://example.com/codeflow?start=8&length=12)

NIT: Looks like `flush` is good candidate for `Drop` trait.

makubacki:
> With the newest Windbgx, the configuration for UEFI has been updated such
that the GDB extension will wait until a monitor command is completed with
an OK packet to signal that all responses have been sent. This means that buffering
the entire response into a single buffer is no longer required and the message
may be sent over multiple GDB packets. So the monitor buffer that is pre-allocated
is no longer required.

Should the Windbgx version be documented with these changes?

core/patina_debugger/src/arch.rs
```diff
@@ -73,7 +73,7 @@ pub trait DebuggerArch {
     fn get_page_table() -> Result<Self::PageTable, ()>;
```
Javagedes, line 73: Can you update some documentation somewhere to specify that the minimum supported version of Windbg is 2507?

cfernald: Right now the use of windbg is really not documented anywhere in Patina. I have a TODO to add this to the book, so will make sure it's documented there when I do.

core/patina_debugger/src/debugger.rs
```diff
@@ -239,7 +225,19 @@ impl<T: SerialIO> PatinaDebugger<T> {
         while !target.is_resumed() {
             gdb = match gdb {
                 GdbStubStateMachine::Idle(mut gdb) => {
-                    let byte = gdb.borrow_conn().read().unwrap();
+                    let byte = loop {
```
Javagedes, line 228: If you required `IntoIterator<Item = u8>` for `ConnectionExt`, you could use `.position(|&b| b != 0x0)`. Just theoretical. Not sure if it's worth the trouble.

cfernald: Because this is a hardware port, I think I'd rather each read be explicit rather than being implicit in some more data-oriented function.

core/patina_debugger/src/arch.rs
cfernald: https://github.com/OpenDevicePartnership/patina/issues/662

PR #585, opened 2025-07-14
patina_performance: Refactor and clean up [Rebase & FF]

makubacki:
## Description

A "wave" of cleanup and reorganization of code in `patina_performance`. This does not solve all debt in the code and will have future iterations of changes. It is the most essential refactoring needed to minimize functional disruption, better support current code organization guidelines, and improve maintainability of the code for upcoming future changes.

---

**Move content from `patina_performance` to `patina_sdk`**

A lot of content in the `patina_performance` component is not solely
within the scope of the component. It makes understanding and
maintaining the component more difficult and places like
`patina_dxe_core` have dependencies on the component module to access
content that should be in a more broadly accessible area.

This change moves content out of `patina_performance` to resolve these
issues and support an upcoming refactor of the component.

Note: This also makes some additional clean up and changes such as
restructuring the namespace, increasing documentation, and some
minor code changees. However, this is not meant to resolve all of
the work that should be done to refactor the performance code. For
example, the code depends on atomic globals for maintaining some
state. An alternative solution would be cleaner. This change simply
collects those static globals into a `globals` module so they are
clearly visible as such.

---

**Refactor patina_performance per crate organization guidelines**

Updates the content of `patina_performance` to follow the most recent
organization format as defined in RFC 9.

https://github.com/OpenDevicePartnership/patina/blob/main/docs/src/rfc/text/0009-standardize-component-crate.md#standard-crate-layout

---

- [x] Impacts functionality?
- [ ] Impacts security?
- [x] Breaking change?
- [x] Includes tests?
- [x] Includes documentation?

## How This Was Tested

- `cargo make all`
- Boot with `BLD_*_TRACE_ENABLE=TRUE` in the `patina-qemu` build with the `Performance` component present in the Q35 Patina DXE Core binary

## Integration Instructions

Review updated documentation and changes in content locations to integrate these changes. At this time, it is expected that a very small population of platforms are using the performance code, so, in practice, the overall impact should be relatively minimal.


makubacki:
@Javagedes, @vineelko, @berlin-with0ut-return, @magravel, I know this is a rather large change, but I have another change coming on top of it to unblock Patina DXE Core integration on Intel platforms. Please try to prioritize review if you can. Thanks!

vineelko:
Reviewed it. LGTM.

components/patina_performance/src/config.rs
```diff
@@ -0,0 +1,47 @@
...
+//! Copyright (C) Microsoft Corporation.
+//!
+//! SPDX-License-Identifier: BSD-2-Clause-Patent
+//!
+
+use patina_sdk::performance::Measurement;
+
+/// A wrapper to generate a mask of all enabled measurements.
+#[derive(Debug, Default)]
+pub struct EnabledMeasurement(pub &'static [Measurement]);
```
Javagedes, line 40: I would add a `all()` function to this that returns a static list of everything.

Javagedes:
```rust
impl EnabledMeasurement {
    const fn all() -< &'static [Measurement] {
        &[
            patina_sdk::performance::Measurement::DriverBindingStart,
            patina_sdk::performance::Measurement::DriverBindingStop,
            patina_sdk::performance::Measurement::DriverBindingSupport,
            patina_sdk::performance::Measurement::LoadImage,
            patina_sdk::performance::Measurement::StartImage,
        ]
    }
}
```

makubacki: Done. I also added unit tests for `EnabledMeasurement`.

PR #586, opened 2025-07-14
[REBASE&FF] Add component HOB consumption example

Javagedes:
## Description

Adds an example of how to consume HOBs in a component.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

CI

## Integration Instructions

N/A

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -52,7 +52,7 @@ Example:
 alias = "build-std"
 
 [tasks.build-std]
-description = """Builds crates in the repository with the standard library.
+description = """Builds crates in the repository with the standard library, and their examples.
 
 Customizations:
     -p [development|release]: Builds in debug or release. Default: development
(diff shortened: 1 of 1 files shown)


sdk/patina_sdk/examples/consume_hob.rs
makubacki, line 1: Since this is an "intro" example, can you please explain at the top of the file that this only demonstrates HOB consumption as HOB production is not allowed at all in the Patina DXE Core? That's obvious to us but may quickly address someone wondering why that's not shown.

Ignore, if that's already done and I missed it.

makubacki, line 1: Does this address anywhere what to do if the HOB is not a GUID HOB? I may have missed it, but that will be a valid scenario under "HOB consumption".

Javagedes: Only GUIDed HOBs are made available to a component. Do we want to support providing non-guided HOBs to consumers? If so, I can create a task and look into it.

For now, however, I've updated the example to specifically say we only support dependency injection for GUIDed HOBs.

Javagedes: Updated to specifically state that HOB production is not supported, and that it is expected that all HOBs are retrieved via the HOB list passed from a pre-DXE firmware.

makubacki: I was just looking for the support to be described here. Imo, we can wait until a use case comes along that justifies more extensive HOB support.

makubacki:
Can you please point to this example somewhere in the component documentation?

Javagedes:
> Can you please somewhere in the component documentation to point to this example?

Added documentation to point back to the examples. Thanks!

sdk/patina_sdk/Cargo.toml
```diff
@@ -16,6 +16,9 @@ harness = false
 name = "bench_service"
 harness = false
 
+[[example]]
+name = "basic_hob_usage"
```
cfernald, line 20: Will these be built in our current CI?

Javagedes: No. But I just updated the `build-std` command to also build the examples :)

PR #589, opened 2025-07-14
Add dynamic configuration provider for Performance component [Rebase & FF]

### LINKED ISSUE #578: [Feature]: Support dynamic configuration of the `patina_performance` component
### Feature Overview

Right now, the `patina_performance` component is configurable via a policy in the bin .rs file:

```rust

// ...

Core::default()
 // ...
 .with_config(patina_performance::EnabledMeasurement(&[
        patina_performance::Measurement::DriverBindingStart, // Adds driver binding start measurements.
        patina_performance::Measurement::DriverBindingStop, // Adds driver binding stop measurements.
        patina_performance::Measurement::DriverBindingSupport, // Adds driver binding support measurements.
        patina_performance::Measurement::LoadImage, // Adds load image measurements.
        patina_performance::Measurement::StartImage, // Adds start image measurements.
    ]))
 .with_component(patina_performance::Performance)
 .start()
 .unwrap();

// ...
```

The problem is that (as noted in the integration instructions of https://github.com/OpenDevicePartnership/patina/pull/405) that if a platform is not built for performance the component should not be addeed (or an asert will happen in FirmwarePerformanceDxe) and if the platform is built for performance then it should be included.

Our Patina DXE Core model provides a single pre-built binary to platforms. Therefore, performance needs to be configurable dynamically based on platform needs. This will allow a platform to consume a single Patina DXE Core and when `BLD_*_PERF_TRACE_ENABLE=TRUE`, they can dynamically configure the Patina Performance component to execute.

### Solution Overview

Add a new component called `PerformanceConfigurationProvider` with the following entry point signature:

```rust
    pub fn entry_point(
        self,
        perf_config_hob: Hob<PerformanceConfigHob>,
        mut config_mut: ConfigMut<config::PerfConfig>,
    ) -> patina_sdk::error::Result<()>
```

This component will only dispatch if a newly introduced HOB is produced (`PerformanceConfigHob`) and modify the performance config to reflect the HOB values. The component is essentially a gasket between the HOB and config that is purely consumed by `Performance`.

### Alternatives Considered

_No response_

### Urgency

Low

I will implement the feature

No maintainer feedback needed

### Anything else?

_No response_

makubacki:
## Description

Closes #578 

---

**patina_sdk: Add bitwise OR support to `Measurement`**

To ease performing bitwise operations on `Measurement` values add
support for various bitwise OR combinations.

---

**components/patina_performance: Add config provider component**

Add a new component called `performance_config_provider` that a
platform can optionally use to convert configuration HOB information
to a configuration consumed by the `patina_performance` component.

Prior to this, the `patina_performance` component only supported
static configuration when invoking the Patina DXE Core `start()`
call. This is a problem because the performance component needs to
be included in the Patina DXE Core binary depending on whether the
platform has enabled performance measurements. So, the reality is
that a single Patina DXE Core binary needs a dynamic interface to
determine whether to activate the main Patina Performance component.

> Platform Integration Detail: Most platforms control performance
> enabling via a PCD. If the platform opts into using the
> `performance_config_provider` component, it should produce the config
> GUID HOB (`{fd87f2d8-112d-4640-9c00-d37d2a1fb75d}`) in PEI with the
> HOB contents set according to platform PCD values.

---

- [x] Impacts functionality?
- [ ] Impacts security?
- [x] Breaking change?
- [ ] Includes tests?
- [x] Includes documentation?

## How This Was Tested

- `cargo make all`
- Boot with performance dynamically enabled and disabled in QEMU Patina FW

## Integration Instructions

- Review the instructions for changing how the `Performance` component is configured
- Review whether `performance_config_provider` is necessary in your Patina DXE Core build and how to integrate it via the mardown file.

### DIFF
--- a/components/patina_performance/src/component.rs
+++ b/components/patina_performance/src/component.rs
@@ -8,6 +8,7 @@
 //!
 
 pub mod performance;
+pub mod performance_config_provider;
 
 // Re-export the Performance component for easier access.
 pub use performance::Performance;
--- a/components/patina_performance/src/component/performance.rs
+++ b/components/patina_performance/src/component/performance.rs
@@ -44,13 +44,18 @@ impl Performance {
     #[cfg(not(tarpaulin_include))] // This is tested via the generic version, see _entry_point.
     pub fn entry_point(
         self,
-        enabled_measurements: Config<config::EnabledMeasurement>,
+        config: Config<config::PerfConfig>,
         boot_services: StandardBootServices,
         runtime_services: StandardRuntimeServices,
         records_buffers_hobs: Option<Hob<HobPerformanceData>>,
         mm_comm_region_hobs: Option<Hob<MmCommRegion>>,
     ) -> Result<(), EfiError> {
-        set_perf_measurement_mask(enabled_measurements.mask());
+        if !config.enable_component {
+            log::warn!("Patina Performance Component is not enabled, skipping entry point.");
+            return Ok(());
+        }
+
+        set_perf_measurement_mask(config.enabled_measurements);
 
         let fbpt = set_static_state(StandardBootServices::clone(&boot_services))
             .expect("Static state should only be initialized here!");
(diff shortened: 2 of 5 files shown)


makubacki:
@cfernald, @berlin-with0ut-return, @magravel, @vineelko, can I please get a review here when you get a chance? This is needed to unblock Intel Patina integration.

PR #590, opened 2025-07-16
patina_dxe_core: Enable Null Pointer Detection

os-d:
## Description

This commit disallows allocating page 0 in order to leave it unmapped to detect null pointers dereferencing. This also ensured that when EFI_MEMORY_RP is applied via set_memory_space_attributes, that Patina DXE Core will unmap the region, not simply attempt to apply the attribute.

While changing the allocation code, this also changes the parameter name of alignment -> align_shift, except in PI spec defined APIs, as that better reflects the usage of that parameter.

- [x] Impacts functionality?
- [x] Impacts security?
- [x] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested with unit tests, booting Q35 and SBSA to Windows/Linux, and breaking in late in boot and confirming page 0 is unmapped.

## Integration Instructions

Page 0 cannot be allocated at address or in pre-DXE.

### DIFF
--- a/patina_dxe_core/src/allocator.rs
+++ b/patina_dxe_core/src/allocator.rs
@@ -831,6 +831,19 @@ fn process_hob_allocations(hob_list: &HobList) {
                     continue;
                 }
 
+                if desc.memory_length == 0 {
+                    log::warn!("Memory Allocation HOB has a 0 length, ignoring.\n{:#x?}", hob);
+                    continue;
+                }
+
+                if desc.memory_base_address == 0 {
+                    log::warn!(
+                        "Memory Allocation HOB has a 0 base address, ignoring. Page 0 cannot be allocated:\n{:#x?}",
+                        hob
+                    );
+                    continue;
+                }
+
                 //Use allocate_pages here to record these allocations and keep the allocator stats up to date.
                 //Note: PI spec 1.8 III-5.4.1.1 stipulates that memory allocations must have page-granularity,
                 //which allows us to use allocate_pages. Check and warn if an allocation doesn't meet the alignment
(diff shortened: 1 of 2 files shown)


patina_dxe_core/src/allocator.rs
```diff
@@ -831,6 +831,11 @@ fn process_hob_allocations(hob_list: &HobList) {
                     continue;
                 }
 
+                if desc.memory_base_address == 0 || desc.memory_length == 0 {
+                    log::warn!("Memory Allocation HOB has a 0 base address or length, ignoring. Page 0 cannot be allocated:\n{:#x?}", hob);
```
makubacki: nit: `desc.memory.length == 0` alone doesn't necessarily mean that "Page 0" was requested (the warning message could be confusing).

patina_dxe_core/src/gcd/spin_locked_gcd.rs
```diff
@@ -759,18 +789,23 @@ impl GCD {
         log::trace!(target: "allocations", "[{}]   Address: {:#x}", function!(), address);
         log::trace!(target: "allocations", "[{}]   Length: {:#x}", function!(), len);
         log::trace!(target: "allocations", "[{}]   Memory Type: {:?}", function!(), memory_type);
-        log::trace!(target: "allocations", "[{}]   Alignment: {:#x}", function!(), alignment);
+        log::trace!(target: "allocations", "[{}]   Align Shift: {:#x}", function!(), align_shift);
         log::trace!(target: "allocations", "[{}]   Image Handle: {:#x?}", function!(), image_handle);
         log::trace!(target: "allocations", "[{}]   Device Handle: {:#x?}\n", function!(), device_handle.unwrap_or(ptr::null_mut()));
 
+        if address == 0 {
```
makubacki: Is this technically a breaking change?

os-d: I debated whether this was a breaking change or not. I'm fine to mark it as such, in one sense, yes, any allocate at address for page 0 is now going to be blocked. On the other hand, this wouldn't have worked in Mu or edk2 (with proper PCD setting) so I don't think any consumers would have been doing this. That being said, it is low cost to do breaking changes in Patina now, I'll mark it as such.

cfernald: Is this still showing as conventional memory in the memory map? If so I wonder if the more appropriate solution would be to mark the page allocate and then unmap. This would prevent confusion for applications like Windows that will try to "intelligently" allocate though memory map information.

os-d: Yeah, that's fair. I didn't want pre-DXE to allocate this, because then patina can't distinguish whether this is a real memory allocation or just reserving it. But I think you are right that patina then should mark it as allocated. I'll update this.

PR #592, opened 2025-07-16
[FCP] RFC: UEFI Services Component

makubacki:
## Description

RFC for a new component that will abstract "UEFI Services" within the component model.

**Status: FCP**

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [x] Includes documentation?

## How This Was Tested

- `cargo make all`

## Integration Instructions

- N/A

### DIFF
--- a/docs/src/rfc/text/0017-uefi-services-component.md
+++ b/docs/src/rfc/text/0017-uefi-services-component.md
@@ -23,6 +23,7 @@ here, but are not in scope for this RFC.
   can be used as injected dependencies in a component.
 - 2025-10-1: Update the section about Patina component and C driver dispatch to reflect recent changes that interleave
   component and C driver dispatch.
+-2025-10-1: Add mocking note to service interfaces section.
 
 ## Motivation
 
@@ -306,6 +307,9 @@ impl MyComponent {
     }
 }
 ```
+
+> Note: Mocking support via [`mockall`](https://docs.rs/mockall/latest/mockall/index.html) will be implemented on
+> service interfaces.
 
 ## Guide-Level Explanation
 


makubacki:
Note: This is currently marked draft to get high-level feedback on the proposal. I'll move it out of draft after filling out the last two sections.

0000-uefi-services-component.md
```diff
@@ -0,0 +1,165 @@
...
+
+## Technology Background
+
+Patina components (and constituent elements such as "services) are primarily described in
+[Monolithically Compiled Components](https://github.com/OpenDevicePartnership/patina/blob/main/docs/src/component/interface.md).
+
+## Goals
+
+1. Treat UEFI Specification defined services as "component services".
+2. Introduce an Patina component service abstraction layer for interface flexibility and ergonomic component usage.
```
os-d:
```suggestion
2. Introduce a Patina component service abstraction layer for interface flexibility and ergonomic component usage.

```

docs/src/rfc/text/0017-uefi-services-component.md
```diff
@@ -0,0 +1,165 @@
...
+1. Treat UEFI Specification defined services as "component services".
+2. Introduce an Patina component service abstraction layer for interface flexibility and ergonomic component usage.
+   > Note: The abstraction layer also presents an opportunity to instrument telemetry into individual service function
+   usage such as tracking how often signal_event is called and from which component the call originated.
+3. Make services in UEFI Specification defined service tables more granular to:
+   1. Participate more precisely in the component dependency graph.
+   2. Better track with auditing and dependency analysis tools. For example, counting how many component depend on
+      event services.
+
+## Requirements
```
os-d, line 81: Looks like you also missed this section

makubacki: Thanks. I added it.

0000-uefi-services-component.md
```diff
@@ -0,0 +1,159 @@
...
+2. All Boot Services and Runtime Services must be accounted for in the `patina_uefi_services` component.
+3. The `patina_uefi_services` component must not provide any service outside of those within Boot Services and Runtime
+   Services (at this time). In the future, it may be allowable to include other APIs defined in the UEFI Specification
+   as services.
+4. Patina components must use the `patina_uefi_services` component to access any "UEFI Service".
+
+## Unresolved Questions
+
+- Where exactly to draw lines between services.
+- Whether to produce various services from a single component (A) or multiple components (B).
```
os-d: As long as it is easy to disentangle services in the future (say we build a distinct event service for Patina, whatever) and be able to move them to their own component, that having one component produce all of the services makes sense to me.

I will note that this mentions MemoryService, which is already an independent service, I would assume that if anything is not covered there, that we would add it to that service instead of putting something here.

makubacki: > I will note that this mentions MemoryService, which is already an independent service, I would assume that if anything is not covered there, that we would add it to that service instead of putting something here.

Correct. I will update this RFC to meet that expectation (such as updating the requirements) and provide a reference to [RFC: Memory Management Interface](https://github.com/OpenDevicePartnership/patina/blob/55fcb7704b6917d7ccb9744dd5bedeaa261af5c4/docs/src/rfc/text/0002-memory-management-interface.md).

makubacki: Updated.

makubacki:
Note: I've started updating this RFC again.

makubacki:
> This looks good to me. I think this implementation will further necessitate the combined Rust component C driver dispatcher as we've been discussing.
> 
> Some of these services (RuntimeVariableServices, ConsoleServices), will rely on protocols being produced. At the moment, we can create a protocol installation event callback to produce the service, however once this callback executes, we are presumedly already at C driver dispatch and no further Rust components will be dispatched.
> 
> If you agree, I can focus on getting this merged execution out. It's a fairly simple change.

Yes, I ended up at the same conclusion. We'll need the ability for the services to come online as the C-backing drivers are available.

makubacki:
@Javagedes & @os-d, I've moved this to FCP. Can you please revisit?

0000-uefi-services-component.md
```diff
@@ -0,0 +1,376 @@
...
+5. Support earlier dispatch of components. Today, the service table pointers are stored in component storage:
+
+   ```rust
+        unsafe {
+            self.storage.set_boot_services(StandardBootServices::new(&*boot_services_ptr));
+            self.storage.set_runtime_services(StandardRuntimeServices::new(&*runtime_services_ptr));
+        }
+   ```
+
+   Components are also dispatched prior to C drivers. However, some subsets of services such as variable services in
```
Javagedes: This needs slightly reworded since #762 was merged that somewhat combines dispatch.

makubacki: Updated

PR #600, opened 2025-07-18
Update arm-gic to 0.5.0

joschock:
## Description

Updates patina AARCH64 GIC support to use 0.5.0 of arm-gic crate. There have been significant API changes since arm-gic crate was first integrated. This PR brings patina up to the latest API. In the course of integrating this, found and filed the following bug to upstream (thanks, @kuqin12 ): [https://github.com/google/arm-gic/issues/67](https://github.com/google/arm-gic/issues/67). This issue is not fatal to operation but will cause error messages on systems with GIC RAS features enabled. 

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Built and tested boot on an AARCH64 hardware platform. 

## Integration Instructions

N/A

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,8 @@ publish = ["patina-fw"]
 
 [workspace.dependencies]
 alloc-no-stdlib = { version = "~2.0" }
-arm-gic = { version = "^0.1.2" }
+arm-gic = { version = "0.5" }
+safe-mmio = { version = "0.2.5" }
 bitfield-struct = { version = "0.10" }
 brotli-decompressor = { version = "4.0.0", default-features = false }
 cfg-if = { version = "1" }
--- a/core/patina_internal_cpu/src/interrupts/aarch64/gic_manager.rs
+++ b/core/patina_internal_cpu/src/interrupts/aarch64/gic_manager.rs
@@ -1,7 +1,10 @@
-use arm_gic::gicv3::{GicV3, IntId, Trigger, registers::GICD};
-use core::ptr::{addr_of, addr_of_mut, write_volatile};
+use arm_gic::{
+    IntId, Trigger,
+    gicv3::{GicV3, InterruptGroup},
+};
 use patina_sdk::error::EfiError;
 use r_efi::efi;
+use safe_mmio::field;
 
 use crate::interrupts::aarch64::sysreg::{read_sysreg, write_sysreg};
 
@@ -68,19 +71,6 @@ pub fn get_system_gic_version() -> GicVersion {
     GicVersion::ArmGicV2
 }
 
-/// Get the maximum interrupt number supported by the GIC.
-///
-/// # Safety
-///
-/// This function reads the GICD_TYPER register, which is set during core
-/// initialization and is not expected to change during runtime.
-///
-pub unsafe fn get_max_interrupt_number(gicd: *mut GICD) -> u32 {
-    let max_num = unsafe { addr_of!((*gicd).typer).read_volatile() & 0x1f };
-
-    if max_num == 0x1f { 1020 } else { (max_num + 1) * 32 }
-}
-
 pub fn get_mpidr() -> u64 {
     unsafe { read_sysreg!(mpidr_el1) }
 }
@@ -99,7 +89,7 @@ pub fn set_binary_point_reg(value: u64) {
 /// initialized during core initialization and not expected to change during
 /// runtime.
 ///
-pub unsafe fn gic_initialize(gicd_base: *mut u64, gicr_base: *mut u64) -> Result<GicV3, EfiError> {
+pub unsafe fn gic_initialize<'a>(gicd_base: *mut u64, gicr_base: *mut u64) -> Result<GicV3<'a>, EfiError> {
     let gic_v = get_system_gic_version();
     if gic_v == GicVersion::ArmGicV2 {
         debug_assert!(false, "GICv2 is not supported");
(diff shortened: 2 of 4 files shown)


sdk/patina_sdk/src/guid.rs
```diff
@@ -90,13 +90,13 @@ pub const HARDWARE_INTERRUPT_PROTOCOL: efi::Guid =
 /// This protocol provides a means of registering and unregistering interrupt handlers for AARCH64 systems.
 /// This protocol extends the Hardware Interrupt Protocol to support interrupt type query.
 ///
-/// (`32898322-2D1A-474A-BAAA-F3F7CF569470`)
+/// (`32898322-2DA1-474A-BAAA-F3F7CF569470`)
```
kuqin12: I thought this is something Chris has fixed?

https://github.com/OpenDevicePartnership/patina/pull/591

joschock: ah, missed that. rebased to main and picked it up. Will fix description too.

Cargo.toml
```diff
@@ -13,7 +13,8 @@ publish = ["patina-fw"]
 
 [workspace.dependencies]
 alloc-no-stdlib = { version = "~2.0" }
-arm-gic = { version = "^0.1.2" }
+arm-gic = { version = "^0.5.0" }
```
os-d: nit: While here, drop `^` to follow recommended Rust dependency syntax: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#caret-requirements.

Also, should we only specify `0.5` to pick up bugfixes automatically? Or do we think there is enough churn that we should lock onto a single version? Same question on safe-mmio.

core/patina_internal_cpu/src/interrupts/aarch64/gic_manager.rs
```diff
@@ -110,135 +100,107 @@ pub unsafe fn gic_initialize(gicd_base: *mut u64, gicr_base: *mut u64) -> Result
...
+            return efi::Status::INVALID_PARAMETER;
         };
-        self.gic_v3.enable_interrupt(int_id, true);
+
+        self.gic_v3.enable_interrupt(int_id, Some(0), true);
+
         efi::Status::SUCCESS
     }
 
     pub fn disable_interrupt_source(&mut self, interrupt_source: u64) -> efi::Status {
```
os-d, line 147: While we are here, do all of these need to be `pub` and do they need to return `efi::Status` instead of `EfiError`?

core/patina_internal_cpu/src/interrupts/aarch64/gic_manager.rs
```diff
@@ -110,135 +100,107 @@ pub unsafe fn gic_initialize(gicd_base: *mut u64, gicr_base: *mut u64) -> Result
...
-                addr_of_mut!((*self.gic_v3.sgi_ptr()).isenabler0).read_volatile() & bit != 0
-            } else {
-                addr_of_mut!((*self.gic_v3.gicd_ptr()).isenabler[index]).read_volatile() & bit != 0
-            }
+        if interrupt_source < 32 {
+            let mut sgi = self.gic_v3.sgi_ptr(0);
+            field!(sgi, isenabler0).read() & bit != 0
+        } else {
+            let mut gicd = self.gic_v3.gicd_ptr();
+            field!(gicd, isenabler).get(index).unwrap().read() & bit != 0
```
os-d, line 166: Instead of `unwrap()` should we handle any potential error returned?

core/patina_internal_cpu/src/interrupts/aarch64/gic_manager.rs
```diff
@@ -110,135 +100,107 @@ pub unsafe fn gic_initialize(gicd_base: *mut u64, gicr_base: *mut u64) -> Result
...
-        // redistributor interface.
-        let level = unsafe {
-            if interrupt_source < 32 {
-                addr_of_mut!((*self.gic_v3.sgi_ptr()).icfgr[index]).read_volatile() & bit == 0
-            } else {
-                addr_of_mut!((*self.gic_v3.gicd_ptr()).icfgr[index]).read_volatile() & bit == 0
-            }
+        let level = if interrupt_source < 32 {
+            let mut sgi = self.gic_v3.sgi_ptr(0);
+            field!(sgi, icfgr).get(index).unwrap().read() & bit != 0
```
os-d, line 186: Same unwrap questions

Cargo.toml
joschock: There has been a lot of API churn in the arm-gic repo, but they seem to be bumping the minor each time (i.e. the last 3 releases have been 3.0, 4.0, 5.0).  safe_mmio I want to match the version used by arm-gic exactly to avoid issues when accessing safe-mmio structs that arm-gic is producing.  I'll update arm-gic to "0.5"

core/patina_internal_cpu/src/interrupts/aarch64/gic_manager.rs
joschock: These are member function implementations for the various HwInterupt protocols:

for example: 
patina_dxe_core/src/hw_interrupt_protocol.rs:
```
type HardwareInterruptEnableV2 = unsafe extern "efiapi" fn(*mut EfiHardwareInterruptV2Protocol, u64) -> efi::Status;
```
The don't necessarily have to be pub, but the do have to return efi::Status to match the proto. I'm inclined to leave them "pub" as well to reflect the fact that they are exposed to the world via the Protocol Interface.

os-d: Sure, I figured that was the case for some of these, wasn't sure if all of them. I'm also fine to leave them as such.

joschock: acutally, I was mistaken. They are called by wrapper functions thru a lock, so we could change them. However, I think a refactor/cleanup of that nature should be a separate PR to divide our rework from integration for potential bug bisect.

core/patina_internal_cpu/src/interrupts/aarch64/gic_manager.rs
joschock: Ressonable, but I will defer to a future PR to address this separately from bringing in 0.5.0 - since it would require changing the signatures of the function to allow proper error return.

core/patina_internal_cpu/src/interrupts/aarch64/gic_manager.rs
joschock: Likewise.

core/patina_internal_cpu/src/interrupts/aarch64/gic_manager.rs
joschock: https://github.com/OpenDevicePartnership/patina/issues/601 assigned to me to track.

Cargo.toml
joschock: done.

PR #606, opened 2025-07-18
Refactor GIC code for better error handling

joschock:
## Description

Updates gic_manager and hw_interrupt_protocol code for cleaner error handling using Result<T, EfiError> returns instead of efi::Status and cleaning up visibility of various functions. 

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Verfiied GIC initialization succeeds on AArch64 hw as before this refactor. 

## Integration Instructions

N/A - no changes to interfaces consumed outside of patina_internal_cpu.

### DIFF
--- a/core/patina_internal_cpu/src/interrupts/aarch64/gic_manager.rs
+++ b/core/patina_internal_cpu/src/interrupts/aarch64/gic_manager.rs
@@ -102,6 +102,9 @@ pub unsafe fn gic_initialize<'a>(gicd_base: *mut u64, gicr_base: *mut u64) -> Re
 
     // Set binary point reg to 0x7 (no preemption)
     // Safety: this is a legal value for BPR1 register.
+    // Refer to "Arm Generic Interrupt Controller Architecture Specification GIC
+    // architecture version 3 and Version 4" (Arm IHI 0069H.b ID041224)
+    // 12.2.5: "ICC_BPR1_EL1, Interrupt Controller Binary Point Register 1"
     unsafe {
         write_sysreg!(ICC_BPR1_EL1, 0x7u64);
     }


core/patina_internal_cpu/src/interrupts/aarch64/gic_manager.rs
```diff
@@ -123,83 +116,85 @@ pub struct AArch64InterruptInitializer<'a> {
...
+
+        if int_id.is_private() {
             let mut sgi = self.gic_v3.sgi_ptr(0);
-            field!(sgi, isenabler0).read() & bit != 0
+            Ok(field!(sgi, isenabler0).read() & bit != 0)
         } else {
             let mut gicd = self.gic_v3.gicd_ptr();
-            field!(gicd, isenabler).get(index).unwrap().read() & bit != 0
+            //source_to_intid() validates the interrupt source number, so index computed must be valid.
+            Ok(field!(gicd, isenabler).get(index).unwrap().read() & bit != 0)
```
os-d, line 164: Return error if unwrap fails?

core/patina_internal_cpu/src/interrupts/aarch64/gic_manager.rs
```diff
@@ -123,83 +116,85 @@ pub struct AArch64InterruptInitializer<'a> {
...
         let index = (interrupt_source / 16) as usize;
         let bit = 1 << (interrupt_source % 16);
 
-        let level = if interrupt_source < 32 {
+        // validates the interrupt source
+        let int_id = self.source_to_intid(interrupt_source)?;
+
+        let level = if int_id.is_private() {
             let mut sgi = self.gic_v3.sgi_ptr(0);
             field!(sgi, icfgr).get(index).unwrap().read() & bit != 0
```
os-d, line 184: Same unwrap questions here and a couple lines below

core/patina_internal_cpu/src/interrupts/aarch64/gic_manager.rs
joschock: the comment is indicating that the unwrap only fails due to a programming error - i.e. you can only get here and fail the unwrap() if source_to_intid() isn't written correctly. I elected to leave it as an unwrap since it's a programmer error if it occurs. Happy to change it to an expect("fix source_to_intid() pls") if you like :)

os-d: Didn’t read the comment :). In general I think expect is better than unwrap, but either is fine here.

core/patina_internal_cpu/src/interrupts/aarch64/gic_manager.rs
```diff
@@ -111,8 +99,13 @@ pub unsafe fn gic_initialize<'a>(gicd_base: *mut u64, gicr_base: *mut u64) -> Re
     for spi in 0..gic_v3.typer().num_spis() {
         gic_v3.set_interrupt_priority(IntId::spi(spi), None, 0x80);
     }
+
     // Set binary point reg to 0x7 (no preemption)
-    set_binary_point_reg(0x7);
+    // Safety: this is a legal value for BPR1 register.
```
makubacki, line 104: Is there reference documentation that can be linked here?

core/patina_internal_cpu/src/interrupts/aarch64/gic_manager.rs
joschock: same as above, I think.

core/patina_internal_cpu/src/interrupts/aarch64/gic_manager.rs
joschock: I could point to arm GIC ref manual, if desired.

makubacki: I'm looking at the PR for the Rust changes (not familiar with GIC details) but a clear connection to evidence would be helpful to establish.

joschock: Added a reference to the appropriate arch manual.

PR #607, opened 2025-07-18
patina_debugger: Improve the info (monitor ?) response

cfernald:
## Description

Adds instruction address and clarifies the exception type info. This is invoked by calling `!uefiext.info` or `!uefiext.monitor ?` in windbgx.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on Q35

## Integration Instructions

N/A

### DIFF
--- a/core/patina_debugger/src/lib.rs
+++ b/core/patina_debugger/src/lib.rs
@@ -248,17 +248,18 @@ pub fn add_monitor_command(cmd: &'static str, function: MonitorCommandFn) {
 }
 
 /// Exception information for the debugger.
-#[derive(Debug)]
 #[allow(dead_code)]
 struct ExceptionInfo {
     /// The type of exception that occurred.
     pub exception_type: ExceptionType,
+    /// The instruction pointer address.
+    pub instruction_pointer: u64,
     /// The system context at the time of the exception.
     pub context: ExceptionContext,
 }
 
 /// Exception type information.
-#[derive(Debug, PartialEq, Eq)]
+#[derive(PartialEq, Eq)]
 #[allow(dead_code)]
 enum ExceptionType {
     /// A break due to a completed instruction step.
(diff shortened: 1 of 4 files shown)


core/patina_debugger/src/lib.rs
```diff
@@ -248,17 +248,18 @@ pub fn add_monitor_command(cmd: &'static str, function: MonitorCommandFn) {
 }
 
 /// Exception information for the debugger.
-#[derive(Debug)]
 #[allow(dead_code)]
 struct ExceptionInfo {
     /// The type of exception that occurred.
     pub exception_type: ExceptionType,
+    /// The type of exception that occurred.
```
makubacki:
```suggestion
    /// The instruction pointer address.
```

core/patina_debugger/src/lib.rs
```diff
@@ -272,3 +273,17 @@ enum ExceptionType {
     /// A break due to an exception type not handled by the debugger. The exception type is provided.
     Other(u64),
 }
+
+impl core::fmt::Display for ExceptionType {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        match self {
+            ExceptionType::Step => write!(f, "Debug Step"),
+            ExceptionType::Breakpoint => write!(f, "Breakpoint"),
+            ExceptionType::AccessViolation(addr) => write!(f, "Access Violation at {:#X}", addr),
```
os-d, line 282: Do we already print the esr and the exception data for x64? It would be nice to print the exception data meaning (e.g. we faulted because of a read). The esr is obviously more complicated to decode and I think that relying on the user to use external tools to decode is reasonable.

os-d:
@cfernald I think you may have forgotten about this :)

core/patina_debugger/src/lib.rs
cfernald: This is what the `ExceptionType` is really intended to abstract. Its created from the ESR by the ARM code, but this is the architecture agnostic enum for the general reasons. The raw ESR is available for more architecture specifics though robust translation of that is probably better done in the debug extension then on-device.

cfernald: These are available through `arch regs` so will leave it out of the generic info to keep this as the intended simplified view.

os-d: It would be very nice to know, at least on X64, whether the violation was a write, read, or execute violation. But if that is too much work to plumb we can leave it for the user to decode.

PR #608, opened 2025-07-19
Add AdvLoggerLocator and Variable Policy Interface

kouchekiniad:
## Description

This change adds the publishing and policy-setting of an AdvLoggerLocator variable to match the Tianocore-based implementation of the core component of the Advanced Logger found in Mu's AdvLoggerPkg. To facilitate setting the policies of the AdvLoggerLocatorVariable, this change also implements a partial interface for the Mu Variable Policy Protocol.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Booted to Windows on physical platform, confirming that the AdvLoggerLocator variable has been successfully published and write-protected in Windows.

## Integration Instructions

N/A

### DIFF
--- a/components/patina_adv_logger/src/component.rs
+++ b/components/patina_adv_logger/src/component.rs
@@ -18,10 +18,8 @@ use patina_sdk::{
     error::{EfiError, Result},
     runtime_services::{RuntimeServices, StandardRuntimeServices},
     serial::SerialIO,
-    uefi_protocol::{
-        self, ProtocolInterface,
-        variable_policy::{BasicVariablePolicy, MuVariablePolicyProtocol, VariablePolicy},
-    },
+    uefi_protocol::ProtocolInterface,
+    variable_policy::{BasicVariablePolicy, VariablePolicy, VariablePolicyProtocol},
 };
 use r_efi::efi::{self, Guid};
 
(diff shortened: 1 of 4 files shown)


sdk/patina_sdk/src/uefi_protocol/mu_variable_policy.rs
makubacki, line 1: I'd rather avoid too much Mu-specific content leaking throughout `patina` repo, especially branded as Mu (and in the SDK). Advanced logger is a Mu invention. As a component, it gets isolated within its component boundaries. This is replicating interfaces though that are controlled by Project Mu and expected to match the Mu author's definition. Therefore, I'd prefer it to go somewhere like mu_rust_helpers. Then, only the `patina_adv_logger` needs to take a dependency on that (it already does) and it's clear that this component depends on something from Project Mu.

(I know `patina_sdk` already has a `mu_rust_helpers` dependency. I'd like to minimize and eliminate that in the future).

kouchekiniad: That'd requiring adding `patina_sdk` as a dependency of `mu_rust_helpers` which is a dependency of `patina_sdk`. Should I be worried about the circular dependency?

makubacki: Are the two dependencies `EfiError` and `ProtocolInterface`? If so, are those things you can just hookup in the component after using the main content from the Mu repo?

kouchekiniad: @makubacki Would you have an objection to adding only the aspects of variable policy found in Tianocore EDK2's VariablePolicyLib to patina_sdk/as a component service?

kouchekiniad: @makubacki The associated issue is low priority, but would you mind taking a look at this PR again?

makubacki:
@kouchekiniad to create an issue with changes needed to help facilitate this change.

makubacki:
I'm closing this PR since it is based on a branch in the `patina` repo and we're removing all personal branches from the `patina` repo. Please recreate the PR from a fork if you'd like to keep it open.

PR #609, opened 2025-07-21
RFC: Patina Test Execution Configuration

Javagedes:
## Description

Current State: FCP

Additional `patina_test` macro configuration to control when a patina_test executes. Would add additional timing for a patina_test to be executed, including on a protocol installation, on an event group, on a timer, and the normal execution.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

N/A

## Integration Instructions

N/A


docs/src/rfc/text/0000-system-tests.md
```diff
@@ -0,0 +1,169 @@
+# RFC: Patina Test Execution Configuration
+
+Add the ability to control on-system unit test execution to a limited degree, to include (1) Tests executing multiple
+times via event triggers and (2) Allow execution at different points of DXE, rather than only during monolithic
+component execution prior to normal EDKII style component dispatch
```
os-d:
```suggestion
component execution prior to normal EDK II style component dispatch

```
nit: during the docs scrub all instances were changed to this. And for other instances below.

docs/src/rfc/text/0000-system-tests.md
```diff
@@ -0,0 +1,169 @@
...
+## Technology Background
+
+This feature will rely almost entirely on the EDKII style eventing system, however it will not be implemented until
+[Uefi Services](https://github.com/OpenDevicePartnership/patina/pull/592) is implemented, which will provide an
+abstraction for eventing.
+
+## Goals
+
+- #[patina_test] macro configuration for specifying the event that a component can trigger for rather than in the
+  monolithic
```
os-d:
```suggestion
- #[patina_test] macro configuration to specify an event that triggers component execution rather than one time execution during normal dispatch

```
Or something like that, the original wording was just confusing

docs/src/rfc/text/0000-system-tests.md
```diff
@@ -0,0 +1,169 @@
...
+### `TestCase` Changes
+
+`TestCase` will be changed to include a trigger type to inform the component on how to register / run a component
+
+``` rust
+pub enum Trigger {
+    /// A Test that runs immediately during Test component execution.
+    Component,
+    /// A test that runs multiple times, every 'X' seconds.
+    Timer(f32)
```
os-d: Would it really be a float?

makubacki: The time unit should probably be reduced to milliseconds. Firmware can execute from first UEFI execution to OS handoff in ~2.5 seconds on some systems. Seconds is a large interval especially exclusive to code running in DXE right now.

Javagedes: Depends on what unit of time we want to represent. If it is `s`, then yes, a float makes sense. If it is `ms` then maybe not. I'm open to either unit (or ns if necessary).

docs/src/rfc/text/0018-system-tests.md
```diff
@@ -0,0 +1,169 @@
...
+
+Currently, on-system unit tests via patina-test has no support for running tests at any other time then monolithic
+component execution.
+
+## Alternatives
+
+Each entity that wants these type(s) of test would directly write and register their own component that in-turn
+register an event callback that runs their validation testing
+
+## Rust Code Design
```
makubacki, line 62: Given that we're trying to replace protocols with component services, should there be another trigger on a component service being produced?

---

My understanding is that in the alternative:

> Each entity that wants these type(s) of test would directly write and register their own component that in-turn
register an event callback that runs their validation testing

That would naturally be possible via depdendency injection (test component declares dependency on the service), but this proposal is currently focusing on more granular triggers for tests within a broader set of code.

os-d: I think that dependency injection would be a nice way to execute these tests, though currently we can't do dependency injection multiple times, right? i.e. in the case of a test that was going to run at ready to boot and EBS or something. Maybe you could have multiple of the patina tests registered for different dependencies?

docs/src/rfc/text/0000-system-tests.md
```diff
@@ -0,0 +1,169 @@
...
+                          │register an event callback │
+                          └───────────────────────────┘
+```
+
+## Guide-Level Explanation
+
+This feature allows a user to select how a particular `patina_test` is executed. In addition to the default execution
+time of a `patina_test` (which is during `TestRunner` component execution), developers can now chose to delay execution
+of a particular unit test to execute periodically based off a timer using `#[on(timer = X)]`, when a protocol is
+registered using `#[on(protocol = "GUID")]`, or whenever a particular event is triggered  using
```
makubacki:
```suggestion
registered using `#[on(protocol = "GUID")]`, or whenever a particular event is triggered using
```

docs/src/rfc/text/0000-system-tests.md
```diff
@@ -0,0 +1,169 @@
...
+`#[on(event = "GUID")]`.
+
+This new functionality will not provide any breaking changes to the consumer, only allowing additional configuration.
+This change will delay the reporting of all test results until a certain event group runs (undetermined which group,
+but possibly ExitBootServices), even if tests run prior to that.
```
makubacki: We often only boot to EFI shell and `ExitBootServices` does not run booting to shell. I think at least making results at `OnReadyToBoot` available would be helpful for cases we're not booting to an OS.

makubacki: People may want periodic or event-based test results up to Exit Boot Services, and that's fine, they could be dumped again, or a delta of new test results since the last dump point (e.g. `OnReadyToBoot`) be provided at that time.

docs/src/rfc/text/0000-system-tests.md
```diff
@@ -0,0 +1,169 @@
...
+## Requirements
+
+- Support Timer based validation tests (Trigger every X (s, ms, etc))
+- Support Protocol installation based validation tests
+- Support Event based validation tests
+
+## Unresolved Questions
+
+- Should The test runner report multiple executions of the same test as a single result, as individual tests, or
+  as a single test that just has an execution count field?
```
makubacki: What is the current proposal? I may have missed it somewhere else in the RFC.

docs/src/rfc/text/0000-system-tests.md
makubacki: Given that the EFI event interface works off units of 100ns, I suggest to either match that or move to a millisecond. Millisecond probably makes sense. And avoid trying to convert floats into those units (`UINT64`).

Javagedes:
> but is a function call that kicks off a test because some code path was taken.

@os-d I'm curious how much this is needed? We could do something like this, but it would turn the TestRunner component more so into a global static where we then provide `test!(<test_name>)` macro's that are compiled out on release. However that is pretty close to someone just adding something like this to their code

``` rust
#[cfg(debug_assertions)]
{
   // Run some test code here
}
```

os-d:
> > but is a function call that kicks off a test because some code path was taken.
> 
> @os-d I'm curious how much this is needed? We could do something like this, but it would turn the TestRunner component more so into a global static where we then provide `test!(<test_name>)` macro's that are compiled out on release. However that is pretty close to someone just adding something like this to their code
> 
> ```rust
> #[cfg(debug_assertions)]
> {
>    // Run some test code here
> }
> ```

Let's table it here to not muddy this RFC. I'll talk to you offline.

docs/src/rfc/text/0000-system-tests.md
Javagedes: I did not add a proposal for this unresolved question. 

My proposal is to have an extra field in the table for number of executions. Something like

| test name | executions | status |
|---|---|---|
| my_test | 3 | Failed [1 time(s)] |
| other_test | 10 | Success |

docs/src/rfc/text/0000-system-tests.md
Javagedes: `ms` sounds good to me. will update.

docs/src/rfc/text/0018-system-tests.md
makubacki: To be clear, I'm not necessarily suggesting the alternative would be a better approach, it just provides a straightforward answer to a service dependency. As I understand the current proposal, you could treat these like unit tests where they're still local to the component crate and can access crate scoped items and stuff like that which is useful.

makubacki:
> Let's table it here to not muddy this RFC. I'll talk to you offline.

A proposal would be interesting, but I'm not particularly in favor of introducing code complexity in production code implementation to offshoot tests. Not only for the code maintainability and review complexity, but it might become a "crutch" where someone says, "this is easier than actually mocking things the right way". Just my high-level concern with the general concept.

If this is entirely outside the main code path and "hooks" onto by its path or something (which may have been what is proposed), that would alleviate the concern.

os-d:
> > Let's table it here to not muddy this RFC. I'll talk to you offline.
> 
> A proposal would be interesting, but I'm not particularly in favor of introducing code complexity in production code implementation to offshoot tests. Not only for the code maintainability and review complexity, but it might become a "crutch" where someone says, "this is easier than actually mocking things the right way". Just my high-level concern with the general concept.

Yeah, valid concern. My concern is the opposite that someone says, oh, I can't write this as a test, so now I have to write it as production code (that was the origin of this conversation, I wanted to do the heap guard implementation as a periodic pool scanner and Sean pointed out that without hooks we couldn't do it after every alloc/free, so you would need the production code to do this instead of a test or have the production code do heap guard (which still might happen anyway, that's orthogonal to this). So, it led to an interesting discussion of what could be done to try to isolate tests from the production code, but obviously there are tradeoffs.

docs/src/rfc/text/0018-system-tests.md
Javagedes: To Oliver's statement, He is correct, we cannot run the same component multiple times as I created struct components to be `self` and not `&self`. This could probably be changed, and we could theoretically run components multiple times if we wished.

Regardless of that sentiment, I've purposefully not added the ability for components to register other components, as I feared the negative impacts of indirection and hidden components running / executing. That can be changed if we have the justification (like this), and possibly add some better metadata for debugging so we can track who registered components, services, etc. Then it would be possible to trigger a test on service installation based off of dependency injection.

docs/src/rfc/text/0000-system-tests.md
Javagedes: Updated Guide-Level explanation to specify we will post results at `OnReadyToBoot` with a diff of results at `ExitBootServices`, if the platform gets to that point. Thanks!

docs/src/rfc/text/0018-system-tests.md
makubacki: My initial comment was not about a component running multiple times or changing the registration process for components from an end user perspective. But to allow the ability to run a triggered test as defined in this RFC based on a service being produced (which may be optional on the owning component).

Without that, it seems like there is a gap where moving a protocol to a service prevents a test from being able to be triggered on that external interface. Is that not the case?

Javagedes: I agree that there is a gap, I think the solution @os-d was hinting at was to basically just register the test as a component and then the component will just execute eventually (not immediately) after the component is registered. However, at this point in time we cannot do that because components cannot register other components.

I'm sure we can do it a different way too. Not sure what that way would look like off the top of my head.

makubacki: I'd be okay with noting it as an unresolved question for now and addressing it later if needed.

makubacki:
> Yeah, valid concern. My concern is the opposite that someone says, oh, I can't write this as a test, so now I have to write it as production code (that was the origin of this conversation, I wanted to do the heap guard implementation as a periodic pool scanner and Sean pointed out that without hooks we couldn't do it after every alloc/free, so you would need the production code to do this instead of a test or have the production code do heap guard (which still might happen anyway, that's orthogonal to this). So, it led to an interesting discussion of what could be done to try to isolate tests from the production code, but obviously there are tradeoffs.

Makes sense. Certain core "services" like memory allocation are also somewhat unique (more frequently called, more essential) to most other code, so I think we could also consider having special hooks for these as opposed to a widespread system anyone can tap into. In any case, it would be a good RFC for discussion.

os-d:
> Makes sense. Certain core "services" like memory allocation are also somewhat unique (more frequently called, more essential) to most other code, so I think we could also consider having special hooks for these as opposed to a widespread system anyone can tap into. In any case, it would be a good RFC for discussion.

I'll think on it some more and write an RFC for it, even if it becomes a rejected RFC I agree I think it would be good to reason through.

docs/src/rfc/text/0000-system-tests.md
Javagedes: Updated documentation to reflect proposed table format.

Javagedes:
Moving to FCP for any final comments.

Javagedes:
I will be merging this soon barring any comments.

docs/src/rfc/text/0018-system-tests.md
```diff
@@ -0,0 +1,188 @@
...
+fn my_test() -> Result { ... }
+
+// Run every "X"
+#[patina_test]
+#[on(timer = 1)]
+fn my_test() -> Result { ... }
+
+// Run on protocol installation
+#[patina_test]
+#[on(protocol = "GUID")]
```
cfernald, line 78: For my education, why create a seperate mechanism instead of investing in dep-inj for these scenarios?

Javagedes: That is actually a good idea I did not think about.

As it stands, the limitation is we cannot dependency inject Protocols, nor do we have the infrastructure setup to register dependency injectable functions for event callbacks.

Both are absolutely doable, just not implemented. I've also been limiting the scope of dependency injectible variables to include UEFI / EDKII specific things (So Protocols, Events, etc). If we decide we will do that, then all of this is possible

PR #610, opened 2025-07-21
patina_dxe_core: Only Unmap Page 0 If It Exists

os-d:
## Description

This patch updates the logic to unmap page 0 after resource HOBs are added. It previously attempted to always do it, but on some platforms page 0 does not exist (in which case the page table will not map it anyway). This checks if the memory exists before attempting to unmap, so it does not cause a debug_assert.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on a physical platform that did not have page 0 mapped and was hitting the debug_assert. It no longer is. Also tested on Q35 and SBSA booting to Windows.

## Integration Instructions

N/A.

### DIFF
--- a/patina_dxe_core/src/gcd/spin_locked_gcd.rs
+++ b/patina_dxe_core/src/gcd/spin_locked_gcd.rs
@@ -11,7 +11,10 @@ use alloc::{boxed::Box, slice, vec, vec::Vec};
 use core::{fmt::Display, ptr};
 use patina_sdk::{base::DEFAULT_CACHE_ATTR, error::EfiError};
 
-use mu_pi::{dxe_services, hob};
+use mu_pi::{
+    dxe_services::{self, GcdMemoryType},
+    hob,
+};
 use mu_rust_helpers::function;
 use patina_internal_collections::{Error as SliceError, Rbt, SliceKey, node_size};
 use patina_sdk::{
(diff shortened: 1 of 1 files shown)


patina_dxe_core/src/gcd/spin_locked_gcd.rs
```diff
@@ -2145,11 +2148,19 @@ impl SpinLockedGcd {
         }
 
         // make sure we didn't map page 0 if it was reserved or MMIO, we are using this for null pointer detection
-        if let Err(err) = self.set_memory_space_attributes(0, UEFI_PAGE_SIZE, efi::MEMORY_RP) {
-            // if we fail to set these attributes we can continue to boot, but we will not be able to detect null
-            // pointer dereferences.
-            log::error!("Failed to unmap page 0, which is reserved for null pointer detection. Error: {:?}", err);
-            debug_assert!(false);
+        // only do this if page 0 actually exists
+        if let Ok(descriptor) = self.get_memory_descriptor_for_address(0) {
```
Javagedes, line 2152: This is a lot of nested ifs. Seems like we should use the opposite pattern of this and turn this into a function that we bail out of early using `let Some(x) = y else { return };`

Javagedes:
``` rust
let Ok(descriptor) = self.get_memory_descriptor_for_address(0) else {
  return;
};

if descriptor.memory_type == GcdMemoryType::NonExistent {
  return;
};

if let Err(err) = self.set_memory_space_attributes(0, UEFI_PAGE_SISZE, efi::MEMORY_RP) {
  log::error!(
    "Failed to unmap page 0, which is reserved for null pointer detection. Error: {:?}",
    err
  );
  debug_assert!(false);
}
```

os-d: What I really want is chained lets, but Rust tells me that is an unstable feature :). I don't mind changing this, however, this logic is at the end of this function by chance, not design. So I don't want to add returns, because that could have unintended consequences when this is updated later and is no longer the end. I can refactor it into a match statement with an if arm wrapping the original if statement, though that is really the same thing

Javagedes: let chains were just stabilized in [1.88](https://releases.rs/docs/1.88.0/). I'm not against you adding that as an unstable feature here, until we upgrade past 1.85 :)

makubacki: It technically does not fall under our unstable guidelines (there is an alternative).

https://github.com/OpenDevicePartnership/patina/blob/4c321c078e7cf873781f155dea2c21433ad7a9fa/docs/src/dev/unstable.md?plain=1#L13-L21

I'm not entirely opposed to it though. But at least filing a tracking issue would be nice if you go down that path to explain the usage and resolution.

os-d: I think that let chains are something we could propagate across patina and tighten up some code structures (though it is really just a visual thing, not a real change to code flow). I'm not sure this warrants bringing in the unstable feature, though certainly better to do since it is already stabilized. I will file a backlog item that notes this case and suggests moving it and similar code patterns to let chains when we move to 1.88. It could be a good first issue.

os-d: https://github.com/OpenDevicePartnership/patina/issues/613

PR #619, opened 2025-07-23
[REBASE&FF] Prevent panic on failed consumption of PageAllocation

Javagedes:
## Description

A PageAllocation can be converted to different types in multiple different ways. Some of these ways are fallable, if the type being converted to is too large. In a scenario where the conversion failed, the code was found to panic because the `drop` implementation is called, which purposefully panics. 

To fix this, on a failed conversion, we manually free the pages and forget the instance, so that `drop` is not called and the panic does not occur.

The other option would be to add a new field to the page allocation that tracked if the page allocation failed to convert, and use that to determine if we should panic. This was found to be a poor solution.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Added CI tests to ensure the panic no longer occurs.

## Integration Instructions

N/A

### DIFF
--- a/sdk/patina_sdk/src/component/service/memory.rs
+++ b/sdk/patina_sdk/src/component/service/memory.rs
@@ -599,7 +599,7 @@ impl PageAllocation {
     }
 
     /// Frees the allocated pages of memory this struct manages.
-    /// 
+    ///
     /// This is not a public method as it invalidates `Self` without consuming `self`.
     /// This should only be used internally to free the memory when dropping `self`.
     fn free_pages(&mut self) {
@@ -912,5 +912,33 @@ mod tests {
         let page = service.allocate_pages(1, AllocationOptions::new()).unwrap();
         let my_thing = page.try_leak_as(42).unwrap();
         assert_eq!(*my_thing, 42);
+    }
+
+    #[test]
+    fn test_failed_try_into_box_does_not_panic() {
+        let mm = Service::mock(Box::new(StdMemoryManager::new()));
+
+        let page = mm
+            .allocate_pages(1, AllocationOptions::default())
+            .unwrap_or_else(|e| panic!("Failed to allocate pages: {:?}", e));
+
+        assert!(
+            page.try_into_box([42_u8; UEFI_PAGE_SIZE + 1]).is_none(),
+            "Expected allocation to fail due to insufficient size for Box<[u8]>"
+        );
+    }
+
+    #[test]
+    fn test_failed_try_leak_as_does_not_panic() {
+        let mm = Service::mock(Box::new(StdMemoryManager::new()));
+
+        let page = mm
+            .allocate_pages(1, AllocationOptions::default())
+            .unwrap_or_else(|e| panic!("Failed to allocate pages: {:?}", e));
+
+        assert!(
+            page.try_leak_as([42_u8; UEFI_PAGE_SIZE + 1]).is_none(),
+            "Expected allocation to fail due to insufficient size for [u8]"
+        );
     }
 }


sdk/patina_sdk/src/component/service/memory.rs
```diff
@@ -585,10 +597,12 @@ impl PageAllocation {
...
     }
-}
 
-impl Drop for PageAllocation {
-    fn drop(&mut self) {
+    /// Frees the allocated pages of memory this struct manages.
+    ///
+    /// This is not a public method as it invalidates `Self` without consuming `self`.
+    /// This should only be used internally to free the memory when dropping `self`.
+    fn free_pages(&mut self) {
```
cfernald, line 605: Since this doesn't consume `self`, which makes sense for use in drop, we should zero out the address and size so that we can detect and panic if free_pages is called more then once. e.g. implementation forgets to call `forget`

Javagedes: This is outside the scope of the memory management trait or custom implementations. The only way I can see "someone" not calling forget, is for one of us editing the PageAllocation code and messing up.

I'm not sure how I feel about wasted performance (however minor) to check code that we purely own. Seems like something we can just cover with tests.

Is there an example I'm not thinking of, where this would be in the purview of a memory manager implementation or consumer?

sdk/patina_sdk/src/component/service/memory.rs
```diff
@@ -508,9 +508,15 @@ impl PageAllocation {
...
-    #[cfg(feature = "alloc")]
-    pub fn try_into_box<T>(self, value: T) -> Option<Box<T, PageFree>> {
+    #[cfg(any(test, feature = "alloc"))]
+    pub fn try_into_box<T>(mut self, value: T) -> Option<Box<T, PageFree>> {
         if self.byte_length() < size_of::<T>() {
+            // This is an intentional case where the struct is being dropped,
+            // but we want to avoid triggering the panic in its `Drop` implementation.
+            // To handle this safely, we manually free the pages and then call `forget`
+            // to prevent `drop` from running.
+            self.free_pages();
```
os-d, line 518: Should we actually free the memory here? We don't know what the consumer will do after this try fails, they may try to use the memory another way, which I think is valid, if they allocated it, they keep it until they free it (or drop, which frees it). I think it might be odd to free it here. But I guess since this consumes self, the consumer wouldn't compile if they tried to use the original memory after that try call.

sdk/patina_sdk/src/component/service/memory.rs
```diff
@@ -556,8 +562,14 @@ impl PageAllocation {
...
     ///
     #[must_use]
-    pub fn try_leak_as<T>(self, value: T) -> Option<&'static T> {
+    pub fn try_leak_as<T>(mut self, value: T) -> Option<&'static T> {
         if self.byte_length() < size_of::<T>() {
+            // This is an intentional case where the struct is being dropped,
+            // but we want to avoid triggering the panic in its `Drop` implementation.
+            // To handle this safely, we manually free the pages and then call `forget`
+            // to prevent `drop` from running.
+            self.free_pages();
```
os-d, line 571: Same question here.

sdk/patina_sdk/src/component/service/memory.rs
Javagedes: I talked to @cfernald about this. The current interface is that we consume the page allocation whether it passes or fails. So there is no opportunity to do something else with it.

I talked about possibly changing the interface to a result that returns back the page allocation, in which none of this change is necessary.

But Chris brought up a valid point that if it fails to allocate, it's not really a "use case" to try and re-use that allocation for something else. It's more so just an implementation error for not allocating enough in the first place.

Since it failed, the normal solution would be to create a new allocation with more space, since the only way to fail at the moment is for the size to be too small.

os-d: I think that's fair, I'm assuming the expected use case is `mm.alloc().try_into_box().unwrap()`, i.e. it being chained, in which case that makes sense. Now I do think there may be use cases in the future of `mm.alloc()` -> do something to the page that makes sense to that component -> `page.try_into_box().unwrap()`, which is the scenario I was thinking about. I'm fine to address this when it comes up or just to say no, the interface doesn't allow that, which is probably okay.

sdk/patina_sdk/src/component/service/memory.rs
cfernald: That is fair, and you are correct that it would only be internal code that could cause it. I'm alright leaving it out since it would be an internal error in a relatively small amount of code.

sdk/patina_sdk/src/component/service/memory.rs
Javagedes: At this point in time they cannot manipulate the page table in any way other than a way that consumes `self` and converts it to another type. You can get the `byte_length` and the `page_count`, but you cannot get the underlying address without converting it to some type, which consumes `self`.

I do agree, if we introduce a way to manipulate the bytes without consume `self`, then yes, we may need to re-evaluate. Particularly if we provide a `realloc()` method. In that case doing something like 

```rust
let mut pa = mm.allocate_pages(1, default());

let my_type = loop {
    match pa.try_into_raw_ptr() {
      Ok(val) => break val,
      Err(p) => { p.realloc(p.page_count * 2); pa = p; };
    }
};
```

Issue #621, opened 2025-07-24
[Bug]: PageAllocation::into_boxed_slice calls `Drop` on uninitialized memory


Javagedes:
### Is there an existing issue for this?

- [x] I have searched existing issues

### Current Behavior

When calling `into_boxed_slice::<T>()` on a `PageAllocation`, the `Drop` implementation for `T` is called on every uninitialized entry in the slice as we fill the slice with T::default(). This is because `fill_with` assumes that the data is initialized, which means it thinks it is replacing an existing value with the new value (T::defualt()).

### Expected Behavior

`Drop` should not be called when filling the slice with default values.

### Steps To Reproduce

The following test detected the error:

```rust
#[test]
fn test_into_boxed_slice_will_call_drop_properly() {
  static DROP_COUNT: AtomicUsize = AtomicUsize::new(0);

  #[derive(Default)]
  struct MyStruct(usize);
  impl MyStruct {
      fn value(&self) -> usize {
          self.0
      }
  }

  impl Drop for MyStruct {
      fn drop(&mut self) {
          DROP_COUNT.fetch_add(1, std::sync::atomic::Ordering::SeqCst);
      }
  }

  let mm = Box::leak(Box::new(StdMemoryManager::new()));

  let pa = mm.allocate_pages(1, AllocationOptions::default()).expect("Should not fail for test.");
  {
      let boxed_slice = pa.into_boxed_slice::<MyStruct>();
      assert_eq!(boxed_slice.len(), UEFI_PAGE_SIZE / size_of::<MyStruct>());

      let mut i = 0;
      boxed_slice.iter().for_each(|item| {
          assert_eq!(item.value(), 0, "Default value of MyStruct should be zero");
          i += 1;
      });
  }
  assert_eq!(
    DROP_COUNT.load(std::sync::atomic::Ordering::SeqCst),
    UEFI_PAGE_SIZE / size_of::<MyStruct>(),
    "Drop should be called for each item in the boxed slice"
  );
}
```

### Build Environment

```markdown
- OS(s): Any
- Tool Chain(s): Any
- Targets Impacted: Any
```

### Version Information

```text
4029abb
```

### Urgency

Medium

### Are you going to fix this?

I will fix it

### Do you need maintainer feedback?

No maintainer feedback needed

PR #623, opened 2025-07-24
patina_sdk: PageAllocation: Do not drop uninitialized memory

Javagedes:
## Description

This commit does the following two things:

1. Brings the user interface for leaking to be similar to [Box::leak](https://doc.rust-lang.org/std/boxed/struct.Box.html#method.leak) which is to say we no longer require the lifetime of the leaked data to be `static`, but still clearly state that dropping the reference will result in a memory leak.

2. Similar to #622 and the issue #621, when filling a slice, the existing memory is treated as initialized and is thus dropped when replaced with the default value. #622 resolved the issue in `into_boxed_slice` but did not fix `leak_as_slice`, which this commit resolves by moving the logic of `into_boxed_slice` into `leak_as_slice` then re-using `leak_as_slice` in `into_boxed_slice`

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

CI passes

## Integration Instructions

N/A

### DIFF
--- a/sdk/patina_sdk/src/component/service/memory.rs
+++ b/sdk/patina_sdk/src/component/service/memory.rs
@@ -552,13 +552,13 @@ impl PageAllocation {
 
     /// Converts the allocation and leaks the memory as a mutable `T`.
     ///
-    /// This function is similar to [Box::lea] in terms of caller responsibility for memory
+    /// This function is similar to [Box::leak] in terms of caller responsibility for memory
     /// management. Dropping the returned reference will cause a memory leak.
     ///
     /// # Returns
     ///
     /// - `None` if the size of the value is larger than the allocation.
-    /// - `Some(&T)` of the initialized value.
+    /// - `Some(&mut T)` of the initialized value.
     ///
     #[must_use]
     pub fn try_leak_as<'a, T>(mut self, value: T) -> Option<&'a mut T> {


sdk/patina_sdk/src/component/service/memory.rs
```diff
@@ -540,16 +540,14 @@ impl PageAllocation {
     /// of bytes in the allocation divided by the size of `T`.
     #[must_use]
     #[cfg(any(test, feature = "alloc"))]
-    pub fn into_boxed_slice<T: Default>(self) -> Box<[T], PageFree> {
+    pub fn into_boxed_slice<T: Default + 'static>(self) -> Box<[T], PageFree> {
```
os-d: Is requiring this to be static okay? The leak function returning a static obj makes sense, leaking is of course static. But will the allocate return something that is static?

Javagedes: Yes, the underlying allocation will always be static. That is a promise of `core::mem::alloc` / the Allocator trait. Something somewhere has to be responsible for calling the dealloc function. This was verified via the test added in this PR. You can see that drop is never called on any of the underlying slice entries, which means nothing deallocated the bytes.

Really all this `T: +'static` does is make it such that you cannot pass in an object that has a lifetime other than 'static associated with it. An example would be if you have a struct like `MyStruct<'a>(&'a str);`, the only way you could use this function with it is to have the `'a` be static. You cannot associate a lifetime other than 'static for it because then you have a conflict - `MyStruct` will be alive for the lifetime of the program, but the lifetime `&'a str` could theoretically not live as long as the lifetime of the program, which would be undefined behavior.

Honestly it was probably a bug that we didn't require 'static before-hand, I'm guessing we got around it with some of the pointer casting we were doing. With my changes, rust realized what we were doing and yelled at me that this needed to be 'static. I'd consider this a memory safety improvement.

os-d: That makes sense to me, if alloc is returning something static. I am still a bit confused on how the lifetimes work, because we may expect to alloc something and have it drop later, but I guess the drop must call dealloc? I'll follow up offline.

PR #625, opened 2025-07-25
RFC: Move to cargo-llvm-cov for Code Coverage

os-d:
## Description

Status: FCP

This introduces an RFC that proposes to move to cargo-llvm-cov for code coverage. Details in the RFC.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [x] Includes documentation?

## How This Was Tested

N/A.

## Integration Instructions

N/A.

### DIFF
--- a/cspell.yml
+++ b/cspell.yml
@@ -55,6 +55,7 @@ words:
   - getrsp
   - getsp
   - granularities
+  - grcov
   - gread
   - guids
   - gwrite
--- a/docs/src/rfc/text/0013-llvm-cov-coverage.md
+++ b/docs/src/rfc/text/0013-llvm-cov-coverage.md
@@ -1,0 +1,102 @@
+# RFC: Code Coverage Using llvm-cov
+
+This RFC proposes using the [cargo-llvm-cov crate](https://crates.io/crates/cargo-llvm-cov) for code coverage instead
+of [cargo-tarpaulin](https://crates.io/crates/cargo-tarpaulin), the current solution.
+
+## Change Log
+
+- 2025-07-25: Initial draft of RFC
+
+## Motivation
+
+Currently, Patina uses `cargo-tarpaulin` to calculate code coverage from unit and integration tests. `cargo-tarpaulin`
+has been observed to miscalculate line coverage, even when using the `llvm` backend, which is purported to be the
+same as `cargo-llvm-cov`. When running the same configuration and code under test with `cargo-llvm-cov`, line coverage
+has been observed to be much more accurate.
+
+We desire to have accurate code coverage results as such, this RFC proposes moving to `cargo-llvm-cov`.
+
+## Technology Background
+
+[Code Coverage](https://en.wikipedia.org/wiki/Code_coverage) is a standard concept in software engineering.
+
+## Goals
+
+1. Provide accurate code coverage of our unit and integration tests.
+
+## Requirements
+
+1. **Line Coverage Results** - We must show a line coverage percentage to track how much of Patina has been tested.
+2. **Maintain Test Framework** - We must maintain our existing build and test frameworks.
+
+## Unresolved Questions
+
+- What is the best configuration for getting accurate coverage results? Small scale testing and research has given this
+  profile:
+
+  ```rust
+  [profile.test]
+  opt-level = 0
+  debug = true
+  debug-assertions = true
+  overflow-checks = true
+  lto = false
+  incremental = false
+  codegen-units = 1
+  ```
+
+- Does `cargo-llvm-cov` bring in any nightly features that we are not okay with? The
+  [documentation](https://crates.io/crates/cargo-llvm-cov) notes that the following require nightly/are unstable:
+
+  - `llvm-tools-preview` is required to execute `cargo llvm-cov`, but it is not required to run the tests under a
+    nightly toolchain
+  - [#[coverage(off)] attribute](https://github.com/rust-lang/rust/issues/84605), required to disable code coverage on
+    code, including test code
+  - branch coverage (optional)
+  - doc test coverage (optional)
+
+- `cargo-llvm-cov` is not as widely used as `cargo-tarpaulin`. Does it meet our standards? Recent versions of
+  `cargo-llvm-cov` have ~80k downloads, `cargo-tarpaulin` has ~1.7m. Both appear to have a single maintainer who is the
+  primary contributors.
+
+- Are there performance differences? Research indicates that `cargo-llvm-cov` is faster, but has not been verified.
+
+## Prior Art
+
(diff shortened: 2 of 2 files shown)


docs/src/rfc/text/0000-llvm-cov-coverage.md
```diff
@@ -0,0 +1,99 @@
...
+- **Use other coverage tools** - There may be better options, either other Rust crates or non-Rust tools, but research
+  would need to be done to verify.
+
+## Rust Code Design
+
+This is mostly a CI change. The only Rust code changes would be:
+
+- Adding a `profile.test` in `Cargo.toml` as noted above to disable optimizations for tests
+- Adding the `#[coverage(off)]` attribute to test modules and non-testable functions
+- Updating documentation, pipelines, makefiles, etc. to run `cargo-llvm-cov` instead of `cargo-tarpaulin`
```
Javagedes: not sure if we want to document it in the RFC but the Makefile.toml and rust-toolchain.toml needs updated

docs/src/rfc/text/0000-llvm-cov-coverage.md
```diff
@@ -0,0 +1,99 @@
...
+
+- `cargo-llvm-cov` is not as widely used as `cargo-tarpaulin`. Does it meet our standards? Recent versions of
+  `cargo-llvm-cov` have ~80k downloads, `cargo-tarpaulin` has ~1.7m. Both appear to have a single maintainer who is the
+  primary contributors.
+
+- Are there performance differences? Research indicates that `cargo-llvm-cov` is faster, but has not been verified.
+
+## Prior Art
+
+As noted, the current code coverage solution used `cargo-tarpaulin`.
```
Javagedes: Did cargo-tarpaulin include doc tests as coverage? There is a feature for cargo-llvm-cov to include doc tests in the coverage calculateion: `--doctests`

docs/src/rfc/text/0000-llvm-cov-coverage.md
```diff
@@ -0,0 +1,99 @@
...
+## Change Log
+
+- 2025-07-25: Initial draft of RFC
+
+## Motivation
+
+Currently, Patina uses `cargo-tarpaulin` to calculate code coverage from unit and integration tests. `cargo-tarpaulin`
+has been observed to miscalculate line coverage, even when using the `llvm` backend, which is purported to be the
+same as `cargo-llvm-cov`. When running the same configuration and code under test with `cargo-llvm-cov`, line coverage
+has been observed to be much more accurate.
```
makubacki: Have you seen any case where `cargo-llvm-cov` is not accurate? Based on the initial results, it shows more coverage than tarpaulin. I just want to make sure it doesn't have a higher rate of false positives that are getting mixed into the data with some of the misses from tarpaulin now being covered.

os-d: I have not observed any false positives. That being said, I have done a small usage of it. It would be great if other folks want to test drive it as well. It does include more source lines in its calculation than `cargo-tarpaulin` does, but it seems to correctly also include them in covered lines (things like blank lines and comments). So it may skew the coverage a bit there, but from one `cargo-llvm-cov` result to another it should be consistent.

os-d:
The team is looking at grcov as well before deciding on a coverage option. This RFC will be updated to either note grcov and why cargo-llvm-cov is the chosen option or choose grcov.

docs/src/rfc/text/0000-llvm-cov-coverage.md
vineelko: @os-d, I think it would be helpful to add some more details in the RFC on how to install/configure `llvm-cov` and finally how to view the report. Any `Makefile.toml` changes required for running it inplace of `cargo make test`. Feel free to share a private branch(it doesn't have to be fully flushed) so that it is easy for others to give it a shot.

os-d: @vineelko , all I did was this:

```rust
diff --git a/Cargo.toml b/Cargo.toml
index 3a82b38..e1c6567 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,3 +18,13 @@ cfg-if = "1.0.0"

 [features]
 doc = []
+
+[profile.test]
+opt-level = 0
+debug = true
+debug-assertions = true
+overflow-checks = true
+lto = false
+incremental = false
+codegen-units = 1
+
diff --git a/Makefile.toml b/Makefile.toml
index 122c2bc..6092af2 100644
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -6,7 +6,7 @@ RUSTC_BOOTSTRAP = 1
 BUILD_FLAGS = "--profile ${RUSTC_PROFILE} -Zbuild-std=core,compiler_builtins,alloc -Zbuild-std-features=compiler-builtins-mem -Zunstable-options --timings=html"
 UEFI_CRATES = "-p patina_paging"
 BUILD_CRATES = "-p patina_paging"
-COV_FLAGS = { value = "--out html --exclude-files **/tests/*", condition = { env_not_set = ["COV_FLAGS"] } }
+COV_FLAGS = { value = "--lcov --profile test --ignore-filename-regex **/tests/*", condition = { env_not_set = ["COV_FLAGS"] } }

 [env.development]
 RUSTC_PROFILE = "dev"
@@ -80,6 +80,13 @@ clear = true
 command = "cargo"
 args = ["tarpaulin", "@@split(COV_FLAGS, )", "--output-dir", "${CARGO_MAKE_WORKSPACE_WORKING_DIRECTORY}/target"]

+[tasks.llvm_coverage]
+description = "Build and run all tests and calculate coverage."
+install_crate = false
+clear = true
+command = "cargo"
+args = ["llvm-cov", "@@split(COV_FLAGS, )", "--output-path", "${CARGO_MAKE_WORKSPACE_WORKING_DIRECTORY}/target/lcov.info"]
+
 [tasks.clippy]
 description = "Run cargo clippy."
 clear = true
```

in patina-paging. That let me run `cargo make llvm_coverage` vs `cargo make coverage` to compare (note that the `COV_FLAGS` were still toggled, you could make those separate for easier comparison.

vineelko: I have hard time getting it working on patina repo. It always fails with below error with no report. Will investigate it little latter. 

warning: 1 functions have mismatched data

docs/src/rfc/text/0000-llvm-cov-coverage.md
os-d: Yes, tarpaulin includes doctests, but not function, branch, or region coverage. I'll note that if I didn't already.

os-d:
Updating status to FCP.

PR #627, opened 2025-07-28
RFC: SMBIOS

kat-perez:
Proposes a new `SmbiosProvider` service

## Description

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [x] Includes documentation?

## How This Was Tested

N/A

## Integration Instructions

N/A

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,7 +4,7 @@ resolver = "2"
 members = ["components/*", "core/*", "sdk/*", "patina_dxe_core"]
 
 [workspace.package]
-version = "11.3.2"
+version = "11.3.3"
 license = "Apache-2.0"
 edition = "2024"
 rust-version = "1.89"
@@ -31,21 +31,21 @@ linkme = { version = "^0.3.29" }
 log = { version = "0.4", default-features = false }
 mu_rust_helpers = { version = "3.0.2" }
 num-traits = { version = "0.2", default-features = false }
-patina = { version = "11.3.2", path = "sdk/patina" }
-patina_debugger = { version = "11.3.2", path = "core/patina_debugger" }
-patina_ffs = { version = "11.3.2", path = "sdk/patina_ffs" }
-patina_ffs_extractors = { version = "11.3.2", path = "sdk/patina_ffs_extractors" }
-patina_internal_collections = { version = "11.3.2", path = "core/patina_internal_collections", default-features = false }
-patina_internal_cpu = { version = "11.3.2", path = "core/patina_internal_cpu" }
-patina_internal_depex = { version = "11.3.2", path = "core/patina_internal_depex" }
-patina_internal_device_path = { version = "11.3.2", path = "core/patina_internal_device_path" }
+patina = { version = "11.3.3", path = "sdk/patina" }
+patina_debugger = { version = "11.3.3", path = "core/patina_debugger" }
+patina_ffs = { version = "11.3.3", path = "sdk/patina_ffs" }
+patina_ffs_extractors = { version = "11.3.3", path = "sdk/patina_ffs_extractors" }
+patina_internal_collections = { version = "11.3.3", path = "core/patina_internal_collections", default-features = false }
+patina_internal_cpu = { version = "11.3.3", path = "core/patina_internal_cpu" }
+patina_internal_depex = { version = "11.3.3", path = "core/patina_internal_depex" }
+patina_internal_device_path = { version = "11.3.3", path = "core/patina_internal_device_path" }
 patina_lzma_rs = { version = "0.3.1", default-features = false }
-patina_macro = { version = "11.3.2", path = "sdk/patina_macro" }
+patina_macro = { version = "11.3.3", path = "sdk/patina_macro" }
 patina_mtrr = { version = "1.0.0" }
 patina_paging = { version = "9" }
-patina_performance = { version = "11.3.2", path = "components/patina_performance" }
-patina_pi = { version = "11.3.2", path = "sdk/patina_pi" }
-patina_stacktrace = { version = "11.3.2", path = "core/patina_stacktrace" }
+patina_performance = { version = "11.3.3", path = "components/patina_performance" }
+patina_pi = { version = "11.3.3", path = "sdk/patina_pi" }
+patina_stacktrace = { version = "11.3.3", path = "core/patina_stacktrace" }
 proc-macro2 = { version = "1" }
 quote = { version = "1" }
 r-efi = { version = "5.0.0", default-features = false }


docs/src/rfc/text/0014-smbios.md
```diff
@@ -0,0 +1,1082 @@
...
+
+    fn version(&self) -> (u8, u8) {
+        (self.major_version, self.minor_version)
+    }
+}
+```
+
+### SMBIOS Component
+
+A single component `SmbiosInit` initializes the SMBIOS tables, registers the SMBIOS provider service,
```
makubacki: Can you add more detail about the relationship between `SmbiosManager` and `SmbiosInit`?

docs/src/rfc/text/0014-smbios.md
```diff
@@ -0,0 +1,1082 @@
...
+}
+```
+
+## Guide-Level Explanation
+
+All interaction with SMBIOS records will be mediated by the `SmbiosProvider` trait interface.
+Consumers will access this service as follows:
+
+```rust
+pub fn component(smbios_provider: Service<dyn SmbiosProvider>) -> Result<()> {
```
makubacki: I've typically tried to call components "provider" as it is the entity "providing" something. In this case, I'd consider something like `SmbiosRecords` for the service name if it is entirely scoped to record operations.

docs/src/rfc/text/0014-smbios.md
```diff
@@ -0,0 +1,1082 @@
...
+
+1. The API should provide all necessary SMBIOS functionality as a service to components
+2. The API should utilize Rust best practices, particularly memory safety and error handling
+3. The SMBIOS service should produce protocols equivalent to the current C implementations, preserving existing C functionality
+4. Support both SMBIOS 2.x (32-bit) and SMBIOS 3.x (64-bit) table formats
+5. Provide safe string manipulation for SMBIOS records
+
+## Unresolved Questions
+
+- Should the API expose separate interfaces for 32-bit and 64-bit SMBIOS tables, or handle this internally?
```
makubacki: Can you add the pros and cons to each option here?

docs/src/rfc/text/0014-smbios.md
```diff
@@ -0,0 +1,1082 @@
...
+- Is there value in exposing lower-level table construction functionality to advanced users?
```
makubacki: Can you state whether this is required for today's use cases for what you have in mind? If so, provide an example use case.

docs/src/rfc/text/0014-smbios.md
```diff
@@ -0,0 +1,1082 @@
...
+- Should we provide typed interfaces for specific SMBIOS record types (Type 0, Type 1, etc.)?
```
makubacki: Can you provide example usage of the current design and an example of this and note the pros and cons?

docs/src/rfc/text/0014-smbios.md
```diff
@@ -0,0 +1,1082 @@
...
+
+## Prior Art (Existing PI C Implementation)
+
+The Rust-based SMBIOS implementation follows the SMBIOS protocol
```
makubacki: Given this is the priror art section, which "Rust-based SMBIOS implementation" is being referenced here? Can you clarify in the RFC text?

docs/src/rfc/text/0014-smbios.md
makubacki: If a similar name like that is used, you could also consider renaming functions to remove redundancy. For example, `add_record` becomes `add`.

docs/src/rfc/text/0014-smbios.md
```diff
@@ -0,0 +1,1082 @@
...
+
+/// SMBIOS record type
+pub type SmbiosType = u8;
+
+/// SMBIOS string maximum length per specification
+pub const SMBIOS_STRING_MAX_LENGTH: usize = 64;
+
+/// Enhanced error handling
+#[derive(Debug, Clone, PartialEq)]
+pub enum SmbiosError {
```
makubacki: An advantage of custom error types is that the enum variants can be specific to the use case. You can consider dropping the comments next to some variants. For example:

```rust
AlreadyStarted, // Handle already in use
```

Becomes:

```rust
HandleAlreadyInUse,
```

docs/src/rfc/text/0014-smbios.md
makubacki: This section states:

> A single component `SmbiosInit` initializes the SMBIOS tables, registers the SMBIOS provider service, and installs the SMBIOS protocol.

The next section states:

> The global `SmbiosManager` will produce a C protocol to support the existing EDK II `EFI_SMBIOS_PROTOCOL`

Some clarification would be helpful.

docs/src/rfc/text/0014-smbios.md
AnhDLuong: I would like to ask another question - can we not use efi::Status as the Error (I understand efi::Status has success and other errors that Smbios does not need) instead of creating a new enum SmbiosError, and then need to add a function to convert from SmbiosError to efi::Status ?

kat-perez: @AnhDLuong We want to use custom error types here so that we can have more precise error handling. We should be using EFI status codes for functions that use `efiapi`.

docs/src/rfc/text/0014-smbios.md
kat-perez: I meant that the "Rust-based SMBIOS implementation" which is what this RFC for is based on the existing UEFI SmbiosProtocol mentioned in the Protocol section. Should I word this differently?

makubacki: I suggest "This Patina-based SMBIOS implementation...".

kat-perez: updated

docs/src/rfc/text/0014-smbios.md
kat-perez: updated to more specific error names

docs/src/rfc/text/0014-smbios.md
kat-perez: Makes sense, updating SmbiosProvider to SmbiosRecords

docs/src/rfc/text/0014-smbios.md
kat-perez: decided to include the SmbiosInit functionality into SmbiosManager instead

docs/src/rfc/text/0014-smbios.md
```diff
@@ -0,0 +1,1163 @@
+# RFC: `SMBIOS`
+
+This RFC proposes a Rust-based interface for SMBIOS table management providing the functionality described in the
+`EFI_SMBIOS_PROTOCOL` service defined in the UEFI Specification.
+It introduces an SmbiosRecords trait to manage addition, removal, updates, and retrieval of SMBIOS records,
+encapsulating the SMBIOS protocol behind a safe, idiomatic Rust API.
+The goal is to maintain compatibility with the UEFI specification
+while improving memory safety and simplifying integration for components.
```
joschock: nitpick: line breaks seem inconsistent.

docs/src/rfc/text/0014-smbios.md
```diff
@@ -0,0 +1,1163 @@
...
+1. The API should provide all necessary SMBIOS functionality as a service to components
+2. The API should utilize Rust best practices, particularly memory safety and error handling
+3. The SMBIOS service should produce protocols equivalent to the current C implementations, preserving existing C functionality
+4. Support both SMBIOS 2.x (32-bit) and SMBIOS 3.x (64-bit) table formats
+5. Provide safe string manipulation for SMBIOS records
+
+## Unresolved Questions
+
+ Should the API expose separate interfaces for 32-bit and 64-bit SMBIOS tables, or handle this internally?
+  Summary: Prefer a unified public API that handles 32/64 internally, plus an opt-in low-level escape hatch for experts.
```
joschock: would love a discussion about simply dropping support for the 32-bit entry point.

docs/src/rfc/text/0014-smbios.md
```diff
@@ -0,0 +1,1163 @@
...
+  - Easier to evolve for future SMBIOS updates.
+  Cons:
+  - More internal complexity and runtime branching.
+  - Advanced users lose direct control without escape hatches.
+  - Version-specific constraints can be hidden/surprising.
+  - Requires thorough tests to catch format-specific bugs.
+
+  Recommendation
+  Patina exposes a single, ergonomic SMBIOS API that’s format-agnostic by default. The SmbiosRecords service lets components add, update, remove, and enumerate records without choosing between SMBIOS 2.x (32-bit) and 3.x (64-bit). The service selects and publishes the appropriate tables per platform and policy. For advanced firmware tooling, an opt-in “raw” module provides explicit 32/64 entry-point construction and publishing primitives, so experts retain full control when needed.
+
```
joschock: Agree with this recommendation in general; the API exposed to other services should hide and make opaque the details of table construction. IMHO, the goal should be a safe, idiomatic, opaque SMBIOS service that can be used for "normal" usage. The "raw" module muddies the waters a bit, since it opens the door to the internals. If such a thing is needed, I'd recommend doing it separately from the `Service` being defined here.

docs/src/rfc/text/0014-smbios.md
```diff
@@ -0,0 +1,1163 @@
...
+  Tiny API sketch
+  ```rust
+  /// Optional hint when you need to force a target format.
```
joschock: see comment above; I'd love to see us start with a simple API that hides the target knowledge. If it turns out to be required, then we can always make an extension to the API to add targeting.

docs/src/rfc/text/0014-smbios.md
```diff
@@ -0,0 +1,1163 @@
...
+          /// Build a 64-bit SMBIOS entry-point from a serialized table buffer.
+          pub fn build_entry_point64(_table: &[u8]) -> SmbiosEntryPoint64 { unimplemented!() }
+
+          /// Publish entry-points to the system configuration table (unsafe by nature).
+          pub unsafe fn install_config_table_32(_ep: &SmbiosEntryPoint32) -> Result<(), ()> { unimplemented!() }
+          pub unsafe fn install_config_table_64(_ep: &SmbiosEntryPoint64) -> Result<(), ()> { unimplemented!() }
+      }
+  }
+  ```
+- Is there value in exposing lower-level table construction functionality to advanced users?
```
joschock: my opinion on this is: "don't do it until it is demanded"

docs/src/rfc/text/0014-smbios.md
```diff
@@ -0,0 +1,1163 @@
...
+- Should we provide typed interfaces for specific SMBIOS record types (Type 0, Type 1, etc.)?
```
joschock: I think it would be useful to provide typed records. However, you have to deal somehow with the fact that each new SMBIOS spec updates some of the types - so, for example, SMBIOS 3.0 Type0 is different (has fewer fields) than SMBIOS 3.1 Type0. The implementation would also need to take care to be forward-compatible - i.e. if you get a Type0 with more  data than you were expecting, it may be that the implementation is out of date.

docs/src/rfc/text/0014-smbios.md
```diff
@@ -0,0 +1,1163 @@
...
+
+In C, `SMBIOS_INSTANCE` provides the core management structure,
+`EFI_SMBIOS_ENTRY` represents individual SMBIOS records,
+and `SMBIOS_HANDLE_ENTRY` tracks allocated handles.
+These are roughly replicated by the Rust structs described in [SMBIOS Records](#smbios-records).
+
+### Dependencies on C Protocols
+
+While the final outcome should be a purely Rust-based interface,
+a Rust implementation of SMBIOS services currently relies on C protocols like `BootServices.InstallConfigurationTable`
```
joschock: FWIW, I think BootServices is now available in patina as a rust component/service.

docs/src/rfc/text/0014-smbios.md
```diff
@@ -0,0 +1,1163 @@
...
+    type Iter: Iterator<Item = &SmbiosRecord>;
+
+    /// Adds an SMBIOS record to the SMBIOS table.
+    /// If handle is SMBIOS_HANDLE_PI_RESERVED (0xFFFE), a unique handle will be assigned.
+    fn add(
+        &mut self,
+        producer_handle: Option<Handle>,
+        smbios_handle: &mut SmbiosHandle,
+        record: &SmbiosTableHeader,
+    ) -> Result<(), SmbiosError>;
```
joschock: consider taking the smbios_handle as an `Option` and returning it in the result
```suggestion
    fn add(
        &mut self,
        producer_handle: Option<Handle>,
        smbios_handle: Option<SmbiosHandle>,
        record: &SmbiosTableHeader,
    ) -> Result<SmbiosHandle, SmbiosError>;
```

joschock: (instead of using 0xFFFE as sentinel value, I mean)

docs/src/rfc/text/0014-smbios.md
```diff
@@ -0,0 +1,1163 @@
...
+        &mut self,
+        smbios_handle: SmbiosHandle,
+        string_number: usize,
+        string: &str,
+    ) -> Result<(), SmbiosError>;
+
+    /// Removes an SMBIOS record from the SMBIOS table.
+    fn remove(&mut self, smbios_handle: SmbiosHandle) -> Result<(), SmbiosError>;
+
+    /// Discovers SMBIOS records, optionally filtered by type.
```
joschock: This seems redundant with the iterator.

docs/src/rfc/text/0014-smbios.md
```diff
@@ -0,0 +1,1163 @@
...
+pub struct SmbiosManager {
+    records: Vec<SmbiosRecord>,
+    allocated_handles: HashSet<SmbiosHandle>,
+    major_version: u8,
+    minor_version: u8,
+    entry_point_32: Option<Box<SmbiosEntryPoint>>,
+    entry_point_64: Option<Box<Smbios30EntryPoint>>,
+    table_32_address: Option<PhysicalAddress>,
+    table_64_address: Option<PhysicalAddress>,
+    lock: Mutex<()>,
```
joschock: if a mutex is part of the design, that probably merits some discussion above.

docs/src/rfc/text/0014-smbios.md
```diff
@@ -0,0 +1,1163 @@
...
+                record.push(0); // Null terminator
+            }
+            record.push(0); // Double null terminator
+        }
+        
+        Ok(record)
+    }
+
+    fn allocate_handle(&mut self) -> Result<SmbiosHandle, SmbiosError> {
+        for handle in 1..0xFF00 {
```
joschock: recommend just maintaining a counter for this rather than searching the hashset each time.

docs/src/rfc/text/0014-smbios.md
```diff
@@ -0,0 +1,1163 @@
...
+
+```rust
+/// SMBIOS table header structure
+#[repr(C, packed)]
+#[derive(Debug, Clone)]
+pub struct SmbiosTableHeader {
+    pub record_type: u8,
+    pub length: u8,
+    pub handle: SmbiosHandle,
+    // Variable-length data follows
```
joschock: comment: variable length data on the struct is not very rusty.

docs/src/rfc/text/0014-smbios.md
```diff
@@ -0,0 +1,1163 @@
...
+    }
+    
+        // Remove the record
+    smbios_records.remove(handle)?;
+    Ok(())
+}
+```
+
+When adding a record with `add`, if the handle is set to `SMBIOS_HANDLE_PI_RESERVED`,
+a unique handle will be automatically assigned and returned via the mutable reference.
```
joschock: recommend (as noted elsewhere) not using the "reserved handle" as a sentinel value, but instead explicitly use Option or similar to indicate that a new handle is desired. Does it ever make sense to do an `add` on an existing handle? If not, consider omitting the argument entirely

docs/src/rfc/text/0014-smbios.md
```diff
@@ -0,0 +1,1163 @@
...
+
+        // Build the complete record with strings
+        let strings = vec![
+            config.vendor.as_str(),
+            config.version.as_str(),
+            config.release_date.as_str(),
+        ];
+
+        let record_data = SmbiosManager::build_record_with_strings(
+            &bios_info.header,
```
joschock: this is a bit dangerous, since the call expects that &header as the full structure data that follows it (i.e. is a header followed by data). This is giving up some memory safety and should be marked unsafe, probably. Consider this buggy code:

```
let bogus_header:SmbiosTableHeader  = SmbiosTableHeader  { length: 0x1234, /*other fields*/};
let record_data = SmbiosManager::build_record_with_strings (&bogus_header, strings); //compiler will allow this call without safety check, but will over-index because there is nothing after bogus_header.
```

docs/src/rfc/text/0014-smbios.md
```diff
@@ -0,0 +1,1163 @@
...
+- Implement pure Rust configuration table installation
+- Complete migration of all SMBIOS-related components
+
+## Performance Considerations
+
+### Memory Efficiency
+
+- Rust's zero-cost abstractions provide equivalent performance to C
+- Smart pointer usage minimizes memory allocation overhead
+- Compile-time optimization of record structure layouts
```
joschock: you probably don't want the record structure layouts to be optimized, unless they are intermediate representations not connected to the actual spec record layouts.

docs/src/rfc/text/0014-smbios.md
rogurr: It appears version 3.x of the spec (64-bit support) was primarily designed to allow the UEFI configuration table to point to the entry point structure anywhere in addressable space and allow all structures pointed to by the entry point to reside in memory above 4GB.

This differs from version 2.x of the spec (32-bit) that expects the entry point structure to be paragraph aligned in Segment 0xF000 (below 1MB) and the structure array being limited to RAM below 4GB.

Since version 3.0 was designed to support UEFI and 3.0 can be configured to support the 1MB and 4GB limits, I would vote we don't support anything below 3.0, stick with only the 64-bit entry point structure.

docs/src/rfc/text/0014-smbios.md
rogurr: Since this is an RFC, throwing in my $0.02.  I would vote for no access to lower-level construction to make sure the code is secure.

Since the entry point structure is only 24 bytes long and populated with specific data representing the code in this crate such as spec version supported, table address, etc., there should be no reason to change this code.  As for how the tables are created, there is a clearly defined 4-byte header that indicates the length of the fixed size values that is followed by a double NULL terminated string pool that should not be changed if they are following the specification.

If an OEM does need to divert from the specification, they can always create a local override of the crate.

docs/src/rfc/text/0014-smbios.md
rogurr: I added a comment to Michael's first comment above, can we dog-pile there?

docs/src/rfc/text/0014-smbios.md
rogurr: Instead of typed interfaces, what if we use a byte array for the interfaces and provide sized structures for spec defined records?

For instance if a user wanted to add a type 0 record, use the defined structure you have below that they instantiate:
``` rust
pub struct type0_platform_firmware_information{
    header: SmbiosHeader,
    vendor: u8,
    firmware_version: u8,
    bios_starting_address_segment: u16,
    firmware_release_date: u8,
    firmware_rom_size: u8,
    // ... other fields defined in DSP0134_3.9.0.pdf
    string_pool: Vec<String>,
}
```

Then they send that structure as a byte array to the trait add function:
``` rust
fn add(
    &mut self,
    record_data: &[u8],
) -> Result<(), SmbiosError>;
```

And in the function implementation it would confirm:
* record_data is at least 4 bytes so it can evaluate the header
* The header->length is <= (record_data.length - 2) to confirm the string pool can be present
* There are only single nulls in the string pool area
* A double NULL resides at the end of record_data

This would then give the OEMs examples of how they can define their private records (ID 0x80 to 0xFE) and still use the generic add function.

docs/src/rfc/text/0014-smbios.md
```diff
@@ -0,0 +1,1163 @@
...
+    pub doc_rev: u8,
+    pub revision: u8,
+    pub reserved: u8,
+    pub table_max_size: u32,
+    pub table_address: u64,
+}
+
+/// BIOS Information (Type 0) - typed interface
+#[repr(C, packed)]
+pub struct BiosInformation {
```
rogurr: I owned the SMBIOS spec in a previous life and it was about 50/50 on how people refer to these records.  Can we name the structures by using both the type ID and the exact name as it is written in the spec?

[DSP0134_3.9.0.pdf](https://www.dmtf.org/sites/default/files/standards/documents/DSP0134_3.9.0.pdf)

For instance, use something like "type0_platform_firmware_information"?

docs/src/rfc/text/0014-smbios.md
joschock: How does `type0_platform_firmware_information` become a `&[u8]` - are you proposing some kind of serialization function?

docs/src/rfc/text/0014-smbios.md
kat-perez: updated

docs/src/rfc/text/0014-smbios.md
kat-perez: updated

docs/src/rfc/text/0014-smbios.md
kat-perez: updated

docs/src/rfc/text/0014-smbios.md
kat-perez: updated

docs/src/rfc/text/0014-smbios.md
rogurr: This comment is mostly focused on the SMBIOS component functions (Add, Remove, and GetNext).  Since we must support generic passing of data to support OEM defined records, I thought why not use super generic functions that work with a sized byte array buffer and remove any knowledge of specific SMBIOS records.

I wanted to give OEMs the option to create a SMBIOS table using something as simple as feeding in a hard coded byte array, moving up to our defined record modules, or creating a super complex module with reporting functions, update functions, etc.

When we write the record modules defined in the spec, we should adhere to some commonality but maybe not rely on the SMBIOS component.  We already have the struct defined, so I thought to just add an export function to all of them.

```rust
pub struct type0_platform_firmware_information {
    header: SmbiosHeader,
    vendor: u8,
    firmware_version: u8,
    ...
    string_pool: Vec<String>,
}

impl type0_platform_firmware_information {
    fn export(&self) -> Vec<u8> {
        ...
    }
```

I waffled back and forth between passing around a Vec<u8> or something like this:
``` rust
pub struct smbios_record {
    header: SmbiosHeader,
    pub record_data: Vec<u8>,
}
```

But trying to confine things always seems to make it more complicated to understand or bite us later.  So I stuck with the Vec<u8> and for the GetNext( ) function return a generic record module:
```rust
pub struct type_n_generic_record {
    header: SmbiosHeader,
    pub record_data: Vec<u8>,
}

impl type_n_generic_record {
    fn export(&self) -> Vec<u8> {
        ...
    }
```

docs/src/rfc/text/0014-smbios.md
rogurr: The SMBIOS spec has only a few specific requirements.  The entry point structure must be present, and each record in the table must have a header (ID, length, handle) and a double null after the string pool.  Since those are pretty minimal and work for all records, I would agree with John and keep it out of the module.

docs/src/rfc/text/0014-smbios.md
rogurr: Can we separate all records away from the SMBIOS component?  The spec was written so the entire table can be parsed and examined in binary format without any knowledge of each record.  So maybe we can follow that in our SMBIOS component?

Then maybe create a file for each record type and have specific new functions to return a different struct for each version of the spec that changes the fixed data size?

docs/src/rfc/text/0014-smbios.md
rogurr: I just got to the part where you defined the iterator.  I really like that even though it would probably make passing the Vec<u8> a bad idea.  I see you have the SmbiosRecord already includes data:Vec<u8> for the complete record.  I think that would work well for the 'export' functions in each record module instead of a Vec<u8>.

docs/src/rfc/text/0016-smbios.md
```diff
@@ -0,0 +1,1328 @@
...
+        }
+
+        let mut record_header = *header;
+        record_header.handle = smbios_handle;
+
+        let smbios_record = SmbiosRecord {
+            header: record_header,
+            producer_handle,
+            data,
+            string_count: 0, // Would be calculated from actual strings
```
AnhDLuong: Can you please elaborate how do you calculate the string_count ?

docs/src/rfc/text/0016-smbios.md
```diff
@@ -0,0 +1,1328 @@
...
+**Safe alternatives:**
+
+- **`add_from_bytes()`**: Takes the complete record as a byte slice, avoiding unsafe pointer arithmetic
+- Create the complete record data as a byte vector before passing to the API
+
+**Example of unsafe usage that could cause memory corruption:**
+
+```rust
+// DANGEROUS: This only creates a header, not a complete record
+let bogus_header = SmbiosTableHeader { length: 0x1234, /* other fields */ };
```
AnhDLuong: In my opinion, if the length is passed in (as part of the SmbiosTableHeader) and is corrupted, there is no safe way to parse the record. Sounds like the only way to make it safe is the patina-smbios has to construct the SmbiosTableHeader itself.

docs/src/rfc/text/0014-smbios.md
kat-perez: updated RFC to address SMBIOS 3.0+ support

docs/src/rfc/text/0014-smbios.md
kat-perez: have added a suggestion for serialization to the latest update of this RFC

docs/src/rfc/text/0014-smbios.md
kat-perez: removed

docs/src/rfc/text/0014-smbios.md
kat-perez: added a design suggestion for typed records in the RFC

docs/src/rfc/text/0014-smbios.md
kat-perez: updated to counter solution

docs/src/rfc/text/0016-smbios.md
kat-perez: new function added called `count_strings_in_record`

docs/src/rfc/text/0016-smbios.md
kat-perez: the safe pattern is listed below this example in the RFC. check the function `add_from_bytes` to see safe usage. lemme know if that still could potentially corrupt memory

### Recommended Safe Pattern

```rust
// SAFE: Create complete record as validated bytes first
let mut record_bytes = Vec::new();
record_bytes.extend_from_slice(&header_bytes);
record_bytes.extend_from_slice(&structured_data_bytes);
let handle = smbios.add_from_bytes(None, &record_bytes)?;
```

docs/src/rfc/text/0014-smbios.md
kat-perez: proposed a safer method with `add_from_bytes` function in the updated RFC

docs/src/rfc/text/0014-smbios.md
kat-perez: updated the RFC with an example how we can serialize

docs/src/rfc/text/0016-smbios.md
kat-perez: updated with function for counting the strings

docs/src/rfc/text/0016-smbios.md
kat-perez: also have provided a method to parse the record

docs/src/rfc/text/0014-smbios.md
kat-perez: updated

docs/src/rfc/text/0018-smbios.md
```diff
@@ -0,0 +1,3781 @@
+# RFC 0018: `SMBIOS`
+
+This RFC proposes a Rust-based interface for SMBIOS table management, providing the functionality described in the
+`EFI_SMBIOS_PROTOCOL` service defined in the UEFI Specification.
+
+It introduces an `SmbiosRecords` trait to manage addition, removal, updates, and retrieval of SMBIOS records,
+encapsulating the SMBIOS protocol behind a safe, idiomatic Rust API.
+
+The goal is to maintain compatibility with the UEFI specification while improving memory safety and simplifying
```
makubacki: "maintain compatibility with the UEFI specification" leaves a bit of room for interpretation of what "compatibility" means. The primary goal, as I understand, is to preserve semantics of SMBIOS support in the UEFI Specification but change the interfaces for Rust integration. Since this is the intro paragraph, I'd call that out a little bit more here to set expectations.

docs/src/rfc/text/0018-smbios.md
```diff
@@ -0,0 +1,3781 @@
...
+    pub extended_bios_size: Option<u16>,
+    
+    // Track which version this was parsed with
+    parsed_version: (u8, u8),
+    
+    // Preserve unknown future fields
+    unknown_data: Vec<u8>,
+}
+
+impl VersionedSmbiosRecord for BiosInformation {
```
makubacki: I don't think you need implementation in the RFC. Mostly just public structures and interfaces.

makubacki: It makes the RFC more difficult to read and maintain and is more likely to drift from the actual code implementation.

docs/src/rfc/text/0018-smbios.md
```diff
@@ -0,0 +1,3781 @@
...
+#### Performance Characteristics
+
+```rust
+// The generic serializer is zero-cost at runtime:
+// - Field layouts are computed at compile time
+// - No dynamic dispatch or reflection
+// - Unsafe code is isolated and well-tested
+// - Memory layout matches exactly what manual code would produce
+
+#[cfg(test)]
```
makubacki: Benchmarking code like this can also be taken out and moved to a benchmark file.

docs/src/rfc/text/0018-smbios.md
```diff
@@ -0,0 +1,3781 @@
...
+0x17   | FF                             | EC Minor (not supported)
+0x18   | 00 00                          | Extended ROM size
+-------|--------------------------------|------------------------
+0x1A   | 50 61 74 69 6E 61 20 46 69... | "Patina Firmware Corp\0"
+       | 76 32 2E 35 2E 31 2D 70 61... | "v2.5.1-patina\0"  
+       | 30 39 2F 31 38 2F 32 30 32... | "09/18/2025\0"
+       | 00                             | Double null terminator
+```
+
+#### Validation During add_from_bytes()
```
makubacki: Sections like this should also be removed from the RFC.

docs/src/rfc/text/0018-smbios.md
makubacki, line 1: @kat-perez, general feedback is to please reduce the verbosity of the RFC. I'll revisit after that is done.

- Remove any validation strategy unless it is significant to the design
- Remove any non-public interfaces, including structures, functions, implementation blocks, and insignificant constants
- Remove any "peripheral" code like benchmarking examples

Clean up any explanations about the kind of content.

I suggest moving most of the public structures and important traits to a "code appendix" and refer to that when needed. Keeping the source code relatively light and focused in the design sections. That will allow the primary and alternate designs to be easier to mentally parse and compare.

makubacki, line 1: nit: The RFC number should remain "0000" until the RFC is ready to be merged.

docs/src/rfc/text/0018-smbios.md
kat-perez: makes sense. slimming it down

kat-perez:
@makubacki I have greatly reduced the verbosity in this RFC and I also removed all the sections you noted. Any important public APIs are now moved to an Appendix. Thanks for the review and let me know if you have any further suggestions!

docs/src/rfc/text/0000-smbios.md
```diff
@@ -0,0 +1,467 @@
...
+- Versioned typed record interfaces
+- Typed record methods (`add_typed_record()`, `get_typed_record()`, `iter_typed_records()`)
+- Forward/backward compatibility with version-aware parsing
+- SMBIOS protocol FFI bindings (C ABI extern functions)
+
+These features are described in Future Work sections below.
+
+## Future Work
+
+The current byte-based API is production-ready. Potential future enhancements include:
```
makubacki: Missing the list.

```suggestion
The current byte-based API is production-ready. Potential future enhancements include:
```

kat-perez: added

docs/src/rfc/text/0014-smbios.md
kat-perez: i've left the implementation out of the RFC but I am following that convention in the latest updates I have coming in for the patina_smbios component. will look something like this:

```
pub struct Type0PlatformFirmwareInformation {
    pub header: SmbiosTableHeader,
    pub vendor: u8,           // String index
    pub firmware_version: u8, // String index
    pub bios_starting_address_segment: u16,
    pub firmware_release_date: u8, // String index
    pub firmware_rom_size: u8,
    pub characteristics: u64,
    pub characteristics_ext1: u8,
    pub characteristics_ext2: u8,
    pub system_bios_major_release: u8,
    pub system_bios_minor_release: u8,
    pub embedded_controller_major_release: u8,
    pub embedded_controller_minor_release: u8,
    pub extended_bios_rom_size: u16,

    /// String pool containing the actual string content.
    ///
    /// **IMPORTANT**: This field is NOT part of the SMBIOS table binary layout.
    /// It is Rust metadata that gets converted to null-terminated bytes during serialization.
    /// Never attempt to directly cast this struct to bytes or use it in FFI - always use
    /// `SmbiosSerializer::serialize()`.
    pub string_pool: Vec<String>,
}
```

PR #628, opened 2025-07-28
RFC: Patina release process [FCP]

### LINKED ISSUE #594: [Task]: Create "Release Strategy" RFC
### Request Description

Patina consumers in production need a reliable way to get hot fixes and other changes on top of previously released Patina versions. This issue tracks an RFC explaining the release process that will accommodate this.

For example, "a release branch" strategy will likely be a part of the RFC.

I will make the change

No maintainer feedback needed

### Anything else?

_No response_

makubacki:
## Description

Closes #594 

An RFC to define the Patina release process.

Status: Final

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [x] Includes documentation?

## How This Was Tested

N/A

## Integration Instructions

N/A


docs/src/rfc/text/0000-patina-release-process.md
```diff
@@ -0,0 +1,171 @@
...
+## Prior Art (Existing PI C Implementation)
+
+Various projects have incorporated semantic versioning in their release processes. This process is largely based on
+the Semantic Versioning 2.0.0 specification. In terms of related open source firmware project, Tianocore follows a
+[stable tag](https://github.com/tianocore/tianocore.github.io/wiki/EDK-II-Release-Planning) process that releases on
+a quarterly interval with freeze periods preceding the release point. [Project Mu](https://microsoft.github.io/mu/) has
+an [automated release note process](https://github.com/microsoft/mu_devops?tab=readme-ov-file#release-drafting) that
+portions of this proposal are based on.
+
+## Proposed Release Process
```
cfernald: I think we need to address Major releases more explicitly. Specifically, I think we need to be able to support the current Major version while also leaving room for making breaking changes pending the next major release. We may want to begin making breaking or interface changes well before we intend on releasing them. Now I can see a couple of ways to achieve this but would like to get your thoughts.

1. Fork the previous major version and main is permanently a pre-release version of the next major release using the prerelease tagging with a counter indicating the prerelease version.

image

2. We leverage a automatic swap of interfaces for the next major version swap using a "unstable" and "deprecated" swap at the major version. That is, make a duplicated interface that leverages something like the unstable declaration to make that compile time removed for non-interested consumers but that swaps at the next major release. This is a higher code complexity cost, but requires a less divergent code tree.

docs/src/rfc/text/0000-patina-release-process.md
```diff
@@ -0,0 +1,171 @@
...
+
+Patina makes official releases on as the project determines necessary. This RFC covers three types of releases:
```
cfernald: Do we need to address the expected versioning of the different crates in the repo? e.g. are SDK, core, and core crates versioned in lock step? How about Components? etc.

makubacki: I should have put that under "Unresolved Questions". It's within the scope of this PR, so I'm fine to add it. Right now, the RFC assumes the current release model is followed - all repo crate versions are updated together.

In general, I'm in favor of individual crate versions to track their changes more precisely, but then all those individual crates need to be reconciled against the GitHub repo version and release notes.

A way to approach that would be to continue with automated release notes and Patina repo versioning, but Patina repo versions are made up of crate updates. For example:

- Patina Repo Release: `v0.0.1`
  - `patina_dxe_core v0.0.1 -> v0.0.2`
  - `patina_sdk v0.2.0 -> 0.2.1`
- Patina Repo Release: `v0.1.0`
  - `patina_dxe_core v0.0.1 -> v0.1.0`
  - `patina_stacktrace v0.0.2 -> 0.0.3`

There the "Patina Repo version" tracks all updates in the repo where its version is INCREMENT_VERSION_COMPONENT_MAX(crate versions). The version element in the repo version is updated by the maximum version component updated in the crates in the release. The repo release notes could be broken up into sections for individual crate updates.

---

Does that sound reasonable as a starting point for the RFC? Other suggestions?

makubacki: I'm also not entirely convinced we need to change this right now. I'm interested to see what others think.

cfernald: My opinion would be that all of the "core" crates should be versioned together and the semantic versioning should reflect the collective consumer facing interfaces. So that is SDK, DXE-Core, and Core/* should be versioned together. Otherwise, I feel like it will be difficult to track the inter dependencies between the crates. Specifically, because the core crates will have breaking dependencies between them such as:
1. SDK introducing a new interface that the core has to implement
2. a patina_internal_* crate changes an interface used by dxe-core

This will mean for practical reasons the core crates will all have a specific version requirement for the sub crates unless we are constantly creating breaking changes. With this in mind it seems simpiler to just track one semantic verison and have them all be in lock step instead of the complexity of tracking all the version that _sometimes_ have to move in lock step. Long winded way of saying, I dont see a lot of benefit of splitting them up but I see a lot of complexity/cost.

Now components are different. Component by their nature should be treated as external to our public API. For this reason I could definitely see the components being tracked with their own individual versions. Though I don't know how we do with will within the monorepo honestly.

cfernald: To your point, i don't think we need to change this but I think we should make a statement about the expectation since I don't believe we have currently addressed it in an RFC.

docs/src/rfc/text/0000-patina-release-process.md
```diff
@@ -0,0 +1,171 @@
...
+
+  In this case, a new branch is created from `base change` with the minimum commits needed to include `Exact Changes
+  Needed`. The branch follows the naming convention `hotpatch/<Patina Base Version>/<Hot Fix Number>`. Hot fix crate
+  versions are versioned as `<Patina Base Version>-hotfix.<Hot Fix Number>`. For example, if a hot fix is requested on
+  Patina v4.1.0 and it is the second hot fix requested on that base version, the following branches will exist with
+  the following versions produced from those branches:
+
+  - Request 1:
+    - Branch: `hotpatch/4.1.0/1`
+    - Version: `4.1.0-hotfix.1`
```
cfernald: To be pedantic, wouldn't `-hotfix.1` indicate a "prelease" version in semantic versioning?  Why not just use the next available patch version? Is the concern that a platform may want to use 1.0.1 with a fix without going to 1.0.3?

docs/src/rfc/text/0000-patina-release-process.md
```diff
@@ -0,0 +1,171 @@
...
+  Friday at 11am for a release on Friday. At the earliest, the release will be made on Monday by 5pm. All times are
+  based on Pacific time unless noted otherwise. The Patina team will acknowledge the expected release date in the
+  discussion section of the work item request. Note that the requested date may not be feasible, try to plan a
+  special release request as early as possible.
+
+  Please consider if you can simply take the latest code as an early release and avoid a hot fix.
+
+  ## Hot Fix Process
+
+  Hot fixes are made when an out-of-cycle release is requested on a `base change` in which the `main` branch already
```
cfernald: Is this the main avenue of providing fixes for existing releases? 

If I understand this correctly, I worry about the support cost and the problem of missing fixes. My concern is that lets say we have 3 platforms on 3 releases, 1.0.1, 1.0.2, 1.0.3. We have to then know to fix all of these versions with their appropriate "-hotfix.N" release. Additionally, not all platforms may be aware they even need such a fix. While I think there will always be space for specific requesting of things like patches, I wonder if we need a default avenue for supporting a previous release.

I wonder if instead we should attempt to automatically cherry-pick bug fixes back to N previous Major releases. I'm not a fan of having to do with for all possible minor versions though.

docs/src/rfc/text/0000-patina-release-process.md
makubacki: > My opinion would be that all of the "core" crates should be versioned together and the semantic versioning should reflect the collective consumer facing interfaces. So that is SDK, DXE-Core, and Core/* should be versioned together.

I see what you're saying from an interface dependency perspective but drawing lines around subsections of the repo creates complex fragmentation in overall version management and the release process. I'm not necessarily opposed to it or anything but I'd still want to maintain a consistent GitHub release process in the repo using something like I described above to capture updates in the overall repo.

I'm a little hesitant to make updates to this process at this time unless the value is clear. Process change like this come with a lot of tooling and communication overhead. I'll leave the thread open and add this to the RFC as an unresolved question.

docs/src/rfc/text/0000-patina-release-process.md
makubacki: > We have to then know to fix all of these versions with their appropriate "-hotfix.N" release.

What do you mean by this? It is the platform team's responsibility to say what they want cherry-picked in the GitHub issue. They would have to request all 3 in the hot fix request issue, or the hot fix is not made. We just perform the release for them.

---

For these:

> Additionally, not all platforms may be aware they even need such a fix.

> While I think there will always be space for specific requesting of things like patches, I wonder if we need a default avenue for supporting a previous release.

> I wonder if instead we should attempt to automatically cherry-pick bug fixes back to N previous Major releases. I'm not a fan of having to do with for all possible minor versions though.

I'm expecting the hot fix process to be a consumer-driven process. I don't think the Patina team has the resources to make forward progress while supporting too many combinations of changes on prior releases and managing automation policy and implementation.

A platform team needs to be aware of the Patina releases they have rejected to integrate and determine what is critical for their platform on their current release that justifies a hot fix release.

Automatic cherry-picks introduce a lot of additional support overhead with little value to the Patina project in addition to potential for merge conflicts. For example, if `5.0.1` is a bug fix for a bug introduced in `5.0.0` it will have a merge conflict being automatically cherry-picked to pre-`5.0.0` code. In addition, `N` may not be large enough and need a manual release anyway.

The hot fix request process in a GitHub issue provides an opportunity to clearly document the hot fix and discuss alternatives. For example, if a platform is on `1.2.0` and requests a hot fix from `1.2.3`, the answer is probably going to be to move to `1.2.3`.

docs/src/rfc/text/0000-patina-release-process.md
makubacki: Yes. That's why the guidance was to update to an exact version when using hot fixes.

> Is the concern that a platform may want to use 1.0.1 with a fix without going to 1.0.3?

Yes. And that independent requests may be made to hot fix `1.0.1`.

If we had a rule that hot fixes are only made on the latest patch version, we could do that but that implies a long-lived continuation of patches on that minor version that are common to all consumers of that minor version. The goal here is to get people what they critically need and get out of the way, not support a long-term maintenance branch of the `1.0` release.

docs/src/rfc/text/0000-patina-release-process.md
Javagedes: Having a pre-release branch for each major version makes sense to me. Then any `semver:major` could throw a CI check error if the target branch is not some `x.0.0-abc` format.

The one painpoint would be making sure branches stay compatible between say `1.0.1` and `2.0.0-abc` We could have CI always run on both, and only merge if both pass. And any `semver:minor` or `semver:patch` would be immediately ported to the `x.0.0-abc` branch.

I'd say this complexity would only be worth it if we strive to keep `semver:major` level changes to greater than every week. Maybe every month or something? Not sure. With our unstable feature "feature", that would be doable.

This would also only make sense if we kept all crates in the repo at the same version, however. If we don't, I'm not sure how we could do any of this.

docs/src/rfc/text/0000-patina-release-process.md
Javagedes: Having differently versioned crates in this repo is the most accurate way to do it, however it would add a lot of complexities when it comes to staging breaking changes, documenting releases on github, and handling the cascade of breaking changes between crates (It brings back, in a slightly less intrusive way, the same cascading version issues we were having when the crates were in separate repos).

docs/src/rfc/text/0000-patina-release-process.md
Javagedes: I think my main concern is if we are at 6.0.0 and a hotfix comes in for 5.0.3, we have to apply the hotfix to `5.0.4`, `5.1.0`, `5.1.2`, `5.1.3`, `5.2.0`, `6.0.0`, Otherwise we are now in a torn state where one very specific version is fixed, but anything else above that may not be fixed.

While it is a platform driven initiative, I feel like the right answer is to fix all releases after.

docs/src/rfc/text/0000-patina-release-process.md
os-d: I think that in combination with the idea that when open sourcing, all of our crates will reset to v0 and that we go through a stabilization period before the 1.0.0 release, that keeping crates in lockstep is probably okay.

The main downside of not doing independent crate versioning is that consumers of one crate will miss out on bugfixes if they don't take breaking changes, e.g. if they depend on patina_crate v0.* and some other patina crate takes a breaking change, so all patina crates go to v1.0.0, then patina_crate gets a critical bugfix, the consumer doesn't get the bug fix unless they upgrade all of patina.

Now, we can say this scenario falls under the purview of hotfix branches, which is fine, but then is more maintenance on our side for something that otherwise would require 0 maintenance. So, while I agree there are maintenance concerns with individual crate versioning, I think there is maintenance required without it, also. Or consumer pain.

I also think that the monorepo eases that pain, because we can use the repo Cargo.toml to control which version of other patina crates they support, right? Or maybe this is where you would use the crate level Cargo.toml? I'm not sure that's a good way to handle this, but something to think about anyway.

If we can find a compelling way to resolve the crate interdependency issue, my preference would be individual crate versioning. If not, I would still be in favor of Chris's suggestion that components are versioning independently and everything else is versioned together.

I do think this RFC should pick a current direction, since this will greatly affect our release process. It may well pick that we are versioning everything together and a future RFC can change that, but I do think we should document it.

docs/src/rfc/text/0000-patina-release-process.md
os-d: To me, this seems like something that we don't currently have a use case for and I would personally defer from this RFC, to be addressed in a future RFC if we come up with a use case. Right now, this RFC (and how we currently do it in both patina and mu) requires instant breaking change releases. edk2 does this slightly differently, though they don't have real releases, but the general idea is queuing up all changes until a big release.

I don't see an issue with continuing main as the development branch and forking release branches off of it. For any large scale interface changes, we would probably be using the unstable feature anyway, until such a time that it stabilizes, in which case we will make a major release. In the meantime, those can be minor version releases that are iterating on it.

I think in the early days of an open source patina we should be avoiding overly complex release patterns, because they will certainly change and the majority of our effects are going to be spent on integration and bug fixes/stabilization, and we won't want to spend time on release management.

These are just my thoughts, if we have a specific use case now, no complaints from me to make this more prescribed.

docs/src/rfc/text/0000-patina-release-process.md
makubacki: Why would it need to be applied against the others?

If a platform has fixed themselves onto `5.0.3` and made a request for a change to be cherry-picked (hot fix applied) onto that release, then:

1. A branch called `hotpatch/5.0.3/P` is created from the `5.0.3` tag where `P` is the hot patch request count on the `5.0.3` release.
2. The change requested in the hot fix issue is cherry-picked onto that branch.
3. The tip of that branch is released as `5.0.3-hotfix.1`

The hot fix process is for the case where a platform team has locked their software configuration for a production release but is impacted by a critical bug that needs to be fixed in that configuration. Typically, they want exactly what they ask for, not miscellaneous other changes bundled in. Even though those are meant to be helpful fixes, from their perspective they introduce opportunity for further regression from the validation they've already performed. This process is intended to make the intention of the request explicit and clear and only the exact work necessary for a consumer that is going to actually use the hot fix is performed.

docs/src/rfc/text/0000-patina-release-process.md
```diff
@@ -0,0 +1,175 @@
...
+
+## Proposed Release Process
+
+Patina makes official releases on as the project determines necessary. This RFC covers three types of releases:
+
+- **Regular Releases**: A release is made on Monday each week. If there are no changes, the release is not made. The
+  release process is automated including the release notes, versioning, and publishing to crates.io. However, a Patina
+  team member with proper access can always make a release at any time.
+  > Note: Monday was chosen as the provides the maximum amount of time for the Patina team to be available for feedback
+  and support of the release during the given work week.
```
os-d: nit: I think for all your notes you need to use `>` on each line of the note

e.g.
```suggestion
  > and support of the release during the given work week.

```

docs/src/rfc/text/0000-patina-release-process.md
```diff
@@ -0,0 +1,175 @@
...
+    - `Requested Delivery Date: <date>`
+      - Examples
+        - `<reason(s)>`: Needed to enable unblock feature a on x and y platforms
+        - `<change(s)>`: Include commit ae4b134 from the main branch
+        - `<date>`: 8/1/2025
+
+  `<date>` must be at least the next business day from the request date. For example, a request cannot be made on
+  Friday at 11am for a release on Friday. At the earliest, the release will be made on Monday by 5pm. All times are
+  based on Pacific time unless noted otherwise. The Patina team will acknowledge the expected release date in the
+  discussion section of the work item request. Note that the requested date may not be feasible, try to plan a
```
os-d:
```suggestion
  discussion section of the work item request. Note that the requested date may not be feasible; try to plan a

```

docs/src/rfc/text/0000-patina-release-process.md
os-d: I think that consumer driven is the right approach (and maybe in the future if the community is large enough there can be a separate backporting approval group who manages that, a la the stable trees in Linux). I'm personally not too worried about torn states in older releases, by definition they are always distinct from every other release in what they contain and the release notes can always be viewed and give consumers all the information they need.

I would propose that we only support hot fixing the last patch version in any given minor release, e.g. if we have released 5.1.5 then 5.2.0, that we only allow hot fixing 5.1.5, not 5.1.4 and previous. Any patch updates should be safe to take (and if not then they should be hot fixed anyway) and would prevent trying to hotfix 5.1.2 with the bug fixes in 5.1.3, 5.1.4, and 5.1.5. For the current release (5.2.0 in this example) we would just release a regular patch, not a hot fix and not until 5.3.0 (or higher) is released would we apply hot fixes to the 5.1.x series.

That being said, I'm also okay to start with the current proposal and if it becomes a maintenance burden, we can do an RFC to change to something like what I am proposing (or something else).

docs/src/rfc/text/0000-patina-release-process.md
os-d: As noted above I would advocate for only hot fixing the latest patch version. To the concern of having to support long term maintenance of 1.0, I agree we don't want to do that, but as the arbiters of releasing hot fixes, we can reject the request and say no, you need to move forward, that is unsupported. Which, that is probably worth calling out if there is a release window we support hot fixing and otherwise do not support. E.g. we hot fix N-5 major releases, whatever.

docs/src/rfc/text/0000-patina-release-process.md
makubacki: My main concern is (1) we get into a `release` vs `dev` branch split where the process becomes so complicated we're not following it correctly and consumers are not using it as expected (2) the process defers updating to the latest release.

That said, if we think this is worth trying, I'm fine to add it to the RFC.

docs/src/rfc/text/0000-patina-release-process.md
makubacki: Logically, only supporting the latest patch version in a given minor release makes sense. I was initially hesitant to require that given we might need to patch something several years old, but if we make that known up front, we can set the expectation and enforce it. I'll add that requirement.

docs/src/rfc/text/0000-patina-release-process.md
makubacki: I'm in favor of more granular crate versioning. But, I'd like the rules to be simple and extensible. If the rule is "everything but components has a single version" and "components are versioned independently", that's simple enough and scales across new crates being added over time.

However, I don't want to lose the benefit of GitHub releases - a repo release version and release notes that provide a clean release history in the repo and show up as a GitHub notification to those watching the repo. In that case, I'd lean toward the process I mentioned  in the first reply to this thread.

docs/src/rfc/text/0000-patina-release-process.md
makubacki, line 1: ### Major Version Criteria Thread

Something we discussed in today's meeting is clarifying the criteria for a major release to be made. For example, somthing like changing the type of a parameter in a public API might not justify a major version increment.

Can we use this thread to capture and close on what changes we'd like to impact the major version?

makubacki: @joschock, can you please continue your thoughts from the meeting here as a starting point?

docs/src/rfc/text/0000-patina-release-process.md
makubacki: Based on today's conversation outside of the RFC, I added a variation of this request to the RFC. I'm closing this thread to focus further conversation on the current content.

docs/src/rfc/text/0000-patina-release-process.md
makubacki: I updated the RFC with the decision to continue versioning all changes together and to reevaluate if that direction needs to change in the future.

docs/src/rfc/text/0000-patina-release-process.md
makubacki: It will be treated as the same block if there is not a line break.

docs/src/rfc/text/0000-patina-release-process.md
makubacki: I didn't entirely forbid a given version from being hot fixed. It is possible there will be scenarios in which something in a later version may not be acceptable to a platform. But it is now described as "strongly recommended" to only hot fix the latest patch version for a given major release and a discussion should be had at the time of the request about why that is not possible if a hot fix is requested on a different version.

The current crate naming convention is retained instead of continuing to version items for hot fixes as hot fixes are intended to only target a specific release in special circumstances. The use case of continuing to iterate on a previous release is available for other use cases.

docs/src/rfc/text/0000-patina-release-process.md
makubacki: I didn't entirely forbid a given version from being hot fixed. It is possible there will be scenarios in which something in a later version may not be acceptable to a platform. But it is now described as "strongly recommended" to only hot fix the latest patch version for a given major release and a discussion should be had at the time of the request about why that is not possible if a hot fix is requested on a different version.

docs/src/rfc/text/0000-patina-release-process.md
joschock: To summarize: the semantics of versioning communicates information that is useful in scoping and planning. It's weird if integrating a version bump from `1.0.0` to `2.0.0` is half-an hour work because of a type change from `u32`->`usize` in one API, but `2.0.0`->`3.0.0` is a month's work of work because of a major rewrite of a large API surface. Strictly speaking, both are "breaking" changes in the semver sense, but it seems useful to differentiate between them somehow.

I see a few potential ways to approach this (there are probably more):
1) simple changes that are breaking (i.e. u32->usize) are simply deferred until such time as we are gearing up for a major release. Pros: adheres more closely to the spirit of semver. Proponents of Semver might ask: "do you really need to change that API from u32->usize - why not add a new API and deprecate the old one to be removed in future major release"? Cons: managing deprecated APIs/API bloat and or overhead of tracking in "minor" changes to make sure they intercept the next "major" release.
2) we diverge somewhat from semver and stipulate that "minor" changes to the API (e.g. in this example the API is semantically identical, but the type signature has changed) do not roll the major version and we communicate the expectation that integrating a minor version may require some integration work. We write down at least some rules of thumb to draw a line between what is "minor" and what is "major"
3) we adhere strictly to semver and also check in "minor" changes to the API as they come up. Integrators refer to release notes/other project resources to forecast volume of actual change and scope of integration work. Pros: most semver. Cons: major version increments more rapidly and doesn't provide a clean gauge of actual change. 

Option 1 seems okay to me, but does involve more project management-type work.

docs/src/rfc/text/0000-patina-release-process.md
```diff
@@ -0,0 +1,269 @@
...
+  patina = { version = "4.1.0-hotfix.2" }
+  ```
+
+  This ensures that the consumer is using the exact version of the crate that was released as a hot fix and avoids
+  potential issues with future releases that may not include the hot fix changes.
+
+## Open Source Versioning Plan
+
+When Patina is made open source, it will begin publishing to crates.io (instead of the `patina-fw` public registry).
+At that time, the first release will begin at version `0.1.0` for all Patina crates. Per the Semantic Versioning
```
rogurr: With all of Patina using the same version release number, do we have a plan to show what pieces are affected in a specific drop?

For instance, I understand the primary goal of Patina is a bootable UEFI and not a general collection of crates.  But if an OEM does find a good use for a component crate and doesn't yet need to ingest all of Patina, will there be an easy way to determine if a weekly drop has any changes in that specific component?

docs/src/rfc/text/0000-patina-release-process.md
os-d: I definitely agree with the need to define this and I tend to agree with John's overall sentiment that it feels odd to have a very small breaking change that requires essentially no integration work be the same as something major. Again to his point on UEFI 2.0 was very different from UEFI 1.0, but UEFI 2.11 is not that different than UEFI 2.10, but contains some breaking changes.

I think there is one additional consideration here: we have historically been making breaking changes for things that don't have consumers outside of patina. So I think another angle to think about is: if an internal only crate makes a breaking API change, do we increment the major version? I would say no, that an internal crate is not intended to be consumed externally and therefore is not a breaking change to the exposed API surface of Patina.

I personally would advocate for John's option 2. I think that is the easiest to track and gives consumers a realistic expectation on how much work consuming a new Patina* crate version is. If it is a minor version, you expect some level of small changes that don't require much integration work. If it is a major version update, then you expect to invest significant time to integrate.

This does violate the very basic rules of semver, but tbh I think it is more realistic and useful than the consumer having no idea if a version change indicates anything meaningful or not. If we want to say we are going to follow semver word for word, that's fine, I understand the sentiment, then I would agree with option 1. That puts more maintenance burden on us and more confusion for the consumer (which API version do I use, etc.), but lets us be semver compliant.

Option 3 is okay with me, too, it is the same very low amount of work as option 1, but does make the consumer story worse, imo. But if we say option 2 is unacceptable, there is an attractiveness to the lack of work required for option 3 as compared to option 1.

docs/src/rfc/text/0000-patina-release-process.md
os-d: Our GitHub release notes will be the source of truth for changes between versions (perhaps there is a way for crates.io to reflect this, I'm not familiar).

I do agree with Ron that with lockstepped crates, we should amend our release drafter to change the summary of commits to be per crate, if possible. That would help understand the scenarios we'll run into, like some crate not changing for 16 major versions, etc.

os-d: Will we start at `0.1.0` or `0.0.1`?

docs/src/rfc/text/0000-patina-release-process.md
```diff
@@ -0,0 +1,269 @@
...
+  - Any release made from a non-`main` branch must have a pre-release identifier in the version.
+- A `feature` branch may be created for an involved feature or enhancement.
+  - The `feature` branch is merged into the `main` branch when the feature is complete and ready for release.
+  - The `feature` branch is deleted after it is merged.
+  - **Branch name**: `feature/<feature-name>`
+  - **Version Format**: `<major>.<minor>.<patch>-feature<feature-name>.<feature_release_number>`
+    - **Example**: `4.1.0-featureabc.1`
+  - Changes must be submitted by PR, however, PR requirements may vary from those in the `main` branch. The
+    final feature merge to `main` will be subject to the normal `main` branch PR requirements.
+- Prior to an upcoming major releases, any breaking changes or other changes not considered stable for release
```
os-d: Is the intention that there will be one stabilization branch per major release?

This approach seems a bit different than what I recall us discussing in the Rust meeting. I thought the idea was to have `main` be the development branch and we were forking off branches that corresponded to major releases. It was very similar to the `dev`/`release` split.

That being said, I'm not opposed to this approach, a few things I want to clarify, though:
- As asked, is there one stabilization branch per major release? If we have multiple `stabilization` branches that could be okay, but I worry that folks will end up developing on different branches and the various `stabilization` branches will be out of date with each other and so when integrated into main be less stable than intended.
- One concern with this approach is that even with a single `stabilization` branch, I don't think we end up with a single development branch. Because if I have contributed features for a `stabilization` branch, chances are I am working there. If I haven't, I'll probably work on `main`. We could say `stabilization` is the development branch, though that's not much different from having `main` be the development branch, though maybe I like keeping main as the release branch more.
- Do `unstable` features needs be go into the `stabilization` branch as they are inherently not stable? Or are they candidates for `main` because they are gated?

As I ramble, I guess the proposal I have is to have a single `stabilization/<major-release-ver>` per major release and we use `main` as the primary development branch. When we are ready for a new major release, we merge in the `stabilization` branch and do the release. I think there is a conversation about do we cherry-pick changes from main to the `stabilization` branch (I think so, or else it is likely to be less stable).

This is an inverted `dev`/`release` split and there are many other ways to take this, some of which I started to type out, but ended up typing this version out as it was the shortest delta from what is written. I am open to other variations (or something else).

docs/src/rfc/text/0000-patina-release-process.md
```diff
@@ -0,0 +1,269 @@
...
+        - `<change(s)>`: Include commit ae4b134 from the main branch
+        - `<date>`: 8/1/2025
+
+  `<date>` must be at least the next business day from the request date. For example, a request cannot be made on
+  Friday at 11am for a release on Friday. At the earliest, the release will be made on Monday by 5pm. All times are
+  based on Pacific time unless noted otherwise. The Patina team will acknowledge the expected release date in the
+  discussion section of the work item request. Note that the requested date may not be feasible; try to plan a
+  special release request as early as possible.
+
+  Please consider if you can simply take the latest code as an early release and avoid a hot fix.
```
os-d:
```suggestion
  Please consider if you can simply take a newer release and avoid a hot fix.
```
nit: the language seemed a little odd here (I wasn't sure how it was an early release if the release was made some time ago)

docs/src/rfc/text/0000-patina-release-process.md
makubacki: I'm fine to explore option 2. I think strictly sticking to semver (Option 3) is more realistic with per-crate versioning where the blast radius of API changes is smaller and we've chosen not to do that.

Option 1 would require custom infrastructure to be put in place and process management I'd prefer to avoid.

Let's focus on the criteria for a "minor breaking change" versus "major breaking change" if there is no opposition to option 2.

docs/src/rfc/text/0000-patina-release-process.md
makubacki: The plan (mentioned somewhere else in this PR) is to have per-crate sections in the release notes. I'll check if the RFC needs to be updated to include that.

makubacki: This is following the semantic versioning FAQ guidance to start the initial development release at `0.1.0`.

docs/src/rfc/text/0000-patina-release-process.md
makubacki: `main` is the development branch in this proposal.

The detail that was not closed in the meeting was which branch releases are made from. This currently defines that as `main`.

This proposes a single `stabilization` branch per release. Multiple stabilization branches would be chaotic and unsustainable.

My understanding is that the `stabilization` branch discussed was that, a place for changes that may need to be deferred from release to stabilize.

If we want a simple release branch model with `release` not `stabilization` branches, I'm fine with that too. Simple to understand and follow. But, that requires that `main` be ready for release at anytime as opposed to changes being able to mature together in a single branch outside `main`.

docs/src/rfc/text/0000-patina-release-process.md
makubacki: This is in reference to the releases from `main` being made on a weekly cadence. The simpler language still makes sense for that.

docs/src/rfc/text/0000-patina-release-process.md
makubacki: Added a note to the crate versioning section.

docs/src/rfc/text/0000-patina-release-process.md
joschock: Here's a draft set of criteria (not exhaustive):
1. Changes or removals of APIs considered private to Patina (i.e. that are exposed from internal patina crates for use by other internal patina crates) are not considered breaking changes. <TODO: identify those APIs considered "public" and not exempt, and publish that list>
2. Changes to an API that change the API but preserve semantic behavior are not considered breaking changes. Examples: 
- type signature changes that improve correctness but preserve semantics ("u32->usize"), 
- addition of new parameters to an API for extended behavior with an option that allows preservation of previous behavior. ("do_something()->do_something(extended_behavior: bool)" where `extended_behavior=false` yeilds original semantics)
- API renames preserving behavior: ("do_soemthing() -> do_something()")
3. Addition of new public APIs is not considered a breaking change.
4. Removal of existing public APIs _is_ considered a breaking change.
5. Refactoring of a public API with significant semantic difference is considered a breaking change. For example: `do_something() -> do_something_with(complicated_to_construct: thing)` is breaking.

os-d: I agree with John's list. A few other things that come to mind:
6. Moving a component into the critical boot path is considered a breaking change. E.g.
  - Moving core functionality to a component that now requires that component for booting
  - Moving enough C functionality to a component such that the component is required for Patina to operate
7. The converse is also a breaking change: when a component is removed or a C driver is removed (e.g. when we deprecated CpuDxe) in such a way that running the component or C driver would cause Patina to fail to operate correctly
  - If the component/driver can still be run without affecting Patina's operation, this is not a breaking change.
8. C FFI layer exposed by Patina is always part of the "public" API set. It follows the same rules for breaking changes as Rust APIs (I thought about saying C FFI changes are always breaking changes, which for something coming from edk2 makes sense, but a patina-edk2 HOB changing or something doesn't necessarily mean we should do a breaking change; open for ideas here)
9. Refactoring the body of a public API function (e.g. usage remains the same but behavior is different) is:
  - Not a breaking change unless:
  - Calling this refactored API without changes from the caller will result in significant semantic differences (e.g. if Patina was being run with core_allocate_pages returning pages that are marked RWX and that changed to return pages that are marked RP)

docs/src/rfc/text/0000-patina-release-process.md
os-d: Yeah, all I meant was that, besides clarifying there is a single `stabilization` branch, is that in the current proposal, the stabilization branch name is said to be: `**Branch name**: `stabilization/<major>.<minor>`` and I'm wondering if we actually want this to just be `stabilization/<major>` and that stabilization just gets merged back when we are doing a new release? Otherwise we would potentially have new minor releases on a weekly cadence which would necessitate new stabilization branches.

But, I may not be following completely the goal here. My recollection of the conversation was to have a branch to defer breaking changes to so that we can coalesce them into a single version update. Which is what I thought was the `stabilization` branches purpose. But perhaps it is just for stabilization of features that may not be breaking changes?

makubacki: Your understanding of the branch's goal is the same as mine. I took that to mean:

1. Development occurs in `main`.
2. A change comes up that is not suitable for the upcoming release and it needs to be deferred while "stabilization" occurs.
3. The author of those changes creates a new `stabilization` branch if it does not already exist from the last release.
  3.1 Anyone may work in this `stabilization` branch as they need.
4. Release time comes for `main`.
  4.1. Authors of changes have to choose whether to move changes from the `stabilization` branch to `main`. If they do not their changes are not part of that release.
  4.2. That `stabilization` branch is locked.
5. `main` is released.

---

My original thought was that the `stabilization` branch is locked at the time of release. And the authors of changes there are forced to create a new one and migrate any changes from the previous `stabilization` branch necessary as an incentive to not sit in the `stabilization` branch forever and move their changes to the main release cycle.

---

I'm not opposed to other approaches though, this was just an interpretation of the discussion. I also realize that I didn't put some of this detail in the RFC that would have made the approach easer to follow.

os-d: Gotcha. I think that's reasonable. I do kind of like the idea of being able to only take some changes from `stabilization` to `main` and make a release, but I do wonder if that flow makes sense. I am seeing `stabilization` as more of a `pending_release` branch as opposed to actually stabilizing. For features that need true stabilization, I would imagine they would follow one of the two other paths: a `feature` branch or an `unstable` feature.

If we don't take all changes from `stabilization` into `main`, I'm not quite sure how `feature` branches and `unstable` features play into how a developer chooses where to target something. I was imagining (but open to changing that thought process) that we would merge `stabilization` as a whole into `main` when we are ready for a major version rolling, i.e. that everything in `stabilization` is actually stable and agreed upon. If it weren't, it wouldn't be included in `stabilization`.

The developer flow would then be (referring to below diagram):
- Start 1: If a large feature that requires coordinated development, create a `feature` branch and develop on it
- Start 2: If many changes are expected to the functionality being added but the change is complete or does not require coordination to develop, create it as an unstable feature (with the lines indicating whether to create it in `main` or `stabilization`.
- Start 3: If a breaking change is occurring that does not fall into one of the first categories, make it directly in `stabilization`
- Start 4: If not any of the other categories, go directly to `main`, do pass Go and do collect $200

image

I'm open to other versions of how the flow goes, but I guess it feels like it would be useful to have a (much better formatted) diagram documenting this, which doesn't per se have to be part of this RFC, but should be an outcome of this RFC.

docs/src/rfc/text/0000-patina-release-process.md
joschock: a couple of other examples for 9 would be changing the TPL level that a routine raises to or changing the "side" effects as it pertains to contents of protocol or event databases.

makubacki: Included custom criteria for determining what constitutes a major version update.

docs/src/rfc/text/0000-patina-release-process.md
makubacki: Simplified this down to a `major` branch to hold breaking changes until they are merged into `main` for a release.

docs/src/rfc/text/0000-patina-release-process.md
```diff
@@ -0,0 +1,360 @@
...
+### Criteria for Breaking Changes
+
+#### Non-Breaking Changes
+
+The following changes are **not** considered breaking:
+
+1. **Modifications to Private APIs**
+
+   Changes or removals of APIs considered private to Patina (i.e., exposed from internal Patina crates for use by other
+   internal crates) are not considered breaking changes.
```
os-d:
```suggestion
   Changes or removals of APIs considered private to Patina (i.e., exposed from Patina crates for use by other
  Patina crates) are not considered breaking changes.

```

nit: I thought using "internal" here may confuse the issue a little, i.e. only `patina_internal_*` crates can create or consume private APIs

docs/src/rfc/text/0000-patina-release-process.md
```diff
@@ -0,0 +1,360 @@
...
+    - Increment to Patch:
+      - 'impact:non-functional'
+      - 'semver:patch'
+      - 'type:bug'
+      - 'type:documentation'
+    - Default: Update patch
+
+### Release Process Details
+
+- The release process is automated and triggered by a GitHub action at 7am Pacific time on Tuesday on cadence decided
```
os-d:
```suggestion
- The release process is automated and triggered by a GitHub action at 7am Pacific time on Tuesday on a cadence decided

```

PR #630, opened 2025-07-30
[REBASE&FF] patina_sdk: Improve pointer provenance

Javagedes:
## Description

This commit enhances the pointer provenance tracking for allocations
made through the MemoryManager service in the Patina SDK. Pointer
provenance is crucial for ensuring memory safety, preventing undefined
behavior, allowing for enhanced compiler optimizations, and most
importantly, better evaluation of memory access patterns by static
analysis tools such as miri.

For more information on pointer provenance, see:
https://doc.rust-lang.org/std/ptr/index.html#provenance

With this commit, provenance is now exposed within the
`core_allocate_pages` function via the `expose_provenance` method, which
means any conversions of an address in the bounds of that allocation
will gain that pointer's provenance so long as it is converted via the
`with_exposed_provenance` or `with_exposed_provenance_mut` methods.
Simple conversions like `as *mut T` or `as *const T` should also pick up
the provenance, however, it is better to use the provided methods to
ensure provenance is tracked correctly.

While this does help with provenance tracking once the allocation is
exposed to the caller of `core_allocate_pages`, further work should
eventually be done to either expose the pointer's provenance earlier in
the call-stack, or to generate and return a pointer type such as
`NonNull<T>` in the rust core, rather than a `usize`, as provenance is
maintained with these types.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

CI

## Integration Instructions

N/A

### DIFF
--- a/sdk/patina_sdk/src/component/service/memory.rs
+++ b/sdk/patina_sdk/src/component/service/memory.rs
@@ -420,7 +420,18 @@ impl PageAllocation {
     /// Returns an appropriate `Err` if the address is not page aligned or the
     /// page count is zero.
     ///
-    /// # Safety
+    /// ## Pointer Provenance
+    ///
+    /// As the function interface does not take a pointer, and instead takes a usize
+    /// representing the address, there is no pointer provenance metadata during
+    /// build by default. This function uses [with_exposed_provenance_mut] to allow
+    /// the compiler / other tools to attempt to associate the address with the
+    /// original pointer's provenance. It is imperative that the caller exposes the
+    /// original pointer's provenance before passing the address to this function via
+    /// one of the means described in [pointer provenance](https://doc.rust-lang.org/std/ptr/index.html#provenance)
+    /// such as `expose_provenance`.
+    ///
+    /// ## Safety
     ///
     /// The caller is responsible for ensuring the provided address and page count
     /// are valid and are read/write accessible. Producing a `PageAllocation` will


patina_dxe_core/src/allocator.rs
```diff
@@ -495,7 +495,7 @@ pub fn core_allocate_pages(
             };
 
             if let Ok(ptr) = result {
-                unsafe { memory.write(ptr.as_ptr() as *mut u8 as u64) }
+                unsafe { memory.write(ptr.cast::<u8>().as_ptr().expose_provenance() as u64) }
```
os-d, line 498: Can't `expose_provenance` simply be called on self (e.g. not doing the cast and .as_ptr()`?

sdk/patina_sdk/src/component/service/memory.rs
```diff
@@ -483,7 +498,7 @@ impl PageAllocation {
     #[inline(always)]
     pub fn into_raw_ptr<T>(self) -> *mut T {
         // Move this struct to manual management and return the address.
-        ManuallyDrop::new(self).address as *mut T
+        ManuallyDrop::new(self).blob.cast::<T>().as_ptr()
```
os-d, line 500: Would this be a place to use `.addr()` instead of `.as_ptr()`.

sdk/patina_sdk/src/component/service/memory.rs
```diff
@@ -599,12 +614,13 @@ impl PageAllocation {
     /// This is not a public method as it invalidates `Self` without consuming `self`.
     /// This should only be used internally to free the memory when dropping `self`.
     fn free_pages(&mut self) {
+        let address = self.blob.as_ptr().addr();
```
os-d: Can't `.addr()` simply be called on self, no `.as_ptr()`, e.g. NonNull's `.addr()` method.

sdk/patina_sdk/src/component/service/memory.rs
```diff
@@ -644,18 +660,21 @@ unsafe impl Allocator for PageFree {
...
+        if self.blob != ptr {
+            log::error!(
+                "PageFree was not used to free the correct memory! Leaking memory at {:?}!",
+                self.blob.as_ptr()
+            );
             debug_assert!(false, "PageFree was not used to free the correct memory!");
             return;
         }
 
+        let address = self.blob.as_ptr().addr();
```
os-d: Same questions on whether .as_ptr() is needed

sdk/patina_sdk/src/component/service/memory.rs
```diff
@@ -459,10 +477,7 @@ impl PageAllocation {
             for i in 0..self.page_count {
                 // SAFETY: The address is page aligned and the page count is valid.
                 //         This is the responsibility of the caller of PageAllocation::new().
-                core::ptr::write_volatile(
-                    (self.address + uefi_pages_to_size!(i)) as *mut [u8; UEFI_PAGE_SIZE],
-                    [0_u8; UEFI_PAGE_SIZE],
-                );
+                self.blob.cast::<UefiPage>().add(i).write_volatile(UefiPage::zeros());
```
cfernald: Do you know if this would put [0; 0x1000] on the stack? This is the things that makes me nervous about large type writes. It would seem crazy for the compiler to put this on the stack, but it seems implementation specifics.

patina_dxe_core/src/allocator.rs
Javagedes: yes, It can. I missed that `NonNull<T>` had these functions. I will update. Thanks :)

sdk/patina_sdk/src/component/service/memory.rs
os-d: I was also worried about that, but thought that the original implementation had the same potential. Either way, we should confirm and amend this if needed.

patina_dxe_core/src/allocator.rs
Javagedes: Actually, I cannot. The nightly feature for this is stabilized in 1.89, which has not been released yet. Will create a tracking issue once this is merged.

https://github.com/rust-lang/rust/issues/135243
https://github.com/rust-lang/rust/pull/144509/files

sdk/patina_sdk/src/component/service/memory.rs
Javagedes: No, because in this scenario we are returning the actual pointer, not a `usize`. actual pointers keep their provenance. It's only when we convert to a `usize` that we lose it.

sdk/patina_sdk/src/component/service/memory.rs
cfernald: agreed, this just a general question. Not blocking

Javagedes: @cfernald @os-d I dumped the assembly with `cargo-show-asm` before and after this change, for zero_pages.

The previous implementation does in-fact create a stack of a bit over 4k. However this implementation creates a stack of over 12K, so, uhhh. yeah. I'm going to change this.

Javagedes: From offline conversation, we decided to change this to `write_bytes` now that our understanding of why the rust compiler optimized write in the past. Turned this into a rebase & ff in case we need to revert that portion because it still optimizes the zeroing.

PR #653, opened 2025-08-01
allocator: Remove static memory type info table

Javagedes:
## Description

This commit removes the static memory type info table. Due to the need
to keep all memory type information in a single table (as it is
published as a system table), this commit opts to give ownership of
the entire table to the GCD. When the GCD creates a new allocator, it
passes a pointer to the memory type info for that allocator, to the
allocator. This allows the allocator to self update it's memory type
information as needed, while still keeping the entire table together.

For dynamically generated allocators (e.g. the allocators not a part
of the 17 tracked memory types), a new MemoryTypeInfo is created and
leaked, for them to use.

Additionally, this removes the page_change_callback in favor of a
function that does the same.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

1. CI Passes
2. Verified boot to UEFI Shell

## Integration Instructions

N/A

### DIFF
--- a/patina_dxe_core/src/allocator.rs
+++ b/patina_dxe_core/src/allocator.rs
@@ -72,7 +72,7 @@ cfg_if::cfg_if! {
 #[cfg_attr(target_os = "uefi", global_allocator)]
 pub(crate) static EFI_BOOT_SERVICES_DATA_ALLOCATOR: UefiAllocator = UefiAllocator::new(
     &GCD,
-    GCD.memory_type_info(efi::BOOT_SERVICES_DATA),
+    NonNull::from_ref(GCD.memory_type_info(efi::BOOT_SERVICES_DATA)),
     protocol_db::EFI_BOOT_SERVICES_DATA_ALLOCATOR_HANDLE,
     DEFAULT_PAGE_ALLOCATION_GRANULARITY,
 );
@@ -82,14 +82,14 @@ pub(crate) static EFI_BOOT_SERVICES_DATA_ALLOCATOR: UefiAllocator = UefiAllocato
 // the other allocators use.
 pub static EFI_LOADER_CODE_ALLOCATOR: UefiAllocator = UefiAllocator::new(
     &GCD,
-    GCD.memory_type_info(efi::LOADER_CODE),
+    NonNull::from_ref(GCD.memory_type_info(efi::LOADER_CODE)),
     protocol_db::EFI_LOADER_CODE_ALLOCATOR_HANDLE,
     DEFAULT_PAGE_ALLOCATION_GRANULARITY,
 );
 
 pub static EFI_BOOT_SERVICES_CODE_ALLOCATOR: UefiAllocator = UefiAllocator::new(
     &GCD,
-    GCD.memory_type_info(efi::BOOT_SERVICES_CODE),
+    NonNull::from_ref(GCD.memory_type_info(efi::BOOT_SERVICES_CODE)),
     protocol_db::EFI_BOOT_SERVICES_CODE_ALLOCATOR_HANDLE,
     DEFAULT_PAGE_ALLOCATION_GRANULARITY,
 );
@@ -98,7 +98,7 @@ pub static EFI_BOOT_SERVICES_CODE_ALLOCATOR: UefiAllocator = UefiAllocator::new(
 // passed in
 pub static EFI_RUNTIME_SERVICES_CODE_ALLOCATOR: UefiAllocator = UefiAllocator::new(
     &GCD,
-    GCD.memory_type_info(efi::RUNTIME_SERVICES_CODE),
+    NonNull::from_ref(GCD.memory_type_info(efi::RUNTIME_SERVICES_CODE)),
     protocol_db::EFI_RUNTIME_SERVICES_CODE_ALLOCATOR_HANDLE,
     RUNTIME_PAGE_ALLOCATION_GRANULARITY,
 );
@@ -107,7 +107,7 @@ pub static EFI_RUNTIME_SERVICES_CODE_ALLOCATOR: UefiAllocator = UefiAllocator::n
 // passed in
 pub static EFI_RUNTIME_SERVICES_DATA_ALLOCATOR: UefiAllocator = UefiAllocator::new(
     &GCD,
-    GCD.memory_type_info(efi::RUNTIME_SERVICES_DATA),
+    NonNull::from_ref(GCD.memory_type_info(efi::RUNTIME_SERVICES_DATA)),
     protocol_db::EFI_RUNTIME_SERVICES_DATA_ALLOCATOR_HANDLE,
     RUNTIME_PAGE_ALLOCATION_GRANULARITY,
 );
@@ -318,7 +318,7 @@ impl AllocatorMap {
             // from the GCD. Otherwise, we will just leak a new memory type info struct with the given memory type and
             // have the allocator use it.
             let memory_type_info = if (memory_type as usize) <= GCD.memory_type_info_table().len() {
-                GCD.memory_type_info(memory_type)
+                NonNull::from_ref(GCD.memory_type_info(memory_type))
             } else {
                 NonNull::from_ref(Box::leak(Box::new(EFiMemoryTypeInformation { memory_type, number_of_pages: 0 })))
             };
--- a/patina_dxe_core/src/allocator/uefi_allocator.rs
+++ b/patina_dxe_core/src/allocator/uefi_allocator.rs
@@ -313,7 +313,7 @@ mod tests {
             static GCD: SpinLockedGcd = SpinLockedGcd::new(None);
             let ua = UefiAllocator::new(
                 &GCD,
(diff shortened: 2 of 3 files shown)


patina_dxe_core/src/allocator.rs
```diff
@@ -320,7 +313,18 @@ impl AllocatorMap {
...
+            // If this is one of the memory types tracked by the system table, we will use the memory type info struct
+            // from the GCD.
+            if (memory_type as usize) < GCD.memory_type_info_table().len() {
+                return Box::leak(Box::new(UefiAllocator::new(&GCD, memory_type, handle, granularity)));
+            }
+
+            // Otherwise, we will just leak a new memory type info struct with the given memory type and have the
+            // allocator use it.
+            let memory_type_info = EFiMemoryTypeInformation { memory_type, number_of_pages: 0 };
+            let memory_type_info = NonNull::new(Box::leak(Box::new(memory_type_info))).unwrap();
```
os-d: unwrap is probably fine here because if we are out of memory this early we won't be able to do anything else, but still, maybe not a bare unwrap?

patina_dxe_core/src/allocator.rs
```diff
@@ -757,63 +761,10 @@ pub fn terminate_memory_map(map_key: usize) -> Result<(), EfiError> {
...
-
 pub fn install_memory_type_info_table(system_table: &mut EfiSystemTable) -> Result<(), EfiError> {
     // SAFETY: This is safe because we are initializing the table with a static array
-    #[allow(static_mut_refs)]
-    config_tables::core_install_configuration_table(
-        guid::MEMORY_TYPE_INFORMATION,
-        unsafe { MEMORY_TYPE_INFO_TABLE.as_mut_ptr() as *mut c_void },
-        system_table,
-    )
+    let table_ptr = GCD.memory_type_info_table().as_ptr() as *mut EFiMemoryTypeInformation as *mut c_void;
```
os-d: Does this need to be 
```suggestion
    let table_ptr = GCD.memory_type_info_table_mut().as_mut_ptr() as *mut EFiMemoryTypeInformation as *mut c_void;
```

since we are casting as mut pointer? This would require getting mutable access from the GCD.

patina_dxe_core/src/allocator/uefi_allocator.rs
```diff
@@ -51,18 +51,44 @@ impl UefiAllocator {
         gcd: &'static SpinLockedGcd,
         memory_type: efi::MemoryType,
         allocator_handle: efi::Handle,
-        page_change_callback: PageChangeCallback,
+        page_allocation_granularity: usize,
+    ) -> Self {
+        debug_assert!(
```
os-d: What happens in a release build if this fails?

Javagedes: In a release build we would buffer overflow and have a `NonNull<EFiMemoryTypeInformation>` pointing at some other part of the GCD or even further past it.

However that is not really possible at runtime because of the `if (memory_type as usize) < GCD.memory_type_info_table().len()` check in the changes to `patina_dxe_core/src/allocator.rs`

This debug_assert! is mainly to catch if we create a static UefiAllocator that is an EFI memory type past 17, which should hypothetically never happen. I did not exactly want to have a check for that at runtime on production builds. It seemed unecessary.

patina_dxe_core/src/allocator.rs
Javagedes: I'm not sure how much it matters for this. We are handing off a pointer to this table to the black abyss called EDKII and crossing the C FFI boundary. Whether we pass it a *const or *mut does not mean much.

Additionally, we cannot access the static GCD mutably (that is what this is trying to fix). I could wrap it in a mutex to get it mutably, but it ultimately does not matter because we are handing off the raw pointer to a C FFI who will not use the mutex to access the underlying data safely.

os-d: Well this is not passed to the C FFI boundary. This is passed to a Rust function `core_install_configuration_table` which does *something* and then writes it to memory. I agree once it is written to memory, whatever, but my concern is the Rust side of this. If we have something that is a mutable pointer, it should actually have come from a mutable source, at least as our conversations with Wesley Weiser went. When we didn't do that is when we ended up with UB and the compiler throwing away writes, etc. I'm not actually sure why `core_install_configuration_table` needs a mutable pointer, I assume it could be const? I didn't take a look though.

Javagedes: Used `NonNull::as_ptr` to create a `*mut <Ptr>` instead of the UB of just casting a *const to a *mut.

os-d: I'm a bit confused, isn't this calling `memory_type_info_table()` here:

image

Which is then an immutable reference being cast to a mutable pointer? Don't we need to change that function to return a mutable reference (or add a new function for a mutable reference)? Or am I misreading the abstraction layers ?

patina_dxe_core/src/allocator/uefi_allocator.rs
```diff
@@ -51,18 +51,44 @@ impl UefiAllocator {
...
+            memory_type as usize <= gcd.memory_type_info_table().len(),
+            "Memory Type is not tracked in the GCD, use `new_dynamic` instead."
+        );
+
+        // Get the pointer to the first memory type info struct in the table, then offset it by the memory type, which
+        // is also the index in the table. The debug_assert above ensures that the memory type is tracked by the GCD.
+        let memory_type_info = NonNull::new(unsafe {
+            (gcd.memory_type_info_table().as_ptr() as *mut EFiMemoryTypeInformation).add(memory_type as usize)
```
os-d: Same question here

Javagedes: This can be updated to `NonNull::from_ref(...)` once we update to 1.89. If we go this route for this PR, I will hold back on merging this PR until we update to 1.89

Javagedes: I have confirmed via a miri test below, that there is no UB regarding using `NonNull::from_ref(...)`

```rust
    #[test]
    fn test_miri() {
        with_locked_state(|| {
            static GCD: SpinLockedGcd = SpinLockedGcd::new(None);

            let base = init_gcd(&GCD, 0x400000);

            let ua = UefiAllocator::new(&GCD, efi::RUNTIME_SERVICES_DATA, 1 as _, DEFAULT_PAGE_ALLOCATION_GRANULARITY);

            let pointer = ua.allocate_pages(AllocationStrategy::TopDown(None), 2, 0x4000).unwrap();
            assert_eq!(ua.number_of_pages(), 2);

            let pointer2 = ua.allocate_pages(AllocationStrategy::TopDown(None), 2, 0x4000).unwrap();
            assert_eq!(ua.number_of_pages(), 4);

            unsafe { ua.free_pages(pointer.cast::<u8>().as_ptr() as usize, 2).unwrap() };
            unsafe { ua.free_pages(pointer2.cast::<u8>().as_ptr() as usize, 2).unwrap() };
            assert_eq!(ua.number_of_pages(), 0);

            unsafe { System.dealloc(
                base as *mut u8,
                Layout::from_size_align(0x400000, UEFI_PAGE_SIZE).unwrap_or_else(|err| {
                    panic!("Allocation layout error: {:#?}", err)
                })
            ) };
        });
    }
```

patina_dxe_core/src/allocator.rs
Javagedes: I have confirmed via a miri test below, that there is no UB regarding using `NonNull::from_ref(...)`

```rust
    #[test]
    fn test_miri() {
        with_locked_state(|| {
            static GCD: SpinLockedGcd = SpinLockedGcd::new(None);

            let base = init_gcd(&GCD, 0x400000);

            let ua = UefiAllocator::new(&GCD, efi::RUNTIME_SERVICES_DATA, 1 as _, DEFAULT_PAGE_ALLOCATION_GRANULARITY);

            let pointer = ua.allocate_pages(AllocationStrategy::TopDown(None), 2, 0x4000).unwrap();
            assert_eq!(ua.number_of_pages(), 2);

            let pointer2 = ua.allocate_pages(AllocationStrategy::TopDown(None), 2, 0x4000).unwrap();
            assert_eq!(ua.number_of_pages(), 4);

            unsafe { ua.free_pages(pointer.cast::<u8>().as_ptr() as usize, 2).unwrap() };
            unsafe { ua.free_pages(pointer2.cast::<u8>().as_ptr() as usize, 2).unwrap() };
            assert_eq!(ua.number_of_pages(), 0);

            unsafe { System.dealloc(
                base as *mut u8,
                Layout::from_size_align(0x400000, UEFI_PAGE_SIZE).unwrap_or_else(|err| {
                    panic!("Allocation layout error: {:#?}", err)
                })
            ) };
        });
    }
```

Javagedes:
@os-d @makubacki with the 1.89 update, this is now ready for a final review and merge when you have some time :)

PR #658, opened 2025-08-04
allocator: Prevent panic in linked_list_allocator edge case

### LINKED ISSUE #616: [Bug]: Assert in `fixed_size_block_allocator` on physical Intel platform

### Current Behavior

First boot to Windows succeeds but an assert is seen after rebooting the system on an Intel physical platform:

```
  ERROR - panicked at .\patina_dxe_core5.0.0\src\allocator\fixed_size_block_allocator.rs:818:25:
  assertion failed: false
```

This began after a recent Windows update in the device LKG.

- Pass: 250630.0.26200 (OS Build 26200.5670)
- Fail: 250714.0.26200 (OS Build 26200.5702)

### Expected Behavior

Window boot succeeds on all boots.

### Steps To Reproduce

1. Flash debug UEFI image with Patina DXE Core (based on patina v5.0.0)
2. Boot into OS for the first time
3. Restart OS <-- Issue happens here

### Build Environment

```markdown
- OS(s): Windows 11 
- Tool Chain(s): VS2022. Rust 1.85. patina v5.0.0.
- Targets Impacted: debug
```

### Version Information

```text
v5.0.0
```

### Urgency

Medium

Someone else needs to fix it

Maintainer feedback requested

### Anything else?

_No response_

makubacki:
## Description

Fixes #616 

Ensure enough space is available for hole list metadata.

https://docs.rs/linked_list_allocator/latest/linked_list_allocator/hole/struct.HoleList.html#method.new

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- Verified Windows allocation failure no longer occurs with change
- Unit test

## Integration Instructions

- N/A

### DIFF
--- a/patina_dxe_core/src/allocator/fixed_size_block_allocator.rs
+++ b/patina_dxe_core/src/allocator/fixed_size_block_allocator.rs
@@ -291,8 +291,13 @@ impl FixedSizeBlockAllocator {
         }
 
         // Determine how much additional memory is required
-        let additional_mem_required =
-            layout.pad_to_align().size() + Layout::new::<AllocatorListNode>().pad_to_align().size();
+        //
+        // Per the `linked_list_allocator::hole::HoleList::new` documentation, depending on the alignment of the
+        // hole_addr pointer, the minimum size for storing required metadata is between 2 * size_of::<usize> and
+        //  3 * size_of::<usize>. The size reservation for `additional_mem_required` assumed the largest size.
+        let additional_mem_required = layout.pad_to_align().size()
+            + Layout::new::<AllocatorListNode>().pad_to_align().size()
+            + 3 * size_of::<usize>();
         let additional_mem_required = align_up_size(additional_mem_required, align_of::<AllocatorListNode>());
 
         Err(FixedSizeBlockAllocatorError::OutOfMemory(additional_mem_required))
@@ -1607,7 +1612,7 @@ mod tests {
             assert_eq!(stats.page_allocation_calls, 0);
             assert_eq!(stats.page_free_calls, 0);
             assert_eq!(stats.reserved_size, MIN_EXPANSION * 2);
-            assert_eq!(stats.reserved_used, MIN_EXPANSION);
+            assert_eq!(stats.reserved_used, MIN_EXPANSION + uefi_pages_to_size!(1));
             assert_eq!(stats.claimed_pages, uefi_size_to_pages!(MIN_EXPANSION * 2));
 
             unsafe {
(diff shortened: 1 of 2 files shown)


patina_dxe_core/src/allocator/fixed_size_block_allocator.rs
```diff
@@ -291,8 +291,13 @@ impl FixedSizeBlockAllocator {
         }
 
         // Determine how much additional memory is required
-        let additional_mem_required =
-            layout.pad_to_align().size() + Layout::new::<AllocatorListNode>().pad_to_align().size();
+        //
+        // Per the `linked_list_allocator::hole::HoleList::new` documentation, depending on the alignment of the
+        // hole_addr pointer, the minimum size for storing required metadata is between 2 * size_of::<usize> and
+        //  3 * size_of::<usize>. The size reservation for `additional_mem_required` assumed thes largest size.
+        let additional_mem_required = layout.pad_to_align().size()
```
Javagedes, line 298: `Layout::new::<AllocatorListNode>().pad_to_align().size() + 3 * size_of::<usize>()` is all `const` where as `layout.pad_to_align().size()` will not be const.

I assume the compiler is smart enough to recognize this and evaluate that this line is actually just `layout.pad_to_align().size() + MAGIC_NUMBER`, but we should review the assembly of this function using https://github.com/pacak/cargo-show-asm to double check that this is true and we are not doing unnecessary work.

makubacki: This only modifies the value of the existing constant expression. Wouldn't this also vary according to optimization levels. In the `dev` profile, it appears the constant values are added here:

```
// patina_dxe_core\src\allocator\fixed_size_block_allocator.rs:298
        add rdi, 64     // Layout::new::<AllocatorListNode>().pad_to_align().size()
        jb .LBB353_20
        add rdi, 24     // 3 * size_of::<usize>()
```

Javagedes: I don't agree `layout.pad_to_align().size` is constant because that variable comes from the function, and that function is not constant. However, `Layout::new::<AllocatorListNode>().pad_to_align().size() + 3 * size_of::<usize>()` is definitely constant. My thought was if it added many assembly commands (such as the adds, the multiplies, etc), you could do something like 

```rust
const METADATA_SIZE: usize = Layout::new::<AllocatorListNode>().pad_to_align().size()
  + 3 * size_of::<usize>();
  
let additional_mem_required = layout.pad_to_align().size() + METADATA_SIZE;
```

Which should realistically drop it down to just two adds. But from your check, this is 3 adds verse 2 adds, so not sure its a big deal. Feel free to resolve this (regardless of if you keep it the same or change it) and I'll approve :)

makubacki: By "This only modifies the value of the existing constant expression" I was referring to `Layout::new::<AllocatorListNode>().pad_to_align().size()` as the constant expression. I'm not suggesting `layout.pad_to_align().size()` is constant.

I think the main thing shown in the assembly is that the constant items are being treated as their integer values which should allow straightforward optimization of those values.

Issue #620, opened 2025-07-23
[Bug]: into_raw_ptr in MemoryManager should check size


cfernald:
### Is there an existing issue for this?

- [x] I have searched existing issues

### Current Behavior

The `into_raw_ptr` function in `MemoryManager` does not currently check the size of the pointer for sized types matches the size of the page allocation. This is currently the responsibility of the caller, but this causes unnecessary risk and would be better done by the implementaiton.

### Expected Behavior

The implementation should change `into_raw_ptr` to return an `Option<*mut T>` conditional on the size check. An additional function could be created `into_raw_ptr_unchecked` to allow for unchecked or unsized ptr conversion.

### Steps To Reproduce

N/A

### Build Environment

```markdown
- OS(s): N/A
- Tool Chain(s): N/A
- Targets Impacted: N/A
```

### Version Information

```text
N/A
```

### Urgency

Medium

### Are you going to fix this?

I will fix it

### Do you need maintainer feedback?

No maintainer feedback needed

PR #661, opened 2025-08-05
patina_sdk: Add size check to page allocation's into_raw_ptr

cfernald:
## Description

Other routines in the Memory Manager API will return `None` in the event that the type of the pointer does not fit into the allocated pages. This moves that logic down to the ptr so that all returnes will do size checks. A `into_raw_ptr_unchecked` could be added in the future if a truely unchecked version is needed.

This change additionally removes the `try_` prefixes as the panic versions no longer exist, and this makes the interfaces more consistent.

CLOSES: #620 

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- [x] Q35 with patina tests
- [x] Unit tests

## Integration Instructions

Update usage to remote `try_` prefix and handle option from `into_raw_ptr`

### DIFF
--- a/sdk/patina_sdk/src/component/service/memory.rs
+++ b/sdk/patina_sdk/src/component/service/memory.rs
@@ -96,18 +96,18 @@ pub trait MemoryManager {
     /// fn component(memory_manager: &dyn MemoryManager) -> Result<(), MemoryError> {
     ///     // Allocate a page of memory and leak it.
     ///     let alloc = memory_manager.allocate_pages(1, AllocationOptions::new())?;
-    ///     let static_u64 = alloc.try_leak_as(42).unwrap();
+    ///     let static_u64 = alloc.leak_as(42).unwrap();
     ///
     ///     // Allocate a page and convert it into a Box.
     ///     let alloc = memory_manager.allocate_pages(1, AllocationOptions::new())?;
     ///     #[cfg(feature = "alloc")]
-    ///     let boxed_value = alloc.try_into_box(42).unwrap();
+    ///     let boxed_value = alloc.into_box(42).unwrap();
     ///     // Memory will be safely freed when the Box is dropped.
     ///
     ///     // Allocate a page of memory and manually manage it, making sure
     ///     // to free it.
     ///     let alloc = memory_manager.allocate_pages(1, AllocationOptions::new())?;
-    ///     let ptr = alloc.into_raw_ptr::<u8>();
+    ///     let ptr = alloc.into_raw_ptr::<u8>().unwrap();
     ///     unsafe { memory_manager.free_pages(ptr as usize, 1)? };
     ///
     ///     Ok(())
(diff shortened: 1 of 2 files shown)


sdk/patina_sdk/src/component/service/memory.rs
```diff
@@ -494,10 +494,19 @@ impl PageAllocation {
     /// the type fits in the allocation before dereferencing. The memory will not
     /// be initialized and the caller is responsible ensuring the type is valid.
     #[must_use]
-    #[inline(always)]
-    pub fn into_raw_ptr<T>(self) -> *mut T {
+    pub fn into_raw_ptr<T>(mut self) -> Option<*mut T> {
```
os-d, line 502: This is only ever intended to be called in a chain, right? e.g. `MM.alloc().into_raw_ptr()`? That use case seems appropriate to free if an error occurs but it would be wild to call `into_raw_ptr()` and have it free the memory if it failed. But I don't think there would be any use case like that.

sdk/patina_sdk/src/component/service/memory.rs
```diff
@@ -575,18 +574,8 @@ impl PageAllocation {
...
-            // but we want to avoid triggering the panic in its `Drop` implementation.
-            // To handle this safely, we manually free the pages and then call `forget`
-            // to prevent `drop` from running.
-            self.free_pages();
-            core::mem::forget(self);
-            return None;
-        }
-
-        let ptr = self.into_raw_ptr::<T>();
+    pub fn leak_as<'a, T>(self, value: T) -> Option<&'a mut T> {
```
os-d, line 590: Shouldn't this stay as `try_leak_as`, since it can fail?

sdk/patina_sdk/src/component/service/memory.rs
makubacki: In any case, the function description should be updated to describe the behavior. It currently says:

```
    /// Consumes the allocation and returns the raw address which must be manually
    /// freed using the [MemoryManager::free_pages] routine. If the caller fails
    /// to free the memory, it will leak. The caller is responsible for assuring
    /// the type fits in the allocation before dereferencing. The memory will not
    /// be initialized and the caller is responsible ensuring the type is valid.
```

Which does not cover the case this function performs the `free` operation and returns `None`.

makubacki: This does seem like a weird side effect. The function returning an `Option` is intuitive but this seems like something that would be handled by the caller based on the result from `into_raw_ptr()`.

sdk/patina_sdk/src/component/service/memory.rs
cfernald: Typically the try_ pattern seems to be used in the method where there is a corresponding panic version. Since we don't have those we can make them all try_s, e.g. try_into_raw_ptr, or we can just leave them all as is and the option is clarification on that. 

I don't have a strong preference but obviously want all the function names to be consistent.

sdk/patina_sdk/src/component/service/memory.rs
cfernald: The problem is that this function really has to consume the object, or else we dont have a strong way to prevent memory allocations to be reused. So the question is then, what do you do if it fails? We have roughyl 3 options.
1. current, free the memory and return `None`
2. leak the memory and return `None`
3. Return `Result<*mut ptr, PageAllocation>` 

I would personally prefer to free the memory then to leak it. The `Result` is interesting, but also then requires more robust error handling for each and every invocation where this should really never happen in expected code paths. option 1 just makes the most sense to me. I will update the description though.

sdk/patina_sdk/src/component/service/memory.rs
cfernald: as discussed offline, this is for consistency. We cna move the rest to try_ but will leave it as is for unless anyone is opposed.

sdk/patina_sdk/src/component/service/memory.rs
makubacki: The choice itself seems reasonable. The issue was mostly around communicating the behavior. The documentation update seems sufficient to clarify that.

PR #672, opened 2025-08-08
[REBASE & FF] Implement a memory poke test for debugger

### LINKED ISSUE #655: [Feature]: Debugger: Implement graceful handling of data faults
### Feature Overview

Page tables sometime have valid mapping to invalid memory, especially on AArch64. When this happens the debugger can hit a nested fault due to the debugger application requesting to read such an address. This happens especially on windbgx which will try to read different addresses on first connection. 

### Solution Overview

Create a function that will set a static, try to touch the memory, and return a boolean indicating if a fault was taken. This will require the exception handler check for that static, update it, and increment the instruction pointer before returning from the nested exception.

### Alternatives Considered

_No response_

### Urgency

Low

I will implement the feature

No maintainer feedback needed

### Anything else?

_No response_

cfernald:
## Description

The debugger will attempt to check the page tables to determine if memory is safe to access, but this is not 100% affective as memory will often be mapped that is not actually real or accessible. This can happen in the inherited page tables or for MMIO regions that may not be present.

The poke test is an architecture specific function that will set a marker and touch the memory. During the exception entry if the marker is set then it can be concluded that the memory is bad and we can jump past the faulting instruction and return. This prevents nested faults when actually accessing the memory.

CLOSES #655 

### Interrupt handler fix 

This change exposed a bug in the x64 interrupt handling, where the previously the implementation incorrectly tried to preserve the exception code on the stack for interrupt vectors that originally pushed the code. The SDM details in volume 3 section 6.13 that "Note that the error code is not popped when the IRET instruction is executed to return from an exception handler, so the handler must remove the error code before executing a return"

This changes the hint to be simply for alignment and always pops the exception code whether it is real or not.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested
- [x] Q35
- [x] SBSA
- [x] Physical x64 device

## Integration Instructions

N/A

### DIFF
--- a/core/patina_debugger/src/arch/aarch64.rs
+++ b/core/patina_debugger/src/arch/aarch64.rs
@@ -1,4 +1,9 @@
-use core::{arch::asm, num::NonZeroUsize, ops::Shr};
+use core::{
+    arch::asm,
+    num::NonZeroUsize,
+    ops::Shr,
+    sync::atomic::{AtomicBool, Ordering, compiler_fence},
+};
 
 use gdbstub::arch::{RegId, Registers};
 use patina_internal_cpu::interrupts::ExceptionContext;
@@ -32,6 +37,8 @@ const MDSCR_MDE: u64 = 0x8000;
 const MDSCR_KDE: u64 = 0x2000;
 
 const DAIF_DEBUG_MASK: u64 = 0x200;
+
+static POKE_TEST_MARKER: AtomicBool = AtomicBool::new(false);
 
 macro_rules! read_sysreg {
   ($reg:expr) => {{
@@ -139,14 +146,14 @@ impl DebuggerArch for Aarch64Arch {
         let mut daif = read_sysreg!("daif");
         daif |= DAIF_DEBUG_MASK;
         write_sysreg!("daif", daif);
-        core::sync::atomic::compiler_fence(core::sync::atomic::Ordering::SeqCst);
+        compiler_fence(Ordering::SeqCst);
 
         // Clear the OS lock if needed
         let oslsr_el1 = read_sysreg!("oslsr_el1");
         if oslsr_el1 & 1 != 0 {
             unsafe { asm!("msr oslar_el1, xzr") };
         }
-        core::sync::atomic::compiler_fence(core::sync::atomic::Ordering::SeqCst);
+        compiler_fence(Ordering::SeqCst);
 
         // Enable kernel and monitor debug bits
         let mut mdscr_el1 = read_sysreg!("mdscr_el1");
@@ -157,13 +164,13 @@ impl DebuggerArch for Aarch64Arch {
         for i in 0..NUM_WATCHPOINTS {
             write_dbg_wcr(i, Wcr::from(0));
         }
-        core::sync::atomic::compiler_fence(core::sync::atomic::Ordering::SeqCst);
+        compiler_fence(Ordering::SeqCst);
 
         // Enable debug exceptions in DAIF
         daif = read_sysreg!("daif");
         daif &= !DAIF_DEBUG_MASK;
         write_sysreg!("daif", daif);
-        core::sync::atomic::compiler_fence(core::sync::atomic::Ordering::SeqCst);
+        compiler_fence(Ordering::SeqCst);
     }
 
     fn add_watchpoint(address: u64, length: u64, access_type: gdbstub::target::ext::breakpoints::WatchKind) -> bool {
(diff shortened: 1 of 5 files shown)


core/patina_debugger/src/arch/aarch64.rs
```diff
@@ -73,9 +75,18 @@ impl DebuggerArch for Aarch64Arch {
         }
     }
 
-    fn process_entry(_exception_type: u64, context: &mut ExceptionContext) -> ExceptionInfo {
+    fn process_entry(_exception_type: u64, context: &mut ExceptionContext) -> Result<ExceptionInfo, DebugError> {
+        // First check if this was due to a poke-test
+        if POKE_TEST_MARKER.load(core::sync::atomic::Ordering::SeqCst) {
```
Javagedes: Theoretically, this is not atomic. You probably need to do something like

```rust
if POKE_TEST_MARKER.swap(false, core::sync::atomic::Ordering::seqCst) {
  context.elr += 4
  return Err(DebugError::IgnoreException);
}
```

This will always swap the stored pointer for false, but if the previous value is true, you can do your special handling.

Javagedes: maybe possibly `compare_exchange` also. Not sure the implications.

core/patina_debugger/src/arch/aarch64.rs
```diff
@@ -73,9 +75,18 @@ impl DebuggerArch for Aarch64Arch {
...
+            // Toggle the boolean to indicate a fault occurred.
+            POKE_TEST_MARKER.store(false, core::sync::atomic::Ordering::SeqCst);
+            // Increment the instruction pointer to step past the load
+            context.elr += 4;
+            return Err(DebugError::IgnoreException);
```
Javagedes: It seems a bit of a misnomer to return an error here. Would it not be more accurate to update Exception Info (If its not a standardized struct) to have a debugger or ignore field or something?

core/patina_debugger/src/arch/aarch64.rs
```diff
@@ -266,6 +277,28 @@ impl DebuggerArch for Aarch64Arch {
...
+        // This will raise a page fault if the address is not accessible.
+
+        let _value: u64;
+        // SAFETY: The safety of this is dubious and may cause a page fault, but
+        // the exception handler will catch it and resolve it by stepping beyond
+        // the exception.
+        unsafe { asm!("ldr {}, [{}]", out(reg) _value, in(reg) address, options(nostack)) };
+
+        // Check if the marker was cleared, indicating a page fault.
+        if POKE_TEST_MARKER.load(core::sync::atomic::Ordering::SeqCst) {
```
Javagedes: same issue not being fully atomic. should check and store in the same command.

core/patina_debugger/src/arch/aarch64.rs
cfernald: I don't actually care if the change of value is atomic or not, but I agree the compare exchange is cleaner either way and so can change it.

cfernald: This is because the clear is just to reset the state and does not interact with the async access to the static.

core/patina_debugger/src/arch/aarch64.rs
cfernald: Mentioned in the other thread. I don't actually care if the set to false is atomic with the check, but may change this for brevity.

core/patina_debugger/src/arch/aarch64.rs
cfernald: Maybe i'll just split out this functionality actually.

core/patina_debugger/src/arch/x64.rs
```diff
@@ -170,6 +176,33 @@ impl DebuggerArch for X64Arch {
...
+
+        // Check if the marker was cleared, indicating a page fault. Reset either way.
+        if POKE_TEST_MARKER.swap(false, Ordering::SeqCst) { Ok(()) } else { Err(()) }
+    }
+
+    fn check_memory_poke_test(context: &mut ExceptionContext) -> bool {
+        let poke_test = POKE_TEST_MARKER.swap(false, Ordering::SeqCst);
+        if poke_test {
+            // We need to increment the instruction pointer to step past the load
+            context.rip += 3;
```
os-d, line 201: Why 3 vs 4 for aarch64? Just a difference of where these point to by one byte?

core/patina_debugger/src/memory.rs
```diff
@@ -98,10 +98,28 @@ pub fn write_memory<Arch: DebuggerArch>(address: u64, buffer: &[u8]) -> Result<(
...
+    page_table: &Arch::PageTable,
+    address: u64,
+    length: usize,
+) -> Result<usize, ()> {
+    // Check the page tables first.
+    let len = check_paging_range(page_table, address, length)?;
+
+    // Poke it with a stick. This will just check the first address
+    // to try to catch bogus memory ranges that are still mapped, which is common
+    // on debuggers initial breakpoint on the inherited page tables.
```
os-d:
```suggestion
    // on the debugger's initial breakpoint on the inherited page tables.
```

core/patina_debugger/src/arch/x64.rs
cfernald: Just the joys of x86 being a variable instruction length ISA. the MOV is only 3 bytes in length where as arm uses a fixed 32 bits.

cfernald:
> Approved, but did you test on physical HW? Would be good to confirm.

Tested on x64 but not aarch64 due to lack of device. I suspect there may be issues there as these "not real PA" issues can sometimes cause async errors, but in the end this is best effort as we are about to touch the memory anyways.

PR #684, opened 2025-08-11
Fix end of BootServices not reported.

magravel:
## Description

Before the end of exit boot services was not reported. This was caused by some status code not executed. After some investigation, I found that the status code listed was unregister in an event run at the end of boot services. This causes status code to set the end of exit boot services to not be called.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

A perf record was generated.

## Integration Instructions

N/A

### DIFF
--- a/docs/src/dxe_core/images.md
+++ b/docs/src/dxe_core/images.md
@@ -138,6 +138,9 @@ flowchart TD
   style image2 fill:#caa
 ```
 
+See the following compatibility note -
+[ConnectController() Must Explicitly Be Called For Handles Created/Modified During Image Start](../integrate/patina_dxe_core_requirements.md#33-connectcontroller-must-explicitly-be-called-for-handles-createdmodified-during-image-start).
+
 ## Exiting an Image
 
 The `exit()` routine is used to transfer control from the image execution context back to the `core_start_image` context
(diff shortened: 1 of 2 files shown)


sdk/patina_sdk/src/performance/measurement.rs
```diff
@@ -80,19 +84,31 @@ pub mod event_callback {
...
             efi::Guid::clone(&EDKII_FPDT_EXTENDED_FIRMWARE_PERFORMANCE),
             fbpt_address,
         );
-
         if status.is_err() {
             log::error!("Performance: Fail to report FBPT status code.");
         }
 
+        // Enable status code capability in Firmware Performance DXE.
+        let status = p.report_status_code(
```
makubacki: Why is this status code being reported from here?

magravel: I put it there because it needed to be called before the end of boot service. Do you have a better place to add this?

makubacki: This is PI Spec defined progress code that was previously being produced by the DXE Core right before BDS entry:

https://github.com/tianocore/edk2/blob/d46aa46c8361194521391aa581593e556c707c6e/MdeModulePkg/Core/Dxe/DxeMain/DxeMain.c#L582-L588

So, it's strange for such a generic broadcast code to be produced in this performance callback function. Can it just be produced before calling BDS entry in the Rust code?

The C FirmwarePerformanceDxe driver is one listener for the code. But that doesn't mean it is limited to performance usage only. Other listeners may hook on it.

PR #687, opened 2025-08-11
Create patina_sdk Signature macro.

apop5:
## Description
Add rust macro uefi_signature which can be used to create a u16, u32 or u64 signature, similar to
SIGNATURE_16, SIGNATURE_32 or SIGNATURE_64.

Ensure endianness of created signature.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Unit tests to verify endianness and that created signatures match 
expectations.

## Integration Instructions
No integration necessary.

### DIFF
--- a/sdk/patina_sdk/src/base.rs
+++ b/sdk/patina_sdk/src/base.rs
@@ -306,6 +306,68 @@ where
     let aligned_length = aligned_end - aligned_base;
     Ok((aligned_base, aligned_length))
 }
+
+/// Generates a UEFI-style signature from between 1 to 8 bytes, packing them into a u16, u32
+/// or u64 as appropriate for the parameters passed.
+///
+/// # Examples
+///
+/// ```rust
+/// use patina_sdk::signature;
+/// const SIG: u32 = signature!('A', 'B', 'C', 'D');
+/// assert_eq!(SIG, 0x44434241);
+/// ```
+///
+/// # Note
+/// This macro is typically used to create signatures for UEFI structures
+/// and is the equivalent of the SIGNATURE_16, SIGNATURE_32 and SIGNATURE_64
+/// macros from EDK2.
+#[allow(unused_macros)]
+#[macro_export]
+macro_rules! signature {
+    ($a:literal) => {
+        ($a as u16)
+    };
+    ($a:literal, $b:literal) => {
+        ($a as u16) | (($b as u16) << 8)
+    };
+    ($a:literal, $b:literal, $c:literal) => {
+        ($a as u32) | (($b as u32) << 8) | (($c as u32) << 16)
+    };
+    ($a:literal, $b:literal, $c:literal, $d:literal) => {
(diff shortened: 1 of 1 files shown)


sdk/patina_sdk/src/base.rs
```diff
@@ -306,6 +306,72 @@ where
     let aligned_length = aligned_end - aligned_base;
     Ok((aligned_base, aligned_length))
 }
+
+
+/// Generates a UEFI-style signature from up to 8 bytes, packing them into a u32 or u64 as appropriate.
+///
+/// # Examples
+///
+/// ```
```
Javagedes:
```suggestion
/// ```rust
```

sdk/patina_sdk/src/base.rs
```diff
@@ -306,6 +306,72 @@ where
...
+/// assert_eq!(SIG, 0x44434241);
+/// ```
+///
+/// # Note
+/// This macro is typically used to create signatures for UEFI structures.
+/// and is the equivalent of the SIGNATURE_16, SIGNATURE_32 and SIGNATURE_64
+/// macros from EDK2.
+#[allow(unused_macros)]
+#[macro_export]
+macro_rules! uefi_signature {
```
Javagedes: I have no real opinion here, mainly saying this to get others opinions. Do we need `uefi_` here? would `signature!` be just as clear, but more concise?

sdk/patina_sdk/src/base.rs
```diff
@@ -306,6 +306,72 @@ where
     let aligned_length = aligned_end - aligned_base;
     Ok((aligned_base, aligned_length))
 }
+
+
+/// Generates a UEFI-style signature from up to 8 bytes, packing them into a u32 or u64 as appropriate.
```
Javagedes: nit: per your code below it technically packs them into a `u8`, `u16`, `u32` or `u64` as appropriate

sdk/patina_sdk/src/base.rs
makubacki: I agree with `signature`. It's in` patina_sdk` so I'd consider the "Patina SDK signature macro".

sdk/patina_sdk/src/base.rs
```diff
@@ -361,4 +423,21 @@ mod tests {
 
         assert!(align_range(100u64, 100u64, 3u64).is_err()); // not power of two
     }
+    #[test]
```
makubacki, line 427: There are some cases missing in the unit test for an odd number of arguments ("1 -> u8", "3 -> u32", "5 -> u64").

apop5: I added the test cases, but started running into clippy errors.

In rust, a 'A' is a unicode scalar value, which is 4 bytes.

Looking back through uses of the signatures, there are not usecase for  a single byte 'signature', so I'm dropping the u8 case, and leaving it will u16, u32 and u64.

PR #688, opened 2025-08-12
Add MaxAddress capability to memory manager

berlin-with0ut-return:
## Description
Add the ability to allocate with a maximum address for the memory manager.

The immediate use case of this is to unblock ACPI component development, although it may be useful for other memory-restricted operations.

- [X] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [X] Includes tests?
- [ ] Includes documentation?

## How This Was Tested
Added a new allocation as part of `memory_manager_allocations_test`. Allocates and boots with no issues.

## Integration Instructions
N/A.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -71,3 +71,9 @@ opt-level = 3
 
 [profile.test]
 opt-level = 0
+debug = true
+debug-assertions = true
+overflow-checks = true
+lto = false
+incremental = false
+codegen-units = 1
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -6,7 +6,7 @@ RUSTC_BOOTSTRAP = 1
 NO_STD_FLAGS = "--profile ${RUSTC_PROFILE} -Zbuild-std=core,compiler_builtins,alloc -Zbuild-std-features=compiler-builtins-mem -Zunstable-options --timings=html"
 STD_FLAGS = "--profile ${RUSTC_PROFILE} --features std"
 TEST_FLAGS = { value = "--features enable_patina_tests", condition = { env_not_set = ["TEST_FLAGS"] } }
-COV_FLAGS = { value = "--features enable_patina_tests --workspace --out html --out xml --exclude-files **/tests/* --exclude-files **/benches/*", condition = { env_not_set = ["COV_FLAGS"] } }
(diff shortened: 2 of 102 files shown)


patina_dxe_core/src/memory_manager.rs
```diff
@@ -46,6 +46,10 @@ impl MemoryManager for CoreMemoryManager {
                 address = requested_address as efi::PhysicalAddress;
                 efi::ALLOCATE_ADDRESS
             }
+            PageAllocationStrategy::MaxAddress(max_address) => {
```
os-d, line 49: @cfernald originally had a range option here instead of mimicking the max addr/min addr logic, which doesn't fit all use cases (i.e. sometimes you need both greater than this and less than that). Chris, instead of this approach, should we just bring the range back? That would also support this use case.

cfernald: I was thinking about it some, I can't think of a scenario where someone actually cares if memory is "too low." obviously we should try to reserve <1MB for special cases, but that's an internal policy that the caller wouldn't care about. Can you think of any scenarios?

os-d: No, I thought you had a scenario :). But I guess that's why we nuked range in the first place. edk2 only uses a MinAddress to allocate within buckets which is also an internal policy. No complaints with this then.

cfernald: Yeah, exactly why I nuked it, got caught in the "oh this would be cool" without any good reason 😅

I think adding just MaxAddress to be amended later if needed seems reasonable

patina_dxe_core/src/memory_manager.rs
```diff
@@ -232,7 +236,16 @@ fn memory_manager_allocations_test(mm: Service<dyn MemoryManager>) -> patina_sdk
...
 
-    // Get an allocator
+    // Allocate with a max address limit.
+    let max_address = 0x1000_0000;
+    let result =
+        mm.allocate_pages(1, AllocationOptions::new().with_strategy(PageAllocationStrategy::MaxAddress(max_address)));
+    u_assert!(result.is_ok(), "Failed to allocate with max address limit.");
+    let allocation = result.unwrap();
+    let address = allocation.into_raw_ptr::<u8>().unwrap() as usize;
+    u_assert!(address <= max_address, "Allocated address exceeds max address limit.");
```
cfernald: You should check that the entire range is less then max_address. That is instead of `address <= max_address` it should be `(address + PAGE_SIZE - 1) <= max_address`

PR #689, opened 2025-08-12
[REBASE&FF] Resolve all Failing Boot Services SCTs

Javagedes:
## Description

This pull request resolves ~160 SCTs in the BootServices test suite. The errors are related to two core parts of patina as discussed in the commit comment.

1. Marking efi::UNUSABLE_MEMORY as an invalid memory type for allocations, which goes against specification
2. Returning incorrect status codes during close_protocol: 1. returning EFI_UNSUPPORTED when the protocol instance could not be found (expected: EFI_NOT_FOUND). Returning EFI_SUCCESS when the protocol instance was found, but no matching usage was found (expected: EFI_NOT_FOUND).

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

SCTs now pass and all CI tests still succeed.

## Integration Instructions

N/A

### DIFF
--- a/patina_dxe_core/src/protocol_db.rs
+++ b/patina_dxe_core/src/protocol_db.rs
@@ -409,9 +409,10 @@ impl ProtocolDb {
 
         let key = handle as usize;
         let handle_instance = self.handles.get_mut(&key).expect("valid handle, but no entry in self.handles");
-        let instance = handle_instance.get_mut(&OrdGuid(protocol)).ok_or(EfiError::Unsupported)?;
+        let instance = handle_instance.get_mut(&OrdGuid(protocol)).ok_or(EfiError::NotFound)?;
 
-        if let Some(idx) = instance
+        let mut status = Err(EfiError::NotFound);
+        while let Some(idx) = instance
             .usage
             .iter()
             .rposition(|x| (x.agent_handle == agent_handle) && (x.controller_handle == controller_handle))
@@ -427,10 +428,10 @@ impl ProtocolDb {
             if (usage.attributes & efi::OPEN_PROTOCOL_EXCLUSIVE) != 0 {
                 instance.opened_by_exclusive = false;
             }
-            Ok(())
-        } else {
-            Err(EfiError::NotFound)
+            status = Ok(());
         }
+
+        status
     }
 
     fn get_open_protocol_information_by_protocol(
--- a/patina_dxe_core/src/protocols.rs
+++ b/patina_dxe_core/src/protocols.rs
@@ -425,13 +425,9 @@ extern "efiapi" fn close_protocol(
         }
     };
 
-    // Remove all usages matching the specified controller and agent handles.
-    loop {
-        match PROTOCOL_DB.remove_protocol_usage(handle, unsafe { *protocol }, Some(agent_handle), controller_handle) {
-            Err(EfiError::NotFound) => return efi::Status::SUCCESS,
-            Err(err) => return err.into(),
-            Ok(_) => continue,
-        }
+    match PROTOCOL_DB.remove_protocol_usage(handle, unsafe { *protocol }, Some(agent_handle), controller_handle) {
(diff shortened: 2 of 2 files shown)


patina_dxe_core/src/allocator.rs
```diff
@@ -1386,11 +1380,6 @@ mod tests {
                 efi::Status::INVALID_PARAMETER
             );
 
-            assert_eq!(
```
makubacki, line 1391: Should this be modified to verify the allocation is successful now (instead of removing it from the unit test)?

makubacki: I guess the main ask is that a unit test actively confirms this behavior so we don't have to wait until the SCT is run to find a regression in the future. If that's done somewhere else, that's fine.

Javagedes: Makes sense. I've updated the tests to expect success now.

patina_dxe_core/src/protocol_db.rs
```diff
@@ -409,9 +409,10 @@ impl ProtocolDb {
 
         let key = handle as usize;
         let handle_instance = self.handles.get_mut(&key).expect("valid handle, but no entry in self.handles");
-        let instance = handle_instance.get_mut(&OrdGuid(protocol)).ok_or(EfiError::Unsupported)?;
+        let instance = handle_instance.get_mut(&OrdGuid(protocol)).ok_or(EfiError::NotFound)?;
 
-        if let Some(idx) = instance
+        let mut status = Err(EfiError::NotFound);
+        while let Some(idx) = instance
```
joschock, line 415: I'm not sure I understand the difference between `if let Some(idx)` and `while let Some(idx)` given that the right hand side should be an `Option<usize>` as opposed to an iterable. Can you elaborate on why the `while` is needed here?

Javagedes: `if let Some(idx)` was only calling the functionality once (if `Some(idx)` was returned), and you were relying on the outer loop calling `remove_protocol_usage` until there was a "not found" returned.

`while let Some(idx)` will call the inner handler multiple times on each individual `idx` until `rposition` returns `None` instead of a `Some(idx)`.

patina_dxe_core/src/allocator.rs
```diff
@@ -1570,6 +1564,16 @@ mod tests {
                 ),
                 efi::Status::INVALID_PARAMETER
             );
+
+            assert_eq!(
+                allocate_pages(
+                    efi::ALLOCATE_ANY_PAGES,
```
cfernald, line 1572: I wonder if we should artificially restrict that unusable memory can only be done through allocate by address. Is there a reason people would allocate unusable memory from anywhere?

os-d: The only expected use case is via a memory allocation HOB. But we also don't have a good grasp of how platforms are actually using this.

Javagedes: While I would say that is arguably the correct way to use it, and allocating any page does not make sense, per the spec and SCTs, restricting that would go against spec. I'm not against going against spec, but we would need to clearly document that we acknowledge this goes against spec, and these specific SCT failures are expected.

patina_dxe_core/src/protocol_db.rs
joschock: Ah - got it. it's the side effect of instance.usage.remove() that you want to hit repeatedly. Makes sense now.

patina_dxe_core/src/allocator.rs
Javagedes: We could add this as a restriction in the MemoryManager Service :)

cfernald: We can defer this conversation

PR #690, opened 2025-08-13
systemtables: Checksum on specific protocol install

Javagedes:
## Description

Adds a core component that registers an event that will re-checksum
the system tables when specific protocols (that would normally replace
functions in the system tables) are installed.

This fixes two SCT failures that were caused by the boot services
table and runtime services table having invalid checksums as we were
only performing a checksum on the tables before patina components or
UEFI spec compliant components were executed.
- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

N/A

## Integration Instructions

N/A

### DIFF
--- a/patina_dxe_core/src/systemtables.rs
+++ b/patina_dxe_core/src/systemtables.rs
@@ -725,22 +725,22 @@ pub fn register_checksum_on_protocol_install_events(
     }
 
     const GUIDS: [efi::Guid; 16] = [
-        efi::Guid::from_bytes(&uuid::uuid!("A46423E3-4617-49F1-B9FF-D1BFA9115839").to_bytes_le()), // gEfiSecurityArchProtocolGuid
+        efi::Guid::from_bytes(&uuid::uuid!("1DA97072-BDDC-4B30-99F1-72A0B56FFF2A").to_bytes_le()), // gEfiMonotonicCounterArchProtocolGuid
+        efi::Guid::from_bytes(&uuid::uuid!("1E5668E2-8481-11D4-BCF1-0080C73C8881").to_bytes_le()), // gEfiVariableArchProtocolGuid
         efi::Guid::from_bytes(&uuid::uuid!("26BACCB1-6F42-11D4-BC7E-0080C73C8881").to_bytes_le()), // gEfiCpuArchProtocolGuid
         efi::Guid::from_bytes(&uuid::uuid!("26BACCB2-6F42-11D4-BCE7-0080C73C8881").to_bytes_le()), // gEfiMetronomeArchProtocolGuid
         efi::Guid::from_bytes(&uuid::uuid!("26BACCB3-6F42-11D4-BCE7-0080C73C8881").to_bytes_le()), // gEfiTimerArchProtocolGuid
-        efi::Guid::from_bytes(&uuid::uuid!("665E3FF6-46CC-11D4-9A38-0090273FC14D").to_bytes_le()), // gEfiBdsArchProtocolGuid
-        efi::Guid::from_bytes(&uuid::uuid!("665E3FF5-46CC-11D4-9A38-0090273FC14D").to_bytes_le()), // gEfiWatchdogTimerArchProtocolGuid
-        efi::Guid::from_bytes(&uuid::uuid!("B7DFB4E1-052F-449F-87BE-9818FC91B733").to_bytes_le()), // gEfiRuntimeArchProtocolGuid
-        efi::Guid::from_bytes(&uuid::uuid!("1E5668E2-8481-11D4-BCF1-0080C73C8881").to_bytes_le()), // gEfiVariableArchProtocolGuid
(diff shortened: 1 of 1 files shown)


patina_dxe_core/src/systemtables.rs
```diff
@@ -714,6 +715,48 @@ pub fn init_system_table() {
     _ = SYSTEM_TABLE.lock().insert(table);
 }
 
+/// A component to register a callback that recalculates the CRC32 checksum of the system table
+/// when certain protocols are installed.
+pub fn register_checksum_on_protocol_install_events(
+    bs: patina_sdk::boot_services::StandardBootServices,
+) -> patina_sdk::error::Result<()> {
+    extern "efiapi" fn callback(_event: efi::Event, _: *mut c_void) {
+        SYSTEM_TABLE.lock().as_mut().expect("System Table is initialized").checksum_all();
```
os-d, line 724: Why is this a component? Can the protocol database not just update the checksum after installing one of these?

Javagedes: IMO separation of duties and it compartmentalizes this. While I'm open to it not being a component, I'm not a big fan of doing it directly in the protocol db. it seems weird to me to have the protocol DB update the system table checksum. I also don't like adding a "if guid in guids then checksum" for every installed protocol

os-d: My main concern is that it feels odd to me to have the core launch a component that then updates a part of the core. I'd rather have two parts of the core talk to each other to accomplish something (which doesn't have to be the protocol dabase directly updating this, it can be a call here or the system table listening).

I'm not completely opposed to this approach, it just concerns me if we are saying drivers can change core state directly. That seems to violate that separation. I missed that we have the core launching its own components at all, I'm a bit concerned with that approach in general. If we go to a kernel/userspace model how does this work, etc.

> I also don't like adding a "if guid in guids then checksum" for every installed protocol

This also does that, just in a more roundabout way

Javagedes: I don't have a huge opinion on the core registering its own components. I set it up, but I just saw it as a way to separate functionality in a clean manner. We can revert that, which moves these standalone components back underneath their associated `init_*` functions, which are directly called in `Core::initialize_system_table`.

I will say, as one note - The component is just registering a event on a protocol install, that updates the system table checksums. This is exactly what the C DXE core does today.

Again, open to that not being the case though. We can always do better :)

patina_dxe_core/src/systemtables.rs
```diff
@@ -714,6 +715,48 @@ pub fn init_system_table() {
...
+    }
+
+    const GUIDS: [efi::Guid; 16] = [
+        efi::Guid::from_bytes(&uuid::uuid!("A46423E3-4617-49F1-B9FF-D1BFA9115839").to_bytes_le()), // gEfiSecurityArchProtocolGuid
+        efi::Guid::from_bytes(&uuid::uuid!("26BACCB1-6F42-11D4-BC7E-0080C73C8881").to_bytes_le()), // gEfiCpuArchProtocolGuid
+        efi::Guid::from_bytes(&uuid::uuid!("26BACCB2-6F42-11D4-BCE7-0080C73C8881").to_bytes_le()), // gEfiMetronomeArchProtocolGuid
+        efi::Guid::from_bytes(&uuid::uuid!("26BACCB3-6F42-11D4-BCE7-0080C73C8881").to_bytes_le()), // gEfiTimerArchProtocolGuid
```
makubacki, line 732: Seeing these in some sense of order makes the poor choice of GUID generation more apparent. Nit, can you go ahead and just sort all the array members (so they're sorted by GUID)?

patina_dxe_core/src/systemtables.rs
makubacki: I don't know if this is really an issue. I like the separation of this to a component. It's clear from component inventory it's happening and it isolates protocol-specific logic which we want out of the "real core" anyway. It's also triggered by a global event to alter a global data table so it's not necessarily in the core or operating directly on core internals.

joschock: Not really opinionated about component vs. no, but wanted to note that not everyone is squeamish about updating the internal core ST checksums on a whim: https://github.com/microsoft/mu_basecore/blob/a9a137aaa1a9821ea83ea01faa78f1f509a5cb79/MdeModulePkg/Library/UefiBootManagerLib/BmConsole.c#L776

In any case, I think a triggered callback like this is a good way to do it, regardless of how it is installed. I'd second the argument for keeping the protocol_db unaware of the special meanings of any of the protocols it is managing.

os-d: I'll drop my complaints. I'm just hesitant about this approach because it allows deep coupling between components and the core. I would have just had the system table code register directly for notification, not through BS. But, other folks like it, I'll resolve this.

Issue #662, opened 2025-08-06
[Documentation]: Add debugging instructions to book


cfernald:
### Request Description

Include detail: min version 2507

### Are you going to make the change?

I will make the change

### Do you need maintainer feedback?

No maintainer feedback needed

PR #691, opened 2025-08-13
mdbook: Add developer pages for debugging

cfernald:
## Description

Adds developer pages detailing setting up the debugger and using windbg.

CLOSES: #662

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [x] Includes documentation?

## How This Was Tested

N/A

## Integration Instructions

N/A


docs/src/dev/debugging/windbg_debugging.md
```diff
@@ -0,0 +1,166 @@
+# Windbg Debugging
+
+Windbg is the primary debugging software recommended for use in Patina. While other
+debugging environments such as GDB do work, Windbg provides native support for
+PDB symbols and a robust bare-metal debugging experience. For this reason, the Patina
+team has invested most in this solution, so it has the most robust tooling.
```
apop5:
```suggestion
team has invested heavily in this solution, so it has the most robust tooling.
```
 There is a most later in the sentence

docs/src/dev/debugging/windbg_debugging.md
```diff
@@ -0,0 +1,166 @@
...
+
+If you frequently debug Patina, add `!uefiext.init` to the startup commands. This is
+found in `File->Settings->Debugging settings->Startup`. This will cause modules and
+symbols to be automatically resolved on connection.
+
+![Windbg Extension Startup](res/windbg_startup.png)
+
+### Installing the Extension
+
+To install the extension, download it from the most recent
```
apop5:
```suggestion
To install the extension, download it from the most recent
 release
```

docs/src/dev/debugging/windbg_debugging.md
```diff
@@ -0,0 +1,166 @@
...
+| `g`                | Resume execution                                       |
+| `p`                | Single step                                            |
+| `t`                | Step in                                                |
+| `gu`               | Step out                                               |
+| `dv`               | Display local variables                                |
+| `dx`               | Display object, used for globals                       |
+| `d[b/d/q]`         | Display bytes/dword/qword at provided address          |
+| `bp`               | Set a breakpoint at a provided address or symbol       |
+| `bl`               | List current breakpoints                               |
+| `ba`               | Set an access breakpoint                               |
```
apop5: Maybe add r for list registers?

cfernald: I knew I was forgetting something basic here!

docs/src/dev/debugging.md
```diff
@@ -0,0 +1,143 @@
...
+- **Unexpected Behavior** – Step through functions to analyze execution.
+
+### Advantages over a hardware debugger
+
+Hardware debuggers (JTAG) are powerful but need special hardware, configuration, and licenses.
+The self-hosted debugger is lightweight and tightly integrated with Patina, offering features like:
+
+- Breaking on module load
+- Module and symbol enumeration
+- Catching exceptions, panics, and asserts directly
```
Javagedes: custom debugger command registration

docs/src/dev/debugging/windbg_debugging.md
cfernald: I think the format of the suggestion got messed up. What were you asking for?

docs/src/dev/debugging.md
```diff
@@ -0,0 +1,143 @@
...
+```rust
+static DEBUGGER: patina_debugger::PatinaDebugger<UartPl011> =
+    patina_debugger::PatinaDebugger::new(UartPl011::new(0x6000_0000))
+        .without_transport_init()
+        .with_force_enabled(false);
+```
+
+### Step 2: Install the debugger
+
+In the platform `_start` routine, call `set_debugger` to install the debugger
```
Javagedes: nit `_start` is not required nomenclature, the start symbol just needs to be exported as `efi_main`. You may need to loosen this wording.

docs/src/dev/debugging.md
```diff
@@ -0,0 +1,143 @@
...
+```rust
+patina_debugger::set_debugger(&DEBUGGER);
+```
+
+Just because the debugger is installed, does not mean that the debugger is enabled
+or active. Installing is a no-op without enablement.
+
+### Step 3: Enable the debugger
+
+Enable the debugger at compile time with `.with_force_enabled(true)`. This causes Patina to
```
Javagedes: should probably mention the `with_exception_types` configuration here, or maybe in an "advanced" section.

docs/src/dev/debugging/windbg_debugging.md
apop5: That you specify to download the binary from from the releases, otherwise they might think they need to build the project.

docs/src/dev/debugging.md
```diff
@@ -0,0 +1,143 @@
...
+
+Monitor commands are interpreted by the Patina debugger. They allow dynamic actions from the
+debugger. Use `!monitor <command>` in WinDbg or `monitor <command>` in GDB. Useful commands:
+
+| Command     | Description                                           |
+|-------------|-------------------------------------------------------|
+| `help`      | Lists monitor commands                                |
+| `?`         | Shows debugger info and current break                 |
+| `mod`       | Module functions: list modules, break on load         |
+| `arch`      | Architecture-specific functions, e.g., dump registers |
```
Javagedes: I think we also have `gcd` and `system_table_ptr` built in by default, right? and platforms / components can add their own.

docs/src/dev/debugging/windbg_debugging.md
```diff
@@ -0,0 +1,168 @@
+# Windbg Debugging
+
+Windbg is the primary debugging software recommended for use in Patina. While other
+debugging environments such as GDB do work, Windbg provides native support for
+PDB symbols and a robust bare-metal debugging experience. For this reason, the Patina
```
Javagedes: It may be worth mentioning the uefi extension - described below - here also, as it is a benefit

docs/src/dev/debugging/windbg_debugging.md
```diff
@@ -0,0 +1,168 @@
...
+debugging environments such as GDB do work, Windbg provides native support for
+PDB symbols and a robust bare-metal debugging experience. For this reason, the Patina
+team has invested heavily in this solution, so it has the most robust tooling.
+
+More information and downloads can be found on the
+[Windbg learn page](https://learn.microsoft.com/en-us/windows-hardware/drivers/debugger/),
+or Windbg can be installed using the following winget command:
+
+```powershell
+winget install Microsoft.WinDbg
```
Javagedes: It might be worth mentioning NOT downloading the version from the microsoft store.

docs/src/dev/debugging/windbg_debugging.md
```diff
@@ -0,0 +1,168 @@
...
+
+After the script has started, you can connect the debugger to the forwarded port.
+
+## Connecting Windbg
+
+Windbg GDB remote support is provided through the EXDI interface. To connect through
+the UI, navigate to `Start debugging`, select `Attach to kernel`, and go to `EXDI`.
+This can be done faster using the `Ctrl+K` shortcut. From the EXDI tab, select the
+`UEFI` target type, your correct architecture, and the address:port you want to
+connect to, such as `localhost:5555`.
```
Javagedes: Even though it is probably obvious, it might be beneficial to show example of connecting to a system on the local network.

```suggestion
connect to, such as `localhost:5555`, `192.168.100.1:5555`, or `OTHER_NETWORKED_PC:5555`
```

docs/src/dev/debugging.md
```diff
@@ -0,0 +1,143 @@
...
+GDB also works, but symbols may not resolve since Patina uses PE images with PDB symbols.
+
+### Step 6: Set up the panic handler
+
+To break into the debugger on a panic, add a manual breakpoint in the panic handler. Only do this
+when the debugger is enabled:
+
+```rust
+if patina_debugger::enabled() {
+    patina_debugger::breakpoint();
```
Javagedes: It might be worth mentioning that this `patina_debugger::breakpoint()` would be a useful way to create a breakpoint exactly where you want, rather than needing to break in, init and find all the modules, find the address / name of the exact function you want to break in to, break into it, then step over to the exact position you want.

docs/src/dev/debugging/windbg_debugging.md
cfernald: Actually the store version is probably fine, but is a little behind. By time this goes public it should be fine. Internal is better because it just has more tools, but that isn't an option for reader here.

cfernald: I'm pretty sure the store and winget pull from the same place anyways

docs/src/dev/debugging.md
cfernald: I was wanting to avoid doing a full enumeration that needs to be updated. Instead was thinking more a sampling and recommending checking `!help`. I will clarify this.

docs/src/dev/debugging.md
```diff
@@ -0,0 +1,143 @@
...
+You can also enable the debugger at runtime using the `Configure` routine, but use caution.
+Runtime enablement can skip the initial breakpoint and may cause security issues. For development,
+prefer force enablement.
+
+### Step 4: Verify the transport
+
+After the initial breakpoint, monitor the debug port for:
+
+```text
+$T05thread:01;#07
```
os-d: This doesn't always come through. In fact, I think more often than not, I don't see it. So maybe noting that or adjusting the language to just describe what this is

docs/src/dev/debugging/windbg_debugging.md
```diff
@@ -0,0 +1,168 @@
...
+```
+
+It is recommended to use Windbg version >= 1.2507 for full support. There are known
+issues with older versions.
+
+## Serial Port Forwarding
+
+Windbg currently only supports connecting to the TCP-based GDB server. To support
+COM and named pipe-based transports, use the forwarding script
+[ComToTcpServer.py](https://github.com/microsoft/mu_feature_debugger/blob/main/Scripts/ComToTcpServer.py)
```
os-d: At some point we should host this and uefiext in Patina (generically a non-mu repo)

docs/src/dev/debugging.md
cfernald: If you are talking about Q35, then that is not the "debug port". Do you mean you dont see it on other platforms?

docs/src/dev/debugging/windbg_debugging.md
cfernald: Yeah, I was thinking about that. Not sure if its something we would want to do before opening or anything to avoid that association.

docs/src/dev/debugging.md
cfernald: Yeah, i'm wanting to avoid detailing too much code interfaces here that may just get out of sync, but also we dont have a way to link the rust docs yet.

docs/src/dev/debugging.md
makubacki: Yeah. I think the fact that code is authored specific to Patina DXE Core internals to aid debugging and that can be extended is a primary feature.

docs/src/dev/debugging.md
os-d: I think it is inconsistent on the physical platforms, too. I could be wrong.

makubacki:
@cfernald, have you reconciled this against https://github.com/OpenDevicePartnership/patina-qemu/blob/main/Platforms/Docs/Common/windbg-qemu-uefi-debugging.md?

docs/src/dev/debugging/windbg_debugging.md
cfernald: It does mention release, just on the next line in markdown

docs/src/dev/debugging.md
cfernald: I think I know what you are talking about. That is a tooling issue with our console software for some of our hardware devices and not something general. I might mention it's best to use putty to do this monitoring.

docs/src/dev/debugging/windbg_debugging.md
makubacki: How would you want to host it? I assume we'd want a single copy of debug collateral that is shared between Patina, Mu, and potentially others.

cfernald: Agreed. I don't want to split the functionality and most of the C dependencies in uefiext really come from EDKII and not MU. I think if we wanted to we could split this out into it's own repo of "uefi debugging tools" or maybe group together as "patina tools".

cfernald:
> @cfernald, have you reconciled this against https://github.com/OpenDevicePartnership/patina-qemu/blob/main/Platforms/Docs/Common/windbg-qemu-uefi-debugging.md?

I did look through that, and I like that as more of a specific case study on QEMU with QEMU specific details whereas here I'm trying to keep it generic and use-case driven.

docs/src/dev/debugging/windbg_debugging.md
cfernald: Going to close this, but we should discuss offline

os-d: Yeah, this wasn’t meant to hold up this PR, just as a general note to discuss. I also think one central repo that both Mu and Patina refer to. But I don’t care where it is hosted and ODP would be fine to me.

makubacki:
> > @cfernald, have you reconciled this against https://github.com/OpenDevicePartnership/patina-qemu/blob/main/Platforms/Docs/Common/windbg-qemu-uefi-debugging.md?
> 
> I did look through that, and I like that as more of a specific case study on QEMU with QEMU specific details whereas here I'm trying to keep it generic and use-case driven.

I think there are a few updates that could be made there for this content. Mainly just acknowledging that this exists and linking to it, but also removing any overlap in that document.

PR #694, opened 2025-08-15
image: bugfix unload_image

Javagedes:
## Description

Resolves a use-after-free error that originates in unload image. The
unload_image attempt to uninstall the image's loaded image protocol was
fallable due to not calling disconnect_controller first. In this
scenario, the loaded image protocol was not uninstalled, but the
protocol itself was freed, leading to a use-after-free error.

This resolves ~10K individual test failures, however it also reduces greatly reduces the amount of tests executed. This is because this test suite is run in two parts. The first part loads over 2500 images in different ways, then unloads them. The second half of the suite then runs validation tests on all of the loaded_image protocols. Due to the bug mentioned above, the 2500 images never had their loaded_image protocols removed (but they were de-allocated), so the test suite was running the test on each of the 2500 deallocated loaded_image protocols.

With this fix, the test suite now only runs on the ~200 images associated with the platform, and all 200 tests now pass.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

SCTs pass, unit tests pass

## Integration Instructions

N/A

### DIFF
--- a/patina_dxe_core/src/protocols.rs
+++ b/patina_dxe_core/src/protocols.rs
@@ -30,7 +30,7 @@ pub fn core_install_protocol_interface(
     protocol: efi::Guid,
     interface: *mut c_void,
 ) -> Result<efi::Handle, EfiError> {
-    log::info!("InstallProtocolInterface: {:?} @ {:#x?}", guid_fmt!(protocol), interface);
+    log::debug!("InstallProtocolInterface: {:?} @ {:#x?}", guid_fmt!(protocol), interface);
     let (handle, notifies) = PROTOCOL_DB.install_protocol_interface(handle, protocol, interface)?;
 
     let mut closed_events = Vec::new();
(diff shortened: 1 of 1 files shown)


patina_dxe_core/src/protocols.rs
```diff
@@ -72,35 +72,31 @@ extern "efiapi" fn install_protocol_interface(
...
+    protocol: efi::Guid,
     interface: *mut c_void,
-) -> efi::Status {
-    if protocol.is_null() {
-        return efi::Status::INVALID_PARAMETER;
-    }
-
-    let caller_protocol = *(unsafe { protocol.as_mut().expect("previously null-checked pointer is null") });
+) -> Result<(), EfiError> {
+    log::info!("UninstallProtocolInterface: {:?} @ {:#x?}", guid_fmt!(protocol), interface);
```
makubacki: Do you think this needs to remain an info level message?

Javagedes: This is a new log statement I added to mirror the existing `log::info!` for installing a protocol interface.

I'd be happy to move them both to debug (or only my newly added one) if you like. I don't have a preference :)

Javagedes: I've updated both the install and uninstall log statements to be debug :)

PR #702, opened 2025-08-22
[REBASE&FF] Always produce UefiDecompress protocol / section extractor

Javagedes:
## Description

This PR converts the `SectionExtractor` trait to be consumed as a service rather than as a special method in the Core. This makes the section extractor optional, so a core section extractor implementation is now provided which always provide Tiano / UEFI decompression functionality. Registering a `SectionExtractor` implementation will add additional capabilities to the core section extractor.

In addition to this, the EfiDecompress protocol is now always produced by the core, in compliance with the SCTs

- [ ] Impacts functionality?
- [ ] Impacts security?
- [x] Breaking change?
- [ ] Includes tests?
- [x] Includes documentation?

## How This Was Tested

Continued boot with Q35

## Integration Instructions

1. `patina_section_extractor::NullSectionExtractor` has been removed, as it is the same as not registering the service.

2. platforms must now register their section extractor with `.with_service(<extractor>)` or `.with_component(<comp_to_produce_service>)` instead of relying on the default implementation when initializing the core. If a platform is using `.with_section_extractor(...)`, it must be removed. If they are using `.with_section_extractor(...)` with a section extractor from `patina_section_extractor`, then they can simply move it after the `init_memory` method and update it to `.with_service(...)`

```rust
// Old
Core::default()
  .with_section_extractor(patina_section_extactor::CompositeSectionExtractor::default())
  .init_memory(physical_hob_list)
  .start()
```

```rust
// New
Core::default()
  .init_memory(physical_hob_list)
  .with_service(patina_section_extractor::CompositeSectionExtractor::default())
  .start()
```

### DIFF
--- a/docs/src/SUMMARY.md
+++ b/docs/src/SUMMARY.md
@@ -34,9 +34,9 @@
 
 - [Patina DXE Core Requirements](integrate/patina_dxe_core_requirements.md)
 - [Setting up the Patina DXE Core](integrate/dxe_core.md)
-- [Integrating an Externally Compiled EFI Binary to the Platform](integrate/platform_external.md)
 - [Compiling External to the Platform](integrate/compile_external.md)
-- [Compiling Local to the Platform (Deprecated)](integrate/platform_local.md)
+- [Compiling Local to the Platform (Deprecated)](integrate/compile_local.md)
+- [Integrating an Externally Compiled EFI Binary to the Platform](integrate/platform_external.md)
 
 # Patina Component Model
 
--- a/docs/src/integrate/dxe_core.md
+++ b/docs/src/integrate/dxe_core.md
@@ -101,9 +101,9 @@ not meant to be exhaustive, but rather a starting example for showing how to set
 `SectionExtractor` is an abstraction point that allows a platform to specify the specific section extraction methods it
 supports. As an example, a platform may only compress its sections with brotli, so it only needs to support brotli
 extractions. A platform may create their own extractor; it only needs to implement the
-[SectionExtractor](https://github.com/microsoft/mu_rust_pi/blob/c8dd7f990d87746cfae9a5e821ad69501c46f346/src/fw_fs.rs#L77)
-trait. However, multiple implementations are provided via [patina_ffs_extractors](https://github.com/OpenDevicePartnership/patina/tree/main/core/patina_ffs_extractors),
-such as brotli, crc32, uefi_decompress, etc.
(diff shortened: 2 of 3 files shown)


patina_dxe_core/src/fv.rs
```diff
@@ -45,7 +45,7 @@ enum PrivateDataItem {
 
 struct PrivateGlobalData {
     fv_information: BTreeMap<*mut c_void, PrivateDataItem>,
-    section_extractor: Option<Box<dyn SectionExtractor>>,
+    section_extractor: Option<Service<dyn SectionExtractor>>,
```
os-d: I'm a bit concerned about having the core rely on functionality from a component: that was one of the major design decisions we said we didn't want from edk2 and have done some large changes to make that the case.

Now, maybe I'm misunderstanding the component/service part here. In some places this is referred to as a component, in some places as a service. If this is code that lives in the core and just produces a service, that is great and makes a lot of sense to me, like in the memory code. But is this actually running as a component?

Javagedes: I can see the confusion because of the naming, and I agree it needs some re-wording. I originally updated each of the section extractors to be components where only thing the component did was register itself as a service. In my head, I thought that was dumb. Therefore, I exposed a `with_service` function from the core where a platform can directly register a service during initialization, instead of dispatching a component to do it for you.

With that said, this is purely a service providing optional extra Section Extraction functionality. The core, before dispatching, checks to see if the platform has custom section extraction functionality; if it does, it will use it (so think our NULL libs to attach extra supported decompression algorithms in the core).

In a following update (assuming we don't nix this implementation), This is further shown because the `section_extractor` variable gets updated to be a `CoreExtractor` where the uefi/tiano decompress is always present, and the platform can elect to attach additional decompression support.

patina_dxe_core/src/lib.rs
```diff
@@ -281,25 +246,28 @@ where
...
-{
+impl Core<Alloc> {
     /// Registers a component with the core, that will be dispatched during the driver execution phase.
     #[inline(always)]
     pub fn with_component<I>(mut self, component: impl IntoComponent<I>) -> Self {
         self.insert_component(self.components.len(), component.into_component());
         self
     }
 
+    /// Registers a service with the core, that will be available to all monolithic components.
```
makubacki: If `with_service` is kept, can you please update the component documentation to state the design rationale and guidelines for when it should be used?

Javagedes: See #703 🙂

Javagedes:
Putting this on hold to wait for #706 to complete.

PR #703, opened 2025-08-25
[REBASE&FF] Allow services to be registered with the core

Javagedes:
## Description

This pull request allows services that have no dependencies to be directly registered with the core during Core setup, which prevents the need for a wrapper component to simply instantiate and register a component. This makes the service immediately available and reduces the number of components that must be executed. This supports the ability for the core to directly take soft dependencies on services.

Additionally, This PR adds documentation to clarify that if a service can be directly registered with the core, the service implementation should go in the crates `service` module.

Additionally, this adds more documentation on Components and services to clarify to what a component is, what makes up a component, and how a component differs from configuration and services.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

N/A

## Integration Instructions

N/A

### DIFF
--- a/docs/src/component/getting_started.md
+++ b/docs/src/component/getting_started.md
@@ -1,0 +1,74 @@
+# Getting Started with Components
+
+Components are the mechanism used to attach additional functionality to the Patina core while keeping each piece
+decoupled from the others. In a systems programming context, components are akin to drivers. Patina uses dependency
+injection through each component's entry point function. The component's entry point defines what it requires before it
+can run. As an example, a component whose interface is `fn entry_point() -> Result<()>` has **no dependencies** where
+as a component with the interface `fn entry_point(service: Service<dyn Interface>) -> Result<()>` has a dependency on
+the `dyn Interface` service being registered. The latter service will not execute until this service has been produced.
+
+This architecture ensures:
+
+- Loose coupling - Components do not depend directly on each other.
+- Explicit dependencies - Components declare the services or configuration they need.
+- Flexibility - It does not matter who provides a service or sets a configuration, only that it is available.
+- Maintainability - Each component can be developed and maintained independently.
+
+At runtime, the dependency injection model allows the patina core to track which component provides which service(s)
+and identify the missing dependency that is preventing any given component from running, making it easy to determine
+which components are not executing and why.
+
+When it comes to understanding components and how they interact with each other, there are three main topics that you
+must understand - (1) Components, (2) Configuration, and (3) Services. Each will be discussed below in broad scopes,
+but more details can be found in this mdbook, and in the component documentation for [patina_sdk](https://github.com/OpenDevicePartnership/patina/tree/main/sdk/patina_sdk).
+
+## Components
+
+As mentioned above, Components are a way to attach additional functionality to the core that is executed in a
+controlled manner based off of the component function interface. Components can be used to set configuration, create
+services, communicate with physical devices, and many other things. The components section of the [patina_sdk](https://github.com/OpenDevicePartnership/patina/tree/main/sdk/patina_sdk)
+goes into much more detail regarding components.
+
+## Configuration
+
+Configuration comes in two types of flavors - public configuration and private configuration. As the name suggests,
+public configuration can be accessed by any executing component, and is typically set by the platform to generically
+configure multiple components at once. Private configuration is configuration set when registering the component with
+the core.
+
+Public configuration is consumed by components by using the `Config<Type>` in the function interface for the component
+(See [Interface](https://opendevicepartnership.github.io/patina/component/interface.html#component-params)).
+Configuration is typically set by the platform using the [Core::with_config](https://github.com/OpenDevicePartnership/patina),
+however each configuration type must implement `Default`, so configuration will always be available to components.
+
+Private configuration is the configuration set when instantiating a component that is registered with the core using
+[Core::with_component](https://github.com/OpenDevicePartnership/patina). Not all components will have private
+configuration; it depends on the component implementor and the needs of the component.
+
+## Services
+
+Services are the mechanism in which to share functionality between components via a well-defined interface (trait)
+while allowing the underlying implementation to have different implementations per platform. This enables platforms
+to switch implementations without directly breaking any other components that depend on that functionality. Services
+may be registered by the core itself, by components, or by the platform via the [Core::with_service](https://github.com/OpenDevicePartnership/patina)
+during Core setup. See the Patina [Service](https://opendevicepartnership.github.io/patina/component/interface.html#servicet)
+or [Patina Sdk](https://github.com/OpenDevicePartnership/patina) crate documentation for more information.
+
+``` admonish note
+The core may take an **optional** dependency on some services. These services will be directly communicated in the
+inline documentation from the core and must be directly registered using [Core::with_service](https://github.com/OpenDevicePartnership/patina).
+If not, there is no guarantee that the service will be available before the core needs it. The core must be able to
+operate (albeit with potentially reduced capabilities) if no services are provided via [Core::with_service](https://github.com/OpenDevicePartnership/patina).
+
+If the core requires platform-specific functionality mandatory for **core** operation, it will be enforced via
+mechanisms other than the [Core::with_service] as missing services can only be determined at runtime. Typically this
+will involve using an API exposed from the core that will cause a build break if a platform fails to provide the
+required functionality. 
+```
+
+Services can be registered and made available to components in a few different ways. The first way is that the core
+itself produces some services directly, such as the Memory Manager. This is a way to expose controlled access to
+internal functionality. The second is that a service can be registered directly with a core using [Core::with_service](https://github.com/OpenDevicePartnership/patina).
+This is only available for services that have no external dependencies, and can be instantiated directly. Finally, a
+component can register a service by using [Storage::add_service](https://github.com/OpenDevicePartnership/patina),
+which is used when a service has a dependency, be it another service, configuration, or something else.
--- a/docs/src/SUMMARY.md
+++ b/docs/src/SUMMARY.md
@@ -38,6 +38,12 @@
 - [Compiling External to the Platform](integrate/compile_external.md)
 - [Compiling Local to the Platform (Deprecated)](integrate/platform_local.md)
 
+# Patina Component Model
+
+- [Getting Started with Components](component/getting_started.md)
+- [Component Crate Requirements](component/requirements.md)
+- [Component Interface](component/interface.md)
+
 # Patina DXE Core Subsystems
 
 - [Theory and Operation](dxe_core/operation.md)
@@ -53,11 +59,6 @@
   - [Synchronization](dxe_core/synchronization.md)
   - [Testing](dxe_core/testing.md)
   - [UEFI Driver Model](dxe_core/driver_model.md)
-
-# Patina Component Model
-
-- [Component Crate Requirements](component/requirements.md)
-- [Component Interface](component/interface.md)
 
 # Component Documentation
 
--- a/docs/src/component/interface.md
+++ b/docs/src/component/interface.md
@@ -148,6 +148,11 @@ latter as it simplifies mocking functionality for host-based unit tests, however
 two main drawbacks are (1) functionality is accessed via a v-table, causing some performance degradation, and (2) dyn
 trait objects do not support generics in their function interfaces.
 
+Service implementations can be registered in two distinct ways. If the service implementation can be directly
+instantiated with no other dependencies, then it can be registered directly during core instantiation with [Core::with_service](https://github.com/OpenDevicePartnership/patina),
+which makes the service immediately available. If the service implementation does have dependencies, then a component
+will be used to request those dependencies, instantiate the service, and pass it to the storage with [Storage::add_service](https://github.com/OpenDevicePartnership/patina).
+
 If function generics are needed / wanted, it is suggested that most functionality be provided via a typical, mockable
 trait object service, with a lightweight concrete struct Service Wrapper to support generics. This allows for easy
 mocking of the underlying functionality, but provides an easy to use interface as seen below:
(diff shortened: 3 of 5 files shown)


docs/src/component/getting_started.md
```diff
@@ -0,0 +1,52 @@
+# Getting Started with Components
+
+Components are a way to attach additional functionality to the Patina core while keeping each them decoupled from
+eachother. This makes it simple to maintain each component separately, understand it's dependencies, and allow any
```
os-d:
```suggestion
Components are a way to attach additional functionality to the Patina core while keeping them decoupled from
each other. This makes it simple to maintain each component separately, understand it's dependencies, and allow any
```

It also kind of feels like we are dancing around saying a component is a driver. Which it is. I think we may want to explicitly say that. It puts it in the frame of not just edk2, but also system development where you have a core, drivers, and applications.

docs/src/component/getting_started.md
```diff
@@ -0,0 +1,52 @@
...
+given platform to easily pick and choose the components necessary for it. The Patina core uses dependency injection
+via the component's entry point to fully define it's dependencies to successfully execute. e.g. a component registered
+with `entry_point() -> Result<()>` has no dependencies where as `entry_point(service: Service<dyn Interface>)` will
+not be executed until the `dyn Interface` service is registered. This makes it simple to understand what functionality
+must exist for a component to execute and debugging / logging what functionality is missing at runtime that is
```
os-d:
```suggestion
must exist for a component to execute and to debug/log what functionality is missing at runtime that is
```

docs/src/component/getting_started.md
```diff
@@ -0,0 +1,52 @@
...
+preventing a component from executing.
+
+When it comes to understanding components and how they interact with eachother, there are three main topics that you
```
os-d:
```suggestion
When it comes to understanding components and how they interact with each other, there are three main topics that you
```

docs/src/component/getting_started.md
```diff
@@ -0,0 +1,52 @@
...
+must understand - (1) Components, (2) Configuration, and (3) Services. Each will be discussed below in broad
+scopes, but more details can be found in this document, and in the component documentation for [patina_sdk](https://github.com/OpenDevicePartnership/patina/tree/main/sdk/patina_sdk).
```
os-d:
```suggestion
scopes, but more details can be found in this mdbook, and in the component documentation for [patina_sdk](https://github.com/OpenDevicePartnership/patina/tree/main/sdk/patina_sdk).
```
Or something? Saying this document makes it seem like this specific file contains more info

docs/src/component/getting_started.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Components
+
+As mentioned above, Components are a way to attach additional functionality to the core that is executed in
+a controlled manner based off of the component function interface. It is easiest to think of a component as an EDKII
+driver, however it is more than that. Components can be used to set configuration, create services, communicate with
```
os-d: As noted above, I think we can make the comparison to edk2 drivers, but again a driver is a systems programming context, I think we can frame it in that manner as opposed to just that it isn't an edk2 driver, but kind of is.

docs/src/component/getting_started.md
```diff
@@ -0,0 +1,52 @@
...
+the core.
+
+Public configuration is consumed by components by using the `Config<Type>` in the function interface for the component
+(See [Interface](https://opendevicepartnership.github.io/patina/component/interface.html#component-params)).
+Configuration is typically set by the platform using the [Core::with_config](https://github.com/OpenDevicePartnership/patina),
+however each configuration type must implement `Default`, so configuration will always be available to components.
+
+Private configuration is the configuration set when instantiating a component that is registered with the core using
+[Core::with_component](https://github.com/OpenDevicePartnership/patina). Not all components will have private
+configuration; it depends on the component implementor and needs of the component.
```
os-d:
```suggestion
configuration; it depends on the component implementor and the needs of the component.
```

docs/src/component/getting_started.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Services
+
+Services are how components share functionality with eachother. Services are a way to abstract functionality by having
```
os-d:
```suggestion
Services are how components share functionality with each other. Services are a way to abstract functionality by having
```

Also, this statement isn't completely true, right? The core can also produce services, as you note below

docs/src/component/getting_started.md
```diff
@@ -0,0 +1,52 @@
...
+a well-defined interface (using a trait), and allowing the underlying implementation to differ, without affecting
+consumers of the functionality. See the Patina [Service](https://opendevicepartnership.github.io/patina/component/interface.html#servicet)
+documentation, or the crate documentation for more information.
+
+Services can be registered and made available to components a few different ways. The first way is that the core itself
```
os-d:
```suggestion
Services can be registered and made available to components in a few different ways. The first way is that the core itself
```

docs/src/component/getting_started.md
```diff
@@ -0,0 +1,52 @@
...
+produces some services directly, such as the Memory Manager. This is a way to expose some limited internal
```
os-d: Why does this need to be limited?

Javagedes: Limited in the sense that we don't want to expose direct access to the GCD or to other purely internal mechanisms. The service interface allows us to expose a controlled subset of internal functionality. I've reworded a bit. Let me know if it needs additional changes.

docs/src/component/getting_started.md
Javagedes: Fair - I've updated to specify that Servies are a mechanism to access internal functionality from the core or to share functionality between components.

docs/src/component/getting_started.md
Javagedes: Makes sense. I've completely re-worked the first paragraph and specifically called it a driver. Feel free to re-read and provide any additional comments in a new conversation :)

docs/src/component/getting_started.md
```diff
@@ -0,0 +1,64 @@
...
+
+Private configuration is the configuration set when instantiating a component that is registered with the core using
+[Core::with_component](https://github.com/OpenDevicePartnership/patina). Not all components will have private
+configuration; it depends on the component implementor and the needs of the component.
+
+## Services
+
+Services are the mechanism to access internal functionality from the core or to share functionality between components.
+Services are a way to abstract functionality by having a well-defined interface (using a trait) while the underlying
+implementation can have different implementations per platform, without effecting consumers of the functionality. See
```
joschock:
```suggestion
implementation can have different implementations per platform, without affecting consumers of the functionality. See
```

docs/src/component/getting_started.md
```diff
@@ -0,0 +1,64 @@
...
+the Patina [Service](https://opendevicepartnership.github.io/patina/component/interface.html#servicet) documentation,
+or the crate documentation for more information.
+
+``` admonish note
+The core may take a soft dependency on some services. These services will be directly communicated in the inline
+documentation from the core and must be directly registered with the [Core::with_service](https://github.com/OpenDevicePartnership/patina).
+If not, there is no guarantee that the service will be available before the core needs it.
+
+A hard dependency, e.g. required dependency, will never be made on a service. Any external functionality that creates
```
joschock: I find this sentence a bit hard to follow; consider rewording?

docs/src/component/getting_started.md
```diff
@@ -0,0 +1,64 @@
+# Getting Started with Components
+
+Components are a way to attach additional functionality to the Patina core while keeping each decoupled from each
+other. In a systems programming context, components are the equivalent of drivers. The Patina core uses dependency
+injection via the component's entry point to fully define it's dependencies to successfully execute. As an example, a
```
makubacki:
```suggestion
injection via the component's entry point to fully define its dependencies to successfully execute. As an example, a
```

docs/src/component/getting_started.md
```diff
@@ -0,0 +1,64 @@
...
+component registered with `entry_point() -> Result<()>` has no dependencies where as
+`entry_point(service: Service<dyn Interface>)` has a dependency on a the `dyn Interface` service being implemented and
+registered. This component will not be executed until this happens.
+
+This design makes it simple to maintain each component as separate entities, agnostic and independent from eachother.
```
makubacki: I think this was done elsewhere already.

```suggestion
This design makes it simple to maintain each component as separate entities, agnostic and independent from each other.
```

docs/src/component/getting_started.md
```diff
@@ -0,0 +1,64 @@
+# Getting Started with Components
+
+Components are a way to attach additional functionality to the Patina core while keeping each decoupled from each
+other. In a systems programming context, components are the equivalent of drivers. The Patina core uses dependency
+injection via the component's entry point to fully define it's dependencies to successfully execute. As an example, a
+component registered with `entry_point() -> Result<()>` has no dependencies where as
+`entry_point(service: Service<dyn Interface>)` has a dependency on a the `dyn Interface` service being implemented and
```
makubacki:
```suggestion
`entry_point(service: Service<dyn Interface>)` has a dependency on the `dyn Interface` service being implemented and
```

docs/src/component/getting_started.md
makubacki: There are a few other places this could be updated too.

docs/src/component/getting_started.md
makubacki: This is a little confusing because they may not be independent from each other depending on what another component does like produce a service that another component depends on. Technically, the component is dependent on the service, not the other component, but I still think this should describe components more as "independent sets of functionality" as opposed independence from each other.

docs/src/component/getting_started.md
```diff
@@ -0,0 +1,64 @@
...
+## Configuration
+
+Configuration comes in two types of flavors - public configuration and private configuration. As the name suggests,
+public configuration can be accessed by any executing component, and is typically set by the platform to generically
+configure multiple components at once. Private configuration is configuration set when registering the component with
+the core.
+
+Public configuration is consumed by components by using the `Config<Type>` in the function interface for the component
+(See [Interface](https://opendevicepartnership.github.io/patina/component/interface.html#component-params)).
+Configuration is typically set by the platform using the [Core::with_config](https://github.com/OpenDevicePartnership/patina),
```
makubacki, line 41: Are these links meant to go back to the main repo ("https://github.com/OpenDevicePartnership/patina")?

docs/src/component/getting_started.md
makubacki: I think just sticking to "required" and "optional" instead of mixing in "hard" and "soft" would be clearer.

docs/src/component/getting_started.md
```diff
@@ -0,0 +1,64 @@
...
+however each configuration type must implement `Default`, so configuration will always be available to components.
+
+Private configuration is the configuration set when instantiating a component that is registered with the core using
+[Core::with_component](https://github.com/OpenDevicePartnership/patina). Not all components will have private
+configuration; it depends on the component implementor and the needs of the component.
+
+## Services
+
+Services are the mechanism to access internal functionality from the core or to share functionality between components.
```
makubacki: This is worded in a way that it seems like services could be interpreted to "break abstractions by allowing outsiders to access internals". Was something like "Services abstract functionality behind a well-defined interface and can be provided by the core or components" intended?

docs/src/component/getting_started.md
Javagedes: Fixed in next push. Thanks

docs/src/component/getting_started.md
Javagedes: Fixed in next push. Thanks!

docs/src/component/getting_started.md
Javagedes: Yes, they are basically placeholders until we go open source and I can start linking against docs.rs. I can do something else if you prefer, just let me know. If this is OK, then close the issue :)

docs/src/component/getting_started.md
Javagedes: Fixed in next push. Thanks!

docs/src/component/getting_started.md
Javagedes: Updated to:

```
If the core **requires** platform specific functionality to exist, then the core will expose a mechanism to expose this
requirement such that a compilation failure will occur if the functionality is not provided. What this means is that
required platform specific functionality will **never** be registered via services or [Core::with_service](https://github.com/OpenDevicePartnership/patina).
```

If this is still needs work, then please re-open this issue, or open another :)

docs/src/component/getting_started.md
Javagedes: Updated

docs/src/component/getting_started.md
Javagedes: fixed in next push. Thanks!

docs/src/component/getting_started.md
joschock: I understand it better now, though:

>If the core requires platform-specific functionality to exist, the core will expose a mechanism to expose this requirement such that compilation failure will occur if functionality is not provided

-> that's a lot of exposes. 

How about "If the core has a hard requirement on the platform to provide a platform-specific implementation of functionality mandatory for core operation, that will be enforced via mechanisms other than the [Core::with_service] mechanism described here, typically using an API that will cause a build break if a platform fails to provide the required functionality. The core must be able to operate (albeit with potentially reduced capabilities) if no services are provided via [Core::with_serivce]."

docs/src/component/getting_started.md
Javagedes: I completely re-wrote this section :)

docs/src/component/getting_started.md
Javagedes: Updated!

Javagedes:
@makubacki @os-d if you could squeeze in a final review today for this, that would be awesome. I'd like to get my uefi-decompress PR finalized

PR #706, opened 2025-08-27
Refactor FFS support per RFC 14

joschock:
## Description

Initial draft PR for FFS refactor based on [RFC 14](https://github.com/OpenDevicePartnership/patina/pull/522)

- [x] Impacts functionality?
- [ ] Impacts security?
- [x] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Passes all unit tests, tested boot on hardware platform.

## Integration Instructions

To follow: main change is instantiation of section extractor.

(skip cargo release dry run)

### DIFF
--- a/patina_dxe_core/src/dispatcher.rs
+++ b/patina_dxe_core/src/dispatcher.rs
@@ -344,22 +344,16 @@ fn add_fv_handles(new_handles: Vec<efi::Handle>) -> Result<(), EfiError> {
                 if file.file_type_raw() == ffs::file::raw::r#type::DRIVER {
                     let file = file.clone();
                     let file_name = file.name();
-                    let sections = {
-                        if let Some(extractor) = &dispatcher.section_extractor {
-                            file.sections_with_extractor(extractor.as_ref())?
-                        } else {
-                            file.sections()?
-                        }
+                    let sections = match &dispatcher.section_extractor {
+                        Some(extractor) => file.sections_with_extractor(extractor.as_ref())?,
+                        None => file.sections()?,
                     };
 
                     let depex = sections
                         .iter()
-                        .find_map(|x| {
-                            if x.section_type() == Some(ffs::section::Type::DxeDepex) {
-                                Some(x.try_content_as_slice())
-                            } else {
-                                None
-                            }
+                        .find_map(|x| match x.section_type() {
+                            Some(ffs::section::Type::DxeDepex) => Some(x.try_content_as_slice()),
+                            _ => None,
                         })
                         .transpose()?
                         .map(Depex::from);
@@ -431,22 +425,16 @@ fn add_fv_handles(new_handles: Vec<efi::Handle>) -> Result<(), EfiError> {
                     let file = file.clone();
                     let file_name = file.name();
 
-                    let sections = {
-                        if let Some(extractor) = &dispatcher.section_extractor {
-                            file.sections_with_extractor(extractor.as_ref())?
-                        } else {
-                            file.sections()?
-                        }
+                    let sections = match &dispatcher.section_extractor {
+                        Some(extractor) => file.sections_with_extractor(extractor.as_ref())?,
+                        None => file.sections()?,
                     };
 
                     let depex = sections
                         .iter()
-                        .find_map(|x| {
-                            if x.section_type() == Some(ffs::section::Type::DxeDepex) {
-                                Some(x.try_content_as_slice())
-                            } else {
-                                None
-                            }
+                        .find_map(|x| match x.section_type() {
+                            Some(ffs::section::Type::DxeDepex) => Some(x.try_content_as_slice()),
+                            _ => None,
                         })
                         .transpose()?
                         .map(Depex::from);
@@ -559,7 +547,7 @@ mod tests {
     use super::*;
     use crate::test_collateral;
 
-    // Sample logger for log crate to dump stuff in tests
+    // Simple logger for log crate to dump stuff in tests
     struct SimpleLogger;
     impl log::Log for SimpleLogger {
         fn enabled(&self, metadata: &Metadata) -> bool {
--- a/sdk/patina_ffs/src/file.rs
+++ b/sdk/patina_ffs/src/file.rs
@@ -6,6 +6,13 @@
 //!
 //! Both types help inspect file metadata (name GUID, type, attributes), traverse sections, and
 //! serialize with correct headers, checksums, and alignment.
+//!
+//! ## License
+//!
+//! Copyright (C) Microsoft Corporation.
+//!
+//! SPDX-License-Identifier: BSD-2-Clause-Patent
+//!
 use mu_pi::fw_fs::{
     ffs::{self, attributes, file},
     fv,
@@ -16,7 +23,7 @@ use crate::{
     section::{Section, SectionComposer, SectionExtractor, SectionIterator},
 };
 
-use alloc::{vec, vec::Vec};
+use alloc::vec::Vec;
 use core::{fmt, iter, mem, ptr, slice::from_raw_parts};
 use r_efi::efi;
 
@@ -44,7 +51,7 @@ impl<'a> FileRef<'a> {
     /// - [`FirmwareFileSystemError::InvalidState`]: file state not DATA_VALID.
     /// - [`FirmwareFileSystemError::DataCorrupt`]: data checksum mismatch.
     ///
-    /// Examples
+    /// ## Examples
     ///
     /// ```rust no_run
     /// use patina_ffs::file::{File, FileRef};
@@ -78,9 +85,9 @@ impl<'a> FileRef<'a> {
         let (size, content_offset) = {
             if (header.attributes & attributes::raw::LARGE_FILE) == 0 {
                 //standard header with 24-bit size.
-                let mut size = vec![00u8; 4];
+                let mut size = [00u8; 4];
                 size[0..3].copy_from_slice(&header.size);
-                let size = u32::from_le_bytes(size.try_into().unwrap()) as usize;
+                let size = u32::from_le_bytes(size) as usize;
                 (size, mem::size_of::<file::Header>())
             } else {
                 //extended header with 64-bit size.
@@ -221,7 +228,7 @@ impl<'a> FileRef<'a> {
     ///
     /// Useful for decoding compressed or encapsulated sections during traversal.
     ///
-    /// Examples
+    /// ##Examples
     ///
     /// ```rust no_run
     /// use mu_pi::fw_fs::ffs;
@@ -306,7 +313,7 @@ impl File {
     /// required padding between sections, and computes header/data checksums.
     /// Errors from section serialization are propagated.
     ///
-    /// Examples
+    /// ## Examples
     ///
     /// ```rust no_run
     /// use mu_pi::fw_fs::ffs;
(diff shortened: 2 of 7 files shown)


patina_dxe_core/src/dispatcher.rs
```diff
@@ -415,35 +427,33 @@ fn add_fv_handles(new_handles: Vec<efi::Handle>) -> Result<(), EfiError> {
...
                         .find_map(|x| {
-                            if x.section_type() == Some(FfsSectionType::DxeDepex) {
-                                let data = x.section_data().to_vec();
-                                Some(data)
+                            if x.section_type() == Some(ffs::section::Type::DxeDepex) {
+                                Some(x.try_content_as_slice())
                             } else {
                                 None
                             }
                         })
```
Javagedes: nit: Formatting - Seems a bit cleaner to do something like this:

```rust
.find_map(|x| {
    match x.section_type() {
        Some(ffs::section::Type::DxeDepex) => Some(x.try_content_as_slice()),
        _ => None,
    }
})
```

patina_dxe_core/src/dispatcher.rs
```diff
@@ -415,35 +427,33 @@ fn add_fv_handles(new_handles: Vec<efi::Handle>) -> Result<(), EfiError> {
...
-                if file.file_type_raw() == FfsFileRawType::FIRMWARE_VOLUME_IMAGE {
+                if file.file_type_raw() == ffs::file::raw::r#type::FIRMWARE_VOLUME_IMAGE {
                     let file = file.clone();
                     let file_name = file.name();
 
                     let sections = {
-                        let res = if let Some(extractor) = &dispatcher.section_extractor {
-                            file.section_iter_with_extractor(extractor.as_ref())
-                                .collect::<Result<Vec<_>, efi::Status>>()
+                        if let Some(extractor) = &dispatcher.section_extractor {
```
Javagedes: nit: Formatting - Seems a bit cleaner to do something like this:

```rust
let sections = match &dispatcher.section_extractor {
    Some(extractor) => file.sections_with_extractor(extractor.as_ref())?,
    _ => file.sections()?
};
```

patina_dxe_core/src/dispatcher.rs
```diff
@@ -316,44 +328,44 @@ fn add_fv_handles(new_handles: Vec<efi::Handle>) -> Result<(), EfiError> {
...
+                        }
                     };
 
                     let depex = sections
                         .iter()
                         .find_map(|x| {
-                            if x.section_type() == Some(FfsSectionType::DxeDepex) {
-                                let data = x.section_data().to_vec();
-                                Some(data)
+                            if x.section_type() == Some(ffs::section::Type::DxeDepex) {
```
Javagedes: nit: Formatting - seems a bit cleaner to do something like:

```rust
.find_map(|x| {
    match x {
        Some(ffs::section::Type::DxeDepex) => Some(x.try_content_as_slice()),
        _ => None
    }
})
```

patina_dxe_core/src/dispatcher.rs
```diff
@@ -316,44 +328,44 @@ fn add_fv_handles(new_handles: Vec<efi::Handle>) -> Result<(), EfiError> {
...
+            for file in fv.files() {
+                let file = file?;
+                if file.file_type_raw() == ffs::file::raw::r#type::DRIVER {
                     let file = file.clone();
                     let file_name = file.name();
                     let sections = {
-                        let res = if let Some(extractor) = &dispatcher.section_extractor {
-                            file.section_iter_with_extractor(extractor.as_ref())
-                                .collect::<Result<Vec<_>, efi::Status>>()
+                        if let Some(extractor) = &dispatcher.section_extractor {
```
Javagedes: nit: Formatting - seems like something like below may be more clean:

```rust
let sections = match &dispatcher.section_extractor {
    Some(extractor) => file.sections_with_extractor(extractor.as_ref())?,
    _ => file.sections()?
}
```

patina_dxe_core/src/fv.rs
Javagedes, line 1: I think a lot of the error handling in this file could be significantly cleaned up if you made a non "efiapi" function that returns a `Result<>`, then call that from the "efiapi" function using the `From` implementation of EfiError. There are many places where you are calling `.into()` to convert an error type to a u32, however if you have it return a Result<(), EfiError>, they can all be reduced from 4-5 lines to a single line.

One example is:

```rust
    let fv = match unsafe { VolumeRef::new_from_address(fvb_data.physical_address) } {
        Ok(fv) => fv,
        Err(err) => return err.into(),
    };
```

can just be:

```rust
    let fv = match unsafe { VolumeRef::new_from_address(fvb_data.physical_address) }?;
```

sdk/patina_ffs/src/file.rs
```diff
@@ -0,0 +1,644 @@
+//! Firmware File System (FFS) file parsing and composition.
+//!
+//! This module provides:
+//! - `FileRef`: a zero-copy, read-only view over a serialized FFS file backed by a byte slice.
+//! - `File`: an owned, mutable builder for creating and serializing FFS files from sections.
+//!
+//! Both types help inspect file metadata (name GUID, type, attributes), traverse sections, and
+//! serialize with correct headers, checksums, and alignment.
```
Javagedes, line 8: nit: need license

sdk/patina_ffs/src/file.rs
```diff
@@ -0,0 +1,644 @@
...
+    /// let section = Section::new_from_header_with_data(
+    ///     SectionHeader::Standard(ffs::section::raw_type::RAW, data.len() as u32),
+    ///     data,
+    /// ).unwrap();
+    /// file.sections_mut().push(section);
+    /// let bytes = file.serialize().unwrap();
+    /// let file_ref = FileRef::new(&bytes).unwrap();
+    /// assert_eq!(file_ref.file_type_raw(), 0x07);
+    /// ```
+    pub fn new(buffer: &'a [u8]) -> Result<Self, FirmwareFileSystemError> {
```
Javagedes, line 75: It may be more canonical to implement  `TryFrom<&[u8]>` instead of this, along with your other try from's. But I'm not saying it needs to be done. Up to you - you could just close this.

sdk/patina_ffs/src/file.rs
```diff
@@ -0,0 +1,644 @@
...
+                    content.extend(iter::repeat_n(0u8, pad_length));
+                }
+            }
+        }
+
+        let mut header = {
+            if ((self.attributes & attributes::raw::LARGE_FILE) != 0)
+                || content.len() > 0xffffff - mem::size_of::<ffs::file::Header>()
+            {
+                let mut file_header = ffs::file::Header2 {
```
Javagedes: In my opinion, it would be easier to comprehend this code to separate these this into two separate functions associated with `Header` and `Header2` struct. Something like

```rust
Ok(if true /* the logic to do Header2 */ {
    Header2::as_bytes(..., content)?
} else {
    Header::as_bytes(..., content)?
})
```

sdk/patina_ffs/src/lib.rs
```diff
@@ -0,0 +1,70 @@
...
+extern crate alloc;
+
+pub mod file;
+pub mod section;
+pub mod volume;
+
+use patina_sdk::error::EfiError;
+use r_efi::efi;
+/// Error definitions for Firmware File System
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
```
Javagedes: typically these get put in an error.rs file. I don't care too much though.

sdk/patina_ffs/src/section.rs
```diff
@@ -0,0 +1,631 @@
+//! Section parsing and composition utilities for UEFI Firmware File System (FFS) sections.
+//!
+//! This module models a single FFS section (leaf or encapsulation) and provides utilities to
+//! parse from raw bytes, compose/serialize, and traverse immediate sub-sections.
+//!
```
Javagedes, line 5: nit: need license

sdk/patina_ffs/src/section.rs
```diff
@@ -0,0 +1,631 @@
...
+    pub fn section_type_raw(&self) -> u8 {
+        match self {
+            SectionHeader::Pad(_) => ffs::section::raw_type::FFS_PAD,
+            SectionHeader::Standard(raw_type, _) => *raw_type,
+            SectionHeader::Compression(_, _) => ffs::section::raw_type::encapsulated::COMPRESSION,
+            SectionHeader::GuidDefined(_, _, _) => ffs::section::raw_type::encapsulated::GUID_DEFINED,
+            SectionHeader::Version(_, _) => ffs::section::raw_type::VERSION,
+            SectionHeader::FreeFormSubtypeGuid(_, _) => ffs::section::raw_type::FREEFORM_SUBTYPE_GUID,
+        }
+    }
```
Javagedes, line 115: nit: new line between these two functions.

sdk/patina_ffs/src/section.rs
```diff
@@ -0,0 +1,631 @@
...
+                ffs::section::raw_type::MM_DEPEX => Some(ffs::section::Type::MmDepex),
+                _ => None,
+            },
+            SectionHeader::Compression(_, _) => Some(ffs::section::Type::Compression),
+            SectionHeader::GuidDefined(_, _, _) => Some(ffs::section::Type::GuidDefined),
+            SectionHeader::Version(_, _) => Some(ffs::section::Type::Version),
+            SectionHeader::FreeFormSubtypeGuid(_, _) => Some(ffs::section::Type::FreeformSubtypeGuid),
+        }
+    }
+    /// Serialize the header into bytes suitable for prefixing the section content.
```
Javagedes, line 142: nit: newline between these two functions

sdk/patina_ffs/src/section.rs
```diff
@@ -0,0 +1,631 @@
...
+                };
+                (freeform_slice.to_vec(), *content_size)
+            }
+        };
+
+        let mut section_header = ffs::section::Header { section_type: self.section_type_raw(), size: [0xffu8; 3] };
+
+        let section_size = mem::size_of_val(&section_header) + header_data.len() + (content_size as usize);
+
+        if section_size < 0x1000000 {
```
Javagedes: should replace this magic number with a const.

sdk/patina_ffs/src/section.rs
```diff
@@ -0,0 +1,631 @@
...
+
+        section_vec.extend(header_data);
+
+        section_vec
+    }
+}
+
+#[derive(Clone)]
+struct LeafSectionData {
+    data: Vec<u8>,
```
Javagedes, line 204: Does this data get large? Seems like `Clone` may not be the right move if the data gets large (Same question with the other structs)

sdk/patina_ffs/src/section.rs
```diff
@@ -0,0 +1,631 @@
...
+            Self::new_from_buffer(&buffer)
+        }
+    }
+
+    /// Parse a serialized section from `buffer`.
+    ///
+    /// Validates the common and variant-specific headers, sets the content size accordingly, and
+    /// stores raw content bytes. Encapsulation sections start with `extracted = false` and no
+    /// populated sub-sections.
+    pub fn new_from_buffer(buffer: &[u8]) -> Result<Self, FirmwareFileSystemError> {
```
Javagedes, line 272: Should canonically probably just be a `TryFrom` implementation

sdk/patina_ffs/src/section.rs
```diff
@@ -0,0 +1,631 @@
...
+
+    /// Extract sub-sections of an encapsulation section via `extractor`.
+    ///
+    /// If the extractor returns `Unsupported`, the method is a no-op. Otherwise, the returned
+    /// bytes are parsed into immediate sub-sections and marked as extracted.
+    pub fn extract(&mut self, extractor: &dyn SectionExtractor) -> Result<(), FirmwareFileSystemError> {
+        match &self.data {
+            SectionData::Encapsulation(x) if !x.extracted => (),
+            _ => return Ok(()), //nothing to do.
+        };
```
Javagedes:
```suggestion
        if !matches!(&self.data, SectionData::Encapsulation(x) if !x.extracted) {
            return Ok(());
        }
```

sdk/patina_ffs/src/volume.rs
```diff
@@ -0,0 +1,1447 @@
...
+        // content_offset to account for this.
+        let content_offset =
+            align_up(content_offset as u64, 8).map_err(|_| FirmwareFileSystemError::InvalidHeader)? as usize;
+
+        Ok(Self { data: buffer, fv_header, ext_header, block_map, content_offset })
+    }
+
+    /// Instantiate a new FirmwareVolume from a base address.
+    ///
+    /// ## Safety
```
Javagedes, line 209: nit: newline between header and contents

sdk/patina_ffs/src/volume.rs
```diff
@@ -0,0 +1,1447 @@
...
+    /// Caller must ensure that base_address is the address of the start of a firmware volume.
+    /// Caller must ensure that the lifetime of the buffer at base_address is longer than the
+    /// returned VolumeRef.
+    ///
+    /// Examples
```
Javagedes: nit: missing the `##`

sdk/patina_ffs/src/volume.rs
```diff
@@ -0,0 +1,1447 @@
...
+            let ext_header_data_start = self.fv_header.ext_header_offset as usize + header_size;
+            let ext_header_end = ext_header_data_start + ext_header.ext_header_size as usize - header_size;
+            let header_data = self.data[ext_header_data_start..ext_header_end].to_vec();
+            (ext_header, header_data)
+        })
+    }
+
+    /// The Firmware Volume name GUID from the extended header, if available.
+    pub fn fv_name(&self) -> Option<efi::Guid> {
+        self.ext_header().map(|x| x.0.fv_name)
```
Javagedes:
```suggestion
        self.ext_header().map(|(hdr, _)| hdr.fv_name)
```

sdk/patina_ffs/src/volume.rs
```diff
@@ -0,0 +1,1447 @@
...
+            attributes: 0,
+            ext_header: None,
+            block_map,
+            files: Vec::new(),
+            capacity: Capacity::Unbounded,
+        }
+    }
+
+    /// Read-only access to the list of FFS files contained in this FV.
+    pub fn files(&self) -> &Vec<File> {
```
Javagedes: I'm not a fan of the fact that `files()` here returns a vector while `files()` in the struct above returns an iterator. We should have it standardized.

sdk/patina_ffs/src/volume.rs
```diff
@@ -0,0 +1,1447 @@
+//! Firmware Volume (FV) parsing, inspection, and composition.
+//!
+//! This module provides:
+//! - `VolumeRef`: a zero-copy, read-only view over a serialized FV backed by a byte slice.
+//! - `Volume`: an owned builder for assembling FVs from a block map and FFS files and serializing.
+//!
+//! It validates FV headers and block maps, iterates contained files, and serializes with proper
+//! alignment, optional extended headers, and checksum calculation per the PI specification.
```
Javagedes, line 8: need license info for file.

sdk/patina_ffs/src/volume.rs
joschock: I can return an iterator in both for this.

sdk/patina_ffs_extractors/src/uefi_decompress.rs
```diff
@@ -0,0 +1,59 @@
...
+            }
+            _ => return Err(FirmwareFileSystemError::Unsupported),
+        };
+
+        //sanity check the src data
+        if src.len() < 8 {
+            Err(FirmwareFileSystemError::DataCorrupt)?;
+        }
+
+        let compressed_size = u32::from_le_bytes(src[0..4].try_into().unwrap()) as usize;
```
os-d: Here and line 51, return an error instead of panicking?

patina_dxe_core/src/dispatcher.rs
```diff
@@ -632,29 +664,38 @@ mod tests {
 
     #[test]
     fn test_init_dispatcher() {
+        set_logger();
```
os-d, line 655: Do we want to keep all of these? I assume this was for debugging, but is it going to spam prints while executing tests?

sdk/patina_ffs/src/file.rs
```diff
@@ -0,0 +1,644 @@
...
+        // safety: buffer is large enough to contain file header.
+        let header = unsafe { ptr::read_unaligned(buffer.as_ptr() as *const file::Header) };
+
+        // determine actual size and content_offset
+        let (size, content_offset) = {
+            if (header.attributes & attributes::raw::LARGE_FILE) == 0 {
+                //standard header with 24-bit size.
+                let mut size = vec![00u8; 4];
+                size[0..3].copy_from_slice(&header.size);
+                let size = u32::from_le_bytes(size.try_into().unwrap()) as usize;
```
os-d: Return error instead of unwrap?

sdk/patina_ffs/src/section.rs
```diff
@@ -0,0 +1,631 @@
...
+                (freeform_slice.to_vec(), *content_size)
+            }
+        };
+
+        let mut section_header = ffs::section::Header { section_type: self.section_type_raw(), size: [0xffu8; 3] };
+
+        let section_size = mem::size_of_val(&section_header) + header_data.len() + (content_size as usize);
+
+        if section_size < 0x1000000 {
+            section_header.size = (section_size as u32).to_le_bytes()[0..3].try_into().unwrap();
```
os-d, line 183: Does `section_header.size` need to be set when the size is >= 0x1000000?

joschock:
> I haven't reviewed the mu_pi code, but my understanding is that you moved that code to patina, then refactored it. Does it make sense to have this be a rebase & FF PR with two commits, one that just moves the code into patina, then the other that does your refactor? It would make the review of the refactor easier and keep the original implementation in the patina history for comparison.
> 
> If this doesn't make sense (because it is a completely different implementation, etc.) then I'm okay with saying that the original history lives in mu_pi.


Well, I re-coded it from scratch using the mu_pi version as a semantic reference. So to reconstruct it as a "move, then refactor" clean set of commits would be a bit of work, but doable. Not sure it is worth it?

patina_dxe_core/src/fv.rs
joschock: this is good feedback, but I would prefer to implement it as a separate PR since this guy is already big enough as is. Created and self-assigned https://github.com/OpenDevicePartnership/patina/issues/715 to do it in a follow up PR.

patina_dxe_core/src/dispatcher.rs
joschock: fixed in next push

patina_dxe_core/src/dispatcher.rs
joschock: fixed in next push.

patina_dxe_core/src/dispatcher.rs
joschock: fixed in next push

patina_dxe_core/src/dispatcher.rs
joschock: fixed in next push

sdk/patina_ffs/src/file.rs
joschock: fixed in next push

sdk/patina_ffs/src/file.rs
joschock: had cake and ate it: implemented TryFrom() and retained new.

sdk/patina_ffs/src/file.rs
joschock: associating it with `ffs::file::Header/Header2` struct would move the implementation into back into the mu_rust_pi crate (which I would like to avoid per the discussion in the RFC about keeping it strictly spec structs). But I agree that splitting it into functions is cleaner, so I'll do that as private methods on `File`.

joschock: fixed in next push.

sdk/patina_ffs/src/section.rs
joschock: fixed in next push.

sdk/patina_ffs/src/lib.rs
joschock: fixed in next push.

sdk/patina_ffs/src/section.rs
joschock: fixed in next push

makubacki:
> > I haven't reviewed the mu_pi code, but my understanding is that you moved that code to patina, then refactored it. Does it make sense to have this be a rebase & FF PR with two commits, one that just moves the code into patina, then the other that does your refactor? It would make the review of the refactor easier and keep the original implementation in the patina history for comparison.
> > If this doesn't make sense (because it is a completely different implementation, etc.) then I'm okay with saying that the original history lives in mu_pi.
> 
> Well, I re-coded it from scratch using the mu_pi version as a semantic reference. So to reconstruct it as a "move, then refactor" clean set of commits would be a bit of work, but doable. Not sure it is worth it?

I don't think so. We can go reference the old mu_pi code if needed.

patina_dxe_core/src/dispatcher.rs
```diff
@@ -542,12 +552,34 @@ mod tests {
...
     use std::{fs::File, io::Read, vec};
 
+    use log::{Level, LevelFilter, Metadata, Record};
     use patina_internal_device_path::DevicePathWalker;
     use uuid::uuid;
 
     use super::*;
     use crate::test_collateral;
 
+    // Sample logger for log crate to dump stuff in tests
```
makubacki: nit: Sample could work, but was "simple" meant here?

```suggestion
    // Simple logger for log crate to dump stuff in tests
```

sdk/patina_ffs/src/section.rs
joschock: Large is relative, but most sections are going to be some where between bytes and hundreds of KB. None of our FVs that I know of have sections that exceed the max standard size (16MB). Largest sections tend to be PE32 code sections and most of those are in the <100K range. 

Do you mean something like an `Rc` or a `Cow`? I considered something like that, even prototyped it a bit; but it added complexity in handling the sections (particularly around lifetime handling for extracted sections) so I traded it off for simplicity. We could do some profiling to see if further work is justified here from a perf standpoint, but I think that's probably beyond the scope of this initial implementation PR.

sdk/patina_ffs/src/section.rs
joschock: fixed in next push

sdk/patina_ffs/src/section.rs
joschock: fixed in next push

sdk/patina_ffs/src/section.rs
joschock: Cake+Ate. Added TryFrom wrapper for new_from_buffer.

sdk/patina_ffs/src/section.rs
Javagedes: Fair. I was thinking of Cow, but if it adds extra complexity then it's probably not worth it. This is not a very hot part of code:)

sdk/patina_ffs/src/section.rs
joschock: fixed in next push.

sdk/patina_ffs/src/volume.rs
joschock: fixed in next push

sdk/patina_ffs/src/volume.rs
joschock: fixed in next push.

sdk/patina_ffs/src/volume.rs
joschock: fixed in next push.

sdk/patina_ffs/src/volume.rs
joschock: fixed in next push. Made them both iterators, but the owned version doesn't need to be able to return an error, so the Iterator item is different.

sdk/patina_ffs/src/volume.rs
joschock: fixed in next push

sdk/patina_ffs_extractors/src/uefi_decompress.rs
joschock: fixed in next push.

patina_dxe_core/src/dispatcher.rs
joschock: yes, they don't spam prints in normal text execution as that suppresses stdout.  This allows examination of log on failure, which is convenient if you're looking at an error e.g. from server build. This functions same as println in tests.

sdk/patina_ffs/src/file.rs
joschock: converted size to a slice so that the conversion is infallible. fixed in next push.

sdk/patina_ffs/src/section.rs
joschock: If the size is greater than 16MB, then the extended size field has the actual size, and the regular size field is set to 0xFFFFFF. The latter is already done as part of header initialization:
```
let mut section_header = ffs::section::Header { section_type: self.section_type_raw(), size: [0xffu8; 3] };
```

patina_dxe_core/src/dispatcher.rs
joschock: simple was meant - fixed in latest release.

os-d:
> Well, I re-coded it from scratch using the mu_pi version as a semantic reference. So to reconstruct it as a "move, then refactor" clean set of commits would be a bit of work, but doable. Not sure it is worth it?

Makes sense, I agree, it’s fine as is. I couldn’t tell how much was new vs old

PR #714, opened 2025-08-28
[REBASE & FF] Fix asm! Macro Usage in Patina

### LINKED ISSUE #483: [Bug]: Review Inline ASM Usage

### Current Behavior

A review of inline asm options, explicit register use, etc. is needed to ensure hard to catch bugs are not produced.

### Expected Behavior

n/a

### Steps To Reproduce

n/a

### Build Environment

```markdown
- OS(s):
- Tool Chain(s):
- Targets Impacted:
```

### Version Information

```text
top of tree
```

### Urgency

Medium

I will fix it

No maintainer feedback needed

### Anything else?

_No response_

os-d:
## Description

This PR contains two commits that fix two classes of errors w.r.t. inline asm usage in Patina. Closes #483 in conjunction with PRs in patina-paging and patina-mtrr.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

<_Describe the test(s) that were run to verify the changes._>

## Integration Instructions

<_Describe how these changes should be integrated. Use N/A if nothing is required._>

### DIFF
--- a/core/patina_debugger/src/arch/aarch64.rs
+++ b/core/patina_debugger/src/arch/aarch64.rs
@@ -2,7 +2,7 @@ use core::{
     arch::asm,
     num::NonZeroUsize,
     ops::Shr,
-    sync::atomic::{AtomicBool, Ordering, compiler_fence},
+    sync::atomic::{AtomicBool, Ordering},
 };
 
 use gdbstub::arch::{RegId, Registers};
@@ -40,6 +40,11 @@ const DAIF_DEBUG_MASK: u64 = 0x200;
 
 static POKE_TEST_MARKER: AtomicBool = AtomicBool::new(false);
 
+/// This enum is used to specify the type of barrier to use when writing to a system register and in which order.
+enum BarrierType {
+    Instruction,
+}
+
 macro_rules! read_sysreg {
   ($reg:expr) => {{
     let value: u64;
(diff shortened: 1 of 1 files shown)


os-d:
Note, keeping this PR open until @cfernald is back, want to discuss implications with him and review the debugger's usage of `compiler_fence` instead of asm synchronization barriers. Per the [docs](https://doc.rust-lang.org/std/sync/atomic/fn.compiler_fence.html), `compiler_fence` does not generate any machine code, just is a hint to the compiler to not reorder this code. Which may be fine, but if we need actual HW synchronization barriers in those places, that is not going to work. It may be safest to convert to writing the barriers, but want to discuss.

core/patina_debugger/src/arch/x64.rs
```diff
@@ -133,8 +131,9 @@ impl DebuggerArch for X64Arch {
     fn reboot() {
         // Reset the system through the keyboard controller IO port.
         unsafe {
-            asm!("cli");
-            asm!("out dx, al", in("dx") 0x64, in("al") 0xFE_u8);
+            asm!("cli", "out dx, al", in("dx") 0x64, in("al") 0xFE_u8);
```
joschock, line 134: FWIW, port 0xCF9 is a bit more ~~traditional~~ common/modern reset generator than the old port 64 keyboard reset. you might consider writing 0x6 to port 0xCF9 in the event that the old 60/64 ports are not emulated.

os-d: I'll defer that to @cfernald who wrote this portion

cfernald: This is probably a good idea. I did notice that the legacy keyboard reset doesnt work on physical x64 devices. I think for the sake of this PR you can leave it alone, but I filed #720.

os-d:
> Note, keeping this PR open until @cfernald is back, want to discuss implications with him and review the debugger's usage of `compiler_fence` instead of asm synchronization barriers. Per the [docs](https://doc.rust-lang.org/std/sync/atomic/fn.compiler_fence.html), `compiler_fence` does not generate any machine code, just is a hint to the compiler to not reorder this code. Which may be fine, but if we need actual HW synchronization barriers in those places, that is not going to work. It may be safest to convert to writing the barriers, but want to discuss.

Per offline discussion, this does need to be HW synchronized, so moving these to asm instructions in a separate commit

PR #724, opened 2025-09-03
protocols: Allow protocol usage removal include attributes

makubacki:
## Description

`remove_protocol_usage` removes all usages that match the same agent handle and controller handle. This can cause a failure in the for loop iteration because the current usage targeted for removal may have already been deleted in a previous iteration, resulting in a `NotFound` error.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- Q35/SBSA boot
- Unit tests
- Verify system issue is fixed - removing USB in front page does not hang

## Integration Instructions

- N/A

### DIFF
--- a/patina_dxe_core/src/protocols.rs
+++ b/patina_dxe_core/src/protocols.rs
@@ -133,8 +133,13 @@ pub fn core_uninstall_protocol_interface(
                     | efi::OPEN_PROTOCOL_TEST_PROTOCOL)
                 != 0
             {
-                let result =
-                    PROTOCOL_DB.remove_protocol_usage(handle, protocol, usage.agent_handle, usage.controller_handle);
+                let result = PROTOCOL_DB.remove_protocol_usage(
+                    handle,
+                    protocol,
+                    usage.agent_handle,
+                    usage.controller_handle,
+                    Some(usage.attributes),
+                );
                 if result.is_err() {
                     unclosed_usages = true;
                 }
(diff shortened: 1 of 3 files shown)


patina_dxe_core/src/protocols.rs
```diff
@@ -135,8 +135,13 @@ pub fn core_uninstall_protocol_interface(
             {
                 let result =
                     PROTOCOL_DB.remove_protocol_usage(handle, protocol, usage.agent_handle, usage.controller_handle);
-                if result.is_err() {
-                    unclosed_usages = true;
+                if let Err(err) = result {
```
Javagedes: This seems more so of an indicator of a different bug - either that `add_protocol_usage` is adding a usage when it shouldn't (because it already exists with the same settings), or that `remove_protocol_usage` need an optional `attributes` to filter by.

As a secondary if the above is not true, it seems like we should instead do some type of reduction / filter of the usages so we only call `remove_protocol_usage` the correct number of times.

The third and final would just be that we should do a match

``` rust
match result {
   Ok(_) | Err(EfiError::NotFound) => {}
   _ => unclosed_usages = true;
}
```

makubacki: It's in draft while trying a few different options. I did try using attributes. That worked and is likely what I'll go with.

makubacki: Updated to add `attributes` to `remove_protocol_usage`.

PR #732, opened 2025-09-09
[REBASE & FF] dxe_core: Introduce 32 bit memory prioritization option & resolve allocator issues

### LINKED ISSUE #191: Page allocator should have option to be consistent with EDKII

### Current Behavior

The current page allocator default is to do a pure top-down allocation. This was done to avoid allocating the <1MB memory on x64 which is needed for MP support. However, this allocation scheme is different from the C code allocator which does a scope bottom-up allocation. This has exposed a number of bugs across UEFI and Windows, but before releasing it would be good to go for consistency to avoid incompatibility with 3rd party software and drivers.

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Build Environment

```markdown
- OS(s):
- Tool Chain(s):
- Targets Impacted:
```

### Version Information

```text

```

### Urgency

Low

Someone else needs to fix it

None

### Anything else?

_No response_

cfernald:
## Description

This pull request adds functionality for prioritize allocating 32-bit address for platforms that need to support buggy external code. This change includes a few incidental changes details below for consistency and fixes related to this functionality.

# Introduce a prioritize_32_memory option for the core & GCD

Some platforms may encounter issues using the Patina default top-down allocation strategy as it will prioritize high memory and some components may not properly handle 64 bit addresses. This could be because of fixed memory sizes or use of 32 bit variables to store addresses.

This new routine will attempt tp priorritize 32 bit memory when used. This specifically applies to top down allocations as bottom up will inherently prioritize low memory and address allocations are not applicable.

# Switch TopDown to use max address for PI spec

The PI specified allocation type for TopDown specifies a max address and not a min address for both memory and IO in section 7 as "EfiGcdAllocateMaxAddressSearchTopDown". The correct definition was used, but converted into an inverse type. This commit resolves this inconsistency and converts all top down searches to use a max instead of a min.

# Change ALLOCATE_MAX_ADDRESS to still use TopDown allocations

For consistency this commit changes MAX_ADDRESS allocations to also use TopDown but with the max address set. This ensures that low memory is reserved for when it is actually needed and prevents fragmenting memory.

Additionally, this change removed a condition in the allocate_pages_should_allocate_pages test as it incorrectly assumes MAX_ADDRESS follows a bottom up scheme.

CLOSES #191 

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- [x] Q35 OS Boot
- [ ] SBSA OS Boot
- [ ] Intel platform OS Boot.

## Integration Instructions

N/A

### DIFF
--- a/patina_dxe_core/src/allocator.rs
+++ b/patina_dxe_core/src/allocator.rs
@@ -482,7 +482,7 @@ pub fn core_allocate_pages(
                 efi::ALLOCATE_ANY_PAGES => allocator.allocate_pages(DEFAULT_ALLOCATION_STRATEGY, pages, alignment),
                 efi::ALLOCATE_MAX_ADDRESS => {
                     let address = unsafe { memory.as_ref().expect("checked non-null is null") };
-                    allocator.allocate_pages(AllocationStrategy::BottomUp(Some(*address as usize)), pages, alignment)
+                    allocator.allocate_pages(AllocationStrategy::TopDown(Some(*address as usize)), pages, alignment)
                 }
                 efi::ALLOCATE_ADDRESS => {
                     let address = unsafe { memory.as_ref().expect("checked non-null is null") };
@@ -1453,18 +1453,6 @@ mod tests {
                 efi::Status::SUCCESS
             );
             free_pages(buffer_ptr as u64, 0x10);
-
-            //test unsuccessful allocate_max where max is less than the address that was just freed.
-            buffer_ptr = buffer_ptr.wrapping_sub(0x12 * 0x1000);
-            assert_eq!(
-                allocate_pages(
-                    efi::ALLOCATE_MAX_ADDRESS,
-                    efi::BOOT_SERVICES_DATA,
-                    0x10,
-                    core::ptr::addr_of_mut!(buffer_ptr) as *mut efi::PhysicalAddress
-                ),
-                efi::Status::NOT_FOUND
-            );
 
             //test invalid allocation type
             assert_eq!(


patina_dxe_core/src/gcd/spin_locked_gcd.rs
```diff
@@ -59,7 +59,7 @@ enum InternalError {
 pub enum AllocateType {
     // Allocate from the lowest address to the highest address or until the specify address is reached (max address).
     BottomUp(Option<usize>),
-    // Allocate from the highest address to the lowest address or until the specify address is reached (min address).
+    // Allocate from the highest address to the lowest address starting at the specified address (max address).
```
os-d: nit: Starting at the optionally specified address (max address) or top of memory.

patina_dxe_core/src/gcd/spin_locked_gcd.rs
```diff
@@ -721,6 +724,17 @@ impl GCD {
...
         ensure!(len > 0, EfiError::InvalidParameter);
 
+        // For top down requests specifically, if prioritize 32 bit memory is set, then first
+        // try with an artificial max.
+        if self.prioritize_32_bit_memory && max_address > u32::MAX as usize {
+            match self.allocate_top_down(memory_type, align_shift, len, image_handle, device_handle, u32::MAX as usize)
+            {
+                Ok(addr) => return Ok(addr),
+                Err(EfiError::NotFound) => {}
+                Err(e) => return Err(e),
```
os-d: If we hit an error here, should we bail or should we try again always? I think there should be debug_asserts in the path, for sure, to catch the errors, but for a release boot, probably should try again above 32 bits?

patina_dxe_core/src/allocator.rs
makubacki, line 1: Can you update the docs to show how a platform should use this?

patina_dxe_core/src/gcd/spin_locked_gcd.rs
```diff
@@ -717,52 +720,59 @@ impl GCD {
...
         log::trace!(target: "gcd_measure", "search");
-        let mut current = memory_blocks.last_idx();
+        let mut current = memory_blocks.get_closest_idx(&(max_address as u64));
         while let Some(idx) = current {
             let mb = memory_blocks.get_with_idx(idx).expect("idx is valid from prev_idx");
-            if mb.len() < len {
+
+            // Account for if the block is truncated by the max_address. Max address
+            // is inclusive, but end() is exclusive so subtract 1 from end.
+            let usable_len = if mb.end() - 1 > max_address { max_address - mb.start() + 1 } else { mb.len() };
```
makubacki: Is `mb.start()` guaranteed to be less than `max_address` here?

patina_dxe_core/src/gcd/spin_locked_gcd.rs
makubacki: I'd just make it clear.

```suggestion
    // Allocate from the highest address to the lowest address.
    // Some(address) => Start at the specified address (inclusive max address).
    // None => Start at top of memory
```

patina_dxe_core/src/gcd/spin_locked_gcd.rs
cfernald: debatable, really the failures that could occur other than NotFound are all errors not necessarily to the limit. A debug assert probably isn't right because it could just be a regular failure that will occur in the nested call. Falling back is fine, but perhaps redundant.

patina_dxe_core/src/gcd/spin_locked_gcd.rs
cfernald: It should, the semanics of get_closest_idx is the closest IDX rounded down, so we should not get a start that is greater than the max_address. I will make this a checked sub to be same though.

cfernald: Though, thinking about it, panicing since that is a broken internal assumption makes some sense.

PR #734, opened 2025-09-11
Initial smbios header code conversion

ansleythompson:
## Description

SmbiosDxe.h and Smbios.h conversion from c to rust. Combined both of these into smbios.rs for the time being.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

NA

## Integration Instructions

NA

### DIFF
--- a/components/patina_smbios/src/smbios.rs
+++ b/components/patina_smbios/src/smbios.rs
@@ -2619,7 +2619,6 @@ pub struct DeviceStruct {
 ///
 #[repr(C, packed)]
 /// OEM Strings (Type 11)
-#[repr(C, packed)]
 pub struct SmbiosTableType11 {
     /// Structure header
     pub hdr: SmbiosStructure,
@@ -2634,7 +2633,6 @@ pub struct SmbiosTableType11 {
 ///
 #[repr(C, packed)]
 /// System Configuration Options (Type 12)
-#[repr(C, packed)]
 pub struct SmbiosTableType12 {
     /// Structure header
     pub hdr: SmbiosStructure,
@@ -2649,7 +2647,6 @@ pub struct SmbiosTableType12 {
 ///
 #[repr(C, packed)]
 /// BIOS Language Information (Type 13)
-#[repr(C, packed)]
 pub struct SmbiosTableType13 {
     /// Structure header
     pub hdr: SmbiosStructure,
@@ -2668,7 +2665,6 @@ pub struct SmbiosTableType13 {
 ///
 #[repr(C, packed)]
 /// Group Item Entry
-#[repr(C, packed)]
 pub struct GroupStruct {
     /// Item type
     pub item_type: u8,
@@ -2685,7 +2681,6 @@ pub struct GroupStruct {
 ///
 #[repr(C, packed)]
 /// Group Associations (Type 14)
-#[repr(C, packed)]
 pub struct SmbiosTableType14 {
     /// Structure header
     pub hdr: SmbiosStructure,
@@ -4796,7 +4791,6 @@ impl SmbiosHandleEntry {
 
 #[repr(C, packed)]
 /// EFI SMBIOS Table End Structure
-#[repr(C, packed)]
 pub struct EfiSmbiosTableEndStructure {
     /// Table end header
     pub header: EfiSmbiosTableHeader,


rogurr:
Quick ACK for anyone reading this later, we do know there are "TODO" tags in the file, but due to the size we wanted to check it in before working on any further steps.

components/patina_smbios/Cargo.toml
```diff
@@ -10,6 +10,9 @@ publish.workspace = true
 [dependencies]
 log = {workspace = true}
 patina_sdk = {workspace = true}
+bitfield = "0.19.2"
```
Javagedes: Dependencies should be defined in the workspace toml file and then consumed in the individual crates using `workspace = true`. See examples in the same file.

components/patina_smbios/Cargo.toml
```diff
@@ -10,6 +10,9 @@ publish.workspace = true
...
+r-efi = "5.3.0"
+mu_pi = {git = "https://github.com/microsoft/mu_rust_pi"}
```
Javagedes: You cannot use git dependencies in this repository as they are published to a feed (Which does not allow git dependencies). mu_pi is already defined in the workspace.toml file. Please use `workspace = true`.

components/patina_smbios/src/smbios.rs
Javagedes, line 1: This new module (`smbios`) needs to be added to `lib.rs` (`pub mod smbios.rs`) so that it can be accessed by other parts of the component, or publically :)

components/patina_smbios/src/smbios.rs
```diff
@@ -0,0 +1,3161 @@
...
+    pub security_status: u8,
+    pub oem_defined: [u8; 4],
+    pub height: u8,
+    pub numberof_power_cords: u8,
+    pub contained_element_count: u8,
+    pub contained_element_record_length: u8,
+    //
+    // Can have 0 to (ContainedElementCount * ContainedElementRecordLength) contained elements
+    //
+    pub contained_elements: [ContainedElement; 1],
```
joschock: This reserves space for at least one contained element. Is contained_element_count always non-zero per spec? Otherwise, consider `[ContainedElement; 0]`

components/patina_smbios/src/smbios.rs
```diff
@@ -0,0 +1,3161 @@
...
+    //
+    pub segment_group_num: u16,
+    pub bus_num: u8,
+    pub dev_func_num: u8,
+    //
+    // Add for smbios 3.2
+    //
+    pub data_bus_width: u8,
+    pub peer_grouping_count: u8,
+    pub peer_groups: [MiscSlotPeerGroup; 1],
```
joschock: Consider `[MiscSlotPeerGroup; 0]` unless `peer_grouping_count` is always > 0.

components/patina_smbios/src/smbios.rs
```diff
@@ -0,0 +1,3161 @@
...
+}
+*/
+
+/// SMBIOS record Header
+///
+/// An SMBIOS internal Record is an EFI_SMBIOS_RECORD_HEADER followed by
+/// (RecordSize - HeaderSize) bytes of data. The format of the data is
+/// defined by the SMBIOS spec.
+#[repr(C, packed)]
+pub struct EfiSmbiosRecordHeader {
```
joschock, line 4734: Are these internal for Rust usage, or still used with C FFI? If they are only intended for rust usage, consider making safer by avoiding types with header + untyped/unconstrained data after them and using e.g. a slice (or a Vec<u8>) and accessor routines. If they are intended for use in FFI, then this is probably fine to match the C style approach.

components/patina_smbios/Cargo.toml
kat-perez: added to workspace dependencies instead

components/patina_smbios/Cargo.toml
kat-perez: updated to `workspace = true`

components/patina_smbios/src/smbios.rs
kat-perez: added `pub mod smbios` to lib.rs

components/patina_smbios/src/smbios.rs
kat-perez: I think it does say that per SMBIOS spec that the ContainedElementCount may be zero. So, setting this to 0 instead

components/patina_smbios/src/smbios.rs
kat-perez: also setting this to `[MiscSlotPeerGroup; 0]` since the spec allows Peer Grouping Count to be zero

components/patina_smbios/src/smbios.rs
kat-perez: going to add this as a TODO comment for now so we can reevaluate it later

PR #738, opened 2025-09-12
dxe_core: Move runtime event handling into runtime module

### LINKED ISSUE #337: [Bug]: Crash in set_virtual_address_map

### Current Behavior

There is a periodic crash in set_virtual_address_space map with the following signature:

```
INFO - EBS completed successfully. # Note this is a print about to be added
ERROR - panicked at C:\Users\olsmithd\.rustup\toolchains\1.84.0-x86_64-pc-windows-msvc\lib\rustlib\src\rust\library\alloc\src\collections\btree\node.rs:863:9:
assertion failed: idx <= node.len()
INFO - Dumping stack trace with PC: 0x7ea0eadf, SP: 0xffffffffaac039a0
INFO -       # Child-SP              Return Address         Call Site
INFO -       0 FFFFFFFFAAC039A0      000000007E9B41D1       qemu_q35_dxe_core-657b42e768b0b19e+69ADF
INFO -       1 FFFFFFFFAAC039D0      000000007EA9AF1F       qemu_q35_dxe_core-657b42e768b0b19e+F1D1
INFO -       2 FFFFFFFFAAC03AC0      000000007EA9AFC9       qemu_q35_dxe_core-657b42e768b0b19e+F5F1F
INFO -       3 FFFFFFFFAAC03B00      000000007EA51A5F       qemu_q35_dxe_core-657b42e768b0b19e+F5FC9
INFO -       4 FFFFFFFFAAC03B70      000000007EA2E65C       qemu_q35_dxe_core-657b42e768b0b19e+ACA5F
INFO -       5 FFFFFFFFAAC03BA0      000000007EA2F2B7       qemu_q35_dxe_core-657b42e768b0b19e+8965C
INFO -       6 FFFFFFFFAAC03C70      000000007E9F68A9       qemu_q35_dxe_core-657b42e768b0b19e+8A2B7
INFO -       7 FFFFFFFFAAC03D30      FFFFFFFFA86F0818       qemu_q35_dxe_core-657b42e768b0b19e+518A9
```

Which when run through the stacktrace script gives:

```
 # Source Path                                                           Child-SP         Return Address   Call Site
 0 [C:\Users\olsmithd\.cargo\registry\src\pkgs.dev.azure.com-f1ddb803f8268cb0\stacktrace-10.0.3\src\stacktrace.rs @  115] FFFFFFFFAAC039A0 000000007E9B41D1 qemu_q35_dxe_core-657b42e768b0b19e!union enum2$<core::result::Result<tuple$<>,enum2$<stacktrace::error::Error> > > stacktrace::stacktrace::StackTrace::dump()+F
 1 [d:\mu\mu_dev\qemu_rust_bins\bin\q35_dxe_core.rs              @   24] FFFFFFFFAAC039D0 000000007EA9AF1F qemu_q35_dxe_core-657b42e768b0b19e!void qemu_q35_dxe_core::panic(struct core::panic::panic_info::PanicInfo *)+D1
 2 [C:\Users\olsmithd\.rustup\toolchains\1.84.0-x86_64-pc-windows-msvc\lib\rustlib\src\rust\library\core\src\panicking.rs @   76] FFFFFFFFAAC03AC0 000000007EA9AFC9 qemu_q35_dxe_core-657b42e768b0b19e!void core::panicking::panic_fmt(struct core::fmt::Arguments, struct core::panic::location::Location *)+1F
 3 [C:\Users\olsmithd\.rustup\toolchains\1.84.0-x86_64-pc-windows-msvc\lib\rustlib\src\rust\library\core\src\panicking.rs @  148] FFFFFFFFAAC03B00 000000007EA51A5F qemu_q35_dxe_core-657b42e768b0b19e!void core::panicking::panic(struct ref$<str$>, struct core::panic::location::Location *)+49
 4 [C:\Users\olsmithd\.rustup\toolchains\1.84.0-x86_64-pc-windows-msvc\lib\rustlib\src\rust\library\alloc\src\collections\btree\navigate.rs @  559] FFFFFFFFAAC03B70 000000007EA2E65C qemu_q35_dxe_core-657b42e768b0b19e!struct tuple$<ref$<usize>,ref_mut$<dxe_core::event_db::Event> > alloc::collections::btree::node::Handle<alloc::collections::btree::node::NodeRef<alloc::collections::btree::node::marker::ValMut,usize,dxe_core::event_db::Event,enum2$<alloc::collections::btree::node::marker::Leaf> >,enum2$<alloc::collections::btree::node::marker::Edge> >::next_unchecked<usize,dxe_core::event_db::Event>()+6F
 5 [D:\mu\mu_dev\uefi-dxe-core\dxe_core\src\event_db.rs          @  342] FFFFFFFFAAC03BA0 000000007EA2F2B7 qemu_q35_dxe_core-657b42e768b0b19e!void dxe_core::event_db::EventDb::signal_group(struct r_efi::base::Guid)+10C
 6 [D:\mu\mu_dev\uefi-dxe-core\dxe_core\src\event_db.rs          @  594] FFFFFFFFAAC03C70 000000007E9F68A9 qemu_q35_dxe_core-657b42e768b0b19e!void dxe_core::event_db::SpinLockedEventDb::signal_group(struct r_efi::base::Guid)+87
 7 [D:\mu\mu_dev\uefi-dxe-core\dxe_core\src\runtime.rs           @   87] FFFFFFFFAAC03D30 FFFFFFFFA86F0818 qemu_q35_dxe_core-657b42e768b0b19e!static struct r_efi::base::Status dxe_core::runtime::set_virtual_address_map(unsigned __int64, unsigned __int64, unsigned int, struct r_efi::system::MemoryDescriptor *)+F9
```

### Expected Behavior

We should not crash in set_virtual_address_map

### Steps To Reproduce

Boot top of tree of all repos to Linux. Occasionally we will hit this.

### Build Environment

```markdown
- OS(s): Linux
- Tool Chain(s):
- Targets Impacted:
```

### Version Information

```text
top of tree
```

### Urgency

Medium

Someone else needs to fix it

No maintainer feedback needed

### Anything else?

_No response_

cfernald:
## Description

This commit moves the handling of runtime events into the runtime.rs module and allocates them out of runtime memory. Runtime events, by definition and standard, are intended to be allocated in runtime memory and handled by a runtime driver. Since Patina handles the set_virtual_address_map itself, it must try to isolate that logic as much as possible from leveraging boot services resources.

This commit specifically resolves an issue on Windows where the set virtual address map is not called from the full firmware context. In this seperate context, the runtime services memory is mapped, but newer boot services data allocations may not be mapped. Moving the allocations to runtime memory works around potential access violations when traversing the tree that contains events that may have been allocated from a later heap.

CLOSES #337 

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Boot to Windows on Q35

## Integration Instructions

N/A

### DIFF
--- a/patina_dxe_core/src/event_db.rs
+++ b/patina_dxe_core/src/event_db.rs
@@ -20,7 +20,7 @@ use core::{cmp::Ordering, ffi::c_void, fmt};
 use patina_sdk::error::EfiError;
 use r_efi::efi;
 
-use crate::tpl_lock;
+use crate::{runtime, tpl_lock};
 
 /// Defines the supported UEFI event types
 #[repr(u32)]
@@ -188,7 +188,7 @@ impl Eq for TaggedEventNotification {}
 // whereas efi::Event is used as the public index or handle into the event database.
 // In the code below efi::Event is used to qualify the index/handle type, where as `Event` with
 // scope qualification refers to this private type.
-struct Event {
+pub(crate) struct Event {
     event_id: usize,
     event_type: EventType,
     event_group: Option<efi::Guid>,
@@ -204,6 +204,11 @@ struct Event {
     trigger_time: Option<u64>,
     period: Option<u64>,
 }
+
+// SAFETY: This structure is used within a lock on a single core and is not mutated
+// after creation. It is safe to share references to it.
+unsafe impl Sync for crate::event_db::Event {}
+unsafe impl Send for crate::event_db::Event {}
 
 impl fmt::Debug for Event {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
@@ -266,6 +271,26 @@ impl Event {
             period: None,
         })
     }
+
+    pub fn event_id(&self) -> usize {
+        self.event_id
+    }
+
+    pub fn efi_event(&self) -> efi::Event {
+        self.event_id as efi::Event
+    }
+
+    pub fn event_group(&self) -> Option<efi::Guid> {
+        self.event_group
+    }
+
+    pub fn notify_fn(&self) -> Option<efi::EventNotify> {
+        self.notify_function
+    }
+
+    pub fn notify_context(&self) -> Option<*mut c_void> {
+        self.notify_context
+    }
 }
 
 struct EventDb {
@@ -280,6 +305,9 @@ struct EventDb {
 }
 
 impl EventDb {
+    /// Tag used for runtime event IDs.
+    const RT_EVENT: usize = 1 << (usize::BITS - 1);
+
     const fn new() -> Self {
         EventDb { events: BTreeMap::new(), next_event_id: 1, pending_notifies: BTreeSet::new(), notify_tags: 0 }
     }
@@ -292,16 +320,38 @@ impl EventDb {
         notify_context: Option<*mut c_void>,
         event_group: Option<efi::Guid>,
     ) -> Result<efi::Event, EfiError> {
-        let id = self.next_event_id;
+        if self.next_event_id == Self::RT_EVENT {
+            debug_assert!(false, "Event ID space exhausted.");
+            return Err(EfiError::OutOfResources);
+        }
+
+        let runtime =
+            (event_type & efi::EVT_RUNTIME) != 0 || event_group == Some(efi::EVENT_GROUP_VIRTUAL_ADDRESS_CHANGE);
+
+        // Tag runtime event specially as they are tracked seperately.
+        let id = if runtime { self.next_event_id | Self::RT_EVENT } else { self.next_event_id };
         self.next_event_id += 1;
+
         let event = Event::new(id, event_type, notify_tpl, notify_function, notify_context, event_group)?;
-        self.events.insert(id, event);
+        if runtime {
+            // Runtime events are managed by the runtime module to reduce dependencies
+            // on boot services.
+            runtime::add_runtime_event(event)?;
+        } else {
+            self.events.insert(id, event);
+        }
+
         Ok(id as efi::Event)
     }
 
     fn close_event(&mut self, event: efi::Event) -> Result<(), EfiError> {
         let id = event as usize;
-        self.events.remove(&id).ok_or(EfiError::InvalidParameter)?;
+        if (id & Self::RT_EVENT) != 0 {
+            runtime::remove_runtime_event(id)?;
+        } else {
+            self.events.remove(&id).ok_or(EfiError::InvalidParameter)?;
+        }
+
         Ok(())
     }
 
@@ -310,7 +360,7 @@ impl EventDb {
         if event.event_type.is_notify_signal() || event.event_type.is_notify_wait() {
             pending_notifies.insert(TaggedEventNotification(
                 EventNotification {
-                    event: event.event_id as efi::Event,
+                    event: event.efi_event(),
                     notify_tpl: event.notify_tpl,
                     notify_function: event.notify_function,
                     notify_context: event.notify_context,
--- a/patina_dxe_core/src/runtime.rs
+++ b/patina_dxe_core/src/runtime.rs
@@ -6,19 +6,20 @@
 //!
 //! SPDX-License-Identifier: Apache-2.0
 //!
+use alloc::collections::BTreeMap;
 use core::{
     ffi::c_void,
     mem, ptr,
     sync::atomic::{AtomicBool, Ordering},
 };
 use mu_pi::{list_entry, protocols::runtime};
-use patina_sdk::base::UEFI_PAGE_SIZE;
+use patina_sdk::{base::UEFI_PAGE_SIZE, error::EfiError};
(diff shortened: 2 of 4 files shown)


patina_dxe_core/src/runtime.rs
```diff
@@ -42,7 +43,12 @@ impl RuntimeData {
 unsafe impl Sync for RuntimeData {}
 unsafe impl Send for RuntimeData {}
 
+unsafe impl Sync for crate::event_db::Event {}
```
Javagedes: This should be implemented in the same place `Event` is defined. Also, this is an unsafe implementation, so you need to describe what makes this structure safe to send and update asynchronously.

patina_dxe_core/src/runtime.rs
```diff
@@ -42,7 +43,12 @@ impl RuntimeData {
...
+unsafe impl Send for crate::event_db::Event {}
+
 static RUNTIME_DATA: Mutex<RuntimeData> = Mutex::new(RuntimeData::new());
+static RUNTIME_EVENTS: Mutex<BTreeMap<usize, crate::event_db::Event, &'static crate::allocator::UefiAllocator>> =
```
Javagedes, line 47: Is there any reason _not_ to add this to the existing runtime data structure? It seems like it would be better to reduce the number of statics we have.

Additionally, in the C code, the events get added to the event list inside the runtime protocol. Are there any additional concerns that we are not doing that, and instead have a different structure?

patina_dxe_core/src/event_db.rs
```diff
@@ -188,17 +188,17 @@ impl Eq for TaggedEventNotification {}
 // whereas efi::Event is used as the public index or handle into the event database.
 // In the code below efi::Event is used to qualify the index/handle type, where as `Event` with
 // scope qualification refers to this private type.
-struct Event {
-    event_id: usize,
-    event_type: EventType,
-    event_group: Option<efi::Guid>,
+pub(crate) struct Event {
+    pub event_id: usize,
```
Javagedes: It does not look like all of the attributes here need to be pub. maybe a few of them, but those can be `inline(always)` functions. instead of direct access to them.

patina_dxe_core/src/runtime.rs
makubacki: I believe the idea was to eliminate dependency on C-like and protocol structures where possible. C uses the protocol to communicate between two binaries - RuntimeDxe and the DXE Core and we don't need to do that.

patina_dxe_core/src/event_db.rs
```diff
@@ -280,6 +280,9 @@ struct EventDb {
 }
 
 impl EventDb {
+    /// Tag used to runtime event IDs.
```
makubacki:
```suggestion
    /// Tag used for runtime event IDs.
```

patina_dxe_core/src/event_db.rs
```diff
@@ -292,16 +295,38 @@ impl EventDb {
...
+            self.events.insert(id, event);
+        }
+
         Ok(id as efi::Event)
     }
 
     fn close_event(&mut self, event: efi::Event) -> Result<(), EfiError> {
         let id = event as usize;
-        self.events.remove(&id).ok_or(EfiError::InvalidParameter)?;
+        if (id & Self::RT_EVENT) != 0 {
```
os-d, line 349: Do we need the RT_EVENT tag that artifically reduces the EVENT space (yes from an absolutely enormous number)? Seems like the only thing it is used for is this check here and we could do the same check that we did to set this (e.g. 

```rust
let runtime =
            (event_type & efi::EVT_RUNTIME) != 0 || event_group == Some(efi::EVENT_GROUP_VIRTUAL_ADDRESS_CHANGE);
```

Could be a macro if we want. Not a big deal to keep the tag, but seems simpler and not having to reduce the event space to not have it?

cfernald: We don't know the event type here, until we know what list to search. We could search both, but I feel like it's better to enforce the expectation this way.

os-d: Oh, I was mixing up efi::event and struct Event, that does have this info. This is fine and I prefer to just pass an id to the caller instead of the actual pointer to the struct event.

patina_dxe_core/src/runtime.rs
cfernald: Yeah, This falls under the category of "lets use rusty things when we can" nothing external needs to consume that C structure, so lets not use it.

Javagedes: Yeah my main concern on the latter half of the question was targeted about consumers of the runtime protocol. Thanks for clarifying that is not a concern.

As for the first half of the question, is there a reason you put it in its own static instead of as a member in runtime data?

kuqin12:
Tested on hardware platform and looks good as well.

PR #741, opened 2025-09-15
RFC: Remove `FunctionComponent` Component implementation

Javagedes:
## Description

RFC to remove the `FunctionComponent` implementation so that only structures (with or without internal state) can be used as a component

Current RFC State: FCP

PR #744, opened 2025-09-15
Remove `FunctionComponent` implementation

Javagedes:
## Description

Reference implementation for #741 

This commit removes the `FunctionComponent` struct, associated blanket
`IntoComponent` implementation for functions, tests, and examples.
Additionally, it updates all relevant documentation to reflect this
change.

- [x] Impacts functionality?
- [ ] Impacts security?
- [x] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

CI

## Integration Instructions

Remove any usage of function components. They can easily be made into struct components in the follwoing example:

```rust
fn reference_function_component(Config<MyConfig>) ->Result<()> {
    Ok(())
}

// Struct component keeping the same name
#[derive(IntoComponent)]
#[entry_point(path = Self::reference_function_component)]
struct NewComponent;

impl NewComponent {
  fn reference_function_component(&self, Config<MyConfig>) ->Result<()> {
      Ok(())
  }
}

// Struct component changing the name
#[derive(IntoComponent)]
struct NewComponent;

impl NewComponent {
  fn entry_point(&self, Config<MyConfig>) ->Result<()> {
      Ok(())
  }
}

```

### DIFF
--- a/patina_dxe_core/src/lib.rs
+++ b/patina_dxe_core/src/lib.rs
@@ -520,12 +520,13 @@ impl Core<Alloc> {
     }
 
     /// Registers core provided components
+    #[allow(clippy::default_constructed_unit_structs)]
     fn add_core_components(&mut self) {
-        self.insert_component(0, decompress::DecompressProtocolInstaller.into_component());
-        self.insert_component(0, systemtables::SystemTableChecksumInstaller.into_component());
-        self.insert_component(0, cpu_arch_protocol::CpuArchProtocolInstaller.into_component());
+        self.insert_component(0, decompress::DecompressProtocolInstaller::default().into_component());
+        self.insert_component(0, systemtables::SystemTableChecksumInstaller::default().into_component());
+        self.insert_component(0, cpu_arch_protocol::CpuArchProtocolInstaller::default().into_component());
         #[cfg(all(target_os = "uefi", target_arch = "aarch64"))]
-        self.insert_component(0, hw_interrupt_protocol::HwInterruptProtocolInstaller.into_component());
+        self.insert_component(0, hw_interrupt_protocol::HwInterruptProtocolInstaller::default().into_component());
     }
 
     /// Starts the core, dispatching all drivers.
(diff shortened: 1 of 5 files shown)


components/patina_samples/src/function_component.rs
makubacki, line 1: @Javagedes, I plan to review this on Thursday.

patina_dxe_core/src/lib.rs
```diff
@@ -513,11 +521,11 @@ impl Core<Alloc> {
...
     fn add_core_components(&mut self) {
-        self.insert_component(0, decompress::install_decompress_protocol.into_component());
-        self.insert_component(0, systemtables::register_checksum_on_protocol_install_events.into_component());
-        self.insert_component(0, cpu_arch_protocol::install_cpu_arch_protocol.into_component());
+        self.insert_component(0, decompress::DecompressProtocolInstaller.into_component());
+        self.insert_component(0, systemtables::SystemTableChecksumInstaller.into_component());
+        self.insert_component(0, cpu_arch_protocol::CpuArchProtocolInstaller.into_component());
         #[cfg(all(target_os = "uefi", target_arch = "aarch64"))]
-        self.insert_component(0, hw_interrupt_protocol::install_hw_interrupt_protocol.into_component());
+        self.insert_component(0, hw_interrupt_protocol::HwInterruptProtocolInstaller.into_component());
```
makubacki: It would be more consistent to derive `Default` on these types and then have something like:

```rust
        self.insert_component(0, decompress::DecompressProtocolInstaller::default().into_component());
        self.insert_component(0, systemtables::SystemTableChecksumInstaller::default().into_component());
        self.insert_component(0, cpu_arch_protocol::CpuArchProtocolInstaller::default().into_component());
        #[cfg(all(target_os = "uefi", target_arch = "aarch64"))]
        self.insert_component(0, hw_interrupt_protocol::HwInterruptProtocolInstaller::default().into_component());
```

My understanding is that this works now because these are unit structures but the instance will not be created automatically like this if they're no longer unit structs in the future.

makubacki:
I think the changes in `basic_hob_usage.rs`, most clearly show how this is makes component usage in the code more obvious and is a good change.

patina_dxe_core/src/lib.rs
Javagedes: Updated, thanks!

PR #745, opened 2025-09-16
ParamFunction: Support Additional Self Types

Javagedes:
## Description

This commit enhances the blanket `ParamFunction` trait implementation to support functions that take `&Self::In` and `&mut Self::In` in addition to `Self::In`.

What this translates to is that struct / enum components using the `#[derive(IntoComponent)]` macro can now define entry point functions that take `&self` or `&mut self` as their first parameter (in addition to `self`). This enhancement paves the way for future support of multi-callable dependency injectable types, including but not limited to components.

Additionally, this commit introduces a custom error message when a function is being used as a `ParamFunction` but does not meet the function signature requirements. This should help users quickly identify and rectify issues with their function signatures.

Improves the error message when the function you declare as your entry point does not meet the requirements of being a `ParamFunction`

image

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

CI continues to pass. Existing components continue to execute as expected.

## Integration Instructions

N/A - But users may now have functions with the interface `&self` or `&mut self`

### DIFF
--- a/sdk/patina_sdk/src/component/params.rs
+++ b/sdk/patina_sdk/src/component/params.rs
@@ -163,9 +163,15 @@ pub unsafe trait Param {
     fn init_state(storage: &mut Storage, meta: &mut MetaData) -> Self::State;
 }
 
-/// A hidden marker to differentiate between functions that have input and those that do not.
+/// A hidden marker for functions that consume their input (take `In` by value).
+/// These functions can only be executed once since they take ownership.
 #[doc(hidden)]
-pub struct HasInput;
+pub struct RunOnce;
+
+/// A hidden marker for functions that borrow their input (take `&In` or `&mut In`).
+/// These functions can be executed multiple times since they don't consume the input.
+#[doc(hidden)]
+pub struct RunMany;
 
 /// A trait that must be implemented by all components that have input.
 #[doc(hidden)]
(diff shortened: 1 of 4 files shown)


sdk/patina_sdk/src/component/params.rs
```diff
@@ -163,9 +163,13 @@ pub unsafe trait Param {
     fn init_state(storage: &mut Storage, meta: &mut MetaData) -> Self::State;
 }
 
-/// A hidden marker to differentiate between functions that have input and those that do not.
+/// A hidden marker to differentiate between functions that take `self` and `&self` / `&mut self`.
```
makubacki: Please let me know if I misunderstand, but I thought these was the intention of these marker traits.

```suggestion
/// A hidden marker for functions that consume their input (take `In` by value).
/// These functions can only be executed once since they take ownership.
```

sdk/patina_sdk/src/component/params.rs
```diff
@@ -163,9 +163,13 @@ pub unsafe trait Param {
...
 #[doc(hidden)]
-pub struct HasInput;
+pub struct RunOnce;
+
+/// A hidden marker to differentiate between functions that take `self` and `&self` / `&mut self`.
```
makubacki:
```suggestion
/// A hidden marker for functions that borrow their input (take `&In` or `&mut In`).
/// These functions can be executed multiple times since they don't consume the input.
```

sdk/patina_sdk/src/component/params.rs
```diff
@@ -238,7 +249,61 @@ macro_rules! impl_param_function {
...
                 }
                 let ($($param,)*) = param_value;
-                call_inner(self, input, $($param),*)
+                call_inner(self, input.take().unwrap(), $($param),*)
+            }
+        }
+
+        #[allow(unused_variables)]
+        #[allow(non_snake_case)]
+        impl<In, Out, Func, $($param: Param),*> ParamFunction<(RunMany, fn(&mut In, $($param,)*)->Out)> for Func
```
makubacki, line 260: It seems like there should be some additional tests for these signatures?

Javagedes: This is difficult to codify via pass / fail tests as these are macros that just generate other code. But I've now added tests in `struct_component.rs` that specifically attempt to make components that have entry points of `self`, `&self` and `&mut self`. These will result in compilation failures rather than test failures if we lose that capability due to changes.

PR #746, opened 2025-09-16
Refactor Patina Perf MM Code and Use the MM Comm Service [Rebase & FF]

### LINKED ISSUE #444: [Task]: Update the `patina_performance` component to use the MM Communication service
### Feature Overview

The `patina-performance` component directly interacts with the MM Communication region today. Update it to use the MM Communicate service.

### Urgency

Medium

I will implement the feature

No maintainer feedback needed

### Anything else?

_No response_

makubacki:
## Description

Closes #444 

Refactors performance code interacting with MM to be more ergonomic and use the MM Communication service (instead of the MM Communication protocol). Use zerocopy in some places to reduce overhead from scroll. General other improvements for error handling and adding more testing.

---

**perf: Update dependencies for MM communication support**

Add patina_mm dependency to workspace and performance component to
enable MM communication. Replace scroll with zerocopy for improved
serialization performance and reduced dependency weight.

- Add patina_mm dependency to workspace Cargo.toml
- Add patina_mm and zerocopy dependencies to performance component
- Remove scroll dependency in favor of zerocopy for serialization
- Add zerocopy with derive features to SDK

This establishes the foundation for modernizing MM communication in the
performance component by replacing HOB-based communication with a
service-based approach.

---

**patina_performance: Add MM module**

Add a dedicated MM communication module for performance related
MM definitions and functionality such as record fetching.

- SmmCommHeader is zerocopy wire format compatible
- Adds GetRecordSize for querying total performance record size
- Adds GetRecordDataByOffset for chunked record retrieval wit
  configurable buffer sizes

---

**patina_sdk/performance: Improve perf record serialization infra**

- Adds the PerformanceRecordHeader struct to represent and manage
  performance record headers.
- Replaces scroll::Error with Error::Serialization so scroll errror
  types are not leaking into performance code.
- Implement direct little-endian serialization helpers for certain
  primitives
- Favors zerocopy instead of scroll in relevant places.
- Adds improved bounds checking for serialization operations.

---

**patina_performance: Update configuration with explicit defaults**

Enhance `PerfConfig` with explicit default values and improved
documentation to provide clearer configuration semantics.

- Add explicit `DEFAULT_ENABLE_COMPONENT` and
  `DEFAULT_ENABLED_MEASUREMENTS` constants
- Implement custom `Default` trait instead of derive to better
  control defaults
- Improve documentation
- Improve `PerfConfig` usage ergonomics

---

**patina_performance: Use the MM Communication service**

Drop HOB-based MM comm buffers and the Communicate protocol being
used in the performance component to instead use the Communicate
service.

Implements more robust error handling, general code cleanup, and
additional tests.

- Uses the `MmCommunication` service
- Adds better error handling using the `MmPerformanceError` enum
- Implements chunked MM record fetching with configurable buffer
  sizes
- Adds `PerformanceRecordIterator` for more memory-efficient perf
  record processing

- Adds unit tests covering different scenarios such as zero records,
  single record, and multi-chunk scenarios
- Mocks the `MmCommunication` service implementations to better
  isolate testing

---

**patina_sdk: Refactor perf measurement module to drop MM comm protocol**

Remove protocol-based MM communication in the patina_sdk perf code
and clean up the `event_callback` namespace.

- Remove Vec import as it's no longer needed for temporary buffers
- Remove fetch_and_add_mm_performance_records from event_callback
  module
- Clean up unused imports and simplify module dependencies

---

**patina_sdk/performance: Remove temp SMM module**

Remove the temporary _smm module that provided scroll-based MM
communication.

The new MM communication approach in the mm.rs module provides
the same functionality with improved error handling, better
resource management, and integration with the Patina service
architecture.

---

**patina_performance: Add records module**

Adds structures that represent performance records so they can
be parsed and printed in debug messages.

---

**patina_mm: Add MM communication buffer update protocol**

Implements the MM communication buffer update protocol to support`
dynamic updates to MM communication buffers during boot. This is
an option that can be enabled in the MM Comm buffer config. The
protocol is currently published the Mu MM Supervisor Standalone MM
implementation.

High-level changes:

- Add support for the MmCommBufferUpdate protocol
- Update the patina_mm component to support MM Comm buffer updates
- Update patina_mm config to support MM Comm buffer updates
- Integration with SW MMI manager for buffer update handling
- Component integration tests for buffer update functionality

An example of a pending MM Communication buffer being applied when
the perf component triggers a SW MMI:

```
DEBUG - Starting MM communication: buffer_id=1, data_size=40, recipient=C095791A-3001-47B2-80C9-EAC7319F2FA4
INFO - Pending buffer update detected, applying now
INFO - Adding new comm buffer 1: addr=0x7ddd9000, size=0x10000
INFO - Successfully applied pending comm buffer 1 update
TRACE - patina_mm\src\component\communicator.rs:466: Buffer validation: buffer_len=65536, required_len=64
TRACE - patina_mm\src\component\communicator.rs:473: Resetting the comm buffer and internal tracking state
TRACE - patina_mm\src\component\communicator.rs:476: Setting up communication buffer for MM request
TRACE - patina_mm\src\config.rs:569: Setting message info for buffer 1: recipient=C095791A-3001-47B2-80C9-EAC7319F2FA4
TRACE - patina_mm\src\config.rs:540: Validating capacity for buffer 1: buffer_size=65536, message_size=0
TRACE - patina_mm\src\config.rs:558: Buffer 1 capacity validation passed
TRACE - patina_mm\src\config.rs:527: Buffer 1 state consistency was verified successfully
TRACE - patina_mm\src\config.rs:586: Message info set successfully for buffer 1
TRACE - patina_mm\src\config.rs:598: Setting message for buffer 1: message_size=40
TRACE - patina_mm\src\config.rs:540: Validating capacity for buffer 1: buffer_size=65536, message_size=40
TRACE - patina_mm\src\config.rs:558: Buffer 1 capacity validation passed
TRACE - patina_mm\src\config.rs:610: Buffer 1: writing header and message data
TRACE - patina_mm\src\config.rs:527: Buffer 1 state consistency was verified successfully
DEBUG - Buffer 1 message set successfully: header_size=24, message_size=40
DEBUG - Outgoing MM communication request: buffer_id=1, data_size=40, recipient=C095791A-3001-47B2-80C9-EAC7319F2FA4
DEBUG - Request Data (hex): [01, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00]
TRACE - patina_mm\src\component\communicator.rs:488: Comm buffer before request: CommunicateBuffer(id: 0x1. len: 0x10000)
. . .
```

---

**Add additional MM Communicator unit tests**

Coverage for:

- Debug output
- Default trait implementation
- Using an MM executor

---

Improve MM perf record debuggability

Prints additional information about MM performance records retrieved,
including record-specific information. That allows better visibility
into the records themselves and MM communication between the Patina
Performance component the MM performance code.

An example of two MM performance records received from the MM UEFI
variable driver:

```
DEBUG - MM communication response received: size=156
INFO - Performance: Processing 116 bytes of MM performance data
DEBUG - Performance: MM record #1 - type: 0x1011 (Dynamic String Event), length: 58, revision: 1, data_len: 54
DEBUG -   Record #1: progress_id: 0x0030, apic_id: 0, timestamp: 4105960673, guid: 7EE2C0C1-C21A-4113-A53A-66824A95696F
DEBUG -     String: "MmVariableServiceInitia"
DEBUG - Performance: MM record #2 - type: 0x1011 (Dynamic String Event), length: 58, revision: 1, data_len: 54
DEBUG -   Record #2: progress_id: 0x0031, apic_id: 0, timestamp: 4157443753, guid: 7EE2C0C1-C21A-4113-A53A-66824A95696F
DEBUG -     String: "MmVariableServiceInitia"
INFO - Performance: MM record summary - total: 2, added: 2, failed: 0
```

---

- [x] Impacts functionality?
- [ ] Impacts security?
- [x] Breaking change?
- [x] Includes tests?
- [x] Includes documentation?

## How This Was Tested

- `cargo make all`
- Boot on QEMU Q35 with default Q35 build and run settings
- Boot on QEMU Q35 with `PERF_TRACE_ENABLE`
- Add two MM perf records for testing and verify they are found using MM communication
- Verify the MM communicate buffer can be updated successfully when the MM Communication Buffer update protocol is published and the following performance configuration is used:

```
  enable_comm_buffer_updates: true,
  updatable_buffer_id: Some(1),
```

## Integration Instructions

Update performance configuration to include the following fields:

```
  enable_comm_buffer_updates: false,
  updatable_buffer_id: None,
```

If the MM Communication Buffer Update protocol is installed on the platform (protocol with GUID `2a22e38f-9d1c-49d0-bdce-7ddac16da45d`), then use the following configration:

```
  enable_comm_buffer_updates: true,
  updatable_buffer_id: Some(1),
```

Where `updatable_buffer_id` is the ID of the comm buffer that should be updated with the information in the protocol.

### DIFF
--- a/components/patina_performance/src/component/performance.rs
+++ b/components/patina_performance/src/component/performance.rs
@@ -30,6 +30,7 @@ use patina::{
         record::{
             GenericPerformanceRecord, PerformanceRecordHeader,
             hob::{HobPerformanceData, HobPerformanceDataExtractor},
+            print_record_details, record_type_name,
         },
         table::FirmwareBasicBootPerfTable,
     },
@@ -358,13 +359,35 @@ where
         return Ok(());
     }
 
+    log::info!("Performance: Processing {} bytes of MM performance data", record_data.len());
+
     let record_iter = PerformanceRecordIterator::new(&record_data);
+    let mut record_count = 0;
+    let mut success_count = 0;
+    let mut error_count = 0;
 
     for record_result in record_iter {
         match record_result {
             Ok(record) => {
+                record_count += 1;
+
+                log::debug!(
+                    "Performance: MM record #{} - type: 0x{:04X} ({}), length: {}, revision: {}, data_len: {}",
+                    record_count,
+                    record.record_type,
+                    record_type_name(record.record_type),
+                    record.length,
+                    record.revision,
+                    record.data.len()
+                );
+                // Print detailed record information based on type
+                print_record_details(record.record_type, record_count, record.data);
+
                 if let Err(e) = fbpt.lock().add_record(record) {
-                    log::error!("Performance: Failed adding MM record: {:?}", e);
+                    error_count += 1;
+                    log::error!("Performance: Failed adding MM record #{}: {:?}", record_count, e);
+                } else {
+                    success_count += 1;
                 }
             }
             Err(e) => {
@@ -373,6 +396,13 @@ where
             }
         }
     }
+
+    log::debug!(
+        "Performance: MM record summary - total: {}, added: {}, failed: {}",
+        record_count,
+        success_count,
+        error_count
+    );
 
     Ok(())
 }


components/patina_performance/src/component/performance.rs
```diff
@@ -87,7 +87,7 @@ impl Performance {
         boot_services: BB,
         runtime_services: RR,
         records_buffers_hobs: Option<P>,
-        mm_comm_region: Option<MmCommRegion>,
+        mm_comm_service: Option<Service<dyn MmCommunication>>,
         fbpt: &'static TplMutex<'static, F, B>,
     ) -> Result<(), EfiError>
     where
```
Javagedes, line 97: This entry point has been made overly abstract for testing purposes. It can be reduced to just 

```rust
fn entry_point<B, R, F>(
    self,
    boot_services: B,
    runtime_services: R,
    records_buffers_hobs: Option<Hob<HobPerformanceData>>,
    fbpt: &'static TplMutex<'static, F, B>,
) -> Result<(), EfiError>
where
    B: BootServices + Clone + 'static,
    R: RuntimeServices + Clone + 'static,
    F: FirmwareBasicBootPerfTable,
{ ... }
```

I did a test sweep of the change and was able to compile. I did not test further. Not sure if we want this to be a change in this PR or I can make an issue and add it in after - but this interface (and a few others that cascade from it) are overly complex.

components/patina_performance/src/component/performance.rs
```diff
@@ -135,19 +135,21 @@ impl Performance {
...
                 EventType::NOTIFY_SIGNAL,
                 Tpl::CALLBACK,
-                Some(event_callback::fetch_and_add_mm_performance_records),
-                Box::new((BB::clone(&boot_services), mm_comm_region, fbpt)),
+                Some(fetch_and_add_mm_performance_records::<BB, B, F>),
+                Box::new((BB::clone(&boot_services), fbpt, mm_comm_service)),
                 &EVENT_GROUP_READY_TO_BOOT,
             )?;
         } else {
             log::info!(
```
Javagedes: would we want this to be a warning possibly?

components/patina_performance/src/component/performance.rs
```diff
@@ -167,25 +169,296 @@ impl Performance {
     }
 }
 
+/// Error types for MM performance record operations
+#[derive(Debug)]
+enum MmPerformanceError {
+    Communication(patina_mm::component::communicator::Status),
```
Javagedes, line 183: even though this is not a public type, so our doc build will not complain about undocumented enum variants, we should still probably document these

components/patina_performance/src/component/performance.rs
```diff
@@ -167,25 +169,296 @@ impl Performance {
...
+            MmPerformanceError::StatusError(status) => {
+                write!(f, "MM operation failed with status: 0x{:x}", status.as_usize())
+            }
+            MmPerformanceError::RecordError(msg) => write!(f, "Record processing error: {msg}"),
+        }
+    }
+}
+
+/// Configuration for MM performance record fetching
+struct MmPerformanceConfig {
```
Javagedes: Is this struct necessary? I don't see a way to configure this configuration, and the values are consts and the only way to instantiate is through `default()` which uses the const values

components/patina_performance/src/component/performance.rs
```diff
@@ -167,25 +169,296 @@ impl Performance {
...
+impl Default for MmPerformanceConfig {
+    fn default() -> Self {
+        Self { chunk_size: SMM_FETCH_CHUNK_BYTES, max_boot_record_bytes: mm::MAX_SMM_BOOT_RECORD_BYTES }
+    }
+}
+
+/// Fetches the total size of MM performance records
+fn fetch_mm_record_size(comm_service: &Service<dyn MmCommunication>) -> Result<usize, MmPerformanceError> {
+    let mut size_req_buf = [0u8; mm::SMM_COMM_HEADER_SIZE];
+    mm::GetRecordSize::new().write_into(&mut size_req_buf).unwrap();
```
Javagedes: should bubble up this unwrap

components/patina_performance/src/component/performance.rs
```diff
@@ -167,25 +169,296 @@ impl Performance {
...
+/// Fetches a chunk of MM performance record data
+fn fetch_mm_record_chunk(
+    comm_service: &Service<dyn MmCommunication>,
+    offset: usize,
+    chunk_size: usize,
+) -> Result<Vec<u8>, MmPerformanceError> {
+    let mut data_req = mm::GetRecordDataByOffset::<SMM_FETCH_CHUNK_BYTES>::new(offset);
+    data_req.boot_record_data_size = chunk_size;
+    let mut data_req_buf = [0u8; mm::SMM_COMM_HEADER_SIZE];
+    data_req.write_into(&mut data_req_buf).unwrap();
```
Javagedes: should bubble up this unwrap

components/patina_performance/src/config.rs
```diff
@@ -36,11 +36,22 @@
 //! SPDX-License-Identifier: Apache-2.0
 //!
 
+/// Default: component disabled unless explicitly enabled by platform/HOB.
+pub const DEFAULT_ENABLE_COMPONENT: bool = false;
```
Javagedes, line 40: I'm not against this, but this seems like overkill :)

components/patina_performance/src/mm.rs
```diff
@@ -0,0 +1,180 @@
...
+
+    pub fn read_from(src: &[u8]) -> Option<(Self, usize)> {
+        let (header, consumed) = SmmCommHeader::read_from(src, Self::FUNCTION_ID)?;
+        Some((Self { boot_record_size: header.boot_record_size(), return_status: header.return_status() }, consumed))
+    }
+}
+
+/// MM communicate function to return a BUFFER_SIZE about of byte at an offset.
+#[derive(Debug, Copy, Clone)]
+pub struct GetRecordDataByOffset<const BUFFER_SIZE: usize> {
```
Javagedes: You could set the default buffer size to reduce boilerplate in other parts of the codebase, if you wanted.

```rust
pub struct GetRecordDataByOffset<const BUFFER_SIZE: usize = SMM_FETCH_CHUNK_BYTES> {
```

sdk/patina_sdk/src/performance/record/extended.rs
```diff
@@ -282,14 +267,82 @@ impl PerformanceRecord for GuidQwordStringEventRecord<'_> {
...
+        write_u64_le(buff, offset, self.timestamp)?;
+        write_bytes(buff, offset, self.guid.as_bytes())?;
+        write_u64_le(buff, offset, self.qword)?;
+        write_bytes(buff, offset, self.string.as_bytes())?;
+        write_u8(buff, offset, 0)?;
         Ok(())
     }
 }
+
+trait WriteLittleEndian {
```
Javagedes: I think this is a good use case for a simple macro - 

```rust
trait IntoLeBytes {
    type Bytes: AsRef<[u8]>;
    fn to_le_bytes(self) -> Self::Bytes;
}

macro_rules! impl_into_le_bytes {
    ($($t:ty),*) => {
        $(
            impl IntoLeBytes for $t {
                type Bytes = [u8; core::mem::size_of::<$t>()];
                fn to_le_bytes(self) -> Self::Bytes {
                    <$t>::to_le_bytes(self)
                }
            }
        )*
    };
}

impl_into_le_bytes!(u8, u16, u32, u64, u128);
```

you could change the macro a bit if you want so that it ends up like this:

```rust
impl_into_le_bytes!(u8)
impl_into_le_bytes!(u16)
...
```

sdk/patina_sdk/src/performance/record.rs
```diff
@@ -13,47 +13,98 @@ pub mod known;
...
+}
+
+impl PerformanceRecordHeader {
+    /// Create a new performance record header.
+    pub const fn new(record_type: u16, length: u8, revision: u8) -> Self {
+        Self { record_type, length, revision }
+    }
+
+    /// Size of the header structure in bytes
+    pub const SIZE: usize = core::mem::size_of::<Self>();
```
Javagedes: nit: while there is no enforced style, I think it makes more sense to put a const value above the function (but at a minimum not in-between functions)

sdk/patina_sdk/src/performance/record.rs
```diff
@@ -13,47 +13,98 @@ pub mod known;
...
+    pub fn try_from_bytes(bytes: &[u8]) -> Option<Self> {
+        if bytes.len() < Self::SIZE {
+            return None;
+        }
+
+        Some(Self { record_type: u16::from_le_bytes([bytes[0], bytes[1]]), length: bytes[2], revision: bytes[3] })
+    }
+
+    /// Convert the header to little-endian bytes.
+    pub fn to_le_bytes(self) -> [u8; mem::size_of::<Self>()] {
```
Javagedes: This seems like something Zero copy should be doing. Can it handle packed types? At a minimum, we need a safety comment on the `unsafe` block

sdk/patina_sdk/src/performance/record.rs
```diff
@@ -13,47 +13,98 @@ pub mod known;
...
-        let record_size_offset = writing_offset;
-        buff.gwrite(0_u8, &mut writing_offset)?;
-        buff.gwrite(self.revision(), &mut writing_offset)?;
+    /// Write just the record payload (not including the common header)
+    /// into `buff` at `offset`, advancing `offset` on success.
+    fn write_data_into(&self, buff: &mut [u8], offset: &mut usize) -> Result<(), Error>;
+
+    /// Serialize the full record (header + payload) into `buff` at `offset`.
+    ///
+    /// On success returns the total size (header + payload). Fails with:
```
Javagedes: Per patina documenting quick reference, errors should go in a `## Errors` section

See: https://sturdy-adventure-nv32gqw.pages.github.io/dev/documenting/reference.html#function-template

components/patina_performance/src/config.rs
makubacki: I can see that. I mainly added it to make the documentation and behavior a little more clear for those less familiar with deriving the `Default` trait. Since platform integrators might end up here that are trying to understand how to configure performance for their platform. If you don't have a strong objection, then I'll leave it in.

sdk/patina_sdk/src/performance/record.rs
makubacki: Moved it to the top of the impl block.

sdk/patina_sdk/src/performance/record.rs
makubacki: It works. I forgot to come back to it after deciding to include zerocopy in these changes. Just needed to derive some zerocopy traits on `PerformanceRecordHeader`.

sdk/patina_sdk/src/performance/record.rs
makubacki: Updated

sdk/patina_sdk/src/performance/record/extended.rs
makubacki: Updated.

components/patina_performance/src/component/performance.rs
makubacki: It stuck around after an earlier refactor. I'll refer to the constants directly and drop it.

components/patina_performance/src/component/performance.rs
makubacki: I think so. Updated.

components/patina_performance/src/component/performance.rs
makubacki: I noticed that too :)

image

Let's defer that. I don't mind making the changes, but I didn't want to increase the scope of this refactor round much more and I'll continue to satisfy clippy for now.

components/patina_performance/src/component/performance.rs
makubacki: Added

components/patina_performance/src/component/performance.rs
makubacki: Done

components/patina_performance/src/component/performance.rs
makubacki: Done

components/patina_performance/src/component/performance.rs
makubacki: Filed https://github.com/OpenDevicePartnership/patina/issues/750 to track.

components/patina_performance/src/component/performance.rs
```diff
@@ -167,25 +167,288 @@ impl Performance {
...
+    type Item = Result<GenericPerformanceRecord<&'a [u8]>, MmPerformanceError>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.bytes.len() < PerformanceRecordHeader::SIZE {
+            return None;
+        }
+
+        let header = match PerformanceRecordHeader::try_from_bytes(self.bytes) {
+            Some(h) => h,
+            None => return Some(Err(MmPerformanceError::RecordError("Failed to parse record header".into()))),
```
joschock: This error state doesn't advance `self.bytes` or otherwise set a flag that could terminate the iterator; this could lead to an infinite loop on error? e.g.
```rust
for record in PerformanceRecordIterator::new(bogus_bytes) {
   print!("Record: {:?}", record);
}
```
would loop forever printing "Failed to parse record header" if the bytes are not perfect I think.

joschock: (I note the usage below breaks the loop manually on error, but might be worth a comment given that it means the iterator needs to be treated specially by the user)

components/patina_performance/src/mm.rs
```diff
@@ -0,0 +1,197 @@
...
+    efi::Guid::from_fields(0xc095791a, 0x3001, 0x47b2, 0x80, 0xc9, &[0xea, 0xc7, 0x31, 0x9f, 0x2f, 0xa4]);
+
+/// MM communicate function to return performance record size info.
+#[derive(Debug, Default, Copy, Clone)]
+pub struct GetRecordSize {
+    pub return_status: efi::Status,
+    pub boot_record_size: usize,
+}
+
+impl GetRecordSize {
```
joschock, line 113: why is the impl for GetRecordSize split?

components/patina_performance/src/mm.rs
```diff
@@ -0,0 +1,197 @@
...
+            Self::FUNCTION_ID,
+            self.return_status,
+            self.boot_record_size,
+            0, // boot_record_data (unused for GetRecordSize)
+            0, // boot_record_offset (unused for GetRecordSize)
+        );
+        header.write_into(dest)
+    }
+
+    pub fn read_from(src: &[u8]) -> Option<(Self, usize)> {
```
joschock: `Option` vs. `Result` -> in this case, seems like you might want `Result`?

components/patina_performance/src/mm.rs
```diff
@@ -0,0 +1,197 @@
...
+        let (header, consumed) = SmmCommHeader::read_from(src, Self::FUNCTION_ID)?;
+        Some((Self { boot_record_size: header.boot_record_size(), return_status: header.return_status() }, consumed))
+    }
+}
+
+/// MM communicate function to return a BUFFER_SIZE about of byte at an offset.
```
joschock: comment wording "about of byte at an offset" is unclear.

sdk/patina_sdk/src/performance/record.rs
```diff
@@ -13,47 +13,103 @@ pub mod known;
...
+    pub const SIZE: usize = core::mem::size_of::<Self>();
+
+    /// Create a new performance record header.
+    pub const fn new(record_type: u16, length: u8, revision: u8) -> Self {
+        Self { record_type, length, revision }
+    }
 
-/// Common behavior of every performance records.
+    /// Try to parse a header from the beginning of a byte slice
+    pub fn try_from_bytes(bytes: &[u8]) -> Option<Self> {
```
joschock: Why not implement the `TryFrom` trait?

components/patina_performance/src/component/performance.rs
makubacki: @joschock, I think some logic to break out of the loop could help. Since I would dissolve the changes back into commits (since this is Rebase & FF), I broke it out as a separate commit here for review https://github.com/OpenDevicePartnership/patina/commit/4b059e4225cfd52c56832620782b9818ffefe471.

Can you check if that's what you had in mind?

components/patina_performance/src/mm.rs
makubacki: It doesn't need to be. Just didn't combine it after a refactor.

components/patina_performance/src/component/performance.rs
joschock: Yep, that looks like it would do the trick.

joschock: Though one thing is that it would keep processing (incrementing by one byte) on malformed header.  Might be better to just kick it to the end if a malformed header is found and not process the rest of the buffer. Chances of it getting back on track after one malformed header seem kind of arbitrary. But I don't feel strongly about this and am fine if you want to keep that approach.

components/patina_performance/src/mm.rs
makubacki: Yeah, I think there is an opportunity to add a custom `mm::ParseError` type and differentiate between the different error conditions in the return value. I'll move it to `Result`.

makubacki: I think it's cleaner: https://github.com/OpenDevicePartnership/patina/commit/114c2fda047660494c0bf57540004b2d119bbcd9

Will merge into PR branch commits.

components/patina_performance/src/mm.rs
makubacki: Updated

sdk/patina_sdk/src/performance/record.rs
makubacki: This just seemed simpler at the time, but `TryFrom` also provides for better error handling. Broke this change out here as https://github.com/OpenDevicePartnership/patina/commit/9e2d1045b184f99e8a46a63236307024ea06d313.

components/patina_performance/src/component/performance.rs
makubacki: I wasn't quite sure how to best address that either I thought incrementing by one byte it might get there and the error is returned. If you don't feel strongly then I'll probably keep it this way for now.

makubacki:
I'm holding merging of this PR until a few others go in that this one will build on top of.

Currently blocked on https://github.com/microsoft/mu_feature_mm_supv/issues/541 for compat with MM Supervisor Standalone MM.

makubacki:
> I'm holding merging of this PR until a few others go in that this one will build on top of.
> 
> Currently blocked on [microsoft/mu_feature_mm_supv#541](https://github.com/microsoft/mu_feature_mm_supv/issues/541) for compat with MM Supervisor Standalone MM.

Marking the PR as draft until that is done. Then, I'll rebase and update the PR as needed.

makubacki:
@joschock, @Javagedes, @os-d, @magravel, @berlin-with0ut-return, this PR is now updated to work with the MM Supervisor, taking advantage of a new protocol that updates the comm buffer protocol during boot. That change in particular is made in https://github.com/OpenDevicePartnership/patina/pull/746/commits/ef054ea58be58754bc847382eeef76f74f3a4953.

The new commits since the last review period are:

- [patina/performance: Add record structures](https://github.com/OpenDevicePartnership/patina/pull/746/commits/1796b404b618a95e26c026e453f959f9735eb2bb)
- [patina_mm: Add MM communication buffer update protocol](https://github.com/OpenDevicePartnership/patina/pull/746/commits/ef054ea58be58754bc847382eeef76f74f3a4953)
- [Add additional MM Communicator unit tests](https://github.com/OpenDevicePartnership/patina/pull/746/commits/69b290f5c068438a1f08b2793bd4303fbe26d71b)
- [Improve MM perf record debuggability](https://github.com/OpenDevicePartnership/patina/pull/746/commits/c22a615505c952ed06a671741eaa0cd7d4901064)

makubacki:
Can you please take a look through the current PR? I'll complete the PR after getting two approvals on the latest changes.

components/patina_performance/src/config.rs
```diff
@@ -36,11 +36,22 @@
...
+pub const DEFAULT_ENABLED_MEASUREMENTS: u32 = 0;
+
 /// The configuration for the Patina Performance component.
-#[derive(Debug, Default)]
+#[derive(Debug, Clone, Copy)]
 pub struct PerfConfig {
     /// Indicates whether the Patina Performance component is enabled.
     pub enable_component: bool,
-    /// A wrapper to generate a mask of all enabled measurements.
+    /// Bitmask of enabled measurements (see `patina_sdk::performance::Measurement`).
```
Javagedes: This is `patina` now. It may also be worth doing an actual link to crates.io patina latest

sdk/patina/Cargo.toml
```diff
@@ -45,7 +45,7 @@ serde = { workspace = true, optional = true }
 serde_yaml = { workspace = true, optional = true }
 serde_json = { workspace = true, optional = true }
 spin = { workspace = true }
-zerocopy = { workspace = true }
+zerocopy = { workspace = true, features = ["derive"] }
```
Javagedes: zerocopy suggests that we stick to having zerocopy without "derive" and to use zerocopy-derive separately, to help with compliation speeds. Here is a blurb from their readme

```
derive Provides derives for the core marker traits via the zerocopy-derive crate. These derives are re-exported
from zerocopy, so it is not necessary to depend on zerocopy-derive directly.

However, you may experience better compile times if you instead directly depend on both zerocopy and
zerocopy-derive in your Cargo.toml, since doing so will allow Rust to compile these crates in parallel. To do so,
do not enable the derive feature, and list both dependencies in your Cargo.toml with the same leading non
zero version number; e.g:

[dependencies]
zerocopy = "0.X"
zerocopy-derive = "0.X"
To avoid the risk of [duplicate import errors](https://github.com/google/zerocopy/issues/1587) if one of your dependencies enables zerocopy's derive feature, import derives as use zerocopy_derive::* rather than by name (e.g., use zerocopy_derive::FromBytes).
```

components/patina_mm/src/component/communicator.rs
```diff
@@ -213,6 +264,13 @@ impl MmCommunication for MmCommunicator {
     fn communicate<'a>(&self, id: u8, data_buffer: &[u8], recipient: Guid<'a>) -> Result<Vec<u8>, Status> {
         log::debug!(target: "mm_comm", "Starting MM communication: buffer_id={}, data_size={}, recipient={:?}", id, data_buffer.len(), recipient);
 
+        // Check for and apply any pending buffer updates from a potential protocol callback
+        if let Some(context) = self.notify_context {
+            let mut comm_buffers = self.comm_buffers.borrow_mut();
+            comm_buffer_update::apply_pending_buffer_update(context, &mut comm_buffers);
+            drop(comm_buffers);
```
Javagedes: I don't think this explicit drop is necessary because we are in the scope of the `if let Some(...)`

sdk/patina/Cargo.toml
makubacki: Removed the `derive` feature from zerocopy.

components/patina_performance/src/config.rs
makubacki: Updated with a link to the crate documentation.

components/patina_mm/src/component/communicator.rs
makubacki: Yeah, it's not needed. Dropped it.

makubacki:
If anyone would like to review, please comment in the next hour or so, otherwise I'll rebase and merge it.

makubacki:
> If anyone would like to review, please comment in the next hour or so, otherwise I'll rebase and merge it.

@os-d reached out about taking a look. The PR will stay open.

components/patina_mm/src/component/communicator.rs
```diff
@@ -150,47 +155,93 @@ pub trait MmCommunication {
...
+    /// # Coverage
+    ///
+    /// This function is marked with `#[coverage(off)]` because it requires StandardBootServices
+    /// which is not available in unit tests. It is tested through integration tests.
+    #[coverage(off)]
     fn entry_point(
         mut self,
         storage: &mut Storage,
         sw_mmi_trigger: Service<dyn SwMmiTrigger>,
+        boot_services: StandardBootServices,
```
joschock, line 194: Is it worth making this a Box<dyn BootServices> so you can mock for testing?

components/patina_mm/src/component/communicator/comm_buffer_update.rs
```diff
@@ -0,0 +1,481 @@
...
+    // Clear the pending flag
+    context.has_pending_update.store(false, Ordering::Release);
+    true
+}
+
+/// Protocol notification callback function
+///
+/// This callback is triggered when the MM Communication Buffer Update Protocol is installed.
+/// It reads the protocol data, validates the communication buffer information, and stores
+/// the buffer update The update will be applied by communicate().
```
os-d:
```suggestion
/// the buffer update. The update will be applied by communicate().
```

os-d (approved):

components/patina_mm/src/component/communicator.rs
Javagedes: A box cannot be dependency injected. However, the boot_services can be mocked like below:

```rust
fn entry_point(
    mut self,
    storage: &mut Storage,
    sw_mmi_trigger: Service<dyn SwMmiTrigger>,
    boot_services: StandardBootServices,
    entry_point_inner(...)
}

fn entry_point_inner(
    mut self,
    storage: &mut Storage,
    sw_mmi_trigger: Service<dyn SwMmiTrigger>,
    boot_services: impl BootServices
    ...
}

#[test]
fn my_test() {
    let comp = ...
    let storage = Storage::mock();
    let sw_mmi_trigger = Service::mock(Box::new(MockSwMmiTrigger::new()));
    // Configure in / out of mockall mock of SwMmiTrigger
    let boot_services = MockBootServices::new();
   // Configure in / out of mockall mock of BootServices

  comp.entry_point_inner(
    storage,
    sw_mmi_trigger,
    boot_services.
  );
}
```

makubacki: Coverage was turned off on this particular `entry_point()` function because I didn't see much value. It's basically integrating other pieces unit tested or tested in the [MM integration tests](https://github.com/OpenDevicePartnership/patina/tree/main/components/patina_mm/tests). I can add some coverage if there is a strong preference for that though.

makubacki: I'll proceed as-is and follow up with a test there if we'd like.

PR #755, opened 2025-09-22
Set toolchain to 1.90 and fix issues with allocator_api and clippy.

joschock:
## Description

Sets the toolchain to 1.90, fixes issues with allocator_api and clippy (none of which are expected to be functional/behavior changes). 

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Unit Tests and clippy check pass, confirmed boot on aarch64 hardware. 

## Integration Instructions

N/A

### DIFF
--- a/core/patina_internal_cpu/src/interrupts/x64/interrupt_manager.rs
+++ b/core/patina_internal_cpu/src/interrupts/x64/interrupt_manager.rs
@@ -46,7 +46,7 @@ lazy_static! {
         // means external caller cannot register for double fault call backs.
         // Fix it: Below line is excluded from std builds because rustc fails to
         //        compile with following error "offset is not a multiple of 16"
-        unsafe { idt.double_fault.set_handler_fn(double_fault_handler).set_stack_index(0) };
+        unsafe { idt.double_fault.set_handler_addr(VirtAddr::new(double_fault_handler as *const () as u64)).set_stack_index(0) };
 
         // Initialize the error code vectors. the x86_64 crate does not allow these
         // to be indexed.
@@ -113,11 +113,11 @@ impl InterruptManager for InterruptsX64 {}
 
 /// Handler for double faults.
 ///
-/// Handler for doubel faults that is configured to run as a direct interrupt
+/// Handler for double faults that is configured to run as a direct interrupt
 /// handler without using the normal handler assembly or stack. This is done to
 /// increase the diagnosability of faults in the interrupt handling code.
 ///
-extern "x86-interrupt" fn double_fault_handler(stack_frame: InterruptStackFrame, _error_code: u64) -> ! {
+extern "x86-interrupt" fn double_fault_handler(stack_frame: InterruptStackFrame, _error_code: u64) {
     panic!("EXCEPTION: DOUBLE FAULT\n{stack_frame:#x?}");
 }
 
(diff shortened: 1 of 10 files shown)


core/patina_internal_cpu/src/interrupts/x64/interrupt_manager.rs
```diff
@@ -46,7 +46,7 @@ lazy_static! {
         // means external caller cannot register for double fault call backs.
         // Fix it: Below line is excluded from std builds because rustc fails to
         //        compile with following error "offset is not a multiple of 16"
-        unsafe { idt.double_fault.set_handler_fn(double_fault_handler).set_stack_index(0) };
+        unsafe { idt.double_fault.set_handler_addr(VirtAddr::new(double_fault_handler as *const () as u64)).set_stack_index(0) };
```
Javagedes, line 49: you should probably use core::ptr's `addr_of!` macro here, if it works. not 100% sure it will work in this situation though.

joschock: Ok. This is a weird bug in clippy, I think. I see some chatter in rust-lang about it; but I don't think it is fixed yet.

Cargo.toml
```diff
@@ -7,7 +7,7 @@ members = ["components/*", "core/*", "sdk/*", "patina_dxe_core"]
 version = "8.1.0"
 license = "Apache-2.0"
 edition = "2024"
-rust-version = "1.89"
+rust-version = "1.90"
```
joschock: jumped the gun here; 1.89 still works. will revert in a subsequent push.

core/patina_internal_cpu/src/interrupts/x64/interrupt_manager.rs
joschock: I tried `VirtAdd::new(core::ptr::addr_of!(doubl_fault_handler))`, but it gave the following error:
```
error[E0745]: cannot take address of a temporary
  --> core/patina_internal_cpu/src/interrupts/x64/interrupt_manager.rs:49:86
   |
49 |         unsafe { idt.double_fault.set_handler_addr(VirtAddr::new(core::ptr::addr_of!(double_fault_handler) as u64)).set_stack_index(0) };
   |                                                                                      ^^^^^^^^^^^^^^^^^^^^ temporary value

Cargo.toml
joschock: reverted.

core/patina_internal_cpu/src/interrupts/x64/interrupt_manager.rs
joschock: context: the original construction won't work because clippy says that double_fault_handler can't be `-> !`.  (as part of clippy compilation, not as a lint). But the type definition for double_fault_handler requires that it is, and `set_handler_fn` enforces that. So I elected to change to set_handler_addr() because that allows me to remove the `-> !` from double_fault_handler (to satisfy clippy).

Javagedes: If its easier you could just disable that lint check for that singular line rather than the jumping around, you are doing. I don't have a huge preference :)

joschock: the problem is that (as far as I can tell from the error message) there is no actual lint to disable. i.e. it is causing clippy to fail, but not as the result of a lint check.

makubacki:
@joschock, thank you for holding the PR for extended team review. I think you're good to merge now when you're ready.

PR #757, opened 2025-09-22
docs: Add general docs about memory safety

makubacki:
## Description

Adds documentation that focuses on the fundamental advantages of Rust memory safety and how it applies to Patina.

- `memory_safety_strategy.md`: Meant to be a starting point for a reader to learn (1) what problems exist in C-based FW, (2) how Rust prevents these issues (3) how the Patina DXE Core decision specifically helps address these most immediately in existing firmware implementations.

- `uefi_memory_safety_case_studies.md`: Shows practical examples of where classes of memory safety issues have been a problem in C with sample code of how this could be handled better in Rust.

These are linked to in "patina.md". The goal for these documents is to explain fundamental memory safety offered by Rust and how that differs from approaches available in C and clarify what unsafe means and its overall exposure and role in Patina Rust firmware.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [x] Includes documentation?

## How This Was Tested

- `cargo make all`

## Integration Instructions

- N/A

### DIFF
--- a/docs/src/background/uefi_memory_safety_case_studies.md
+++ b/docs/src/background/uefi_memory_safety_case_studies.md
@@ -42,235 +42,6 @@ vulnerability classes:
 
 ## Detailed CVE Analysis
 
-### CVE-2022-36765: Integer Overflow in CreateHob()
-
-- **CVE Details**: [CVE-2022-36765](https://nvd.nist.gov/vuln/detail/CVE-2022-36765)
-- **CVSS Score**: 7.0 (HIGH)
-- **Vulnerability Type**: [CWE-680](https://cwe.mitre.org/data/definitions/680.html)
-  (Integer Overflow to Buffer Overflow)
-- [Integer Overflow in CreateHob() could lead to HOB OOB R/W](https://github.com/tianocore/edk2/security/advisories/GHSA-ch4w-v7m3-g8wx)
-
-**The Vulnerability**: "EDK2's `CreateHob()` function was susceptible to integer overflow when calculating HOB
-alignment, allowing attackers to trigger buffer overflows."
-
-**Attack Scenario**: An attacker provides `HobLength = 0xFFFA`:
-
-1. `HobLength + 0x7 = 0x10001` (65537) - overflows UINT16 to `0x0001`
-2. `(0x0001) & (~0x7) = 0x0000` - aligned length becomes 0
-3. Function allocates 0 bytes but caller expects 65530 bytes
-4. Subsequent HOB access overflows the HOB buffer
-
-**C Problem**:
-
-```c
-EFI_STATUS
-PeiCreateHob (
-  IN CONST EFI_PEI_SERVICES  **PeiServices,
-  IN UINT16                  Type,
-  IN UINT16                  Length,
-  IN OUT VOID                **Hob
-  )
-{
-  // Vulnerable: No overflow checking
-  HobLength = (UINT16) ((Length + 0x7) & (~0x7));
-  // ... buffer overflow when accessing memory beyond allocated size
-}
-```
-
-**How Rust Prevents This (Quick Defensive Translation)**:
-
-If a similar function signature were retained in a relatively straightforward port of the C code, the code could be
-more defensively written as:
-
-```rust
-impl HobAllocator {
-    pub fn create_hob(&mut self, hob_type: u16, length: u16) -> Result<*mut HobHeader, HobError> {
-        // Checked arithmetic prevents overflow
-        let aligned_length = length
-            .checked_add(7)
-            .ok_or(HobError::LengthOverflow)?
-            & !7;
-
-        // Bounds checking ensures allocation safety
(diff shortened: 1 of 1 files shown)


makubacki:
Filed a bug to track the unrelated CI failure https://github.com/OpenDevicePartnership/patina/issues/759.

docs/src/background/memory_safety_strategy.md
```diff
@@ -0,0 +1,272 @@
...
+executed during a typical boot, and how many of these services are now implemented in memory-safe Rust:
+
+| Type | Service | Implemented in Pure Rust | Call Count (QEMU X64) |
+|------|---------|---------------------------|------------------------|
+| **Driver Support** | ConnectController() | **Yes** | 517 |
+| | DisconnectController() | **Yes** | 0 |
+| **Event** | CheckEvent() | **Yes** | 27,347 |
+| | CloseEvent() | **Yes** | 2,082 |
+| | CreateEvent() | **Yes** | 2,153 |
+| | CreateEventEx() | **Yes** | (combined with ext()) |
```
os-d: I'm a little confused what `ext()` here is. Should that be `CreateEvent()`?

docs/src/background/uefi_memory_safety_case_studies.md
```diff
@@ -0,0 +1,1189 @@
...
+let hob = builder.build(&mut allocator)?;
+```
+
+**How This Helps**:
+
+1. **Compile-Time Overflow Prevention**: `HobLength::new()` uses `checked_add()`, preventing overflows
+2. **Type-Level Guarantees**: The type system ensures only valid lengths can be used to create HOBs
+3. **Ownership-Based Safety**: `HobRef<T>` *owns* its memory region, preventing use-after-free
+
+#### What is `PhantomData` and Why is it Needed Here?
```
os-d: Part of me wonders if this example should come last, because I worry readers will see the first example as using something deeply Rust (PhantomData) and get spooked (no pun intended). This document is for people at the very first stage of their Rust FW journey and I think a simpler example may be better suited for the first entry. But, also fine to overwhelm them with Rust to show how different Rust is in approaching these types of issues.

docs/src/background/uefi_memory_safety_case_studies.md
```diff
@@ -0,0 +1,1189 @@
...
+    unsafe {
+        // Safety: We checked for null above
+        Some(*ptr)
+    }
+}
+```
+
+#### 2. Unsafe Functions Require Safety Documentation
+
+The Rust compiler and tools in the ecosystem enforce that unsafe functions document their safety requirements:
```
os-d, line 351: If there isn't one already, we should file an issue to turn on this clippy lint in patina

docs/src/background/uefi_memory_safety_case_studies.md
```diff
@@ -0,0 +1,1189 @@
...
+6. **Type-Safe Validation**: Variable headers and names are validated before use, preventing corruption-based attacks
+
+**Summary**: The C vulnerability existed because it used unbounded iteration through variable stores without
+checking if the iteration had reached the end of valid memory. The primary attack vector was corrupted variable
+headers that could cause infinite loops or out-of-bounds memory access during variable store traversal. Rust prevents
+this class of vulnerability by preventing invalid accesses in safe code - you cannot access invalid slice
+indices, and iterators automatically handle bounds checking. The `zerocopy` approach also ensures that binary layout
+parsing matches C structures while providing memory safety.
+
+## Additional CVEs Prevented by Rust's Safety Guarantees
```
os-d:
```suggestion
## Additional CVEs Preventable by Rust's Safety Guarantees

```

We haven't actually mitigated any of these, so maybe should say we could.

docs/src/background/memory_safety_strategy.md
makubacki: Yes, that was search/replace error.

docs/src/background/uefi_memory_safety_case_studies.md
makubacki: That makes sense, I also had a similar concern, but I thought it was interesting to cover to give more depth to what Rust offers.

os-d: I could see it either way, the gentler approach vs hitting them over the head with Rust off the bat. Whichever you prefer.

makubacki: I moved it.

docs/src/background/uefi_memory_safety_case_studies.md
makubacki: Updated

docs/src/background/uefi_memory_safety_case_studies.md
makubacki: Tracked in https://github.com/OpenDevicePartnership/patina/issues/761

PR #760, opened 2025-09-22
Fix allocation bug for variable-sized ACPI tables

berlin-with0ut-return:
## Description
Fix a bug where variable-sized ACPI tables (those for which `size_of::<T> < header.length` due to trailing data bytes) were being incorrectly allocated.

- [X] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested
Existing unit and integration tests still pass. miri no longer catches undefined behavior in regards to this allocation (other miri bugs still need to be fixed).

## Integration Instructions
N/A.

### DIFF
--- a/components/patina_acpi/src/acpi_table.rs
+++ b/components/patina_acpi/src/acpi_table.rs
@@ -408,7 +408,7 @@ impl AcpiTable {
             .map_err(|_e| AcpiError::AllocationFailed)?;
 
         // Get the raw pointer to the allocated memory for copying.
-        let dest_alloc = table_page_alloc.into_raw_ptr().ok_or(AcpiError::AllocationFailed)? as *mut u8;
+        let dest_alloc = table_page_alloc.into_raw_ptr().ok_or(AcpiError::AllocationFailed)?;
 
         // Copy entire table into the new allocation.
         // SAFETY: If function preconditions are met, the pointer is valid and points to a valid ACPI table header.


components/patina_acpi/src/acpi_table.rs
```diff
@@ -360,18 +361,55 @@ impl AcpiTable {
...
-        let table = NonNull::from(Box::leak(Box::new_in(
-            table,
-            mm.get_allocator(allocator_type).map_err(|_e| AcpiError::AllocationFailed)?,
-        )))
-        .cast::<Table>();
+        // The current Windows implementation uses the legacy 32-bit FACS pointer in the FADT.
+        // As such, the FACS must be allocated in the lower 32-bit address space.
+        // This workaround can be removed when Windows no longer relies on this field.
+        // The FACS is always allocated in NVS memory.
+        if unsafe { table.signature } == signature::FACS {
```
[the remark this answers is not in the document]
os-d: I haven't checked if this .signature() function already exists, but I was going to suggest having a safe wrapper if not

components/patina_acpi/src/acpi_table.rs
```diff
@@ -360,18 +361,55 @@ impl AcpiTable {
...
+            let facs_alloc = mm
+                .allocate_pages(
+                    1,
+                    AllocationOptions::new()
+                        .with_memory_type(EfiMemoryType::ACPIMemoryNVS)
+                        .with_strategy(PageAllocationStrategy::MaxAddress(0x1000_0000)),
```
os-d:
```suggestion
                        .with_strategy(PageAllocationStrategy::MaxAddress(SIZE_4GB)),
```
from patina_sdk definitions

components/patina_acpi/src/acpi_table.rs
```diff
@@ -360,18 +361,55 @@ impl AcpiTable {
...
+                )
+                .map_err(|_e| AcpiError::AllocationFailed)?;
+            let new_facs_ptr = facs_alloc.into_raw_ptr().unwrap();
+            unsafe { ptr::copy_nonoverlapping(table.as_ref() as *const T as *const u8, new_facs_ptr, total_len) };
```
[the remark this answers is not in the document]
os-d: Agreed that we should align on approach here, whether creating the slice and populating it or doing a box of the type, etc.

components/patina_acpi/src/acpi_table.rs
```diff
@@ -360,18 +361,55 @@ impl AcpiTable {
     pub unsafe fn new<T: 'static>(table: T, mm: &Service<dyn MemoryManager>) -> Result<Self, AcpiError> {
         // SAFETY: If the caller preconditions are met, the signature, header, and table fields of the union are valid.
         let table = unsafe { Table::new(table) }?;
+        let total_len = unsafe { table.header.length } as usize;
```
Javagedes: Is this function now just `AcpiTable::new_from_ptr`, except it takes a concrete type instead of a pointer? If so, we should look making one wrap the other to reduce code duplication

Javagedes: I would add a `fn len() -> usize` function to the `Table` struct, so you don't have to do unsafe here.

Javagedes: At a minimum the code specially for the FACS is basically just the `new_from_ptr` code. I would de-duplicate it somehow.

components/patina_acpi/src/acpi_table.rs
```diff
@@ -360,18 +361,55 @@ impl AcpiTable {
...
-        )))
-        .cast::<Table>();
+        // The current Windows implementation uses the legacy 32-bit FACS pointer in the FADT.
+        // As such, the FACS must be allocated in the lower 32-bit address space.
+        // This workaround can be removed when Windows no longer relies on this field.
+        // The FACS is always allocated in NVS memory.
+        if unsafe { table.signature } == signature::FACS {
+            let facs_alloc = mm
+                .allocate_pages(
+                    1,
```
Javagedes: For safety, seems like you should calculate this hardcoded `1` via a `total_len`. If `total_len` is more than 4K, You would end up with the same issue this is fixing.

components/patina_acpi/src/acpi_table.rs
```diff
@@ -360,18 +361,55 @@ impl AcpiTable {
...
+            return Ok(Self {
+                table: NonNull::new(new_facs_ptr).ok_or(AcpiError::NullTablePtr)?.cast::<Table>(),
+                type_id: TypeId::of::<AcpiFacs>(),
+            });
+        }
+
+        let data_len = total_len - mem::size_of::<AcpiTableHeader>();
+        // Table is header followed by trailing data.
+        // `header.length` specifies the full length of the table, including the header.
+        let tbl_layout = Layout::new::<AcpiTableHeader>()
```
Javagedes: I know I said it may be better to do the `Layout` way of allocating, but doing it this way is preventing you from having the same code path for any type of table. At the moment your code path splits in two based off of if the table is the FACS, either doing a page allocation or a layout allocation, but ultimately doing the same thing (allocating space for the table on the heap)

I think to reduce code complexity you should always do the page allocation method, and have a single, small if statement to add a `PageAllocationStrategy::MaxAddress(SIZE_4GB)` if the table is a FACS

components/patina_acpi/src/acpi_table.rs
berlin-with0ut-return: the logic for `new_from_ptr` and `new` with the new changes is similar. i will look into making a wrapper

components/patina_acpi/src/acpi_table.rs
```diff
@@ -388,53 +377,54 @@ impl AcpiTable {
...
+            ptr::copy_nonoverlapping(header_ptr as *const u8, dest_alloc, table_length);
         }
 
         // Leak the allocated bytes.
-        let raw_header = Box::into_raw(table_bytes.into_boxed_slice()) as *mut AcpiTableHeader;
+        let table = NonNull::new(dest_alloc.cast::<Table>()).ok_or(AcpiError::NullTablePtr)?;
+
+        // Store the table type for convenience.
+        // If the type is unknown (for example, coming over C FFI interface), use AcpiTableHeader as a fallback.
+        let type_id = if let Some(tid) = type_id { tid } else { TypeId::of::<AcpiTableHeader>() };
```
Javagedes:
```suggestion
        let type_id = type_id.unwrap_or(TypeId::of::<AcpiTableHeader>());
```

components/patina_acpi/src/acpi_table.rs
```diff
@@ -388,53 +377,54 @@ impl AcpiTable {
...
+            PageAllocationStrategy::Any
+        };
 
-        // Allocate table based on the length given in the header field.
-        let allocator = mm.get_allocator(EfiMemoryType::ACPIReclaimMemory).map_err(|_e| AcpiError::AllocationFailed)?;
-        let mut table_bytes = Vec::with_capacity_in(table_length, allocator);
+        // Allocate memory in appropriate ACPI region, up to page granularity.
+        let table_page_alloc = mm
+            .allocate_pages(
+                (table_length + UEFI_PAGE_SIZE - 1) / UEFI_PAGE_SIZE,
```
os-d:
```suggestion
                uefi_size_to_pages!(table_length),
```
from patina_sdk

berlin-with0ut-return:
converting to draft please do not review until all merge issues are worked out

PR #762, opened 2025-09-23
patina_dxe_core: Combine Core and UEFI dispatch

Javagedes:
## Description

This commit combines the core component dispatcher and the UEFI driver dispatcher into a single dispatch loop. This loop is broken into two phases. The first phase performs a single pass of core component dispatch while the second phase performs a single pass of UEFI driver dispatch. Should either phase result in at least a single dispatch, the loop will continue. This loop will only exit when a full iteration of both phases results in no dispatches.

This functionality is necessary to produces services that are rust-ified wrappers around C-functionality (such as protocols). While this is not a suggested way of producing services, it is an acknowledged stop-gap until pure rust implementations can be provided. This is specifically necessary for #592, which requires certain protocols to exist before a service can be produced (such as Console support, variable services, etc).

Note: There is an intent to move the core UEFI dispatcher into a multi-dispatchable component, however RFC work needs to be performed to come up with an adequate solution for multi-dispatchable components, and this is needed now to support the aforementioned RFC.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Verified boot to UEFI shell on virtual platform.

## Integration Instructions

N/A

### DIFF
--- a/patina_dxe_core/src/lib.rs
+++ b/patina_dxe_core/src/lib.rs
@@ -378,17 +378,17 @@ impl Core<Alloc> {
         len != self.components.len()
     }
 
-    /// Performs a combined dispatch of core components and UEFI drivers.
+    /// Performs a combined dispatch of Patina components and UEFI drivers.
     ///
     /// This function will continue to loop and perform dispatching until no components have been dispatched in a full
     /// iteration. The dispatching process involves a loop of two distinct dispatch phases:
     ///
-    /// 1. A single iteration of dispatching core components, retaining those that were not dispatched.
+    /// 1. A single iteration of dispatching Patina components, retaining those that were not dispatched.
     /// 2. A single iteration of dispatching UEFI drivers via the dispatcher module.
     fn core_dispatcher(&mut self) -> Result<()> {
         perf_function_begin(function!(), &CALLER_ID, create_performance_measurement);
         loop {
-            // Core component dispatch
+            // Patina component dispatch
             let dispatched = self.dispatch_components();
 
             // UEFI driver dispatch


patina_dxe_core/src/lib.rs
```diff
@@ -349,38 +354,57 @@ impl Core<Alloc> {
...
-                        true // Component dispatched, even if it did fail, so remove from self.components to avoid re-dispatch.
-                    }
+    fn dispatch(&mut self) -> bool {
+        let len = self.components.len();
+        self.components.retain_mut(|component| {
+            // Ok(true): Dispatchable and dispatched returning success
+            // Ok(false): Not dispatchable at this time.
+            // Err(e): Dispatchable and dispatched returning failure
+            let name = component.metadata().name();
+            log::info!("DISPATCH_ATTEMPT BEGIN: Id = [{:?}]", name);
```
os-d: Probably don't need these at info level, probably trace, and just a single info level log for dispatched component x

patina_dxe_core/src/lib.rs
```diff
@@ -349,38 +354,57 @@ impl Core<Alloc> {
...
+
+    /// Performs a combined dispatch of core components and UEFI drivers.
+    ///
+    /// This function will continue to loop and perform dispatching until no components have been dispatched in a full
+    /// iteration. The dispatching process involves a loop of two distinct dispatch phases:
+    ///
+    /// 1. A single iteration of dispatching core components, retaining those that were not dispatched.
+    /// 2. A single iteration of dispatching UEFI drivers via the dispatcher module.
+    fn core_dispatcher(&mut self) -> Result<()> {
+        perf_function_begin(function!(), &CALLER_ID, create_performance_measurement);
```
os-d, line 389: no perf measurements on component dispatch?

patina_dxe_core/src/lib.rs
```diff
@@ -349,38 +354,57 @@ impl Core<Alloc> {
...
+        loop {
+            // Core component dispatch
+            let dispatched = self.dispatch();
+
+            // UEFI driver dispatch
+            let dispatched = dispatched
+                | dispatcher::dispatch().inspect_err(|err| log::error!("UEFI Driver Dispatch error: {err:?}"))?;
```
os-d:
```suggestion
                || dispatcher::dispatch().inspect_err(|err| log::error!("UEFI Driver Dispatch error: {err:?}"))?;
```
I think you are trying to boolean OR here, not bitwise

patina_dxe_core/src/lib.rs
Javagedes: This is actually removing the perf measurements from the C dispatcher and moving it to the overarching dispatcher, which includes the Rust components and C drivers. Maybe a name other than `core_dispatcher` would be more appropriate?

os-d: Oh, yep, I see. Naming is a bit confusing since we have self.dispatcher, dispatcher::dispatcher, and core_dispatcher :). But I'm not hugely against it. I'll resolve this.

joschock:
General comment: looks like component dispatch is only invoked on the first "auto-dispatch" pass; subsequent dispatcher passes (invoked by `gDS->Dispatch()`) will not invoke components. Is that expected?

Javagedes:
> General comment: looks like component dispatch is only invoked on the first "auto-dispatch" pass; subsequent dispatcher passes (invoked by `gDS->Dispatch()`) will not invoke components. Is that expected?

Yes that was my intent. I'm happy to discuss including it, but I was hoping to keep the two concepts separate from eachother

patina_dxe_core/src/lib.rs
```diff
@@ -349,38 +354,54 @@ impl Core<Alloc> {
...
+                    true // Component dispatched, even if it did fail, so remove from self.components to avoid re-dispatch.
                 }
-            });
-            if self.components.len() == len {
+            }
+        });
+        len != self.components.len()
+    }
+
+    /// Performs a combined dispatch of core components and UEFI drivers.
```
makubacki: What is meant by "core components" here? This is dispatching all components, right? If that's the case, "core components" makes this sound like a special subset of components implementing core functionality.

Javagedes: Ah yes good catch. I meant "Patina components". I've updated it in this upcoming push.

PR #769, opened 2025-09-24
patina_dxe_core: Switch to use external RuntimeDxe implementation

cfernald:
## Description

The current core implements SetVirtualAddressMap within the core itself and does not properly fill the runtime protocol. This means that when the OS calls back into SetVirtualAddressMap, they are invoking code that is executing in boot services memory and will often leverage boot service memory allocated structures. This can cause problems as there is no requirement that this memory is still valid and mapped at the point where SetVirtualAddressMap was called.

This commit changes the runtime module to instead of exposing runtime functions, to populate the runtime protocol produced by the RuntimeDxe module to provide the event, and image data are available at runtime. The expectation is now that there will be a separate module that will implement the core runtime routines. For now, this will mean that consumers will need to add the traditional RuntimeDxe driver or similar to handle this logic.

This commit intentionally does not populate the memory map fields of the runtime protocol as they are currently unused. This may change in the future if other implementations are introduced.

- [x] Impacts functionality?
- [ ] Impacts security?
- [x] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- [x] Q35 (MSVC)
- [x] Q35 (GCC)
- [x] SBSA (GCC)
- [x] AArch64 platform
- [x] Intel Platform

## Integration Instructions

Consumers must introduce a traditional RuntimeDxe driver.


cfernald:
Relies on: https://github.com/microsoft/mu_rust_pi/pull/93

patina_dxe_core/src/pecoff.rs
```diff
@@ -314,6 +314,46 @@ pub fn relocate_image(
...
+                IMAGE_REL_BASED_DIR64 => {
+                    flat_data.extend_from_slice(&reloc.value.to_le_bytes());
+                }
+                _ => todo!(), // Other fixups not implemented at this time
+            }
+        }
+    }
+
+    // Double check that the capacity calculation was correct.
+    debug_assert!(flat_data.capacity() == size && flat_data.len() == size);
```
makubacki, line 353: Returning a `Result` would help callers decide if they want to do something different in case of an error. Even if it is just a debug assert.

patina_dxe_core/src/runtime.rs
```diff
@@ -6,401 +6,326 @@
...
+            memory_descriptor_size: 0,
+            memory_descriptor_version: 0,
+            memory_map_size: 0,
+            memory_map_physical: ptr::null_mut(),
+            memory_map_virtual: ptr::null_mut(),
+            virtual_mode: AtomicBool::new(false),
+            at_runtime: AtomicBool::new(false),
+        };
+        let mut data = RuntimeData::new();
+        data.runtime_arch_ptr = &mut protocol;
```
makubacki: Isn't this making `data.runtime_arch_ptr` a dangling pointer? Since `protocol` is local.

patina_dxe_core/src/runtime.rs
```diff
@@ -6,401 +6,326 @@
...
+    Ok(())
 }
 
-pub fn add_runtime_event(event: crate::event_db::Event) -> Result<(), EfiError> {
-    let event_id = event.event_id();
-    RUNTIME_EVENTS.lock().insert(event_id, event);
+pub fn add_runtime_image(
+    image_base: *mut c_void,
+    image_size: u64,
+    relocation_data: &Vec<RelocationBlock>,
```
makubacki: Does the need to be a reference to a `Vec` specifically?

`relocation_data: &[RelocationBlock]` would be more flexible.

patina_dxe_core/src/runtime.rs
makubacki: I think we'd normally leak a box in situations like this in test code.

patina_dxe_core/src/pecoff.rs
cfernald: For this specific check, it is purely checking for internal consistency to make sure it didnt have to expand or over-allocate. I don't think it's necessary for the caller to be involved.

cfernald: Will resolve for now, let me know if you disagree.

makubacki: I was thinking more for the case that debug asserts are disabled so the caller is aware something is not correct. I agree leaving it as-is is also fine.

kuqin12:
Tested and functional on proprietary ARM64 hardware platform.

PR #773, opened 2025-09-25
[Feature]: Add support to load an image from a FV device path

AnhDLuong:
## Description

Port part of GetFileBufferFromFv in EDK2 and rename it as add get_file_buffer_from_fw.

- [X] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Verified booting to UEFI shell in qemu and on hardware.

## Integration Instructions

N/A

### DIFF
--- a/patina_dxe_core/src/image.rs
+++ b/patina_dxe_core/src/image.rs
@@ -7,7 +7,7 @@
 //! SPDX-License-Identifier: Apache-2.0
 //!
 use alloc::{boxed::Box, collections::BTreeMap, string::String, vec, vec::Vec};
-use core::{convert::TryInto, ffi::c_void, mem::transmute, slice::from_raw_parts};
+use core::{convert::TryInto, ffi::c_void, mem::transmute, slice, slice::from_raw_parts};
 use goblin::pe::section_table;
 use patina::base::{DEFAULT_CACHE_ATTR, UEFI_PAGE_SIZE, align_up};
 use patina::error::EfiError;
@@ -17,7 +17,11 @@ use patina::performance::{
 };
 use patina::{guids, uefi_pages_to_size, uefi_size_to_pages};
 use patina_internal_device_path::{DevicePathWalker, copy_device_path_to_boxed_slice, device_path_node_count};
-use patina_pi::hob::{Hob, HobList};
+use patina_pi::{
+    fw_fs::FfsSectionRawType::PE32,
+    hob::{Hob, HobList},
+    protocols::firmware_volume,
+};
 use r_efi::efi;
 
 use crate::{
@@ -39,6 +43,7 @@ use crate::{
     tpl_lock,
 };
 
+use efi::Guid;
 use uefi_corosensei::{
     Coroutine, CoroutineResult, Yielder,
     stack::{MIN_STACK_SIZE, STACK_ALIGNMENT, Stack, StackPointer},
@@ -765,8 +770,9 @@ fn get_buffer_by_file_path(
         Err(EfiError::InvalidParameter)?;
     }
 
-    //TODO: EDK2 core has support for loading an image from an FV device path which is not presently supported here.
-    //this is the only case that Ok((buffer, true, authentication_status)) would be returned.
+    if let Ok((buffer, device_handle)) = get_file_buffer_from_fw(file_path) {
+        return Ok((buffer, true, device_handle, 0));
+    }
 
     if let Ok((buffer, device_handle)) = get_file_buffer_from_sfs(file_path) {
         return Ok((buffer, false, device_handle, 0));
@@ -786,6 +792,57 @@ fn get_buffer_by_file_path(
     }
 
     Err(EfiError::NotFound)
+}
+
+fn get_file_guid_from_device_path(path: *mut efi::protocols::device_path::Protocol) -> Result<Guid, EfiError> {
+    let mut walker = unsafe { DevicePathWalker::new(path) };
+    let file_path_node = walker.next().ok_or(EfiError::InvalidParameter)?;
+    if file_path_node.header().r#type != efi::protocols::device_path::TYPE_MEDIA
+        || file_path_node.header().sub_type != efi::protocols::device_path::Media::SUBTYPE_PIWG_FIRMWARE_FILE
+    {
+        return Err(EfiError::InvalidParameter);
+    }
+    Ok(Guid::from_bytes(file_path_node.data().try_into().map_err(|_| EfiError::BadBufferSize)?))
+}
+
+fn get_file_buffer_from_fw(
+    file_path: *mut efi::protocols::device_path::Protocol,
+) -> Result<(Vec<u8>, efi::Handle), EfiError> {
+    // Locate the handles to a device on the file_path that supports the firmware volume protocol
+    let (remaining_file_path, handle) = core_locate_device_path(firmware_volume::PROTOCOL_GUID, file_path)?;
+
+    // For FwVol File system there is only a single file name that is a GUID.
+    let fv_name_guid = get_file_guid_from_device_path(remaining_file_path)?;
+
+    // Get the firmware volume protocol
+    let fv_ptr =
+        PROTOCOL_DB.get_interface_for_handle(handle, firmware_volume::PROTOCOL_GUID)? as *mut firmware_volume::Protocol;
+    if fv_ptr.is_null() {
+        debug_assert!(!fv_ptr.is_null(), "ERROR: get_interface_for_handle returned NULL ptr for FirmwareVolume!");
+        return Err(EfiError::InvalidParameter);
+    }
+    let fw_vol = unsafe { fv_ptr.as_ref().unwrap() };
+
+    // Read image from the firmware file
+    let mut buffer: *mut u8 = core::ptr::null_mut();
+    let buffer_ptr: *mut *mut c_void = &mut buffer as *mut _ as *mut *mut c_void;
+    let mut buffer_size = 0;
+    let mut authentication_status = 0;
+    let authentication_status_ptr = &mut authentication_status;
+    let status = (fw_vol.read_section)(
+        fw_vol,
+        &fv_name_guid,
+        PE32,
+        0, // Instance
+        buffer_ptr,
+        core::ptr::addr_of_mut!(buffer_size),
+        authentication_status_ptr,
+    );
+
+    EfiError::status_to_result(status)?;
+
+    let section_slice = unsafe { slice::from_raw_parts(buffer, buffer_size) };
+    Ok((section_slice.to_vec(), handle))
 }
 
 fn get_file_buffer_from_sfs(


patina_dxe_core/src/image.rs
```diff
@@ -767,6 +768,9 @@ fn get_buffer_by_file_path(
 
     //TODO: EDK2 core has support for loading an image from an FV device path which is not presently supported here.
```
rogurr: There are no other comments for each call in this function and I don't think we need the statement about this being the only 'Ok" case, so can we just remove this?

At the same time, can we also update this function's header?  Line 759 was incorrect prior to this change and we probably need to update line 760 to state only 'authentication_status' is not supported.

patina_dxe_core/src/image.rs
```diff
@@ -788,6 +792,93 @@ fn get_buffer_by_file_path(
...
+    device_path: *const efi::protocols::device_path::Protocol,
+) -> Option<efi::Guid> {
+    let fv_device_path_node_ptr = device_path as *const MediaFwVolFilePathDevicePath;
+    if fv_device_path_node_ptr.is_null() {
+        return None;
+    }
+
+    let fv_device_path_node = unsafe { *fv_device_path_node_ptr };
+
+    // if fv_device_path_node.header.r#type == efi::protocols::device_path::TYPE_MEDIA && fv_device_path_node.header.sub_type == efi::protocols::device_path::Media::SUBTYPE_FIRMWARE_VOLUME_FILE {
```
rogurr: Looks like this was a first implementation of the code starting at line 816 and a remnant that can be removed?

patina_dxe_core/src/image.rs
```diff
@@ -788,6 +792,93 @@ fn get_buffer_by_file_path(
     Err(EfiError::NotFound)
 }
 
+#[repr(C)]
+#[derive(Clone, Copy, Debug)]
+pub struct MediaFwVolFilePathDevicePath {
```
rogurr: Not a bug, just curios on what would be a good location for this code.

Started [discussion 774](https://github.com/OpenDevicePartnership/patina/discussions/774) to determine if the Patina end goal is to have a central device path crate or if things like this should reside locally.

patina_dxe_core/src/image.rs
```diff
@@ -788,6 +792,93 @@ fn get_buffer_by_file_path(
...
+    // Read image from the firmware file
+    let mut buffer: *mut u8 = core::ptr::null_mut();
+    let buffer_ptr: *mut *mut c_void = &mut buffer as *mut _ as *mut *mut c_void;
+    let mut buffer_size = 0;
+    let mut authentication_status = 0;
+    let authentication_status_ptr = &mut authentication_status;
+    let status = ((*fw_vol).read_section)(
+        fw_vol,
+        &fv_name_guid,
+        16, // EFI_SECTION_PE32
```
rogurr: You are using mu_pi, correct?  It looks like there is a define for this in the crate if you want to use the pre-defined values instead of a hard coded value.

https://docs.rs/mu_pi/6.0.2/src/mu_pi/fw_fs/ffs/section.rs.html#27

patina_dxe_core/src/image.rs
```diff
@@ -788,6 +792,93 @@ fn get_buffer_by_file_path(
...
+                    return None;
+                }
+            }
+        },
+        _ => {
+            return None;
+        }
+    }
+}
+
```
kouchekiniad:
```suggestion
fn get_file_guid_from_device_path(path: *mut efi::protocols::device_path::Protocol) -> Result<Guid, EfiError> {
    let mut walker = unsafe { DevicePathWalker::new(path) };
    let file_path_node = walker.next().ok_or(EfiError::InvalidParameter)?;

    if file_path_node.header().r#type != efi::protocols::device_path::TYPE_MEDIA ||
        file_path_node.header().sub_type != efi::protocols::device_path::Media::SUBTYPE_PIWG_FIRMWARE_FILE {
        return Err(EfiError::InvalidParameter);
    }

    return Ok(Guid::from_bytes(file_path_node.data().try_into().map_err(|_| EfiError::BadBufferSize)?));
}

```

kouchekiniad: Hi @AnhDLuong, I'd suggest this simpler implementation that leverages DevicePathWalker

patina_dxe_core/src/image.rs
```diff
@@ -788,6 +792,93 @@ fn get_buffer_by_file_path(
...
+    let firmware_volume_protocol_guid = &firmware_volume::PROTOCOL_GUID as *const _ as *mut efi::Guid;
+    let mut fw_vol: *mut firmware_volume::Protocol = core::ptr::null_mut();
+    let fw_vol_ptr: *mut *mut c_void = &mut fw_vol as *mut _ as *mut *mut c_void;
+    let status = handle_protocol(handle, firmware_volume_protocol_guid, fw_vol_ptr);
+    if status != efi::Status::SUCCESS {
+        return Err(EfiError::NotFound);
+    }
+
+    let fw_vol =
+        unsafe { (fw_vol as *mut firmware_volume::Protocol).as_mut().ok_or(EfiError::Unsupported)? };
```
kouchekiniad: Hi @AnhDLuong, I'd suggest leveraging `PROTOCOL_DB.get_interface_for_handle` here:

```suggestion
    let fv_ptr = PROTOCOL_DB.get_interface_for_handle(fv_handle, firmware_volume::PROTOCOL_GUID)? as *mut firmware_volume::Protocol;
    let fv = unsafe {
        fv_ptr.as_ref().expect("get_interface_for_handle returned NULL ptr for FirmwareVolume")
    };
```

patina_dxe_core/src/image.rs
```diff
@@ -788,6 +792,93 @@ fn get_buffer_by_file_path(
...
+        0,  // Instance
+        buffer_ptr,
+        core::ptr::addr_of_mut!(buffer_size),
+        authentication_status_ptr,
+    );
+    
+    unsafe {
+        let file_buffer = Vec::from_raw_parts(buffer, buffer_size, buffer_size);
+        EfiError::status_to_result(status).map(|_| (file_buffer, handle))
+    }
```
kouchekiniad: Hi @AnhDLuong, I think we'll need to check status before trying to construct a buffer; If `read_section` fails and buffer_ptr isn't set, this code would be constructing a Vec at address 0. 

Also, since a mutable Vec is requested instead of a &[u8], I'd lean towards copying the data, rather than returning a Vec over potentially RO memory.

```suggestion
    EfiError::status_to_result(status)?;
    
    let section_slice = unsafe { slice::from_raw_parts(buffer, buffer_size) };
    return Ok((section_slice.to_vec(), handle));
```

patina_dxe_core/src/image.rs
AnhDLuong: Agree

patina_dxe_core/src/image.rs
AnhDLuong: Agree

patina_dxe_core/src/image.rs
AnhDLuong: Good point!

patina_dxe_core/src/image.rs
AnhDLuong: Good point!

patina_dxe_core/src/image.rs
AnhDLuong: Good point!

patina_dxe_core/src/image.rs
AnhDLuong: Good point!

patina_dxe_core/src/image.rs
```diff
@@ -43,6 +43,9 @@ use uefi_corosensei::{
     Coroutine, CoroutineResult, Yielder,
     stack::{MIN_STACK_SIZE, STACK_ALIGNMENT, Stack, StackPointer},
 };
+use mu_pi::{protocols::firmware_volume};
```
Javagedes: nit: please add these definitions to the other `mu_pi` definitions towards the top of the file

patina_dxe_core/src/image.rs
```diff
@@ -43,6 +43,9 @@ use uefi_corosensei::{
...
+use mu_pi::fw_fs::FfsSectionRawType::PE32;
+use patina_sdk::component::service::Guid;
```
Javagedes: #788 is adding a new standard (for the core) GUID type. Depending on whose PR goes in first (This or #788), one of them will have to update this code :)

patina_dxe_core/src/image.rs
```diff
@@ -788,6 +792,59 @@ fn get_buffer_by_file_path(
...
+
+fn get_file_buffer_from_fw (
+    file_path: *mut efi::protocols::device_path::Protocol,
+) -> Result<(Vec<u8>, efi::Handle), EfiError> {
+    // Locate the handles to a device on the file_path that supports the firmware volume protocol
+    let (remaining_file_path, handle) =
+        core_locate_device_path(firmware_volume::PROTOCOL_GUID, file_path)?;
+
+    // For FwVol File system there is only a single file name that is a GUID.
+    let fv_name_guid = get_file_guid_from_device_path(remaining_file_path);
```
Javagedes: You should be able to reduce all of the code to get the fv_name_guid (currently L813 - L818) to just this:

```suggestion
    let fv_name_guid = get_file_guid_from_device_path(remaining_file_path)
       .map_err(|_| EfiError::InvalidParameter)?;
```

You could also bubble up the existing error, as they are the same type, instead of mapping it to always InvalidParameter. But I'll leave that up to you.

patina_dxe_core/src/image.rs
```diff
@@ -788,6 +792,59 @@ fn get_buffer_by_file_path(
...
+    if fv_name_guid.is_err() {
+        return Err(EfiError::InvalidParameter);
+    }
+
+    let fv_name_guid = fv_name_guid.unwrap();
+
+    // Get the firmware volume protocol
+    let fv_ptr = PROTOCOL_DB.get_interface_for_handle(handle, firmware_volume::PROTOCOL_GUID)? as *mut firmware_volume::Protocol;
+    let fw_vol = unsafe {
```
Javagedes: We should do a null ptr check here.

AnhDLuong: @Javagedes , can you please elaborate? "fv_ptr.as_ref().expect("get_interface_for_handle returned NULL ptr for FirmwareVolume")" will panic if fv_ptr is null. Hence, the fw_vol should be a valid pointer. Is it not the case?

Javagedes: Yes but at least on release builds we don't want to panic, which stops everything - we instead want to return in error so we continue running.

Typically we do something like 

```
if ptr.is_null() {
  debug_assert!(...)
  // return an error on release builds
  return Err(EfiError::InvalidParameter);
}
```

patina_dxe_core/src/image.rs
```diff
@@ -788,6 +794,57 @@ fn get_buffer_by_file_path(
...
+    let fv_name_guid = get_file_guid_from_device_path(remaining_file_path)?;
+
+    // Get the firmware volume protocol
+    let fv_ptr =
+        PROTOCOL_DB.get_interface_for_handle(handle, firmware_volume::PROTOCOL_GUID)? as *mut firmware_volume::Protocol;
+    if fv_ptr.is_null() {
+        debug_assert!(fv_ptr.is_null(), "ERROR: get_interface_for_handle returned NULL ptr for FirmwareVolume!");
+        return Err(EfiError::InvalidParameter);
+    }
+    let fw_vol = unsafe { fv_ptr.as_ref().expect("get_interface_for_handle returned NULL ptr for FirmwareVolume") };
```
kouchekiniad: Nit: `.expect("get_interface_for_handle returned NULL ptr for FirmwareVolume")` can be changed to `.unwrap()` now that we have an explicit check

patina_dxe_core/src/image.rs
kouchekiniad: @AnhDLuong FYI it looks like #788 went in

patina_dxe_core/src/image.rs
AnhDLuong: will fix it

patina_dxe_core/src/image.rs
AnhDLuong: @kouchekiniad , thanks for informing me. I did discuss with @Javagedes and he agreed to let me use efi::Guid in this PR.

patina_dxe_core/src/image.rs
AnhDLuong: Will bubble up the error

patina_dxe_core/src/image.rs
AnhDLuong: will check null ptr

patina_dxe_core/src/image.rs
AnhDLuong: thanks, will change to .unwrap()

patina_dxe_core/src/image.rs
```diff
@@ -788,6 +794,57 @@ fn get_buffer_by_file_path(
...
+    let (remaining_file_path, handle) = core_locate_device_path(firmware_volume::PROTOCOL_GUID, file_path)?;
+
+    // For FwVol File system there is only a single file name that is a GUID.
+    let fv_name_guid = get_file_guid_from_device_path(remaining_file_path)?;
+
+    // Get the firmware volume protocol
+    let fv_ptr =
+        PROTOCOL_DB.get_interface_for_handle(handle, firmware_volume::PROTOCOL_GUID)? as *mut firmware_volume::Protocol;
+    if fv_ptr.is_null() {
+        debug_assert!(fv_ptr.is_null(), "ERROR: get_interface_for_handle returned NULL ptr for FirmwareVolume!");
```
makubacki:
```suggestion
        debug_assert!(!fv_ptr.is_null(), "ERROR: get_interface_for_handle returned NULL ptr for FirmwareVolume!");
```

makubacki:
@AnhDLuong, is there a reason this PR is still in draft state? Also, please confirm in the "How This Was Tested" section of the PR description that you have tested this in patina-qemu.

AnhDLuong:
@makubacki , this PR is ready

PR #777, opened 2025-09-26
Increase stack space for AARCH64 synchronous exceptions.

joschock:
## Description

The current implementation for the aarch64 exception handler reserves 8K of stack for exceptions where the stack pointer may have been corrupted at the time of exception. In practice (especially when optimizations are disabled) this is not enough for all use cases, in particular when the debugger is active. 

This PR increases the space for the alternate stack to 64K. 

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Fixes debugger behavior on aarch64 platform used for test.

## Integration Instructions

N/A

### DIFF
--- a/core/patina_internal_cpu/src/interrupts/aarch64/exception_handler.asm
+++ b/core/patina_internal_cpu/src/interrupts/aarch64/exception_handler.asm
@@ -91,10 +91,10 @@
 
   .global sp_el0_end
 
-# Stack for SP_EL0 of 0x2000 bytes. Also set to 8KB aligned, which corresponds to BIT13.
+# Stack for SP_EL0 of 0x10000 bytes. Also set to 8KB aligned, which corresponds to BIT13.
   .align 13
 sp_el0_start:
-  .space 0x2000
+  .space 0x10000
 sp_el0_end:
 
   .section .text


core/patina_internal_cpu/src/interrupts/aarch64/exception_handler.asm
```diff
@@ -91,10 +91,10 @@
 
   .global sp_el0_end
 
-# Stack for SP_EL0 of 0x2000 bytes. Also set to 8KB aligned, which corresponds to BIT13.
+# Stack for SP_EL0 of 0x10000 bytes. Also set to 8KB aligned, which corresponds to BIT13.
   .align 13
 sp_el0_start:
-  .space 0x2000
```
cfernald, line 97: Does this get populated in the image size or is it covered in the expansion region? I'm wondering if its better to allocate a second stack later then add to the image size. I did a quick test, and it seems to actually increase the binary size.

joschock: I think it does increase the image size. No objection to a future better solution; but any concerns about this as a stopgap in the meantime? I can file a bug for future improvement for runtime stack allocation - but how that stack is allocated and by whom is a matter that I think deserves some architectural thought.

cfernald: Yeah, go ahead with this and please file a bug for us to address this

joschock: https://github.com/OpenDevicePartnership/patina/issues/781 filed.

PR #786, opened 2025-09-29
Handle potentially unaligned pointers from FFI

joschock:
## Description

The UEFI spec does not clearly require that pointers passed to UEFI spec APIs (like AllocatePages) must be aligned. 
This pull request changes various FFI routines to assume that raw pointers passed to them may not be aligned. 

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Unit tests pass. On systems passing unaligned pointers to core routines, no longer observe crashes. Note: this doesn't constitute broad coverage of these changes; this was only observed with a subset of the routines in the real world. 

## Integration Instructions

N/A

### DIFF
--- a/core/patina_internal_device_path/src/lib.rs
+++ b/core/patina_internal_device_path/src/lib.rs
@@ -26,7 +26,8 @@ use r_efi::efi;
 ///
 /// ## Safety
 ///
-/// device_path input must be a valid pointer to a well-formed device path.
+/// device_path input must be a valid pointer (i.e. not null) that points to
+/// a well-formed device path that conforms to UEFI spec 2.11 section 10.
 ///
 /// ## Examples
 ///
@@ -75,6 +76,8 @@ pub fn device_path_node_count(
         return Err(efi::Status::INVALID_PARAMETER);
     }
     loop {
+        // Safety: caller must guarantee that device_path is a valid pointer to
+        // a well-formed device path as described in the function documentation above.
         let current_node = unsafe { current_node_ptr.read_unaligned() };
         let current_length: usize = u16::from_le_bytes(current_node.length).into();
         node_count += 1;
--- a/patina_dxe_core/src/allocator.rs
+++ b/patina_dxe_core/src/allocator.rs
@@ -400,6 +400,7 @@ extern "efiapi" fn allocate_pool(pool_type: efi::MemoryType, size: usize, buffer
 
     match core_allocate_pool(pool_type, size) {
         Err(err) => err.into(),
+        // Safety: caller must ensure that buffer is a valid pointer. It is null-checked above.
         Ok(allocation) => unsafe {
             buffer.write_unaligned(allocation);
             efi::Status::SUCCESS
@@ -481,10 +482,12 @@ pub fn core_allocate_pages(
             let result = match allocation_type {
                 efi::ALLOCATE_ANY_PAGES => allocator.allocate_pages(DEFAULT_ALLOCATION_STRATEGY, pages, alignment),
                 efi::ALLOCATE_MAX_ADDRESS => {
+                    // Safety: caller must ensure that "memory" is a valid pointer. It is null-checked above.
                     let address = unsafe { memory.read_unaligned() };
                     allocator.allocate_pages(AllocationStrategy::TopDown(Some(address as usize)), pages, alignment)
                 }
                 efi::ALLOCATE_ADDRESS => {
+                    // Safety: caller must ensure that "memory" is a valid pointer. It is null-checked above.
                     let address = unsafe { memory.read_unaligned() };
                     allocator.allocate_pages(AllocationStrategy::Address(address as usize), pages, alignment)
                 }
@@ -492,6 +495,7 @@ pub fn core_allocate_pages(
             };
 
             if let Ok(ptr) = result {
+                // Safety: caller must ensure that "memory" is a valid pointer. It is null-checked above.
                 unsafe { memory.write_unaligned(ptr.cast::<u8>().as_ptr().expose_provenance() as u64) }
                 Ok(())
             } else {
@@ -709,13 +713,16 @@ extern "efiapi" fn get_memory_map(
     }
 
     if !descriptor_size.is_null() {
+        // Safety: caller must ensure that descriptor_size is a valid pointer if it is not null.
         unsafe { descriptor_size.write_unaligned(mem::size_of::<efi::MemoryDescriptor>()) };
     }
 
     if !descriptor_version.is_null() {
+        // Safety: caller must ensure that descriptor_version is a valid pointer if it is not null.
         unsafe { descriptor_version.write_unaligned(efi::MEMORY_DESCRIPTOR_VERSION) };
     }
 
+    // Safety: caller must ensure that memory_map_size is a valid pointer. It is null-checked above.
     let map_size = unsafe { memory_map_size.read_unaligned() };
 
     let efi_descriptors = match get_memory_map_descriptors(false) {
@@ -727,6 +734,7 @@ extern "efiapi" fn get_memory_map(
 
     let required_map_size = efi_descriptors.len() * mem::size_of::<efi::MemoryDescriptor>();
 
+    // Safety: caller must ensure that memory_map_size is a valid pointer. It is null-checked above.
     unsafe { memory_map_size.write_unaligned(required_map_size) };
 
     if map_size < required_map_size {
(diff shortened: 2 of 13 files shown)


patina_dxe_core/src/events.rs
```diff
@@ -126,19 +126,21 @@ extern "efiapi" fn wait_for_event(
...
-    //get the events list as a slice
-    let event_list = unsafe { core::slice::from_raw_parts(event_array, number_of_events) };
-
     //spin on the list
     loop {
-        for (index, event) in event_list.iter().enumerate() {
-            match check_event(*event) {
+        let mut event_ptr = event_array;
+        for index in 0..number_of_events {
+            let event = unsafe { event_ptr.read_unaligned() };
```
os-d, line 137: nit: safety comment (and anywhere else throughout)

patina_dxe_core/src/events.rs
```diff
@@ -126,19 +126,21 @@ extern "efiapi" fn wait_for_event(
...
                 efi::Status::NOT_READY => (),
                 status => {
-                    unsafe { *out_index = index };
+                    unsafe {
+                        out_index.write_unaligned(index);
+                    };
                     return status;
                 }
             }
+            event_ptr = unsafe { event_ptr.add(1) };
```
os-d, line 149: Your changes look the same as what was there, but this does look like a difference in behavior from edk2?

Two things, one is at the end of the outer spin loop, we don't signal the idle event, does that matter? https://github.com/tianocore/edk2/blob/38370cf4924205df97b8bcaf0f184983a4b34f66/MdeModulePkg/Core/Dxe/Event/Event.c#L697

More importantly, shouldn't we check if `out_index` is null before dereferencing? https://github.com/tianocore/edk2/blob/38370cf4924205df97b8bcaf0f184983a4b34f66/MdeModulePkg/Core/Dxe/Event/Event.c#L686

It would be embarrassing if C could catch a null pointer and Rust couldn't ;).

As I noted, I think you preserved the behavior, so these changes could come separately, if needed.

patina_dxe_core/src/protocols.rs
```diff
@@ -309,7 +311,13 @@ extern "efiapi" fn locate_handle(
             }
 
             //copy handle list into output buffer
-            unsafe { slice::from_raw_parts_mut(handle_buffer, list.len()).copy_from_slice(&list) };
+            unsafe {
+                core::ptr::copy(
```
os-d, line 318: I think copy can't be used, right?

https://doc.rust-lang.org/std/ptr/fn.copy.html#safety

Image

So maybe `write_unaligned()` has to be used?

patina_dxe_core/src/protocols.rs
```diff
@@ -464,9 +476,13 @@ extern "efiapi" fn open_protocol_information(
     match core_allocate_pool(efi::BOOT_SERVICES_DATA, buffer_size) {
         Err(err) => err.into(),
         Ok(allocation) => unsafe {
-            entry_buffer.write(allocation as *mut efi::OpenProtocolInformationEntry);
-            *entry_count = open_info.len();
-            slice::from_raw_parts_mut(*entry_buffer, open_info.len()).copy_from_slice(&open_info);
+            entry_buffer.write_unaligned(allocation as *mut efi::OpenProtocolInformationEntry);
+            entry_count.write_unaligned(open_info.len());
+            core::ptr::copy(
```
os-d, line 492: Same on copy

patina_dxe_core/src/protocols.rs
joschock: I think it is okay due to the cast to *u8, which makes the args alignment=1

os-d: Makes sense, missed that.

patina_dxe_core/src/events.rs
joschock: For the first one, agree, that looks like a miss. I will file an issue to fix it separately before I close this PR, but since it is off-topic, would prefer to do separately. On the second one, out_index is null-checked at the start of the function, so I think that's covered. 
```rust
    if number_of_events == 0 || event_array.is_null() || out_index.is_null() {
    }
```

patina_dxe_core/src/protocols.rs
joschock: same - cast to *u8 makes align=1

patina_dxe_core/src/events.rs
joschock: will take a safety comment pass thru before completing PR.

patina_dxe_core/src/events.rs
os-d: Totally fine to do separately.

For the out_index one, I don't have a preference, maybe this is the right way to do it, but it is a behavior change from edk2, which would still return success here (just not change the null index pointer). The UEFI spec doesn't envision returning EFI_INVALID_PARAMETER here, so maybe worth also changing.

joschock: Makes sense to operate on out_index as an optional pointer. I'll include both in the issue and make a PR to change the behavior separately.

patina_dxe_core/src/events.rs
joschock: safety comments: activated.

patina_dxe_core/src/events.rs
joschock: filed https://github.com/OpenDevicePartnership/patina/issues/791

core/patina_internal_device_path/src/lib.rs
```diff
@@ -75,7 +75,8 @@ pub fn device_path_node_count(
         return Err(efi::Status::INVALID_PARAMETER);
     }
     loop {
-        let current_node = unsafe { *current_node_ptr };
+        // Safety: caller must guarantee that device_path is a valid pointer to a well-formed device path.
```
makubacki: Since this is a caller requirement, it seems like it should be added to the function doc comment safety section. Maybe the comment here then references that requirement being upheld by the caller.

patina_dxe_core/src/allocator.rs
```diff
@@ -481,18 +482,21 @@ pub fn core_allocate_pages(
...
+                    let address = unsafe { memory.read_unaligned() };
+                    allocator.allocate_pages(AllocationStrategy::Address(address as usize), pages, alignment)
                 }
                 _ => Err(EfiError::InvalidParameter),
             };
 
             if let Ok(ptr) = result {
-                unsafe { memory.write(ptr.cast::<u8>().as_ptr().expose_provenance() as u64) }
+                // Safety: caller must ensure that "memory" is a valid pointer. It is null-checked above.
+                unsafe { memory.write_unaligned(ptr.cast::<u8>().as_ptr().expose_provenance() as u64) }
```
makubacki, line 499: How often are unaligned addresses being passed for page allocation?

core/patina_internal_device_path/src/lib.rs
joschock: there is a doc safety comment in the function header already: 
```
/// ## Safety
///
/// device_path input must be a valid pointer to a well-formed device path.
```
did you mean changing this comment to somehow reference that?

makubacki: As we've discussed in the past, these should be `unsafe` functions with clear caller requirements. All of that doesn't need to be done in this PR or anything, but I think placing the safety requirements together that are present is at least a step toward that for now.

patina_dxe_core/src/allocator.rs
joschock: So far the only examples I've seen in the wild of unaligned pointers are for gBS->AllocatePool(), but I think it's just as likely for AllocatePages. 

Note that this is the alignment of the pointer to receive the allocated address or communicate the limit (as opposed to the pointer contents), which is generally going to be naturally aligned by most compilers if it's just a local variable or something - so I would expect it to be rare.  

The main case I can think of in C is if the pointer is part of a packed struct and isn't aligned on (in this case) an 8-byte boundary within that struct e.g:
`gBS->AllocatePages(AllocateAnyPages, EfiBootServicesData, 0x100, &PackedStruct.UnalignedPointerField);`

core/patina_internal_device_path/src/lib.rs
joschock: Agree - but I'm not following what change you are suggesting for the current PR?
do you mean something like:
`// Safety: caller must guarantee that device_path is a valid pointer to a well-formed device path as specified in the function requirements.`?

makubacki: Yes, but also to clarify what "valid pointer" means in the context of this function's requirements.

patina_dxe_core/src/allocator.rs
makubacki: Makes sense. I also realized after posting that this was for the location of storing the allocated address, so the unaligned address is not trickling further down into anything. As you said, that would seem very rare, but not too concerning from an overall safety perspective here.

core/patina_internal_device_path/src/lib.rs
joschock: ok, clarified both the comment here and the doc comment with a reference to spec definition for device path to clarify what "well-formed" means. Valid just means "not bogus" - but clarified that that includes non-null.

PR #790, opened 2025-09-29
[REBASE & FF] Consume patina-paging v9

os-d:
## Description

patina-paging release v9 which is a breaking change. It created a more permissive API in that crate, requiring consumers to drive policy decisions.

This updates patina to query memory regions and decide if the GCD and page table state is aligned.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested booting to Windows on Q35, SBSA, and a physical Intel platform. Testing on a physical ARM64 platform to be done soon.

## Integration Instructions

N/A.

### DIFF
--- a/core/patina_debugger/src/debugger.rs
+++ b/core/patina_debugger/src/debugger.rs
@@ -297,7 +297,6 @@ impl<T: SerialIO> Debugger for PatinaDebugger<T> {
                 }
                 else {
                     internal.gdb_buffer = unsafe { Some(&*(GDB_BUFFER.as_ptr() as *mut [u8; GDB_BUFF_LEN])) };
-                    internal.monitor_buffer = unsafe { Some(&*(MONITOR_BUFFER.as_ptr() as *mut [u8; MONITOR_BUFF_LEN])) };
                 }
             }
         }


os-d:
This depends on this PR completing first: https://github.com/OpenDevicePartnership/patina-paging/pull/120

patina_dxe_core/src/gcd/spin_locked_gcd.rs
```diff
@@ -1873,9 +1873,55 @@ impl SpinLockedGcd {
...
+                    // it is not an error if the range is fully not mapped, we just need to map it, unless we are
+                    // trying to unmap the region, which we will check for below
+                    unmapped = true;
+                }
+                Err(e) => {
+                    log::error!(
+                        "query memory region {base_address:#x?} of length {len:#x?} with attributes {attributes:#x?}. Status: {e:#x?}",
+                    );
+                    log::error!("GCD and page table are out of sync. This is a critical error.");
+                    log::error!("GCD {GCD}");
```
cfernald: Maybe error is too high for printing the GCD. If a configuration has error printing on but not debug_assert this could cause a lot of prints to happen if these got out of sync. Consider making this a trace or info.

patina_dxe_core/src/gcd/spin_locked_gcd.rs
```diff
@@ -1873,9 +1873,55 @@ impl SpinLockedGcd {
...
+                    "Memory region {base_address:#x?} of length {len:#x?} with attributes {attributes:#x?}. No paging action taken: Region already mapped with these attributes.",
+                );
+                return Ok(());
+            }
+
             // EFI_MEMORY_RP is a special case, we don't actually want to set it in the page table, we want to unmap
             // the region
             if paging_attrs & MemoryAttributes::ReadProtect == MemoryAttributes::ReadProtect {
+                if unmapped {
+                    // the region is already unmapped, this indicates that the GCD and page table are out of sync
```
cfernald, line 1918: If this indicates we are out of sync, should this not also be a debug assert? Or an error/warning log?

patina_dxe_core/src/gcd/spin_locked_gcd.rs
```diff
@@ -2446,11 +2445,33 @@ impl SpinLockedGcd {
...
+                            .set_memory_space_attributes(
+                                current_base as usize,
+                                current_len as usize,
+                                descriptor.attributes,
+                            )
+                            .is_err()
+                        {
+                            // well, we did our best. The GCD and page table are now out of sync, which is a critical error.
+                            log::error!(
+                                "Failed to roll back GCD attributes after page table attribute set failure. This is a critical error. GCD and page table are now out of sync."
```
cfernald: should probably log the error that occurred.

patina_dxe_core/src/gcd/spin_locked_gcd.rs
os-d: Yeah, that is true. At the same time they will probably hit a page fault before they get too far :). I'll move this to info. It is very helpful to have in the log to debug with (especially in the next PR that adds the pt dump also).

patina_dxe_core/src/gcd/spin_locked_gcd.rs
os-d: I think so. Let me test that. I thought I had it that way and then maybe ran into an error.

PR #792, opened 2025-09-29
docs/rust_tools: Add Patina tool overview

makubacki:
## Description

Adds a document to the Patina background section to explain the various tools used in Patina and why they are used. To help readers better relate these tools to what is commonly used in C firmware development, some comparisons are made throughout the document.

The goal of this document is to explain key tools used in Patina, the unifying interface behind the commands using cargo, and how these align under a single Patina tool philosophy.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [x] Includes documentation?

## How This Was Tested

- `cargo make all`
- Local markdown lint check

## Integration Instructions

- N/A

### DIFF
--- a/docs/src/background/rust_tools.md
+++ b/docs/src/background/rust_tools.md
@@ -484,11 +484,11 @@ args = ["clippy", "--all-targets", "--all-features", "--", "-D", "warnings"]
 
 - [Clippy Lints](https://rust-lang.github.io/rust-clippy/master/index.html)
 
-### cargo miri (Undefined Behavior Detection for Unsafe Code)
+### cargo miri (Undefined Behavior Detection)
 
 **Purpose**: Executes tests under the [Miri](https://github.com/rust-lang/miri) interpreter to detect memory safety
-issues resulting in undefined behavior (UB) in `unsafe` code. For example, out-of-bounds pointer arithmetic,
-use-after-free, invalid use of uninitialized data, alignment violations, etc.
+issues resulting in undefined behavior (UB). This is particularly useful in `unsafe` code. For example, out-of-bounds
+pointer arithmetic, use-after-free, invalid use of uninitialized data, alignment violations, etc.
 
 In Patina, the goal is to write as little `unsafe` code as possible, but some low-level operations require it. Patina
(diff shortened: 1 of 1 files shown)


makubacki:
Note: This is part 2 of 3 parts planned for the Patina Overview/Background section. The final part will be one focused on platform firmware integration/philosophy in Patina and the impact of that on current platform integration.

docs/src/background/rust_tools.md
```diff
@@ -0,0 +1,619 @@
...
+/// # Examples
+/// ```
+/// let mut buffer = vec![1, 2, 3, 4];
+/// process_buffer(&mut buffer);
+/// assert_eq!(buffer, vec![0, 0, 0, 0]);  // This example is automatically tested!
+/// ```
+///
+/// # Safety
+/// This function is memory-safe due to Rust's slice bounds checking.
+pub fn process_buffer(buffer: &mut [u8]) {
```
os-d, line 311: small nit: May be nice to show the same function implemented in C vs Rust and what comments are like, for more direct comparison.

docs/src/background/rust_tools.md
```diff
@@ -0,0 +1,619 @@
...
+[tasks.clippy]
+command = "cargo"
+args = ["clippy", "--all-targets", "--all-features", "--", "-D", "warnings"]
+```
+
+**Further Documentation**:
+
+- [Clippy Lints](https://rust-lang.github.io/rust-clippy/master/index.html)
+
+### cargo miri (Undefined Behavior Detection for Unsafe Code)
```
os-d: Miri technically works on more than just unsafe code, e.g. it can tell you about memory leaks. Unsafe evaluation is definitely the major benefit, but may be worth reframing this section as undefined behavior detection that extends what the Rust compiler is able to check for or that is only able to be caught dynamically.

makubacki: Clarified in the latest push.

docs/src/background/rust_tools.md
makubacki: Updated

PR #794, opened 2025-09-30
x86_64 cpu_init: Initialize / Configure FPU

Javagedes:
## Description

Initializes / configures the FPU by setting the x87 FPU Control Word and MXCSR Control Status Register to the required state per UEFI SCTs and existing C implementation.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Both SCTs now show the expected values for these two configurations, however one test is still failing as the precision sticky flag is being set to 1 somewhere in the BDS entry point before entering the Shell. This indicates that some type of floating-point math was performed where the result could not be accurately described by a given number and rounding occurred (example: `1.0 / 3.0`). This still needs to be further investigated, but the general configurations ignoring this bit is accurate (ref: #796)

## Integration Instructions

N/A

### DIFF
--- a/core/patina_internal_cpu/src/cpu/x64/cpu.rs
+++ b/core/patina_internal_cpu/src/cpu/x64/cpu.rs
@@ -116,7 +116,7 @@ impl EfiCpuX64 {
                 temp = out(reg) _,
                 FPU_CONTROL_WORD = sym FPU_CONTROL_WORD,
                 MMX_CONTROL_WORD = sym MMX_CONTROL_WORD,
-                BIT9 = const 1 << 9,
+                BIT9 = const patina_sdk::bit!(9),
                 options(nostack, preserves_flags)
             );
         }


core/patina_internal_cpu/src/cpu/x64/cpu.rs
```diff
@@ -91,6 +92,35 @@ impl EfiCpuX64 {
...
+                // Set OSFXSR (bit 9) in CR4 to enable SSE instructions
+                "mov {temp}, cr4",
+                "or {temp}, {BIT9}",
+                "mov cr4, {temp}",
+
+                "ldmxcsr [{MMX_CONTROL_WORD}]",
+                temp = out(reg) _,
+                FPU_CONTROL_WORD = sym FPU_CONTROL_WORD,
+                MMX_CONTROL_WORD = sym MMX_CONTROL_WORD,
+                BIT9 = const 1 << 9,
```
os-d: I'm fine with this, but we should add the BIT definitions to patina_sdk

cspell.yml
os-d: The additions here shouldn't be required with this ignore list, probably the regex is incorrect. Do you want to fix that up instead?

cspell.yml
```diff
@@ -51,6 +51,8 @@ words:
   - efiapi
   - efierror
   - eflags
+  - fldcw
+  - finit
```
cfernald: Clearly my regex above for catching asm isn't working. Might be worth trying to fix that so this file doesnt just get filled with asm commands.

cfernald: Oh, just saw Oliver's comment

cspell.yml
Javagedes: will investigate

cspell.yml
Javagedes: I will investigate the regex :)

Javagedes:
> I'm a bit hesitant to check this in before we get the SCT passing, since that is the only reason we are doing this. In theory this doesn't affect the place where you are seeing the failure, but still only seeing it in patina is concerning.

Sure that is fine. I can wait to merge this until I determine who is causing the precision exception sticky flag. I was mainly wanting a review on the assembly regarding this PR :)

cspell.yml
Javagedes: resolved in #802

Javagedes:
> > I'm a bit hesitant to check this in before we get the SCT passing, since that is the only reason we are doing this. In theory this doesn't affect the place where you are seeing the failure, but still only seeing it in patina is concerning.
> 
> Sure that is fine. I can wait to merge this until I determine who is causing the precision exception sticky flag. I was mainly wanting a review on the assembly regarding this PR :)

From offline conversation we will merge this, and I will continue to investigate on my own.

It is fairly low priority because we are not trapping this exception, and cannot even purposefully trap if we try, so it should not directly affect a platform, rather than the single SCT not currently passing.

core/patina_internal_cpu/src/cpu/x64/cpu.rs
Javagedes: https://github.com/OpenDevicePartnership/patina/pull/803

os-d:
> > > I'm a bit hesitant to check this in before we get the SCT passing, since that is the only reason we are doing this. In theory this doesn't affect the place where you are seeing the failure, but still only seeing it in patina is concerning.
> > 
> > 
> > Sure that is fine. I can wait to merge this until I determine who is causing the precision exception sticky flag. I was mainly wanting a review on the assembly regarding this PR :)
> 
> From offline conversation we will merge this, and I will continue to investigate on my own.
> 
> It is fairly low priority because we are not trapping this exception, and cannot even purposefully trap if we try, so it should not directly affect a platform, rather than the single SCT not currently passing.

tbh, I think we should just close the associated issues as won't fix. This is not a patina issue and does not affect anything meaningfully. We will continue to periodically run the SCTs and can re-evaluate failures at that time.

Javagedes:
> > > > I'm a bit hesitant to check this in before we get the SCT passing, since that is the only reason we are doing this. In theory this doesn't affect the place where you are seeing the failure, but still only seeing it in patina is concerning.
> > > 
> > > 
> > > Sure that is fine. I can wait to merge this until I determine who is causing the precision exception sticky flag. I was mainly wanting a review on the assembly regarding this PR :)
> > 
> > 
> > From offline conversation we will merge this, and I will continue to investigate on my own.
> > It is fairly low priority because we are not trapping this exception, and cannot even purposefully trap if we try, so it should not directly affect a platform, rather than the single SCT not currently passing.
> 
> tbh, I think we should just close the associated issues as won't fix. This is not a patina issue and does not affect anything meaningfully. We will continue to periodically run the SCTs and can re-evaluate failures at that time.

I'm definitely on board with that :) Will close it once this goes in.

PR #797, opened 2025-09-30
Return Handles in Creation order; adjust verbosity for protocol install/uninstall messages. [Rebase & FF]

joschock:
## Description

This PR makes two commits to change the protocol DB implementation:

Handle Database behavior updates
* Add support for tracking creation time for handles so that they are returned in creation order for query functions like `LocateHandleBuffer`
* (minor) `registered_protocols` updated to return only unique protocols (no duplicates)

Handle Database message verbosity
* (minor) Adjust verbosity of install/uninstall protocol interface messages to align with traditional verbosity in EDK2 core. 

Returning handles in creation order is not a UEFI spec requirement, but returning them in creation order is behavior that a number of subsystems (even within patina) rely on for reasonable behavior. An example within patina is if multiple FVs are defined via HOBs at the start of DXE, they should be dispatched in the order that they are reported to the core. Without creation ordering on the handle query for new FVs, the dispatcher will discover and process the FVs in arbitrary order. 

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Passes unit tests, executes on hardware platform without observed issues.

## Integration Instructions

N/A

### DIFF
--- a/patina_dxe_core/src/protocol_db.rs
+++ b/patina_dxe_core/src/protocol_db.rs
@@ -144,10 +144,53 @@ pub struct ProtocolNotify {
     fresh_handles: BTreeSet<efi::Handle>,
 }
 
+struct Handle {
+    order: usize,
+    protocols: BTreeMap<OrdGuid, ProtocolInstance>,
+}
+
+impl Handle {
+    fn new(order: usize) -> Self {
+        Handle { order, protocols: BTreeMap::new() }
+    }
+
+    fn keys(&self) -> impl Iterator<Item = &OrdGuid> {
+        self.protocols.keys()
+    }
+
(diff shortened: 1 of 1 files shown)


makubacki:
Even though this isn't spec-defined behavior, I added the "type:design-change" label to help call this change out in the release notes.

patina_dxe_core/src/protocol_db.rs
makubacki, line 1: Would it be possible to add a unit test for this behavior?

joschock: sure, good idea. will do.

joschock: Added in latest push.

Issue #749, opened 2025-09-17
[Documentation]: Justify each unstable feature in Patina docs


makubacki:
### Request Description

The Patina policy for using unstable features is defined in [Use of Unstable Rust Features in Patina](https://github.com/OpenDevicePartnership/patina/blob/1baec125c96e8376b4f3fae87d8239201227e608/docs/src/dev/unstable.md).

Consistent with that policy, Patina uses unstable Rust features. This issue tracks clearly enumerating each unstable feature with an accompanying justification so Patina consumers can understand the rationale for using the feature.

One GitHub tracking issue should be filed per unstable feature currently in use.

### Are you going to make the change?

Someone else needs to make the change

### Do you need maintainer feedback?

No maintainer feedback needed

Javagedes:
I've updated the process as documented in https://opendevicepartnership.github.io/patina/dev/unstable.html, which has also created a issue template for creating a tracker for all used and actively proposed rust feature gates. With this, each unstable feature currently being used in this repository now has a tracking issue created for it with the `rustc-feature-gate` label attached.

Closing.

PR #800, opened 2025-10-01
Add rustc feature gate tracking issue requirements

Javagedes:
## Description

Updates the documentation to clarify the process and requirements for introducing and tracking unstable rustc feature usage in the Patina project. Specifically, this includes the need for an active and maintained tracking issue for each unstable feature used. This tracking issue should remain open until the feature usage is rejected, stabilized, or removed from the codebase.

The tracking issue, specifically marked with the label `rustc-feature-gate`, allows maintainers to easily monitor and manage the lifecycle of unstable rustc features within the project. It also provides a single place for tracking justification and usage of such features.

Having an overall tracking issue for each unstable feature was requested via #749 but it does not close it.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [x] Includes documentation?

## How This Was Tested

CI and validated the issue template renders properly.

## Integration Instructions

The `rustc-feature-gate` label needs to be created.

### DIFF
--- a/docs/src/dev/unstable.md
+++ b/docs/src/dev/unstable.md
@@ -10,7 +10,8 @@ and unstable state, such as: being largely no-std, implementing and using its ow
 and unsafe operations. Below are the guiding principles and practices for working with unstable Rust features in
 Patina.
 
-You can always track all active unstable rustc [feature usage](https://github.com/OpenDevicePartnership/patina/issues?q=is%3Aissue%20state%3Aopen%20label%3Astate%3Arustc-feature-gate).
+All active rustc feature usage is tracked with the [`state:rustc-feature-gate`](https://github.com/OpenDevicePartnership/patina/issues?q=is%3Aissue%20state%3Aopen%20label%3Astate%3Arustc-feature-gate)
+label.
 
 ## When Unstable Rust Features May Be Used
 
--- a/.github/ISSUE_TEMPLATE/rustc_feature_gate.yml
+++ b/.github/ISSUE_TEMPLATE/rustc_feature_gate.yml
@@ -25,6 +25,14 @@ body:
       placeholder: "generic_associated_types"
     validations:
       required: true
+  - type: input
+    id: feature_link
+    attributes:
+      label: Unstable rustc feature issue link
+      description: The link to the unstable rust feature tracking issue (github).
+      placeholder: https://github.com/rust-lang/rfcs/pull/1598
+    validations:
+      required: true
   - type: checkboxes
     id: status
     attributes:


docs/src/dev/unstable.md
```diff
@@ -33,31 +35,74 @@ perform the following tasks to mitigate risk:
...
+- **Risks:** Any special risks the project may incur due to use of this feature.
+
+Once created, the community will review and comment on the RFC, following the RFC process. The RFC will eventually
+either be accepted or denied and then merged.
+
+### Create and Merge Implementation
+
+If the RFC is merged, it is now your responsibility to implement it's usage and create a pull-request for review. The
+pull-request should reference that this is the implementation for your accepted RFC. Once merged, **you must continue**
+**to monitor** the stability of the feature. Notably, if any changes to the feature's API occur, you will be responsible
```
os-d, line 91: Rust has stabilization issues for each feature, right? Can we use those to link against our issues, also? Is there any way to have those alert us when the Rust side stabilization feature is closed (or otherwise status changed I guess)? May not be worth it, but the linking would be nice.

Javagedes: Yes this makes sense. I'll add it as a required field in the template and mention it in the markdown doc!

docs/src/dev/unstable.md
```diff
@@ -33,31 +35,74 @@ perform the following tasks to mitigate risk:
...
+for adjusting the usage of the feature in the codebase.
+
+### Stabilization
+
+The final step of the process is stabilization (or removal, either from our codebase or from rustc) of the feature. If
+the feature is stabilized, then this step is short. The `#![feature(...)]` must be removed on the next rustc version
+update in this project that includes the feature stabilization. If any final changes need to be made, they will be made
+here.
+
+```admonish warning
```
apop5, line 101: stabilization can be tied to a rust tool chain update.

Is there a process in place for reexamining the unstable features for each tool chain update?

docs/src/dev/unstable.md
```diff
@@ -10,6 +10,8 @@ and unstable state, such as: being largely no-std, implementing and using its ow
 and unsafe operations. Below are the guiding principles and practices for working with unstable Rust features in
 Patina.
 
+You can always track all active unstable rustc [feature usage](https://github.com/OpenDevicePartnership/patina/issues?q=is%3Aissue%20state%3Aopen%20label%3Astate%3Arustc-feature-gate).
```
makubacki: I think this is clearer. It also calls out the grouping label for future reference.

```suggestion
All active rustc feature usage is tracked with the [`state:rustc-feature-gate`](https://github.com/OpenDevicePartnership/patina/issues?q=is%3Aissue%20state%3Aopen%20label%3Astate%3Arustc-feature-gate) label.
```

docs/src/dev/unstable.md
makubacki: We could run an action (e.g. on a schedule) that gathers the active stabilization issues from our open issues and adds a comment to it if a major action occurs in the upstream issue, like it is closed. But I think just linking to it is fine for now.

docs/src/dev/unstable.md
Javagedes: We will actually fail to compile if we update to a toolchain where a feature has been stabilized or removed. If the API has changed then we will probably fail to compile there too, but that's iffier, depending on the change.

docs/src/dev/unstable.md
Javagedes: Added a required section to link the rust-lang stabilization issue. Good idea.

@makubacki do you want me to add a backlog item for your comment?

Closing this, but feel free to comment / reopen.

makubacki: > @makubacki do you want me to add a backlog item for your comment?

@Javagedes, I don't think it's necessary at this time. We can put more scaffolding in place depending on how things go.

PR #803, opened 2025-10-01
provide a bit! macro to match make setting a particular bit easy.

Javagedes:
## Description

We unfortunately cannot directly match C's `#define` functionality as it is type-less, and const definitions in rust require a type. To resolve this, we instead make a macro that does the left shifting for you, allowing you to specify the underlying binary type (u32, i32, etc).

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

CI

## Integration Instructions

N/A

### DIFF
--- a/sdk/patina_sdk/src/base.rs
+++ b/sdk/patina_sdk/src/base.rs
@@ -11,7 +11,7 @@
 use num_traits;
 use r_efi::efi;
 
-use crate::{bit, error::EfiError};
+use crate::error::EfiError;
 
 pub mod guid;
 


sdk/patina_sdk/src/base.rs
```diff
@@ -27,125 +27,133 @@ pub const UEFI_PAGE_MASK: usize = UEFI_PAGE_SIZE - 1;
 pub const UEFI_PAGE_SHIFT: usize = 12;
 
 /// 1KB, 1024 bytes, 0x400, 2^10
-pub const SIZE_1KB: usize = 0x400;
+pub const SIZE_1KB: usize = bit!(10);
```
os-d: I would be inclined to leave these definitions as is. This is the same, but I think it is confusing, when using bit!() you are trying to get a mask, typically, for these definitions we just want a size, nothing to do with a mask.

makubacki: Yeah. In this example, `0x400` is visually familiar/obvious as 1KB. The bit macro introduces some mental shifting that I don't think is needed.

sdk/patina_sdk/src/base.rs
```diff
@@ -27,125 +27,133 @@ pub const UEFI_PAGE_MASK: usize = UEFI_PAGE_SIZE - 1;
...
+pub const SIZE_128TB: usize = bit!(47);
 
 /// 256TB, 0x1000000000000, 2^48
-pub const SIZE_256TB: usize = 0x1000000000000;
+pub const SIZE_256TB: usize = bit!(48);
 
 /// Patina uses write back as the default cache attribute for memory allocations.
 pub const DEFAULT_CACHE_ATTR: u64 = efi::MEMORY_WB;
 
+/// A macro to generate a bit mask with the nth bit set.
```
makubacki, line 149: Some additional guidance probably wouldn't hurt.

```suggestion
/// A macro to generate a bit mask with the nth bit set.
///
/// This macro should generally be used to simplify bit references in
/// in masking operations where bit position is significant.
```

sdk/patina_sdk/src/base.rs
Javagedes: Sounds good, I will change the consts back to their original values!

PR #810, opened 2025-10-02
RFC: Remove Atomics

joschock:
## Description

This PR presents an RFC for the removal of atomics from Patina.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

N/A

## Integration Instructions

N/A

### DIFF
--- a/docs/src/rfc/text/0018-remove-atomics.md
+++ b/docs/src/rfc/text/0018-remove-atomics.md
@@ -1,0 +1,236 @@
+# RFC: `Remove Atomic from Patina`
+
+This RFC proposes removing the use of `Atomic` operations from Patina in favor of other forms of mutual exclusion.
+
+## Change Log
+
+Initial Revision.
+
+- 2025-10-01: Initial RFC created.
+- 2025-10-15: Updated with feedback from PR, moved to FCP.
+- 2025-10-23: Updated with learnings from initial draft implementation.
+
+## Motivation
+
+Presently `core::sync::atomic` module types are used in several locations in Patina to allow for thread/interrupt-safe
+internal mutability (often to satisfy the rust compiler more than to actually provide additional safety). While these
+primitives provide a relatively simple approach to managing concurrency within Patina, they have two significant
+drawbacks:
+
+1. **Compatibility** - Atomics require the use of special processor instructions. Not all architectures support these
+instructions, or may have issues with them (especially for early-in-development silicon). Use of atomics limits the
+potential portability of Patina.
+2. **Performance** - Executing atomic instructions typically has a performance impact. In a single-cpu, interrupt-only
+model such as UEFI mutual exclusion can be accomplished via interrupt disable which may have less of a performance
+impact in the UEFI context.
+
+The following table gives a general sense of the impact of removing atomics from the Red-
+Black Tree implementation that is used as the backing collection for the GCD. These were collected using `cargo make
+bench -p patina_internal_collections` on relatively recent aarch64 and x64 hardware:
+
+| Operation         | Architecture | Speed Improvement |
+| :---------------- | :----------: | :---------------: |
+| add/rbt/32bit     | aarch64      | 49%               |
+| add/rbt/32bit     | x64          | 35%               |
+| add/rbt/128bit    | aarch64      | 49%               |
+| add/rbt/128bit    | x64          | 30%               |
+| add/rbt/384bit    | aarch64      | 48%               |
+| add/rbt/384bit    | x64          | 32%               |
+| delete/rbt/32bit  | aarch64      | 50%               |
+| delete/rbt/32bit  | x64          | 34%               |
+| delete/rbt/128bit | aarch64      | 58%               |
+| delete/rbt/128bit | x64          | 31%               |
+| delete/rbt/384bit | aarch64      | 51%               |
+| delete/rbt/384bit | x64          | 27%               |
+| search/rbt/32bit  | aarch64      | 8%                |
+| search/rbt/32bit  | x64          | 10%               |
+| search/rbt/128bit | aarch64      | 5%                |
+| search/rbt/128bit | x64          | 8%                |
+| search/rbt/384bit | aarch64      | 6%                |
+| search/rbt/384bit | x64          | 3%                |
+
+While these performance benchmarks are narrow and somewhat synthetic, they do illustrate a material performance
+improvement from removal of atomics.
+
+## Technology Background
+
+The general topic of concurrency and the use of atomic operations therein is a large one. A simple primer is available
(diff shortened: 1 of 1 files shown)


docs/src/rfc/text/0018-remove-atomics.md
```diff
@@ -0,0 +1,178 @@
+# RFC: `Remove Atomic from Patina`
```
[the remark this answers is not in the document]
joschock: Intentional

joschock:
> I'm hesitant to say completely that atomics are not allowed. Maybe if it was scoped to the core, but components (as more easily overridden by a platform) are allowed to use them, perhaps. In general, I think the advice should be do not use them unless you have a very good reason. For sure the way we were using them was inappropriate. An attempt to escape unsafe in a non-performant way.

Yeah, I decided to open with "no atomics" as a start position, but I would not be hard-opposed to an escape clause for components.

docs/src/rfc/text/0018-remove-atomics.md
```diff
@@ -0,0 +1,180 @@
...
+
+  CoreRestoreTpl (Tpl);
+}
+```
+
+## Alternatives
+
+- Why is this design the best in the space of possible designs?
+
+The status quo of using atomics throughout the core has the drawbacks of lack of portability and performance impact as
```
makubacki, line 153: @joschock is their quantifiable performance data that can be referenced to support the perf claims? I'm not challenging them or anything, I think it'd just make the RFC more complete and have a more compelling argument.

Javagedes: I actually ran my benchmarks on `patina_internal_collections` and saw the following performance improvements (but I do not have real-world platform performance improvement data)

image
image
image

joschock: Thanks @Javagedes  - @makubacki would Joey's numbers suffice? I suspect the real-world impact is going to be quite variable across architectures depending on how atomics are implemented in each arch.

makubacki: I think those numbers are fine, but to your point, having them for both supported architectures would help complete the data.

docs/src/rfc/text/0018-remove-atomics.md
```diff
@@ -0,0 +1,180 @@
...
+If this concurrency protection is in fact required, this global state can be protected with interrupt-based concurrency
+protection (such as `tpl_mutex` in the sdk).
+
+In addition to the above, there is a large amount of test code that uses atomics. Modifications of test code are not in
+view since the primary drawbacks being addressed in this RFC (portability and performance) largely don't apply to unit
+tests executing in the build environment.
+
+It is possible that introduction of a new concurrency primitive based solely on interrupt manipulation
+(independently of TPL) to live alongside `tpl_lock` may be beneficial in addressing some of the above usages of
+atomics. Desirability of this potential approach will be determined as part of implementation of the RFC.
```
cfernald: I think that this would be beneficial. As it stands the TPL lock services two purposes, 1. locking the specific resource in question and 2. locking other resources similarly guarded by the TPL. I believe in many of the use cases where "we just need a static" it would be beneficial to have a lock that uses interrupts and a lock specific volatile. This way we do not correlate locks that are fundamentally independent.

In fact, I would take this a bit further to say this should be our primary locking mechanism, and say that unless TPL is actually needed, we should use an instance specific volatile marker protected by interrupt disablement.

joschock: I agree in general, at least within the core. However, there are two different usages of the TPL lock in the core today: TPL_HIGH (which essentially kills interrupts for the whole critical section - basically as long as the `TplGuard` hasn't been dropped) - and TPL_NOTIFY - which allows interrupts to occur, but only services handlers that are at TPL_HIGH. For most well-behaved code outside the core that doesn't run at TPL_HIGH, this is essentially the same thing, but there could be subtle differences. Do you think we need to mirror this with two different interrupt-based locks (i.e. the difference being whether interrupts are disabled for the whole critical section)?

cfernald: I agree, there is a difference between guarding in prevention of re-entrant behavior. That is, for some functionality like GCD manipulation we probably want to prevent interrupts because it will impact core functionality interrupts may need.

e.g. `InterruptLock` could suspend interrupts (if not already disabled) for the duration of the Guard lifetime tracking acquired with a volatile bool to protection from reentrance or exception handlers, but `ReentrantLock` will merely use a volatile (perhaps with temporary int disable) to protect against reentrancy for a given resource. So Reentrant lock would be the same as atomic mutexes, just without the atomic.

I just want to avoid the concepts of TPLs internally for relatively random static state since we may nest dependencies and TPL can be sensitive to nested state transitions. Or the other way, lock the resource and not the system.

Though thinking about it, for the interrupt case, it may be appropriate for those to stay TPL since that is more or less the same problem the "global" locking of TPL is supposed to solve.

makubacki:
@joschock, just a reminder to please keep the [Synchronization ](https://github.com/OpenDevicePartnership/patina/blob/main/docs/src/dxe_core/synchronization.md) doc updated as things progress.

joschock:
> I'm hesitant to say completely that atomics are not allowed. Maybe if it was scoped to the core, but components (as more easily overridden by a platform) are allowed to use them, perhaps. In general, I think the advice should be do not use them unless you have a very good reason. For sure the way we were using them was inappropriate. An attempt to escape unsafe in a non-performant way.

Updated to allow `Atomics` in components in unique circumstances (see requirement 3).

docs/src/rfc/text/0018-remove-atomics.md
joschock: Added a table of sample perf measurements to the discussion in the motivation section.

docs/src/rfc/text/0018-remove-atomics.md
joschock: Added a section describing the addition of new lock primitives as per this discussion and updated the recommended set of changes to reference them.

joschock:
> @joschock, just a reminder to please keep the [Synchronization ](https://github.com/OpenDevicePartnership/patina/blob/main/docs/src/dxe_core/synchronization.md) doc updated as things progress.

Added this to the plan described in the RFC:
```
As part of implementing this RFC, issues will be filed for all of the above items as appropriate to track the work of
implementing atomic removal and updating documentation as needed to describe the new primitives and recommended approach to synchronization.
```

cfernald (approved):

joschock:
Updated with review feedback, moving to FCP.

joschock:
Based on initial implementation work, plan to update RFC with some remarks on TPL handling and to rework the part about removing TplMutex.

PR #818, opened 2025-10-02
add patina_smbios component with safe Rust SMBIOS table management

kat-perez:
## Description

This PR implements a pure Rust-based SMBIOS (System Management BIOS) component for Patina firmware, providing functionality equivalent to the `EFI_SMBIOS_PROTOCOL` defined in the UEFI specification. The implementation introduces an `SmbiosRecords` trait that encapsulates SMBIOS table management behind a safe, idiomatic Rust API while maintaining compatibility with existing UEFI protocols.

### Key Features Implemented

**Core Service Architecture**
- Component/service pattern using `SmbiosProviderManager` with `#[derive(IntoComponent, IntoService)]`
- Service registration via Commands pattern for consumption by other components
- Configurable SMBIOS version support (defaults to SMBIOS 3.9+)

**Memory Safety & Validation**
- Safe `add_from_bytes()` API as the primary interface for adding SMBIOS records
- Comprehensive string validation with multi-layer checks (individual strings, string pools, and complete records)
- Protected unsafe methods retained only for legacy UEFI protocol compatibility (strongly discouraged)
- All validation follows SMBIOS specification constraints (64-byte string limits, proper null termination)

**Handle Management**
- Optimized counter-based handle allocation with O(1) average case performance
- Sequential allocation from 1 with wraparound logic (resets at 0xFEFF)
- Automatic skipping of reserved handles (0, 0xFFFE, 0xFFFF)
- Handle reuse optimization when records are removed

**String Operations**
- Full `update_string()` implementation that parses, updates, and rebuilds record data
- Support for 1-indexed string references per SMBIOS specification
- Complete record reconstruction with updated string pools

**Thread Safety**
- Mutex protection using `spin::Mutex` for all mutable operations
- Trait object safety for service injection compatibility

**SMBIOS 3.9+ Focus**
- Exclusive support for 64-bit entry point structures
- Eliminates legacy 32-bit format complexity
- Designed specifically for modern UEFI environments with flexible memory layouts

- [x] Impacts functionality?
- [x] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [x] Includes documentation?

## How This Was Tested

Integrated into `patina-qemu`

The `patina_smbios` component includes:

- Detailed usage examples showing record creation (Type 0 BIOS Information, Type 1 System Information, OEM custom records)
- Complete data flow documentation showing byte-level structure
- Validation logic with comprehensive safety checks
- Integration examples for component usage

Testing performed through:
- Integration with platform-specific DXE core implementations
- Validation of SMBIOS table generation and protocol installation
- Verification of string manipulation and handle allocation
- Compliance testing against SMBIOS 3.0+ specification

Co-authored-by: Anh Luong <anh.luong@dell.com>

### DIFF
--- a/components/patina_smbios/src/smbios_record.rs
+++ b/components/patina_smbios/src/smbios_record.rs
@@ -42,7 +42,13 @@ pub trait SmbiosRecordStructure {
     fn string_pool_mut(&mut self) -> &mut Vec<String>;
 }
 
-/// Generic SMBIOS record serializer using reflection-like techniques
+/// Generic SMBIOS record serializer using compile-time field introspection
+///
+/// This serializer uses a reflection-like pattern implemented via traits and macros
+/// to introspect struct fields at compile-time. Each SMBIOS record type implements
+/// `SmbiosFieldLayout` which provides metadata about field types and offsets, allowing
+/// the serializer to generically extract and serialize primitive fields while handling
+/// the `Vec<String>` string pool separately.
 pub struct SmbiosSerializer;
 
 impl SmbiosSerializer {
@@ -87,10 +93,6 @@ impl SmbiosSerializer {
         // This field layout mechanism only extracts the primitive fields (u8, u16, u32, u64, uuid)
         // that form the structured portion of the SMBIOS table. The string pool is handled separately
         // by serialize_string_pool().
-        //
-        // This is why we can't use zerocopy::IntoBytes on the entire struct - it contains non-primitive
-        // types. The field layout approach allows us to selectively serialize only the C-compatible
-        // primitive fields while keeping the convenient Vec<String> API for users.
         for field_info in &layout.fields {
             match field_info.field_type {
                 FieldType::U8(offset) => {
@@ -101,19 +103,16 @@ impl SmbiosSerializer {
                 FieldType::U16(offset) => {
                     let ptr = (record as *const T as *const u8).wrapping_add(offset);
                     let value = unsafe { core::ptr::read_unaligned(ptr as *const u16) };
-                    // Use zerocopy for endian-safe serialization
                     bytes.extend_from_slice(value.as_bytes());
                 }
                 FieldType::U32(offset) => {
                     let ptr = (record as *const T as *const u8).wrapping_add(offset);
                     let value = unsafe { core::ptr::read_unaligned(ptr as *const u32) };
-                    // Use zerocopy for endian-safe serialization
                     bytes.extend_from_slice(value.as_bytes());
                 }
                 FieldType::U64(offset) => {
                     let ptr = (record as *const T as *const u8).wrapping_add(offset);
                     let value = unsafe { core::ptr::read_unaligned(ptr as *const u64) };
-                    // Use zerocopy for endian-safe serialization
                     bytes.extend_from_slice(value.as_bytes());
                 }
                 FieldType::ByteArray { offset, len } => {


kat-perez:
putting this temporarily into draft state until I resolve some more build errors

os-d:
@kat-perez , happy to review once you've put this back in non-draft. Please just re-request reviews when you are ready

components/patina_smbios/examples/vendor_oem.rs
```diff
@@ -0,0 +1,203 @@
+use patina_smbios::smbios_derive::{SMBIOS_HANDLE_PI_RESERVED, SmbiosManager, SmbiosTableHeader};
+use patina_smbios::smbios_record::{
+    FieldInfo, FieldLayout, FieldType, SmbiosFieldLayout, SmbiosRecordStructure, Type0PlatformFirmwareInformation,
+    Type1SystemInformation, Type2BaseboardInformation, Type3SystemEnclosure,
+};
+use std::string::String;
```
os-d: We shouldn't use std in the example, since that cannot actually be used by production code. These two should be switched to the core lib.

components/patina_smbios/examples/vendor_oem.rs
```diff
@@ -0,0 +1,203 @@
...
+        system_bios_minor_release: 4,          // BIOS minor version
+        embedded_controller_major_release: 1,  // EC major version
+        embedded_controller_minor_release: 0,  // EC minor version
+        extended_bios_rom_size: 0x0000,        // No extended size needed
+        string_pool: vec![String::from("ACME BIOS Corp"), String::from("v2.4.1"), String::from("09/26/2025")],
+    };
+
+    let bios_bytes = bios_rec.to_bytes();
+    let bios_header: SmbiosTableHeader =
+        unsafe { core::ptr::read_unaligned(bios_bytes[..header_size].as_ptr() as *const SmbiosTableHeader) };
```
os-d: This might be a place to use the [zerocopy crate](https://docs.rs/zerocopy/latest/zerocopy/). We plan to use this extensively in patina and have been converting existing unsafe manipulations to use it. Same anywhere bytes are being converted to a structure or vice versa.

components/patina_smbios/examples/vendor_oem.rs
```diff
@@ -0,0 +1,203 @@
...
+    let enclosure_header: SmbiosTableHeader =
+        unsafe { core::ptr::read_unaligned(enclosure_bytes[..header_size].as_ptr() as *const SmbiosTableHeader) };
+
+    let _enclosure_handle = unsafe { manager.add(None, &enclosure_header).expect("enclosure add failed") };
+
+    // Verify all five records were added
+    let mut search = SMBIOS_HANDLE_PI_RESERVED;
+    let (found, _) = manager.get_next(&mut search, Some(VendorOemRecord::RECORD_TYPE)).expect("get_next failed");
+    assert_eq!(found.record_type, VendorOemRecord::RECORD_TYPE);
+    println!("Added vendor record handle: {}", search);
```
os-d: using log::info! instead of println!

components/patina_smbios/src/component.rs
```diff
@@ -9,15 +9,120 @@
...
+pub struct SmbiosConfiguration {
+    /// SMBIOS major version (e.g., 3 for SMBIOS 3.x)
+    pub major_version: u8,
+    /// SMBIOS minor version (e.g., 0 for SMBIOS 3.0)
+    pub minor_version: u8,
+}
+
+impl Default for SmbiosConfiguration {
+    fn default() -> Self {
+        Self { major_version: 3, minor_version: 0 }
```
os-d, line 37: 3.0 is pretty old now, should we default to 3.9, the latest?

components/patina_smbios/src/component.rs
```diff
@@ -9,15 +9,120 @@
...
-    fn entry_point(self) -> Result<()> {
-        log::info!("Hello from SmbiosProviderManager");
+    /// Create a new SMBIOS provider manager with default SMBIOS 3.0 version
+    pub fn new() -> Self {
+        Self { manager: SmbiosManager::new(3, 0) }
+    }
+
+    /// Initialize the SMBIOS provider and register it as a service
+    fn entry_point(mut self, config: Option<Config<SmbiosConfiguration>>, mut commands: Commands) -> Result<()> {
+        log::info!("Initializing SMBIOS Provider...");
```
os-d: I recommend having entry and exit statements at trace level. Almost always a platform won't want to see these messages, unless they are debugging something. While this is in a feature branch the higher level is okay.

components/patina_smbios/src/component.rs
```diff
@@ -9,15 +9,120 @@
...
+pub struct SmbiosProviderManager {
+    manager: SmbiosManager,
+}
 
 impl SmbiosProviderManager {
-    fn entry_point(self) -> Result<()> {
-        log::info!("Hello from SmbiosProviderManager");
+    /// Create a new SMBIOS provider manager with default SMBIOS 3.0 version
+    pub fn new() -> Self {
+        Self { manager: SmbiosManager::new(3, 0) }
```
os-d: Use the SmbiosConfiguration::default() for consistency?

components/patina_smbios/src/component.rs
```diff
@@ -9,15 +9,120 @@
...
+
+        // Register the service so other components can consume it
+        commands.add_service(self);
+
+        log::info!("SMBIOS Provider initialized successfully");
         Ok(())
     }
 }
+
+impl Default for SmbiosProviderManager {
```
os-d: Is Default really needed here? It doesn't do anything differently than calling new()

components/patina_smbios/src/lib.rs
```diff
@@ -1,15 +1,54 @@
-//! SMBIOS Support
+//! SMBIOS (System Management BIOS) component for Patina
 //!
-//! ## License
-//!
-//! Copyright (C) Microsoft Corporation. All rights reserved.
-//!
-//! SPDX-License-Identifier: BSD-2-Clause-Patent
-//!
-#![cfg_attr(not(feature = "std"), no_std)]
```
os-d, line 9: Lost the license

components/patina_smbios/src/lib.rs
```diff
@@ -1,15 +1,54 @@
...
 
-mod component;
-/// SMBIOS table and type definitions
-pub mod smbios;
+        // Print bytes as hex; run tests with `-- --nocapture` to see this output
+        print!("Record bytes ({}):", bytes.len());
+        for b in &bytes {
+            print!(" {:02X}", b);
+        }
+        println!();
```
os-d: We shouldn't have prints in tests. Those can be added locally for debugging, but in checked in code we should only really on asserts and the like.

components/patina_smbios/src/smbios_derive.rs
```diff
@@ -0,0 +1,958 @@
...
+    StringTooLong,
+    BufferTooSmall,
+}
+
+pub trait SmbiosRecords<'a> {
+    /// Adds an SMBIOS record to the SMBIOS table.
+    ///
+    /// # Safety
+    ///
+    /// **WARNING: This method is unsafe and should be avoided in favor of `add_from_bytes`.**
```
os-d: Then why do we have it at all? :). I know the C protocol has the Add function, but in our efiapi wrapper of this, is it possible to manipulate the incoming parameters such that the safe version can be used? If not, we should ensure that this function is not accessible from Rust code, only the C protocol layer.

components/patina_smbios/src/smbios_derive.rs
```diff
@@ -0,0 +1,958 @@
...
+
+    /// Discovers SMBIOS records, optionally filtered by type.
+    fn get_next(
+        &self,
+        smbios_handle: &mut SmbiosHandle,
+        record_type: Option<SmbiosType>,
+    ) -> Result<(&SmbiosTableHeader, Option<Handle>), SmbiosError>;
+
+    /// Provides an iterator over all SMBIOS records.
+    fn iter(&self) -> Box<dyn Iterator<Item = &'a SmbiosRecord> + 'a>;
```
os-d: Is there any need for a mutable iterator?

components/patina_smbios/src/smbios_derive.rs
```diff
@@ -0,0 +1,958 @@
...
+            table_64_address: None,
+            lock: Mutex::new(()),
+        }
+    }
+
+    fn validate_string(s: &str) -> Result<(), SmbiosError> {
+        if s.len() > SMBIOS_STRING_MAX_LENGTH {
+            return Err(SmbiosError::StringTooLong);
+        }
+        if s.contains('\0') {
```
os-d, line 205: Just confirming that you are expecting the strings to not have null terminators?

components/patina_smbios/src/smbios_derive.rs
```diff
@@ -0,0 +1,958 @@
...
+
+        let mut count = 0;
+        let mut i = 0;
+        let data_end = len - 2; // Exclude the final double-null
+
+        while i < data_end {
+            let start = i;
+
+            // Find the next null terminator (end of current string)
+            while i < data_end && string_pool_area[i] != 0 {
```
os-d: I am not a Rust string expert, but it seems like there would be some Rust core functions to do this string parsing?

components/patina_smbios/src/smbios_derive.rs
```diff
@@ -0,0 +1,958 @@
...
+        // Validate all strings first
+        for s in strings {
+            Self::validate_string(s)?;
+        }
+
+        let mut record = Vec::new();
+
+        // Add the structured data
+        let header_bytes =
+            unsafe { core::slice::from_raw_parts(header as *const _ as *const u8, header.length as usize) };
```
os-d: zerocopy?

components/patina_smbios/src/smbios_derive.rs
```diff
@@ -0,0 +1,958 @@
...
+
+            // Move past the null terminator
+            i += 1;
+        }
+
+        Ok(count)
+    }
+
+    #[allow(dead_code)]
+    fn build_record_with_strings(header: &SmbiosTableHeader, strings: &[&str]) -> Result<Vec<u8>, SmbiosError> {
```
os-d: I believe &str is a fat pointer, is this intended to be put in the actual SMBIOS table? If so that would cause an issue when C code tries to interpret it?

components/patina_smbios/src/smbios_derive.rs
```diff
@@ -0,0 +1,958 @@
...
+        // Start from next_handle, try sequential allocation up to 0xFEFF
+        // If all are allocated, wrap around from 1
+        let mut attempts = 0u32;
+        const MAX_ATTEMPTS: u32 = 0xFEFF;
+
+        loop {
+            let candidate = self.next_handle;
+
+            // Skip reserved handles (0xFFFE, 0xFFFF, 0)
+            if candidate == 0 || candidate >= 0xFEFF {
```
os-d: Are 0xFF00 and up not valid? The comment seems out of sync with the check.

components/patina_smbios/src/smbios_derive.rs
```diff
@@ -0,0 +1,958 @@
...
+                self.next_handle = 1;
+                attempts += 1;
+                if attempts >= MAX_ATTEMPTS {
+                    return Err(SmbiosError::OutOfResources);
+                }
+                continue;
+            }
+
+            if !self.is_handle_allocated(candidate) {
```
os-d: Is it worth keeping track of allocated handles separately so you can just query what the next one is instead of searching all records each time? That feels odd to iterate through all possible handles and all records when we already know (just don't keep track of) which handles are in use.

To save space, it could just be using next_handle and then when remove is called (which I'm assuming is the only time a handle gets freed) it gets added to a freed_handle list kind of thing. If next_handle == highest_allowed_handle: pop from free list.

I assume this doesn't happen very often.

components/patina_smbios/src/smbios_derive.rs
```diff
@@ -0,0 +1,958 @@
...
+        }
+    }
+
+    #[allow(dead_code)]
+    fn install_configuration_table(&self) -> Result<(), SmbiosError> {
+        // This would interact with UEFI Boot Services to install
+        // the SMBIOS table in the system configuration table
+        // Implementation depends on your UEFI framework
+
+        // For SMBIOS 2.x
```
os-d: Should this even support SMBIOS 2.0? The last SMBIOS spec (if copilot is correct) was released in 2013. Unless we have an active use case, I would say this implementation should only support the latest SMBIOS spec. Anything legacy can use the edk2 driver.

components/patina_smbios/src/smbios_derive.rs
```diff
@@ -0,0 +1,958 @@
...
+        producer_handle: Option<Handle>,
+        record_data: &[u8],
+    ) -> Result<SmbiosHandle, SmbiosError> {
+        // Step 1: Validate minimum size for header (at least 4 bytes)
+        if record_data.len() < core::mem::size_of::<SmbiosTableHeader>() {
+            return Err(SmbiosError::BufferTooSmall);
+        }
+
+        // Step 2: Parse and validate header
+        let header = unsafe { &*(record_data.as_ptr() as *const SmbiosTableHeader) };
```
os-d: This again feels like a place we can use zerocopy to handle the unsafe checks for us.

components/patina_smbios/src/smbios_derive.rs
```diff
@@ -0,0 +1,958 @@
...
+            SmbiosTableHeader { record_type: header.record_type, length: header.length, handle: smbios_handle };
+        record_header.handle = smbios_handle;
+
+        // Update the handle in the actual data
+        let mut data = record_data.to_vec();
+        let handle_bytes = smbios_handle.to_le_bytes();
+        data[2] = handle_bytes[0]; // Handle is at offset 2 in header
+        data[3] = handle_bytes[1];
+
+        let smbios_record = SmbiosRecord {
```
os-d: consider a new() function here to match the style of the rest of this

components/patina_smbios/src/smbios_derive.rs
```diff
@@ -0,0 +1,958 @@
...
+
+        // Extract existing strings from the string pool
+        let string_pool_start = header_length;
+        let string_pool = &record.data[string_pool_start..];
+
+        let mut existing_strings = Vec::new();
+        let mut current_string = Vec::new();
+        let mut null_count = 0;
+
+        for &byte in string_pool {
```
os-d: It seems like we could do this at the string level instead of the byte level

components/patina_smbios/src/smbios_derive.rs
```diff
@@ -0,0 +1,958 @@
...
+        Err(SmbiosError::HandleNotFound)
+    }
+
+    fn iter(&self) -> Box<dyn Iterator<Item = &'static SmbiosRecord> + 'static> {
+        // We need to use unsafe here because we're extending the lifetime of the iterator.
+        // This is safe because the SmbiosManager is 'static and the records vector
+        // is only modified through &mut self methods.
+        let records_ptr = self.records.as_ptr();
+        let len = self.records.len();
+        Box::new((0..len).map(move |i| unsafe { &*records_ptr.add(i) }))
```
os-d: self.records is a Vec, which you can already iterate over. Why not use self.records.iter() (or similar) here?

components/patina_smbios/src/smbios_derive.rs
```diff
@@ -0,0 +1,958 @@
...
+    pub length: u8,
+    pub handle: SmbiosHandle,
+}
+
+impl SmbiosTableHeader {
+    pub fn new(record_type: SmbiosType, length: u8, handle: SmbiosHandle) -> Self {
+        Self { record_type, length, handle }
+    }
+
+    // fn get_smbios_structure_size(&self) -> (usize, usize) {
```
os-d: Remove?

components/patina_smbios/src/smbios_derive.rs
```diff
@@ -0,0 +1,958 @@
...
+
+    // }
+}
+
+/// Internal SMBIOS record representation
+pub struct SmbiosRecord {
+    pub header: SmbiosTableHeader,
+    pub producer_handle: Option<Handle>,
+    pub data: Vec<u8>, // Complete record including strings
+    // record_size: usize,
```
os-d: Remove if this is not needed?

components/patina_smbios/src/smbios_derive.rs
```diff
@@ -0,0 +1,958 @@
...
+    string_count: usize,
+    pub smbios32_table: bool,
+    pub smbios64_table: bool,
```
os-d: Presumably it can't be both of these at the same time? If not, consider an enum here to avoid invalid input

components/patina_smbios/src/smbios_derive.rs
```diff
@@ -0,0 +1,958 @@
...
+    strings: Vec<String>,
+}
+
+impl SmbiosRecordBuilder {
+    pub fn new(record_type: u8) -> Self {
+        Self { record_type, data: Vec::new(), strings: Vec::new() }
+    }
+
+    pub fn add_field<T: Copy>(mut self, value: T) -> Self {
+        let bytes = unsafe { core::slice::from_raw_parts(&value as *const T as *const u8, core::mem::size_of::<T>()) };
```
os-d: zerocopy?

components/patina_smbios/src/smbios_derive.rs
```diff
@@ -0,0 +1,958 @@
...
+        let mut record = Vec::new();
+
+        // Add header
+        let header = SmbiosTableHeader {
+            record_type: self.record_type,
+            length: (core::mem::size_of::<SmbiosTableHeader>() + self.data.len()) as u8,
+            handle: SMBIOS_HANDLE_PI_RESERVED,
+        };
+
+        let header_bytes = unsafe {
```
os-d: zerocopy?

components/patina_smbios/src/smbios_derive.rs
```diff
@@ -0,0 +1,958 @@
...
+}
+
+unsafe impl ProtocolInterface for SmbiosProtocol {
+    const PROTOCOL_GUID: efi::Guid = efi::Guid::from_fields(
+        0x03583ff6,
+        0xcb36,
+        0x4940,
+        0x94,
+        0x7e,
+        &[0xb9, 0xb3, 0x9f, 0x4a, 0xfa, 0xf7], // ✅ Corrected GUID
```
os-d: This comment can be dropped :)

components/patina_smbios/src/smbios_derive.rs
```diff
@@ -0,0 +1,958 @@
...
+            update_string: Self::update_string_ext,
+            remove: Self::remove_ext,
+            get_next: Self::get_next_ext,
+            major_version,
+            minor_version,
+        }
+    }
+
+    #[allow(dead_code)]
+    extern "efiapi" fn add_ext(
```
os-d, line 740: This is the level where I would take the unsafe C input, do a single unsafe operation to convert to Rust structures and then call the Rust safe API. You can then entirely drop the add function.

components/patina_smbios/src/smbios_record.rs
```diff
@@ -0,0 +1,406 @@
+extern crate alloc;
+use crate::smbios_derive::{SMBIOS_HANDLE_PI_RESERVED, SmbiosError, SmbiosTableHeader};
+use alloc::string::String;
+use alloc::vec::Vec;
+
+macro_rules! vec {
```
os-d: Why is this here? I assume this is the same or close to what the core vec! already does? Why not use that?

components/patina_smbios/src/smbios_record.rs
```diff
@@ -0,0 +1,406 @@
...
+
+    /// Convert the structure to a complete SMBIOS record byte array
+    // fn to_bytes(&self) -> Vec<u8>;
+    fn to_bytes(&self) -> Vec<u8>
+    // where Self: SmbiosFieldLayout
+    where
+        Self: SmbiosFieldLayout,
+        Self: Sized,
+    {
+        SmbiosSerializer::serialize(self)
```
os-d, line 32: This would again potentially be a place for zerocopy instead of custom logic

components/patina_smbios/src/smbios_record.rs
```diff
@@ -0,0 +1,406 @@
...
+    fn string_pool(&self) -> &[String];
+
+    /// Get mutable access to the string pool
+    fn string_pool_mut(&mut self) -> &mut Vec<String>;
+}
+
+/// Generic SMBIOS record serializer using reflection-like techniques
+pub struct SmbiosSerializer;
+
+impl SmbiosSerializer {
```
os-d, line 54: Can this entire thing be replaced with zerocopy?

components/patina_smbios/src/smbios_record.rs
```diff
@@ -0,0 +1,406 @@
...
+            }
+            bytes.push(0); // Double null terminator
+        }
+
+        bytes
+    }
+}
+
+/// Field layout description for generic serialization
+pub trait SmbiosFieldLayout {
```
os-d, line 149: Same here is this needed if zerocopy can be used?

components/patina_smbios/src/component.rs
```diff
@@ -9,15 +9,120 @@
...
+        IntoComponent,
+        params::{Commands, Config},
+        service::IntoService,
+    },
+    error::Result,
+};
+
+/// Configuration for SMBIOS service
+#[derive(Debug, Clone)]
+pub struct SmbiosConfiguration {
```
os-d, line 28: Nothing in this seems to actually depend on the version being set (maybe in non-changed code). Does this actually need to be configurable? I'd assume the component should just report the version it actually supports and not let that be configurable.

components/patina_smbios/examples/vendor_oem.rs
kat-perez: SmbiosTableHeader now uses zerocopy::FromBytes

components/patina_smbios/examples/vendor_oem.rs
kat-perez: updated

components/patina_smbios/examples/vendor_oem.rs
kat-perez: updated to use core

components/patina_smbios/src/component.rs
kat-perez: agree. removing since it's not necessary

components/patina_smbios/src/component.rs
kat-perez: updated to latest

components/patina_smbios/src/component.rs
kat-perez: modifying to use trace-level instead

components/patina_smbios/src/component.rs
kat-perez: updated

components/patina_smbios/src/lib.rs
kat-perez: oops, added it back in

components/patina_smbios/src/lib.rs
kat-perez: prints removed

components/patina_smbios/src/component.rs
kat-perez: added it back in since clippy suggests there should be a Default implementation and same for `unrwap_or_default()`

components/patina_smbios/src/smbios_derive.rs
kat-perez: removing the add function completely in favor of the add_from_bytes function instead

components/patina_smbios/src/smbios_derive.rs
kat-perez: removing it since it is unused

components/patina_smbios/src/smbios_derive.rs
kat-perez: removed 2.x support

components/patina_smbios/src/smbios_derive.rs
kat-perez: dropping

components/patina_smbios/src/smbios_derive.rs
kat-perez: add function is dropped

components/patina_smbios/src/smbios_record.rs
kat-perez: removed the macro and using core vec! instead

components/patina_smbios/src/smbios_derive.rs
kat-perez: yes, updated

components/patina_smbios/src/smbios_derive.rs
kat-perez: the strings shouldn't have null terminators. they're added when building the string pool

components/patina_smbios/src/smbios_derive.rs
kat-perez: updated

components/patina_smbios/src/smbios_derive.rs
kat-perez: updated

components/patina_smbios/src/smbios_derive.rs
kat-perez: these record structs were designed to not be C-compatible but adding documentation to make that clear

components/patina_smbios/src/smbios_derive.rs
kat-perez: fixed

components/patina_smbios/src/smbios_derive.rs
kat-perez: removed

components/patina_smbios/src/smbios_derive.rs
kat-perez: removed

components/patina_smbios/src/smbios_derive.rs
kat-perez: removed it instead since we only want to support >3.x

components/patina_smbios/src/smbios_derive.rs
kat-perez: updated

components/patina_smbios/src/smbios_derive.rs
kat-perez: also updated

components/patina_smbios/src/smbios_derive.rs
kat-perez: added optimization

components/patina_smbios/src/smbios_derive.rs
kat-perez: added new()

components/patina_smbios/src/smbios_derive.rs
kat-perez: refactored and added a helper function

components/patina_smbios/src/smbios_derive.rs
kat-perez: updated

components/patina_smbios/src/smbios_record.rs
kat-perez: some parts cannot be changes because the struct contains Vec<String> which isn't part of the binary format

os-d (approved):
Per offline comment, thanks for resolving all of my comments. Approving to allow the feature branch merge to come through and when the PR to bring the feature branch to the mainline comes through we can do another deep review. Happy to review further feature branch PRs before then, of course!

PR #820, opened 2025-10-03
patina_mm: Improve API, validation, and logging [Rebase & FF]

makubacki:
## Description

Series of commits to make general improves to `patina_mm`.

> Note: I'm going to move `patina_mm` to use Patina GUIDs in a future change, but that's not here.

---

**patina_mm: Enhance CommunicateBuffer memory safety and API**

Add some improvements to the CommunicateBuffer API.

Memory Safety:

- Improve validation in `from_raw_parts`
- Add page alignment checks and address overflow validation
- Add new `from_firmware_region` method with additional checks
  for using a comm buffer with a given address and size
- Replace panics with `Result` return types

API Improvements:

- Make `EfiMmCommunicateHeader` fields private with safe accessors
- Add `Copy` trait to `EfiMmCommunicateHeader`
- Add a constructor
- Add compile-time constants for the comm recipient GUID and message
  length

Buffer Management:

- Add `validate_capacity` method for upfront validation
  - Use in `set_message()`
- Make internal buffer access methods private
- Add zero-initialization for new buffers

Error Handling:

- Add more error types to `CommunicateBufferStatus` to better
  identify specific error cases
- Add additional validation for some operations

Testing:

- Update all tests to use the Result-based APIs
- Include page-aligned buffer testing and test alignment

---

**patina_mm: Add Display trait for MmCommunicationConfiguration**

Implement the `Display` trait for `MmCommunicationConfiguration` so
MM Communication Configuration structures are easier to read.

---

**patina_mm: Add `MmiPort` and `AcpiBase` display traits**

Adds a `Display` implementation for:

- `MmiPort`
- `AcpiBase`

With simple unit tests to check formatting.

---

**patina_mm: Add targeted logging to MM communication flow**

Adds a `mm_comm` log target to MM communication related flows to be
able to trace communicate behavior in more detail.

---

**patina_mm: Add targeted logging to SW MMI management**

Adds detailed logging message to `sw_mmi_manager` under the `sw_mmi`
log target. These can be used to trace the SW MMI flow.

---

**patina_mm: Improve CommunicateBuffer state management**

Adds private state tracking to `CommunicateBuffer` so the comm buffer
in memory can be regularly audited during API operations against the
internal private data.

- Adds `private_recipient` and `private_message_length` fields for
  independent state verification. Allows consistency to be verified
  between in-memory structures and API state
- Updated `get_message()` return type to allow better error handling
  (return `Result` instead of just a `Vec<u8>`)
- Add more logging through the 'mm_comm' target
- Improved `set_message_info()` and `set_message()` validation and
  added consistency checking to their implementation
- Updated the `MmCommunicator` service to better handle `get_message()`
  responses with a new `Status::InvalidResponse` error

---

**patina_mm: Add comm buffer state verification**

Adds a new method to `CommunicateBuffer` called `verify_state_consistency`
that actually does the verification of internal state against the comm
buffer.

Update some operations like `set_message_info()`, `set_message()`,
`get_message()`, `get_header_guid()`, and `get_messgage_length()`
to verify state consistency when they're called.

---

- [x] Impacts functionality?
- [x] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [x] Includes documentation?

## How This Was Tested

- `cargo make all`
- Q35 boot with MM Communicate service

## Integration Instructions

- Review the changes to the API to understand what has changed. Not marked as breaking as this does not require platforms using Patina to make any changes.

### DIFF
--- a/components/patina_mm/src/config.rs
+++ b/components/patina_mm/src/config.rs
@@ -330,6 +330,60 @@ impl CommunicateBuffer {
         self.len().saturating_sub(Self::MESSAGE_START_OFFSET)
     }
 
+    /// Verifies that the internal state matches what is in the memory buffer.
+    /// This is intended to catch corruption and ensure the buffer actually matches what has
+    /// been requested through the MM Communication API.
+    ///
+    /// Returns `Ok(())` if state verification passes, otherwise returns the appropriate error.
+    fn verify_state_consistency(&self) -> Result<(), CommunicateBufferStatus> {
+        if self.len() < Self::MESSAGE_START_OFFSET {
+            log::error!(target: "mm_comm", "Buffer {} is too small for the communicate header", self.id);
+            return Err(CommunicateBufferStatus::TooSmallForHeader);
+        }
+
+        let header_slice = &self.as_slice()[..Self::MESSAGE_START_OFFSET];
+
+        let memory_guid = unsafe {
+            // SAFETY: We've validated the buffer has at least MESSAGE_START_OFFSET bytes
+            // and efi::Guid has a stable #[repr(C)] layout
+            core::ptr::read(header_slice.as_ptr() as *const efi::Guid)
+        };
+
+        let memory_message_length = unsafe {
+            // SAFETY: We've validated the buffer has sufficient bytes for usize at correct offset
+            core::ptr::read(header_slice.as_ptr().add(16) as *const usize)
+        };
+
+        // Verify that thee recipient matches
+        match self.private_recipient {
+            Some(expected_guid) => {
+                if memory_guid != expected_guid {
+                    log::error!(target: "mm_comm", "Buffer {} GUID mismatch: private={:?}, memory={:?}",
+                        self.id, expected_guid, memory_guid);
+                    return Err(CommunicateBufferStatus::InvalidRecipient);
+                }
+            }
+            None => {
+                // If no recipient is set privately, the memory should contain all zeros for the GUID
+                let zero_guid = efi::Guid::from_fields(0, 0, 0, 0, 0, &[0; 6]);
+                if memory_guid != zero_guid {
+                    log::error!(target: "mm_comm", "Buffer {} unexpected GUID in memory when none set privately", self.id);
+                    return Err(CommunicateBufferStatus::InvalidRecipient);
+                }
+            }
+        }
+
+        // Verify message length matches
+        if memory_message_length != self.private_message_length {
+            log::error!(target: "mm_comm", "Buffer {} message length mismatch: private={}, memory={}",
+                self.id, self.private_message_length, memory_message_length);
+            return Err(CommunicateBufferStatus::TooSmallForMessage);
+        }
+
+        log::trace!(target: "mm_comm", "Buffer {} state consistency was verified successfully", self.id);
+        Ok(())
+    }
+
     /// Validates that the buffer can accommodate a header and message of the given size.
     ///
     /// ## Arguments
@@ -381,6 +435,9 @@ impl CommunicateBuffer {
         let header_bytes = header.as_bytes();
         self.as_slice_mut()[..Self::MESSAGE_START_OFFSET].copy_from_slice(header_bytes);
 
+        // Verify state consistency after update
+        self.verify_state_consistency()?;
+
         log::trace!(target: "mm_comm", "Message info set successfully for buffer {}", self.id);
         Ok(())
     }
@@ -416,6 +473,9 @@ impl CommunicateBuffer {
         self.as_slice_mut()[Self::MESSAGE_START_OFFSET..Self::MESSAGE_START_OFFSET + message.len()]
             .copy_from_slice(message);
 
+        // Verify state consistency after update
+        self.verify_state_consistency()?;
+
         log::debug!(target: "mm_comm", "Buffer {} message set successfully: header_size={}, message_size={}",
             self.id, Self::MESSAGE_START_OFFSET, message.len());
         Ok(())
@@ -426,6 +486,9 @@ impl CommunicateBuffer {
     ///
     /// Note: This method extracts the actual message content using verified state tracking.
     pub fn get_message(&self) -> Result<Vec<u8>, CommunicateBufferStatus> {
+        // Verify state consistency before proceeding
+        self.verify_state_consistency()?;
+
         if self.private_message_length == 0 {
             log::trace!(target: "mm_comm", "Buffer {} has zero-length message", self.id);
             return Ok(Vec::new());
@@ -444,6 +507,29 @@ impl CommunicateBuffer {
         let message = self.as_slice()[start_offset..end_offset].to_vec();
         log::trace!(target: "mm_comm", "Retrieved message from buffer {}: message_size={}", self.id, message.len());
         Ok(message)
+    }
+
+    /// Returns the header GUID from the current communicate buffer.
+    /// This method uses the internal state and verifies consistency with memory.
+    ///
+    /// Returns `None` if no recipient has been set.
+    pub fn get_header_guid(&self) -> Result<Option<efi::Guid>, CommunicateBufferStatus> {
+        // Verify state consistency first
+        self.verify_state_consistency()?;
+
+        log::trace!(target: "mm_comm", "Buffer {} header GUID retrieved from private state", self.id);
+        Ok(self.private_recipient)
+    }
+
+    /// Returns the message length from the current communicate buffer.
+    /// This method uses the internal state and verifies consistency with memory.
+    pub fn get_message_length(&self) -> Result<usize, CommunicateBufferStatus> {
+        // Verify state consistency first
+        self.verify_state_consistency()?;
+
+        log::trace!(target: "mm_comm", "Buffer {} message length retrieved from private state: len={}",
+            self.id, self.private_message_length);
+        Ok(self.private_message_length)
     }
 }
 
@@ -601,6 +687,10 @@ mod tests {
             Guid::from_fields(0x12345678, 0x1234, 0x5678, 0x90, 0xAB, &[0xCD, 0xEF, 0x01, 0x23, 0x45, 0x67]);
 
         assert!(comm_buffer.set_message_info(recipient_guid).is_ok());
+
+        // Test that state verification works
+        assert!(comm_buffer.get_header_guid().is_ok());
+        assert_eq!(comm_buffer.get_header_guid().unwrap(), Some(recipient_guid));
     }
 
     #[test]
@@ -661,7 +751,10 @@ mod tests {
 
         // Test that we can retrieve the message
         let retrieved_message = comm_buffer.get_message().unwrap();
-        assert_eq!(retrieved_message[..message.len()], *message);
+        assert_eq!(retrieved_message, message);
+
+        // Test that state verification is successful
+        assert_eq!(comm_buffer.get_message_length().unwrap(), message.len());
     }
 
     #[test]
@@ -700,7 +793,7 @@ mod tests {
         assert!(comm_buffer.set_message(MESSAGE).is_ok(), "Failed to set the message");
 
         let retrieved_message = comm_buffer.get_message().unwrap();
-        assert_eq!(retrieved_message[..MESSAGE.len()], *MESSAGE);
+        assert_eq!(retrieved_message, MESSAGE.to_vec());
     }
 
     #[test]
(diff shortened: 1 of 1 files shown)


components/patina_mm/src/config.rs
```diff
@@ -52,32 +53,98 @@ impl Default for MmCommunicationConfiguration {
...
+        } else {
+            for buffer in &self.comm_buffers {
+                writeln!(f, "    Buffer {:#04X}: ptr={:p}, len=0x{:X}", buffer.id(), buffer.as_ptr(), buffer.len(),)?;
+            }
+            Ok(())
+        }
+    }
+}
+
+fn format_mmi_port(port: &MmiPort) -> String {
```
Javagedes: I think these formatting functions for `mmi_port` and `acpi_base` would be better off as `fmt::Display` implementations for them.

components/patina_mm/src/config.rs
```diff
@@ -52,32 +53,98 @@ impl Default for MmCommunicationConfiguration {
...
-    /// Returns the communicate header as a slice of bytes.
+    /// Create a new communicate header with the specified GUID and message length.
+    pub const fn new(header_guid: efi::Guid, message_length: usize) -> Self {
+        Self { header_guid, message_length }
+    }
+
+    /// Returns the communicate header as a slice of bytes using safe conversion.
+    ///
+    /// Useful if byte-level access to the header structure is needed.
     pub fn as_bytes(&self) -> &[u8] {
```
Javagedes, line 101: would it be more appropriate to derive zerocopy for this struct?

components/patina_mm/src/config.rs
```diff
@@ -52,32 +53,98 @@ impl Default for MmCommunicationConfiguration {
...
     pub const fn size() -> usize {
         core::mem::size_of::<Self>()
     }
+
+    /// Returns the header GUID from this communicate header.
+    ///
+    /// # Returns
+    ///
+    /// The GUID that identifies the registered MM handler recipient.
+    #[allow(dead_code)]
```
Javagedes, line 118: Are these functions actually not being used, or is the `#[allow(dead_code)]` outdated? If they are not being used, do we need them? This struct is internal to the crate, so if the crate is not using it, it seems unnecessary.

components/patina_mm/src/config.rs
```diff
@@ -122,18 +199,28 @@ impl CommunicateBuffer {
     /// - The buffer must not be used by any other code.
     pub unsafe fn new(buffer: Pin<&'static mut [u8]>, id: u8) -> Self {
         let length = buffer.len();
+        log::debug!(target: "mm_comm", "Creating new CommunicateBuffer: id={}, size=0x{:X}", id, length);
```
Javagedes, line 180: This is a great idea, and we should probably start doing this more with our components. You should add documentation to the crate stating you can configure logging for this crate via the `mm_comm` target.

components/patina_mm/src/config.rs
```diff
@@ -122,18 +199,28 @@ impl CommunicateBuffer {
     /// - The buffer must not be used by any other code.
     pub unsafe fn new(buffer: Pin<&'static mut [u8]>, id: u8) -> Self {
```
Javagedes: I think the casting of the pointer from a `Pin<&'static mut [u8]>` to a `&'static mut [u8]` might be UB. I also think it is unnecessary. Since we are consuming the `Pin`, we can have mutable access to the `Pin` if we want it, so we should be able to just call `buffer.fill(0)` directly, and remove this casting.

```rust
pub unsafe fn new(mut buffer: Pin<&'static mut [u8]>, id: u8) -> Self {
    let length = buffer.len();
    log::debug!(target: "mm_comm", "Creating new CommunicateBuffer: id={}, size=0x{:X}", id, length);
    buffer.fill(0);

    log::trace!(target: "mm_comm", "CommunicateBuffer {} created successfully at address {:p}", id, buffer_ptr.as_ptr());
    Self { buffer: buffer_ptr, id, length, private_recipient: None, private_message_length: 0 }
}
```

components/patina_mm/src/config.rs
```diff
@@ -122,18 +199,28 @@ impl CommunicateBuffer {
     /// - The buffer must not be used by any other code.
```
Javagedes: I don't think I agree with most of these safety checks (They also may just be outdated). By requiring a `Pin<&'static mut [u8]>`, most of these checks are true by default. The only way it would not be, would be by them using one of the `unchecked` methods. In that scenario they need to satisfy the safety checks there.

However by definition

1. a `Pin<&'static mut [u8]>` is a valid pointer, and the size is baked into the [u8] fat pointer
2. The buffer cannot be null. You pass in the valid type with `new` or `set`
3. The type of the buffer input forces it to be static. You cannot get around that
4. The `Pin<...>` ensures it is not moved
5. The `Pin<...>` consumes the leaked buffer, so rusts lifetime static analysis should handle this.

components/patina_mm/src/config.rs
```diff
@@ -122,18 +199,28 @@ impl CommunicateBuffer {
...
     pub unsafe fn new(buffer: Pin<&'static mut [u8]>, id: u8) -> Self {
         let length = buffer.len();
+        log::debug!(target: "mm_comm", "Creating new CommunicateBuffer: id={}, size=0x{:X}", id, length);
+
         let buffer_ptr =
```
Javagedes: you should be able to get rid of this `expect` call by passing a reference to the already instantiated object

```rust
let ptr: NonNull<[u8]> = NonNull::from_mut(Pin::into_inner(buffer));
let ptr: NonNull<u8> = ptr.cast::<u8>();
```

components/patina_mm/src/config.rs
```diff
@@ -122,18 +199,28 @@ impl CommunicateBuffer {
...
+        slice.fill(0);
+
+        log::trace!(target: "mm_comm", "CommunicateBuffer {} created successfully at address {:p}", id, buffer_ptr.as_ptr());
+        Self { buffer: buffer_ptr, id, length, private_recipient: None, private_message_length: 0 }
     }
 
     /// Returns a reference to the buffer as a slice of bytes.
-    pub fn as_slice(&self) -> &'static [u8] {
+    /// This is only used for internal operations.
+    fn as_slice(&self) -> &[u8] {
```
Javagedes, line 191: I'm not sure how this affects the rest of the code, but if you store the value as a `NonNull<[u8]>` instead of a `NonNull<u8>`, you can just do `self.buffer.as_ref()` and `self.buffer.as_mut()`

components/patina_mm/src/config.rs
```diff
@@ -146,14 +233,82 @@ impl CommunicateBuffer {
     /// - The buffer must have a static lifetime.
     /// - The buffer must not be moved in memory while it is being used.
     /// - The buffer must not be used by any other code.
-    pub unsafe fn from_raw_parts(buffer: *mut u8, size: usize, id: u8) -> Self {
-        if size == 0 {
-            panic!("CommunicateBuffer::from_raw_parts: size is zero");
+    /// - The buffer must be page (4k) aligned so paging attributes can be applied to it.
```
Javagedes, line 214: I don't agree with a few of these existing safety checks. Safety checks are meant for things we have to trust the user is ensuring. We can self validate a few of these (and we already are). Namely we validate the size, and the pointer is not null.

components/patina_mm/src/config.rs
```diff
@@ -146,14 +233,82 @@ impl CommunicateBuffer {
...
+            return Err(CommunicateBufferStatus::NotAligned);
+        }
+
+        if buffer as usize > usize::MAX - size {
+            log::error!(target: "mm_comm", "Buffer {} address overflow: ptr=0x{:X}, size=0x{:X}", id, buffer as usize, size);
+            return Err(CommunicateBufferStatus::AddressValidationFailed);
+        }
+
+        log::debug!(target: "mm_comm", "CommunicateBuffer {} validation passed, creating buffer", id);
+        unsafe { Ok(Self::new(Pin::new_unchecked(core::slice::from_raw_parts_mut(buffer, size)), id)) }
```
Javagedes: From reading, I do not think you need to use `Pin::new_unchecked`. `&mut [u8]` is a valid type to `Pin`. Therefore the unsafe `core::slice::from_raw_parts_mut` is the unsafe part that produces a valid `&mut [u8]` that can be passed to the safe `Pin::new`.

nit - need safety comments, which is probbaly just saying something along the lines of the safety requirements for `are passed onto the caller.

components/patina_mm/src/config.rs
```diff
@@ -146,14 +233,82 @@ impl CommunicateBuffer {
...
+    /// The caller must ensure:
+    /// - The memory region is valid and accessible throughout buffer lifetime
+    /// - The memory is not used by other components concurrently
+    /// - The firmware has guaranteed the memory region is stable and properly mapped
+    pub unsafe fn from_firmware_region(
+        address: u64,
+        size_bytes: usize,
+        buffer_id: u8,
+    ) -> Result<Self, CommunicateBufferStatus> {
+        let ptr = address as usize as *mut u8;
```
Javagedes: This needs a comment - It took me some time to realize what you were doing. If I understand correctly, you are just making sure the address fits into the addressible memory range (32bit vs 64bit).

I think a cleaner way to do this would be:

```rust
/// Check that memory address is addressable on this system. A 32bit system will fail this if the address is over 4GB
let address = usize::try_from(address) else {
    return Err(CommunicateBufferStatus::AddressValidationFailed);
}

if address.checked_add(size_bytes).is_none() {
    return Err(CommunicateBufferStatus::AddressValidationFailed);
}

let ptr = address as *mut u8;

...
```

components/patina_mm/src/config.rs
```diff
@@ -171,61 +326,241 @@ impl CommunicateBuffer {
...
+    pub fn message_capacity(&self) -> usize {
+        self.len().saturating_sub(Self::MESSAGE_START_OFFSET)
+    }
+
+    /// Verifies that the internal state matches what is in the memory buffer.
+    /// This is intended to catch corruption and ensure the buffer actually matches what has
+    /// been requested through the MM Communication API.
+    ///
+    /// Returns `Ok(())` if state verification passes, otherwise returns the appropriate error.
+    fn verify_state_consistency(&self) -> Result<(), CommunicateBufferStatus> {
```
Javagedes, line 338: Regarding my suggestion above, I think zerocopy would make this checking easier.

components/patina_mm/src/config.rs
```diff
@@ -171,61 +326,241 @@ impl CommunicateBuffer {
...
+        if self.len() < Self::MESSAGE_START_OFFSET {
+            log::error!(target: "mm_comm", "Buffer {} is too small for the communicate header", self.id);
+            return Err(CommunicateBufferStatus::TooSmallForHeader);
+        }
+
+        let header_slice = &self.as_slice()[..Self::MESSAGE_START_OFFSET];
```
Javagedes, line 344: This will panic if the slice is not long enough, which makes multiple of the checks below invalid.

components/patina_mm/src/component/communicator.rs
```diff
@@ -138,32 +143,67 @@ impl Debug for MmCommunicator {
...
+            log::error!(target: "mm_comm", "Failed to set message data: {:?}", err);
+            Status::CommBufferInitError
+        })?;
+
+        log::debug!(target: "mm_comm", "Outgoing MM communication request: buffer_id={}, data_size={}, recipient={:?}", id, data_buffer.len(), recipient);
+        log::debug!(target: "mm_comm", "Request Data (hex): {:02X?}", &data_buffer[..core::cmp::min(data_buffer.len(), 64)]);
+        log::trace!(target: "mm_comm", "Comm buffer before request: {:?}", comm_buffer);
+
+        log::debug!(target: "mm_comm", "Triggering SW MMI for MM communication");
+        unsafe {
```
Javagedes, line 198: safety comment

components/patina_mm/src/config.rs
Javagedes: I would also consider the `NonNull<[u8]>` rather than a `NonNull<u8>`

components/patina_mm/src/config.rs
makubacki: I tried initially. The ultimate problem is that I haven't moved this code to the Patina GUID yet (see note in PR description) and `r_efi::base::Guid` doesn't implement the necessary zerocopy traits (like `IntoBytes`).

components/patina_mm/src/config.rs
makubacki: I have this staged locally with another set of commits that adds integration tests. I did not include those in this PR to reduce its overall size. These are used in the integration test.

I can drop it from this PR if you'd like, but I still needed `[allow(dead_code)]` even when it's used in the integration test.

components/patina_mm/src/config.rs
makubacki: Yeah, those were left around after a few refactors from using raw points before `Pin<&'static mut [u8]>`. I updated the first commit in the series to remove them. I also removed `unsafe` from the method so the `Safety` section is not required in the doc comment.

components/patina_mm/src/config.rs
makubacki: I agree. Updated `CommunicateBuffer::buffer` to be `NonNull<[u8]>` and made consequential changes for that which was not much since it's private.

components/patina_mm/src/config.rs
makubacki: Added the target info the module doc for this and the `sw_mmi_manager` to describe their targets.

components/patina_mm/src/config.rs
makubacki: Got rid of it.

components/patina_mm/src/config.rs
makubacki: I'm intending to inform the user what is expected regardless of internal validation detail. I'd still prefer they understand these requirements and not hit the errors cases.

I find that to be clearer but if convention is not to mention that, I'll drop the items being checked.

components/patina_mm/src/config.rs
makubacki: Agree. Updated to `Pin::new()` with a safety comment.

components/patina_mm/src/config.rs
makubacki: That's cleaner. Uses `usize::try_from()` with some slight adjustments for error propagation.

components/patina_mm/src/config.rs
makubacki: As I mentioned there, I'll revisit zerocopy after switching to the Patina GUID type.

components/patina_mm/src/config.rs
makubacki: If I understand your question correctly, that's what the slice length check just above was intended to check.

components/patina_mm/src/config.rs
makubacki: Moved them to `fmt::Display` implementation on their respective types.

components/patina_mm/src/component/communicator.rs
makubacki: Thanks, added.

components/patina_mm/src/config.rs
makubacki: Done

components/patina_mm/src/config.rs
Javagedes: I'm fine with keeping it.

components/patina_mm/src/config.rs
Javagedes: yes. Maybe I was looking at an outdated version of the PR or just had a brain fart, but all of this looks good now.

PR #829, opened 2025-10-03
Makefile.toml: Add `patina-test` task

makubacki:
## Description

Right now, `cargo make test` does not support common test arguments
like `-p`:

```
  error: invalid character `;` in package name: `;patina`, the first
  character must be a Unicode XID start character (most letters or `_`)

  Error while executing command, exit code: 101
```

It is useful to only against test in a given package during
development. This change proposes `cargo make test` behaving similar
to `cargo test` so users don't have to know when to switch between
the two and can use `cargo make` as their common command root.

Now something like this works:

> `cargo make test -p patina_mm -- --nocapture`

---

In the course of changing the test command, this is also fixed:

`enable_patina_tests` is a feature that is used to build test code
to execute on platforms. It can be thought of conceptually like
UEFI application testing on a given platform. It is not related
to unit testing on a host machine.

This change removes the feature from the `test` command entirely
and builds code with that feature under its own `patina-test`
task using the underlying `build` command.

Also, not all crates support `enable_patina_tests`. Previously,
you'd run into errors like this even if individual crates were
supported in the test command if the crate didn't support the
feature:

```
error: the package 'patina_adv_logger' does not contain this feature:
enable_patina_tests
```

That feature can be enabled against all workspace:

```
cargo make patina-test
```

The `all` task now includes both `test` (for host-based tests) and
`patina-tests` (for building the on-platform tests).

`patina-test` can be run against an individual package that supports
the feature. For example:

```
cargo make patina-test -p patina_macro
```

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- `cargo make test`
- `cargo make test -p patina_mm`
- `cargo make test -p patina_mm -- --nocapture`
- `cargo make patina-test`
- `cargo make patina-test -p patina_macro`
- `cargo make all`

## Integration Instructions

- No integration needed. But developers should review how each test command is used.

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -5,8 +5,7 @@ default_to_workspace = false
 RUSTC_BOOTSTRAP = 1
 NO_STD_FLAGS = "--profile ${RUSTC_PROFILE} -Zbuild-std=core,compiler_builtins,alloc -Zbuild-std-features=compiler-builtins-mem -Zunstable-options --timings=html"
 STD_FLAGS = "--profile ${RUSTC_PROFILE} --features std"
-TEST_FLAGS = { value = "--features enable_patina_tests", condition = { env_not_set = ["TEST_FLAGS"] } }
-COV_FLAGS = { value = "--features enable_patina_tests --workspace --lcov --profile test --ignore-filename-regex .*test.*", condition = { env_not_set = ["COV_FLAGS"] } }
+COV_FLAGS = { value = "--workspace --lcov --profile test --ignore-filename-regex .*test.*", condition = { env_not_set = ["COV_FLAGS"] } }
 RUSTDOCFLAGS = "-D warnings -D missing_docs"
 
 [env.development]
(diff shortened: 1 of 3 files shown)


Javagedes:
@vineelko @makubacki `enable_patina_test` is used to compile patina_test's during a UEFI build so that the on-system tests are compiled into the binary and executed. It should not be set for host-based unit tests, which is what `cargo make test` is for. I do not think TEST_FLAGS or COV_FLAGS should have ever had `--features enable_patina_tests` added to it.

Am I misunderstanding something?

makubacki:
> @vineelko @makubacki `enable_patina_test` is used to compile patina_test's during a UEFI build so that the on-system tests are compiled into the binary and executed. It should not be set for host-based unit tests, which is what `cargo make test` is for. I do not think TEST_FLAGS or COV_FLAGS should have ever had `--features enable_patina_tests` added to it.
> 
> Am I misunderstanding something?

That's my understanding. I want `cargo make test` to work as described in this PR. `cargo make patina-test` is to check those tests as shown. `cargo make all` to check that they compile with everything else.

Are you suggesting that they be moved out of a `test`-like command entirely?

Javagedes:
The `enable_patina_tests` should never be enabled or associated with the underlying `cargo test` command, which is what the new `cargo make patina-test` is continuing to do.

`cargo make patina-test` is still running host-based unit tests. But `--features enable_patina_tests` is a feature used when compiling in on-platform tests. Basically the `enable_patina_tests` feature flag is for UEFI_APPICATION-like testing on a physical platform, but you are enabling it when running HOST_APPLICATION-like tests with `cargo test`

If anything, `cargo make build-x64` / `cargo make build--aarch64` should be the commands that have that feature turned on. And the current state (`enable_patina_tests` enabled for `cargo make test` and `cargo make cov` should have never been a thing)

makubacki:
> > @vineelko @makubacki `enable_patina_test` is used to compile patina_test's during a UEFI build so that the on-system tests are compiled into the binary and executed. It should not be set for host-based unit tests, which is what `cargo make test` is for. I do not think TEST_FLAGS or COV_FLAGS should have ever had `--features enable_patina_tests` added to it.
> > Am I misunderstanding something?
> 
> That's my understanding. I want `cargo make test` to work as described in this PR. `cargo make patina-test` is to check those tests as shown. `cargo make all` to check that they compile with everything else.
> 
> Are you suggesting that they be moved out of a `test`-like command entirely?

This is going with what was already here. But, I think you're saying is that was a bug. `enable_patina_tests` is normal code, so use the normal build command in the `patina-test` task?

makubacki:
Okay. I'll fix that bug while at it.

Javagedes:
> This is going with what was already here. But, I think you're saying is that was a bug. enable_patina_tests is normal code, so use the normal build command in the patina-test task?

Yes :)

makubacki:
Updated.

Issue #816, opened 2025-10-02
[Bug]: `test_image_list_consistency` unit test re-entrant locks for "FsbLock" issue


makubacki:
### Is there an existing issue for this?

- [x] I have searched existing issues

### Current Behavior

Unit tests are currently failing occasionally with this error:

```
test tpl_lock::tests::tpl_mutex_should_change_tpl_if_bs_available ... ok

failures:

---- runtime::tests::test_image_list_consistency stdout ----

thread 'runtime::tests::test_image_list_consistency' panicked at patina_dxe_core/src/tpl_lock.rs:74:43:
Re-entrant locks for "FsbLock" not permitted.
stack backtrace:
   0:     0x55a0f73e9d92 - std::backtrace_rs::backtrace::libunwind::trace::h2d45396358f41939
                               at /rustc/1159e78c4747b02ef996e55082b704c09b970588/library/std/src/../../backtrace/src/backtrace/libunwind.rs:117:9
   1:     0x55a0f73e9d92 - std::backtrace_rs::backtrace::trace_unsynchronized::hffcefc0b67f1d6e2
                               at /rustc/1159e78c4747b02ef996e55082b704c09b970588/library/std/src/../../backtrace/src/backtrace/mod.rs:66:14
   2:     0x55a0f73e9d92 - std::sys::backtrace::_print_fmt::hd72f71d23b436b92
                               at /rustc/1159e78c4747b02ef996e55082b704c09b970588/library/std/src/sys/backtrace.rs:66:9
   3:     0x55a0f73e9d92 - <std::sys::backtrace::BacktraceLock::print::DisplayBacktrace as core::fmt::Display>::fmt::hdcfcb6d4c8489523
                               at /rustc/1159e78c4747b02ef996e55082b704c09b970588/library/std/src/sys/backtrace.rs:39:26
   4:     0x55a0f740f7d3 - core::fmt::rt::Argument::fmt::h2c56b3114963061a
                               at /rustc/1159e78c4747b02ef996e55082b704c09b970588/library/core/src/fmt/rt.rs:173:76
   5:     0x55a0f740f7d3 - core::fmt::write::h8a494366950f23bb
                               at /rustc/1159e78c4747b02ef996e55082b704c09b970588/library/core/src/fmt/mod.rs:1468:25
   6:     0x55a0f73e6d13 - std::io::default_write_fmt::h7b8824096454f323
[...]

PR #834, opened 2025-10-06
patina_dxe_core: Use global lock for runtime tests

cfernald:
## Description

The runtime tests use a static runtime allocator which in turn uses a TPL lock. This can cause panics if the locks run at in parallel and try to acquire the TPL lock at the same time.

Closes: #816 

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

cargo make test

## Integration Instructions

N/A

### DIFF
--- a/docs/src/SUMMARY.md
+++ b/docs/src/SUMMARY.md
@@ -2,6 +2,7 @@
 
 [Introduction](introduction.md)
 [Patina Background](patina.md)
+[RFC Lifecycle](rfc_lifecycle.md)
 [Code Organization](dev/code_organization.md)
 
 # Background Information
--- a/docs/src/background/rust_tools.md
+++ b/docs/src/background/rust_tools.md
@@ -6,6 +6,8 @@ The Patina project relies on Rust's tooling ecosystem to maintain consistent qua
 UEFI firmware development. This document summarizes the tooling in use, highlights how it is integrated into Patina's
 workflow, and contrasts the approach with common practices in C-based firmware projects where relevant.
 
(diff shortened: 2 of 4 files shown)


patina_dxe_core/src/runtime.rs
```diff
@@ -210,121 +211,127 @@ mod tests {
 
     #[test]
     fn test_image_list_consistency() {
-        let mut data = setup_protocol_and_data();
-        let link_offset = size_of::<u64>() * 4;
+        with_global_lock(|| {
```
makubacki, line 216: In think context in the code will help others.

```suggestion
        // Runtime tests require global synchronization due to shared static allocators
        // that use TPL locks, which cannot be acquired concurrently
        with_global_lock(|| {
```

patina_dxe_core/src/runtime.rs
```diff
@@ -210,121 +211,127 @@ mod tests {
...
+                    assert_eq!(entry.link.back_link, prev as *mut _);
+                    count += 1;
+                    protocol_link = entry.link.forward_link;
+                    prev = &entry.link as *const _;
+                    assert!(count <= 5, "Too many entries in the image list.");
+                }
+                assert_eq!(count, 5, "Not all entries were found in the image list.");
+            }
+        })
+        .unwrap();
```
makubacki: There is an opportunity to make the panic message more clear.

```suggestion
        .unwrap_or_else(|e| panic!("Test failed with runtime allocator conflict: {:?}", e));
```

PR #838, opened 2025-10-06
Signal events in reverse order of creation

joschock:
## Description

Presently Patina signals events in order of creation (assuming that they are signaled at the same time with the same notify TPL). However, the traditional EDK2 core signals events in reverse order of creation (i.e. the most recently created events are signaled before older events). This PR changes the signaling order to match EDK2 behavior. 

The UEFI spec does not specify what order events should be signaled if they are signaled at the same time and have the same notify TPL, but this aligns Patina with existing implementations for maximum compatibility. 

This PR also adds a test to verify the ordering. 

This PR also contains a minor refactor of the `Ord::cmp` implementation for `TaggedEventNotification`. This does not functionally change the comparison order, but cleans up the style to avoid redundant comparisons. 

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Existing and new unit tests pass, verified on platform hardware that ordering of platform event signaling now matches EDK2 core. 

## Integration Instructions

N/A

### DIFF
--- a/patina_dxe_core/src/event_db.rs
+++ b/patina_dxe_core/src/event_db.rs
@@ -165,12 +165,17 @@ impl PartialOrd for TaggedEventNotification {
 
 impl Ord for TaggedEventNotification {
     fn cmp(&self, other: &Self) -> Ordering {
-        if self.0.event == other.0.event {
-            Ordering::Equal
-        } else if self.0.notify_tpl == other.0.notify_tpl {
-            self.1.cmp(&other.1)
-        } else {
-            other.0.notify_tpl.cmp(&self.0.notify_tpl)
+        match self.0.event.cmp(&other.0.event) {
+            // events are unique - so if event ids match, they are the same event.
+            Ordering::Equal => Ordering::Equal,
+            _ =>
+            // Higher TPL has precedence; if TPLs are equal, then earlier insertion has precedence.
+            {
+                match other.0.notify_tpl.cmp(&self.0.notify_tpl) {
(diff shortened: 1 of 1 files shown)


patina_dxe_core/src/event_db.rs
```diff
@@ -165,12 +165,17 @@ impl PartialOrd for TaggedEventNotification {
...
-            self.1.cmp(&other.1)
-        } else {
-            other.0.notify_tpl.cmp(&self.0.notify_tpl)
+        match self.0.event.cmp(&other.0.event) {
+            // events are unique - so if event ids match, they are the same event.
+            Ordering::Equal => Ordering::Equal,
+            _ =>
+            //Higher TPL has precedence; if TPLs are equal, then earlier insertion has precedence.
+            {
+                match other.0.notify_tpl.cmp(&self.0.notify_tpl) {
```
Javagedes, line 174: you could technically do something like this:

```rust
fn cmp(&self, other: &Self) -> Ordering {
        if self.0.event == other.0.event {
            Ordering::Equal
        } else {
            // Higher TPL first, then earlier insertion
            (other.0.notify_tpl, self.1).cmp(&(self.0.notify_tpl, other.1))
        }
    }
```

But I'm honestly not sure that is much more legible haha. Do whichever you prefer, just providing an alternative :)

joschock: I'm going to stick with what I have - the construction you provided took extra brain power for me to understand ;)

PR #842, opened 2025-10-07
Generate lcov and html coverage reports

makubacki:
## Description

Currently, only lcov reports are generated by default. However, HTML reports provide nice, simple visualization of coverage data. While the HTML report can be generated manually today, it can also be generated efficiently in the `coverage` task without users having to understand (1) that the command is available and (2) how to use it to generate the HTML report.

Coverage data will be calculated once and then both reports will be generated from that data. Individual cargo-make tasks can be run as desired if only report type is needed in a given flow or development scenario. However, the main use case remains `cargo make coverage` which will now generate both reports.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- `cargo make coverage`
- `cargo make coverage-collect`
- `cargo make coverage-lcov`
- `cargo make coverage-html`

## Integration Instructions

- N/A - Impacts local repo developer operations

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -5,7 +5,7 @@ default_to_workspace = false
 RUSTC_BOOTSTRAP = 1
 NO_STD_FLAGS = "--profile ${RUSTC_PROFILE} -Zbuild-std=core,compiler_builtins,alloc -Zbuild-std-features=compiler-builtins-mem -Zunstable-options --timings=html"
 STD_FLAGS = "--profile ${RUSTC_PROFILE} --features std"
-COV_FLAGS = { value = "--workspace --lcov --profile test --ignore-filename-regex .*test.*", condition = { env_not_set = ["COV_FLAGS"] } }
+COV_FLAGS = { value = "--workspace --profile test --ignore-filename-regex .*test.*", condition = { env_not_set = ["COV_FLAGS"] } }
 RUSTDOCFLAGS = "-D warnings -D missing_docs"
 
 [env.development]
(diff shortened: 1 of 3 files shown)


Makefile.toml
```diff
@@ -150,14 +150,32 @@ private = true
...
-args = ["llvm-cov", "@@split(COV_FLAGS, )", "--output-path", "${CARGO_MAKE_WORKSPACE_WORKING_DIRECTORY}/target/lcov.info"]
+args = ["llvm-cov", "@@split(COV_FLAGS, )", "--no-report"]
 dependencies = ["individual-package-targets"]
 
+[tasks.coverage-lcov]
+description = "Generate an LCOV coverage report from collected data."
+install_crate = false
+clear = true
+command = "cargo"
+args = ["llvm-cov", "report", "--lcov", "--output-path", "${CARGO_MAKE_WORKSPACE_WORKING_DIRECTORY}/target/lcov.info"]
```
Javagedes, line 166: Should we consider moving this to `target/coverage/lcov.info` ?

Javagedes: Nevermind this is for tools like Code Gutters.

makubacki:
> Do we want to make all of the tasks except for `coverage` private, so that they cannot be run manually?

I was going to leave them accessible as I thought there is little downside to doing so. But I also don't know how realistic it is that people are going to use them. I'll let you decide.

Javagedes:
> > Do we want to make all of the tasks except for `coverage` private, so that they cannot be run manually?
> 
> I was going to leave them accessible as I thought there is little downside to doing so. But I also don't know how realistic it is that people are going to use them. I'll let you decide.

It's fine the way it is. we can always lock it down later

PR #843, opened 2025-10-07
Docs: Add Recommendation to Use Debugger Feature Flag

os-d:
## Description

After analysis with cargo-bloat, it has been determined that the debugger, even when disabled, adds 60k - 200k of file size to a platform, depending on the platform. This updates the patina docs to recommend a feature flag to enable the debugger in order to save this space when it is not in use.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [x] Includes documentation?

## How This Was Tested

N/A.

## Integration Instructions

N/A.

### DIFF
--- a/docs/src/dev/debugging.md
+++ b/docs/src/dev/debugging.md
@@ -44,14 +44,19 @@ for examples.
 Example setup:
 
 ```rust
+#[cfg(feature = "enable_debugger")]
 static DEBUGGER: patina_debugger::PatinaDebugger<UartPl011> =
     patina_debugger::PatinaDebugger::new(UartPl011::new(0x6000_0000))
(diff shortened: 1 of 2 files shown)


makubacki:
This is a great update to documentation, but that size bump is significant enough to impact fitting this into a firmware image with debug enabled.

os-d:
> This is a great update to documentation, but that size bump is significant enough to impact fitting this into a firmware image with debug enabled.

Agreed. I think this is an easy win for production FW, but we'll continue analyzing at a deeper level how to make this smaller for debug FW too. As you mentioned previously, the FV will still have to be large enough to accommodate this.

PR #844, opened 2025-10-07
Docs: Update Debugger Recommendations

os-d:
## Description

This updates the patina docs to reflect that it is recommended to have a PCR7 measurement reflecting that the debugger is enabled and that for FV size considerations, it is recommended that debug builds have the debugger enabled to accommodate the size needs easily.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [x] Includes documentation?

## How This Was Tested

N/A.

## Integration Instructions

N/A.

### DIFF
--- a/docs/src/dev/debugging.md
+++ b/docs/src/dev/debugging.md
@@ -45,18 +45,25 @@ Example setup:
 
 ```rust
 #[cfg(feature = "enable_debugger")]
+const _ENABLE_DEBUGGER: bool = true;
+#[cfg(not(feature = "enable_debugger"))]
+const _ENABLE_DEBUGGER: bool = false;
+
+#[cfg(feature = "build_debugger")]
 static DEBUGGER: patina_debugger::PatinaDebugger<UartPl011> =
     patina_debugger::PatinaDebugger::new(UartPl011::new(0x6000_0000))
         .without_transport_init()
-        .with_force_enabled(true);
+        .with_force_enabled(_ENABLE_DEBUGGER);
 ```
 
 Debugging configuration is critical to proper functionality. Read the [Patina Debugger documentation](https://github.com/OpenDevicePartnership/patina/blob/main/core/patina_debugger/src/debugger.rs)
 for full configuration options.
 
-> Note: It is recommended to use a compile time feature flag to enable/disable the debugger, including instantiating the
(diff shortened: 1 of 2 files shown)


docs/src/dev/debugging.md
```diff
@@ -44,19 +44,26 @@ for examples.
 Example setup:
 
 ```rust
+#[cfg(feature = "initial_breakpoint")]
+const _ENABLE_INITIAL_BREAKPOINT: bool = true;
+#[cfg(not(feature = "initial_breakpoint"))]
+const _ENABLE_INITIAL_BREAKPOINT: bool = false;
```
cfernald: This is misleading since with_force_enabled doesnt only control the initial breakpoint, but full enablement. and enable_debugger in this case will create the object but will not enable it in any way. I think for this to be affective we need to change the enablement story slightly.

os-d: Yeah, that's fair. I mostly wanted to enable three scenarios:
- Not instantiating the struct to save space on production builds
- Instantiating the struct in debug builds but not actually enabling the debugger in order to accommodate space in the FV with every debug build
- Instantiating the struct and enabling the debugger to break in at the initial breakpoint. This is nice to be able to do via CLI.

We can change the naming or the semantics of how we accomplish these three things.

kuqin12 (approved):

docs/src/dev/debugging.md
cfernald: I think a more appropriate leveling would be

feat debugger -> compiles the debugger in
feat debugger_enabled (dependent on debugger) -> sets force enabled.

I'm not fully convinced that have the debugger enabled but without the initial breakpoint is really the most ergonomic or intuitive way to operate.

os-d: That was an ask from the platform teams.

cfernald: So we would have 

> [features]
> debugger = []
> debugger_enabled = [debugger]
> debug_build_features = [debugger]

or whatever we want to call the debug_build_features

os-d: Updated.

cfernald (approved):

PR #845, opened 2025-10-07
patina_debugger: Remove initial breakpoint configurability.

cfernald:
## Description

The `configure` function in PatinaDebugger currently allows the platform to dynamically enable the debugger without the initial breakpoint. Currently this input is ignored as having the debugger enable in a silent configuration can be potentially dangerous for having the debugger enabled without notice in non-development scenarios. This removes this option and changes `configure` to `enable` for clarity.

Not marking as breaking as the configure function is not currently used.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

N/A

## Integration Instructions

N/A

### DIFF
--- a/core/patina_debugger/src/debugger.rs
+++ b/core/patina_debugger/src/debugger.rs
@@ -102,11 +102,7 @@ impl<T: SerialIO> PatinaDebugger<T> {
             log_policy: DebuggerLoggingPolicy::SuspendLogging,
             no_transport_init: false,
             exception_types: SystemArch::DEFAULT_EXCEPTION_TYPES,
-            config: spin::RwLock::new(DebuggerConfig {
-                enabled: false,
-                initial_break: false,
-                initial_break_timeout: 0,
-            }),
+            config: spin::RwLock::new(DebuggerConfig { enabled: false, initial_break: true, initial_break_timeout: 0 }),
             internal: Mutex::new(DebuggerInternal { gdb_buffer: None, gdb: None }),
             system_state: Mutex::new(SystemState::new()),
         }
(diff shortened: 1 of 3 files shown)


core/patina_debugger/src/debugger.rs
```diff
@@ -148,22 +144,16 @@ impl<T: SerialIO> PatinaDebugger<T> {
...
+    /// core is invoked.
     ///
     /// Enabled - Whether the debugger is enabled, and will install itself into the system.
     ///
-    /// Initial Break - Whether the debugger should break on initialization.
-    ///
-    /// Initial Break Timeout - A duration in seconds for the debugger to wait for a connection.
-    /// 0 indicates no timeout and will wait indefinitely
-    ///
-    pub fn configure(&self, enabled: bool, _initial_break: bool, _initial_break_timeout: u32) {
```
makubacki, line 162: I think this should at least be explicitly called out as a "design change" (the `type:design-change` GitHub label).

docs/src/dev/debugging.md
```diff
@@ -88,9 +88,8 @@ ERROR - ************************************
 This means the debugger is waiting for a connection. If you do not see this hang,
 then confirm that the debugger is enabled and installed prior to calling the core.
 
-You can also enable the debugger at runtime using the `Configure` routine, but use caution.
-Runtime enablement can skip the initial breakpoint and may cause security issues. For development,
-prefer force enablement.
+You can also enable the debugger at runtime using the `enable` routine, but use caution.
+Dynamic enablement should be carefully thought through to ensure proper platform security.
```
makubacki, line 92: Can you please end this so it's less open ended? Either point to an existing location talking about source debugger security considerations or briefly mention them here.

core/patina_debugger/src/debugger.rs
cfernald: Yeah, wasn't sure the scare required to really qualify as a "design change", added the label

docs/src/dev/debugging.md
cfernald: Fair enough, will do

cfernald: Done, added a section adding a bit more details.

PR #848, opened 2025-10-08
component params: Bugfix and additional testing

Javagedes:
## Description

While writing tests in this same commit, a bug was found in the param initialization / validation logic for ConfigMut<T>. The wrong panic message was being displayed when there was a conflict in parameter accesses for a param function. This commit fixes that bug.

This commit also increases test coverage for the param module.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

CI

## Integration Instructions

N/A

### DIFF
--- a/sdk/patina/src/component/params.rs
+++ b/sdk/patina/src/component/params.rs
@@ -413,21 +413,19 @@ unsafe impl<T: Default + 'static> Param for Config<'_, T> {
     fn init_state(storage: &mut Storage, meta: &mut MetaData) -> Self::State {
         let id = storage.add_config_default_if_not_present::<T>();
 
-        if meta.access().has_writes_all_configs() {
-            assert!(
-                !meta.access().has_config_write(id),
-                "Config<{0}> in component {1} conflicts with a previous &mut Storage access.",
-                core::any::type_name::<T>(),
-                meta.name(),
-            );
-        } else {
-            assert!(
-                !meta.access().has_config_write(id),
-                "Config<{0}> in component {1} conflicts with a previous ConfigMut<{0}> access.",
-                core::any::type_name::<T>(),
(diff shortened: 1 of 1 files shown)


sdk/patina/src/component/params.rs
```diff
@@ -413,17 +413,17 @@ unsafe impl<T: Default + 'static> Param for Config<'_, T> {
     fn init_state(storage: &mut Storage, meta: &mut MetaData) -> Self::State {
         let id = storage.add_config_default_if_not_present::<T>();
 
-        if !meta.access().has_writes_all_configs() {
+        if meta.access().has_writes_all_configs() {
```
os-d: There are a lot of asserts/panics in this file. I understand that failures here indicate an issue with the component inner workings, but is there not a graceful way to handle it and use debug_asserts to catch the internal logic errors?

sdk/patina/src/component/metadata.rs
makubacki, line 1: @Javagedes, my review will be delayed, but can you please hold the PR for me to submit a review?

sdk/patina/src/component/params.rs
Javagedes: I was actually just thinking about this.

These panics / asserts are only in code that is static. These checks make sure your component function parameters won't cause UB due to mutable aliasing. As a simple example, a function with two parameters `Config<u32>` and `ConfigMut<u32>` would be UB.

I think it is completely reasonable to switch to debug asserts for this. It will save some performance, but probably a decent amount of size due to the strings we create.

I can do this on another PR

Javagedes: #849

PR #851, opened 2025-10-09
add SMBIOS 3.x Configuration Table publication support

kat-perez:
## Description

Implement complete SMBIOS 3.0 table publication to UEFI

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Integrated with patina-qemu with q35 patina-dxe-core-qemu

## Integration Instructions

Same as above

### DIFF
--- a/components/patina_smbios/src/manager.rs
+++ b/components/patina_smbios/src/manager.rs
@@ -2181,4 +2181,218 @@ mod tests {
         let manager2 = SmbiosManager::new(255, 255);
         assert_eq!(manager2.version(), (255, 255));
     }
+
+    #[test]
+    fn test_smbios_handle_constants() {
+        // Test the reserved handle value
+        assert_eq!(SMBIOS_HANDLE_PI_RESERVED, 0xFFFE);
+    }
+
+    #[test]
+    fn test_smbios_error_clone_and_eq() {
+        let err1 = SmbiosError::InvalidParameter;
+        let err2 = err1.clone();
+        assert_eq!(err1, err2);
+
+        let err3 = SmbiosError::OutOfResources;
+        assert_ne!(err1, err3);
+    }
+
+    #[test]
+    fn test_smbios_record_builder_multiple_fields() {
+        let record = SmbiosRecordBuilder::new(2)
+            .add_field(0x12u8)
+            .add_field(0x3456u16)
+            .add_field(0x789ABCDEu32)
+            .build()
+            .expect("build failed");
+
+        // Check type
+        assert_eq!(record[0], 2);
+        // Verify fields are present (header + u8 + u16 + u32 = 4 + 1 + 2 + 4 = 11 bytes minimum)
+        assert!(record.len() >= 11);
+    }
+
+    #[test]
+    fn test_smbios_record_builder_with_multiple_strings() {
+        let record = SmbiosRecordBuilder::new(1)
+            .add_string(String::from("String1"))
+            .expect("add failed")
+            .add_string(String::from("String2"))
+            .expect("add failed")
+            .add_string(String::from("String3"))
+            .expect("add failed")
+            .build()
+            .expect("build failed");
+
+        // Verify record type
+        assert_eq!(record[0], 1);
+
+        // Verify strings are null-terminated and present
+        let record_str = core::str::from_utf8(&record[4..]).unwrap_or("");
+        assert!(record_str.contains("String1"));
+        assert!(record_str.contains("String2"));
+        assert!(record_str.contains("String3"));
+    }
+
+    #[test]
+    fn test_build_record_with_strings_empty_string() {
+        let header = SmbiosTableHeader::new(1, 4, SMBIOS_HANDLE_PI_RESERVED);
+        let strings = &[""];
+        let record = SmbiosManager::build_record_with_strings(&header, strings).expect("build failed");
+
+        // Empty string should still be null-terminated
+        assert_eq!(record[4], 0); // First string terminator
+        assert_eq!(record[5], 0); // Double null terminator
+    }
+
+    #[test]
+    fn test_add_from_bytes_with_max_handle() {
+        let manager = SmbiosManager::new(3, 9);
+
+        // Create a record with handle at max valid value
+        let header = SmbiosTableHeader::new(1, 4, 0xFEFF);
+        let record_bytes = SmbiosManager::build_record_with_strings(&header, &[]).expect("build failed");
+
+        let result = manager.add_from_bytes(None, &record_bytes);
+        assert!(result.is_ok());
+    }
+
+    #[test]
+    fn test_allocate_handle_with_gaps() {
+        let manager = SmbiosManager::new(3, 9);
+
+        // Add some records and get their actual handles
+        let header1 = SmbiosTableHeader::new(1, 4, SMBIOS_HANDLE_PI_RESERVED);
+        let bytes1 = SmbiosManager::build_record_with_strings(&header1, &[]).expect("build failed");
+        let _h1 = manager.add_from_bytes(None, &bytes1).expect("add failed");
+
+        let header2 = SmbiosTableHeader::new(1, 4, SMBIOS_HANDLE_PI_RESERVED);
+        let bytes2 = SmbiosManager::build_record_with_strings(&header2, &[]).expect("build failed");
+        let h2 = manager.add_from_bytes(None, &bytes2).expect("add failed");
+
+        let header3 = SmbiosTableHeader::new(1, 4, SMBIOS_HANDLE_PI_RESERVED);
+        let bytes3 = SmbiosManager::build_record_with_strings(&header3, &[]).expect("build failed");
+        let _h3 = manager.add_from_bytes(None, &bytes3).expect("add failed");
+
+        // Remove middle record
+        manager.remove(h2).expect("remove failed");
+
+        // Allocate a new handle - should reuse h2
+        let header4 = SmbiosTableHeader::new(1, 4, SMBIOS_HANDLE_PI_RESERVED);
+        let bytes4 = SmbiosManager::build_record_with_strings(&header4, &[]).expect("build failed");
+        let h4 = manager.add_from_bytes(None, &bytes4).expect("add failed");
+
+        assert_eq!(h4, h2); // Should reuse freed handle
+    }
+
+    #[test]
+    fn test_get_next_with_nonexistent_type() {
+        let manager = SmbiosManager::new(3, 9);
+
+        // Add a Type 1 record
+        let header = SmbiosTableHeader::new(1, 4, SMBIOS_HANDLE_PI_RESERVED);
+        let bytes = SmbiosManager::build_record_with_strings(&header, &[]).expect("build failed");
+        manager.add_from_bytes(None, &bytes).expect("add failed");
+
+        // Try to find Type 99 (doesn't exist)
+        let mut handle = 0xFFFF; // Start from beginning
+        let result = manager.get_next(&mut handle, Some(99));
+
+        assert_eq!(result, Err(SmbiosError::HandleNotFound));
+    }
+
+    #[test]
+    fn test_update_string_with_empty_string() {
+        let manager = SmbiosManager::new(3, 9);
+
+        // Add a record with strings
+        let header = SmbiosTableHeader::new(1, 4, SMBIOS_HANDLE_PI_RESERVED);
+        let bytes = SmbiosManager::build_record_with_strings(&header, &["Original"]).expect("build failed");
+        let handle = manager.add_from_bytes(None, &bytes).expect("add failed");
+
+        // Update to empty string
+        let result = manager.update_string(handle, 1, "");
+        assert!(result.is_ok());
+    }
+
+    #[test]
+    fn test_remove_last_record() {
+        let manager = SmbiosManager::new(3, 9);
+
+        // Add single record
+        let header = SmbiosTableHeader::new(1, 4, SMBIOS_HANDLE_PI_RESERVED);
+        let bytes = SmbiosManager::build_record_with_strings(&header, &[]).expect("build failed");
+        let handle = manager.add_from_bytes(None, &bytes).expect("add failed");
+
+        // Remove it
+        manager.remove(handle).expect("remove failed");
+
+        // Verify empty
+        assert_eq!(manager.records.borrow().len(), 0);
+    }
+
+    #[test]
+    fn test_validate_string_with_valid_characters() {
+        // Test with various valid ASCII characters
+        assert!(SmbiosManager::validate_string("Hello World 123!").is_ok());
+        assert!(SmbiosManager::validate_string("Valid-String_With.Symbols").is_ok());
+        assert!(SmbiosManager::validate_string("UPPERCASE").is_ok());
+        assert!(SmbiosManager::validate_string("lowercase").is_ok());
+    }
+
+    #[test]
+    fn test_smbios_table_header_fields() {
+        let header = SmbiosTableHeader::new(42, 100, 0x1234);
+
+        // Verify all fields - copy to avoid unaligned reference
+        let record_type = header.record_type;
+        let length = header.length;
+        let handle = header.handle;
+
+        assert_eq!(record_type, 42);
+        assert_eq!(length, 100);
+        assert_eq!(handle, 0x1234);
+    }
+
+    #[test]
+    fn test_calculate_checksum_zero() {
+        // Create entry point with all zeros
+        let entry_point = Smbios30EntryPoint {
+            anchor_string: [0, 0, 0, 0, 0],
+            checksum: 0,
+            length: 0,
+            major_version: 0,
+            minor_version: 0,
+            docrev: 0,
+            entry_point_revision: 0,
+            reserved: 0,
+            table_max_size: 0,
+            table_address: 0,
+        };
+
+        let checksum = SmbiosManager::calculate_checksum(&entry_point);
+        // Sum of zeros should give checksum of 0
+        assert_eq!(checksum, 0);
+    }
+
+    #[test]
+    fn test_smbios_record_with_string_count() {
+        let header = SmbiosTableHeader::new(1, 4, 10);
+        let data = vec![1, 4, 10, 0, 0, 0]; // Minimal data
+        let record = SmbiosRecord::new(header, None, data, 5);
+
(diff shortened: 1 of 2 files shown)


os-d:
@kat-perez looks like this is coming from a new base onto the feature branch, so you are bringing 210 commits, mostly newer commits from the mainline. That makes this challenging to review.

Can you either:
- Split this into two PRs, one that merges the mainline into the feature branch, the other that is new changes to the SMBIOS feature
- Only do the new changes to the SMBIOS feature and defer the mainline merge until later

Or, are you intending to target main from the feature branch?

kat-perez:
> @kat-perez looks like this is coming from a new base onto the feature branch, so you are bringing 210 commits, mostly newer commits from the mainline. That makes this challenging to review.
> 
> Can you either:
> 
> * Split this into two PRs, one that merges the mainline into the feature branch, the other that is new changes to the SMBIOS feature
> * Only do the new changes to the SMBIOS feature and defer the mainline merge until later
> 
> Or, are you intending to target main from the feature branch?

I will split this into two PRs. I had to merge main since I believe this would likely be the last PR before committing to main. I had to make sure these changes worked with main before publishing the PR that targets main from the feature branch

kat-perez:
@os-d this PR has been split now and only contains the relevant commits for the SMBIOS config table support

components/patina_smbios/src/component.rs
```diff
@@ -57,12 +62,13 @@ impl Default for SmbiosProviderManager {
 impl SmbiosProviderManager {
     /// Create a new SMBIOS provider manager with default SMBIOS 3.9 version
     pub fn new() -> Self {
-        Self { manager: SmbiosManager::new(3, 9) }
+        log::trace!("SMBIOS provider manager new");
```
os-d: I don't think entry/exit logs, even at trace level, are useful in production. I would recommend removing any.

components/patina_smbios/src/smbios_derive.rs
```diff
@@ -98,110 +113,117 @@ pub trait SmbiosRecords<'a> {
...
     minor_version: u8,
     #[allow(dead_code)]
     entry_point_64: Option<Box<Smbios30EntryPoint>>,
     #[allow(dead_code)]
     table_64_address: Option<PhysicalAddress>,
-    lock: Mutex<()>,
 }
 
 impl Clone for SmbiosManager {
+    fn clone(&self) -> Self {
```
os-d: In what situations does the SmbiosManager get cloned? I assumed it would be a single instance for all of boot?

components/patina_smbios/src/smbios_derive.rs
```diff
@@ -254,104 +276,51 @@ impl SmbiosManager {
...
-        // Validate each string
-        for string_bytes in &strings {
-            if string_bytes.is_empty() {
-                // Empty slice means consecutive nulls (invalid)
-                return Err(SmbiosError::InvalidParameter);
-            }
-            if string_bytes.len() > SMBIOS_STRING_MAX_LENGTH {
-                return Err(SmbiosError::StringTooLong);
+            // Find the next null terminator (end of current string)
+            while i < data_end && string_pool_area[i] != 0 {
```
os-d: I haven't made it all the way through yet, but I would expect that we would receive bytes from the protocol and as quickly as possible convert it to Strings (or str, perhaps) so we can use standard operations on those instead of continuing to parse strings as bytes. Does it make sense to do that before this point? Is my mental model flawed?

components/patina_smbios/src/smbios_derive.rs
```diff
@@ -254,104 +276,51 @@ impl SmbiosManager {
...
         }
 
         let mut record = Vec::new();
 
-        // Add the structured data using zerocopy
-        use zerocopy::IntoBytes;
-        record.extend_from_slice(header.as_bytes());
+        // Add the structured data
+        let header_bytes =
+            unsafe { core::slice::from_raw_parts(header as *const _ as *const u8, header.length as usize) };
```
os-d: This was zerocopy before, did that not work?

components/patina_smbios/src/smbios_derive.rs
```diff
@@ -368,65 +337,180 @@ impl SmbiosManager {
...
+        let total_table_size: usize = inner.records.iter().map(|r| r.data.len()).sum();
+
+        if total_table_size == 0 {
+            log::warn!("No SMBIOS records to install");
+            return Err(SmbiosError::InvalidParameter);
         }
 
-        // No freed handles available, use next_handle
-        let candidate = self.next_handle;
+        log::info!("Building SMBIOS table: {} records, {} bytes", inner.records.len(), total_table_size);
```
os-d: All of these prints are probably not that useful in production. Some information, say what table is being produced, etc., would be reasonable to print, but that x bytes was allocated, etc., probably not useful.

components/patina_smbios/src/smbios_derive.rs
```diff
@@ -368,65 +337,180 @@ impl SmbiosManager {
...
+
+        // Step 2: Allocate memory for the table (using UEFI Boot Services memory allocation)
+        let table_pages = total_table_size.div_ceil(4096);
```
os-d: uefi_size_to_pages!

components/patina_smbios/src/smbios_derive.rs
```diff
@@ -368,65 +337,180 @@ impl SmbiosManager {
...
+            let table_ptr = table_address as *mut u8;
+            let mut offset = 0;
 
-        // Check if we've exhausted the handle space
-        // Valid handles are 1..=0xFEFF (0xFFFE and 0xFFFF are reserved)
-        if candidate >= 0xFFFE {
-            // All handles exhausted
-            return Err(SmbiosError::OutOfResources);
+            for record in &inner.records {
+                core::ptr::copy_nonoverlapping(record.data.as_ptr(), table_ptr.add(offset), record.data.len());
```
os-d: zerocopy?

components/patina_smbios/src/smbios_derive.rs
```diff
@@ -368,65 +337,180 @@ impl SmbiosManager {
...
             return Err(SmbiosError::BufferTooSmall);
         }
 
-        // Step 2: Parse and validate header using zerocopy
-        use zerocopy::Ref;
-        let (header_ref, _rest) =
-            Ref::<&[u8], SmbiosTableHeader>::from_prefix(record_data).map_err(|_| SmbiosError::InvalidParameter)?;
-        let header: &SmbiosTableHeader = &header_ref;
+        // Step 2: Parse and validate header
+        let header = unsafe { &*(record_data.as_ptr() as *const SmbiosTableHeader) };
```
os-d: Same here, did zerocopy not work?

kat-perez:
> @kat-perez it looks like this branch might be stale? It is removing some of the changes you made in your last PR, like the zerocopy changes and the changes to optimize getting handles. Can you please check this is the intended set of changes? I think in the split some old state was kept accidentally. I have stopped my review partway through (but published the comments).

reworking, sorry I didn't realize I lost all of those. I was working off the last branch that I knew worked with the patina integration and then lost all the other changes from the other PR. will update you again when that's ready

kat-perez:
moving back into draft until I resolve the missing commits on the feature/smbios-component branch from PR #818

components/patina_smbios/src/smbios_derive.rs
kat-perez: updated

components/patina_smbios/src/smbios_derive.rs
kat-perez: reworked the entire PR, so this should have the zerocopy updates

components/patina_smbios/src/smbios_derive.rs
kat-perez: reworked the entire PR, so this should have the zerocopy updates

components/patina_smbios/src/smbios_derive.rs
kat-perez: removed all log::info

components/patina_smbios/src/smbios_derive.rs
kat-perez: i've reworked this to remove clone from SmbiosManager

components/patina_smbios/src/smbios_derive.rs
kat-perez: i think i lost this in a previous commit so i reworked it to have the changes from the previous PR

components/patina_smbios/src/smbios_derive.rs
kat-perez: made sure to have the zeroopy changes from last PR in here

components/patina_smbios/src/component.rs
```diff
@@ -55,78 +59,87 @@ impl Default for SmbiosProviderManager {
...
             Ok(handle) => {
-                log::info!("SMBIOS C protocol installed successfully at handle {:?}", handle);
+                log::info!("SMBIOS protocol installed successfully at handle {:?}", handle);
             }
             Err(e) => {
-                log::warn!("Failed to install SMBIOS C protocol: {:?}", e);
-                // Continue anyway - the Rust service will still work
+                log::error!("Failed to install SMBIOS protocol: {:?}", e);
+                // Cannot proceed without the manager - this is a fatal error
+                panic!("SMBIOS manager installation failed");
```
os-d: I don't think we should panic here. debug_assert and bail is reasonable, but in production we'd want to be able to continue if possible.

components/patina_smbios/src/manager.rs
```diff
@@ -82,104 +93,73 @@ pub trait SmbiosRecords<'a> {
...
-    ///
-    /// let record_bytes = bios_info.to_bytes();
-    /// let handle = smbios_records.add_from_bytes(None, &record_bytes)?;
-    /// ```
-    fn add_from_bytes(
-        &mut self,
-        producer_handle: Option<Handle>,
-        record_data: &[u8],
-    ) -> Result<SmbiosHandle, SmbiosError>;
+    fn add_from_bytes(&self, producer_handle: Option<Handle>, record_data: &[u8]) -> Result<SmbiosHandle, SmbiosError>;
```
os-d, line 105: Does it make sense to have the ability for a Rust component to be able to add an SMBIOS table without having to convert it to bytes? Say if the platform wanted to write their SMBIOS table generation in a Rust platform component. Is this SMBIOS component producing a service today? Or it just is handling the protocol? If there is no service today, we can defer this, though that would of course be the goal we want to go for.

components/patina_smbios/src/manager.rs
```diff
@@ -392,31 +348,158 @@ impl SmbiosManager {
...
+    /// - Creating mutable slices to allocated memory
+    /// - Writing the entry point structure to allocated memory
+    /// - Calling the UEFI `install_configuration_table` interface
+    ///
+    /// All memory allocations use UEFI Boot Services and are properly tracked by the firmware.
+    pub fn install_configuration_table(
+        &self,
+        boot_services: &patina::boot_services::StandardBootServices,
+    ) -> Result<(PhysicalAddress, PhysicalAddress), SmbiosError> {
+        use patina::boot_services::allocation::{AllocType, MemoryType};
```
os-d: We've typically been keeping use statements at the top of files, I'm not sure if we have a hard and fast style guideline there, but I think it makes the code easier to read.

components/patina_smbios/src/manager.rs
```diff
@@ -392,31 +348,158 @@ impl SmbiosManager {
...
+        let table_address = boot_services
+            .allocate_pages(
+                AllocType::AnyPage,
+                MemoryType::ACPI_RECLAIM_MEMORY, // SMBIOS tables go in ACPI Reclaim memory
+                table_pages,
+            )
+            .map_err(|_| SmbiosError::OutOfResources)?;
+
+        // Step 3: Copy all records to the table
+        unsafe {
```
os-d: Can zerocopy be used here?

If not and we have to keep the unsafe block, please add safety comments to this (and all unsafe blocks).

components/patina_smbios/src/manager.rs
```diff
@@ -681,6 +779,24 @@ static SMBIOS_PROTOCOL_INTERFACE: AtomicPtr<c_void> = AtomicPtr::new(core::ptr::
...
+/// `None` if `install_smbios_protocol` has not been called yet.
+///
+/// # Safety
+///
+/// This is safe because:
+/// - The manager pointer is only set once via `install_smbios_protocol`
+/// - Once set, it points to a leaked Box with 'static lifetime
+/// - The Mutex provides interior mutability and thread-safety
+pub fn get_global_smbios_manager() -> Option<&'static Mutex<SmbiosManager>> {
+    let ptr = SMBIOS_MANAGER.load(Ordering::Acquire);
```
os-d: I will note that @joschock is about to change this paradigm of using Atomic* in patina, so it may be worth using an UnsafeCell here, instead? Though he may be adding something new, so John, should this stay as is until you are talking atomics all up?

components/patina_smbios/src/manager.rs
```diff
@@ -903,9 +1019,11 @@ impl SmbiosProtocol {
             let manager_lock = manager.lock();
 
             match manager_lock.get_next(&mut handle, type_filter) {
-                Ok((header_ref, prod_handle)) => {
+                Ok((header_value, prod_handle)) => {
                     *smbios_handle = handle;
-                    *record = header_ref as *const SmbiosTableHeader as *mut SmbiosTableHeader;
+                    // Allocate the header on the heap and return a pointer to it
+                    // Note: This leaks memory, but matches the expected C API behavior
+                    *record = Box::into_raw(Box::new(header_value));
```
os-d: This will copy the header to a new location. Does the protocol have an expectation that a caller can call get_next and modify a header in place? Hopefully not

components/patina_smbios/src/manager.rs
```diff
@@ -970,14 +1092,14 @@ pub fn install_smbios_protocol(
...
-            log::info!("SMBIOS Protocol installed at handle {:?}", h);
             Ok(h)
         }
         Err(status) => {
             // Clean up on failure
             unsafe {
                 let _ = Box::from_raw(interface);
-                let _ = Box::from_raw(manager_ptr);
+                let manager_box = Box::from_raw(manager_ptr);
+                drop(manager_box); // Properly drop the manager
```
os-d: This seems odd to have two different approaches here (both of which I think should work). I would recommend unifying the approach unless there is a reason (in which case please comment) on why the different approaches are needed.

components/patina_smbios/src/component.rs
kat-perez: makes sense, removing the panic

components/patina_smbios/src/manager.rs
kat-perez: reworked to use zerocopy wherever possible in this file

components/patina_smbios/src/manager.rs
kat-perez: reworked my approach to use a static buffer instead. We're no longer leaking memory and makes sure callers shouldn't modify the returned header in place

components/patina_smbios/src/manager.rs
kat-perez: add_from_bytes handles adding records but i agree that the approach can be refined without converting to bytes and simply add the records. will rework that

kat-perez: the SMBIOS component is producing a service today and it adds records like this

        match smbios_service.add_from_bytes(None, &bios_bytes) {
            Ok(handle) => log::info!("  Type 0 (BIOS Info) - Handle 0x{:04X}", handle),
            Err(e) => log::warn!("  Failed to add Type 0: {:?}", e),
        }

kat-perez: platform code could do this instead:

```
        // Type 0: BIOS/Firmware Information
        let bios_info = Type0PlatformFirmwareInformation {
            header: SmbiosTableHeader { record_type: 0, length: 0, handle: SMBIOS_HANDLE_PI_RESERVED },
            vendor: 1,
            firmware_version: 2,
            bios_starting_address_segment: 0xE800,
            firmware_release_date: 3,
            firmware_rom_size: 0xFF, // 16MB
            characteristics: 0x08,   // BIOS characteristics
            characteristics_ext1: 0x03,
            characteristics_ext2: 0x03,
            system_bios_major_release: 1,
            system_bios_minor_release: 0,
            embedded_controller_major_release: 0xFF,
            embedded_controller_minor_release: 0xFF,
            extended_bios_rom_size: 0,
            string_pool: vec![String::from("Patina Firmware"), String::from("1.0.0"), String::from("10/15/2025")],
        };

        match smbios_service.add_record(None, &bios_info) {
            Ok(handle) => log::info!("  Type 0 (BIOS Info) - Handle 0x{:04X}", handle),
            Err(e) => log::warn!("  Failed to add Type 0: {:?}", e),
        }

```

kat-perez: also platform code would be responsible for publishing the table on ready_to_boot_event

components/patina_smbios/src/manager.rs
kat-perez: unified the approach

components/patina_smbios/src/manager.rs
kat-perez: i believe since zerocopy works with fixed-size types and we're concatenating a collection of variable-sized records, we would need to copy the actual data

components/patina_smbios/src/manager.rs
joschock: A couple of thoughts: 1) if these are truly write-once, and the write does not depend on a previous read, then they can be UnsafeCell<T>. 2) UEFI is single threaded, but it does have timer interrupts. So that means, for example, that someone could (in theory) access the SmbiosManger from a timer event callback while someone else is already holding the Mutex (which would result in DeadLock). 

So - that means that I don't think Atomics are doing much for you here anyway, and you can change them to UnsafeCell, but I think you probably want to get rid of `spin::Mutex` and use `TplMutex` from the patina SDK instead. That will allow you to set the TPL along with the Mutex, which ensures that you can't be interrupted by another thread at the same or lower TPL, allowing you to avoid the deadlock scenario.

kat-perez: that makes sense to me, thanks! i'm updating my implementation to use UnsafeCell

Issue #519, opened 2025-06-13
[Feature]: Stop Using Resource Descriptor HOB v1


os-d:
### Feature Overview

Once issue #518 has been resolved (or at least platforms have moved to resource descriptor HOB v2s entirely), Patina should ignore resource descriptor HOB v1s.

### Solution Overview

Ignore the v1 HOBs and exclusively use the v2 HOBs.

### Urgency

Low

### Are you going to implement the feature request?

Someone else needs to implement the feature

### Do you need maintainer feedback?

No maintainer feedback needed

os-d:
This is currently blocked on Intel platforms not using resc desc HOB v2.

@makubacki @berlin-with0ut-return @vineelko can one of you help drive converting the Intel platforms to use resc desc HOB v2s? I believe there may already be ongoing work there. That will unblock this task.

makubacki:
Yes. But all three of us are working on time sensitive and program critical work on Intel platforms. This is important, but it will have to come after that work is complete. I've left this unassigned for now so when those tasks are done, the next available person can pick it up.

HappyCoding-2025:
@os-d  @makubacki  @joschock  @Javagedes  @singweeMsftGitHub   - I am working on this based on discussion in today's meeting. 


Here is Scope of Changes
 1. Analyse where all Where Resource Descriptor v1 HOBs Are Currently Used?
2. Stop processing Resource Descriptor v1 HOBs entirely - Remove the conversion logic from v1 to v2
3. Only accept  Resource Descriptor v2 HOBs - Platforms must provide Resource Descriptor HOB v2s
4. Remove v1 compatibility code - Eliminate the ResourceDescriptorV2::from() conversion
5. Update test/example code - Change all test HOB creation to use v2 format

What improvement does this bring in ? 
1. Cache Management: Patina needs explicit cache attributes to properly manage memory protection and cache hierarchies
2. Security: Without proper cache attributes, Patina can't implement secure memory management
3. Simplification: Removes complexity of handling two different HOB formats
4. Future-proofing: Aligns with Patina's architectural requirements and upcoming PI spec changes
5. The change would essentially remove the backwards compatibility with traditional UEFI platforms that only provide v1 HOBs, requiring all platforms to be updated to provide the enhanced v2 HOBs with cache attribute information.

HappyCoding-2025:
[...]

PR #855, opened 2025-10-09
Feature : Convert ResourceDescriptor V1 HOBs to V2 format with cache attributes  Description:

HappyCoding-2025:
https://github.com/OpenDevicePartnership/patina/issues/519
## Breaking Change

- [x] Impacts functionality?
- [ ] Impacts security?
- [x] Breaking change?
- [x] Includes tests?
- [x] Includes documentation?

## Description

- This PR introduces breaking changes to ResourceDescriptor HOB processing.
- By Default it is V2 processing, All V1 processing should enable the feature flag v1_resource_descriptor_support
This feature converts the codebase from using ResourceDescriptor V1 HOBs to the newer V2 format, which includes cache attribute information for improved memory management and security.

1. Removed support for V1 ResourceDescriptor HOBs for security and architectural improvements
2. Updated HOB processing to only handle V2 HOBs
3. Enhanced cache attribute extraction logic to utilize V2 HOB cache information
4. Added documentation comments explaining the migration rationale
5. Added ResourceDescriptorV2 import
6. Updated test HOB list creation to use V2 format
7. Converted all resource descriptors in test functions to V2 format
8. Updated size calculations and copy operations for consistency
9. Applied cache attribute assignments matching production code

Testing -
Build completed fine and was tested in QEMU. 
image

## How This Was Tested

CI, QEMU

Integration documentation

### DIFF
--- a/.github/workflows/calculate-unsafe-code.yml
+++ b/.github/workflows/calculate-unsafe-code.yml
@@ -1,0 +1,25 @@
+# @file calculate-unsafe-code.yml
+#
+# A workflow that counts the amount of unsafe code in the repository with data broken
+# down into various categories.
+#
+##
+# Copyright (c) Microsoft Corporation.
+# SPDX-License-Identifier: Apache-2.0
+##
+name: Calcuate Unsafe Code
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+  schedule:
+    # Run daily at 6 AM UTC
+    - cron: '0 6 * * *'
+
+jobs:
+  calculate_unsafe_code:
+    name: Run
+    uses: OpenDevicePartnership/patina-devops/.github/workflows/CalculateUnsafeCode.yml@v0.0.3
+    secrets: inherit
--- a/.markdownlint.yml
+++ b/.markdownlint.yml
@@ -9,5 +9,5 @@
 {
     "default": true,
     "MD013": {"line_length": 120, "code_blocks": false, "tables": false},
-    "MD033": {"allowed_elements": ["br"]}
+    "MD033": {"allowed_elements": ["br", "center", "div"]}
 }
--- a/README.md
+++ b/README.md
@@ -1,4 +1,3 @@
-
 # Patina
 
 [![release]][_release]
@@ -10,6 +9,20 @@ This repository hosts the Patina project - a Rust implementation of UEFI firmwar
 
 The goal of this project is to serve as a replacement for core UEFI firmware components so they are written in Pure
 Rust as opposed to Rust wrappers around core implementation still written in C.
+
+---
+
+<center>
+<div align="center">
+
+Unsafe Code<br />
+
+[![overall_unsafe_code]][_overall_unsafe_code] [![fn_unsafe_code]][_fn_unsafe_code] [![expr_unsafe_code]][_expr_unsafe_code]
+
+[![impl_unsafe_code]][_impl_unsafe_code] [![traits_unsafe_code]][_traits_unsafe_code] [![methods_unsafe_code]][_methods_unsafe_code]
+
+</div>
+</center>
 
 ## Background
 
@@ -240,3 +253,16 @@ directory.
 [_ci]: https://github.com/OpenDevicePartnership/patina/actions/workflows/ci-workflow.yml
 [cov]: https://codecov.io/gh/OpenDevicePartnership/patina/graph/badge.svg?token=CWHWOUUGY6
 [_cov]: https://codecov.io/gh/OpenDevicePartnership/patina
+
+[overall_unsafe_code]: https://img.shields.io/endpoint?url=https://raw.githubusercontent.com/OpenDevicePartnership/patina/refs/heads/unsafe-code-badges/x86_64-unknown-uefi/badge_overall.json
+[_overall_unsafe_code]: https://img.shields.io/endpoint?url=https://raw.githubusercontent.com/OpenDevicePartnership/patina/refs/heads/unsafe-code-badges/x86_64-unknown-uefi/badge_overall.json
+[fn_unsafe_code]: https://img.shields.io/endpoint?url=https://raw.githubusercontent.com/OpenDevicePartnership/patina/refs/heads/unsafe-code-badges/x86_64-unknown-uefi/badge_functions.json
+[_fn_unsafe_code]: https://img.shields.io/endpoint?url=https://raw.githubusercontent.com/OpenDevicePartnership/patina/refs/heads/unsafe-code-badges/x86_64-unknown-uefi/badge_functions.json
+[expr_unsafe_code]: https://img.shields.io/endpoint?url=https://raw.githubusercontent.com/OpenDevicePartnership/patina/refs/heads/unsafe-code-badges/x86_64-unknown-uefi/badge_exprs.json
+[_expr_unsafe_code]: https://img.shields.io/endpoint?url=https://raw.githubusercontent.com/OpenDevicePartnership/patina/refs/heads/unsafe-code-badges/x86_64-unknown-uefi/badge_exprs.json
+[impl_unsafe_code]: https://img.shields.io/endpoint?url=https://raw.githubusercontent.com/OpenDevicePartnership/patina/refs/heads/unsafe-code-badges/x86_64-unknown-uefi/badge_item_impls.json
+[_impl_unsafe_code]: https://img.shields.io/endpoint?url=https://raw.githubusercontent.com/OpenDevicePartnership/patina/refs/heads/unsafe-code-badges/x86_64-unknown-uefi/badge_item_impls.json
+[traits_unsafe_code]: https://img.shields.io/endpoint?url=https://raw.githubusercontent.com/OpenDevicePartnership/patina/refs/heads/unsafe-code-badges/x86_64-unknown-uefi/badge_item_traits.json
+[_traits_unsafe_code]: https://img.shields.io/endpoint?url=https://raw.githubusercontent.com/OpenDevicePartnership/patina/refs/heads/unsafe-code-badges/x86_64-unknown-uefi/badge_item_traits.json
+[methods_unsafe_code]: https://img.shields.io/endpoint?url=https://raw.githubusercontent.com/OpenDevicePartnership/patina/refs/heads/unsafe-code-badges/x86_64-unknown-uefi/badge_methods.json
+[_methods_unsafe_code]: https://img.shields.io/endpoint?url=https://raw.githubusercontent.com/OpenDevicePartnership/patina/refs/heads/unsafe-code-badges/x86_64-unknown-uefi/badge_methods.json
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -6,6 +6,7 @@ components = ["rust-src", "clippy", "rustfmt", "rust-docs"]
 # A custom section for CI pipelines to install tools
 [tools]
 cargo-deny = "^0.17"
+cargo-geiger = "0.13.0"
 cargo-llvm-cov = "0.6.18"
 cargo-make = "0.37.21"
 cargo-release = "0.25.12"


HappyCoding-2025:
@makubacki  @joschock  @Javagedes  @singweeMsftGitHub  - Have finished implementing the feature support for V1->V2 support - Resource Descriptor Hob. 

Thanks @Javagedes for helping with the breaking change issues encountered while running in QEMU.

patina_dxe_core/src/gcd.rs
```diff
@@ -107,15 +107,15 @@ pub fn add_hob_resource_descriptors_to_gcd(hob_list: &HobList) {
         let mut resource_attributes: u32 = 0;
 
         let mut res_desc_op = None;
-        if let Hob::ResourceDescriptor(t_res_desc) = hob {
-            res_desc_op = Some(ResourceDescriptorV2::from(**t_res_desc));
-        } else if let Hob::ResourceDescriptorV2(t_res_desc) = hob {
+        // This removes support for v1 ResourceDescriptor HOBs for security and architectural improvements
+        if let Hob::ResourceDescriptorV2(t_res_desc) = hob {
```
os-d: Because Resource Descriptor V2 HOBs are new, I think it would be worthwhile to print that we are ignoring a V1 HOB at info level.

os-d: Also, I think this can be refactored to be something like the following flow:

If Hob::ResourceDescriptor {
    log::info!("Skipping v1 Resource Descriptor HOB {addr} {len}");
    continue;
} else if Hob::ResourceDescriptorV2 {
    res_desc_op = ...
} else {
   // not a resource descriptor HOB
   continue;
}

patina_dxe_core/src/gcd.rs
```diff
@@ -196,6 +196,8 @@ pub fn add_hob_resource_descriptors_to_gcd(hob_list: &HobList) {
 
         if gcd_mem_type != GcdMemoryType::NonExistent {
             let memory_attributes = {
```
os-d: This can then be refactored also, because we will always have memory attributes, since we always have V2 HOBs, so we don't need an if/else here or having the memory attributes be an option, we can just directly grab it.

I think we don't need the comment with the GH issue, this logic is already being done.

os-d: While we are cleaning up this section, instead of converting the u64 to MemoryAttributes and back to a u64, we can just always use the u64, e.g. res_desc.attributes & efi::CACHE_ATTRIBUTE_MASK, memory_attributes |= efi::MEMORY_RP.

patina_dxe_core/src/gcd.rs
HappyCoding-2025: Agreed. A log for ignoring V1 Hob at info level would be very helpful than silently ignoring this and if this is V2 , we shall process them. If this is not a resource descriptor HOB, it can continue.. Sure, will make the change.

patina_dxe_core/src/gcd.rs
HappyCoding-2025: YEs, we can simplify this as we always have V2 HOBS. simplification will make this cleaner. sure, will do this. We will remove the GH comment. Sure, we can always use u64. I will upload a new diff to handle the fixes to code review review.

HappyCoding-2025: @os-d Addressed all the code review comments and uploaded the new diff.

patina_dxe_core/src/gcd.rs
HappyCoding-2025: Addressed all the code review comments and uploaded the new diff @os-d

patina_dxe_core/src/gcd.rs
```diff
@@ -103,111 +102,99 @@ pub fn add_hob_resource_descriptors_to_gcd(hob_list: &HobList) {
...
+                )
+                .expect("Failed to add IO space to GCD");
+            }
+            _ => {
+                debug_assert!(false, "Unknown resource type in HOB");
+            }
+        };
+
+        if gcd_mem_type != GcdMemoryType::NonExistent {
+            assert!(res_desc.attributes_valid());
```
os-d: We should have this as a debug_assert. It is important to catch, but in production we can still move forward with 0 attributes, depending on what it is.

patina_dxe_core/src/gcd.rs
```diff
@@ -227,24 +214,27 @@ pub fn add_hob_resource_descriptors_to_gcd(hob_list: &HobList) {
...
+                    // NotReady is expected result here since page table is not yet initialized. In this case GCD
+                    // will be updated with the appropriate attributes which will then be sync'd to page table
+                    // once it is initialized.
+                    Err(EfiError::NotReady) => (),
+                    Ok(()) => {
+                        // Success is also acceptable - means attributes were set immediately
+                        log::debug!("Memory attributes set successfully for {:#X}", split_range.start);
+                    }
+                    Err(err) => {
+                        panic!(
```
os-d: I know this was a panic before, but let's also change this to a debug_assert. At this point, this is just attempting to update the GCD. If it were to fail, we'd want to catch during development, but a system could continue with a torn state here, this is the case in edk2.

HappyCoding-2025:
Dry run is failing for this and it is not reproducible on local system at all. Local system is clean. The Git Hub Helper suggested to take care of dry run issues by this https://github.com/OpenDevicePartnership/patina/pull/858

patina_dxe_core/src/gcd.rs
HappyCoding-2025: sure, will do

patina_dxe_core/src/gcd.rs
HappyCoding-2025: Sure, will do.

HappyCoding-2025: Pushed the new changes with review comments fixed @os-d

patina_dxe_core/src/gcd.rs
HappyCoding-2025: Pushed the new changes with review comments addressed @os-d

Javagedes:
> Dry run is failing for this and it is not reproducible on local system at all. Local system is clean. The Git Hub Helper suggested to take care of dry run issues by this #858

That solution is not valid. It is attempting to commit to your PR. If anything, it should be deleting the created files, not committing them. However, I suspect this occurs due to cargo vet failing its check. In that cause, the reason cargo vet failed (probably due to a crate being released) needs to be investigated and resolved.

makubacki:
> > Dry run is failing for this and it is not reproducible on local system at all. Local system is clean. The Git Hub Helper suggested to take care of dry run issues by this #858
> 
> That solution is not valid. It is attempting to commit to your PR. If anything, it should be deleting the created files, not committing them. However, I suspect this occurs due to cargo vet failing its check. In that cause, the reason cargo vet failed (probably due to a crate being released) needs to be investigated and resolved.

@Javagedes / @HappyCoding-2025, can you summarize what the issue is?

Javagedes:
> > > Dry run is failing for this and it is not reproducible on local system at all. Local system is clean. The Git Hub Helper suggested to take care of dry run issues by this #858
> > 
> > 
> > That solution is not valid. It is attempting to commit to your PR. If anything, it should be deleting the created files, not committing them. However, I suspect this occurs due to cargo vet failing its check. In that cause, the reason cargo vet failed (probably due to a crate being released) needs to be investigated and resolved.
> 
> @Javagedes / @HappyCoding-2025, can you summarize what the issue is?

@makubacki Since this repo has many library crates, not binary crates, we do not commit a Cargo.lock file. This means that every time we run CI, we generate a new Cargo.lock file, which means the most up to date dependencies are selected. Cargo vet runs on two different types of approvals: (1) approvals for specific versions of crates or (2) blanket approvals for authors. We have an intermixing of both in our cargo vet config files.

Ultimately what this boils down to is we randomly fail the cargo vet check whenever a new version of our dependencies is published to crates.io, and that dependency only has approvals for specific versions in our cargo vet config files.

This is the 4th occurrence of this in two days.

makubacki:
> Ultimately what this boils down to is we randomly fail the cargo vet check whenever a new version of our dependencies is published to crates.io, and that dependency only has approvals for specific versions in our cargo vet config files.
> 
> This is the 4th occurrence of this in two days.

Immediate change is here: https://github.com/OpenDevicePartnership/patina/pull/860

patina_dxe_core/examples/std.rs
makubacki, line 1: This PR needs to have the appropriate checkboxes set. Specifically, it is a breaking change and must be marked as breaking with clear integration instructions.

More broadly, this PR needs to be held until we confirm platforms already using Patina have fully transitioned to V2 HOBs. I will close this comment thread when that's the case.

HappyCoding-2025: @makubacki  - As discussed in last week's meeting .I created another feature request for supporting feature flag for supporting V1 for backward compatibility.  By default now it is V2 and V1 is possible when enabled with Flag. I have made additional changes to support this new feature.

patina_dxe_core/src/gcd.rs
```diff
@@ -88,6 +122,19 @@ pub fn init_paging(hob_list: &HobList) {
...
+    #[cfg(feature = "v1_resource_descriptor_support")]
+    {
+        log::warn!("RUST DXE CORE: V1 FEATURE IS ACTIVE!");
+        log::warn!("DXE Core: V1-ONLY ResourceDescriptor HOB processing active - modern V2 HOBs will be ignored");
+    }
+
+    #[cfg(not(feature = "v1_resource_descriptor_support"))]
+    {
+        log::warn!("RUST DXE CORE: V2 FEATURE IS ACTIVE!");
+        log::info!("DXE Core: V2 ResourceDescriptor HOB processing active with V1 backward compatibility");
```
os-d:
```suggestion
        log::info!("DXE Core: Only processing Resource Descriptor HOB V2s, ignoring V1.");
```
I don't think we need the warn log, this isn't a warning. Also, we don't want to process V1 HOBs at all in this configuration.

patina_dxe_core/src/gcd.rs
```diff
@@ -96,118 +143,148 @@ pub fn add_hob_resource_descriptors_to_gcd(hob_list: &HobList) {
         })
         .expect("Failed to find PHIT Hob");
 
+    log::warn!("RUST DXE CORE: PHIT HOB found successfully");
```
os-d: We can remove this, it's not a warning.

patina_dxe_core/src/gcd.rs
```diff
@@ -96,118 +143,148 @@ pub fn add_hob_resource_descriptors_to_gcd(hob_list: &HobList) {
...
+
     let free_memory_start = align_up(phit.free_memory_bottom, 0x1000).expect("Unaligned free memory bottom");
     let free_memory_size =
         align_down(phit.free_memory_top, 0x1000).expect("Unaligned free memory top") - free_memory_start;
 
     //Iterate over the hob list and map resource descriptor HOBs into the GCD.
     for hob in hob_list.iter() {
+        // Simple log to confirm we're processing HOBs
+        log::warn!("RUST DXE CORE: Processing a HOB in the list");
```
os-d: We can remove this, it's not a warning.

patina_dxe_core/src/gcd.rs
```diff
@@ -96,118 +143,148 @@ pub fn add_hob_resource_descriptors_to_gcd(hob_list: &HobList) {
...
+        //   - Processes ONLY V1 HOBs (legacy platform support)
+        //   - Skips V2 HOBs with warning (ensures V1-only behavior)
+        //   - No cache attributes support (V1 limitation)
+        //
+        // Benefits:
+        //   - Zero runtime overhead: No if/else branching in hot path
+        //   - Clean code paths: V1 and V2 logic completely separated
+        //   - Shared business logic: GCD operations remain common
+        //   - Future-proof: Easy to deprecate V1 support later
+        // =====================================================================
```
os-d: We don't need this large comment block, which is repeated from above.

patina_dxe_core/src/gcd.rs
```diff
@@ -96,118 +143,148 @@ pub fn add_hob_resource_descriptors_to_gcd(hob_list: &HobList) {
...
+
+        // Parse ResourceDescriptor HOB using feature-specific function
+        let (res_desc, cache_attributes) = if let Some(parsed) = parse_resource_descriptor_hob(hob) {
+            log::warn!("RUST DXE CORE: Found ResourceDescriptor HOB - processing it");
```
os-d: Don't need the warn log here, not a warning.

patina_dxe_core/src/gcd.rs
```diff
@@ -96,118 +143,148 @@ pub fn add_hob_resource_descriptors_to_gcd(hob_list: &HobList) {
...
-                    hob::EFI_RESOURCE_MEMORY_MAPPED_IO | hob::EFI_RESOURCE_FIRMWARE_DEVICE => {
-                        resource_attributes = res_desc.resource_attribute;
-                        gcd_mem_type = GcdMemoryType::MemoryMappedIo;
+        // =====================================================================
+        // Shared GCD Processing Logic (Common to V1 and V2)
+        // =====================================================================
+        // All memory mapping, attribute setting, and GCD operations below
+        // are identical regardless of HOB version, ensuring consistent
+        // behavior while minimizing code duplication.
+        // =====================================================================
```
os-d: Don't need this big comment block, that isn't needed info.

patina_dxe_core/src/gcd.rs
```diff
@@ -96,118 +143,148 @@ pub fn add_hob_resource_descriptors_to_gcd(hob_list: &HobList) {
...
+            // =====================================================================
+            // Memory Attributes Processing (Version-Agnostic)
+            // =====================================================================
+            // Extract cache attributes from the parsed HOB data:
+            // - V2 HOBs: cache_attributes contains actual cache settings
+            // - V1 HOBs: cache_attributes is 0 (no cache info available)
+            //
+            // This approach ensures consistent behavior regardless of HOB version
+            // while enabling cache attribute support when available.
+            // =====================================================================
```
os-d: Same here, this isn't needed

patina_dxe_core/src/gcd.rs
```diff
@@ -96,118 +143,148 @@ pub fn add_hob_resource_descriptors_to_gcd(hob_list: &HobList) {
...
+
+            // Extract cache attributes and add ReadProtect for system memory
```
os-d:
```suggestion
            // Extract cache attributes and add ReadProtect for system memory. If we are processing V1 HOBs, the cache attributes will be 0; that information is not passed.
```

patina_dxe_core/src/gcd.rs
```diff
@@ -96,118 +143,148 @@ pub fn add_hob_resource_descriptors_to_gcd(hob_list: &HobList) {
...
+            let mut memory_attributes = cache_attributes & efi::CACHE_ATTRIBUTE_MASK;
+            if gcd_mem_type == GcdMemoryType::SystemMemory {
+                // Force all system memory to be RP by default (since none is allocated yet)
+                // This applies to both V1 and V2 HOBs for security
```
os-d:
```suggestion
                // Force all system memory to be RP by default (since none is allocated yet)
```
The v1/v2 HOB decision doesn't need to permeate into all comment blocks. It should really just be a comment in one place.

patina_dxe_core/src/gcd.rs
```diff
@@ -329,6 +413,79 @@ pub(crate) fn activate_compatibility_mode() {
     crate::memory_attributes_protocol::uninstall_memory_attributes_protocol();
 }
 
+// ...existing code...
```
os-d: We don't need this comment

patina_dxe_core/src/gcd.rs
```diff
@@ -329,6 +413,79 @@ pub(crate) fn activate_compatibility_mode() {
...
+// - Future-proof: Easy to remove V1 support when no longer needed
+// =============================================================================
+
+/// Parse ResourceDescriptor HOB for V2 platforms (default, always compiled)
+///
+/// This function only processes V2 HOBs:
+/// - V2 HOBs: Preferred, provides cache attributes for optimal performance
+/// - V1 HOBs: Ignored completely (logs warning)
+///
+/// Returns: Some((ResourceDescriptor, cache_attributes)) or None if not a V2 resource descriptor
```
os-d: I think a lot of the comments in this PR need some work. They do not match our current style and read a bit more AI generated. Please review other function comments in this file and match that style. We don't need the architecture benefits section, e.g.

patina_dxe_core/src/gcd.rs
```diff
@@ -227,24 +304,31 @@ pub fn add_hob_resource_descriptors_to_gcd(hob_list: &HobList) {
...
+                match GCD.set_memory_space_attributes(
+                    split_range.start as usize,
+                    split_range.end.saturating_sub(split_range.start) as usize,
+                    memory_attributes,
+                ) {
+                    // NotReady is expected result here since page table is not yet initialized. In this case GCD
+                    // will be updated with the appropriate attributes which will then be sync'd to page table
+                    // once it is initialized.
+                    Err(EfiError::NotReady) => (),
+                    Ok(()) => {
```
os-d, line 231: Did this case actually arise? I think we should always get EfiError::NotReady here.

patina_dxe_core/src/gcd.rs
```diff
@@ -329,6 +413,79 @@ pub(crate) fn activate_compatibility_mode() {
...
+/// - V2 HOBs: Preferred, provides cache attributes for optimal performance
+/// - V1 HOBs: Ignored completely (logs warning)
+///
+/// Returns: Some((ResourceDescriptor, cache_attributes)) or None if not a V2 resource descriptor
+#[cfg(not(feature = "v1_resource_descriptor_support"))]
+fn parse_resource_descriptor_hob(hob: &Hob) -> Option<(hob::ResourceDescriptor, u64)> {
+    match hob {
+        Hob::ResourceDescriptorV2(v2_res_desc) => {
+            // V2 platforms: Only process V2 HOBs
+            log::info!(
```
os-d: We already print out all the HOBs, we don't need to print them again.

patina_dxe_core/src/gcd.rs
```diff
@@ -329,6 +413,79 @@ pub(crate) fn activate_compatibility_mode() {
...
+                "Processing V2 Resource Descriptor HOB at {:#x} length {:#x} with cache attributes {:#x}",
+                v2_res_desc.v1.physical_start,
+                v2_res_desc.v1.resource_length,
+                v2_res_desc.attributes
+            );
+            Some((v2_res_desc.v1, v2_res_desc.attributes))
+        }
+        Hob::ResourceDescriptor(_) => {
+            // V2 platforms: Ignore V1 HOBs completely
+            log::warn!("Ignoring V1 Resource Descriptor HOB on V2-only platform");
```
os-d: With my suggested change above mentioning we are skipping V1 HOBs, I don't think we have to print that for each V1 HOB.

patina_dxe_core/src/gcd.rs
```diff
@@ -329,6 +413,79 @@ pub(crate) fn activate_compatibility_mode() {
...
+
+/// Parse ResourceDescriptor HOB for V1 platforms (legacy platform support)
+///
+/// This function provides V1-only behavior for legacy platforms:
+/// - V1 HOBs: Processed normally without any migration suggestions
+/// - V2 HOBs: Completely ignored with warning message
+///
+/// Returns: Some((ResourceDescriptor, cache_attributes)) or None if not a V1 resource descriptor
+#[cfg(feature = "v1_resource_descriptor_support")]
+fn parse_resource_descriptor_hob(hob: &Hob) -> Option<(hob::ResourceDescriptor, u64)> {
```
os-d:
```suggestion
fn parse_resource_descriptor_hob(hob: &Hob) -> Option<(hob::ResourceDescriptor, Option<u64>)> {
```
I think we should return an option for the attributes. V1 HOBs return None, V2 HOBs return Some(attributes). That way, we can also validate if a V2 HOB returns Some(0), which is invalid.

patina_dxe_core/src/lib.rs
```diff
@@ -220,6 +220,13 @@ impl Core<NoAlloc> {
     pub fn init_memory(mut self, physical_hob_list: *const c_void) -> Core<Alloc> {
         log::info!("DXE Core Crate v{}", env!("CARGO_PKG_VERSION"));
 
+        // Log which ResourceDescriptor HOB processing mode is compiled
+        #[cfg(feature = "v1_resource_descriptor_support")]
+        log::warn!("DXE Core compiled with V1-ONLY ResourceDescriptor HOB support - V2 HOBs will be ignored!");
+
+        #[cfg(not(feature = "v1_resource_descriptor_support"))]
+        log::info!("DXE Core compiled with V2 ResourceDescriptor HOB support + V1 backward compatibility");
```
os-d: The whole change to this file isn't needed; we already print that when doing HOB processing.

patina_dxe_core/Cargo.toml
```diff
@@ -61,3 +61,21 @@ default = []
...
+#   - Log informational messages when V1 HOBs are encountered
+#   - V1 HOBs are processed for backward compatibility
+#
+# v1_resource_descriptor_support = true:
+#   - Process ONLY V1 ResourceDescriptor HOBs (legacy platforms)
+#   - Skip V2 HOBs completely with warning messages
+#   - No cache attributes support (V1 limitation)
+#
+# This design provides clean separation between V1 and V2 code paths
+# while maintaining shared GCD processing logic for both versions.
```
os-d: A comment is good here (and please also update the documentation to mention this), but this comment is very verbose and not accurate (V2 only doesn't also process V1 HOBs). Please condense this into a smaller, accurate comment.

HappyCoding-2025:
@makubacki  @Javagedes  @joschock  @os-d  - 
Tested on QEMU for V2, it boots fine to shell.

For V1,  ad @makubacki  suggested, had to revert V2 support https://github.com/OpenDevicePartnership/patina-qemu/pull/43 
locally on patina-qemu  and with this reverted  and V1 support on Qemu, It is booting fine to shell with patina supporting V1 with feature flag enabled.

image

patina_dxe_core/src/gcd.rs
HappyCoding-2025: Removed. WIll upload the new diff.

patina_dxe_core/src/gcd.rs
HappyCoding-2025: Removed. Will upload the new diff.

patina_dxe_core/src/gcd.rs
HappyCoding-2025: Removed and added a single comment.

HappyCoding-2025: Will upload the new code diff.

patina_dxe_core/src/gcd.rs
HappyCoding-2025: Fixed. will upload new diff.

patina_dxe_core/src/gcd.rs
HappyCoding-2025: Fixed. Will upload the new diff.

patina_dxe_core/src/gcd.rs
HappyCoding-2025: Fixed. Will upload new diff.

patina_dxe_core/Cargo.toml
HappyCoding-2025: Done, and updated Patina.md with updated documenttion. Will upload the new diff.

patina_dxe_core/src/lib.rs
HappyCoding-2025: Fixed. Will upload the new diff.

patina_dxe_core/src/gcd.rs
HappyCoding-2025: Pushed the fixes.  Have retained only info for V1 and V2 , it becomes easy for debugging that which dxe-core is built with.

HappyCoding-2025:
@makubacki  @Javagedes  @joschock  - could you pls check and review when free. Once this is merged, I would like to pick up another feature.

@os-d  - thanks for approving. I have taken care of all the review comments.

patina_dxe_core/src/gcd.rs
```diff
@@ -88,6 +122,17 @@ pub fn init_paging(hob_list: &HobList) {
...
+    #[cfg(feature = "v1_resource_descriptor_support")]
+    {
+        log::info!("RUST DXE CORE: V1 FEATURE IS ACTIVE!");
+    }
+
+    #[cfg(not(feature = "v1_resource_descriptor_support"))]
+    {
+        log::info!("RUST DXE CORE: V2 FEATURE IS ACTIVE!");
+    }
+
```
joschock: Recommend setting these to `debug` level and making them more specific (i.e. "v1_resource_descriptor feature is active")

patina_dxe_core/src/gcd.rs
```diff
@@ -329,6 +381,67 @@ pub(crate) fn activate_compatibility_mode() {
...
+        }
+        _ => None, // Not a resource descriptor HOB
+    }
+}
+
+/// Parse ResourceDescriptor HOB for V1 platforms (legacy platform support)
+///
+/// This function provides V1-only behavior for legacy platforms:
+/// - V1 HOBs: Processed normally without any migration suggestions
+/// - V2 HOBs: Completely ignored with warning message
```
joschock: comment indicates warning message that is not there in impl.

patina_dxe_core/src/test_support.rs
```diff
@@ -148,95 +148,125 @@ pub(crate) fn build_test_hob_list(mem_size: u64) -> *const c_void {
         reserved: Default::default(),
     };
 
-    let resource_descriptor1 = hob::ResourceDescriptor {
-        header: header::Hob {
-            r#type: hob::RESOURCE_DESCRIPTOR,
-            length: core::mem::size_of::<hob::ResourceDescriptor>() as u16,
-            reserved: 0x00000000,
+    // V2 HOBs use direct field assignments as per latest main branch
```
joschock: not sure what this comment means; consider removing or updating.

patina_dxe_core/Cargo.toml
```diff
@@ -61,3 +61,12 @@ default = []
 std = ["patina/std"]
 doc = ["patina_internal_cpu/doc"]
 compatibility_mode_allowed = []
+
```
joschock: recommend following the documentation style for `compatibility_mode_allowed` -> block comment here is not needed.

docs/src/patina.md
```diff
@@ -206,6 +206,20 @@ and depended on by those drivers are written in Rust.
 | ![UEFI Boot Services Call Count 1](./media/bootserv_call_table_1.png) | ![UEFI Boot Services Call Count 2](./media/bootserv_call_table_2.png)  |
 |---|---|
 
+#### Resource Descriptor HOB Version Support
```
joschock, line 209: This documentation fits better in docs/src/integrate/dxe_core.md section 9. Recommend moving there.

patina_dxe_core/src/gcd.rs
```diff
@@ -1,5 +1,40 @@
 //! DXE Core Global Coherency Domain (GCD)
 //!
+//! ## ResourceDescriptor HOB Version Support
```
joschock: Recommend removing in favor of documentation of the feature flag in docs/src/integrate/dxe_core.md section 9.

patina_dxe_core/src/gcd.rs
HappyCoding-2025: fixed and pushed new changes

patina_dxe_core/src/test_support.rs
HappyCoding-2025: fixed and pushed new changes

docs/src/patina.md
```diff
@@ -206,6 +206,20 @@ and depended on by those drivers are written in Rust.
...
+Patina DXE Core supports two mutually exclusive formats for Resource Descriptor HOBs: V1 (legacy) and V2 (modern,
+with cache attributes). The version supported is selected at compile time using a Cargo feature flag:
+
+- **Default (V2)**: Only V2 Resource Descriptor HOBs are processed. This is the default for modern platforms and
+  enables cache attribute support.
+- **Legacy (V1)**: If the `v1_resource_descriptor_support` feature is enabled, only V1 Resource Descriptor HOBs are
+  processed (for legacy platforms). V2 HOBs are ignored in this mode.
+
+The code paths for V1 and V2 are strictly separated at compile time for performance and maintainability. Shared GCD
+logic is reused for both modes. See the `Cargo.toml` and DXE Core source for details on enabling or disabling this
```
os-d, line 220: We should also include the details here (agree with John on moving to the integrate section), i.e. how you enable the feature.

docs/src/patina.md
HappyCoding-2025: moved to dxe_core.md. fixed and pushed new changes

patina_dxe_core/src/gcd.rs
os-d: Agree

patina_dxe_core/Cargo.toml
HappyCoding-2025: fixed and pushed new changes

patina_dxe_core/src/gcd.rs
```diff
@@ -88,6 +122,17 @@ pub fn init_paging(hob_list: &HobList) {
 }
 
 pub fn add_hob_resource_descriptors_to_gcd(hob_list: &HobList) {
+    // Test if V1 feature flag is working at all
```
os-d: This comment isn't needed, it looks like it was from development.

patina_dxe_core/src/gcd.rs
```diff
@@ -103,111 +148,111 @@ pub fn add_hob_resource_descriptors_to_gcd(hob_list: &HobList) {
...
-        let mut mem_range: Range<u64> = 0..0;
         let mut resource_attributes: u32 = 0;
 
-        let mut res_desc_op = None;
-        if let Hob::ResourceDescriptor(t_res_desc) = hob {
-            res_desc_op = Some(ResourceDescriptorV2::from(**t_res_desc));
-        } else if let Hob::ResourceDescriptorV2(t_res_desc) = hob {
-            res_desc_op = Some(**t_res_desc);
-        }
+        // ResourceDescriptor HOB processing: V1/V2 logic is separated by feature flags. See module-level docs above for details.
```
os-d: Don't need this comment with the docs moving. It is clear enough from the code.

docs/src/patina.md
HappyCoding-2025: Added the how to enable the feature.. this is the way I tested the code.  Wiill push the changes. Added this [features]
default = ["v1_resource_descriptor_support"]
v1_resource_descriptor_support = []

patina_dxe_core/src/gcd.rs
HappyCoding-2025: Correct, it was during my testing :-) Will remove and push new chagnes.

patina_dxe_core/src/gcd.rs
HappyCoding-2025: Fixed and pushed and also added to enable the feature.

patina_dxe_core/src/gcd.rs
HappyCoding-2025: Removed and pushed new changes.

patina_dxe_core/src/gcd.rs
HappyCoding-2025: Fixed and pushed new changes.

patina_dxe_core/examples/std.rs
```diff
@@ -109,95 +109,118 @@ fn build_hob_list() -> *const c_void {
...
 
-    let resource_descriptor6 = hob::ResourceDescriptor {
-        header: header::Hob {
-            r#type: hob::RESOURCE_DESCRIPTOR,
-            length: core::mem::size_of::<hob::ResourceDescriptor>() as u16,
-            reserved: 0x00000000,
+    let resource_descriptor6 = ResourceDescriptorV2 {
+        v1: hob::ResourceDescriptor {
+            header: header::Hob {
+                r#type: hob::RESOURCE_DESCRIPTOR,
```
makubacki: I think a V2 resource descriptor HOB type was intended?

```suggestion
                r#type: hob::RESOURCE_DESCRIPTOR2,
```

patina_dxe_core/examples/std.rs
```diff
@@ -109,95 +109,118 @@ fn build_hob_list() -> *const c_void {
...
 
-    let resource_descriptor7 = hob::ResourceDescriptor {
-        header: header::Hob {
-            r#type: hob::RESOURCE_DESCRIPTOR,
-            length: core::mem::size_of::<hob::ResourceDescriptor>() as u16,
-            reserved: 0x00000000,
+    let resource_descriptor7 = ResourceDescriptorV2 {
+        v1: hob::ResourceDescriptor {
+            header: header::Hob {
+                r#type: hob::RESOURCE_DESCRIPTOR,
```
makubacki:
```suggestion
                r#type: hob::RESOURCE_DESCRIPTOR2,
```

patina_dxe_core/examples/std.rs
HappyCoding-2025: Correct. Fixed.

patina_dxe_core/examples/std.rs
HappyCoding-2025: Correct. Fixed.

patina_dxe_core/src/gcd.rs
HappyCoding-2025: In the current implementation, we always get EfiError::NotReady in this case. The Ok(()) arm is present for completeness, but based on the initialization flow and current code, it is not expected to be reached. All observed test and runtime scenarios return EfiError::NotReady here, as the page table is not yet initialized at this point.

Including the Ok(()) match arm also ensures completeness and future-proofing—if the logic changes later and Ok(()) becomes possible, the code will handle it safely without panics or missed cases.

patina_dxe_core/src/gcd.rs
HappyCoding-2025: Fixed. Will push changes.

patina_dxe_core/src/gcd.rs
HappyCoding-2025: Removed. Will push the changes.

patina_dxe_core/src/gcd.rs
HappyCoding-2025: Added. WIll push changes.

patina_dxe_core/src/gcd.rs
HappyCoding-2025: Fixed and Will push the changes.

patina_dxe_core/src/gcd.rs
HappyCoding-2025: Fixed and will push changes.

patina_dxe_core/src/gcd.rs
HappyCoding-2025: Fixed and updated

os-d:
I should have removed my approval, that was from the original implementation. There were a couple of the comments that I think needed to be addressed in a different way, I'll put up a PR.

makubacki:
> I should have removed my approval, that was from the original implementation. There were a couple of the comments that I think needed to be addressed in a different way, I'll put up a PR.

I was also typing up a comment about some remaining duplication in the docs when the PR was merged. I assume you're aware of and will address that too.

PR #856, opened 2025-10-10
Enable NX for DXE Core stack and Set a guard page for the range with RP attributes

srilathasridharan:
## Description

Enable NX protection for DXE Core stack and set a guard page for the range with RP attributes.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Verified on a Surface arm based platform.

## Integration Instructions

N/A


patina_dxe_core/src/allocator.rs
```diff
@@ -957,6 +959,79 @@ fn process_hob_allocations(hob_list: &HobList) {
...
+        if (stack_address == 0) || (stack_length == 0) {
+            log::error!("Stack base address {:#X} for len {:#X}", stack_address, stack_length);
+            debug_assert!(false);
+        } else {
+            match GCD.get_memory_descriptor_for_address(stack_address) {
+                Ok(gcd_desc) => {
+                    let attributes: u64 = gcd_desc.attributes;
+                    log::trace!("Current Attributes for the stack {:#X?} \n\n", attributes);
+                    // Set Stack region to execute protect.
+                    if attributes != (attributes | (MemoryAttributes::ExecuteProtect).bits()) {
```
os-d: instead of MemoryAttributes::ExecuteProtect.bits(), I would use efi::MEMORY_XP. Same thing, simpler syntax, especially since we need the u64 version, not the MemoryAttributes version.

patina_dxe_core/src/allocator.rs
```diff
@@ -957,6 +959,79 @@ fn process_hob_allocations(hob_list: &HobList) {
...
+                        match GCD.set_memory_space_attributes(
+                            stack_address as usize,
+                            stack_length as usize,
+                            attributes | (MemoryAttributes::ExecuteProtect).bits(),
```
os-d: Same here.

patina_dxe_core/src/allocator.rs
```diff
@@ -957,6 +959,79 @@ fn process_hob_allocations(hob_list: &HobList) {
...
+                                );
+                                debug_assert!(false);
+                            }
+                        }
+                    }
+                    // Set Guard page to read protect.
+                    match GCD.set_memory_space_attributes(
+                        stack_address as usize,
+                        UEFI_PAGE_SIZE as usize,
+                        attributes | (MemoryAttributes::ReadProtect).bits(),
```
os-d: Same but efi::MEMORY_RP

patina_dxe_core/src/allocator.rs
```diff
@@ -957,6 +959,79 @@ fn process_hob_allocations(hob_list: &HobList) {
         };
     }
 
+    // Find the stack hob and set attributes.
+    if let Some(stack_hob) = hob_list.iter().find_map(|x| match x {
```
makubacki, line 962: Can you please check unit test coverage in the new code and add additional tests that may be needed?

makubacki:
This PR branch is quite a bit behind `main`. For example, it is referring to `patina` as `patina_sdk`. Some build tasks have also changed in latest.

patina_dxe_core/src/allocator.rs
srilathasridharan: I have added now.

patina_dxe_core/src/allocator.rs
```diff
@@ -1199,22 +1273,44 @@ mod tests {
...
+
+            // Locate stack hob.
+            if let Some(stack_hob) = hob_list.iter().find_map(|x| match x {
+                patina_pi::hob::Hob::MemoryAllocation(hob::MemoryAllocation { header: _, alloc_descriptor: desc })
+                    if desc.name == HOB_MEMORY_ALLOC_STACK =>
+                {
+                    Some(desc)
+                }
+                _ => None,
+            }) {
```
joschock: the stack hob should exist, so you don't need to use if let, you can unwrap() and fail the test if it doesn't

```suggestion
            let stack_hob = hob_list.iter().find_map(|x| match x {
                patina_pi::hob::Hob::MemoryAllocation(hob::MemoryAllocation { header: _, alloc_descriptor: desc })
                    if desc.name == HOB_MEMORY_ALLOC_STACK =>
                {
                    Some(desc)
                }
                _ => None,
            }).unwrap();
```

patina_dxe_core/src/allocator.rs
```diff
@@ -957,6 +958,79 @@ fn process_hob_allocations(hob_list: &HobList) {
...
+        {
+            Some(desc)
+        }
+        _ => None,
+    }) {
+        log::trace!("Found stack hob {:#X?} of length {:#X?}", stack_hob.memory_base_address, stack_hob.memory_length);
+        let stack_address = stack_hob.memory_base_address;
+        let stack_length = stack_hob.memory_length;
+
+        if (stack_address == 0) || (stack_length == 0) {
```
Javagedes, line 974: This function is humoungous, so I can't quite tell if this is applicable. But if possible, I would do an early return throughout this newly added code to prevent how nested this is getting. I would go as far as saying you should break this out into smaller functions just to allow you to do these early returns and prevent this amount of nesting. One example is this bit of code, but I can see it being applicable throughout. It makes it a bit easier to workout.

```rust
if stack_address == 0 || stack_length == 0 {
    log::error!("Stack base address {stack_address:#X} for len {stack_length:#X}");
    debug_assert!(false);
    return
}

Let Ok(gcd_desc) = match.get_memory_descriptor_for_address(stack_address) else {
    log::error!("Failed to get memory descriptor for address {:#x?} in GCD", stack_address);
    debug_assert!(false);
    return;
}

// so on and so on.
```

srilathasridharan: I agree, I considered having this in a separate function altogether ultimately undid that. I will take this feedback up and throw a separate PR to refactor this if that is alright.

PR #857, opened 2025-10-10
Beautify the Exception Handler Processing

os-d:
## Description

This change makes several updates to the exception handlers for x64 and aarch64:
- Add an AARCH64 synchronous exception handler to dump more granular information, particularly about instruction/data aborts
- Print register dumps in arch specific and generic handlers more concisely, printing multiple registers per line and only useful registers (except at debug error level)
- Dump the page table for the failing address in a page fault instead only using query; this provides much more information.

A couple notes:
- This is somewhat subjective, so happy to take additional feedback
- Doing log::error(""); between lines instead of `\n` inside of another print makes the log look better, with adv logger, because the error string gets appended on the front of the former, but not the latter, making the output look strange.
- Other ARM64 synchronous exception types can be added, but instruction/data aborts are some of the most common and there is easily obtainable information that is useful to retrieve. As time goes on, we could certainly extend it. I am not attempting to replace existing external software that translates the ESR for you. Just getting information we can't otherwise get.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested by injecting faults on Q35 and SBSA and seeing the exception messages.

## Integration Instructions

N/A.

### DIFF
--- a/core/patina_internal_cpu/src/interrupts.rs
+++ b/core/patina_internal_cpu/src/interrupts.rs
@@ -72,6 +72,46 @@ cfg_if::cfg_if! {
 /// specific interrupt type ID.
 pub type ExceptionType = usize;
 
+/// This macro pretty prints registers in groups of four per line.
+/// The expected input is a list of name, value pairs.
+#[macro_export]
+macro_rules! log_registers {
+    ( $( $name:expr, $value:expr ),+ $(,)? ) => {
+        let registers = [$(($name, $value),)+];
+        for chunk in registers.chunks(4) {
+            match chunk {
+                [c1, c2, c3, c4] => {
+                    log::error!(
+                        "{:>4}:  {:#018X}   {:>4}:  {:#018X}   {:>4}:  {:#018X}   {:>4}:  {:#018X}",
+                        c1.0, c1.1, c2.0, c2.1, c3.0, c3.1, c4.0, c4.1
+                    );
+                },
+                [c1, c2, c3] => {
+                    log::error!(
+                        "{:>4}:  {:#018X}   {:>4}:  {:#018X}   {:>4}:  {:#018X}",
+                        c1.0, c1.1, c2.0, c2.1, c3.0, c3.1
+                    );
+                },
+                [c1, c2] => {
+                    log::error!(
+                        "{:>4}:  {:#018X}   {:>4}:  {:#018X}",
+                        c1.0, c1.1, c2.0, c2.1,
+                    );
+                },
+                [c1] => {
+                    log::error!(
+                        "{:>4}:  {:#018X}",
+                        c1.0, c1.1
+                    );
+                },
+                _ => {
+                    log::error!("");
+                }
+            }
+        }
+    };
+}
+
 /// Trait for converting the architecture specific context structures into the
 /// UEFI System Context structure.
 pub(crate) trait EfiSystemContextFactory {
@@ -83,6 +123,9 @@ pub(crate) trait EfiSystemContextFactory {
 pub(crate) trait EfiExceptionStackTrace {
     /// Dump the stack trace for architecture specific context.
     fn dump_stack_trace(&self);
+
+    /// Dump system context registers for architecture specific context.
+    fn dump_system_context_registers(&self);
 }
 
 /// Trait for structs that implement and manage interrupts.
--- a/core/patina_internal_cpu/src/interrupts/aarch64.rs
+++ b/core/patina_internal_cpu/src/interrupts/aarch64.rs
@@ -6,14 +6,16 @@
 //!
 //! SPDX-License-Identifier: Apache-2.0
 //!
+use crate::log_registers;
 use patina::error::EfiError;
 use patina_pi::protocols::cpu_arch::EfiSystemContext;
 use patina_stacktrace::StackTrace;
 
+mod sysreg;
+
 cfg_if::cfg_if! {
     if #[cfg(all(target_os = "uefi", target_arch = "aarch64"))] {
         mod interrupt_manager;
-        mod sysreg;
         pub mod gic_manager;
         pub use interrupt_manager::InterruptsAarch64;
         use core::arch::asm;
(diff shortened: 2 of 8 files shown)


core/patina_internal_cpu/src/interrupts/aarch64.rs
```diff
@@ -38,6 +39,106 @@ impl super::EfiExceptionStackTrace for ExceptionContextAArch64 {
...
         }
     }
+
+    fn dump_system_context_registers(&self) {
+        log::error!("ESR: {:#X?}, ELR: {:#X?}, SPSR: {:#X?}, FAR: {:#X?}", self.esr, self.elr, self.spsr, self.far);
+
+        log::error!("");
+
+        log::error!("General-Purpose Registers");
+        log::error!(
```
cfernald: could we condense this into a loop or a subroutine to cut on duplication?

core/patina_internal_cpu/src/interrupts/aarch64/interrupt_manager.rs
```diff
@@ -130,3 +142,157 @@ fn initialize_exception() -> Result<(), EfiError> {
...
+        log::error!("Page Fault (Instruction Abort)");
+    } else if ec == 0x24 || ec == 0x25 {
+        // Data Abort from a lower EL or same EL
+        log::error!("Page Fault (Data Abort)");
+    }
+
+    log::error!("");
+
+    log::error!("General-Purpose Registers");
+    log::error!(
```
cfernald: isnt this the point of dump_system_context_registers?

core/patina_internal_cpu/src/interrupts/aarch64/interrupt_manager.rs
```diff
@@ -130,3 +142,157 @@ fn initialize_exception() -> Result<(), EfiError> {
...
+    }
+
+    log::debug!("Full Context: {aarch64_context:#X?}");
+
+    log::error!("Dumping Exception Stack Trace:");
+    if let Err(err) = unsafe { StackTrace::dump_with(aarch64_context.elr, aarch64_context.sp) } {
+        log::error!("StackTrace: {err}");
+    }
+
+    // no need to print anything here, we already did
```
cfernald: I'm not sure this is going to always be true. Panic is a point of platform extension and they may have many things they do with panic messages that could be useful. e.g. logging to SEL, force enable logging before printing, etc.. I think it's better to keep the redundant print for that reason.

core/patina_internal_cpu/src/interrupts/exception_handling.rs
```diff
@@ -107,13 +108,22 @@ extern "efiapi" fn exception_handler(exception_type: usize, context: &mut Except
...
         }
     }
 }
 
+#[allow(dead_code)]
+pub(crate) struct FaultAllocator {}
+
+impl PageAllocator for FaultAllocator {
+    fn allocate_page(&mut self, _align: u64, _size: u64, _contiguous: bool) -> patina_paging::PtResult<u64> {
+        unimplemented!()
```
cfernald: Might better to always return an error. We don't want a nested panic in the exception handlers.

core/patina_internal_cpu/src/interrupts/x64/interrupt_manager.rs
```diff
@@ -118,200 +116,232 @@ impl InterruptManager for InterruptsX64 {}
...
-                RBP: {:x?}\n \
-                R8: {:x?}\n \
-                R9: {:x?}\n \
-                R10: {:x?}\n \
-                R11: {:x?}\n \
-                R12: {:x?}\n \
-                R13: {:x?}\n \
-                R14: {:x?}\n \
-                R15: {:x?}",
+        "{:>4}  {:#018X}   {:>4}  {:#018X}   {:>4}  {:#018X}  {:>4}  {:#018X}",
```
cfernald: ditto with using dump_system_context_registers

core/patina_internal_cpu/src/interrupts/x64/interrupt_manager.rs
```diff
@@ -118,200 +116,232 @@ impl InterruptManager for InterruptsX64 {}
...
+
+    dump_pte(x64_context.cr2, x64_context.cr3, paging_type);
+
+    log::error!("Dumping Exception Stack Trace:");
     if let Err(err) = unsafe { StackTrace::dump_with(x64_context.rip, x64_context.rsp) } {
         log::error!("StackTrace: {err}");
     }
 
-    panic!("EXCEPTION: PAGE FAULT");
+    // no need to print anything here, we already did
```
cfernald: ditto with the panic message

core/patina_internal_cpu/src/interrupts/aarch64/interrupt_manager.rs
os-d: Yes, originally there was some problem using that because of where it was placed, but I think the final result can be used. I'll switch over.

core/patina_internal_cpu/src/interrupts/aarch64/interrupt_manager.rs
os-d: Fair enough. I'll leave it.

core/patina_internal_cpu/src/interrupts/aarch64.rs
os-d: Sure.

core/patina_internal_cpu/src/interrupts/exception_handling.rs
os-d: Agreed. This was just copied from what was already done in the x86 side. I'll update it.

core/patina_internal_cpu/src/interrupts/aarch64.rs
joschock: or a macro

os-d: Made a macro

core/patina_internal_cpu/src/interrupts/aarch64/interrupt_manager.rs
os-d: Switched all instances.

core/patina_internal_cpu/src/interrupts.rs
```diff
@@ -72,6 +72,39 @@ cfg_if::cfg_if! {
 /// specific interrupt type ID.
 pub type ExceptionType = usize;
 
+#[macro_export]
+macro_rules! log_registers {
```
vineelko, line 78: Alternate implementation 1: 🦀🦀
```rust
macro_rules! log_registers {
    ( $( $name:expr, $value:expr ),+ $(,)? ) => {
        let registers = [$(($name, $value),)+];
        for chunk in registers.chunks(4) {
            let line = chunk
                .iter()
                .map(|(name, value)| format!("{:>4}  {:#018X}", name, value))
                .collect::<Vec<_>>()
                .join("   ");
            log::error!("{}", line);
        }
    };
}
```
To avoid any allocations due to `format!()/join()` calls. 
Implementation 2: 🦀
```rust
macro_rules! log_registers {
    ( $( $name:expr, $value:expr ),+ $(,)? ) => {
        let registers = [$(($name, $value),)+];
        for chunk in registers.chunks(4) {
            match chunk {
                [c1, c2, c3, c4] => {
                    log::error!(
                        "{:>4}  {:#018X}   {:>4}  {:#018X}   {:>4}  {:#018X}   {:>4}  {:#018X}",
                        c1.0, c1.1, c2.0, c2.1, c3.0, c3.1, c4.0, c4.1
                    );
                },
                [c1, c2, c3] => {
                    log::error!(
                        "{:>4}  {:#018X}   {:>4}  {:#018X}   {:>4}  {:#018X}",
                        c1.0, c1.1, c2.0, c2.1, c3.0, c3.1
                    );
                },
                [c1, c2] => {
                    log::error!(
                        "{:>4}  {:#018X}   {:>4}  {:#018X}",
                        c1.0, c1.1, c2.0, c2.1,
                    );
                },
                [c1] => {
                    log::error!(
                        "{:>4}  {:#018X}",
                        c1.0, c1.1
                    );
                },
                _ => {
                    log::error!("");
                }
            }
        }
    };
}
```
Not a lot of improvement than your implementation but slightly more 🦀.  Not tested!

os-d: Take it up with copilot ;). Yeah, we can't have any memory allocations here (exception handler, system is dead), but I'll update to option two, I agree its cleaner

Issue #511, opened 2025-06-13
[Review: Review GetMemoryMap Implementation


os-d:
### Is there an existing issue for this?

- [x] I have searched existing issues

### Current Behavior

Patina's get_memory_map function has some differences in behavior from edk2 which should be reviewed.

### Expected Behavior

Consistent with UEFI spec expectations and deviations from edk2 understood and intentional.

### Steps To Reproduce

n/a

### Build Environment

```markdown
- OS(s):
- Tool Chain(s):
- Targets Impacted:
```

### Version Information

```text
top of tree
```

### Urgency

Medium

### Are you going to fix this?

I will fix it

### Do you need maintainer feedback?

No maintainer feedback needed

os-d:
@kuqin12 the only part I did not fully review was the recent memory bucket work you did in edk2. Can you please review if Patina needs any updates? I don't believe so, because the buckets are completely owned by the allocators and so should be grouped correctly.

PR #862, opened 2025-10-10
Align EFI_MEMORY_MAP Generation with edk2

### LINKED ISSUE #859: [Bug]: A large amount of main memory may be consumed in the OS

### Current Behavior

It has been observed that a large of "Secure System" memory may be consumed in the OS. This needs to be triaged in this issue.

### Expected Behavior

A large amount of system memory is not consumed.

### Steps To Reproduce

Boot the patina-v11.2.0 release on a physical Intel platform.

### Build Environment

```markdown
N/A
```

### Version Information

```text
Tag: patina-v11.2.0
```

### Urgency

Medium

Someone else needs to fix it

No maintainer feedback needed

### Anything else?

For reference:

Image

os-d:
## Description

OSes have taken a hard dependence on the EFI_MEMORY_MAP as created by edk2. As such, this aligns Patina's
EFI_MEMORY_MAP generation with edk2's.

Namely this:
- Only returns runtime MMIO in the map. Without this large MMIO ranges can cause OSes to needlessly map giant memory regions, causing them to build large PFN databases and consuming a lot of memory.
- Drop checking for ISA_VALID and setting EFI_MEMORY_PORT_IO_SPACE. Both of these are for IPF only, which is not supported in Patina.
- Set EFI_MEMORY_NV for persistent memory.

Closes #859 and #511 .

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on a platform where Windows was consuming large amounts of memory. After this change it no longer does.

## Integration Instructions

N/A.

### DIFF
--- a/patina_dxe_core/src/allocator.rs
+++ b/patina_dxe_core/src/allocator.rs
@@ -644,11 +644,12 @@ pub(crate) fn get_memory_map_descriptors(active_attributes: bool) -> Result<Vec<
 
                     // MMIO. Note: there could also be MMIO tracked by the allocators which would not hit this case.
                     GcdMemoryType::MemoryMappedIo => {
-                        if (descriptor.attributes & efi::MEMORY_ISA_VALID) == efi::MEMORY_ISA_VALID {
-                            Some(efi::MEMORY_MAPPED_IO_PORT_SPACE)
-                        } else {
-                            Some(efi::MEMORY_MAPPED_IO)
+                        // we should only be returning runtime MMIO here
+                        if descriptor.attributes & efi::MEMORY_RUNTIME == 0 {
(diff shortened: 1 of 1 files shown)


cfernald:
If you don't mind holding this PR, I want to understand this issue better before reviewing

cfernald:
FYI, I linked the other both, since I dont think the link tagging works with "and"

os-d:
Tested on two generations of Intel platforms and confirmed boot and that the memory issue was not present. Merging as code coverage for GCD/memory management is being refactored separate from this.

PR #878, opened 2025-10-14
RFC: ESRT

kat-perez:
## Description

Status: Rejected

Proposes a new EsrtRecords service for managing the EFI System Resource Table (ESRT) with Rust-native types and memory safety guarantees.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [x] Includes documentation?

## How This Was Tested

N/A

## Integration Instructions

N/A

### DIFF
--- a/docs/src/rfc/rejected/0023-esrt.md
+++ b/docs/src/rfc/rejected/0023-esrt.md
@@ -1,0 +1,488 @@
+# RFC: `ESRT (EFI System Resource Table)`
+
+This RFC proposes a Rust-based interface for managing the EFI System Resource Table (ESRT), safely
+encapsulating the `ESRT_MANAGEMENT_PROTOCOL` functionality from the UEFI specification behind an
+idiomatic Rust service. It introduces an `EsrtRecords` trait for registering, updating, removing,
+and publishing ESRT entries that describe updatable firmware resources. The design focuses on:
+(1) preserving required UEFI semantics, (2) reducing memory‑unsafe patterns inherent in the C
+protocol, and (3) enforcing a minimal, future‑extensible surface for firmware update tracking.
+
+## Change Log
+
+- 2025-10-14: Initial RFC created.
+- 2025-10-17: Updated based on PR review feedback:
+  - Fixed version validation logic to validate individual field correctness rather than enforce incorrect
+    LSV ≤ FwVersion relationship
+  - Restored automatic FMP synchronization feature to maintain EDK2 compatibility (removed from rejected alternatives)
+  - Clarified platform configuration defaults to reference specific EDK2 PCD values (32 entries for both FMP and Non-FMP)
+  - Added configurable auto-lock behavior with secure default (auto_lock_at_ready_to_boot option, defaults to true)
+  - Enhanced EsrtConfiguration documentation with clear references to EDK2 PCDs and security recommendations
+  - Changed capsule_flags from raw u32 to type-safe CapsuleFlags enum for better API design
+  - Added VendorDefined variant to LastAttemptStatus enum with range-validated VendorStatusCode wrapper
+  - Added design rationale for using private component configuration for capacity limits and ESRT-specific settings
+- 2025-11-03: RFC rejected during Final Comment Period:
+  - Added rejection summary documenting community decision to reject standalone ESRT component
+  - Community consensus favors integrated FMP/ESRT approach over standalone ESRT service
+
+## Motivation
+
+The EFI System Resource Table (ESRT) provides a standardized mechanism for firmware to communicate
+information about updatable system resources to the operating system. This includes system firmware,
+device firmware, and UEFI drivers that can be updated through the UEFI UpdateCapsule mechanism. This
+RFC proposes a pure Rust interface for existing ESRT capabilities, replacing the C-based implementations
+while preserving the semantics and behavior specified in the UEFI specification.
+
+**Compatibility Definition**:
+
+1. **Functional Equivalence**: The Rust implementation provides all functionality required by `ESRT_MANAGEMENT_PROTOCOL`
+2. **Semantic Preservation**: Operations maintain the same behavior, constraints, and error conditions as specified in UEFI
+3. **Data Format Compliance**: ESRT table structures conform exactly to the UEFI specification
+4. **Interface Transformation**: C-style function pointers, raw pointers, and u32 constants are replaced with idiomatic
+   Rust traits, ownership, type safety, and strongly-typed enums
+5. **Interoperability**: The implementation can produce C-compatible protocol bindings when needed for legacy component integration
+
+This approach provides a simpler, safer Rust-based interface while maintaining required ESRT functionality and
+ensuring that the resulting firmware behavior aligns with UEFI specification requirements. By using Rust-native
+types (enums for firmware types and status codes, Guid instead of raw arrays), the API prevents entire classes
+of errors at compile time that would be runtime errors in C.
+
+### Scope
+
+The `EsrtRecords` service implements equivalent functionality for the following protocol:
+
+- `ESRT_MANAGEMENT_PROTOCOL`
+  - `RegisterEsrtEntry`
+  - `UpdateEsrtEntry`
+  - `UnRegisterEsrtEntry`
+  - `GetEsrtEntry`
+  - `SyncEsrtFmp`
+  - `LockEsrtRepository`
+
+Additionally, the service provides ESRT table publication to the UEFI Configuration Table for OS visibility.
+
+## Technology Background
+
+The EFI System Resource Table (ESRT) provides a standardized mechanism for firmware to communicate
+updatable resource information to the operating system. The ESRT contains entries describing firmware
+resources (system firmware, device firmware, UEFI drivers) with version information, update status,
+and rollback protection metadata. See UEFI Specification 2.11, Section 23.4 for complete details.
+
+ESRT resources originate from two sources:
+
+- **FMP (Firmware Management Protocol)**: Dynamic resources from FMP protocol instances
+- **Non-FMP**: Static resources registered by platform code
+
+The UEFI Forum Specifications define the ESRT Management Protocol for managing entries:
+[ESRT_MANAGEMENT_PROTOCOL](https://github.com/tianocore/edk2/blob/master/MdeModulePkg/Include/Protocol/EsrtManagement.h)
+
+## Goals
+
+Create an idiomatic Rust API for ESRT-related protocols (*see [Motivation - Scope](#scope)*).
+
+## Requirements
+
+1. The API should provide all necessary ESRT functionality as a service to components
+2. The API should utilize Rust best practices, particularly memory safety and error handling
+3. The ESRT service should produce protocols equivalent to the current C implementations, preserving existing C functionality
+4. Support both FMP-based and Non-FMP resource registration
+5. Provide automatic FMP synchronization to discover dynamic firmware resources
+6. Enable repository locking to prevent modifications after platform initialization
+7. Publish ESRT table to UEFI Configuration Table for OS consumption
+
+## Unresolved Questions
+
+None at this time.
+
+## Prior Art (Existing PI C Implementation)
+
+This Patina-based ESRT implementation follows the ESRT Management Protocol as implemented in EDK2.
+In C, `ESRT_PRIVATE_DATA` provides the core management structure with separate locks for FMP and
+Non-FMP repositories. Entries are stored in UEFI variables (`EsrtFmp` and `EsrtNonFmp`) and the
+driver publishes the consolidated table to the Configuration Table on the Ready To Boot event.
+
+Key C implementation patterns preserved:
+
+- Separate variable storage for FMP and Non-FMP entries
+- Lock-protected repository access
+- Entry-level granularity for all operations
+- Automatic FMP synchronization on Ready To Boot event
+- Ready To Boot event for table publication
+
+### Dependencies on C Protocols
+
+While the final outcome should be a purely Rust-based interface, the current implementation depends on:
+
+- `FirmwareManagementProtocol` for FMP enumeration
+
+Note: RuntimeServices and BootServices are natively provided by Patina and do not require C FFI.
+
+## Alternatives
+
+### Memory Safety Considerations
+
+**Security-First Approach**: This implementation prioritizes memory safety and specification compliance.
+All ESRT entry operations validate input data before modification to prevent corruption of the firmware
+update metadata that the OS relies upon.
+
+#### Entry Validation Safety Model
+
+The service validates all ESRT entries before accepting them into the repository:
+
+1. **FwClass Uniqueness**: Prevents duplicate GUIDs within the same repository (FMP or Non-FMP)
+2. **Firmware Type Validation**: Ensures FwType is one of the valid values (System/Device/Driver)
+3. **Version Field Validation**: Ensures version fields are valid (non-zero LowestSupportedFwVersion, proper version encoding)
+4. **Repository Capacity**: Enforces maximum entry limits to prevent unbounded growth
+   - **Security Rationale**: Protects against memory exhaustion attacks and ensures predictable memory usage
+   - **EDK2 Compatibility**: Default limits (32 entries each) match existing EDK2 PCD defaults, maintaining
+     established platform expectations and tested deployment patterns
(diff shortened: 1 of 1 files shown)


docs/src/rfc/text/0000-esrt.md
```diff
@@ -0,0 +1,383 @@
...
+All ESRT entry operations validate input data before modification to prevent corruption of the firmware
+update metadata that the OS relies upon.
+
+#### Entry Validation Safety Model
+
+The service validates all ESRT entries before accepting them into the repository:
+
+1. **FwClass Uniqueness**: Prevents duplicate GUIDs within the same repository (FMP or Non-FMP)
+2. **Firmware Type Validation**: Ensures FwType is one of the valid values (System/Device/Driver)
+3. **Version Consistency**: Validates that FwVersion >= LowestSupportedFwVersion
```
joschock: Can you explain this one a bit more? It's entirely possible (since components could be updated out of band, e.g. in factory setting) that the actual firmware version might be lower than the LSV. LSV only declares what UEFI will accept in terms of an update e.g. via capsule, but the actual version if the installed firmware should be reflected in ESRT, even if it is lower than the LSV.

docs/src/rfc/text/0000-esrt.md
```diff
@@ -0,0 +1,383 @@
...
+- **Repository Separation**: Separate FMP/Non-FMP storage preserves trust boundaries and lifecycle management
+- **Write Protection**: `lock_repository()` prevents post-initialization modifications
+
+### Rejected Alternatives
+
+**Single Unified Repository**: Rejected because it loses the trust boundary between platform and third-party code,
+complicates synchronization, and creates a single point of failure.
+
+**Automatic FMP Synchronization**: Rejected to give platforms explicit control over enumeration timing and avoid
+complexity of event-based synchronization.
```
joschock: Requirement 5 above calls this out as a feature requirement, but here it is rejected; evaluate for consistency.

docs/src/rfc/text/0000-esrt.md
```diff
@@ -0,0 +1,383 @@
...
+    commands.add_service(self);
+    Ok(())
+}
+```
+
+Platforms can configure ESRT capacity limits:
+
+```rust
+pub struct EsrtConfiguration {
+    pub max_fmp_entries: u32,     // Defaults to platform-specific value
```
joschock: what does `Defaults to platform-specific value` mean?

docs/src/rfc/text/0000-esrt.md
```diff
@@ -0,0 +1,383 @@
...
+    pub max_non_fmp_entries: u32, // Defaults to platform-specific value
+}
+```
+
+### Service Overview
+
+The ESRT component exposes one primary capability: a resource entry service that collects firmware
+resource metadata, stores it persistently, and (on demand) publishes a consolidated ESRT table to
+the UEFI Configuration Table.
```
joschock: nitpick
```suggestion
resource metadata, stores it persistently, and (on demand) publishes a consolidated ESRT table as a UEFI Configuration Table.
```

docs/src/rfc/text/0000-esrt.md
```diff
@@ -0,0 +1,383 @@
...
+- **FMP Synchronization**: Enumerate FMP instances and update FMP repository
+- **Repository Locking**: Prevent modifications after platform initialization
+- **Automatic Table Publication**: Register Ready To Boot event handler to publish consolidated ESRT
+
+### Public Interaction Pattern (High Level)
+
+1. Platform code registers Non-FMP entries for system firmware resources
+2. After device enumeration, platform calls `sync_fmp()` to discover FMP resources
+3. Platform may update entries to reflect firmware update attempts
+4. Before End of DXE, platform calls `lock_repository()`
```
joschock: I think best practice here is to do the right thing by default, rather than relying on platform implementors to "do the right thing" - Consider making this a "default fail-closed" - i.e. "Unless otherwise configured by the platform, `lock_repository` will automatically be called at Ready To Boot." If it is needed, consider providing a configuration option to diverge from this default behavior.  

Also - per the description above, the point of having sync_fmp() would be to allow the platform to discover all devices with FMP instances via a ConnectAll. While it is possible to initiate driver connects before EndOfDxe, ConnectAll is usually a BDS thing. Some platform trust models defer execution of things like OptionRoms etc. that could produce an FMP until after the EOD trust boundary. This means that you probably don't want to lock the repository (by default) until ReadyToBoot.

docs/src/rfc/text/0000-esrt.md
```diff
@@ -0,0 +1,383 @@
...
+
+```rust
+/// Firmware resource entry
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct SystemResourceEntry {
+    pub fw_class: Guid,
+    pub fw_type: FirmwareType,
+    pub fw_version: u32,
+    pub lowest_supported_fw_version: u32,
+    pub capsule_flags: u32,
```
joschock: consider using an enum for flags?

docs/src/rfc/text/0000-esrt.md
```diff
@@ -0,0 +1,383 @@
...
+    Success,
+    ErrorUnsuccessful,
+    ErrorInsufficientResources,
+    ErrorIncorrectVersion,
+    ErrorInvalidFormat,
+    ErrorAuthError,
+    ErrorPowerEventAc,
+    ErrorPowerEventBattery,
+    ErrorUnsatisfiedDependencies,
+    // Additional status codes per UEFI spec
```
joschock: There are also Oem-defined proprietary LastAttemptStatus codes, so I suspect you're going to need an entry in this enum that is something like `VendorDefined(u32)`

docs/src/rfc/text/0000-esrt.md
makubacki: I was thinking about an auto lock as well. The Ready to Boot default seems reasonable.

docs/src/rfc/text/0000-esrt.md
rogurr: I never liked how the original protocol had a function to find all FMP protocols and create a table.  Are you suggesting here that we remove EsrtDxeSyncFmp from the protocol and require active creation of entries by the FMP drivers?

docs/src/rfc/text/0000-esrt.md
```diff
@@ -0,0 +1,383 @@
...
+        event_context,
+        &EVENT_GROUP_READY_TO_BOOT,
+    )?;
+    
+    commands.add_service(self);
+    Ok(())
+}
+```
+
+Platforms can configure ESRT capacity limits:
```
rogurr: Do we need to implement a max limit?  Are there any situations where a bad actor might be blocked by having limits?

docs/src/rfc/rejected/0023-esrt.md
```diff
@@ -0,0 +1,383 @@
...
+
+## Unresolved Questions
+
+None at this time.
+
+## Prior Art (Existing PI C Implementation)
+
+This Patina-based ESRT implementation follows the ESRT Management Protocol as implemented in EDK2.
+In C, `ESRT_PRIVATE_DATA` provides the core management structure with separate locks for FMP and
+Non-FMP repositories. Entries are stored in UEFI variables (`EsrtFmp` and `EsrtNonFmp`) and the
```
rogurr, line 100: Not saying it's a bad idea, asking that if we remove the callback to collect all FMP instances and rely on runtime creating of the data, do we still need the complexity of two stores?

docs/src/rfc/text/0000-esrt.md
joschock: I'm just saying that the requirements statement and this statement are not aligned - and either we should remove the requirement or re-evaluate rejecting it. 

As to my personal opinion as to whether or not automatic sync of FMP into ESRT is a good idea, architecturally I think I could go either way, but practically, if you materially change the behavior it's going to cause work on the platform side, and work on the platform side raises the cost of adoption. Also, it's not clear what the point of an FMP driver that you _don't_ want sync'd into the ESRT driver would be.

docs/src/rfc/rejected/0023-esrt.md
```diff
@@ -0,0 +1,383 @@
...
+
+### Component Architecture
+
+The ESRT service follows Patina's component/service pattern:
+
+```rust
+#[derive(IntoComponent, IntoService)]
+#[service(dyn EsrtRecords)]
+pub struct EsrtManager {
+    // Internal state for entry management
```
cfernald, line 184: I wonder if the max values in the config below make more sense as configuration in the structure itself instead, it doesnt seem that information needs to be broadly available.

docs/src/rfc/text/0000-esrt.md
cfernald: Potentially heavy handed, but you can enforce correctness on custom type ranges in the enum similar to the memory types here: https://github.com/OpenDevicePartnership/patina/blob/764522e1c70eaf77ce4047257fea1d467b87a00d/sdk/patina/src/efi_types.rs#L56

cfernald (approved):

docs/src/rfc/text/0000-esrt.md
kat-perez: gotcha. i'll update this to reflect field validation instead, meaning that it will check that the fields are well-formed rather than enforcing an incorrrect relationship between them

docs/src/rfc/text/0000-esrt.md
kat-perez: I'm deciding to keep the automatic FMP synchronization to match the current EDK2 implementation since I don't think there's any real gain from changing it. I'll update this to remove this in the "Rejected Alternatives" section for consistency.

docs/src/rfc/text/0000-esrt.md
kat-perez: fixed

docs/src/rfc/text/0000-esrt.md
kat-perez: updated this to say we want to match EDK2 max number of entries=32. the reasoning though is unclear to me but i will stick to matching the current defaults

docs/src/rfc/text/0000-esrt.md
kat-perez: updated this RFC to describe the "auto-lock repository at Ready To Boot" as the default behavior and a component config option to set the value explicitly

docs/src/rfc/text/0000-esrt.md
kat-perez: updated

docs/src/rfc/text/0000-esrt.md
kat-perez: updated

docs/src/rfc/text/0000-esrt.md
kat-perez: agree, updated

docs/src/rfc/text/0000-esrt.md
kat-perez: updated with `VendorDefined(VendorStatusCode)`

docs/src/rfc/rejected/0023-esrt.md
kat-perez: yeah, makes sense. changing this to EsrtConfiguration instead of Option<Config<EsrtConfiguration>>

makubacki:
@kat-perez, my understanding based on our conversation earlier is that you can move this RFC to "Final Comment Period (FCP)" now. Please just add a line to the PR description that says "Status: FCP" when you're ready.

docs/src/rfc/text/0000-esrt.md
kat-perez: i think we should stick to enforcing maximum limits and EDK2 does that too. design rationale is that we can prevent unbounded memory consumption. we can allow platforms the flexibility to override via a private field in the config struct

docs/src/rfc/rejected/0023-esrt.md
kat-perez: We still need the two stores for separating non-fmp and fmp entries since they require different trust domains. Non-FMP entries persist across boots in UEFI variables while FMP entries are rediscovered each boot cycle

kat-perez: i think we still need the two stores to differentiate the different types of resources

rogurr:
> I would like more justification for the need for this module. One of the goals when evaluating bringing modules to Patina is to ensure we are prioritizing necessary functionality.
> 
> For all the platforms I have worked on we have used EsrtFmpDxe rather than EsrtDxe from edk2. I don't see a reason for the service that EsrtDxe supports. Disconnecting the ESRT from the FMP is understandable but introduces a lot of ambiguity about how to connect ESRT to firmware update. I don't see any good reason for this and would instead encourage all platforms to convert their update to FMP.
> 
> The UEFI spec doesn't define ESRT protocol and like above I don't think it is a good idea.

I did a lot of digging, what if we move more to a combination of both?

I like how EsrtFmpDxe is very light and doesn't need a protocol, but it does not account for a UEFI build that supports non-FMP updates for things like IoT devices or SBCs that have proprietary methods.

EsrtDxe does support non-FMP entries, and it has a call to poll all FMPs, but it has a lot of overhead when really a single "add entry" would suffice and does not automatically perform the FMP search at ready-to-boot.

What if we shrink the component trait down to a single function that allows adding non-FMP entries and a feature flag to enable(default) or disable an automatic FMP parsing at ready-to-boot?

spbrogan:
> What if we shrink the component trait down to a single function that allows adding non-FMP entries and a feature flag to enable(default) or disable an automatic FMP parsing at ready-to-boot?

The issue with any side band method for publishing ESRT is that it requires a custom stack across the code base.  It requires numerous touch points and platform overrides across various code modules and that just adds unnecessary overhead.   Overrides are needed in capsule processing and validation and then obviously in the publishing of ESRT and doing fw update.   My suggestion is that those custom entities convert themselves to FMP.  Once behind the FMP abstraction they can still do all that plus they don't need custom capsule processing, validation, or location of their update module.    

If there was a scenario where they just wanted an ESRT node to report version and didn't want update capability they can still do that with FMP.  FMP SetImage, GetImage, CheckImage can return EFI_UNSUPPORTED.  This gives a UEFI author an easy path to expose the version information of some component while not having to provide an update method.  

All this to say, I don't think an ESRT Protocol or even a single platform defined ESRT method is needed or advised.  I understand for those platforms that have already built something and "hacked" their custom code base up to support their needs this will require some change, but it should be relatively easy to make such a transition and now they won't need to maintain custom overrides.

rogurr:
I hate to have a standalone module that is so tightly coupled to another that it requires publishing a Firmware Management Protocol to properly run.

What if instead of making this a stand-alone component, it gets combined into the component that will eventually replace the FmpDxe driver?  If the goal is to couple the two, maybe a better place for the table construction is inside the same component that provides the runtime services callback and does the flag check on boot to call the proper FW updater?

Since we need to move away from inherently unsafe protocols and this porting effort relies on FMPs being published, maybe we should table this module for now?

And as long as there are traditional FMPs in the system, an OEM can still use either the EsrtDxe or EsrtFmpDxe driver as needed.  So, I'm thinking this porting effort doesn't make much sense anymore and the discussion should be evaluated when the FW update component is created.

Thoughts?

kat-perez:
@rogurr @spbrogan From our discussions, we should move this to `Status: Rejected`. The next plan is for Ron to propose an FMP component in a new RFC and I will update this PR with the rejected reason

docs/src/rfc/rejected/0000-esrt.md
```diff
@@ -0,0 +1,495 @@
...
+   platform-specific customizations that increase the cost of adoption. The community prioritized maintaining
+   compatibility with existing EDK2 patterns where ESRT and FMP are integrated.
+
+### Community Decision Process
+
+The rejection emerged from technical review discussions where the community raised
+fundamental architectural concerns about the standalone ESRT approach. The discussion revealed that the proposed
+component would not provide sufficient value over the existing EDK2 implementation patterns and would require
+additional complexity for platform integration.
+
```
makubacki: I think you can drop this section. Its main substance is redundant with the points in the previous section.

```suggestion
```

makubacki (approved):
Approving the update of the RFC to the rejected state. Please don't forget to update the RFC number before completing the PR.

docs/src/rfc/rejected/0000-esrt.md
kat-perez: will remove

Issue #750, opened 2025-09-17
[Task]: Reduce complex interfaces in perf code


makubacki:
### Feature Overview

Tracking issue for reducing overly complex interfaces in performance code.

Reference: https://github.com/OpenDevicePartnership/patina/pull/746#discussion_r2355892440

### Solution Overview

Review the code and simply interfaces.

### Urgency

Low

### Are you going to implement the feature request?

I will implement the feature

### Do you need maintainer feedback?

No maintainer feedback needed

PR #880, opened 2025-10-15
Improve perf timer accuracy [Rebase & FF]

berlin-with0ut-return:
## Description
On Q35 and other virtualized environments, the TSC frequency may not be accessible. This PR introduces a core configuration options, .with_timer_frequency, to allow platforms to perform their own timer frequency calculations.

- [X] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested
Q35 now gives reasonable timings. 

## Integration Instructions
Platforms need to provide a frequency setting if their frequency cannot be calcaulated by default methods (cpuid on x86, aarch64 registers).

For Q35, we provide frequency through the ACPI PM timer in `patina-dxe-core-qemu`: 

```use core::arch::x86_64;

const DEFAULT_ACPI_TIMER_FREQUENCY: u64 = 3_579_545; // 3.579545 MHz

pub fn calibrate_tsc_frequency(pm_timer: PmTimer) -> u64 {
    log::info!("Calibrating TSC frequency using PM timer at {:?}", pm_timer);
    // Wait for a PM timer edge to avoid partial intervals.
    let mut start_pm = read_pm_timer(pm_timer);
    let mut next_pm;
    loop {
        next_pm = read_pm_timer(pm_timer);
        if next_pm != start_pm {
            break;
        }
    }
    start_pm = next_pm;

    // Record starting TSC.
    let start_tsc = unsafe { x86_64::_rdtsc() };

    // Hz = ticks/second. Divided by 20 ~ ticks / 50 ms.
    const TARGET_INTERVAL_SIZE: u64 = 20;
    let target_ticks = (DEFAULT_ACPI_TIMER_FREQUENCY / TARGET_INTERVAL_SIZE) as u32;

    let mut end_pm;
    loop {
        end_pm = read_pm_timer(pm_timer);
        let delta = end_pm.wrapping_sub(start_pm);
        if delta >= target_ticks {
            break;
        }
    }

    // Record ending TSC.
    let end_tsc = unsafe { x86_64::_rdtsc() };

    // Time elapsed based on PM timer ticks.
    let delta_pm = end_pm.wrapping_sub(start_pm) as u64;
    let delta_time_ns = (delta_pm * 1_000_000_000) / DEFAULT_ACPI_TIMER_FREQUENCY;

    // Rdtsc ticks.
    let delta_tsc = end_tsc - start_tsc;

    // Frequency = Rdstc ticks / elapsed time.
    let freq_hz = (delta_tsc * 1_000_000_000) / delta_time_ns;

    freq_hz
}

fn read_pm_timer(pm_timer: PmTimer) -> u32 {
    log::info!("Reading PM timer at {:?}", pm_timer);
    match pm_timer {
        PmTimer::IoPort { port } => {
            let value: u32;
            unsafe {
                core::arch::asm!(
                    "in eax, dx",
                    in("dx") port,
                    out("eax") value,
                    options(nomem, nostack, preserves_flags),
                );
            }
            value
        }
        PmTimer::Mmio { base } => unsafe { core::ptr::read_volatile(base as *const u32) },
    }
}
```

Once this timer is integrated, the `perf_timer` lib in mu_rust_helpers needs to be removed.

### DIFF
--- a/components/patina_adv_logger/src/component.rs
+++ b/components/patina_adv_logger/src/component.rs
@@ -13,7 +13,10 @@ use alloc::boxed::Box;
 use core::{ffi::c_void, ptr};
 use patina::{
     boot_services::{BootServices, StandardBootServices},
-    component::IntoComponent,
+    component::{
+        IntoComponent,
+        service::{Service, perf_timer::ArchTimerFunctionality},
+    },
     error::{EfiError, Result},
     pi::hob::{Hob, PhaseHandoffInformationTable},
     serial::SerialIO,
@@ -114,11 +117,13 @@ where
     ///
     /// Installs the Advanced Logger Protocol for use by non-local components.
     ///
-    fn entry_point(self, bs: StandardBootServices) -> Result<()> {
+    fn entry_point(self, bs: StandardBootServices, timer: Service<dyn ArchTimerFunctionality>) -> Result<()> {
         let Some(address) = self.adv_logger.get_log_address() else {
             log::error!("Advanced logger not initialized before component entry point!");
             return Err(EfiError::NotStarted);
         };
+
+        self.adv_logger.init_timer(timer);
 
         let protocol = AdvancedLoggerProtocolInternal {
             protocol: AdvancedLoggerProtocol::new(Self::adv_log_write, address),
--- a/cspell.yml
+++ b/cspell.yml
@@ -49,6 +49,8 @@ words:
   - clangpdb
   - corosensei
   - coverity
+  - cntfrq
+  - cntpct
   - cpuid
   - ctypes
   - cuviper
@@ -149,6 +151,7 @@ words:
   - regsvr
   - relia
   - reloc
+  - rdtsc
   - rposition
   - ruint
   - rwlock
@@ -188,3 +191,4 @@ words:
   - zbuild
   - zeroize
   - zunstable
+  
--- a/patina_dxe_core/src/perf_timer.rs
+++ b/patina_dxe_core/src/perf_timer.rs
@@ -1,0 +1,146 @@
+//! Performance timer implementation for DXE core and components.
+//!
+//! ## License
+//!
+//! Copyright (c) Microsoft Corporation.
+//!
+//! SPDX-License-Identifier: Apache-2.0
+//!
+
+use core::sync::atomic::{AtomicU64, Ordering};
+
+use patina::component::service::{IntoService, perf_timer::ArchTimerFunctionality};
+
+/// Performance timer implementation.
+#[derive(IntoService)]
+#[service(dyn ArchTimerFunctionality)]
+pub struct PerfTimer {
+    frequency: AtomicU64,
+}
+
+impl ArchTimerFunctionality for PerfTimer {
+    /// Value of the counter (ticks).
+    fn cpu_count(&self) -> u64 {
+        arch_cpu_count()
+    }
+
+    /// Frequency of `cpu_count` increments (in Hz).
+    /// If a platform has provided a custom frequency via `PERF_FREQUENCY`, that value is used.
+    /// Otherwise, an architecture-specific method is attempted to determine the frequency.
+    fn perf_frequency(&self) -> u64 {
+        if self.frequency.load(Ordering::Relaxed) == 0 {
+            self.frequency.store(arch_perf_frequency(), Ordering::Relaxed);
+        }
+        self.frequency.load(Ordering::Relaxed)
+    }
+}
+
+impl PerfTimer {
+    /// Creates a new `PerfTimer` instance.
+    pub fn new() -> Self {
+        Self { frequency: AtomicU64::new(0) }
+    }
+
+    /// Creates a new `PerfTimer` instance with a specified frequency.
+    pub fn with_frequency(frequency: u64) -> Self {
+        Self { frequency: AtomicU64::new(frequency) }
+    }
+
+    /// Sets the performance frequency.
+    pub fn set_frequency(&self, frequency: u64) {
+        self.frequency.store(frequency, Ordering::Relaxed);
+    }
+
+    /// Gets the stored performance frequency.
+    pub fn get_stored_frequency(&self) -> u64 {
+        self.frequency.load(Ordering::Relaxed)
+    }
+}
+
+impl Default for PerfTimer {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+/// Returns the current CPU count using architecture-specific methods.
+///
+/// Skip coverage as any value could be valid, including 0.
+#[coverage(off)]
+pub(crate) fn arch_cpu_count() -> u64 {
+    #[cfg(target_arch = "x86_64")]
+    {
+        use core::arch::x86_64;
+        unsafe { x86_64::_rdtsc() }
+    }
+    #[cfg(target_arch = "aarch64")]
+    {
+        use aarch64_cpu::registers::{self, Readable};
+        registers::CNTPCT_EL0.get()
+    }
+}
+
+/// Returns the performance frequency using architecture-specific methods.
+/// In general, the performance frequency is a configurable value that may be
+/// provided by the platform. This function is a fallback when no
+/// platform-specific configuration is provided.
+///
+/// Skip coverage as any value could be valid, including 0.
+#[coverage(off)]
+pub(crate) fn arch_perf_frequency() -> u64 {
+    let mut frequency = 0u64;
+
+    // Try to get TSC frequency from CPUID (most Intel and AMD platforms).
+    #[cfg(target_arch = "x86_64")]
+    {
+        use core::arch::{x86_64, x86_64::CpuidResult};
+
+        let CpuidResult { eax, ebx, ecx, .. } = unsafe { x86_64::__cpuid(0x15) };
+        if eax != 0 && ebx != 0 && ecx != 0 {
+            // CPUID 0x15 gives TSC_frequency = (ECX * EAX) / EBX.
+            // Most modern x86 platforms support this leaf.
+            frequency = (ecx as u64 * ebx as u64) / eax as u64;
+        } else {
+            // CPUID 0x16 gives base frequency in MHz in EAX.
+            // This is supported on some older x86 platforms.
+            // This is a nominal frequency and is less accurate for reflecting actual operating conditions.
+            let CpuidResult { eax, .. } = unsafe { x86_64::__cpuid(0x16) };
+            if eax != 0 {
+                frequency = (eax * 1_000_000) as u64;
+            }
+        }
+    }
+
+    // Use CNTFRQ_EL0 for aarch64 platforms.
+    #[cfg(target_arch = "aarch64")]
+    {
+        use patina::read_sysreg;
+        frequency = read_sysreg!(CNTFRQ_EL0);
+    }
+
+    frequency
+}
+
+#[cfg(test)]
+#[coverage(off)]
+mod tests {
+    use patina::component::service::perf_timer::ArchTimerFunctionality as _;
+
+    use crate::perf_timer::PerfTimer;
+
+    #[test]
+    fn test_set_frequency() {
+        let frequency = 19191919;
+        let timer = PerfTimer::new();
+        assert_eq!(timer.get_stored_frequency(), 0);
+        timer.set_frequency(frequency);
+        assert_eq!(timer.perf_frequency(), 19191919);
+
+        let frequency = 20252025;
+        let timer = PerfTimer::with_frequency(frequency);
+        assert_eq!(timer.get_stored_frequency(), frequency);
+
+        let timer = PerfTimer::default();
+        assert_eq!(timer.get_stored_frequency(), 0);
+    }
+}
--- a/sdk/patina/src/component/service/perf_timer.rs
+++ b/sdk/patina/src/component/service/perf_timer.rs
@@ -1,0 +1,37 @@
+//! Arch-specific timer functionality
+//! By default, this module attempts to determine the timer frequency via architecture specific methods.
+//! (cpuid for x86, `CNTFRQ_EL0` for aarch64)
+//!
+//! Platforms can override this with a custom performance frequency by providing the Core with the correct frequency:
+//!
+//! <!-- (The below test has to be ignore because `patna` cannot depend on `patina_dxe_core` - circular dependency.) -->
+//! ```rust,ignore
+//!     let frequency_hz: u64 = 1_000_000_000; // Compute with platform-specific methods.
+//!
+//!     Core::default()
+//!        .init_timer_frequency(Some(frequency_hz))
+//!```
+//!
+//! ## License
+//!
+//! Copyright (c) Microsoft Corporation.
+//!
+//! SPDX-License-Identifier: Apache-2.0
+//!
+
+/// Trait that provides architecture-specific timer functionality.
+/// Components that need timing functionality can request this service.
+pub trait ArchTimerFunctionality: Send + Sync {
+    /// Value of the counter (ticks).
+    fn cpu_count(&self) -> u64;
+    /// Value in Hz of how often the counter increment.
+    fn perf_frequency(&self) -> u64;
+    /// Value that the performance counter starts with.
+    fn cpu_count_start(&self) -> u64 {
+        0
+    }
+    /// Value that the performance counter ends with before it rolls over.
+    fn cpu_count_end(&self) -> u64 {
+        u64::MAX
+    }
+}
--- a/sdk/patina/src/performance/measurement.rs
+++ b/sdk/patina/src/performance/measurement.rs
@@ -21,6 +21,7 @@ use core::{
 
 use crate::{
     boot_services::BootServices,
+    component::service::{Service, perf_timer::ArchTimerFunctionality},
     error::EfiError,
     guids::EDKII_FPDT_EXTENDED_FIRMWARE_PERFORMANCE,
     performance::{
@@ -43,7 +44,6 @@ use crate::{
 };
 
 use crate::pi::status_code::{EFI_PROGRESS_CODE, EFI_SOFTWARE_DXE_BS_DRIVER};
-use mu_rust_helpers::perf_timer::{Arch, ArchFunctionality};
 
 use r_efi::{
     efi::{self, Guid},
@@ -207,7 +207,7 @@ pub unsafe extern "efiapi" fn create_performance_measurement(
     identifier: u32,
     attribute: PerfAttribute,
 ) -> efi::Status {
-    let Some((boot_services, fbpt)) = get_static_state() else {
+    let Some((boot_services, fbpt, timer)) = get_static_state() else {
         // If the state is not initialized, it is because perf in not enabled.
         return efi::Status::SUCCESS;
     };
@@ -251,6 +251,7 @@ pub unsafe extern "efiapi" fn create_performance_measurement(
         attribute,
         boot_services,
         fbpt,
+        timer,
     ) {
         Ok(_) => efi::Status::SUCCESS,
         Err(Error::OutOfResources) => {
@@ -285,16 +286,17 @@ fn _create_performance_measurement<B, F>(
     attribute: PerfAttribute,
     boot_services: &B,
     fbpt: &TplMutex<'static, F, B>,
+    timer: &Service<dyn ArchTimerFunctionality>,
 ) -> Result<(), Error>
 where
     B: BootServices,
     F: FirmwareBasicBootPerfTable,
 {
-    let cpu_count = Arch::cpu_count();
+    let cpu_count = timer.cpu_count();
     let timestamp = match ticker {
-        0 => (cpu_count as f64 / Arch::perf_frequency() as f64 * 1_000_000_000_f64) as u64,
+        0 => (cpu_count as f64 / timer.perf_frequency() as f64 * 1_000_000_000_f64) as u64,
         1 => 0,
-        ticker => (ticker as f64 / Arch::perf_frequency() as f64 * 1_000_000_000_f64) as u64,
+        ticker => (ticker as f64 / timer.perf_frequency() as f64 * 1_000_000_000_f64) as u64,
     };
 
     let Ok(known_perf_id) = KnownPerfId::try_from(perf_id) else {
@@ -548,6 +550,8 @@ pub struct MediaFwVolFilepathDevicePath {
 mod tests {
     use super::*;
 
+    use crate as patina;
+
     use alloc::rc::Rc;
     use core::{mem::MaybeUninit, ptr};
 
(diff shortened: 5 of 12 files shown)


sdk/patina/src/perf_timer/component.rs
```diff
@@ -0,0 +1,58 @@
...
+pub struct PerfTimer {}
+
+impl PerfTimer {
+    /// Create a new performance timer component.
+    pub const fn new() -> Self {
+        Self {}
+    }
+
+    // sherry: the problem is that we store a specific timer type but fetch trait object
+    fn entry_point(&self, config: Config<TimerConfig>) -> core::result::Result<(), EfiError> {
```
cfernald: I assume you need `_` since this it's not used in aarch64

sdk/patina/src/perf_timer/config.rs
```diff
@@ -0,0 +1,8 @@
+use crate::perf_timer::timer::PmTimer;
+
+#[derive(Debug, Default)]
```
cfernald: IIRC the dispatcher will use the default of a config when unspecified, for this reason I don't think we should have the default be a actual IO port as this would cause a config for an IO port to be initialized on all platforms, including aarch64. Probably safer to have the default be a no-op implementation to avoid unexpected hardware interactions or errors.

sdk/patina/src/perf_timer/timer.rs
```diff
@@ -0,0 +1,176 @@
+/// Default PM Timer I/O port.
+/// This value was selected based on the QEMU I/O port, since QEMU will most likely be the primary user of the ACPI timer fallback.
+pub const DEFAULT_PM_PORT: u16 = 0x608;
+
+#[derive(Debug, Clone, Copy)]
```
cfernald: Here and throughout this file, for public interfaces we need documenting comments

sdk/patina/src/perf_timer/timer.rs
```diff
@@ -0,0 +1,176 @@
+/// Default PM Timer I/O port.
+/// This value was selected based on the QEMU I/O port, since QEMU will most likely be the primary user of the ACPI timer fallback.
+pub const DEFAULT_PM_PORT: u16 = 0x608;
+
+#[derive(Debug, Clone, Copy)]
+pub enum PmTimer {
```
cfernald: PmTime = PerforMance Timer?

sdk/patina/src/perf_timer/config.rs
```diff
@@ -0,0 +1,8 @@
+use crate::perf_timer::timer::PmTimer;
+
+#[derive(Debug, Default)]
+pub struct TimerConfig {
+    /// The physical address of the RSDP.
+    /// Used to read configuration details from primarily the FADT.
+    pub timer_read_info: PmTimer,
```
cfernald: Do we expect further configurations to be placed here? I'm trying to rationalize the config wrapping the real configuration which is the enum

sdk/patina/src/perf_timer/timer.rs
```diff
@@ -0,0 +1,176 @@
...
+
+#[derive(Debug, Clone, Copy)]
+pub enum PmTimer {
+    IoPort { port: u16 },
+    Mmio { base: u64 },
+}
+
+impl Default for PmTimer {
+    fn default() -> Self {
+        PmTimer::IoPort { port: DEFAULT_PM_PORT }
```
cfernald: As mentioned above, this should probably be some sort of no-op config for safety.

Javagedes (changes requested):
Any component that starts using `PerfTimer::func()` is now locked into your implementation. But what happens when a platform comes along and `perf_frequency` cannot be calculated via `__cpuid` or the `ACPI PM Timer`? Now that platform either cannot use any component that relies on `PerfTimer` (so patina_adv_logger, patina_performacne), or they have to go back and submit a PR to add their necessary way to calculate the frequency to `PerfTimer`. If this happens to 10 platforms, then this `PerfTimer` has 10 different fallback methods to calculating the frequency that is inside every single platform, when their platform only every need one way to calculate the frequency. This is just not sustainable in the long run.

Who knows, maybe it is for this scenario with `PerfTimer`. Maybe there really are only two ways to ever get the frequency. But regardless, it is not a design pattern we want to encourage or highlight in the core. Even at this single override you are providing, this Qemu specific solution is going to be compiled into every single platform.

Now there are two things going on that you need to consider:
1. You need to allow a platform to possibly override frequency.
2. You need to ensure the timer cannot be used until the frequency is correct.

To solve this, I **suggest** a solution to this:

1. You have a component that produces a timer service that must be requested via the component parameter interface, not through a global function call. This prevents usage of the functionality before it has been initialized. Services as static, so any component can take that service and stash it to use later, but this ensures that the service is only produced when everything is available.
2. The core produces the service. We can even make it a struct service, not a dyn trait interface, because we never expect the platform to override it.
3. We have a component in the core produce the service mentioned above, but we create a public configuration knob that allows the platform to override the default configuration for frequency.

This is basically what it would look like:

```rust
// This is the config knob that a platform can override how they want
pub struct PerfFrequency(u64);

impl PerfFrequency(u64) {
    // A platform would use this function to initialize the config knob and give it to the core
    pub fn new(frequency: u64) {
        Self(frequency)
    }
}

impl Default for PerfFrequency {
  fn default() -> Self {
      // The default is always used if a platform does not set a configuration. Due to this, we can calculate the
      // frequency via the default way - e.g. `__cpuid` or `CNTPCT_EL0`
      // However if a platform registers their own configuration of this value, it will be used instead of this
  }
}

#[derive(IntoService, IntoComponent)]
#[service(PerfTimer)]
struct PerfTimer {
  frequency: u64
}

impl PerfTimer {
    // The component entry point sets the frequency then adds the service
    fn entry_point(mut self, commands: Commands, config: Config<PerfFrequency>) -> Result<()> {
        self.frequency = config.0
        commands.add_service(self);
    }
    pub fn cpu_count(&self) -> u64 {
        // target_arch conditional compliation to use the rdtsc or CNTPCT_EL0
    }
    
    pub fn frequency(&self) -> {
        self.frequency
    }
}
```

Now for services that care about getting the perf timer (I'll use patina_adv_logger as an example), they just request the the service, stash that, and use it:

```rust
/// Entry point for AdvancedLoggerComponent
fn entry_point(self, bs: StandardBootServices, timer: Service<PerfTimer>) -> Result<()> {
    let protocol = AdvancedLoggerProtocolInternal {
        protocol: AdvancedLoggerProtocol::new(Self::adv_log_write, address),
        adv_logger: self.adv_logger,
        timer: timer,
    }
}

// The change to the log write function
    pub(crate) fn log_write(&self, error_level: u32, data: &[u8]) {
        ...
        let timestamp = self.timer.cpu_count();
        ...
    }
```

And the main advantage is the configuration for the platforms. For a platform where the default way to calculate the timer frequency works, they do absolutely nothing. However if a platform needs to override the frequency, they can:

``` rust
// Example of a platform That just wants to override with a hardcoded value
Core::default()
    .init_memory(hob_list)
    .with_config(PerfFrequency::new(MY_CPU_FREQUENCY));


// A platform that wants to do some special calculation, like the ACPI table caclulation
fn calc_timer() -> PerfFrequency {
    todo!()
}

Core::default()
    .init_memory(hob_list)
    .with_config(calc_timer());
   
   
// Or even more advanced, a config that requires some other services be available to set it
#[derive(IntoComponent)]
struct MyComponent;

impl MyComponent {
  fn entry_point(self, config: ConfigMut<PerfFrequency>, other_service: Service<dyn OtherService>) -> Result<()> {
      let frequency = other_service.do_something_to_get_frequency();
      *config.frequency = frequency
      config.lock()
  }
}
```

sdk/patina/src/perf_timer/timer.rs
berlin-with0ut-return: it's confusing but it stands for Power Management in the spec https://wiki.osdev.org/ACPI_Timer

sdk/patina/src/perf_timer/component.rs
```diff
@@ -0,0 +1,112 @@
...
+            PerfFrequency(frequency)
+        }
+    }
+}
+
+/// Performance Timer component that provides the ArchTimerFunctionality service.
+/// `Config<PerfFrequency>` provides a way for platforms to override the default frequency determination method.
+#[derive(IntoComponent, IntoService, Default)]
+#[service(dyn ArchTimerFunctionality)]
+pub struct PerfTimer {
```
cfernald: As far as I can tell, PerfTimer is explicitly only initialized by the core, and uses a SDK defined trait. It seems to be the implementation for this should be internal to the core then instead of in the SDK.

sdk/patina/src/perf_timer/timer.rs
```diff
@@ -0,0 +1,41 @@
...
+//!
+//! Copyright (c) Microsoft Corporation.
+//!
+//! SPDX-License-Identifier: Apache-2.0
+//!
+
+/// Represents port configurations for reading the ACPI Power Management (PM) Timer.
+/// This value is stored in the FADT table in UEFI systems, and is constant within the same platform.
+#[derive(Debug, Clone, Copy)]
+pub enum PmTimer {
```
cfernald: Is this still used anywhere?

sdk/patina/src/perf_timer/component.rs
```diff
@@ -0,0 +1,112 @@
...
+use crate::{
+    component::{IntoComponent, params::*},
+    error::EfiError,
+    perf_timer::timer::ArchTimerFunctionality,
+};
+
+/// Performance frequency knob.
+/// Platforms can override this with a custom performance frequency;
+/// otherwise the default implementation will try to determine it via architecture specific methods.
+pub struct PerfFrequency(pub u64);
```
cfernald: We may need to be more explicit about what performance frequency means here. Perhaps calling out the TSC as the example for x64, which is presumably the only place this would be used.

sdk/patina/src/perf_timer/component.rs
```diff
@@ -0,0 +1,112 @@
...
+impl Default for PerfFrequency {
+    /// The default implementation tries to determine the frequency via architecture specific methods
+    /// if a platform does not provide their own configuration. (cpuid for x86, `CNTFRQ_EL0` for aarch64)
+    ///
+    /// However, if a platform registers their own configuration of this value, it will be used instead of this default.
+    fn default() -> Self {
+        let mut frequency = 0u64;
+
+        // Try to get TSC frequency from CPUID (most Intel and AMD platforms).
+        #[cfg(target_arch = "x86_64")]
```
cfernald: I'm not sure if I like putting more complex logic like this into a Default. Perhaps it would be better for this to have an internal Option, e.g. Option<u64>, and if none then the actual component will perform this logic. It just keeps the code more linear and explicit.

sdk/patina/src/perf_timer/component.rs
cfernald: Additionally, it seems pretty likely that parts of the core will need to be able to access this functionality without/before components/services are available. I wonder if part of this a static should be struct in the core similar to the CoreMemoryManager that would allow code at any point in time to get timestamps.

cfernald:
@berlin-with0ut-return , if the PR description still accurate since it seems you have switched to a static freq override?

Javagedes:
> Overall, I think this is an improvement and I think we move forward. However, I wonder if the Config for the frequency works here or if this should be a `.with_timer_frequency()` option in the core init instead. It seems useful for me for the core to have access to a timer and frequency through all stages, so that performance can be captured in all stages of initialization. This could still then expose a service, like we do with the memory manager, so that components can access the same interface.
> 
> 
> 
> 

I think this would be good.

components/patina_adv_logger/src/component.rs
makubacki, line 1: @berlin-with0ut-return, I'd like to review this PR, but I don't have time today. Can you please hold it until at least Monday of next week?

sdk/patina/src/perf_timer/timer.rs
makubacki: It also looks like this could be removed to me as well in the current implementation.

sdk/patina/src/performance/measurement.rs
```diff
@@ -281,16 +283,19 @@ fn _create_performance_measurement<B, F>(
...
+    timer: &Service<dyn ArchTimerFunctionality>,
 ) -> Result<(), Error>
 where
     B: BootServices,
     F: FirmwareBasicBootPerfTable,
 {
-    let cpu_count = Arch::cpu_count();
+    log::info!("before getting timing");
+    let cpu_count = timer.cpu_count();
+    // let cpu_count = Arch::cpu_count();
```
makubacki:
```suggestion
```

sdk/patina/src/performance/measurement.rs
```diff
@@ -281,16 +283,19 @@ fn _create_performance_measurement<B, F>(
...
     boot_services: &B,
     fbpt: &TplMutex<'static, F, B>,
+    timer: &Service<dyn ArchTimerFunctionality>,
 ) -> Result<(), Error>
 where
     B: BootServices,
     F: FirmwareBasicBootPerfTable,
 {
-    let cpu_count = Arch::cpu_count();
+    log::info!("before getting timing");
```
makubacki: This should either be moved to something like trace level or be removed.

```suggestion
```

sdk/patina/src/performance/measurement.rs
```diff
@@ -281,16 +283,19 @@ fn _create_performance_measurement<B, F>(
     attribute: PerfAttribute,
     boot_services: &B,
     fbpt: &TplMutex<'static, F, B>,
+    timer: &Service<dyn ArchTimerFunctionality>,
```
makubacki, line 289: Note: This function signature is very complex. But that's already being tracked in https://github.com/OpenDevicePartnership/patina/issues/750 independent of this change.

sdk/patina/src/perf_timer/component.rs
makubacki: I agree that this should be added independent of the component infrastructure.

makubacki (changes requested):
This PR will fundamentally change after https://github.com/OpenDevicePartnership/patina/pull/880/files#r2441251361  so I'm just marking was changes requested so the review state of the PR is obvious until that is done.

sdk/patina/src/perf_timer/timer.rs
berlin-with0ut-return: it is used by Q35 to read the ACPI PM timer, however i've moved this implementation to patina-dxe-core-qemu so this PmTimer struct can also be moved there

sdk/patina/src/perf_timer.rs
```diff
@@ -0,0 +1,29 @@
+//! Arch-specific timer functionality
```
cfernald, line 1: should this be in component/service?

components/patina_adv_logger/src/component.rs
```diff
@@ -145,6 +148,7 @@ mod tests {
     extern crate std;
     use core::mem::size_of;
 
+    use patina::boot_services::MockBootServices;
```
cfernald: Is this being used here?

patina_dxe_core/src/perf_timer.rs
```diff
@@ -0,0 +1,98 @@
...
+    // Try to get TSC frequency from CPUID (most Intel and AMD platforms).
+    #[cfg(target_arch = "x86_64")]
+    {
+        use core::arch::x86_64;
+        use core::arch::x86_64::CpuidResult;
+
+        let CpuidResult { eax, ebx, ecx, .. } = unsafe { x86_64::__cpuid(0x15) };
+        if eax != 0 && ebx != 0 && ecx != 0 {
+            // CPUID 0x15 gives TSC_frequency = (ECX * EAX) / EBX.
+            // Most modern x86 platforms support this leaf.
```
cfernald, line 101: It may be worth noting here that this may not be a one-size fits all. I know there have been issues in the past using this on AMD processors and they have alternative mechanisms to calibrate TSC frequency there using MSRs.

In fact, considering q35 is technically a  AMD core, have we checked that these MSRs are not supported there?

patina_dxe_core/src/perf_timer.rs
```diff
@@ -0,0 +1,98 @@
...
+    /// Value of the counter (ticks).
+    fn cpu_count(&self) -> u64 {
+        arch_cpu_count()
+    }
+
+    /// Frequency of `cpu_count` increments (in Hz).
+    /// If a platform has provided a custom frequency via `PERF_FREQUENCY`, that value is used.
+    /// Otherwise, an architecture-specific method is attempted to determine the frequency.
+    fn perf_frequency(&self) -> u64 {
+        if let Some(freq) = PERF_FREQUENCY.get() { *freq } else { arch_perf_frequency() }
```
cfernald: Should we try to save the frequency somewhere once it's calibrates? You could consider just having one PERF_FREQUENCY if 0, will be calibrated, otherwise gets used and that is what is set in the `init_timer_frequency`

patina_dxe_core/src/perf_timer.rs
```diff
@@ -0,0 +1,98 @@
...
+                frequency = (eax * 1_000_000) as u64;
+            }
+        }
+    }
+
+    // Use CNTFRQ_EL0 for aarch64 platforms.
+    #[cfg(target_arch = "aarch64")]
+    {
+        use aarch64_cpu::registers::{self, Readable};
+        frequency = registers::CNTFRQ_EL0.get();
```
cfernald: Consider using the new read_sysreg  in the SDK instead

https://github.com/OpenDevicePartnership/patina/blob/main/sdk/patina/src/arch/aarch64.rs

patina_dxe_core/src/perf_timer.rs
cfernald: I did a bit of digging and I don't think anything useful is available as you have mentioned before.

sdk/patina/src/perf_timer.rs
berlin-with0ut-return: i'm actually not sure where to put this since it's a very small service definition and the component part of this is defined in the core, not components/

cfernald: This is true for the Memory Manager too, and that is where it was decided to place that. I'm open to other places, but we should be consistent with wherever we place them.

Personally I think I would prefer 

component/
       |------ component stuff
service/
      |-- memory_manager.rs
      |-- perf_timer.rs

@Javagedes , thoughts?

Javagedes: Yes please place it where chris requested.

sdk/patina/src/component/service/perf_timer.rs
```diff
@@ -0,0 +1,29 @@
+//! Arch-specific timer functionality
+//! By default, this module attempts to determine the timer frequency via architecture specific methods.
+//! (cpuid for x86, `CNTFRQ_EL0` for aarch64)
+//!
+//! Platforms can override this with a custom performance frequency by providing the `PerfFrequency` configuration.
```
makubacki: Is the `PerfFrequency` configuration defined somewhere?

berlin-with0ut-return: that comment is outdated, it should be using .with_frequency in the core

patina_dxe_core/src/perf_timer.rs
```diff
@@ -0,0 +1,102 @@
+use core::sync::atomic::{AtomicU64, Ordering};
+
+use patina::component::service::{IntoService, perf_timer::ArchTimerFunctionality};
+
+/// Cached performance frequency.
+pub(crate) static PERF_FREQUENCY: AtomicU64 = AtomicU64::new(0);
```
makubacki: Could the frequency be stored as a struct member and initialized with a method?

berlin-with0ut-return: yes but i think this would still require either a global PERF_TIMER struct or the Core<NoAlloc> to store a `perf_timer` field (in order to set the frequency and register `PerfTimer` as a service).

berlin-with0ut-return: i think having it as a field is overall better (and changing the global PERF_TIMER=PerfTimer::new(frequency=0) ) since it couples the `PerfTimer` and the frequency which are associated constructs, but want to be sure that's what you mean

makubacki: My understanding is it would be a field in this `Core` structure, right?

```rust
pub struct Core<MemoryState> {
    physical_hob_list: *const c_void,
    hob_list: HobList<'static>,
    components: Vec<Box<dyn Component>>,
    storage: Storage,
    platform_timer_frequency: Option<u64>, // Store frequency here
    _memory_state: core::marker::PhantomData<MemoryState>,
}
```

If so, I think that's better than the global static as it makes the code cleaner and the ownership and lifetime more obvious.

patina_dxe_core/src/perf_timer.rs
makubacki, line 1: Please add a module doc comment for this file.

sdk/patina/src/component/service/perf_timer.rs
```diff
@@ -0,0 +1,38 @@
+//! Arch-specific timer functionality
+//! By default, this module attempts to determine the timer frequency via architecture specific methods.
+//! (cpuid for x86, `CNTFRQ_EL0` for aarch64)
+//!
+//! Platforms can override this with a custom performance frequency by providing the Core with the correct frequency:
+//!
+//!
+//! (The below test has to be ignore because `patna` cannot depend on `patina_dxe_core` - circular dependency.)
```
makubacki: Some minor cleanup.

```suggestion
//!
//! <!-- (The below test has to be ignore because `patna` cannot depend on `patina_dxe_core` - circular dependency.) -->
```

sdk/patina/src/component/params.rs
makubacki, line 1: Can you please separate the changes in this file into a separate commit that comes before the perf timer commit (squash all of the other changes into a single commit and set for rebase & ff)?

core/patina_stacktrace/resolve_stacktrace/pdbhelper.dll
makubacki, line 1: Is `pdbhelper.dll` meant to be in this PR?

makubacki (approved):

core/patina_stacktrace/resolve_stacktrace/pdbhelper.dll
berlin-with0ut-return: it is not, i have removed it

cspell.yml
```diff
@@ -187,4 +190,4 @@ words:
   - xtask
   - zbuild
   - zeroize
-  - zunstable
+  - zunstable
```
makubacki: @berlin-with0ut-return, a few files have the empty new line removed at the end of the file. Can you please ensure each file has one before completing?

berlin-with0ut-return: done

PR #883, opened 2025-10-15
RFC: `AMLParser` UEFI service

berlin-with0ut-return:
## Description
RFC proposes a new AML handling service.

Reviewers: please focus on functional and design considerations as Copilot will catch grammatical/formatting errors.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [X] Includes documentation?

## How This Was Tested
N/A.

## Integration Instructions
Will change the RFC# upon merge.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,7 +41,7 @@ patina_internal_depex = { version = "14.3.2", path = "core/patina_internal_depex
 patina_internal_device_path = { version = "14.3.2", path = "core/patina_internal_device_path" }
 patina_lzma_rs = { version = "0.3.1", default-features = false }
 patina_macro = { version = "14.3.2", path = "sdk/patina_macro" }
-patina_mtrr = { version = "1.0.0" }
+patina_mtrr = { version = "^1.1.4" }
 patina_paging = { version = "10" }
 patina_performance = { version = "14.3.2", path = "components/patina_performance" }
 patina_smbios = { version = "14.3.2", path = "components/patina_smbios" }
--- a/core/patina_debugger/Cargo.toml
+++ b/core/patina_debugger/Cargo.toml
@@ -21,6 +21,9 @@ bitfield-struct = { workspace = true  }
 [dev-dependencies]
 mockall = { workspace = true }
 
+[target.'cfg(all(target_arch="x86_64"))'.dependencies]
+patina_mtrr = { workspace = true }
+
 [features]
 default = ["alloc", "windbg_workarounds"]
 alloc = []
--- a/core/patina_debugger/src/arch.rs
+++ b/core/patina_debugger/src/arch.rs
@@ -21,9 +21,11 @@ use crate::ExceptionInfo;
 
 cfg_if::cfg_if! {
     if #[cfg(target_arch = "x86_64")] {
+        #[coverage(off)]
         mod x64;
         pub type SystemArch = x64::X64Arch;
     } else if #[cfg(target_arch = "aarch64")] {
+        #[coverage(off)]
         mod aarch64;
         pub type SystemArch = aarch64::Aarch64Arch;
     }
--- a/core/patina_debugger/src/arch/aarch64.rs
+++ b/core/patina_debugger/src/arch/aarch64.rs
@@ -237,6 +237,7 @@ impl DebuggerArch for Aarch64Arch {
         match tokens.next() {
             Some("regs") => {
                 print_sysreg!(ttbr0_el2, out);
+                print_sysreg!(mair_el2, out);
                 print_sysreg!(esr_el2, out);
                 print_sysreg!(far_el2, out);
                 print_sysreg!(tcr_el2, out);
--- a/core/patina_debugger/src/arch/x64.rs
+++ b/core/patina_debugger/src/arch/x64.rs
@@ -9,6 +9,7 @@ use gdbstub::{
     target::ext::breakpoints::WatchKind,
 };
 use patina_internal_cpu::interrupts::ExceptionContext;
+use patina_mtrr::Mtrr;
 use patina_paging::PagingType;
 
 use super::{DebuggerArch, UefiArchRegs};
@@ -190,8 +191,28 @@ impl DebuggerArch for X64Arch {
                     asm!("mov {0}, cr3", "mov cr3, {0}", out(reg) _, options(nostack, nomem));
                 }
             }
+            Some("mtrr") => {
+                if let Some(val) = tokens.next() {
+                    let mtrr = patina_mtrr::create_mtrr_lib(0);
+                    let addr = match u64::from_str_radix(val.trim_start_matches("0x"), 16) {
+                        Ok(a) => a,
+                        Err(_) => {
+                            let _ = out.write_str(
+                                alloc::format!("Invalid address format: '{val}'. Expected hex address (e.g. 0x1000).")
+                                    .as_str(),
+                            );
+                            return;
+                        }
+                    };
+
+                    let attr = mtrr.get_memory_attribute(addr);
+                    let _ = write!(out, "{}", attr);
+                } else {
+                    let _ = out.write_str("Usage: mtrr <base_address>");
+                }
+            }
             _ => {
-                let _ = out.write_str("Unknown X64 monitor command. Supported commands: regs, flush_tlb");
+                let _ = out.write_str("Unknown X64 monitor command. Supported commands: regs, flush_tlb, mtrr <addr>");
             }
         }
     }
--- a/core/patina_internal_cpu/src/interrupts/aarch64.rs
+++ b/core/patina_internal_cpu/src/interrupts/aarch64.rs
@@ -13,12 +13,14 @@ use patina_stacktrace::{StackFrame, StackTrace};
 
 cfg_if::cfg_if! {
     if #[cfg(all(target_os = "uefi", target_arch = "aarch64"))] {
+        #[coverage(off)]
         mod interrupt_manager;
         pub mod gic_manager;
         pub use interrupt_manager::InterruptsAarch64;
         use patina::{read_sysreg, write_sysreg};
     } else if #[cfg(feature = "doc")] {
         pub use interrupt_manager::InterruptsAarch64;
+        #[coverage(off)]
         mod interrupt_manager;
     }
 }
--- a/core/patina_internal_cpu/src/interrupts/x64.rs
+++ b/core/patina_internal_cpu/src/interrupts/x64.rs
@@ -14,10 +14,12 @@ use patina_stacktrace::{StackFrame, StackTrace};
 
 cfg_if::cfg_if! {
     if #[cfg(all(target_os = "uefi", target_arch = "x86_64"))] {
+        #[coverage(off)]
         mod interrupt_manager;
         pub use interrupt_manager::InterruptsX64;
     } else if #[cfg(feature = "doc")] {
         pub use interrupt_manager::InterruptsX64;
+        #[coverage(off)]
         mod interrupt_manager;
     }
 }
--- a/core/patina_internal_cpu/src/interrupts/x64/interrupt_manager.rs
+++ b/core/patina_internal_cpu/src/interrupts/x64/interrupt_manager.rs
@@ -15,6 +15,7 @@ use patina::{
     error::EfiError,
     pi::protocols::cpu_arch::EfiSystemContext,
 };
+use patina_mtrr::Mtrr;
 use patina_paging::{PageTable, PagingType};
 use patina_stacktrace::{StackFrame, StackTrace};
 use x86_64::{
@@ -273,4 +274,10 @@ unsafe fn dump_pte(cr2: u64, cr3: u64, paging_type: PagingType) {
     } {
         let _ = pt.dump_page_tables(cr2 & !(UEFI_PAGE_MASK as u64), UEFI_PAGE_SIZE as u64);
     }
+
+    // we don't carry the caching attributes in the page table, so get them from the MTRRs
+    let mtrr = patina_mtrr::create_mtrr_lib(0);
+    log::error!("");
+    log::error!("MTRR Cache Attribute: {}", mtrr.get_memory_attribute(cr2));
+    log::error!("");
 }
--- a/patina_dxe_core/src/allocator/fixed_size_block_allocator.rs
+++ b/patina_dxe_core/src/allocator/fixed_size_block_allocator.rs
@@ -869,7 +869,7 @@ mod tests {
         let layout = Layout::from_size_align(size, UEFI_PAGE_SIZE).unwrap();
         let base = unsafe { System.alloc(layout) as u64 };
         unsafe {
-            gcd.add_memory_space(GcdMemoryType::SystemMemory, base as usize, size, efi::MEMORY_WB).unwrap();
+            gcd.init_memory_blocks(GcdMemoryType::SystemMemory, base as usize, size, efi::MEMORY_WB).unwrap();
         }
         base
     }
--- a/patina_dxe_core/src/allocator/uefi_allocator.rs
+++ b/patina_dxe_core/src/allocator/uefi_allocator.rs
@@ -300,7 +300,7 @@ mod tests {
         let layout = Layout::from_size_align(size, UEFI_PAGE_SIZE).unwrap();
         let base = unsafe { System.alloc(layout) as u64 };
         unsafe {
-            gcd.add_memory_space(dxe_services::GcdMemoryType::SystemMemory, base as usize, size, efi::MEMORY_WB)
+            gcd.init_memory_blocks(dxe_services::GcdMemoryType::SystemMemory, base as usize, size, efi::MEMORY_WB)
                 .unwrap();
         }
         base
--- a/patina_dxe_core/src/gcd.rs
+++ b/patina_dxe_core/src/gcd.rs
@@ -66,19 +66,14 @@ pub fn init_gcd(physical_hob_list: *const c_void) {
     assert!(memory_start < memory_end, "Not enough memory available for DXE core to start.");
 
     // initialize the GCD with an initial memory space. Note: this will fail if GCD.init() above didn't happen.
+    // SAFETY: We are directly using the free memory space from the PHIT HOB, which must be valid and reserved for use
+    // per spec.
     unsafe {
-        GCD.add_memory_space(
+        GCD.init_memory_blocks(
             GcdMemoryType::SystemMemory,
             free_memory_start as usize,
             free_memory_size as usize,
-            efi::MEMORY_UC
-                | efi::MEMORY_WC
-                | efi::MEMORY_WT
-                | efi::MEMORY_WB
-                | efi::MEMORY_WP
-                | efi::MEMORY_RP
-                | efi::MEMORY_XP
-                | efi::MEMORY_RO,
+            efi::MEMORY_ACCESS_MASK | efi::CACHE_ATTRIBUTE_MASK,
         )
         .expect("Failed to add initial region to GCD.");
     }
--- a/patina_dxe_core/src/gcd/spin_locked_gcd.rs
+++ b/patina_dxe_core/src/gcd/spin_locked_gcd.rs
@@ -337,7 +337,7 @@ impl GCD {
         self.maximum_address = 1 << processor_address_bits;
     }
 
-    unsafe fn init_memory_blocks(
+    pub(crate) unsafe fn init_memory_blocks(
         &mut self,
         memory_type: dxe_services::GcdMemoryType,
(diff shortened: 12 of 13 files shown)


docs/src/rfc/text/0017-aml-parser.md
```diff
@@ -0,0 +1,282 @@
...
+    - Cons: `acpi` has limited public interfaces, which may constrain the development of the Rust ACPI service.
+It primarily focuses on looking up and executing AML in the application space,
+with less support for actually walking through and modifying the firmware-side AML object tree.
+
+There is ongoing conversation with the owner of the `acpi` crate about
+borrowing certain implementations and modifying the public interfaces
+to be more friendly to the Patina ACPI implementation.
+This conversation can be tracked through a [Github issue in the `acpi` crate](https://github.com/rust-osdev/acpi/issues/260).
+
+## Rust Code
```
joschock: General thought on implementation: could we code the parser/writer as an independent standalone - rust lib crate? There is prior art on the reader side (https://docs.rs/aml/latest/aml/). But writer/patcher is unique. However, I think it would be neat to have a non-patina crate that just does aml byte code parsing/modification that is then leveraged by this component.

docs/src/rfc/text/0017-aml-parser.md
```diff
@@ -0,0 +1,282 @@
...
+
+This RFC proposes a Rust-based AML (ACPI Machine Language) service that extends the existing ACPI service by
+providing a safe, ergonomic interface for parsing and modifying AML bytecode in firmware during the DXE phase.
+The implementation mirrors the ACPI SDT protocol’s functionality for AML traversal and patching
+and is designed mainly for firmware use rather than OS-level interpretation.
+It defines a structured system of AML handles for navigating AML streams and a trait-based `AmlParser` service for operations
+ such as opening tables, iterating operands, modifying values, and traversing child or sibling nodes.
+ The goal is to replace legacy C-based AML handling with a type-safe Rust service that supports ACPI 2.0+.
+ Future extensions may include extending this infrastructure for application-side AML interpretation 
+ within a `patina-acpi` crate.
```
os-d:
```suggestion
such as opening tables, iterating operands, modifying values, and traversing child or sibling nodes.
The goal is to replace legacy C-based AML handling with a type-safe Rust service that supports ACPI 2.0+.
Future extensions may include extending this infrastructure for application-side AML interpretation 
within a `patina-acpi` crate.
```

docs/src/rfc/text/0017-aml-parser.md
os-d: An additional thought here being perhaps parts of the `acpi` crate could be split out into this utility crate, reducing the amount of new code needed? Though as John says, I suspect this is not doing any writing/parsing, right?

docs/src/rfc/text/0017-aml-parser.md
```diff
@@ -0,0 +1,282 @@
...
+2. Open a DSDT or SSDT with `open_table`.
+3. Traverse as necessary through `get_child`, `get_sibling`, and `get_option`.
+4. Make necessary modifications through `set_option`.
+5. During traversal, close nodes which no longer need to be accessed through `close_handle`.
+6. When traversal / patching is complete, `close_handle` on the root node (originally opened with `open_table`).
+
+## Future Extensions
+
+Eventually, the hope is to provide not only firmware-side implementation of the ACPI SDT protocol,
+but also application-side AML interpretation and execution capabilities through an independent `patina-acpi` crate.
```
os-d: Do we really want interpretation and exececution in FW?

docs/src/rfc/text/0017-aml-parser.md
```diff
@@ -0,0 +1,282 @@
...
+    ...
+}
+```
+
+### AML Trait Interface
+
+The `AmlParser` service generally derives from the ACPI SDT protocol, and allows for traversal of the AML object tree.
+
+```rust
+pub(crate) trait AmlParser {
```
os-d: This feels, and as described, a direct derivation from the SDT protocol, which is obviously needed, though is a very C approach. For the Rust side, would it make sense to just return an iterator that does the get_child/get_sibling in the iteration itself? Feels like that would be more useful and safer. Then its debatable about whether you have an iter_mut that lets the caller directly change it or you force them to call set_option from the iter once they find what they want.

But, in general, trying to think about how we can ensure that this API is defined for a Rust interface and the C protocol is supported but not the primary interface driver.

docs/src/rfc/text/0017-aml-parser.md
joschock: err... I realized after re-reading the introduction that this might already be in view with the mentioned `patina-acpi` crate as a future extension, so take this comment as a "I think we should build it that way to start with."

docs/src/rfc/text/0017-aml-parser.md
joschock: Not in firmware, per-se (although... why not? - forget rust, lets code drivers in ASL :P). but it would be most excellent to be able to write unit tests for ASL using this as an execution framework as a unit test executor.

docs/src/rfc/text/0017-aml-parser.md
berlin-with0ut-return: (if i'm understanding both of you correctly), you are right that the "base" level of AML parsing can be standalone functionality. the patching/traversal functionality i'm writing here is unique, but any AML-related code whether on the firmware or application side needs to do many of the same basic things, ie getting package lengths, reading name strings, etc.

i have tried to contact the owner of `aml` to see if he wants to make this shared functionality publicly accessible (in which case we could simply take `aml` as a dependency) - no response as of yet. but otherwise i can borrow code snippets and repurpose for our crate (`aml` is MIT-licensed`). 

i guess as of now i haven't decided whether this functionality is universal/complex enough to warrant its own crate.

docs/src/rfc/text/0017-aml-parser.md
berlin-with0ut-return: i think there was some brief talk about adding application-side AML functionality to Patina, but i am not sure if this is the right path forward; either way this will be long in the future after the firmware AML protocols are implemented in Rust. (plus the Rust `acpi`  crate already covers lots of the interpretation/execution functionality.)

berlin-with0ut-return:
i am moving this to FCP and will close probably by EOW

makubacki:
@DanielMaddy, I thought you've mentioned some feedback. Have you read the RFC?

berlin-with0ut-return:
> @DanielMaddy, I thought you've mentioned some feedback. Have you read the RFC?

based on a discussion with @DanielMaddy @joschock offline, we discussed 2 main improvements that are being considered on the C side that should be brought to the Rust component

- a helper library that confers the ability to patch a table's AML as a buffer, before publishing. This will coexist with the normal SDT protocol which patches AML starting from the top-level XSDT.
- improved error and type checking, especially validating AML types, sizes, and syntax during patching. (ie - if the region getting patched is a NameSeg of pkglen X, the paramter passed in to patch with must also be a NameSeg of pkglen X). - this is somewhat handled by the `AmlOperand` struct

as a future extension we would like to move away completely from the C SDT protocol as it is difficult to use. right now the Rust API is constrained by the SDT protocol, but with the aforementioned helper library and other improvements we can hopefully transition to a better API once platforms move away from the SDT protocol

docs/src/rfc/text/0017-aml-parser.md
```diff
@@ -0,0 +1,322 @@
...
+  fn set_option(&self, handle: AmlHandle, idx: usize, new_val: AmlOperand) -> Result<(), AmlError>;
+
+  // Returns the first child of an AML node. 
+  fn get_children(&self, handle: AmlHandle) -> Result<impl Iterator<AmlHandle>, AmlError>;
+
+  // Returns the next sibling of an AML node.
+  fn get_sibling(&self, handle: AmlHandle) -> Result<impl Iterator<AmlHandle>, AmlError>;
+
+  // The above two functions are intended to provide a complete traversal implementation.
+  // For example, to get all the children of a node, find the first child through `get_child`, then use `get_sibling` on each subsequent child. In both cases, `None` indicates no child/sibling.
```
os-d: Does it make sense to provide a generic iterator? That will iterate the entire tree without having to do it piecemeal with getting the children, then the siblings of each child, etc.

berlin-with0ut-return: do you mean iterating the tree through something like a BFS/level traversal?

os-d: Yeah. Basically, it seems odd to me to say, I want to traverse the whole tree! Let me get a child, now its siblings, now their children, etc. instead of just

```rust
for node in tree:
   // whatever
```

Because IIUC, right now you would do:

```rust
root = get_root();
while true:
   // do something with root
  siblings = get_siblings(root);
  for node in siblings:
    // do something with node
    recurse(node); // or whatever
```

I.e. it seems much simpler for a consumer to have an iterator over the whole tree instead of breaking it up. Perhaps I am misunderstanding the use case and it is more valuable to have this broken up?

docs/src/rfc/text/0017-aml-parser.md
```diff
@@ -0,0 +1,331 @@
+# RFC: `ACPI SDT AML Handler`
+
+This RFC proposes a Rust-based AML (ACPI Machine Language) service that extends the existing ACPI service by
+providing a safe, ergonomic interface for parsing and modifying AML bytecode in firmware during the DXE phase.
+The implementation mirrors the ACPI SDT protocol’s functionality for AML traversal and patching
+and is designed mainly for firmware use rather than OS-level interpretation.
```
makubacki:
```suggestion
and is designed mainly for firmware use rather than OS-level interpretation of AML bytecode.
```

docs/src/rfc/text/0017-aml-parser.md
```diff
@@ -0,0 +1,331 @@
...
+
+## Motivation
+
+This RFC is an extension of the [ACPI service](0005-acpi.md).
+Similar to the ACPI service, this Rust-based AML service will provide a safer and
+more ergonomic interface for parsing and modifying AML bytecode.
+
+## Technology Background
+
+AML bytecode is encoded mainly in the body of the DSDT and SSDT.
```
makubacki: While the DSDT is a single table, a system can have zero, one, or multiple SSDTs.

```suggestion
Compiled ACPI tables such as the DSDT and SSDTs are composed of AML bytecode.

```

docs/src/rfc/text/0017-aml-parser.md
```diff
@@ -0,0 +1,331 @@
...
+which includes basic functionality for parsing and patching AML bytecode after ACPI tables are installed.  
+
+## Goals
+
+Provide a comprehensive Rust implementation for AML parsing and patching on the firmware DXE side.
+Secondarily, implement the rest of the ACPI SDT protocol relating to AML functionality.  
+
+## Requirements
+
+1. Redesign the existing C firmware AML implementation into a safe, easy-to-use Rust service.
```
makubacki:
```suggestion
1. Redesign existing C firmware AML interaction functionality into a safe, easy-to-use Rust service.
```

docs/src/rfc/text/0017-aml-parser.md
```diff
@@ -0,0 +1,331 @@
...
+while using the `acpi` crate as a dependency and parsing through its public interfaces.
+(This may involve contributing to the `acpi` crate to improve its public interfaces.)
+    - Pros: Less repeated code, especially for parsing.
+    - Cons: `acpi` has limited public interfaces, which may constrain the development of the Rust ACPI service.
+It primarily focuses on looking up and executing AML in the application space,
+with less support for actually walking through and modifying the firmware-side AML object tree.
+
+There is ongoing conversation with the owner of the `acpi` crate about
+borrowing certain implementations and modifying the public interfaces
+to be more friendly to the Patina ACPI implementation.
```
makubacki: Can you elaborate on what "more friendly to the Patina ACPI implementation" means in this context?

docs/src/rfc/text/0017-aml-parser.md
```diff
@@ -0,0 +1,331 @@
...
+        }
+    }
+}
+
+pub type AmlHandle = AmlSdtHandleInternal;
+```
+
+The `size` of an `AmlSdtHandleInternal` refers to its full size, including any children (`TermList`).
+
+The `offset` refers to its offset with the AML stream of the table.
```
makubacki:
```suggestion
The `offset` refers to its offset within the AML stream of the table.
```

docs/src/rfc/text/0017-aml-parser.md
```diff
@@ -0,0 +1,331 @@
...
+pub enum AmlOperand {
+    Opcode(AmlOpcode),
+    Name(AmlNameString), // Represents a NameString (AML path). Not to be confused with a string literal 
+    ...
+}
+```
+
+### AML Trait Interface
+
+The `AmlParser` service generally derives from the ACPI SDT protocol, and allows for traversal of the AML object tree.
```
makubacki: It's been mentioned elsewhere, but I'm not sure that the protocol interface should serve as the primary design point. That was a PI-spec C interface. For the purpose of the RFC, I'd consider inverting this. Define an independent, idiomatic Rust service interface and have the protocol be a wrapper around that.

docs/src/rfc/text/0017-aml-parser.md
```diff
@@ -0,0 +1,331 @@
...
+}
+
+pub enum AmlOperand {
+    Opcode(AmlOpcode),
+    Name(AmlNameString), // Represents a NameString (AML path). Not to be confused with a string literal 
+    ...
+}
+```
+
+### AML Trait Interface
```
makubacki: Can you please provide some actual usage examples of parsing and modifying AML using the interface? You could create an example table or extract one using something like acpidump.

docs/src/rfc/text/0017-aml-parser.md
```diff
@@ -0,0 +1,331 @@
...
+it does not directly deal with ACPI structures like the XSDT.
+Instead, it receives a generic buffer that represents AML bytecode beginning at the start of some ACPI table.
+The advantage of this paradigm is that this code can be reused for other parsing applications,
+and can be reused partially *from* the existing Rust `acpi` crate.
+
+This proposed `patina-acpi` utility crate crate shares functionality with the Rust  `os-dev acpi` crate; as such,
+it remains to be seen if collaboration is possible to expose this shared parsing functionality,
+or if `patina-acpi` will borrow/fork from `acpi`.
+
+### AML Handles
```
makubacki: In general, I think this RFC needs to describe in more detail how the AML services cooperate with the `AcpiProvider` service introduced in RFC 0005-acpi. In particular, how the table key is managed between the two.

For example, my understanding is that you use that service to install an ACPI table (`acpi_provider.install_acpi_table()`) which returns a unique table key. I may have missed it, but somewhere around here I think the connection to the table key in that service should be referenced.

Also, state management between the two. For example, if `acpi_provider.uninstall_acpi_table()` is called on a table key, does this AML service realize that key is not valid? Some simle info like this about the relationship between the two services would be helpful for putting together the whole ACPI design.

docs/src/rfc/text/0017-aml-parser.md
```diff
@@ -0,0 +1,331 @@
...
+  fn close_handle(&self, handle: AmlHandle) -> Result<(), AmlError>;
+
+  // Iterates over the options (operands) of an opened AML handle.
+  fn iter_options(&self, handle: AmlHandle) -> Result<Vec<AmlOperand>, AmlError>;
+
+  // Sets the option (operand) at a particular index to the given value.
+  fn set_option(&self, handle: AmlHandle, idx: usize, new_val: AmlOperand) -> Result<(), AmlError>;
+
+  // Iterates over all nodes of the tree (in a depth-first order).
+  fn iter(&self) -> Result<impl Iterator<AmlHandle>, AmlError>;
```
makubacki: My understanding is that by the `Result<impl Iterator<AmlHandle>, AmlError>` return type wrapping the entire iterator in a `Result`, the error return is happening when the iterator is created as opposed to during iteration where the iterator `Item` would be a `Result<AmlHandle>`. Like this:

```rust
    type Children: Iterator<Item = Result<AmlHandle, AmlError>>
    
    ...

    /// Returns an iterator over immediate children of this node.
    fn children(&self, handle: &AmlHandle) -> Self::Children;
```

docs/src/rfc/text/0017-aml-parser.md
```diff
@@ -0,0 +1,331 @@
...
+  // It points to the first (root) node in the AML stream.
+  unsafe fn open_table(&self, table_key: TableKey) -> Result<AmlHandle, AmlError>;
+
+  // Closes a handle for modification and traversal.
+  // The handle will no longer be valid after it is closed.
+  // Update the corresponding table's checksum if the handle is `modified`. 
+  fn close_handle(&self, handle: AmlHandle) -> Result<(), AmlError>;
+
+  // Iterates over the options (operands) of an opened AML handle.
+  fn iter_options(&self, handle: AmlHandle) -> Result<Vec<AmlOperand>, AmlError>;
```
makubacki: The function name is a little confusing. It may internally iterate over operands to return a vector of items, but that seems like more of an implementation detail. Would something like `get_operands()` be more appropriate?

docs/src/rfc/text/0017-aml-parser.md
```diff
@@ -0,0 +1,331 @@
...
+
+  // Sets the option (operand) at a particular index to the given value.
+  fn set_option(&self, handle: AmlHandle, idx: usize, new_val: AmlOperand) -> Result<(), AmlError>;
```
makubacki: I know the "option" nomenclature comes from the SDT protocol but it seems out of place here. For example, what about something like `set_operand()`?

docs/src/rfc/text/0017-aml-parser.md
```diff
@@ -0,0 +1,339 @@
...
+By spec definition, an AML handle is an opaque handle returned from opening a DSDT or SSDT,
+on which AML traversal and patching operations can be performed.
+
+Internally, each `AmlHandle` object, aliased as `AmlSdtHandleInternal`,
+represents a cursor within an AML stream. Each handle can be conceptualized as a "node"
+in the AML object tree of parent-child-relationships.
+
+```rust
+pub(crate) struct AmlSdtHandleInternal {
+    table_key: TableKey,
```
makubacki: Are public functions like `pub fn table_key(&self) -> TableKey` or `pub fn offset(&self) -> usize` needed? Sorry if I missed it, but is there an easy way to get that from `AmlHandle`? Do you think there needs to be?

docs/src/rfc/text/0017-aml-parser.md
DanielMaddy: I have been using the AcpiSdt Protocol, and it seems that something like the existing FindPath would be helpful to users. The function name seems to be less than ideal but the functionality of taking a valid ACPI path like "\\_SB.ABCD.MNOP.WXYZ" and returning WXYZ would make normal usages easier.

docs/src/rfc/text/0017-aml-parser.md
DanielMaddy: I think the option nomenclature may come from the AML specification, but I agree that it not so easy for users and a bit problematic. The issue seems to be rooted by the tree structure of named objects, but the objects themselves having unnamed parse trees under the objects. This leads to GetOption indexes that are hard find/understand. Presently the general usage flow is GetOption calls with nonobvious indexes, verify the returned buffers, and then SetOption calls.

docs/src/rfc/text/0017-aml-parser.md
berlin-with0ut-return: i can add public getters

docs/src/rfc/text/0017-aml-parser.md
berlin-with0ut-return: thinking about the uninstall case -- does it make sense for AmlHandle to hold a reference to the underlying `AcpiTable` to ensure it doesn't outlive it?

something like:

```rust
pub struct AmlHandle<'t> {
    table: &'t mut AmlTable,
    offset: usize,
    ....
}
```

makubacki: I think that's better for lifetime tracking but also accounting for the fact that a mutable reference exists to that data.

makubacki:
I approved the RFC and it's ultimately up to you, but I'd suggest first implementing this on a feature branch for feedback on the interfaces as tested against some realistic scenarios before moving to the main branch. If anything changes, the RFC can be amended easily without modifying public interfaces in the main branch.

berlin-with0ut-return:
i'll be implementing this as part of the feature/acpi branch.

PR #887, opened 2025-10-16
Add safety comments for patina_ffs module

joschock:
## Description

Add safety comments for patina_ffs module

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Comments only; passes unit tests.

## Integration Instructions

N/A

### DIFF
--- a/sdk/patina_ffs/src/file.rs
+++ b/sdk/patina_ffs/src/file.rs
@@ -78,7 +78,7 @@ impl<'a> FileRef<'a> {
             Err(FirmwareFileSystemError::InvalidHeader)?;
         }
 
-        // safety: buffer is large enough to contain file header.
+        // SAFETY: buffer is large enough to contain file header.
         let header = unsafe { ptr::read_unaligned(buffer.as_ptr() as *const file::Header) };
 
         // determine actual size and content_offset
@@ -94,7 +94,7 @@ impl<'a> FileRef<'a> {
                 if buffer.len() < mem::size_of::<file::Header2>() {
                     Err(FirmwareFileSystemError::InvalidHeader)?;
                 }
-                // safety: buffer is large enough to contain file header.
+                // SAFETY: buffer is large enough to contain file header.
                 let header = unsafe { ptr::read_unaligned(buffer.as_ptr() as *const file::Header2) };
                 (header.extended_size as usize, mem::size_of::<file::Header2>())
             }
@@ -384,7 +384,7 @@ impl File {
         file_header.extended_size = (mem::size_of_val(&file_header) + content.len()) as u64;
 
         // calculate checksum (excludes state and integrity_check_file, set to zero)
-        // safety: file_header is repr(C), safe to represent as byte slice for checksum
+        // SAFETY: file_header is repr(C), safe to represent as byte slice for checksum
         let header_slice =
             unsafe { from_raw_parts(&raw const file_header as *const u8, mem::size_of_val(&file_header)) };
         let sum = header_slice.iter().fold(0u8, |sum, value| sum.wrapping_add(*value));
@@ -405,6 +405,7 @@ impl File {
             file_header.header.state = !file_header.header.state;
         }
 
+        // SAFETY: file_header is repr(C), safe to represent as byte slice for serialization.
         let header_slice =
             unsafe { from_raw_parts(&raw const file_header as *const u8, mem::size_of_val(&file_header)) };
         header_slice.to_vec()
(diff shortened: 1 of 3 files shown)


sdk/patina_ffs/src/volume.rs
```diff
@@ -226,12 +226,19 @@ impl<'a> VolumeRef<'a> {
...
+        // SAFETY: caller must ensure that the base_address is safe to read for enough bytes to read
+        // the fv_header structure.
         let fv_header = unsafe { ptr::read_unaligned(base_address as *const fv::Header) };
         if fv_header.signature != u32::from_le_bytes(*b"_FVH") {
             // base_address is not the start of a firmware volume.
             return Err(FirmwareFileSystemError::DataCorrupt);
         }
-
+        // SAFETY: caller must ensure that the base_address is valid and points
+        // to memory a valid firmware volume as part of the safety contract for
```
os-d:
```suggestion
        // to memory that contains a valid firmware volume as part of the safety contract for
```
nit: while we are here, could keep the newline before this comment

sdk/patina_ffs/src/volume.rs
```diff
@@ -1023,6 +1034,7 @@ mod test {
         // bogus signature.
         let mut fv_bytes = fs::read(root.join("DXEFV.Fv"))?;
         let fv_header = fv_bytes.as_mut_ptr() as *mut fv::Header;
+        // SAFETY: Delibrately corrupting the FV header for test purposes.
```
os-d:
```suggestion
        // SAFETY: Deliberately corrupting the FV header for test purposes.
```

And below

os-d: @joschock , did the test code explicitly fail under the clippy lint? If so @Javagedes is it easy to exclude test code from this lint? A bit burdensome to say to do safety comments for test code, but not the end of the world if we need to.

makubacki: I don't think it's a big deal to require safety comments in test code for consistency (for humans following process and software analyzing code).

sdk/patina_ffs/src/file.rs
```diff
@@ -78,7 +78,7 @@ impl<'a> FileRef<'a> {
             Err(FirmwareFileSystemError::InvalidHeader)?;
         }
 
-        // safety: buffer is large enough to contain file header.
+        // SAFETY: buffer is large enough to contain file header.
```
Javagedes, line 81: Off topic - Just looking at this, this looks like a good candidate for zerocopy. Do you concur? If so, I can add an issue.

sdk/patina_ffs/src/volume.rs
joschock: @os-d - I added `#![warn(clippy::undocumented_unsafe_blocks)]` at the module level and just fixed everything that triggered (I didn't check that in since I assumed we'll eventually enable it globally). If we do enable it that way, then yes, it would trigger for test code.

sdk/patina_ffs/src/file.rs
joschock: Yes, agree.

sdk/patina_ffs/src/volume.rs
```diff
@@ -1031,6 +1043,7 @@ mod test {
         // bogus header_length.
         let mut fv_bytes = fs::read(root.join("DXEFV.Fv"))?;
         let fv_header = fv_bytes.as_mut_ptr() as *mut fv::Header;
+        // SAFETY: Delibrately corrupting the FV header for test purposes.
```
Javagedes: @makubacki @os-d @joschock @cfernald 

Do we have opinions on unsafe blocks in testing? We could just as simply do. I don't have a huge opinion, just throwing it out there.

```rust
#[cfg(test)]
#[allow(clippy::undocumented_unsafe_blocks)]
mod tests {
  ...
}
```

Javagedes (approved):

sdk/patina_ffs/src/volume.rs
makubacki: I left a similar comment in https://github.com/OpenDevicePartnership/patina/pull/887#discussion_r2436427168.

I'm not as concerned about the code being documented but I do feel the overhead to accommodate the linter, impact to static analysis tools that check for undocumented unsafe blocks, and human explanation to a more complex policy is not worth it just to avoid putting comments in test code.

makubacki: That being said, I also don't feel really strongly about this. If someone else does, I think it's also reasonable to allow test writers to use `allow(clippy::undocumented_unsafe_blocks)`. The main thing is that clippy results are clean.

Javagedes: I too have no opinion and leave it up to John if he just wants to close this issue:)

sdk/patina_ffs/src/volume.rs
joschock: Fixed. Thanks!

sdk/patina_ffs/src/volume.rs
joschock: Fixed the spelling issues - made it one place, and copilot helpfully copied it everywhere else.

sdk/patina_ffs/src/volume.rs
joschock: I'm good with leaving it enabled; test code is often an example, so having a safety comment describing what you are doing is useful to people who are looking at the test case trying to figure out if this is the "right" way to do something.

os-d: I’m fine to leave it. Test code is typically doing bad things, but that’s easy to document.

PR #892, opened 2025-10-16
patina_mm: Add missing safety comments

makubacki:
## Description

Adds missing safety comments to the `patina_mm` crate.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- `cargo make all`
- `patina_mm` results from `cargo clippy --all-targets -- -W clippy::undocumented_unsafe_blocks`

## Integration Instructions

- N/A

### DIFF
--- a/components/patina_mm/src/component/communicator.rs
+++ b/components/patina_mm/src/component/communicator.rs
@@ -67,12 +67,10 @@ impl MmExecutor for RealMmExecutor {
     #[coverage(off)]
     fn execute_mm(&self, _comm_buffer: &mut CommunicateBuffer) -> Result<(), Status> {
         log::debug!(target: "mm_comm", "Triggering SW MMI for MM communication");
-        unsafe {
-            self.sw_mmi_trigger_service.trigger_sw_mmi(0xFF, 0).map_err(|err| {
-                log::error!(target: "mm_comm", "SW MMI trigger failed: {:?}", err);
-                Status::SwMmiFailed
-            })
-        }
+        self.sw_mmi_trigger_service.trigger_sw_mmi(0xFF, 0).map_err(|err| {
+            log::error!(target: "mm_comm", "SW MMI trigger failed: {:?}", err);
+            Status::SwMmiFailed
+        })
     }
 }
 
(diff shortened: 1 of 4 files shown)


components/patina_mm/src/component/communicator.rs
```diff
@@ -67,6 +67,7 @@ impl MmExecutor for RealMmExecutor {
     #[coverage(off)]
     fn execute_mm(&self, _comm_buffer: &mut CommunicateBuffer) -> Result<(), Status> {
         log::debug!(target: "mm_comm", "Triggering SW MMI for MM communication");
+        // SAFETY: SW MMI trigger is a platform-provided service with inherent hardware access
```
Javagedes: I'm a bit confused by the safety requirements for `trigger_sw_mmi`. The trait's safety requirements are that the trait itself is responsible for verifying that the system is in a safe state to handle MMIs, however `trigger_sw_mmi` then states that it is the caller's responsibility to ensure that the system is in this safe state.

To me, only one of those two safety requirements need to be met. Either it is up to the implementor of the trait, or it is up to the caller of the trait.

If the safety requirements are distinctly different between the two, then I think the documented safety requirements need to be updated for `SwMmiTrigger`, as I cannot personally tell a difference between the two requirements.

The SAFETY comment I'm attaching my comment to is basically saying this implementation is valid because the platform service implementor says so. All of this is telling me that the `unsafe` can be removed form `trigger_sw_mmi` function, and keep the safety requirements at the trait level, for the trait (platform) implementor to satisfy.

However, if they are separate, and what your documentation is saying is that we are in a valid state because we are being called by the `RealMmExecutor`, then that struct Invariants need to be documented in the struct itself, so that this safety comment can simply something along the lines of `<safety requirement> is satisfied by the nature of this structs invariants`

makubacki: I kept the `SmMmiTrigger` trait as unsafe and updated its docs. I removed unsafe from the `trigger_sw_mmi` method.

makubacki: @Javagedes, I'll leave this thread open for you to close after you can revisit the PR.

makubacki: Going to close it since the issue is addressed.

PR #894, opened 2025-10-16
[feature/esrt-component] add ESRT component scaffold with TODO placeholders

kat-perez:
## Description

add component scaffold for ESRT component

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

n/a

## Integration Instructions

n/a

### DIFF
--- a/components/patina_esrt/examples/integration.rs
+++ b/components/patina_esrt/examples/integration.rs
@@ -1,0 +1,24 @@
+//! Platform integration example for ESRT component
+//!
+//! ## License
+//!
+//! Copyright (c) Microsoft Corporation.
+//!
+//! SPDX-License-Identifier: Apache-2.0
+
+//! TODO: Developer 2 (Phase 4) - Create complete example
+//! Hint: Show component initialization
+//! Hint: Show registering Non-FMP entries
+//! Hint: Show FMP synchronization
+//! Hint: Show entry updates
+//! Hint: Show repository locking
+//! Hint: Show error handling
+fn main() {
+    // TODO: Add example code demonstrating:
+    // 1. Component setup with configuration
+    // 2. Registering system firmware entry
+    // 3. Syncing FMP entries
+    // 4. Updating entry after firmware update
+    // 5. Locking repository before OS handoff
+    // 6. Error handling patterns
+}
--- a/components/patina_esrt/src/service.rs
+++ b/components/patina_esrt/src/service.rs
@@ -1,0 +1,71 @@
+//! EsrtRecords Service Trait
+//!
+//! Service trait for managing ESRT entries. This trait is implemented by the
+//! ESRT component and can be used by other components or drivers.
+//!
+//! ## License
+//!
+//! Copyright (c) Microsoft Corporation.
+//!
+//! SPDX-License-Identifier: Apache-2.0
+
+use crate::{error::EsrtError, types::SystemResourceEntry};
+use r_efi::base::Guid;
+
+/// Service trait for ESRT record management.
+///
+/// This service allows platform code and other components to register,
+/// update, and query ESRT entries for firmware update support.
+pub trait EsrtRecords {
+    /// Register a new ESRT entry.
+    ///
+    /// # Errors
+    /// - `EsrtError::AlreadyExists` if an entry with the same fw_class exists
+    /// - `EsrtError::OutOfResources` if the repository is full
+    /// - `EsrtError::WriteProtected` if the repository is locked
+    fn register_entry(&mut self, entry: SystemResourceEntry) -> Result<(), EsrtError>;
+
+    /// Replace an existing ESRT entry with a new entry.
+    ///
+    /// This function completely replaces the existing entry with the provided entry,
+    /// which will cause the old entry to be dropped. If `SystemResourceEntry` contains
+    /// resources that require special drop handling, consider using field-by-field
+    /// update methods instead.
+    ///
+    /// # Parameters
+    /// - `entry`: The new entry that will replace the existing one
+    ///
+    /// # Returns
+    /// - `Ok(())` if the entry was successfully replaced
+    ///
+    /// # Errors
+    /// - `EsrtError::NotFound` if no entry with matching fw_class exists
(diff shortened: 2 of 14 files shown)


components/patina_esrt/src/component/esrt.rs
```diff
@@ -0,0 +1,73 @@
...
+#[derive(IntoComponent)]
+pub struct Esrt;
+
+impl Esrt {
+    /// Entry point of the ESRT component.
+    ///
+    /// This follows Patina's component pattern similar to `Performance::entry_point`.
+    pub fn entry_point(
+        self,
+        _config: Config<EsrtConfig>,
```
Javagedes: nit: public Config (`Config<T>`, `ConfigMut<T>`) is meant for configuration values that may be consumed by multiple components, or mutated by other components before being locked and consumed by this component. If this configuration is for the `Esrt` component only (and hardcoded), it is probably (but not necessarily) better for it to be private config, which is config set during the instantiation of the `Esrt` component itself.

e.g.

```rust
#[derive(IntoComponent)]
pub struct Esrt {
  max_fmp_entries: u32
}

impl Esrt {
  fn new(max_fmp_entries: u32) -> Self {
    Self { max_fmp_entries }
  }
  fn with_max_fmp_entries(mut self, max: u32) -> Self {
    self.max_fmp_entries = max;
    self
  }
}

impl Default for Esrt {
  fn default() -> Self {
    Self::new(32)
  }
}

// Misc examples
fn main() {
  Core::default()
    .init_memory(hobs)
    .with_component(Esrt::new(64))
    .with_component(Esrt::default().with_max_fmp_entries(128))
}
```

It is ultimetly up to you if it is necessary to leave it as a public config.

components/patina_esrt/src/service.rs
```diff
@@ -0,0 +1,59 @@
...
+    /// - `EsrtError::OutOfResources` if the repository is full
+    /// - `EsrtError::WriteProtected` if the repository is locked
+    fn register_entry(&mut self, entry: SystemResourceEntry) -> Result<(), EsrtError>;
+
+    /// Update an existing ESRT entry.
+    ///
+    /// # Errors
+    /// - `EsrtError::NotFound` if no entry with matching fw_class exists
+    /// - `EsrtError::WriteProtected` if the repository is locked
+    fn update_entry(&mut self, entry: SystemResourceEntry) -> Result<(), EsrtError>;
```
Javagedes, line 45: Depending on the full implementation of `SystemResourceEntry` I am slightly concerned about the naming of this function and the implications of that name.

`update_entry` typically indicates that we are in-place editing the existing value. This means that we do not `Drop` the existing value. However based off the interface, it appears you are going to actually replace the existing value with the new value. This will actually `drop` the existing value.

This is only a concern if the finalized `SystemResourceEntry` has a field that, by dropping it, will cause issues. If that is not going to happen, then this concern can probably be ignored.

If it is a concern, then I could see it going two ways:

1. `replace_entry` and return the replaced value
2. Change the interface to take a reference to the replacing entry, then compare the in-memory entry with the passed in entry, field by field, replacing any that are different.

Javagedes (approved):

components/patina_esrt/src/component/esrt.rs
kat-perez: i agree with you as it as a private config, noted this in the RFC

components/patina_esrt/src/service.rs
kat-perez: I will keep the name here but have added documentation to explain that it performs a complete replacement, which will drop the old entry. I don't think `SystemResourceEntry` contains resources that would require special drop handling. The `EFI_SYSTEM_RESOURCE_ENTRY` struct is

```
typedef struct {
  EFI_GUID    FwClass;                    // 16-byte GUID (Copy type)
  UINT32      FwType;                     // Firmware type enum (Copy type)
  UINT32      FwVersion;                  // Version number (Copy type)
  UINT32      LowestSupportedFwVersion;   // Version number (Copy type)
  UINT32      CapsuleFlags;               // Bitfield flags (Copy type)
  UINT32      LastAttemptVersion;         // Version number (Copy type)
  UINT32      LastAttemptStatus;          // Status code (Copy type)
} EFI_SYSTEM_RESOURCE_ENTRY;
```

All the fields are primitive types so I think it should be fine to not return the replaced value

components/patina_esrt/examples/integration.rs
rogurr, line 1: Do we need a license in every .rs file?

patina_dxe_core/src/allocator/usage_tests.rs
rogurr, line 1: Not a bug, just for my learning.

Why do we use a mod.rs file to define the esrt.rs as the primary module file?  Isn't there a way in the .toml file to force the root file to be different than mod, lib, main, etc.?

Or is this the precedence set by the Patina project?

components/patina_esrt/examples/integration.rs
kat-perez: oh yes, I'll add those

patina_dxe_core/src/allocator/usage_tests.rs
kat-perez: actually this is a good catch, i should have named this component.rs (matching what patina_performance is doing). I think lib.rs is technically the "main" file of this crate. As for the design rationale, I didn't see it explicitly documented anywhere but I was following the structure of the other components in this directory. I'm guessing it's so that it's easy to see from lib.rs what are "public" and "private" modules and what is exported from each module

makubacki: This document has a good explanation of the module system https://doc.rust-lang.org/book/ch07-02-defining-modules-to-control-scope-and-privacy.html.

In most cases, we shouldn't use mod.rs any longer and instead have a file that matches the module name instead. I still have to catch myself occasionally to do that instead of use mod.rs.

kat-perez: will rework this to have a file to match the module name instead

PR #895, opened 2025-10-16
Add safety comments for patina_internal_device_path

apop5:
## Description

Add safety comments for patina_internal_device_path

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested
Comments only. 
Passed `cargo make clippy`
with Makefile.toml modified to include `"-W", "clippy::undocumented_unsafe_blocks"`

## Integration Instructions
No integration necessary.

### DIFF
--- a/components/patina_mm/src/component/communicator.rs
+++ b/components/patina_mm/src/component/communicator.rs
@@ -67,12 +67,10 @@ impl MmExecutor for RealMmExecutor {
     #[coverage(off)]
     fn execute_mm(&self, _comm_buffer: &mut CommunicateBuffer) -> Result<(), Status> {
         log::debug!(target: "mm_comm", "Triggering SW MMI for MM communication");
-        unsafe {
-            self.sw_mmi_trigger_service.trigger_sw_mmi(0xFF, 0).map_err(|err| {
-                log::error!(target: "mm_comm", "SW MMI trigger failed: {:?}", err);
-                Status::SwMmiFailed
-            })
-        }
+        self.sw_mmi_trigger_service.trigger_sw_mmi(0xFF, 0).map_err(|err| {
+            log::error!(target: "mm_comm", "SW MMI trigger failed: {:?}", err);
+            Status::SwMmiFailed
+        })
     }
 }
 
@@ -541,6 +539,7 @@ mod tests {
         mock_executor.expect_execute_mm().times(1).returning(|comm_buffer| {
             // Simulate MM handler corrupting the buffer state by directly writing to memory
             // This should be caught by the state verification
+            // SAFETY: Test intentionally corrupts buffer to verify error detection
             unsafe {
                 let ptr = comm_buffer.as_ptr();
                 *ptr = 0xFF; // Corrupt the first byte of the header
--- a/components/patina_mm/src/component/sw_mmi_manager.rs
+++ b/components/patina_mm/src/component/sw_mmi_manager.rs
@@ -35,17 +35,15 @@ use mockall::automock;
 ///
 /// ## Safety
 ///
-/// This trait is unsafe because an implementation needs to ensure that the service is only invoked after hardware
-/// initialization for MMIs is complete and that the system is in a safe state to handle MMIs.
+/// This trait is unsafe because an implementation needs to:
+///
+/// 1. Ensure that the platform hardware is capable of handling MMIs.
+/// 2. Ensure that the service is only invoked after hardware initialization for MMIs is complete and that the
+///    system is in a safe state to handle MMIs.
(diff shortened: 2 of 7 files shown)


core/patina_internal_device_path/src/lib.rs
```diff
@@ -225,7 +228,9 @@ pub fn remaining_device_path(
     let mut b_ptr = b;
     let mut node_count = 0;
     loop {
+        // SAFETY: Caller must ensure pointer is a valid device_path
         let a_node = unsafe { *a_ptr };
+        // SAFETY: Caller must ensure pointer is a valid device_path
         let b_node = unsafe { *b_ptr };
```
joschock: minor aesthetic - consider:

```suggestion
        // SAFETY: Caller must ensure pointers are valid device_paths
        let (a_node, b_node) = unsafe { (*a_ptr, *b_ptr) };
```

core/patina_internal_device_path/src/lib.rs
```diff
@@ -236,7 +239,12 @@ pub fn remaining_device_path(
 
         let a_length: usize = u16::from_le_bytes(a_node.length).into();
         let b_length: usize = u16::from_le_bytes(b_node.length).into();
+        // SAFETY: caller must assure that device path is valid and that memory will remain
```
makubacki: The function safety comment is currently `/// a and b inputs must be a valid pointers to well-formed device paths`.

The expectation for the lifetime of the data behind the pointer should probably be moved to the function level doc.

core/patina_internal_device_path/src/lib.rs
```diff
@@ -245,14 +253,17 @@ pub fn remaining_device_path(
...
         a_ptr = unsafe { a_ptr.byte_offset(a_offset) };
+        // SAFETY: Caller must ensure that the device path is well formed and valid
         b_ptr = unsafe { b_ptr.byte_offset(b_offset) };
     }
 }
 
 /// Determines whether the given device path points to an end-of-device-path node.
 pub fn is_device_path_end(device_path: *const efi::protocols::device_path::Protocol) -> bool {
     let node_ptr = device_path;
+    // SAFETY: Caller must assure that device_path is valid and aligned
```
makubacki: Technically, the function should be unsafe with this added so it's visible to the caller as part of the caller contract.

core/patina_internal_device_path/src/lib.rs
```diff
@@ -225,8 +228,8 @@ pub fn remaining_device_path(
     let mut b_ptr = b;
     let mut node_count = 0;
     loop {
-        let a_node = unsafe { *a_ptr };
-        let b_node = unsafe { *b_ptr };
+        // SAFETY: Caller must ensure pointers are valid device_paths
```
makubacki, line 233: This function should also be unsafe with the caller requirements called out in the function docs.

core/patina_internal_device_path/src/lib.rs
apop5: Updated in latest.

core/patina_internal_device_path/src/lib.rs
apop5: Updated in latest.

core/patina_internal_device_path/src/lib.rs
apop5: Updated in latest.

core/patina_internal_device_path/src/lib.rs
```diff
@@ -524,7 +544,7 @@ mod tests {
         let device_path_c = device_path_c_bytes.as_ptr() as *const efi::protocols::device_path::Protocol;
 
         // a is a prefix of b.
-        let result = remaining_device_path(device_path_a, device_path_b);
+        let result = unsafe { remaining_device_path(device_path_a, device_path_b) };
```
makubacki, line 547: Does this get called out for not having a safety comment? We've started adding them for unsafe blocks in tests too.

apop5: yes

`error[E0133]: call to unsafe function `remaining_device_path` is unsafe and requires unsafe block
   --> core\patina_internal_device_path\src\lib.rs:547:22
    |
547 |         let result = remaining_device_path(device_path_a, device_path_b);
    |                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ call to unsafe function
    |
    = note: consult the function's documentation for information on how to avoid undefined behavior`

PR #906, opened 2025-10-17
patina_smbios: add SMBIOS component [Rebase && FF]

kat-perez:
## Description

adds a new component for managing SMBIOS tables and adding SMBIOS records with the following features:

- SMBIOS 3.x configuration table publication
- C-compatible protocol interface for platform integration
- Type 0 (Platform Firmware), Type 1 (System), Type 127 (End-of-Table) records
- TplMutex for timer-interrupt-safe synchronization
- Comprehensive unit tests (90 tests, 75%+ coverage)

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [x] Includes documentation?

(skip cargo release dry run)

## How This Was Tested

Make sure everything builds with `cargo make build` and then integrate the component into patina-dxe-core-qemu and test that it runs with patina-qemu

## Integration Instructions

Integrated with an example platform code in patina-dxe-core-qemu and ran with the EDK2 driver removed in patina-qemu. The platform component adds sample records and can read records in UEFI Shell with `smbiosview`

Co-authored-by: Anh Luong <anh.luong@dell.com>

### DIFF
--- a/components/patina_samples/src/lib.rs
+++ b/components/patina_samples/src/lib.rs
@@ -18,3 +18,9 @@
 #![feature(coverage_attribute)]
 
 pub mod component;
+
+#[cfg(feature = "enable_smbios_example")]
+mod smbios_platform;
+
+#[cfg(feature = "enable_smbios_example")]
+pub use smbios_platform::SmbiosExampleComponent;
--- a/components/patina_samples/src/smbios_platform.rs
+++ b/components/patina_samples/src/smbios_platform.rs
@@ -1,0 +1,383 @@
+//! SMBIOS Platform Example Component
+//!
+//! This component demonstrates the recommended SMBIOS integration pattern:
+//! 1. Uses the type-safe `add_record<T>()` API for adding SMBIOS records
+//! 2. Platform component publishes the table after all records are added
+//! 3. Uses structured record types (Type0, Type1, Type2, Type3, Type127, and custom OEM records)
+//! 4. Shows how to create custom vendor-specific SMBIOS records (Type 0x80-0xFF)
+//!
+//! ## Usage
+//!
+//! In your platform's DXE core initialization:
+//!
+//! ```ignore
+//! use patina_smbios::{SmbiosConfiguration, SmbiosProvider};
+//! use patina_samples::SmbiosExampleComponent;
+//!
+//! Core::default()
+//!     // Configure SMBIOS version
+//!     .with_config(SmbiosConfiguration { major_version: 3, minor_version: 9 })
+//!     // Add SMBIOS provider component
+//!     .with_component(SmbiosProvider::new())
+//!     // Add your platform SMBIOS component
+//!     .with_component(SmbiosExampleComponent::new())
+//!     // ... rest of your components
+//! ```
+//!
+//! ## License
+//!
+//! Copyright (c) Microsoft Corporation.
+//!
+//! SPDX-License-Identifier: Apache-2.0
+
+extern crate alloc;
+
+use alloc::string::String;
+use alloc::vec;
+use patina::boot_services::StandardBootServices;
+use patina::component::IntoComponent;
+use patina::component::service::Service;
+use patina::error::Result;
+use patina_smbios::service::{SMBIOS_HANDLE_PI_RESERVED, SmbiosTableHeader};
+use patina_smbios::smbios_record::{
+    SmbiosRecordStructure, Type0PlatformFirmwareInformation, Type1SystemInformation, Type2BaseboardInformation,
+    Type3SystemEnclosure, Type127EndOfTable,
+};
+use patina_smbios::{Smbios, SmbiosError};
+
+/// Example custom vendor-specific SMBIOS record
+///
+/// Demonstrates how to create custom SMBIOS records for vendor-specific data.
+/// SMBIOS types 0x80-0xFF are reserved for OEM/vendor use.
+///
+/// This example shows:
+/// - Defining a custom struct with SmbiosTableHeader
+/// - Implementing SmbiosRecordStructure trait for serialization
+/// - Using a custom type code (0x80) in the vendor range
+/// - Adding custom fields and string pools
+///
+/// # Important: Not C-Compatible
+///
+/// Like standard SMBIOS records, this struct contains a `string_pool: Vec<String>`
+/// which is Rust metadata and NOT part of the SMBIOS binary format. Always use
+/// `to_bytes()` to convert to proper SMBIOS format.
+pub struct VendorOemRecord {
+    pub header: SmbiosTableHeader,
+    pub oem_field: u32, // Example 32-bit vendor-specific field
+    pub string_pool: alloc::vec::Vec<String>,
+}
+
+impl SmbiosRecordStructure for VendorOemRecord {
+    const RECORD_TYPE: u8 = 0x80; // Vendor-specific type (0x80-0xFF reserved for OEM)
+
+    fn to_bytes(&self) -> alloc::vec::Vec<u8> {
+        let mut bytes = alloc::vec::Vec::new();
+
+        // Calculate structured data size (header + oem_field)
+        let structured_size = core::mem::size_of::<SmbiosTableHeader>() + core::mem::size_of::<u32>();
+
+        // Serialize header manually (packed struct: type(u8) + length(u8) + handle(u16))
+        bytes.push(Self::RECORD_TYPE);
+        bytes.push(structured_size as u8);
+        bytes.extend_from_slice(&self.header.handle.to_le_bytes());
+
+        // Serialize oem_field (little-endian u32)
+        bytes.extend_from_slice(&self.oem_field.to_le_bytes());
+
+        // Serialize string pool using the public helper
+        bytes.extend_from_slice(&patina_smbios::smbios_record::serialize_string_pool(&self.string_pool));
+
+        bytes
+    }
+
+    fn validate(&self) -> core::result::Result<(), SmbiosError> {
+        // Validate string lengths
+        for string in &self.string_pool {
+            if string.len() > patina_smbios::SMBIOS_STRING_MAX_LENGTH {
+                return Err(SmbiosError::StringTooLong);
+            }
+        }
+        Ok(())
+    }
+
+    fn string_pool(&self) -> &[String] {
+        &self.string_pool
+    }
+
+    fn string_pool_mut(&mut self) -> &mut alloc::vec::Vec<String> {
+        &mut self.string_pool
+    }
+}
+
+/// Example Platform SMBIOS Component
+///
+/// Demonstrates a real-world platform component that adds SMBIOS records
+/// using the type-safe `add_record<T>()` API.
+///
+/// This example shows:
+/// - Type 0: BIOS/Firmware information
+/// - Type 1: System information with UUID and serial numbers
+/// - Type 2: Baseboard/motherboard information
+/// - Type 3: System chassis/enclosure information
+/// - Type 0x80: Custom vendor-specific record (OEM range)
+/// - Type 127: End-of-table marker
+/// - Publishing the complete table to UEFI Configuration Table
+#[derive(IntoComponent, Default)]
+pub struct SmbiosExampleComponent;
+
+impl SmbiosExampleComponent {
+    /// Create a new SMBIOS platform component
+    pub fn new() -> Self {
+        Self
+    }
+
+    /// Platform SMBIOS initialization entry point
+    ///
+    /// Called by the component framework during DXE phase initialization.
+    /// The framework automatically provides the SMBIOS service as a dependency.
+    #[cfg(feature = "enable_smbios_example")]
+    fn entry_point(self, smbios: Service<Smbios>, boot_services: StandardBootServices) -> Result<()> {
+        log::info!("=== Example Platform SMBIOS Component ===");
+
+        // Verify SMBIOS version
+        let (major, minor) = smbios.version();
+        log::info!("SMBIOS Version: {}.{}", major, minor);
+
+        // Add platform-specific SMBIOS records
+        log::info!("Creating platform SMBIOS records...");
+
+        // Type 0: Platform firmware/BIOS information
+        if let Err(e) = Self::add_bios_information(&smbios) {
+            log::error!("Failed to add BIOS information: {:?}", e);
+            return Err(e);
+        }
+
+        // Type 1: System information
+        if let Err(e) = Self::add_system_information(&smbios) {
+            log::error!("Failed to add system information: {:?}", e);
+            return Err(e);
+        }
+
+        // Type 2: Baseboard/motherboard information
+        if let Err(e) = Self::add_baseboard_information(&smbios) {
+            log::error!("Failed to add baseboard information: {:?}", e);
+            return Err(e);
+        }
+
+        // Type 3: System chassis/enclosure information
+        if let Err(e) = Self::add_system_enclosure(&smbios) {
+            log::error!("Failed to add system enclosure: {:?}", e);
+            return Err(e);
+        }
+
+        // Type 0x80: Platform-specific vendor record
+        if let Err(e) = Self::add_vendor_oem_record(&smbios) {
+            log::error!("Failed to add vendor OEM record: {:?}", e);
+            return Err(e);
+        }
+
+        // Type 127: End-of-table marker
+        if let Err(e) = Self::add_end_of_table(&smbios) {
+            log::error!("Failed to add end-of-table marker: {:?}", e);
+            return Err(e);
+        }
+
+        log::info!("Platform SMBIOS records created successfully");
+
+        // Publish the SMBIOS table to the UEFI Configuration Table
+        log::info!("Publishing SMBIOS table to Configuration Table...");
+        match smbios.publish_table(&boot_services) {
+            Ok((table_addr, entry_point_addr)) => {
+                log::info!("SMBIOS table published successfully");
+                log::info!("  Entry Point: 0x{:X}", entry_point_addr);
+                log::info!("  Table Data: 0x{:X}", table_addr);
+                log::info!("Use 'smbiosview' in UEFI Shell to view records");
+            }
+            Err(e) => {
+                log::error!("Failed to publish SMBIOS table: {:?}", e);
+                // Continue even if publication fails - not critical for boot
+            }
+        }
+
+        log::info!("SMBIOS platform component initialized successfully");
+        Ok(())
+    }
+
+    /// Add Type 0 (Platform Firmware/BIOS Information)
+    ///
+    /// Demonstrates adding firmware version, vendor, and ROM size information.
+    fn add_bios_information(smbios: &Service<Smbios>) -> Result<()> {
+        let bios_info = Type0PlatformFirmwareInformation {
+            header: SmbiosTableHeader::new(0, 0, SMBIOS_HANDLE_PI_RESERVED),
+            vendor: 1,
+            firmware_version: 2,
+            bios_starting_address_segment: 0xE800,
+            firmware_release_date: 3,
+            firmware_rom_size: 0xFF, // 16MB
+            characteristics: 0x08,   // PCI supported
+            characteristics_ext1: 0x03,
+            characteristics_ext2: 0x03,
+            system_bios_major_release: 1,
+            system_bios_minor_release: 0,
+            embedded_controller_major_release: 0xFF,
+            embedded_controller_minor_release: 0xFF,
+            extended_bios_rom_size: 0,
+            string_pool: vec![
+                String::from("Example Firmware Vendor"),
+                String::from("1.0.0"),
+                String::from("10/24/2025"),
+            ],
+        };
+
+        let handle = smbios.add_record(None, &bios_info).map_err(|e| {
+            log::error!("Failed to add BIOS info: {:?}", e);
+            patina::error::EfiError::DeviceError
+        })?;
+
+        log::info!("  Type 0 (BIOS Info) - Handle 0x{:04X}", handle);
+        Ok(())
+    }
+
+    /// Add Type 1 (System Information)
+    ///
+    /// Demonstrates adding system UUID, manufacturer, product name, and serial number.
+    fn add_system_information(smbios: &Service<Smbios>) -> Result<()> {
+        let system_info = Type1SystemInformation {
+            header: SmbiosTableHeader::new(1, 0, SMBIOS_HANDLE_PI_RESERVED),
+            manufacturer: 1,
+            product_name: 2,
+            version: 3,
+            serial_number: 4,
+            uuid: [0x12, 0x34, 0x56, 0x78, 0x9A, 0xBC, 0xDE, 0xF0, 0x12, 0x34, 0x56, 0x78, 0x9A, 0xBC, 0xDE, 0xF0],
+            wake_up_type: 0x06, // Power switch
+            sku_number: 5,
+            family: 6,
+            string_pool: vec![
+                String::from("Example Corporation"),
+                String::from("Example Platform"),
+                String::from("1.0"),
+                String::from("SYS-12345"),
+                String::from("SKU-EXAMPLE-001"),
+                String::from("Example Platform Family"),
+            ],
+        };
+
+        let handle = smbios.add_record(None, &system_info).map_err(|e| {
+            log::error!("Failed to add system info: {:?}", e);
+            patina::error::EfiError::DeviceError
+        })?;
+
+        log::info!("  Type 1 (System Info) - Handle 0x{:04X}", handle);
+        Ok(())
+    }
+
+    /// Add Type 2 (Baseboard/Motherboard Information)
+    ///
+    /// Demonstrates adding motherboard manufacturer, serial number, and asset tag.
+    fn add_baseboard_information(smbios: &Service<Smbios>) -> Result<()> {
+        let baseboard_info = Type2BaseboardInformation {
+            header: SmbiosTableHeader::new(2, 0, SMBIOS_HANDLE_PI_RESERVED),
+            manufacturer: 1,
+            product: 2,
+            version: 3,
+            serial_number: 4,
+            asset_tag: 5,
+            feature_flags: 0x01, // Board is a hosting board
+            location_in_chassis: 6,
+            chassis_handle: 0x0003,
+            board_type: 0x0A, // Motherboard
+            contained_object_handles: 0,
+            string_pool: vec![
+                String::from("Example Corporation"),
+                String::from("Example Baseboard"),
+                String::from("1.0"),
+                String::from("MB-67890"),
+                String::from("ASSET-MB-001"),
+                String::from("Main Board Slot"),
+            ],
+        };
+
+        let handle = smbios.add_record(None, &baseboard_info).map_err(|e| {
+            log::error!("Failed to add baseboard info: {:?}", e);
+            patina::error::EfiError::DeviceError
+        })?;
+
+        log::info!("  Type 2 (Baseboard Info) - Handle 0x{:04X}", handle);
+        Ok(())
+    }
+
+    /// Add Type 3 (System Enclosure/Chassis)
+    ///
+    /// Demonstrates adding chassis type, manufacturer, and serial information.
+    fn add_system_enclosure(smbios: &Service<Smbios>) -> Result<()> {
+        let enclosure_info = Type3SystemEnclosure {
+            header: SmbiosTableHeader::new(3, 0, SMBIOS_HANDLE_PI_RESERVED),
+            manufacturer: 1,
+            enclosure_type: 0x03, // Desktop
+            version: 2,
+            serial_number: 3,
+            asset_tag_number: 4,
+            bootup_state: 0x03,
+            power_supply_state: 0x03,
+            thermal_state: 0x03,
+            security_status: 0x02,
+            oem_defined: 0x00000000,
+            height: 0x00,
+            number_of_power_cords: 0x01,
+            contained_element_count: 0x00,
+            contained_element_record_length: 0x00,
+            string_pool: vec![
+                String::from("Example Corporation"),
+                String::from("Example Chassis v1.0"),
+                String::from("CHASSIS-99999"),
+                String::from("ASSET-CHASSIS-001"),
+            ],
+        };
+
+        let handle = smbios.add_record(None, &enclosure_info).map_err(|e| {
+            log::error!("Failed to add system enclosure: {:?}", e);
+            patina::error::EfiError::DeviceError
+        })?;
+
+        log::info!("  Type 3 (System Enclosure) - Handle 0x{:04X}", handle);
+        Ok(())
+    }
+
+    /// Add custom vendor-specific OEM record (Type 0x80)
+    ///
+    /// Demonstrates how platforms can add custom vendor-specific SMBIOS records.
+    /// Types 0x80-0xFF are reserved for OEM/vendor use.
+    fn add_vendor_oem_record(smbios: &Service<Smbios>) -> Result<()> {
+        let vendor_record = VendorOemRecord {
+            header: SmbiosTableHeader::new(VendorOemRecord::RECORD_TYPE, 0, SMBIOS_HANDLE_PI_RESERVED),
+            oem_field: 0xDEADBEEF, // Example vendor-specific data
+            string_pool: vec![String::from("Example Vendor"), String::from("Platform Data v1.0")],
+        };
+
+        let handle = smbios.add_record(None, &vendor_record).map_err(|e| {
+            log::error!("Failed to add vendor OEM record: {:?}", e);
+            patina::error::EfiError::DeviceError
+        })?;
+
+        log::info!("  Type 0x80 (Vendor OEM) - Handle 0x{:04X}", handle);
+        Ok(())
+    }
+
+    /// Add Type 127 (End of Table) marker
+    ///
+    /// Required marker to indicate the end of the SMBIOS table.
+    fn add_end_of_table(smbios: &Service<Smbios>) -> Result<()> {
+        let end_of_table = Type127EndOfTable {
+            header: SmbiosTableHeader::new(127, 0, SMBIOS_HANDLE_PI_RESERVED),
+            string_pool: vec![],
+        };
+
+        let handle = smbios.add_record(None, &end_of_table).map_err(|e| {
+            log::error!("Failed to add end-of-table: {:?}", e);
+            patina::error::EfiError::DeviceError
+        })?;
+
+        log::info!("  Type 127 (End of Table) - Handle 0x{:04X}", handle);
+        Ok(())
+    }
+}
--- a/components/patina_smbios/Cargo.toml
+++ b/components/patina_smbios/Cargo.toml
@@ -1,0 +1,19 @@
+[package]
+name = "patina_smbios"
+version.workspace = true
+license.workspace = true
+edition.workspace = true
+rust-version.workspace = true
+repository.workspace = true
+description = "System Management BIOS (SMBIOS) support for UEFI components."
+
+[dependencies]
+log = { workspace = true }
+patina = { workspace = true }
+r-efi = { workspace = true }
+spin = { workspace = true }
+zerocopy = { workspace = true }
+zerocopy-derive = { workspace = true }
+
+[features]
+enable_patina_tests = []
--- a/components/patina_smbios/src/component.rs
+++ b/components/patina_smbios/src/component.rs
@@ -1,0 +1,384 @@
+//! SMBIOS Service Implementation
+//!
+//! Defines the SMBIOS provider for use as a service
+//!
+//! ## License
+//!
+//! Copyright (c) Microsoft Corporation.
+//!
+//! SPDX-License-Identifier: Apache-2.0
+//!
+
+extern crate alloc;
+use crate::config::SmbiosConfiguration;
+use crate::error::SmbiosError;
+use crate::manager::{SmbiosManager, install_smbios_protocol};
+use crate::service::SmbiosHandle;
+use alloc::boxed::Box;
+use patina::{
+    boot_services::{StandardBootServices, tpl::Tpl},
+    component::{
+        IntoComponent,
+        params::{Commands, Config},
+        service::IntoService,
+    },
+    error::Result,
+    tpl_mutex::TplMutex,
+};
+use r_efi::efi::Handle;
+
+/// SMBIOS Service Trait (Stable Surface)
+///
+/// This trait defines the stable service interface for SMBIOS record operations.
+/// It provides byte-level record management (CRUD operations).
+///
+/// For type-safe record operations using generics, use the inherent
+/// `add_record<T>()` method on the concrete `Smbios` type.
+///
+/// For table-level operations (version, publishing), use the inherent methods
+/// on the `Smbios` type: `version()` and `publish_table()`.
+pub trait SmbiosRecords {
+    /// Add a pre-serialized SMBIOS record from raw bytes
+    ///
+    /// # Parameters
+    /// - `producer`: Optional producer handle for tracking record ownership
+    /// - `record`: Byte slice containing complete SMBIOS record (header + data + strings)
+    ///
+    /// # Returns
+    /// - `Ok(SmbiosHandle)`: Handle to the newly added record
+    /// - `Err(SmbiosError)`: If validation fails or table is full
+    fn add_from_bytes(
+        &self,
+        producer: Option<Handle>,
+        record: &[u8],
+    ) -> core::result::Result<SmbiosHandle, SmbiosError>;
+
+    /// Update a string in an existing SMBIOS record
+    ///
+    /// # Parameters
+    /// - `handle`: Handle of the record to update
+    /// - `string_number`: 1-based string index (0 means no string)
+    /// - `new_value`: New string value
+    ///
+    /// # Returns
+    /// - `Ok(())`: String updated successfully
+    /// - `Err(SmbiosError)`: If handle invalid, string_number out of range, or string too long
+    fn update_string(
+        &self,
+        handle: SmbiosHandle,
+        string_number: usize,
+        new_value: &str,
+    ) -> core::result::Result<(), SmbiosError>;
+
+    /// Remove an SMBIOS record by handle
+    ///
+    /// # Parameters
+    /// - `handle`: Handle of the record to remove
+    ///
+    /// # Returns
+    /// - `Ok(())`: Record removed successfully
+    /// - `Err(SmbiosError)`: If handle not found
+    fn remove(&self, handle: SmbiosHandle) -> core::result::Result<(), SmbiosError>;
+}
+
+/// Initializes and exposes SMBIOS provider service.
+///
+/// This component provides the `Service<Smbios>` which includes:
+/// - Type-safe record operations: `add_record<T>()`
+/// - Byte-level CRUD operations: `add_from_bytes()`, `update_string()`, `remove()`
+/// - Table management: `version()`, `publish_table()`
+///
+/// The provider creates an SMBIOS manager instance protected by a TplMutex.
+/// A global reference is maintained for C/EDKII protocol compatibility.
+#[derive(IntoComponent)]
+pub struct SmbiosProvider;
+
+impl Default for SmbiosProvider {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl SmbiosProvider {
+    /// Create a new SMBIOS provider
+    pub fn new() -> Self {
+        Self {}
+    }
+
+    /// Initialize the SMBIOS provider and register it as a service
+    #[coverage(off)] // Component integration - tested via integration tests
+    fn entry_point(
+        self,
+        config: Config<SmbiosConfiguration>,
+        mut commands: Commands,
+        boot_services: StandardBootServices,
+    ) -> Result<()> {
+        let cfg = (*config).clone();
+
+        // Create the manager with configured version
+        let manager = SmbiosManager::new(cfg.major_version, cfg.minor_version).map_err(|e| {
+            log::error!("Failed to create SMBIOS manager: {:?}", e);
+            patina::error::EfiError::Unsupported
+        })?;
+
+        // Convert boot_services to 'static by leaking it
+        // SAFETY: This is safe because boot_services remains valid for the entire lifetime
+        // of the system. We leak it intentionally to get a 'static reference.
+        let boot_services_static: &'static StandardBootServices = Box::leak(Box::new(boot_services));
+
+        // Wrap the manager in TplMutex for TPL-aware reentrancy protection at TPL_NOTIFY level
+        let manager_mutex = Box::leak(Box::new(TplMutex::new(boot_services_static, Tpl::NOTIFY, manager)));
+
+        // Install the C/EDKII protocol - this installs a global reference for C compatibility
+        if let Err(e) = install_smbios_protocol(manager_mutex, boot_services_static) {
+            log::error!("Failed to install SMBIOS protocol: {:?}", e);
+        }
+
+        // Register unified SMBIOS service with all operations
+        commands.add_service(Smbios { manager: manager_mutex });
+
+        Ok(())
+    }
+}
+
+/// SMBIOS service providing all SMBIOS operations
+///
+/// This service provides a complete interface for SMBIOS table management:
+/// - **Type-safe operations**: `add_record<T>()` for structured record types
+/// - **Byte-level operations**: `add_from_bytes()`, `update_string()`, `remove()`
+/// - **Table management**: `version()`, `publish_table()`
+///
+/// All operations are protected by a TplMutex at TPL_NOTIFY level for thread safety.
+///
+/// # Example
+///
+/// ```ignore
+/// fn entry_point(
+///     smbios: Service<Smbios>,
+///     boot_services: StandardBootServices,
+/// ) -> Result<()> {
+///     // Add structured records
+///     smbios.add_record(None, &bios_info)?;
+///     
+///     // Publish to configuration table
+///     smbios.publish_table(&boot_services)?;
+///     Ok(())
+/// }
+/// ```
+#[derive(IntoService, Clone, Copy)]
+#[service(Smbios)]
+pub struct Smbios {
+    manager: &'static TplMutex<'static, SmbiosManager, StandardBootServices>,
+}
+
+impl Smbios {
+    /// Gets the SMBIOS version information.
+    ///
+    /// # Returns
+    ///
+    /// A tuple of (major_version, minor_version).
+    pub fn version(&self) -> (u8, u8) {
+        let manager = self.manager.lock();
+        manager.version()
+    }
+
+    /// Publishes the SMBIOS table to the UEFI Configuration Table
+    ///
+    /// This should be called after all records have been added and the table
+    /// is ready to be consumed by the operating system or other firmware components.
+    ///
+    /// # Arguments
+    ///
+    /// * `boot_services` - Reference to UEFI Boot Services for installing
+    ///   the configuration table
+    ///
+    /// # Returns
+    ///
+    /// Returns a tuple of (table_address, entry_point_address) on success:
+    /// - `table_address`: Physical address of the SMBIOS table data
+    /// - `entry_point_address`: Physical address of the SMBIOS 3.x entry point structure
+    ///
+    /// # Errors
+    ///
+    /// Returns `SmbiosError` if:
+    /// - No records have been added
+    /// - Memory allocation fails
+    /// - Configuration table installation fails
+    pub fn publish_table(
+        &self,
+        boot_services: &StandardBootServices,
+    ) -> core::result::Result<(r_efi::efi::PhysicalAddress, r_efi::efi::PhysicalAddress), SmbiosError> {
+        let manager = self.manager.lock();
+        manager.publish_table(boot_services)
+    }
+
+    /// Adds an SMBIOS record to the SMBIOS table from a complete byte representation.
+    ///
+    /// **This is the recommended method for adding SMBIOS records.** It provides memory safety
+    /// and specification compliance by taking the complete record data as a validated byte slice,
+    /// avoiding unsafe pointer arithmetic and potential security vulnerabilities.
+    ///
+    /// # Arguments
+    ///
+    /// * `producer_handle` - Optional handle of the producer creating this record
+    /// * `record_data` - Complete SMBIOS record as a byte slice, including:
+    ///   - Header (4 bytes: type, length, handle)
+    ///   - Structured data (length - 4 bytes)
+    ///   - String pool (null-terminated strings ending with double null)
+    ///
+    /// # Returns
+    ///
+    /// Returns the assigned SMBIOS handle for the newly added record.
+    pub fn add_from_bytes(
+        &self,
+        producer_handle: Option<r_efi::efi::Handle>,
+        record_data: &[u8],
+    ) -> core::result::Result<SmbiosHandle, SmbiosError> {
+        let manager = self.manager.lock(); // TPL raised to NOTIFY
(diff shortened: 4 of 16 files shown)


makubacki:
Before this is opened for review, please consolidate the changes into logically meaningful commits for the main branch.

kat-perez:
> Before this is opened for review, please consolidate the changes into logically meaningful commits for the main branch.

consolidated everything into one commit. please let me know if we need further splits but I believe this is the most logically meaningful commit since it's the minimum required to get SMBIOS tables working in production

components/patina_smbios/examples/vendor_oem.rs
```diff
@@ -0,0 +1,202 @@
...
+        header: SmbiosTableHeader::new(VendorOemRecord::RECORD_TYPE, 0, SMBIOS_HANDLE_PI_RESERVED),
+        oem_field: 0xDEADBEEF,
+        string_pool: vec![String::from("Vendor Extra")],
+    };
+
+    // Serialize and add using the safe add_from_bytes method
+    let bytes = rec.to_bytes();
+
+    // Bring trait into scope so `add_from_bytes` and `get_next` methods are available on the manager
+    use patina_smbios::manager::SmbiosRecords;
```
joschock: nit: stylistically, `use` usually goes at the top rather than in the body of the code.

components/patina_smbios/examples/vendor_oem.rs
```diff
@@ -0,0 +1,202 @@
...
+        ],
+    };
+
+    let enclosure_bytes = enclosure_rec.to_bytes();
+    let _enclosure_handle = manager.add_from_bytes(None, &enclosure_bytes).expect("enclosure add_from_bytes failed");
+
+    // Verify all five records were added
+    let mut search = SMBIOS_HANDLE_PI_RESERVED;
+    let (found, _) = manager.get_next(&mut search, Some(VendorOemRecord::RECORD_TYPE)).expect("get_next failed");
+    assert_eq!(found.record_type, VendorOemRecord::RECORD_TYPE);
```
joschock: maybe add an assert showing that the oem field has deadbeef in it.

components/patina_smbios/src/component.rs
```diff
@@ -0,0 +1,164 @@
...
+
+        // Register the service so other components can consume it
+        commands.add_service(self);
+
+        Ok(())
+    }
+}
+
+// Delegate the SmbiosRecords trait implementation to the global manager
+// All methods acquire TplMutex.lock() which automatically raises TPL to CALLBACK
```
joschock: The edk2 SMBIOS version uses TPL_NOTIFY. if you use TPL_CALLBACK, it means events running at TPL_NOTIFY could be re-entrant to these functions. If that happened, you would get a `panic` when they tried to acquire the TplMutex if it was already locked by code running at TPL_CALLBACK or lower.

If the intent is that these routines must only be invoked at TPL_CALLBACK or lower, then probably need to add that as a requirement on the caller in the docs. I don't know of any actual examples of someone installing SMBIOS records at TPL_NOTIFY, but it would not surprise me to find them, so it might be easiest to simply raise the TPL mutex to NOTIFY. This would make it safe to call from pretty much anywhere besides stuff running at TPL_HIGH_LEVEL, and anything running at TPL_HIGH_LEVEL should not be installing SMBIOS tables.

components/patina_smbios/src/manager.rs
```diff
@@ -0,0 +1,2398 @@
...
+        let table_pages = uefi_size_to_pages!(total_table_size);
+        let table_address = boot_services
+            .allocate_pages(
+                AllocType::AnyPage,
+                MemoryType::ACPI_RECLAIM_MEMORY, // SMBIOS tables go in ACPI Reclaim memory
+                table_pages,
+            )
+            .map_err(|_| SmbiosError::OutOfResources)?;
+
+        // Step 3: Copy all records to the table
```
joschock: nit: safety comment.

components/patina_smbios/src/manager.rs
```diff
@@ -0,0 +1,2398 @@
...
+
+        // Step 5: Allocate memory for entry point structure
+        let ep_pages = 1; // Entry point fits in one page
+        let ep_address = boot_services
+            .allocate_pages(AllocType::AnyPage, MemoryType::ACPI_RECLAIM_MEMORY, ep_pages)
+            .map_err(|_| SmbiosError::OutOfResources)?;
+
+        // Step 6: Copy entry point to allocated memory
+        let ep_bytes = entry_point.as_bytes();
+        let ep_slice = unsafe {
```
joschock: nit: safety comment

components/patina_smbios/src/manager.rs
```diff
@@ -0,0 +1,2398 @@
...
+            core::slice::from_raw_parts_mut(ep_address as *mut u8, core::mem::size_of::<Smbios30EntryPoint>())
+        };
+        ep_slice.copy_from_slice(ep_bytes);
+
+        // Step 7: Install in UEFI Configuration Table
+        unsafe {
```
joschock: nit: safety comment

components/patina_smbios/src/manager.rs
```diff
@@ -0,0 +1,2398 @@
...
+            Ref::<&[u8], SmbiosTableHeader>::from_prefix(record_data).map_err(|_| SmbiosError::InvalidParameter)?;
+        let header: &SmbiosTableHeader = &header_ref;
+
+        // Step 3: Validate header->length is <= (record_data.length - 2) for string pool
+        // The string pool needs at least 2 bytes for the double-null terminator
+        if (header.length as usize + 2) > record_data.len() {
+            return Err(SmbiosError::BufferTooSmall);
+        }
+
+        // Step 4: Validate and count strings in a single efficient pass
```
joschock: nit: comment seems to be describing step 5

components/patina_smbios/src/manager.rs
```diff
@@ -0,0 +1,2398 @@
...
+    unsafe fn get(&self) -> efi::Handle {
+        unsafe { *self.handle.get() }
+    }
+}
+
+static SMBIOS_PROTOCOL_HANDLE: GlobalProtocolHandle = GlobalProtocolHandle::new();
+
+/// Wrapper for static SMBIOS header buffer that implements Sync
+///
+/// SAFETY: This is safe because UEFI DXE runs in a single-threaded environment,
```
joschock: Technically, DXE is signle-threaded with interrupts. so you still have to guard against an interrupt coming in during a critical section. That's what the TplMutex locks are for...

joschock (approved):

components/patina_smbios/Cargo.toml
makubacki, line 1: @kat-perez, I'd like to look at the PR in more detail. Since there is quite a bit to go through, please give reviewers at least a couple of days to have an opportunity to provide feedback.

Javagedes:
I will review this on Monday. Please do not merge until I've had time to review. Thank you:)

components/patina_smbios/examples/vendor_oem.rs
```diff
@@ -0,0 +1,202 @@
...
+
+    // Example 1: Vendor OEM Record
+    let rec = VendorOemRecord {
+        header: SmbiosTableHeader::new(VendorOemRecord::RECORD_TYPE, 0, SMBIOS_HANDLE_PI_RESERVED),
+        oem_field: 0xDEADBEEF,
+        string_pool: vec![String::from("Vendor Extra")],
+    };
+
+    // Serialize and add using the safe add_from_bytes method
+    let bytes = rec.to_bytes();
```
os-d: In one of the PRs into the feature branch, I had raised this point as well, and apologies, I forgot the resolution, but for Rust based SMBIOS addition, it feels like there could be a better, safer way to pass the table from to the SMBIOS component that serializing to bytes and calling add_from_bytes. I think that works well for the C protocol, but for Rust, I wonder if doing something like the ACPI component (which takes a generic table) would fit well here, they are very similar components. It may be simpler for SMBIOS, because IIUC, all SMBIOS tables are primarily a string pool, which the SMBIOS component is going to copy. But also, you have the field layout thing which could describe other fields as needed. Curious @Javagedes and @berlin-with0ut-return 's opinions here, since they worked on the ACPI component's interface.

On the flip side, Joey and Sherry, I'm curious if this FieldLayout struct (I need to look further at the details) would help solve any of the issues in the ACPI component where the incoming tables' fields aren't known (not really sure if it is relevant, but maybe could be an interesting paradigm).

Particularly for the well known SMBIOS tables, it seems like it would be ideal to not pass them as bytes.

components/patina_smbios/examples/vendor_oem.rs
```diff
@@ -0,0 +1,202 @@
...
+    let _enclosure_handle = manager.add_from_bytes(None, &enclosure_bytes).expect("enclosure add_from_bytes failed");
+
+    // Verify all five records were added
+    let mut search = SMBIOS_HANDLE_PI_RESERVED;
+    let (found, _) = manager.get_next(&mut search, Some(VendorOemRecord::RECORD_TYPE)).expect("get_next failed");
+    assert_eq!(found.record_type, VendorOemRecord::RECORD_TYPE);
+    info!("Added vendor record handle: {}", search);
+
+    search = SMBIOS_HANDLE_PI_RESERVED;
+    let (found_bios, _) = manager.get_next(&mut search, Some(0)).expect("get_next failed for bios");
```
os-d: Similarly, for Rust interactions, I think having an iterator be returned (I'll have to review the component again, I thought there was an iterator?) instead of calling the C style get_next seems preferable.

components/patina_smbios/src/manager.rs
```diff
@@ -0,0 +1,2398 @@
...
+    fn publish_table(
+        &self,
+        boot_services: &patina::boot_services::StandardBootServices,
+    ) -> Result<(r_efi::efi::PhysicalAddress, r_efi::efi::PhysicalAddress), SmbiosError>;
+}
+
+/// Extension trait for ergonomic SMBIOS record management
+///
+/// Provides convenient methods for adding SMBIOS records from structured types.
+pub trait SmbiosService {
```
Javagedes: We do our best to not use the word `Service` when defining the trait interface for a service. This is because in the component interface, it ultimetly looks like this:

```rust
fn my_component(&self, smbios: Service<dyn SmbiosService>) -> Result<()> {
    ...
}
```

components/patina_smbios/examples/vendor_oem.rs
```diff
@@ -0,0 +1,202 @@
...
+
+    fn string_pool(&self) -> &[String] {
+        &self.string_pool
+    }
+    fn string_pool_mut(&mut self) -> &mut Vec<String> {
+        &mut self.string_pool
+    }
+}
+
+fn main() {
```
Javagedes: Can you add an additional example in patina_samples that shows a component consuming and using the actual service through the component interface?

components/patina_smbios/src/manager.rs
```diff
@@ -0,0 +1,2398 @@
...
+    fn update_string(&self, smbios_handle: SmbiosHandle, string_number: usize, string: &str)
+    -> Result<(), SmbiosError>;
+
+    /// Removes an SMBIOS record from the SMBIOS table.
+    fn remove(&self, smbios_handle: SmbiosHandle) -> Result<(), SmbiosError>;
+
+    /// Discovers SMBIOS records, optionally filtered by type.
+    fn get_next(
+        &self,
+        smbios_handle: &mut SmbiosHandle,
```
Javagedes: This `IN OUT` variable seems very dangerous, and a C concept that has not really be adoped here. This seems error prone to double usage of the handle, where the handle has changed from one search to another. Even in your example, you are needing to manually reset it back to the default, which may not be obvious to the user. Additionally, this `get_next` iteration appears to be leaving the `SmbiosRecords` in a non-default state after every usage, which can also be error prone depending on usage.

I feel like a better interface would be something along the lines of

``` rust
// Iterate through all records with the SMBIOS handle
fn into_iter(&mut self, smbios_handle: SmbiosHandle) -> SmbiosRecordsIter<'a> { todo!() }
```

components/patina_smbios/src/manager.rs
```diff
@@ -0,0 +1,2398 @@
...
+impl SmbiosService for patina::component::service::Service<dyn SmbiosRecords<'static>> {
+    fn add_record<T>(
+        &self,
+        producer_handle: Option<r_efi::efi::Handle>,
+        record: &T,
+    ) -> Result<SmbiosHandle, SmbiosError>
+    where
+        T: crate::smbios_record::SmbiosRecordStructure + crate::smbios_record::SmbiosFieldLayout,
+    {
+        // Dereference Service twice to get &'static (dyn SmbiosRecords)
```
Javagedes: This dereferencing is crazy to me - I did not even know you could do something like this. But why do you need to do this? Just have two separate services, both provided by the same underlying struct? You would be able to access any of the functions between eachother in your manual implementation and the component writer can request whatever they actually need 

```
pub trait SmBiosService {}
pub trait SmBiosRecords {}

#[derive(IntoService)]
#[service(dyn SmBiosService, dyn SmBiosRecords)]
pub struct MyImplementation { }

fn my_entry_point(service: Service<dyn SmBiosService>) -> Result<()> { todo!() }
fn my_entry_point2(service: Service<dyn SmBiosRecords>) -> Result<()> { todo!() }
```

components/patina_smbios/src/smbios_record.rs
```diff
@@ -0,0 +1,774 @@
...
+pub struct FieldInfo {
+    /// Field name
+    pub name: &'static str,
+    /// Field type and offset
+    pub field_type: FieldType,
+}
+
+/// Field type enumeration for SMBIOS records
+#[derive(Debug, Clone)]
+pub enum FieldType {
```
Javagedes: Is this internal usage only, or will people also be using this? It really seems overly complicated and uses additional structs and struct instantiation then necessary. It seems like you can directly implement a serialize trait with a macro. The macro would give you the string name, and the ability to serialize. This would be faster then the looping that you do on the generic FieldInfo / FieldType data you use. 

```rust
trait SmBiosSerialize {
    pub serialize(self, &mut [u8]) -> Result;
}
```

```rust
impl_serialize!(pub struct Type0PlatformFirmwareInformation {
    /// SMBIOS table header
    pub header: SmbiosTableHeader,
    /// Vendor string index
    pub vendor: u8,
    /// Firmware version string index
    pub firmware_version: u8,
    /// BIOS starting address segment
    pub bios_starting_address_segment: u16,
    /// Firmware release date string index
    pub firmware_release_date: u8,
    /// Firmware ROM size
    pub firmware_rom_size: u8,
    /// BIOS characteristics
    pub characteristics: u64,
    /// BIOS characteristics extension byte 1
    pub characteristics_ext1: u8,
    /// BIOS characteristics extension byte 2
    pub characteristics_ext2: u8,
    /// System BIOS major release
    pub system_bios_major_release: u8,
    /// System BIOS minor release
    pub system_bios_minor_release: u8,
    /// Embedded controller firmware major release
    pub embedded_controller_major_release: u8,
    /// Embedded controller firmware minor release
    pub embedded_controller_minor_release: u8,
    /// Extended BIOS ROM size
    pub extended_bios_rom_size: u16,

    /// String pool containing the actual string content.
    ///
    /// **IMPORTANT**: This field is NOT part of the SMBIOS table binary layout.
    /// It is Rust metadata that gets converted to null-terminated bytes during serialization.
    /// Never attempt to directly cast this struct to bytes or use it in FFI - always use
    /// `SmbiosSerializer::serialize()`.
    pub string_pool: Vec<String>,
})

// Which makes something like

impl SmBiosSerialize for Type0PlatformFirmwareInformation {
    fn serialize(self, bytes: &mut [u8]) -> Result<()> {
        let mut idx = 0;
        bytes[idx..idx + 1].copy_from_slice([self.vendor]);
        idx += 1;
        bytes[idx..idx + 1].copy_from_slice([self.firmware_version]);
        idx += 1;
        bytes[idx..idx + 2].copy_from_slice(self.bios_starting_address_segment.to_le_bytes());
        // Etc. Etc.
    }
}
```

If vendors are expected to use this, then I suggest making it a macro instead:

```rust
#[derive(SmBiosSerialize)]
struct Type0PlatformFirmwareInformation { ... }
```

Javagedes: You can also do things when you see a `String` type. If you use the derive macro then you could even do attribute tags on it. Again this fancy stuff is only needed if the public is expected to use it. If this is only for internal types then I think just some improvement to make the serializer implement a serialize function directly rather than this unecessary layer of creating this field layout information, just to serialize.

components/patina_smbios/src/lib.rs
```diff
@@ -0,0 +1,52 @@
...
+//! Copyright (c) Microsoft Corporation.
+//!
+//! SPDX-License-Identifier: Apache-2.0
+//!
+
+#![no_std]
+
+pub mod component;
+/// SMBIOS manager and protocol implementation
+pub mod manager;
```
Javagedes: nit: This should not be public. We do not want someone to directly use the hardcoded manager. They need to access it via the component interface. Traits that act as the interface need to be public, then consumers will use it in their component itnerface as `service: Service<dyn MyInterface>`

components/patina_smbios/src/lib.rs
```diff
@@ -0,0 +1,52 @@
+//! SMBIOS (System Management BIOS) component for Patina
```
Javagedes, line 1: Need way more docs here, particularly examples and usage, etc. This is the "entry point" into the crate documention on docs.rs

components/patina_smbios/src/manager.rs
```diff
@@ -0,0 +1,2398 @@
...
+    ///
+    /// # Safety
+    ///
+    /// Caller must ensure this is only called during error cleanup
+    unsafe fn clear(&self) {
+        unsafe { *self.manager.get() = None };
+    }
+}
+
+static SMBIOS_MANAGER: GlobalSmbiosManager = GlobalSmbiosManager::new();
```
Javagedes: Why does this need to be global? This should be the private state of the service being produced. If multiple services are accessing this global state, then make all services be implemented off of the same underlying underlying struct. Regardless, one of the intents of services is their is no need for global state.

The only time we have found that global state is necessary is when we are producing a EFI compatible UEFI protocol that does not consume a pointer to self in it's functions. 

If it is necessary, then I just need to know why, and maybe we can find a solution to not have global static state.

components/patina_smbios/src/manager.rs
```diff
@@ -0,0 +1,2398 @@
...
+    /// # Safety
+    ///
+    /// Returns None if not initialized
+    #[allow(dead_code)] // Reserved for future diagnostic access to raw boot services
+    unsafe fn get(&self) -> Option<&'static StandardBootServices> {
+        unsafe { *self.boot_services.get() }
+    }
+}
+
+static BOOT_SERVICES: GlobalBootServices = GlobalBootServices::new();
```
Javagedes: This should also be unecessary. You should be able to stash the BootServices you get from your component interface inside your service implementation.

components/patina_smbios/src/manager.rs
```diff
@@ -0,0 +1,2398 @@
+//! SMBIOS Core Implementation
```
Javagedes: 1300 lines for a single file is getting pretty big (Though it is done elsewhere). Is there any logical places to break this up into multiple modules underneath the manager.rs module? It's OK if not, but it's always a good practice to keep modules smaller if possible.

components/patina_smbios/src/manager.rs
```diff
@@ -0,0 +1,2398 @@
...
+    /// The handle value is invalid
+    InvalidHandle,
+    /// String exceeds maximum allowed length
+    StringTooLong,
+    /// Buffer is too small for operation
+    BufferTooSmall,
+}
+
+/// Core SMBIOS record management operations
+pub trait SmbiosRecords<'a> {
```
Javagedes: The public trait interfaces for services should go in a public service module as specified here:

https://opendevicepartnership.github.io/patina/component/requirements.html

Any public types used in the interface of the service module should go there too, or at least re-exported there for easy consumption for people using the service.

components/patina_smbios/src/manager.rs
Javagedes: Oh I think I see what you are doing. This is enabling generic functions on services. This is an interesting way to get around not being able to have generics on dyn traits. I actually think I like this, but why are you doing this casting? Services have an auto-deref implemented on them, so the service functions should be available automatically.

e.g. Just from looking at the code, the deref of self should theoretically be unecessary

``` rust
fn fn add_record<T>(
        &self,
        producer_handle: Option<r_efi::efi::Handle>,
        record: &T,
    ) -> Result<SmbiosHandle, SmbiosError>
{
    let bytes = record.to_bytes();
    service.add_from_bytes(producer_handle, &bytes)
}
```

Javagedes: Also, are you able to not use a trait implementation, and directly implement on `Service<dyn SmBiosRecords>`?. The only reason I ask is because the caller must make sure `SmbiosService` is in scope for this, which has confused many rust developers as to why certain functions are not available even though they know they exist.

components/patina_smbios/src/smbios_record.rs
Javagedes: You could also completely get rid of all of this manual serialization if you do something similar to the ACPI table where we have a pointer to some random allocated memory, and can cast to different types as needed. The main thing is you would have to fix the Strings to be a char slice or something, rather than a Vec<String> which is a pointer to a vector of pointers to a vector of `char`s.

components/patina_smbios/examples/vendor_oem.rs
Javagedes: Doing it the ACPI way would greatly simplify the serialization process as there would not be any parsing necessary. She could use `zerocopy` on the types if she wanted the bytes directly. However she would be able to store all entry's as type erased structures for management purposes, and still have an interface to give users the original type back. It would look something like this (though I just made this up so it probably won't compile).

```rust

union Entry<T = SmBiosTableHeader> {
    record_type: SmBiosType,
    header: SmBiosTableHeader,
    inner: ManuallyDrop<T>
}

impl<T> Entry<T> {
    pub unsafe fn new(entry: T) -> Self {
        Entry { inner: ManuallyDrop::new(entry) }
    }
}

struct SmBiosEntry {
    entry: NonNull<SmBiosEntry>,
    type_id: TypeId,
}

impl SmBiosEntry {
    fn new<T>(entry: T) -> Self {
      let entry = Box::leak(Box::new(Entry::new(entry))); // Probably do something better here
      SmBiosEntry { NonNull::from_ref(entry), type_id: TypeId::of::<T>() };
    }

    fn record_type(&self) -> SmBiosType {
        unsafe { self.entry.as_ref().record_type }
    }
    
    fn header(&self) -> &SmBiosTableHeader {
        unsafe { self.entry.as_ref().header }
    }
    
    fn as_ref<T>(&self) -> Option<&T> {
        if TypeId::of::<T>() != self.type_id {
            return None
        }
        
        self.entry.cast::<Entry<T>>().as_ref().as_ref()
    }
    
    fn as_mut<T>(&mut self) -> Option<&mut T> {
        if TypeId::of::<T>() != self.type_id {
            return None
        }
        
        self.entry.cast::<Entry<T>>().as_mut().as)mut()
    }
}

// Example usage
fn main() {
  struct MyCustomType<COUNT: usize> {
      pub header: SmbiosTableHeader,
      something: u8,
      another_thing: u16
      string_pool: [char; count]
  }
  
  let my_type: MyCustomType<6> = MyCustomType {
    ...
    string_pool: ['a', 'b', 'c', 'd', 'e', '\0']
  }
  
  // Now see here the type is erased
  let entry: SmBiosEntry = SmBiosEntry::new(my_type).cast::<SmBiosTableHeader>();
  
  assert!(entry.as_ref::<MyCustomType<6>>().is_some());
}

```

components/patina_smbios/src/component.rs
```diff
@@ -0,0 +1,164 @@
...
+        IntoComponent,
+        params::{Commands, Config},
+        service::IntoService,
+    },
+    error::Result,
+};
+
+/// Configuration for SMBIOS service
+#[derive(Debug, Clone)]
+pub struct SmbiosConfiguration {
```
makubacki: This should be moved to a `config` module to follow [0009-standardize-component-crate](https://github.com/OpenDevicePartnership/patina/blob/main/docs/src/rfc/text/0009-standardize-component-crate.md).

components/patina_smbios/src/component.rs
```diff
@@ -0,0 +1,164 @@
...
+/// the boot process. This ensures a single source of truth for SMBIOS data and allows
+/// both Rust services and C/EDKII drivers to access the same SMBIOS tables.
+///
+/// The global instance is thread-safe via an internal Mutex and has 'static lifetime.
+#[derive(IntoComponent, IntoService)]
+#[service(dyn SmbiosRecords<'static>)]
+pub struct SmbiosProviderManager {
+    // No internal state - uses global singleton
+}
+
```
makubacki: Can this just be a unit struct?


```suggestion
pub struct SmbiosProviderManager;

```

components/patina_smbios/src/component.rs
```diff
@@ -0,0 +1,164 @@
...
+pub struct SmbiosConfiguration {
+    /// SMBIOS major version (e.g., 3 for SMBIOS 3.x)
+    pub major_version: u8,
+    /// SMBIOS minor version (e.g., 0 for SMBIOS 3.0)
+    pub minor_version: u8,
+}
+
+impl Default for SmbiosConfiguration {
+    fn default() -> Self {
+        Self { major_version: 3, minor_version: 9 }
```
makubacki: Why is 3.9 chosen as the default?

components/patina_smbios/src/component.rs
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    /// Initialize the SMBIOS provider and register it as a service
+    fn entry_point(
+        self,
+        config: Option<Config<SmbiosConfiguration>>,
+        mut commands: Commands,
+        boot_services: StandardBootServices,
+    ) -> Result<()> {
+        let cfg = config.map(|c| (*c).clone()).unwrap_or_default();
```
makubacki: I think the configuration should be required instead of optional. Right now, this will fall back to the default which is currently going to result in a version of "3.9" and that's probably not going to always make sense for all platforms.

components/patina_smbios/src/smbios_record.rs
```diff
@@ -0,0 +1,774 @@
...
+        //
+        // IMPORTANT: The SMBIOS record structs (Type0, Type1, etc.) contain a `string_pool: Vec<String>`
+        // field which is Rust metadata (fat pointers) and NOT part of the SMBIOS binary format.
+        // This field layout mechanism only extracts the primitive fields (u8, u16, u32, u64, uuid)
+        // that form the structured portion of the SMBIOS table. The string pool is handled separately
+        // by serialize_string_pool().
+        for field_info in &layout.fields {
+            match field_info.field_type {
+                FieldType::U8(offset) => {
+                    let ptr = (record as *const T as *const u8).wrapping_add(offset);
```
makubacki: This `wrapping_add` seems a little dangerous. Is there a bounds check that can be validated against `offset`?

components/patina_smbios/src/component.rs
makubacki: Unless noted as being restricted somewhere, it is like a notification function will eventually be registered at `TPL_NOTIFY` and attempt to modify SMBIOS data.

components/patina_smbios/src/manager.rs
```diff
@@ -0,0 +1,2398 @@
...
+use zerocopy_derive::{FromBytes, Immutable, IntoBytes as DeriveIntoBytes, KnownLayout};
+
+/// SMBIOS record handle type (16-bit identifier)
+pub type SmbiosHandle = u16;
+
+/// Special handle value for automatic assignment
+pub const SMBIOS_HANDLE_PI_RESERVED: SmbiosHandle = 0xFFFE;
+
+/// SMBIOS Protocol GUID: 03583ff6-cb36-4940-947e-b9b39f4afaf7
+pub const SMBIOS_PROTOCOL_GUID: efi::Guid =
```
makubacki: I suggest moving content like this (protocol GUID) to another module (file).

components/patina_smbios/src/manager.rs
```diff
@@ -0,0 +1,2398 @@
...
+
+    /// Discovers SMBIOS records, optionally filtered by type.
+    fn get_next(
+        &self,
+        smbios_handle: &mut SmbiosHandle,
+        record_type: Option<SmbiosType>,
+    ) -> Result<(SmbiosTableHeader, Option<Handle>), SmbiosError>;
+
+    /// Gets the SMBIOS version information.
+    fn version(&self) -> (u8, u8); // (major, minor)
```
makubacki: Some of these APIs seem a little strange to associate directly with a "records" type.

- Make sense: `add`, `update`, `remove`, `get_next`
- Not very cohesive: SMBIOS version, a function to publish the SMBIOS table to the UEFI configuration table

Can those just be a part of `SmbiosManager` without being a part of this `SmbiosRecords` trait? Or, a different trait from `SmbiosRecords` (if that's needed)?

components/patina_smbios/src/manager.rs
```diff
@@ -0,0 +1,2398 @@
...
+        let bytes = record.to_bytes();
+        service.add_from_bytes(producer_handle, &bytes)
+    }
+}
+
+/// SMBIOS 3.0 entry point structure (64-bit)
+/// Per SMBIOS 3.0+ specification section 5.2.2
+#[repr(C, packed)]
+#[derive(Clone, Copy, DeriveIntoBytes, Immutable)]
+pub struct Smbios30EntryPoint {
```
makubacki: A lot of these struct definitions would be better organized in a different file from this `manager.rs` file.

components/patina_smbios/src/manager.rs
```diff
@@ -0,0 +1,2398 @@
...
+    table_64_address: RefCell<Option<PhysicalAddress>>,
+}
+
+impl SmbiosManager {
+    /// Creates a new SMBIOS manager with the specified version
+    ///
+    /// # Arguments
+    ///
+    /// * `major_version` - SMBIOS major version (e.g., 3 for SMBIOS 3.x)
+    /// * `minor_version` - SMBIOS minor version (e.g., 9 for SMBIOS 3.9)
```
makubacki:
```suggestion
    /// * `minor_version` - SMBIOS minor version (e.g., 9 for SMBIOS x.9)
```

components/patina_smbios/src/manager.rs
```diff
@@ -0,0 +1,2398 @@
...
+    ///
+    /// # Returns
+    ///
+    /// Returns `Ok(())` if valid, or an appropriate error if validation fails
+    fn validate_string(s: &str) -> Result<(), SmbiosError> {
+        if s.len() > SMBIOS_STRING_MAX_LENGTH {
+            return Err(SmbiosError::StringTooLong);
+        }
+        // Strings must NOT contain null terminators - they are added during serialization
+        if s.contains('\0') {
```
makubacki: I think this is more explicit about the byte operation being performed, but it's just a consideration, I think what is currently there is also fine.


```suggestion
        if s.bytes().any(|b| b == 0) {
```

components/patina_smbios/src/manager.rs
```diff
@@ -0,0 +1,2398 @@
...
+    ///
+    /// # Arguments
+    ///
+    /// * `header` - The SMBIOS table header and structured data
+    /// * `strings` - Array of string slices to include in the string pool
+    ///
+    /// # Returns
+    ///
+    /// Returns a complete SMBIOS record byte array ready to be added via `add_from_bytes`
+    #[allow(dead_code)]
```
makubacki: Is `#[allow(dead_code)]` needed?

components/patina_smbios/src/manager.rs
```diff
@@ -0,0 +1,2398 @@
...
+        boot_services: &patina::boot_services::StandardBootServices,
+    ) -> Result<(PhysicalAddress, PhysicalAddress), SmbiosError> {
+        let records = self.records.borrow();
+
+        // Step 1: Calculate total table size
+        let total_table_size: usize = records.iter().map(|r| r.data.len()).sum();
+
+        if total_table_size == 0 {
+            log::warn!("No SMBIOS records to install");
+            return Err(SmbiosError::InvalidParameter);
```
makubacki: `InvalidParameter` is usually taken to mean a parameter to the function was not valid. In this case, this is an error in the internally stored `records` data.

components/patina_smbios/src/manager.rs
```diff
@@ -0,0 +1,2398 @@
...
+        if let Some(handle) = self.freed_handles.borrow_mut().pop() {
+            return Ok(handle);
+        }
+
+        // No freed handles available, use next_handle
+        let candidate = *self.next_handle.borrow();
+
+        // Check if we've exhausted the handle space
+        // Valid handles are 1..=0xFEFF (0xFFFE and 0xFFFF are reserved)
+        if candidate >= 0xFFFE {
```
makubacki: A constant for `0xFFFE` would make the code more clear here.

components/patina_smbios/src/manager.rs
```diff
@@ -0,0 +1,2398 @@
...
+    ///
+    /// # Arguments
+    ///
+    /// * `entry_point` - Reference to the SMBIOS 3.0 Entry Point Structure
+    ///
+    /// # Returns
+    ///
+    /// The checksum byte value that makes the structure's byte sum equal to zero
+    ///
+    /// # Safety
```
makubacki: I don't think this safety comment is needed in the function doc. As a caller, I would need expect to see what I need to do to uphold safety expectations. This is simply describing an internal implementation detail.

components/patina_smbios/src/manager.rs
```diff
@@ -0,0 +1,2398 @@
...
+        boot_services: &patina::boot_services::StandardBootServices,
+    ) -> Result<(r_efi::efi::PhysicalAddress, r_efi::efi::PhysicalAddress), SmbiosError> {
+        self.install_configuration_table(boot_services)
+    }
+}
+
+/// SMBIOS table header structure
+#[repr(C, packed)]
+#[derive(Debug, Clone, PartialEq, FromBytes, DeriveIntoBytes, Immutable, KnownLayout)]
+pub struct SmbiosTableHeader {
```
makubacki: Some of these structures like `SmbiosTableHeader` and `SmbiosRecord` could also result in better organization if in a separate file.

components/patina_smbios/src/smbios_record.rs
```diff
@@ -0,0 +1,774 @@
...
+            FieldType::U8(_) => 1,
+            FieldType::U16(_) => 2,
+            FieldType::U32(_) => 4,
+            FieldType::U64(_) => 8,
+            FieldType::ByteArray { len, .. } => len,
+        }
+    }
+}
+
+/// Macro to automatically generate both field layout and SmbiosRecordStructure implementation
```
makubacki: Doc comment needs to be expanded with examples.

components/patina_smbios/Cargo.toml
makubacki, line 1: `manager.rs` has a lot going on. There are many opportunities to leverage the module system to split logical groupings of code in that file into more files to reduce overall file length, a more hierarchical namespace to reduce some redundancy in type names and make types natural to reference, and more doc comments that can be added to make the overall design and logical cohesion of the different content more clear.

On the last note, I think somewhere (perhaps `lib.rs`) needs to have a doc comment that describes that main types involved and their relationships to each other to help guide a reader on how to best connect the relationships between the main types involved and what roles they serve in the larger design.

components/patina_smbios/src/manager.rs
kat-perez: it wasn't explicitly necessary. I have refactored this to remove the global state to store the manager and boot services in the struct `SmbiosProviderManager`

components/patina_smbios/src/component.rs
kat-perez: it was just chosen since it's the latest spec

components/patina_smbios/src/component.rs
kat-perez: refactored this to use  TPL_NOTIFY instead of CALLBACK

components/patina_smbios/src/manager.rs
kat-perez: fixed

components/patina_smbios/src/manager.rs
kat-perez: added

components/patina_smbios/src/manager.rs
kat-perez: added

components/patina_smbios/src/manager.rs
kat-perez: added

components/patina_smbios/src/manager.rs
kat-perez: removing

kat-perez: removed

components/patina_smbios/src/manager.rs
kat-perez: updated to use  SMBIOS_HANDLE_PI_RESERVED

components/patina_smbios/src/manager.rs
kat-perez: adding a new error type `SmbiosError::NoRecordsAvailable` to indicate that no SMBIOS records have been added

kat-perez: i'm going to rework some of the error messages while were here since we decided on more granular error messages in the RFC

components/patina_smbios/src/component.rs
kat-perez: ok, i think that makes sense to make it required instead of optional

components/patina_smbios/examples/vendor_oem.rs
kat-perez: moved up

components/patina_smbios/examples/vendor_oem.rs
kat-perez: done

components/patina_smbios/src/manager.rs
kat-perez: removing this whole function since it's not needed

components/patina_smbios/src/component.rs
kat-perez: moved this into a config module

components/patina_smbios/src/lib.rs
kat-perez: making it private

components/patina_smbios/src/manager.rs
kat-perez: moved this module to a new folder called "manager" and then i separated out parts according to their functional purpose as much as possible

components/patina_smbios/src/lib.rs
kat-perez: adding more docs

components/patina_smbios/src/component.rs
kat-perez: agreed. updating

components/patina_smbios/src/manager.rs
kat-perez: improved the organization

components/patina_smbios/src/smbios_record.rs
kat-perez: it's not used internally, but i've reworked this so that we don't have FieldLayout/FieldInfo/FieldType runtime structures. we now have `to_bytes`

components/patina_smbios/src/manager.rs
kat-perez: also split out the organization for this one

components/patina_smbios/src/manager.rs
kat-perez: reworked in SmbiosManager

components/patina_smbios/Cargo.toml
kat-perez: did a large rework on the structure

components/patina_smbios/src/smbios_record.rs
kat-perez: updated

components/patina_smbios/src/manager.rs
kat-perez: moved

components/patina_smbios/src/smbios_record.rs
kat-perez: got rid of it and am using extend_from_slice

components/patina_smbios/examples/vendor_oem.rs
kat-perez: returned an iterator instead

components/patina_smbios/src/manager.rs
kat-perez: made sure trait interfaces don't have "Service" in the name

components/patina_samples/src/smbios_example.rs
```diff
@@ -0,0 +1,345 @@
...
+    ///
+    /// Demonstrates using the type-safe structured records service.
+    #[cfg(feature = "enable_smbios_example")]
+    fn add_bios_information(smbios: &Service<patina_smbios::SmbiosStructuredRecords>) -> Result<()> {
+        use patina_smbios::service::{SMBIOS_HANDLE_PI_RESERVED, SmbiosTableHeader};
+        use patina_smbios::smbios_record::Type0PlatformFirmwareInformation;
+
+        log::info!("Adding Type 0 (BIOS Information) record...");
+
+        let bios_info = Type0PlatformFirmwareInformation {
```
AnhDLuong: When comparing the smbiosview dump, I found that smbios for type 3, 4, 16, 17, 19, 32 were missing.

components/patina_smbios/examples/vendor_oem.rs
kat-perez: completely reworked the way I serialize. please review the new implementation

components/patina_smbios/src/manager.rs
kat-perez: sorry, I still needed the global state for the C/EDKII protocol interface. But for the Rust component we can still hold the manager directly and will store the manager directly in the service

components/patina_samples/src/smbios_example.rs
kat-perez: per offline conversation with @ansleythompson, we discussed adding support for types that are compatible with SMBIOS spec 3.0+. she will add these

components/patina_smbios/src/manager.rs
kat-perez: moved version and publishing the table out of the `SmbiosRecords` trait

components/patina_samples/src/lib.rs
```diff
@@ -18,3 +18,9 @@
 #![feature(coverage_attribute)]
 
 pub mod component;
+
+#[cfg(feature = "enable_smbios_example")]
```
Javagedes, line 22: I don't think you need to have this hidden behind a feature flag :)

components/patina_smbios/src/component.rs
```diff
@@ -0,0 +1,384 @@
...
+            patina::error::EfiError::Unsupported
+        })?;
+
+        // Convert boot_services to 'static by leaking it
+        // SAFETY: This is safe because boot_services remains valid for the entire lifetime
+        // of the system. We leak it intentionally to get a 'static reference.
+        let boot_services_static: &'static StandardBootServices = Box::leak(Box::new(boot_services));
+
+        // Wrap the manager in TplMutex for TPL-aware reentrancy protection at TPL_NOTIFY level
+        let manager_mutex = Box::leak(Box::new(TplMutex::new(boot_services_static, Tpl::NOTIFY, manager)));
```
Javagedes, line 130: John is currently working on providing additional options other then TplMutex. I would look into using one of those, especially if this is just to protect against re-entrency. We have already seen the TplMutex cause performance issues because when you restore the TPL, you also cause a lot of work behind the scenes with notify callbacks - https://github.com/OpenDevicePartnership/patina/blob/3be0c2d388d4b00031f2c7ad27e8e67cff0ff469/patina_dxe_core/src/events.rs#L228

components/patina_smbios/src/manager/global.rs
```diff
@@ -0,0 +1,204 @@
...
+/// level, preventing timer interrupt reentrancy. TPL is automatically restored when
+/// the lock guard is dropped.
+///
+/// # Safety
+///
+/// - Stores a 'static reference to a TplMutex (created and leaked in component.rs)
+/// - Access is protected by TplMutex.lock() which raises TPL to NOTIFY
+/// - The manager is initialized once during protocol installation
+/// - The reference remains valid for the lifetime of the system
+pub(super) struct GlobalSmbiosManager {
```
Javagedes, line 79: Looking at `manager::protocol`, all four of the protocol functions take a pointer to self. Due to that, this static is unecessary. Each of your protocol function implementations should dereference the pointer as Self.

You can stack information behind the actual protocol that contains your inner state that you are currently using as a static. This is a common practice in the C world. See the advanced logger as an example:

https://github.com/OpenDevicePartnership/patina/blob/3be0c2d388d4b00031f2c7ad27e8e67cff0ff469/components/patina_adv_logger/src/component.rs#L27

```rust
#[repr(C)]
struct AdvancedLoggerProtocolInternal<S>
where
    S: SerialIO + Send + 'static,
{
    // The public protocol that external callers will depend on.
    protocol: AdvancedLoggerProtocol,

    // Internal component access only! Does not exist in C definition.
    adv_logger: &'static AdvancedLogger<'static, S>,
}
```

```rust
fn entry_point(self, bs: StandardBootServices) -> Result<()> {
        let Some(address) = self.adv_logger.get_log_address() else {
            log::error!("Advanced logger not initialized before component entry point!");
            return Err(EfiError::NotStarted);
        };

        let protocol = AdvancedLoggerProtocolInternal {
            protocol: AdvancedLoggerProtocol::new(Self::adv_log_write, address),
            adv_logger: self.adv_logger,
        };

        let protocol = Box::leak(Box::new(protocol));
        match bs.install_protocol_interface(None, &mut protocol.protocol) {
            Err(status) => {
                log::error!("Failed to install Advanced Logger protocol! Status = {status:#x?}");
                Err(EfiError::ProtocolError)
            }
            Ok(_) => {
                log::info!("Advanced Logger protocol installed.");
                Ok(())
            }
        }
    }
```

components/patina_smbios/src/manager/global.rs
```diff
@@ -0,0 +1,204 @@
+//! Global state management for SMBIOS manager
```
Javagedes, line 1: I don't think most of, if any, of these statics are necessary as they can go into your protocol struct, packed behind the protocol itself. You can check out the advanced logger implementation.

These statics should be a last resort. They prevent propper test-ability.

components/patina_samples/src/smbios_platform.rs
```diff
@@ -0,0 +1,383 @@
...
+pub struct VendorOemRecord {
+    pub header: SmbiosTableHeader,
+    pub oem_field: u32, // Example 32-bit vendor-specific field
+    pub string_pool: alloc::vec::Vec<String>,
+}
+
+impl SmbiosRecordStructure for VendorOemRecord {
+    const RECORD_TYPE: u8 = 0x80; // Vendor-specific type (0x80-0xFF reserved for OEM)
+
+    fn to_bytes(&self) -> alloc::vec::Vec<u8> {
```
Javagedes, line 73: It looks like you are now suggesting people manually implement `SmbiosRecordStructure`. This is dangerous if the user messes it up. The user should probably be using the macro_rules macro you created in `smbios_record` to implement any custom SmbiosRecordStructure.

An even better way to do it, which I suggest in this PR or in a fture PR, would be to create a derive macro:

```rust
#[derive(SmbiosRecord)]
pub struct VendorOemRecord {
    pub header: SmbiosTableHeader,
    pub oem_field: u32, // Example 32-bit vendor-specific field
    #[string_pool]
    pub string_pool: alloc::vec::Vec<String>,
}
```

components/patina_smbios/src/smbios_record.rs
```diff
@@ -0,0 +1,987 @@
...
+/// impl_smbios_record!(
+///     Type0PlatformFirmwareInformation,
+///     0,                    // Record type
+///     string_pool,          // String pool field name
+///     vendor: u8,           // Field name: type
+///     firmware_version: u8,
+///     // ... more fields ...
+/// );
+/// ```
+macro_rules! impl_smbios_record {
```
Javagedes, line 478: This is basically a derive macro. I do think you should should turn it into one. If not in this PR then one following after. Implementing SmbiosRecordStructure is dangerous and should not be left to the user. They should use this macro and then use the derive macro once / if created.

components/patina_smbios/src/component.rs
```diff
@@ -0,0 +1,384 @@
...
+}
+
+impl Smbios {
+    /// Gets the SMBIOS version information.
+    ///
+    /// # Returns
+    ///
+    /// A tuple of (major_version, minor_version).
+    pub fn version(&self) -> (u8, u8) {
+        let manager = self.manager.lock();
```
Javagedes, line 181: version should be outside of the lock at a minimum. or using something other then the TplMutex. I just merged a PR that saw an up-to 200% performance improvement by doing somethign similar, due to the overhead of the Tpl Mutex.

https://github.com/OpenDevicePartnership/patina/pull/947

components/patina_smbios/src/component.rs
```diff
@@ -0,0 +1,384 @@
...
+    manager: &'static TplMutex<'static, SmbiosManager, StandardBootServices>,
+}
+
+impl Smbios {
+    /// Gets the SMBIOS version information.
+    ///
+    /// # Returns
+    ///
+    /// A tuple of (major_version, minor_version).
+    pub fn version(&self) -> (u8, u8) {
```
Javagedes, line 180: There is no reason in having the version be a public configuration if users can just access it via the service itself.

The interface should either be (1) Get SmBios version from public config or (2) Get SmBios version from the service itself, but not both.

components/patina_smbios/src/component.rs
```diff
@@ -0,0 +1,384 @@
...
+///     smbios.add_record(None, &bios_info)?;
+///     
+///     // Publish to configuration table
+///     smbios.publish_table(&boot_services)?;
+///     Ok(())
+/// }
+/// ```
+#[derive(IntoService, Clone, Copy)]
+#[service(Smbios)]
+pub struct Smbios {
```
Javagedes, line 170: The SmBios service interface should be in the `service` module.

The `component` should only contain the public component that should be registered with the core.

Consumers of the service expect that the service interface be in <crate>::service per the component crate requirements documentation: https://opendevicepartnership.github.io/patina/component/requirements.html

components/patina_smbios/src/component.rs
```diff
@@ -0,0 +1,384 @@
...
+    ///
+    /// # Errors
+    ///
+    /// Returns `SmbiosError` if:
+    /// - No records have been added
+    /// - Memory allocation fails
+    /// - Configuration table installation fails
+    pub fn publish_table(
+        &self,
+        boot_services: &StandardBootServices,
```
Javagedes, line 209: public interface should not require the user pass in the standardBootServices. Grab and stash the standardBootServices when you are initializing the service in the component entry point.

components/patina_smbios/src/config.rs
```diff
@@ -0,0 +1,61 @@
+//! SMBIOS Configuration
```
Javagedes, line 1: As mentioned in a previous comment, you are giving the user two ways to get the configuration - via the public configuration here or via the struct interface. I do not suggest you do that. Make it accessible via only one mechanism, and document it.

components/patina_smbios/src/manager/protocol.rs
```diff
@@ -0,0 +1,355 @@
...
+    major_version: u8,
+    minor_version: u8,
+}
+
+unsafe impl ProtocolInterface for SmbiosProtocol {
+    const PROTOCOL_GUID: efi::Guid =
+        efi::Guid::from_fields(0x03583ff6, 0xcb36, 0x4940, 0x94, 0x7e, &[0xb9, 0xb3, 0x9f, 0x4a, 0xfa, 0xf7]);
+}
+
+#[allow(dead_code)]
```
Javagedes, line 31: Is this allow(dead_code) necessary here? looks like you are actually using this in the component entry point now, so I would think this is no longer dead code.

kat-perez:
all: i need to remake this PR to use my branch from my fork instead. will close but continue resolving from comments from this PR

components/patina_samples/src/smbios_platform.rs
kat-perez: gotcha. updated so that it looks like this now

```
#[derive(SmbiosRecord)]
#[smbios(record_type = 0x80)]
pub struct VendorOemRecord {
    pub header: SmbiosTableHeader,
    pub oem_field: u32, // Example 32-bit vendor-specific field
    #[string_pool]
    pub string_pool: alloc::vec::Vec<String>,
}
```

sample updated shows usage: https://github.com/kat-perez/patina/blob/0d7e18394928ace2d6b4356dc6558127f580f92c/components/patina_samples/src/smbios_platform.rs#L69

components/patina_smbios/src/smbios_record.rs
kat-perez: impl_smbios_record is removed
@ansleythompson please follow the new derive macro that i've implemented for adding more of the generic SMBIOS record types

components/patina_smbios/src/config.rs
kat-perez: I've made the SmbiosConfiguration fields private (pub(crate)) so users can only access the SMBIOS version through the service interface, not the configuration.

components/patina_smbios/src/manager/protocol.rs
kat-perez: I removed the #[allow(dead_code)] attribute from 5 functions in protocol.rs:

SmbiosProtocol::new() (line 54) - Called by install_smbios_protocol() in the component entry point
add_ext() (line 72) - Referenced in new() when creating the protocol
update_string_ext() (line 156) - Referenced in new() when creating the protocol
remove_ext() (line 198) - Referenced in new() when creating the protocol
get_next_ext() (line 215) - Referenced in new() when creating the protocol

components/patina_smbios/src/component.rs
kat-perez: done!

components/patina_smbios/src/manager/global.rs
kat-perez: refactored the functionality into protocol.rs and removed all global statics

components/patina_smbios/src/component.rs
kat-perez: moved `version` completely outside the TplMutex

components/patina_smbios/src/manager/global.rs
kat-perez: updated with a new `SmbiosProtocolInternal`

components/patina_smbios/src/component.rs
kat-perez: maybe we can keep TplMutex for now and make a new task for later?

joschock: My recommendation is to retain TplMutex here. Incoming refactor of TplMutex in the SDK (see: https://github.com/OpenDevicePartnership/patina/pull/975) will reduce the perf impact (by removing atomics). There is some conversation about whether to expose other locking primitives outside the core besides TplMutex; that's still an open discussion. 

Restoring TPL does fire any pending event callbacks, but that's going to happen regardless - the timer tick interrupt could hit at any time and dispatch those too. That isn't a reason not to use the TplMutex, but it might be a reason to reduce the number of event callbacks elsewhere ;)

Javagedes: Ah. That was my misunderstanding. I thought we were  definitely coming out with additional primitives. This is fine then

Javagedes: Sure, that makes sense to keep TplMutex. I'm fine with contiuing to use that. It was my misunderstanding that the alternate locking primitives were going to be created, not that it is still in discussion :)

PR #914, opened 2025-10-17
[REBASE & FF] Add SAFETY Comments to patina_internal_cpu and patina_debugger

os-d:
## Description

This adds SAFETY comments to patina_internal_cpu and patina_debugger, as well as expanding usage of the read/write_sysreg macros in patina_internal_cpu and moving the unsafe block inside of the macros to reduce unsafe usage, as the operations are safe when the preconditions are met (the preconditions only being provide value register names and values).

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

N/A.

## Integration Instructions

N/A.


cfernald (approved):

kuqin12 (approved):

core/patina_internal_cpu/src/interrupts/aarch64/interrupt_manager.rs
```diff
@@ -192,7 +206,10 @@ extern "efiapi" fn synchronous_exception_handler(_exception_type: isize, context
 }
 
 fn dump_pte(far: u64) {
+    #[cfg(all(not(test), target_arch = "aarch64"))]
     let ttbr0_el2 = read_sysreg!(ttbr0_el2);
+    #[cfg(not(target_arch = "aarch64"))]
```
Javagedes, line 210: This change will cause further issues for John, as there is no definition of `ttbr0_el2` when running host based tests on an AARCH64 platform

Javagedes: Update: There are other changes in this commit that have the same issue. I'll just leave this one comment.

sdk/patina/src/arch/aarch64.rs
```diff
@@ -0,0 +1,129 @@
...
+                );
+            }
+            value
+        }
+    }
+}
+
+/// Writes the given value to the given aarch64 system register.
+/// Usage:
+/// - `write_sysreg!(reg register_name, imm:value)` - Write immediate value
```
Javagedes: nit - documentation is wrong. The final way is `imm value` not `imm:value`. I do kind of like the idea of having a `:` separating it, but this is for internal usage, not external, so I don't think it matters to much.

sdk/patina/src/arch/aarch64.rs
```diff
@@ -0,0 +1,129 @@
...
+/// - `write_sysreg!(reg register_name, imm:value, barrier)` - Write immediate value with barrier
+/// - `write_sysreg!(reg dest_register, reg src_register)` - Copy from one register to sysreg
+/// - `write_sysreg!(reg dest_register, reg src_register, barrier)` - Copy from one register to sysreg with barrier
+/// - `write_sysreg!(reg register_name, value)` - Write literal value
+/// - `write_sysreg!(reg register_name, value, barrier)` - Write literal value with barrier
+#[macro_export]
+macro_rules! write_sysreg {
```
Javagedes, line 41: nit, I find it awkward that the barrier needs to be a string. You can fix that with this:

```rust
// OLD
(reg $dest:ident, imm $imm:literal, $barrier:expr) => {

// NEW
(reg $dest:ident, imm $imm:literal, $($barrier:tt)+) => {
```

And stringifying it looks like this: `core::stringify!($($barrier)+),`

Based off everything else it may even make sense to do this:

``` rust
(reg $dest:ident, imm $imm:literal, barrier $($barrier:tt)+) => {
```

Then your macro becomes 

```rust
//this
write_sysreg!(reg dbgwcr0_el1, value, isb sy);

// Or this if you use barrier
write_sysreg!(reg dbgwcr0_el1, value, barrier isb sy);
```

Javagedes (approved):

sdk/patina/src/arch/aarch64.rs
os-d: I'll play with this. There might be multiple barriers, e.g. isb sy; dsb sy;, which I think your + should allow to work, but need to make sure.

core/patina_internal_cpu/src/interrupts/aarch64/interrupt_manager.rs
os-d: @cfernald and I are talking about how to fix these, because we don't like how this crate has structured its includes, it is not user friendly. I'm going to close this for now and we will resolve the whole crate all up.

sdk/patina/src/arch/aarch64.rs
os-d: Talked offline. Couldn't get the token tree to work with multiple barriers, kept string approach but allows multiple barriers.

makubacki (approved):

makubacki:
It doesn't need to be done here, but this is defining a more structured layout for arch-specific content. Guidance for how to add architecture-specific interfaces and code would probably be a nice addition to the [dev docs](https://github.com/OpenDevicePartnership/patina/tree/main/docs/src/dev).

os-d:
> It doesn't need to be done here, but this is defining a more structured layout for arch-specific content. Guidance for how to add architecture-specific interfaces and code would probably be a nice addition to the [dev docs](https://github.com/OpenDevicePartnership/patina/tree/main/docs/src/dev).

Agreed. @cfernald and I are working out a recommendation and will update both this crate and docs (possibly other crates, too).

PR #925, opened 2025-10-17
patina: test: Update how tests are executed

Javagedes:
## Description

Recent changes to how tests were executed and filtered with the `enable_patina_test` feature resulted in almost no testing actually being performed for the functionality. This pull-request updates how we test the `patina-test` logic, so that it does not rely on the `enable_patina_test` feature being enabled. 

It does this in two parts:

1. Creates a second static collection of patina-tests that only collects fake patina-tests from the local file.
1. Splits the main logic of `TestRunner` into two parts. The first collects the patina-tests, and the second runs the patina-tests. This allowed us properly test the patina-test execution process by passing in our second static collection of controlled patina-tests.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

CI

## Integration Instructions

N/A

### DIFF
--- a/sdk/patina/src/test.rs
+++ b/sdk/patina/src/test.rs
@@ -218,8 +218,14 @@ impl TestRunner {
     }
 
     /// The entry point for the test runner component.
+    #[coverage(off)]
     fn entry_point(self, storage: &mut Storage) -> patina::error::Result<()> {
         let test_list: &[__private_api::TestCase] = __private_api::test_cases();
+        self.run_tests(test_list, storage)
+    }
+
+    /// Runs the provided list of test cases, applying the configuration options set on the TestRunner.
+    fn run_tests(&self, test_list: &[__private_api::TestCase], storage: &mut Storage) -> patina::error::Result<()> {
         let count = test_list.len();
         match count {
             0 => log::warn!("No Tests Found"),
@@ -259,10 +265,13 @@ mod tests {
     use crate::component::{IntoComponent, Storage, params::Config};
 
     // A test function where we mock DxeComponentInterface to return what we want for the test.
-    #[allow(unused)]
     fn test_function(config: Config<i32>) -> Result<(), &'static str> {
         assert!(*config == 1);
         Ok(())
+    }
+
+    fn test_function_fail() -> Result<(), &'static str> {
+        Err("Intentional Failure")
     }
 
     #[test]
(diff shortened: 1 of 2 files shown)


sdk/patina/src/test.rs
```diff
@@ -307,20 +318,56 @@ mod tests {
...
+        let mut test_cases = vec![];
+        let mut storage = Storage::new();
         storage.add_config(1_i32);
 
-        let mut component = super::TestRunner::default().fail_fast(true).into_component();
-        component.initialize(&mut storage);
-        let _ = component.run(&mut storage);
+        let component = super::TestRunner::default().fail_fast(true);
+        let result = component.run_tests(&test_cases, &mut storage);
+        assert!(result.is_ok());
```
apop5, line 356: general question: if a test fails, will this assert stop execution at this point?

If its possible, I think we would still want to execute the rest of the tests so the developer can get a full list of what failed and be able to fix them in one go

vineelko (approved):

sdk/patina/src/test.rs
Javagedes: The patina-test functionality is for on-system unit tests. It provides `assert!` alternatives - namely `u_assert!`, etc. which instead of panicing, immediately return from the test in error.

On the test runner side, you can configure the test runner (that is the `fail_fast` setting) to either immediately return from the component in failure, or to finish all tests and return in failure at that point.

With that said, changes that will be happening soon remove the `fail_fast` functionality, as we will be allowing users to register unit tests to run on certain events. Since that is going to be added, we are removing the concept of failing fast.

makubacki:
@javagedes, is it correct that there are two main changes here?

1. Move the main logic from the `TestRunner` entry point to a separate `run_tests()` function.
2. Creating a separate test slice (`TESTS_TESTS`) to prevent it getting tangled up with `enable_patina_tests` in `TEST_CASES`

If so, the changes look fine, but could you please elaborate on (2) in the PR description (now for the releasee notes)? Also, calling out unit tests where relevant in the PR description would make it a little easier to comprehend given testing is a little overloaded in this area of the code.

Javagedes:
> @Javagedes, is it correct that there are two main changes here?
> 
> 1. Move the main logic from the `TestRunner` entry point to a separate `run_tests()` function.
> 2. Creating a separate test slice (`TESTS_TESTS`) to prevent it getting tangled up with `enable_patina_tests` in `TEST_CASES`
> 
> If so, the changes look fine, but could you please elaborate on (2) in the PR description (now for the releasee notes)? Also, calling out unit tests where relevant in the PR description would make it a little easier to comprehend given testing is a little overloaded in this area of the code.

I missed this. But will be updating the pull request, which should then sync with the release when the release syncer runs. If not, I'll update the release manually.

The focal point of this change is to allow proper testing of this code without the `enable_patina_tests` feature being enabled. It completes this by doing (1) and (2) that you specified.

PR #928, opened 2025-10-17
patina_adv_logger: Add safety comments for unsafe blocks

cfernald:
## Description

This commit adds missing safety comments to all unsafe blocks in the patina_adv_logger component.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

N/A

## Integration Instructions

Callers must now use unsafe on `init_advanced_logger` to ensure a valid hob pointer

### DIFF
--- a/components/patina_samples/src/component.rs
+++ b/components/patina_samples/src/component.rs
@@ -1,0 +1,12 @@
+//! Sample Component Implementations
+//!
+//! This module contains example component implementations demonstrating various
+//! Patina component patterns.
+//!
+//! ## License
+//!
+//! Copyright (c) Microsoft Corporation.
+//!
+//! SPDX-License-Identifier: Apache-2.0
+//!
+pub mod hello_world;
--- a/components/patina_samples/src/component/hello_world.rs
+++ b/components/patina_samples/src/component/hello_world.rs
@@ -1,6 +1,6 @@
-//! A Hello world component implementation example using a struct component.
+//! Hello World Component Examples
 //!
-//! A simple component implementation used to demonstrate how to build a component.
+//! This module demonstrates basic component implementations using both struct and enum types.
 //!
 //! ## License
 //!
--- a/components/patina_samples/src/lib.rs
+++ b/components/patina_samples/src/lib.rs
@@ -1,6 +1,12 @@
-//! Hello World Sample Components
+//! Sample Patina Components
 //!
-//! A simple component used for demonstration.
+//! This crate provides example component implementations demonstrating various
+//! Patina component patterns and usage models.
+//!
+//! ## Examples
+//!
+//! - [`component::hello_world::HelloStruct`]: Demonstrates a struct-based component with default entry point
+//! - [`component::hello_world::GreetingsEnum`]: Demonstrates an enum-based component with custom entry point
 //!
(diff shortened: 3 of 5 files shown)


components/patina_adv_logger/src/component.rs
```diff
@@ -62,6 +62,7 @@ where
     pub fn init_advanced_logger(&self, physical_hob_list: *const c_void) -> Result<()> {
         debug_assert!(!physical_hob_list.is_null(), "Could not initialize adv logger due to null hob list.");
         let hob_list_info =
+            // SAFETY: The caller must provide a valid physical HOB list pointer.
```
makubacki, line 70: The method should be marked unsafe with this in the function's safety doc.

components/patina_adv_logger/src/memory_log.rs
```diff
@@ -582,7 +585,7 @@ mod tests {
             log.add_log_entry(entry).unwrap();
         }
 
-        // adopt the log.
+        // SAFETY: This is the saem buffer as before, still valid.
```
os-d:
```suggestion
        // SAFETY: This is the same buffer as before, still valid.
```

os-d (approved):

components/patina_adv_logger/src/component.rs
cfernald: Done. @makubacki , where did we land on labeling "breaking" changes for things like there where we cause a compile time break but without signifficant interface change?

makubacki: Good reminder that we should follow what was recently agreed upon - [Criteria for Breaking Changes](https://github.com/OpenDevicePartnership/patina/blob/main/docs/src/rfc/text/0015-patina-release-process.md#criteria-for-breaking-changes).

components/patina_adv_logger/src/integration_test.rs
```diff
@@ -24,6 +24,9 @@ fn adv_logger_test(bs: StandardBootServices) -> patina::test::Result {
 
     // Get a reference to the advanced logger buffer. The actual transport does
     // not matter so use the NULL implementation as a stand-in.
+    //
+    // SAFETY: The advanced logger protocol should have been installed by the
+    //         component initialization of the same type.
```
makubacki: Could this be made a little more clear? Does "same type" mean "same instance" here or is it literally referring to a type?

components/patina_adv_logger/src/component.rs
cfernald: Thanks!

components/patina_adv_logger/src/memory_log.rs
```diff
@@ -68,6 +68,7 @@ pub struct AdvancedLog<'a> {
 //         the log. Safety here is checked by the allocation logic in add_log_entry
 //         which relies on atomics to safely allocate portions of the buffer.
 unsafe impl Send for AdvancedLog<'static> {}
+// SAFETY: See above comment
```
makubacki: Might be a little more robust to say:
```suggestion
// SAFETY: See the Send safety comment
```

PR #930, opened 2025-10-17
patina_debugger: Fix OS lock check for aarch64

cfernald:
## Description

When the OS lock register is set, certain debug features like single stepping will not be allowed. This has been observed on aarch64 systems where the OS lock is set initially. This fixes a typo on the initial implementation that checked the wrong bit in the system register that shows the status of the OS lock.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- [x] Tested on physical aarch64 system

## Integration Instructions

N/A


joschock (approved):

makubacki (approved):
nit: The PR description has an incomplete sentence: "When the OS lock register is set, certain debug features like single stepping."

PR #931, opened 2025-10-18
Change the image handle for the EFI System Table Pointer structure to EfiRuntimeServicesData.

joschock:
## Description

This PR changes the image handle for the allocation for the EFI System Table Pointer structure to be EfiBootServicesData; the prior use of the DXE_CORE_HANDLE caused the region to appear as free in the memory map.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Confirmed windows boot with the change, verified memory map by inspection.

## Integration Instructions

N/A

### DIFF
--- a/patina_dxe_core/src/config_tables/debug_image_info_table.rs
+++ b/patina_dxe_core/src/config_tables/debug_image_info_table.rs
@@ -143,7 +143,7 @@ pub(crate) fn initialize_debug_image_info_table(system_table: &mut EfiSystemTabl
         GcdMemoryType::SystemMemory,
         ALIGNMENT_SHIFT_4MB,
         UEFI_PAGE_SIZE,
-        protocol_db::DXE_CORE_HANDLE,
+        protocol_db::EFI_BOOT_SERVICES_DATA_ALLOCATOR_HANDLE,
         None,
     ) {
         Ok(address) => address,


patina_dxe_core/src/config_tables/debug_image_info_table.rs
```diff
@@ -143,7 +143,7 @@ pub(crate) fn initialize_debug_image_info_table(system_table: &mut EfiSystemTabl
         GcdMemoryType::SystemMemory,
         ALIGNMENT_SHIFT_4MB,
         UEFI_PAGE_SIZE,
-        protocol_db::DXE_CORE_HANDLE,
+        protocol_db::EFI_RUNTIME_SERVICES_DATA_ALLOCATOR_HANDLE,
```
os-d: I know we chatted about this offline, but then I looked at edk2, and it is allocated as boot services data there: https://github.com/tianocore/edk2/blob/0fa57975b054b36a90fb8e56f4377c86e672c885/MdeModulePkg/Core/Dxe/Misc/DebugImageInfo.c#L63, so I think we should follow that. This doesn't need to persist into the OS timeframe, we don't even have the exception handlers at some point after the boot loader executes and before EBS, so we should let the OS reclaim this page.

joschock: Cool - will push an update to EBS.

joschock: pushed.

PR #932, opened 2025-10-19
Add Interrupt-based locking primitives and refactor TplLock. 

### LINKED ISSUE #897: [Feature]: Create new locking primitives that do not use Atomics
### Feature Overview

Create new non-Atomics locking primitives as described in the RFC: Remove Atomics (https://github.com/OpenDevicePartnership/patina/pull/810)

Excerpt as follows: 

This RFC proposes the implementation of new "interrupt-only" locking primitives to supplement the existing `tpl_lock`
primitives presently implemented. Two new primitives are proposed that have similar semantics to existing rust mutex
idioms (such as those in `tpl_lock` or from the `spin::mutex` crate) - i.e., acquiring a lock will return a "Guard"
instance, and the lock is released when the "Guard" object is dropped.

`MutexLock<T>`  - This lock will use a volatile bool with temporary suspension of interrupts to detect and panic on
reentrancy to ensure mutual exclusion. Interrupts will only be suspended while attempting to acquire the lock; once the
lock is required interrupts will be enabled for the lifetime of the resulting Guard object.
`InterruptLock<T>` - This lock will suspend interrupts for the lifetime of the corresponding Guard object. This allows
the critical section protected by the guard to execute without potentially being interrupted. As with `MutexLock`, a
volatile bool will be used to detect and panic on reentrancy to ensure mutual exclusion.
`tpl_lock` - This RFC proposes the removal of the existing `tpl_lock` module within the Patina core in favor of the
existing `tpl_mutex` implementation within the Patina SDK which will be reworked to remove atomics.

A new `lock` module will be implemented within the patina core to contain the implementation of these new primitives.
Patina components needing lock functionality are expected to use the `tpl_mutex` within the Patina SDK. Components
are discouraged from directly interacting with interrupts as a means of implementing mutual exclusion and should use a
`tpl_mutex` for that.

### Solution Overview

See description above. 

### Alternatives Considered

See related RFC.

### Urgency

Medium

I will implement the feature

No maintainer feedback needed

### Anything else?

_No response_

joschock:
## Description

Implementation of interrupt-based locking primitives as described in this RFC: https://github.com/OpenDevicePartnership/patina/pull/810

Closes #897 

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Passes unit test and basic boot test on hardware platforms.

## Integration Instructions

N/A.

### DIFF
--- a/patina_dxe_core/src/locks/tpl_mutex.rs
+++ b/patina_dxe_core/src/locks/tpl_mutex.rs
@@ -1,6 +1,12 @@
-//! UEFI Task Priority Level (TPL) Locking support
+//! TplMutex: Mutex implementation that also adjusts UEFI TPL levels
 //!
-//! This module provides a Mutex implementation based on UEFI TPL levels.
+//! This module uses a Mutex to provide mutual exclusion, but also raises and
+//! lowers the UEFI TPL level when the mutex is locked and unlocked.
+//!
+//! This mutex guarantees that the critical section protected by the guard
+//! cannot be interrupted by code running at TPL equal to or lower than the
+//! lock's TPL level. At TPL_HIGH_LEVEL, interrupts are disabled, so that
+//! means that the critical section cannot be interrupted by anything.
 //!
 //! ## License
 //!
@@ -8,48 +14,26 @@
 //!
 //! SPDX-License-Identifier: Apache-2.0
 use core::{
-    cell::UnsafeCell,
     fmt,
     ops::{Deref, DerefMut},
-    sync::atomic::{AtomicBool, AtomicPtr, Ordering},
 };
 
 use r_efi::efi;
 
-static BOOT_SERVICES_PTR: AtomicPtr<efi::BootServices> = AtomicPtr::new(core::ptr::null_mut());
-
-/// Called to initialize the global TplLock BootServices pointer. Prior to this call, TPL locks are collapsed to a basic
-/// lock with no TPL interaction. Afterwards, all TPL locks will adjust TPL according to the TPL they were initialized
-/// with.
-///
-// Design Note: While it would be preferable to avoid a global static BOOT_SERVICES_PTR, the alternative would require
-// boot services to be available whenever a new lock is instantiated. This would have two drawbacks: 1) it would mean
-// that lock instantiation could not be `const` - and therefore could not be used to easily initialize global locked
-// statics (which is a primary use case for this crate), and 2) it would mean that locks could not be instantiated
-// before boot services creation. Since these locks are used in many of the structures that are used to implement boot
-// services, this would introduce a cyclical dependency.
-pub fn init_boot_services(boot_services: *mut efi::BootServices) {
-    BOOT_SERVICES_PTR.store(boot_services, Ordering::SeqCst);
-}
-
-fn boot_services() -> Option<&'static mut efi::BootServices> {
-    let boot_services_ptr = BOOT_SERVICES_PTR.load(Ordering::SeqCst);
-    unsafe { boot_services_ptr.as_mut() }
-}
+use crate::{
+    events::{raise_tpl, restore_tpl},
+    locks::interruptible_mutex::{InterruptibleMutex, InterruptibleMutexGuard},
+};
 
 /// Used to guard data with a locked MUTEX and TPL level.
 pub struct TplMutex<T: ?Sized> {
     tpl_lock_level: efi::Tpl,
-    lock: AtomicBool,
-    name: &'static str,
-    data: UnsafeCell<T>,
+    data: InterruptibleMutex<T>,
 }
 /// Wrapper for guarded data, which can be accessed by Deref or DerefMut on this object.
 pub struct TplGuard<'a, T: ?Sized + 'a> {
-    release_tpl: Option<efi::Tpl>,
-    lock: &'a AtomicBool,
-    name: &'static str,
-    data: *mut T,
+    release_tpl: efi::Tpl,
+    guard: Option<InterruptibleMutexGuard<'a, T>>,
 }
 
 unsafe impl<T: ?Sized + Send> Sync for TplMutex<T> {}
@@ -61,7 +45,7 @@ unsafe impl<T: ?Sized + Send> Send for TplGuard<'_, T> {}
 impl<T> TplMutex<T> {
     /// Instantiates a new TplMutex with the given TPL level, data object, and name string.
     pub const fn new(tpl_lock_level: efi::Tpl, data: T, name: &'static str) -> Self {
-        Self { tpl_lock_level, lock: AtomicBool::new(false), data: UnsafeCell::new(data), name }
+        Self { tpl_lock_level, data: InterruptibleMutex::new(data, name) }
     }
 }
 
@@ -71,21 +55,16 @@ impl<T: ?Sized> TplMutex<T> {
     ///
     /// Safety: Lock reentrance is not supported; attempt to re-lock something already locked will panic.
     pub fn lock(&self) -> TplGuard<'_, T> {
-        self.try_lock().unwrap_or_else(|| panic!("Re-entrant locks for {:?} not permitted.", self.name))
+        self.try_lock().unwrap_or_else(|| panic!("Re-entrant locks for {:?} not permitted.", self.data.name()))
     }
 
     /// Attempts to lock the TplMutex, and if successful, returns a guard object that can be used to access the data.
     pub fn try_lock(&self) -> Option<TplGuard<'_, T>> {
-        let boot_services = boot_services();
-        let release_tpl = boot_services.as_ref().map(|bs| (bs.raise_tpl)(self.tpl_lock_level));
-        if self.lock.compare_exchange(false, true, Ordering::Acquire, Ordering::Relaxed).is_ok() {
-            Some(TplGuard { release_tpl, lock: &self.lock, name: self.name, data: unsafe { &mut *self.data.get() } })
+        let release_tpl = raise_tpl(self.tpl_lock_level);
+        if let Some(guard) = self.data.try_lock() {
+            Some(TplGuard { release_tpl, guard: Some(guard) })
         } else {
-            if let Some(release_tpl) = release_tpl
-                && let Some(bs) = boot_services
-            {
-                (bs.restore_tpl)(release_tpl);
-            }
+            restore_tpl(release_tpl);
             None
         }
     }
@@ -94,8 +73,14 @@ impl<T: ?Sized> TplMutex<T> {
 impl<T: ?Sized + fmt::Debug> fmt::Debug for TplMutex<T> {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match self.try_lock() {
-            Some(guard) => write!(f, "Mutex {{ data: ").and_then(|()| (*guard).fmt(f)).and_then(|()| write!(f, "}}")),
-            None => write!(f, "Mutex {{ <locked> }}"),
+            Some(guard) => write!(
+                f,
+                "TplMutex {{ lock_tpl: {:x?}, release_tpl: {:x?}, data: ",
+                self.tpl_lock_level, guard.release_tpl
+            )
+            .and_then(|()| (*guard).fmt(f))
+            .and_then(|()| write!(f, " }}")),
+            None => write!(f, "TplMutex {{ lock_tpl: {:x?}, data: <locked> }}", self.tpl_lock_level),
         }
     }
 }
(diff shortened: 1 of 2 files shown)


patina_dxe_core/src/locks/interrupt_mutex.rs
```diff
@@ -0,0 +1,228 @@
...
+    ///
+    /// Safety: Lock reentrance is not supported; attempt to re-lock something already locked will panic.
+    pub fn lock(&self) -> InterruptGuard<'_, T> {
+        self.try_lock().unwrap_or_else(|| panic!("Re-entrant locks for {:?} not permitted.", self.name))
+    }
+
+    /// Attempts to lock the InterruptMutex, and if successful, returns a guard object that can be used to access the data.
+    pub fn try_lock(&self) -> Option<InterruptGuard<'_, T>> {
+        let interrupts_enabled = interrupt_control::get_interrupt_state();
+        interrupt_control::disable_interrupts();
```
cfernald: I'm a little paranoid about the compiler doing odd stuff. We have no guarantee that that it wont reorder the interrupt and volatile write do we? Perhaps compiler barriers can help?

patina_dxe_core/src/locks/interrupt_mutex.rs
```diff
@@ -0,0 +1,228 @@
...
+    pub const fn new(data: T, name: &'static str) -> Self {
+        Self { locked: UnsafeCell::new(false), name, data: UnsafeCell::new(data) }
+    }
+}
+
+// Interrupt control functions. Broken out to allow for mocking in tests.
+#[cfg(not(test))]
+mod interrupt_control {
+    pub fn get_interrupt_state() -> bool {
+        patina_internal_cpu::interrupts::get_interrupt_state().expect("Interrupts non-functional.")
```
cfernald: I know it's still a draft, but a question on this. Historically, we have stated that SDK should not depend on core crates. Do you intend on moving this functionality into the SDK for basic interrupt instructions?

joschock: I thought a bit about this, and I might still move these into the SDK depending on how things shake out in removing atomics from the SDK. This would also require moving the CPU stuff into the SDK and possible other entrained dependencies, so I'm hoping not to do that. For now, I'm planning to keep them in the core. I'm also a bit hesitant about exposing the interrupt-based mutexes outside of the core, because that creates a dual locking regime (interrupt vs. TPL), and, in general, I think TPL-based locking is the primary approach that consumers of SDK (or really anything outside the core) should be using.

cfernald: When you say "exposing the interrupt-based mutexes" do you mean just `InterruptMutex` as defined here or also `Mutex`? If you mean just `InterruptMutex` then I agree that, but for `Mutex` the problem is that I think if we are going to say "you shouldnt use atomics" to components, we need to provide a robust suite of atomic-free locks for them to use. For example, components have a lot of reasons to have internal mutability that has nothing to do with TPL or UEFI-constructs, and doing so without unsafes requires something like a Mutex, Once, RwLock, etc.

I think that it's reasonable for us to only expose TplMutex for system wide locking, and a suite of other interior mutability tools that may use `InterruptMutex` internally, but only to avoid atomics. But that also gives us freedom to change how the locks work in the future without breaking consumer abstractions.

joschock: > "components have a lot of reasons to have internal mutability that has nothing to do with TPL or UEFI constructs"

A component that only cares about interior mutability and does not care about concurrency protection can use e.g. `RefCell` or `Cell`.  

For components that _do_ care about concurrency (i.e. they want to have mutability on global state accessible from different interrupt contexts, which is pretty much what UEFI concurrency presently boils down to) - I'm not sure I buy this position yet. The component itself may not care about TPL level at all; but in the UEFI context outside the core (I'd argue this is presently true even for patina components, not just external drivers), the concept of interrupts is pretty tightly bound with the concept of TPL. Using a TPL forces a component to state (at least a coarse level) who it can be interrupted by for various critical sections/APIs. 

Maybe an example of why a component would _want_ to avoid specifying a TPL (or using a boilerplate alias that does so on its behalf) would be useful to helping me think this through?

joschock: Based on offline discussion and findings from actually implementing the locks, TPL is the way to go for most locks as it helps to avoid subtle lock breakage if TPL manipulation is occurring while a lock is held.

patina_dxe_core/src/locks/mutex.rs
```diff
@@ -0,0 +1,200 @@
+//! Mutex: enforces mutual exclusion for a critical section while leaving
```
cfernald: What is your rationale for keeping this internal to dxe_core as opposed to having some standard mutex's like this available in the SDK?

patina_dxe_core/src/locks/interrupt_mutex.rs
```diff
@@ -0,0 +1,242 @@
...
+    pub fn lock(&self) -> InterruptGuard<'_, T> {
+        self.try_lock().unwrap_or_else(|| panic!("Re-entrant locks for {:?} not permitted.", self.name))
+    }
+
+    /// Attempts to lock the InterruptMutex, and if successful, returns a guard object that can be used to access the data.
+    pub fn try_lock(&self) -> Option<InterruptGuard<'_, T>> {
+        let interrupts_enabled = disable_interrupts();
+        // SAFETY: Interrupts are disabled.
+        if unsafe { self.locked.get().read_volatile() } {
+            // Lock is already held, restore interrupt state and fail to acquire lock.
```
cfernald, line 100: should you `debug_assert` that `interrupts_enabled` is false? I think restoring it makes sense for completeness, but theoretically it should always be a no-op unless someone is breaking the expectations of this lock.

patina_dxe_core/src/locks/mutex.rs
```diff
@@ -0,0 +1,200 @@
...
+unsafe impl<T: ?Sized + Sync> Sync for MutexGuard<'_, T> {}
+// Safety: InterruptMutex is used to protect against concurrent access to lock flag.
+unsafe impl<T: ?Sized + Send> Send for MutexGuard<'_, T> {}
+
+impl<T> Mutex<T> {
+    #[allow(unused)]
+    pub const fn new(data: T, name: &'static str) -> Self {
+        // SAFETY: this implementation upholds the safety requirement that interrupts are not enabled while the
+        // InterruptMutex lock is held.
+        unsafe { Self { lock: InterruptMutex::new(false, name), name, data: UnsafeCell::new(data) } }
```
cfernald: since the name is already in the interrupt lock, and we remove it from Mutex and just use the interrupts `name()` to save some redundant space?

patina_dxe_core/src/locks/interruptible_mutex.rs
```diff
@@ -0,0 +1,200 @@
...
+    #[allow(unused)]
+    pub fn lock(&self) -> MutexGuard<'_, T> {
+        self.try_lock().unwrap_or_else(|| panic!("Re-entrant locks for {:?} not permitted.", self.name))
+    }
+
+    /// Attempts to lock the Mutex, and if successful, returns a guard object that can be used to access the data.
+    pub fn try_lock(&self) -> Option<MutexGuard<'_, T>> {
+        let mut lock = self.lock.try_lock()?;
+
+        if *lock {
```
cfernald, line 69: I would clarify in a comment why these aren't volatile accesses like they are in `InterruptMutex`. I assume it's because they are wrapped to a struct that has a lifetime of another locked structure, but it gave me pause when reading through it.

patina_dxe_core/src/locks/mutex.rs
```diff
@@ -0,0 +1,200 @@
...
+    }
+}
+
+impl<T: ?Sized> Drop for MutexGuard<'_, T> {
+    fn drop(&mut self) {
+        // Release the lock.
+        // Note: any other context that might try_lock() (such as an interrupt handler) would be interrupting this
+        // context and would release the lock before resuming this context. This means that try_lock() here should
+        // always succeed. Use a try_lock().expect() to aid in debugging if this is not the case.
+        let mut lock = self.lock.try_lock().expect("Lock acquisition failed during drop");
```
cfernald: Might as well use the name, since we have it.

patina_dxe_core/src/locks/mutex.rs
joschock: The main reason is it would require access to the interrupt primitives which presently don't exist in the SDK. I could make some, or move some to the SDK, but decided not to for now. The standard mutex (which everything outside the core should use or derive from) is the TplMutex. I will have a subsequent PR to remove atomics from that one.

cfernald: My concern with only exposing TplMutex for all non dxe_core crates is, that it's pretty unergonomic since it requires a boot services reference to initialize. This means that in affect, even though it's const, you can't really use it for some of the key scenarios such as locking a static or interior mutability for components. 

It doesn't need to be addressed by this PR, but I think for us to achieve moving folks off atomic's based locks we may need to improve those ergonomics or provide lighter weight locks (that don't disable interrupts outside of their internals).

patina_dxe_core/src/locks/interrupt_mutex.rs
joschock: It's not always the case that interrupts_enabled is true at lock acquisition. Consider:

```
let a = InterruptMutex<()>::new((), "A");
let b = InterruptMutex<()>::new((), "B");
let _guard_a = a.lock();
let _guard_b = b.lock(); //interrupts are disabled here because of 'a' guard. 
drop(_guard_b);
drop(_guard_a);
```

patina_dxe_core/src/locks/mutex.rs
joschock: Yep, I'll make that change.

patina_dxe_core/src/locks/interrupt_mutex.rs
cfernald: My point is that in the scenario where the lock is already acquired (as checked buy line 96) we shouldn't be able to have interrupts enabled, as the contract of this lock is that if the lock is acquired interrupts will be disabled, right?

patina_dxe_core/src/locks/interruptible_mutex.rs
joschock: I am kind of unsteady on ARM's weakly ordered memory model - so I'm open to learning more whether we need ordering here. But - in InterruptMutex, I made the reads/writes volatile to avoid re-order with respect to the interrupt control (mostly out of paranoia - I am not certain that is required). Here, I'm relying on the processor to ensure that data writes in this context are visible before data reads to the same data in this context. `*lock` is an InterruptGuard - so that means that interrupts are disabled here, so the loads and stores to the data behind `lock`, even if they are re-ordered, should be done in a single-threaded context.

patina_dxe_core/src/locks/mutex.rs
joschock: Will do.

patina_dxe_core/src/locks/interrupt_mutex.rs
```diff
@@ -0,0 +1,242 @@
+//! InterruptMutex: A mutex that disables interrupts to provide mutual
+//! exclusion.
+//!
+//! Critical sections are protected by obtaining a `InterruptGuard` object from
```
os-d:
```suggestion
//! Critical sections are protected by obtaining an `InterruptGuard` object from
```

patina_dxe_core/src/locks/interrupt_mutex.rs
```diff
@@ -0,0 +1,242 @@
...
+fn interupts_enabled() -> bool {
+    patina_internal_cpu::interrupts::get_interrupt_state().expect("Interrupts non-functional.")
+}
+
+fn disable_interrupts() -> bool {
+    let current_state = interupts_enabled();
+    if current_state {
+        patina_internal_cpu::interrupts::disable_interrupts();
+    }
+    current_state
```
os-d, line 76: nit: function comment since this is not necessarily intuitive that it returns the state of the interrupts before disabling them.

patina_dxe_core/src/locks/interrupt_mutex.rs
```diff
@@ -0,0 +1,242 @@
...
+}
+
+fn restore_interrupts(state: bool) {
+    if state {
+        patina_internal_cpu::interrupts::enable_interrupts()
+    }
+}
+
+impl<T: ?Sized> InterruptMutex<T> {
+    /// Lock the InterruptMutex and return a InterruptGuard object used to access the data.
```
os-d:
```suggestion
    /// Lock the InterruptMutex and return an InterruptGuard object used to access the data.
```

patina_dxe_core/src/locks/interrupt_mutex.rs
```diff
@@ -0,0 +1,242 @@
...
+    /// violate the safety contract.
+    ///
+    #[allow(dead_code)]
+    pub const unsafe fn new(data: T, name: &'static str) -> Self {
+        Self { locked: UnsafeCell::new(false), name, data: UnsafeCell::new(data) }
+    }
+}
+
+fn interupts_enabled() -> bool {
+    patina_internal_cpu::interrupts::get_interrupt_state().expect("Interrupts non-functional.")
```
os-d, line 67: tbh, get_interrupt_state shouldn't return a result. It can't fail. I would suggest updating that to have no return value.

patina_dxe_core/src/locks/interrupt_mutex.rs
os-d: I haven't finished reading through the PR yet, but what about the case of:

```rust
let a = TplMutex(); // interrupts are disabled
let b = InterruptMutext();
let _ = b.lock(); // interrupts are already disabled
```

Point being, interrupts could disabled outside of this lock, too, right? Or am I missing the semantics?

joschock: you mean if something manages to enable interrupts between line 94 and 97? That would certainly break things. I can add a debug print, but I'm not sure how that is possible?

os-d: Ignore mine, I misread Chris's statement.

I agree with Chris, when `try_lock` is called and the lock is already held, interrupts should be disabled.

i.e.

```rust
let a = InterruptMutex();
let _ = a.lock(); // interrupts are disabled, a is held
let _ = a.try_lock(); // Rust would block this exact example, but point being that if a is attempted to be re-entrantly locked, interrupts should still be disabled
```

So it isn't that interrupts were re-enabled between lines 94 and 97, its that they may be re-enabled between a.lock() and a.try_lock(), without dropping the lock.

os-d: You do check this in the drop routine, so probably don't also need it here, but wouldn't hurt to also debug_assert here.

cfernald: Yes, the assertion is "It should be impossible for the lock to already be acquired but for interrupts to have been enabled" A debug assert on line 97/98 could check that.

patina_dxe_core/src/locks/interruptible_mutex.rs
cfernald: Well the volatile only affects compiler behavior, not the actual ISA, right? So, in this case any use of volatile would be to prevent the compiler for stashing values or swapping with other volatile accesses. I think your logic is fine as it stands because it is tied to the lifetime of the internal lock. 

Perhaps this doesn't need to be addressed and I'm just looking at it too hard and too paranoid 😊.

patina_dxe_core/src/locks/tpl_mutex.rs
```diff
@@ -0,0 +1,212 @@
...
+    /// to the level specified at TplMutex creation.
+    ///
+    /// Safety: Lock reentrance is not supported; attempt to re-lock something already locked will panic.
+    pub fn lock(&self) -> TplGuard<'_, T> {
+        self.try_lock().unwrap_or_else(|| panic!("Re-entrant locks for {:?} not permitted.", self.data.name()))
+    }
+
+    /// Attempts to lock the TplMutex, and if successful, returns a guard object that can be used to access the data.
+    pub fn try_lock(&self) -> Option<TplGuard<'_, T>> {
+        let release_tpl = raise_tpl(self.tpl_lock_level);
```
os-d, line 63: Probably a comment is worthwhile going in here, let me make sure my understanding is correct:

For all TPLs != TPL_HIGH: we use the Mutex to briefly disable interrupts while acquiring the lock, then run with interrupts enabled.

For TPL_HIGH, we do the same thing, but are relying on raise_tpl to disable interrupts and lower_tpl to lower interrupts.

i.e. that InterruptMutex guarantees the interrupt semantics, but TplMutex does not; it relies on [raise|lower]_tpl to handle the interrupts.

patina_dxe_core/src/locks/interrupt_mutex.rs
joschock: So this would protect against this (to extend the example):
```
let a = unsafe {InterruptMutex(...)};
let guard = a.lock();
enable_interrupts(); // bad, voilates the explicit safety contract required to create "a".
let guard2 = a.try_lock(); //should explode.
drop(guard); // will explode due to check in Drop.
```
I'm on board, I'll add a debug_assert!() to the try_lock() fail case if interrupts are observed enabled.

joschock: weird. GH plugin in vscode is scrambling my respoonses.

patina_dxe_core/src/locks/mutex.rs
```diff
@@ -0,0 +1,200 @@
...
+//!
+use core::{
+    cell::UnsafeCell,
+    fmt,
+    ops::{Deref, DerefMut},
+};
+
+use crate::locks::interrupt_mutex::InterruptMutex;
+
+pub struct Mutex<T: ?Sized> {
```
os-d: nit: while this is the best name to describe what we are doing, I'm hesitant to use a well known name as our own name. I'm worried folks will accidentally use core::_::Mutex instead of our Mutex, when they mean to use ours. And we would only catch that in the `use` statement, which is easily overlooked in code review. I would consider naming this something else, though I don't have great suggestions because things like `PatinaMutex` aren't great.

patina_dxe_core/src/locks/tpl_mutex.rs
joschock: Correct. TplMutex makes the guarantee that your critical section (including acquisition of the lock) is guarded against concurrency at or below the lock level specified when you create the TplMutex. This relies on the fact that (aside from special cases that should not be using TplMutex) architecturally there is only one "interrupt" - the timer tick, and handling of it is TPL-aware. 

If you tried to acquire the TplMutex at a TPL higher than specified in the lock, (i.e. you are running at a handler in TPL_NOTIFY, and you try and acquire a lock that is TPL_CALLBACK) you would panic due to TPL inversion.  

Put another way: TplMutex (except at high) don't _need_ to disable interrupts, because the TPL architecture/Timer architecture makes the necessary concurrency guarantees.

os-d: Right, I was coming from the other direction, looking for special handling for TPL_HIGH, but then realizing that that must be handled in raise_tpl

patina_dxe_core/src/locks/interrupt_mutex.rs
joschock: Not opposed, but changing the interrupt manager would be off-topic and breaking (well, not externally breaking, but breaking on the interrupt API to other parts of Patina). I'll file a low-pri bug to fix as a separate PR.

patina_dxe_core/src/locks/interrupt_mutex.rs
joschock: Will fix.

patina_dxe_core/src/locks/interruptible_mutex.rs
joschock: I'm not sure I can formulate a helpful comment. do you have a suggestion?

patina_dxe_core/src/locks/interrupt_mutex.rs
joschock: Changed in latest push.

patina_dxe_core/src/locks/mutex.rs
joschock: Fixed in latest push.

patina_dxe_core/src/locks/mutex.rs
joschock: fixed in latest push.

patina_dxe_core/src/locks/interrupt_mutex.rs
joschock: fixed in latest push.

patina_dxe_core/src/locks/interrupt_mutex.rs
joschock: added function comments in latest push.

patina_dxe_core/src/locks/interrupt_mutex.rs
joschock: fixed in latest push.

patina_dxe_core/src/locks/mutex.rs
joschock: renamed to InterruptibleMutex in latest push.

patina_dxe_core/src/locks/interrupt_mutex.rs
os-d: I'm coming late to this conversation, but I was having an offline conversation with @cfernald on this topic.

I agree with both of you :). But in the sense that I believe:
- For interior mutability only, `*Cell` is the right paradigm for components to use. I would expect the majority of components that do not expose a service or protocol and do not have event notifications would use this approach.
- `InterruptibleMutex` is a useful paradigm for a component that wants to prevent re-entrancy but does not worry about TPLs. For components that offer services but not protocols and do not have event notifications, I would anticipate this would be a useful mechanism to have.
- `TplMutex` is useful to components that do need to worry about TPLs, likely mostly those that produce protocols to interact with C code through or have event notifications.
- `InterruptMutex` should not be offered generically to consumers because of the enumerated dangers

This is the sort of generic view, of course each individual component needs to decide its mutability model and what makes sense for that component. Having a `Cell` in a component that has an event notification is reasonable, depending on the use case.

I think it would be valuable to have `TplMutext` and `InterruptibleMutex` live in the SDK. Now these depends on `InterruptMutex`, which would mean we would have to craft a way for the core to consume it but not other crates, either through code duplication (not ideal), a feature flag (still feel somewhat iffy about), or a redesign (which I don't have any better suggestions)

joschock: I could implement an `InterruptibleMutex` equivalent in the SDK side by just using a lock flag protected by a TplMutex at TPL_HIGH. This would be semantically equivalent - it would briefly halt interrupt (via TPL->HIGH) for lock acquisition and release, but would not hold the TPL during the guard (which is what TplMutex) does today. Consuming component would not need to worry about TPL, because it is always legal to raise to TPL_HIGH.

joschock: However, I have a subsequent PR for the SDK side implementation of atomics removal; so let's defer the SDK-side discussion until that PR, if that's alright.

patina_dxe_core/src/locks/mutex.rs
joschock: Yeah, agree that needing a BS instance to initialize the mutex is kind of a pain (you can see what I had to do in the perf component as in the "remove_atomics_sdk" draft PR:  https://github.com/OpenDevicePartnership/patina/pull/975/commits/dff613e5f272780df554181ef3d1384c64a2f81e).  

But I actuallt prefer the result. As a matter of architecture, I'm kind of opinionated about allowing components to affect global hw state (like interrupt control) without going through central APIs (i.e. TPL for interrupts or GCD for memory map) to allow for proper coordination. in any case, we can discuss more in the SDK PR (feel free to start the convo there now, if you want ;)).

joschock:
Closing this out on the basis of reworked RFC: https://github.com/OpenDevicePartnership/patina/pull/1036 which changes scope: interrupt-based locking proves not to be as performant as expected.

PR #937, opened 2025-10-20
Add codecov configuration

Javagedes:
## Description

This commit configures codecov with the following:
- Delay codecov from sending status until all checks are finished
- Rounds coverage percentage up or down to the nearest 2 decimal palces.
- Updates the range so that less than 80% is red, 80-89% is yellow, and 90%+ is green
- Sets the target coverage for the project and PR patch to 80% rather than keeping it approximitely what was before the commit.
- Hardcodes the comment layout, though these are the default settings.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [x] Includes documentation?

## How This Was Tested

Ran the settings through the codecov validator.

## Integration Instructions

N/A

### DIFF
--- a/docs/src/dev/testing.md
+++ b/docs/src/dev/testing.md
@@ -36,14 +36,14 @@ to help developers track local performance changes when making changes during th
 
 ## Code Coverage
 
-Unit tests and code coverage are an important aspect of our project. It is a simple to consume statistic that gives us
+Unit tests and code coverage are an important aspect of our project. It is a simple-to-consume statistic that gives us
 some limited confidence in the reliability of newly added code. Even more importantly is that it gives us some peace of
 mind that future changes will not cause unexpected regressions or breaking changes, as unit tests that exist to uphold
 certain interface expectations would begin to fail.
 
 ```admonish note
-Needing to change a test's output expectation is a good indication that your change will either impact functionality,
-or be breaking change. Any submitted PR should be marked as such.
+Needing to change a test's output expectation is a good indication that your change will either impact functionality
+or be a breaking change. Any submitted PR should be marked as such.
 ```
 
 Patina's goal is to keep code coverage over 80%. This gives some leniency for code that cannot be tested (e.g. a error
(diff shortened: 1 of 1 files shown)


codecov.yml
```diff
@@ -0,0 +1,24 @@
...
+codecov:
+  require_ci_to_pass: true
+coverage:
+  precision: 2
+  round: nearest
+  range: 80..90 # 0-79% red, 80-89% yellow, 90-100% green
+  status:
+    project:
+      default:
+        target: 80% # Target 80% coverage for the project
```
cfernald, line 18: Considering the amount of uncoverable code we have, Is the assumption here that devs should tag such code with `#coverage(off)`?

makubacki: It'd be nice to take this opportunity to add guidance to https://github.com/OpenDevicePartnership/patina/blob/main/docs/src/dev/testing/unit.md for unit test writers. Both for codcov expectations but also how and when they should ignore coverage.

Javagedes: > Considering the amount of uncoverable code we have, Is the assumption here that devs should tag such code with `#coverage(off)`?

I mean that is exactly why the coverage target is 80% instead of 100%, however yes, you can also use coverage(off). I'll update the docs and have you re-review.

Javagedes: @makubacki @cfernald I've updated docs.

docs/src/dev/testing.md
```diff
@@ -36,16 +36,37 @@ to help developers track local performance changes when making changes during th
...
-in any given repository. We use [cargo-llvm-cov](https://github.com/taiki-e/cargo-llvm-cov) as our code coverage
-reporting tool, as it works well with Windows and Linux, and can generate different report types. Each repository
-must have CI that fails if any code added to the repository has less than 80% coverage, or if the repository as a
-whole is below 80% coverage.
-
-By default, cargo-llvm-cov will produced an lcov report, which is easily consumable in various processing tools like
-[Coverage Gutters](https://marketplace.visualstudio.com/items?itemName=ryanluker.vscode-coverage-gutters). The
-`cargo make coverage` command in Patina also produces an HTML report. This is available in the
-`target/coverage/html` folder.
+Unit tests and code coverage are an important aspect of our project. It is a simple to consume statistic that gives us
```
makubacki: nit: Easier to read with this change.

```suggestion
Unit tests and code coverage are an important aspect of our project. It is a simple-to-consume statistic that gives us
```

docs/src/dev/testing.md
```diff
@@ -36,16 +36,37 @@ to help developers track local performance changes when making changes during th
...
+some limited confidence in the reliability of newly added code. Even more importantly is that it gives us some peace of
+mind that future changes will not cause unexpected regressions or breaking changes, as unit tests that exist to uphold
+certain interface expectations would begin to fail.
+
+```admonish note
+Needing to change a test's output expectation is a good indication that your change will either impact functionality,
+or be breaking change. Any submitted PR should be marked as such.
```
makubacki:
```suggestion
or be a breaking change. Any submitted PR should be marked as such.
```

docs/src/dev/testing.md
```diff
@@ -36,16 +36,37 @@ to help developers track local performance changes when making changes during th
...
+Having a hard target of code coverage (e.g. 80%) can lead to tests being written purely to bump up code coverage. If
+this is happening, consider disabling code coverage for that section of code (`#[coverage(off)]`) with valid
+justification as to why no code coverage is needed.
+```
+
+We use [cargo-llvm-cov](https://github.com/taiki-e/cargo-llvm-cov) as our code coverage reporting tool, as it works
+well with Windows and Linux, and can generate different report types. All of our repository's have CI that calculates
+and uploads code coverage results to [codecov](https://codecov.io) with an expected patch coverage of at least 80%.
+Submitting a PR with less than 80% of changed code coverged by tests will result in the check failing. It is ultimately
+up to the maintainers of patina on if they will merge the PR anyway, or require you to meet the 80% goal.
```
makubacki:
```suggestion
up to the maintainers of Patina whether they will merge the PR as is or require you to meet the 80% goal.
```

makubacki (approved):

PR #946, opened 2025-10-21
Switch to use PageAllocatorStub for page table parsing.

cfernald:
## Description

The exception handler and debugger both have definitions of a dummy page table allocator in order to parse an existing page table without actually allocating any pages. This change switches both to use the common PageAllocatorStub defined in the patina_paging crate. This should prevent some code duplication.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

N/A

## Integration Instructions

N/A

### DIFF
--- a/core/patina_internal_cpu/src/interrupts/x64/interrupt_manager.rs
+++ b/core/patina_internal_cpu/src/interrupts/x64/interrupt_manager.rs
@@ -175,7 +175,8 @@ extern "efiapi" fn page_fault_handler(_exception_type: isize, context: EfiSystem
     let paging_type =
         { if x64_context.cr4 & (1 << 12) != 0 { PagingType::Paging5Level } else { PagingType::Paging4Level } };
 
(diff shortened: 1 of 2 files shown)


core/patina_internal_cpu/src/interrupts/x64/interrupt_manager.rs
```diff
@@ -244,7 +242,13 @@ fn interpret_gp_fault_exception_data(exception_data: u64) {
 }
 
 fn dump_pte(cr2: u64, cr3: u64, paging_type: PagingType) {
-    if let Ok(pt) = unsafe { patina_paging::x64::X64PageTable::from_existing(cr3, FaultAllocator {}, paging_type) } {
+    if let Ok(pt) = unsafe {
```
makubacki, line 253: While you're here, you could add a safety comment for this unsafe block.

PR #950, opened 2025-10-23
patina_internal_depex: Set prints to debug instead of trace

zurcher:
## Description

Change depex prints to `debug!` instead of `trace!` to remove file path and line numbers from depex prints.
Map `Level::Debug` to `DEBUG_LEVEL_INFO` to allow log filtering based on AdvancedLogger init arguments instead of PCD Debug Print levels. By default, the existing Q35 and SBSA instantiations of logger use `max_level: log::LevelFilter::Info`, so they will not see any added prints due to the new Debug level mapping.

Size change of a local build with `max_level_debug` feature flag set: from 1973248 to 1981440 bytes.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Verified target_filters and max_level arguments work correctly to enable and disable Depex prints.

## Integration Instructions

Patina consumers can use the target_filters and max_level arguments of Advanced Logger to enable Depex logs:

`("patina_internal_depex", log::LevelFilter::Debug)`

### DIFF
--- a/components/patina_adv_logger/src/logger.rs
+++ b/components/patina_adv_logger/src/logger.rs
@@ -137,7 +137,7 @@ const fn log_level_to_debug_level(level: Level) -> u32 {
         Level::Warn => memory_log::DEBUG_LEVEL_WARNING,
         Level::Info => memory_log::DEBUG_LEVEL_INFO,
         Level::Trace => memory_log::DEBUG_LEVEL_VERBOSE,
-        Level::Debug => memory_log::DEBUG_LEVEL_VERBOSE,
+        Level::Debug => memory_log::DEBUG_LEVEL_INFO,
     }
 }
 
--- a/core/patina_internal_depex/src/lib.rs
+++ b/core/patina_internal_depex/src/lib.rs
@@ -152,11 +152,11 @@ impl Depex {
     /// Evaluates a DEPEX expression.
     pub fn eval(&mut self, protocols: &[efi::Guid]) -> bool {
         let mut stack = Vec::with_capacity(DEPEX_STACK_SIZE_INCREMENT);
-        log::trace!("Depex:");
+        log::debug!("Depex:");
         for (index, opcode) in self.expression.iter_mut().enumerate() {
             match opcode {
                 Opcode::Before(_) | Opcode::After(_) => {
-                    log::trace!("  {opcode:#x?}");
+                    log::debug!("  {opcode:#x?}");
                     if index != 0 {
                         debug_assert!(false, "Invalid BEFORE or AFTER not at start of depex {:#x?}", self.expression);
                         return false;
@@ -182,7 +182,7 @@ impl Depex {
                     return false;
                 }
                 Opcode::Sor => {
-                    log::trace!("  {opcode:#x?}");
+                    log::debug!("  {opcode:#x?}");
                     if index != 0 {
                         debug_assert!(false, "Invalid SOR not at start of depex.");
                         return false;
@@ -202,7 +202,7 @@ impl Depex {
                         }
                         stack.push(false);
                     }
-                    log::trace!(
+                    log::debug!(
                         "  {opcode:x?} => {:?}, stack ->{:?}",
                         stack.last(),
                         stack.iter().rev().collect::<Vec<_>>()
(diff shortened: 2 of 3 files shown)


joschock (approved):

components/patina_adv_logger/src/logger.rs
```diff
@@ -137,7 +137,7 @@ const fn log_level_to_debug_level(level: Level) -> u32 {
         Level::Warn => memory_log::DEBUG_LEVEL_WARNING,
         Level::Info => memory_log::DEBUG_LEVEL_INFO,
         Level::Trace => memory_log::DEBUG_LEVEL_VERBOSE,
-        Level::Debug => memory_log::DEBUG_LEVEL_VERBOSE,
+        Level::Debug => memory_log::DEBUG_LEVEL_INFO,
```
antklein, line 140: Do you have a rough estimate on how many logs this will now enable?
I like this change, but it might have been set to VERBOSE for a specific reason.

makubacki: Please touch on this in the PR description. And check the impact on Q35 and SBSA if you haven't already.

core/patina_internal_depex/src/lib.rs
```diff
@@ -152,11 +152,11 @@ impl Depex {
     /// Evaluates a DEPEX expression.
     pub fn eval(&mut self, protocols: &[efi::Guid]) -> bool {
         let mut stack = Vec::with_capacity(DEPEX_STACK_SIZE_INCREMENT);
-        log::trace!("Depex:");
+        log::debug!("Depex:");
```
makubacki, line 155: Should these target `patina_internal_depex`?

zurcher: These automatically target patina_internal_depex due to being inside the patina_internal_depex module

components/patina_adv_logger/src/logger.rs
zurcher: Both Q35 and SBSA instantiate the logger with `max_level: log::LevelFilter::Info` so they won't be affected by this change at all, in their default state. When building locally I use `target_filters: &[("patina_internal_depex", log::LevelFilter::Debug)]` to enable these prints.
By changing this to DEBUG_LEVEL_INFO, it allows enabling Depex prints without enabling the entire VERBOSE flag on the EDK2 side.

Javagedes: They will be affected because filtering happens at runtime. Only the log feature flags will compile out those log statements

Javagedes: Talked to @zurcher , I misread the comment. This comment was regarding the number of logs.

My statement was regarding physical size. Moving log level from trace to debug will increase the physical size of any platform that currently uses the log feature flag `max_level_debug` or `release_level_debug` :)

makubacki: To clarify my comment, I mean all aspects of the change - log output at Debug level (now being output at INFO level) before versus now and size impact. Not because the result is a breaker for the change, but to have the info for record keeping if we need to tweak this further in the future.

core/patina_internal_depex/src/lib.rs
makubacki: I missed the module name.

makubacki (approved):

components/patina_adv_logger/src/logger.rs
joschock: The code size increase should be limited to the format strings in the depex code + whatever format logic isn't LTO'd out, correct? I don't think there are very many of them in the depex prints, as voluminous as they are at runtime in terms of actual debug output. should be pretty easy to measure tho.

Javagedes: I agree it should not need much. But would be good to measure and put in the PR just to have :)

zurcher: Size change of a local build with `max_level_debug` feature flag set: from 1973248 to 1981440 bytes.
There was no size change with `max_level_trace` set; are the file paths and line numbers always baked in to debug prints? I figured moving these prints out of the `trace!` category would reduce some size there.

PR #951, opened 2025-10-23
docs: Add Unsafe.md

os-d:
## Description

This adds Patina specific guidance on unsafe usage and how that applies to system software, introducing the concepts of software vs hardware safety.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [x] Includes documentation?

## How This Was Tested

N/A.

## Integration Instructions

N/A.

### DIFF
--- a/docs/src/dev/principles/unsafe.md
+++ b/docs/src/dev/principles/unsafe.md
@@ -106,12 +106,12 @@ invariant that must hold true for that operation. There may be preconditions or 
 
 One simple example of hardware safety is the invalidate cache instruction on x86_64 that invalidates caches without
 writing them back. This operation can be very dangerous if not called at a proper time, but nothing about the software
-can change that; there is no validation that can be done to indicate this is a safe time to call this. In Patina, this
-is hardware safety and does not warrant an unsafe function.
+can change that. In Patina, this is hardware safety and does not warrant an unsafe function.
 
 We may define a wrapper function for it as:
 
-```rust
+```rust,no_run
+# use std::arch::asm;
 fn invalidate_caches() {
   // SAFETY: This is an architecturally defined way to invalidate caches
   unsafe { asm!("invld") }
@@ -120,7 +120,8 @@ fn invalidate_caches() {
 
 The concept of software safety could come into play, say in the below unsafe example:
 
-```rust
+```rust,no_run
+# use std::arch::asm;
 /// Invalidates the cache without writing them back. Also set a memory region to 4.
 ///
 /// # Parameters
@@ -128,31 +129,34 @@ The concept of software safety could come into play, say in the below unsafe exa
 ///
 /// # Safety
 /// The caller must ensure `ptr` points to a valid memory region
-unsafe fn invalidate_caches_and_set_ptr_to_4(ptr: *mut u8) {
-  // Better set this ptr to 4 first!
-  // SAFETY: See function comment, caller assumes responsibility
-  unsafe {ptr.write(4);}
-
+unsafe fn invalidate_caches_and_set_ptr_to_4(ptr: *mut u8) { 
   // SAFETY: This is an architecturally defined way to invalidate caches
-  unsafe { asm!("invld") } 
+  unsafe { asm!("invld"); }
+
+  // Better set this ptr to 4!
+  // SAFETY: See function comment, caller assumes responsibility
+  unsafe { ptr.write(4); }
 }
 ```
 
-However, the function could be made safe again by validating assumptions:
+However, the function could be made safe again by using the standard library to
+validate assumptions (or validating the assumptions ourselves):
 
-```rust
-fn invalidate_caches_and_set_ptr_to_4(ptr: Box<u8>) {
-  // Better set this ptr to 4 first!
-  *ptr = 4;
-
+```rust,no_run
+# use std::arch::asm;
+fn invalidate_caches_and_set_ptr_to_4(mut ptr: Box<u8>) { 
   // SAFETY: This is an architecturally defined way to invalidate caches
-  unsafe { asm!("invld") } 
+  unsafe { asm!("invld"); } 
+
+  // Better set this ptr to 4!
+  *ptr = 4;
 }
 ```
 
 A more complex example, is writing the CR3 register on x64 systems.
 
-```rust
+```rust,no_run
+# use std::arch::asm;
 /// Installs a page table.
 ///
 /// # Parameters
@@ -179,7 +183,7 @@ if this did not point to a page table the hardware understands or if the mapping
 hardware safety aspect. When viewed through this lens, the function can be a safe function, because all a caller could
 do is provide the following safety comment:
 
-```rust
+```rust,ignore
 // SAFETY: Hopefully this works!
 unsafe { install_page_table(cr3);}
 ```
@@ -189,10 +193,36 @@ this u64 provided, when treated as a pointer, will live for the lifetime of the 
 gets reallocated, we have just violated both software and hardware safety. As such, the function is marked unsafe and
 should be given the appropriate function comment describing the caller expectations.
 
+There are other assumptions the software can validate to make this more likely to succeed, even if it cannot predict
+exactly how the hardware will process this write.
+
 This function could also be written to have a safe abstraction over it, e.g.:
 
-```rust
+```rust,no_run
+# use std::arch::asm;
 type PageTableRoot = u64;
+
+enum PageTableInstallError {
+  /// The CR3 address is not 4KB-aligned
+  NotAligned,
+  /// The CR3 address exceeds the maximum physical address width
+  ExceedsMaxPhysAddr,
+  /// Reserved bits are set in CR3
+  ReservedBitsSet,
+  /// PCIDE bit set but CR4.PCIDE is not enabled
+  PcidWithoutCr4Support,
+}
+
+struct PageTableOptions {
+  /// Page-level cache disable (CR3 bit 4)
+  pub cache_disable: bool,
+  /// Page-level write-through (CR3 bit 3)
+  pub write_through: bool,
+  /// Preserve PCID TLB entries (CR3 bit 63, requires CR4.PCIDE=1)
+  pub preserve_pcid: bool,
+  /// PCID value (CR3 bits 11:0 when CR4.PCIDE=1)
+  pub pcid: Option<u16>,
+}
 
 /// Installs a page table.
 ///
@@ -209,10 +239,27 @@ unsafe fn write_cr3(cr3: u64) {
   }
 }
 
-fn install_page_table(cr3: &'static PageTableRoot) {
-   // SAFETY: We have ensured the lifetime is static, that the reference is valid, and we have done our best to ensure
-   // this is a valid page table
-   unsafe { write_cr3(*cr3 as u64) }
+# fn not_aligned(cr3: &PageTableRoot) -> bool { false }
+# fn exceeds_maximum_addr_width (cr3: &PageTableRoot) -> bool { false }
+
+fn install_page_table(cr3: &'static PageTableRoot, options: PageTableOptions) -> Result<(), PageTableInstallError> {
+  if not_aligned(cr3) {
+    return Err(PageTableInstallError::NotAligned);
+  }
+
+  if exceeds_maximum_addr_width(cr3) {
+    return Err(PageTableInstallError::ExceedsMaxPhysAddr);
+  }
+
+  // additional checks such as if any reserved bits are set, if CR4 is configured correctly, etc.
+
+  // Parse options, apply them in CR4, etc, ensure they are self consistent
+
+  // SAFETY: We have ensured the lifetime is static, that the reference is valid, and we have done our best to ensure
+  // this is a valid page table
+  unsafe { write_cr3(*cr3 as u64); }
+
+  Ok(())
 }
 ```
 
@@ -221,7 +268,8 @@ declared as safe in Patina because writing a system register in and of itself is
 the hardware safety. For a given case, say the CR3 case, we may create an unsafe abstraction on top that declares this
 particular register write has memory safety implications that the caller must guarantee.
 
-```rust
+```rust,no_run
+# use std::arch::asm;
 macro_rules! write_sysreg {
   ($dest:ident, $value:expr) => {
     // SAFETY: We have a compile time guarantee that this is a valid register to write to
@@ -233,6 +281,28 @@ macro_rules! write_sysreg {
 }
 
 type PageTableRoot = u64;
+
+enum PageTableInstallError {
+    /// The CR3 address is not 4KB-aligned
+    NotAligned,
+    /// The CR3 address exceeds the maximum physical address width
+    ExceedsMaxPhysAddr,
+    /// Reserved bits are set in CR3
+    ReservedBitsSet,
+    /// PCIDE bit set but CR4.PCIDE is not enabled
+    PcidWithoutCr4Support,
+}
+
+struct PageTableOptions {
+  /// Page-level cache disable (CR3 bit 4)
+  pub cache_disable: bool,
+  /// Page-level write-through (CR3 bit 3)
+  pub write_through: bool,
+  /// Preserve PCID TLB entries (CR3 bit 63, requires CR4.PCIDE=1)
+  pub preserve_pcid: bool,
+  /// PCID value (CR3 bits 11:0 when CR4.PCIDE=1)
+  pub pcid: Option<u16>,
+}
 
 /// Installs a page table.
 ///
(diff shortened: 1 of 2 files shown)


docs/src/dev/principles/unsafe.md
```diff
@@ -0,0 +1,170 @@
+# Unsafe Guidance
```
cfernald, line 1: mentioned offline, but can you link this and all of the principles to the summary page so they are discoverable

os-d: Added this one, but the others are already in the summary page

image

cfernald: so it is, just not under the dev section as I was looking. thanks!

docs/src/dev/principles/unsafe.md
```diff
@@ -0,0 +1,170 @@
...
+principles and applications.
+
+This document expects the reader has a general understanding of the `unsafe` keyword in Rust and how the compiler
+enforces it. The above documentation and the
+[UEFI Memory Safety Case Studies](../../background/uefi_memory_safety_case_studies.md) provide good starting points.
+
+## Unsafe Philosophy
+
+As a general principle, Patina splits the idea of safety into two categories: software safety and hardware safety.
+
```
cfernald, line 18: Can you add a separator row to the table? It messes with at least basic markdown renderers to not have it.

docs/src/dev/principles/unsafe.md
```diff
@@ -0,0 +1,170 @@
...
+  // SAFETY: This is an architecturally defined operation to write the page table root.
+  unsafe {
+      asm!("mov cr3, {0}", in(reg) cr3);
+  }
+
+  ()
+}
+```
+
+There is a precondition here that `cr3` must be a u64 that is the address of a valid page table. This operation is
```
cfernald, line 175: Generally, I agree but want to call attention to this line as to me it's the crux of this document and I want to make sure there is alignment going forward. @makubacki @Javagedes for thoughts.

I mulled over if we should treat pointer-based registers differently, because of the basic requirements of things like memory lifetime, but I don't think it buys us much to try to make that distinction.

docs/src/dev/principles/unsafe.md
```diff
@@ -0,0 +1,170 @@
...
+|Compiler Enforced|✅|❌|
+|Preconditions|✅|✅|
+|Postconditions|✅|✅|
+|Invariants|✅|❌|
+
+Software safety is set of things the compiler can verify and/or the programmer can verify in a pure software
+environment. Hardware safety is the set of things that the compiler cannot verify, the programmer may be able to verify,
+and that interacts with hardware in a direct way.
+
+The reason that Patina splits the idea of safety into these two categories is because it dictates our theory of
```
apop5: This is a little clunky. Maybe:

Patina splits the Software safety and Hardware safety to delineate usage of unsafe: Rust guidance can be applied for software safety, but hardware safety requires more user knowledge and cannot be achieved through software alone.

docs/src/dev/principles/unsafe.md
```diff
@@ -0,0 +1,170 @@
...
+usage of the unsafe keyword. In software safety, we follow closely to the Rust guidance. However, hardware safety does
+not always make sense to follow the guidance.
+
+Below are the breakdowns of the different unsafe usages and how we view software vs hardware safety here.
+
+### Unsafe Code Blocks
+
+Unsafe code blocks are not given much discussion here because the compiler will enforce what operations require the
+unsafe block and cargo-clippy will enforce that unsafe code blocks are only used in cases where the compiler enforces
+it. As such, the general guidance applies here: write as few unsafe operations as possible, constrain them underneath
```
apop5:
```suggestion
it. As general guidance: write as few unsafe operations as possible, constrain them underneath

```

cfernald (approved):

docs/src/dev/principles/unsafe.md
os-d: Same. As I mentioned offline, this is a somewhat nuanced distinction and I'm fine if folks disagree or don't want a general statement on this, but let the developer decide in each situation.

docs/src/dev/principles/unsafe.md
os-d: Yes, thanks for the reminder, I meant to view this rendered and fix it up and I have not.

docs/src/dev/principles/unsafe.md
os-d: Updated with mostly your change.

docs/src/dev/principles/unsafe.md
makubacki, line 1: Please give me a few mins to read through. Doing so now.

docs/src/dev/principles/unsafe.md
```diff
@@ -0,0 +1,170 @@
...
+As a general principle, Patina splits the idea of safety into two categories: software safety and hardware safety.
+
+|                   | Software Safety| Hardware Safety|
+|-------------------|----------------|----------------|
+| Compiler Enforced | ✅            | ❌             |
+| Preconditions     | ✅            | ✅             |
+| Postconditions    | ✅            | ✅             |
+| Invariants        | ✅            | ❌             |
+
+Software safety is set of things the compiler can verify and/or the programmer can verify in a pure software
```
makubacki:
```suggestion
Software safety is the set of things the compiler can verify and/or the programmer can verify in a pure software
```

docs/src/dev/principles/unsafe.md
```diff
@@ -0,0 +1,170 @@
...
+  unsafe {
+      asm!("mov cr3, {0}", in(reg) cr3);
+  }
+
+  ()
+}
+```
+
+There is a precondition here that `cr3` must be a u64 that is the address of a valid page table. This operation is
+unsafe because all assembly is unsafe in Rust. However, in Patina, this need not create an unsafe function.
```
makubacki: I find this a little more natural, up to you.

```suggestion
unsafe because all assembly is unsafe in Rust. However, in Patina, this does not need to
reside in an unsafe function.
```

docs/src/dev/principles/unsafe.md
makubacki: Doesn't this "precondition" rely on caller safety compliance?

`There is a precondition here that `cr3` must be a u64 that is the address of a valid page table.`

The safety comment (`SAFETY: This is an architecturally defined operation to write the page table root.`) is valid if that is true but that is a caller responsibility in the function contract.

makubacki: I understand this the main argument you are making is that this doesn't need to be marked unsafe because it will be given to the hardware to act on, but why exactly would the caller not need to acknowledge the safety expectation for that to succeed?

os-d: This is the nuanced part :). And maybe writing cr3 here is the wrong example. I can choose a more clear cut example like `invd`. Don't mess that up, but there's nothing you can do from a functional safety contract about that. It just has to be called at the right time.

That aside, the general point here was that Rust defines the unsafe keyword as protecting memory safety

image

Well, this is a u64. There's no memory safety about it (which is why Chris's point about register writes that accept a pointer, not a value). It doesn't even need to be mapped, so it isn't "dereferenced" in the way that we talk about with pointers. So, my argument (which I don't feel super strongly about and I'm okay to abandon this if folks don't like it) is that the unsafe keyword as defined by Rust is not very valuable in this situation. It doesn't protect memory safety, it protects hardware safety, which is a much more nebulous concept and using the unsafe keyword on a function isn't going to help us not shoot ourselves in the foot and just adds more usage of the unsafe keyword without value.

joschock: To be pedantic in the cr3 instance: swapping in a new page table (badly or maliciously constructed) could create a memory alias which would break memory safety. Do we care about the nuance that certain hardware behavior could directly impact software safety? An MMIO write to a UART TX buffer register probably doesn't have implications for the rust memory model; changing where addresses go does, as would e.g. specifying a DMA target that scribbles over the heap.

makubacki: I'll start by saying this is nuanced, but I'd like to share my perspective for why I think the function should be `unsafe`.

---

A key reason we are writing firmware from the ground up in Rust is to write code that has correct behavior. And it is important enough that we're willing to transition to an entirely new language, ecosystem, set of tools, and reduce size and performance to make it work.

Various mechanisms are in place to achieve that goal:

- Rust language capabilities: The ownership system, lifetimes, borrow checker, unsafe keyword, etc.
- Project-Imposed Firmware Requirements: No Traditional SMM, no apriori dispatch, memory protections required
- More unit tests and higher quality units than the C code
- Static analysis tools like Miri being run against the code
- And so on

Where incorrect behavior in firmware ranges from bricking the system to eroding security assumptions and other functional but impaired system states.

While we can write "safe abstractions", at some point, we are ultimately responsible for fulfilling the role of software interacting with a direct hardware interface. As software that strives to be correct, we have to determine how we're going to uphold the software-hardware contract with correctness. The hardware resources will vary (here CR3), but ultimately the Software Developer Manual exists to define a contract that allows the hardware to be programmed correctly.

In this example:

```rust
fn install_page_table(cr3: u64) -> () {
  // SAFETY: This is an architecturally defined operation to write the page table root.
  unsafe {
      asm!("mov cr3, {0}", in(reg) cr3);
  }

  ()
}
```

`install_page_table()` has assumed the responsibility of being the software that sets the physical address in CR3 to a table of PDPTEs.

If we accept the spirit of unsafe, in particular an [`unsafe fn`](https://doc.rust-lang.org/std/keyword.unsafe.html#the-different-meanings-of-unsafe) in this case:

image

We could consider whether calling `install_page_table()` is a situation in which the programmer calling the function should explicitly acknowledge (because Rust's language features cannot cover this) that "I have read a contract on how to provide a **correct** u64 value, now use it".

Here, I lean toward that they should acknowledge that:

- The function is called at CPL=0 (not from user mode)
- The `u64` value is a valid physical address in the system that is present so a fault won't occur on accessing it
- The `u64` value follows the software architecture specification requirements for the CR3 value

Because this results in a more likely outcome for "correct" behavior. A (safe) abstraction can be written that calls this function in an unsafe block and that unsafe block can explain how the call upholds the requirements even if the answer is "I don't uphold some right now". In the end, humans debugging can now identify that call point as a potential issue and have more context for what was expected versus delivered in that contract.

makubacki: That said, whatever stance is chosen, the guidance in this document will be helpful to ensure consistency. If you could, please elaborate further on different considerations to help with that decision process.

os-d: We had an offline discussion and I am going to make some changes to the doc, in particular around the CR3 example.

docs/src/dev/principles/unsafe.md
```diff
@@ -0,0 +1,243 @@
...
+  // Better set this ptr to 4 first!
+  // SAFETY: See function comment, caller assumes responsibility
+  unsafe {ptr.write(4);}
+
+  // SAFETY: This is an architecturally defined way to invalidate caches
+  unsafe { asm!("invld") } 
+}
+```
+
+However, the function could be made safe again by validating assumptions:
```
cfernald: In the following example though, it's not declared safe (and shouldn't be because its still not safe to dereference and arbitrary NonNull). Perhaps this was meant to be changed?

docs/src/dev/principles/unsafe.md
```diff
@@ -0,0 +1,243 @@
...
+
+  // SAFETY: This is an architecturally defined way to invalidate caches
+  unsafe { asm!("invld") } 
+}
+```
+
+A more complex example, is writing the CR3 register on x64 systems.
+
+```rust
+// SAFETY: Caller must ensure `cr3` points to a valid page table
```
cfernald: nit: this should probably follow the normal safety comment for a function i.e.

///
/// ...
///
/// ## Safety
// expectations...

docs/src/dev/principles/unsafe.md
cfernald: Resolving this as I think its been addressed.

docs/src/dev/principles/unsafe.md
```diff
@@ -0,0 +1,243 @@
...
+// SAFETY: Caller must ensure `cr3` points to a valid page table and that that pointer lives for the lifetime of
+// being written in CR3.
+unsafe fn write_cr3(cr3: u64) {
+  // SAFETY: This is an architecturally defined operation to write the page table root.
+  unsafe {
+      asm!("mov cr3, {0}", in(reg) cr3)
+  }
+}
+
+fn install_page_table(cr3: &'static u64) {
```
cfernald: Consider using a made-up PageTable type instead of u64, just for clarity.

docs/src/dev/principles/unsafe.md
```diff
@@ -0,0 +1,243 @@
...
+  ($dest:ident, $value:expr) => {
+    // SAFETY: We have a compile time guarantee that this is a valid register to write to
+    unsafe {
+      asm!(concat!("mov ", stringify!($dest), ", {value:x}"),
+      value = in(reg) $value)
+    }
+  }
+}
+
+// SAFETY: Caller must ensure `cr3` points to a valid page table and that that pointer lives for the lifetime of
```
cfernald: ditto with using the traditional function comment style here

docs/src/dev/principles/unsafe.md
os-d: Let me change this.

docs/src/dev/principles/unsafe.md
os-d: I originally did, then I went back to this, but I think that works too

docs/src/dev/principles/unsafe.md
```diff
@@ -0,0 +1,270 @@
...
+  // Better set this ptr to 4 first!
+  // SAFETY: See function comment, caller assumes responsibility
+  unsafe {ptr.write(4);}
+
+  // SAFETY: This is an architecturally defined way to invalidate caches
+  unsafe { asm!("invld") } 
+}
+```
+
+However, the function could be made safe again by validating assumptions:
```
makubacki: > However, the function could be made safe again by validating assumptions:

Isn't this safe not because the function itself is validating assumptions but because `Box<u8>` guarantees that the pointer is not null, properly aligned (in this case only 1 byte alignment), and owned by the caller? In this case there is a transfer of ownership to `invalidate_caches_and_set_ptr_to_4` so it now has exclusive ownership of the box which was not the case for the raw pointer.

docs/src/dev/principles/unsafe.md
```diff
@@ -0,0 +1,270 @@
...
+Whenever possible, programmers should write safe functions that validate all preconditions, postconditions, and
+invariants. Of course, this is not always possible, in which case unsafe functions are called for.
+
+For hardware safety, we don't have invariants, necessarily. For example, in writing a system register, there is no
+invariant that must hold true for that operation. There may be preconditions or postconditions, but not an invariant.
+
+One simple example of hardware safety is the invalidate cache instruction on x86_64 that invalidates caches without
+writing them back. This operation can be very dangerous if not called at a proper time, but nothing about the software
+can change that; there is no validation that can be done to indicate this is a safe time to call this. In Patina, this
+is hardware safety and does not warrant an unsafe function.
```
makubacki: > there is no validation that can be done to indicate this is a safe time to call this. In Patina, this is hardware safety and does not warrant an unsafe function.

I would hold back from such a strong statement. There are cases like executing in Cache-As-RAM that you can check the NEM state might choose to prevent executing `invd` to prevent flushing all of the verified code content and NEM data in the CPU cache. That doesn't occur in DXE, but I also don't think that distinction matters here.

The description doesn't need to be super pedantic but I'd avoid saying "there is no validation that can be done".

docs/src/dev/principles/unsafe.md
os-d: Yeah, I was treating this as the function choice to define its interface that way, forcing safety, but agreed that it is Box that is validating the assumptions. I can either update the statement or I can move this back to a pointer and have the function validate its assumptions. Do you have a preference?

docs/src/dev/principles/unsafe.md
```diff
@@ -0,0 +1,270 @@
...
+    unsafe {
+      asm!(concat!("mov ", stringify!($dest), ", {value:x}"),
+      value = in(reg) $value)
+    }
+  }
+}
+
+type PageTableRoot = u64;
+
+/// Installs a page table.
```
makubacki, line 307: The lifetime of the page table is one aspect, but it seems like there is more that can be done. For example:

```rust
pub unsafe fn install_page_table(
    physical_addr: u64,
    options: Option<PageTableOptions>,
) -> Result<(), PageTableInstallError>
```

Where `PageTableInstallError` could be:

```rust
pub enum PageTableInstallError {
    /// The CR3 address is not 4KB-aligned
    NotAligned,
    /// The CR3 address exceeds the maximum physical address width
    ExceedsMaxPhysAddr,
    /// Reserved bits are set in CR3
    ReservedBitsSet,
    /// PCIDE bit set but CR4.PCIDE is not enabled
    PcidWithoutCr4Support,
}
```

`PageTableOptions` could be something like:

```rust
pub struct PageTableOptions {
    /// Page-level cache disable (CR3 bit 4)
    pub cache_disable: bool,
    /// Page-level write-through (CR3 bit 3)
    pub write_through: bool,
    /// Preserve PCID TLB entries (CR3 bit 63, requires CR4.PCIDE=1)
    pub preserve_pcid: bool,
    /// PCID value (CR3 bits 11:0 when CR4.PCIDE=1)
    pub pcid: Option<u16>,
}
```

So that the `install_page_table()` function can verify basic expectations defined in the Intel Software Developer's Manual. If an abstraction example is being given, I think some more examples of how to ensure a valid page table is being written should be provided.

os-d: Sure, I can add more.

docs/src/dev/principles/unsafe.md
makubacki: I think the box example is fine if the description is updated. The function name is `invalidate_caches_and_set_ptr_to_4()`. But the implementation assigns 4 and invalidates. I think moving the assignment after the invalidate makes sense and would match the function name's order of operations.

docs/src/dev/principles/unsafe.md
os-d: I added more examples, without being thoroughly verbose on the SDM expectations, since that is not the point here, but just that there are additional validations we can make.

docs/src/dev/principles/unsafe.md
os-d: I removed the line.

docs/src/dev/principles/unsafe.md
os-d: Updated.

os-d:
@makubacki I addressed your feedback if you want to re-review

makubacki:
> @makubacki I addressed your feedback if you want to re-review

Thanks

PR #954, opened 2025-10-24
patina_debugger: Move TLB Flush to Monitor Command

os-d:
## Description

Currently, the patina_debugger always issues a TLB flush when exiting the debugger's active session. This can be distracting when debugging issues with incorrect TLB entries as using the debugger can make the problem go away. The debugger does not need to do a TLB flush, when it edits the page tables, the paging crate does all the required TLB operations.

The reason the TLB flush was added to the debugger was to handle the case where a developer is manually editing page tables via the debugger client. This is a rare case, but does require a TLB flush.

To handle this case in a less intrusive way, and to support other uses cases where flushing the TLB is nice, this commit drops always flushing the TLB from the debugger and instead adds a monitor command to flush the TLB.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested running the monitor command and setting breakpoints (to ensure instruction cache flushing was still occuring) on Q35 and SBSA.

## Integration Instructions

`!uefiext.monitor arch flush_tlb` may be run from WinDbg (or other monitor commands from other UIs) to flush the TLB on the DUT.

### DIFF
--- a/core/patina_debugger/src/arch/x64.rs
+++ b/core/patina_debugger/src/arch/x64.rs
@@ -74,13 +74,12 @@ impl DebuggerArch for X64Arch {
             }
         }
 
-        // Always invalidate the TLB in case mappings were changed. The instruction
-        // cache doesn't need to be flushed since it should already be invalidated
-        // by the write according to the Intel SDM Vol 3 section 11.6. The CR3
-        // write is also serializing so no barriers are needed.
-        // SAFETY: This is an architecturally defined operation to flush the TLB.
+        // Always clear the instruction cache since the debugger may have altered instructions. This has the side effect
+        // of also clearing the data cache because on x64 this cannot be done separately.
+        // SAFETY: This is an architecturally defined operation to flush the caches and wbinvd is a serializing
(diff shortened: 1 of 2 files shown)


core/patina_debugger/src/arch/x64.rs
```diff
@@ -74,13 +74,11 @@ impl DebuggerArch for X64Arch {
             }
         }
 
-        // Always invalidate the TLB in case mappings were changed. The instruction
-        // cache doesn't need to be flushed since it should already be invalidated
-        // by the write according to the Intel SDM Vol 3 section 11.6. The CR3
-        // write is also serializing so no barriers are needed.
-        // SAFETY: This is an architecturally defined operation to flush the TLB.
+        // Always clear the instruction cache since the debugger may have altered instructions.
```
cfernald: I would clarify this is also flushing the data cache.

cfernald (approved):

PR #956, opened 2025-10-24
sdk/guid: Add add const string support

Javagedes:
## Description

Add const support for generating a guid from a well formatted string. This updates 	ry_from_string to be const, and adds a from_string method that will panic on an unformatted string.

Additionally removes the manual implementation of PartialOrd in favor of the derive implementation since we manually implement `Ord`, which is the proper way to do it per `PartialOrd` documentation.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [x] Includes documentation?

## How This Was Tested

CI continues to pass, New CI passes

## Integration Instructions

Users may begin creating a const guid (Both BinaryGuid and Guid) from strings:

```rust
const MY_GUID: BinaryGuid = BinaryGuid::from_string("550E8400-E29B-41D4-A716-446655440000");
```

### DIFF
--- a/sdk/patina/src/base/guid.rs
+++ b/sdk/patina/src/base/guid.rs
@@ -603,7 +603,7 @@ const fn guid_from_str(s: &str) -> core::result::Result<r_efi::efi::Guid, GuidEr
         return Err(GuidError::InvalidLength { expected: EXPECTED_HEX_CHARS, actual: char_count });
     }
 
-    let let_low = parse_hex!(chars, 0, 8, u32);
+    let time_low = parse_hex!(chars, 0, 8, u32);
     let time_mid = parse_hex!(chars, 8, 4, u16);
     let time_hi_and_version = parse_hex!(chars, 12, 4, u16);
     let clk_seq_hi_res = parse_hex!(chars, 16, 2, u8);


sdk/patina/src/base/guid.rs
```diff
@@ -608,6 +560,80 @@ impl<'a> TryFrom<&'a str> for OwnedGuid {
...
+        chars[char_count] = bytes[i] as char;
+        char_count += 1;
+        i += 1;
+    }
+
+    if char_count != EXPECTED_HEX_CHARS {
+        return Err(GuidError::InvalidLength { expected: EXPECTED_HEX_CHARS, actual: char_count });
+    }
+
+    let let_low = parse_hex!(chars, 0, 8, u32);
```
makubacki:
```suggestion
    let time_low = parse_hex!(chars, 0, 8, u32);
```

sdk/patina/src/base/guid.rs
```diff
@@ -608,6 +560,80 @@ impl<'a> TryFrom<&'a str> for OwnedGuid {
...
+    let node = [
+        parse_hex!(chars, 20, 2, u8),
+        parse_hex!(chars, 22, 2, u8),
+        parse_hex!(chars, 24, 2, u8),
+        parse_hex!(chars, 26, 2, u8),
+        parse_hex!(chars, 28, 2, u8),
+        parse_hex!(chars, 30, 2, u8),
+    ];
+
+    Ok(r_efi::efi::Guid::from_fields(let_low, time_mid, time_hi_and_version, clk_seq_hi_res, clk_seq_low, &node))
```
makubacki:
```suggestion
    Ok(r_efi::efi::Guid::from_fields(time_low, time_mid, time_hi_and_version, clk_seq_hi_res, clk_seq_low, &node))
```

sdk/patina/src/base/guid.rs
```diff
@@ -608,6 +560,80 @@ impl<'a> TryFrom<&'a str> for OwnedGuid {
...
+}
+
+/// Converts a single hex character (represented as a u8) to it's corresponding u8 value
```
makubacki:
```suggestion
/// Converts a single hex character (represented as a char) to its corresponding u8 value
```

sdk/patina/src/base/guid.rs
```diff
@@ -216,6 +216,28 @@ impl BinaryGuid {
...
+    /// Create a new Guid from a string representation
+    pub const fn try_from_string(s: &str) -> core::result::Result<BinaryGuid, GuidError> {
+        match guid_from_str(s) {
+            Ok(g) => Ok(BinaryGuid(g)),
+            Err(e) => Err(e),
+        }
+    }
+
+    /// Create a new BinaryGuid from a string representation, panicking on invalid input.
+    pub const fn from_string(s: &str) -> BinaryGuid {
```
os-d, line 228: What is the value of `from_string` if it just explicitly panics? It just seems odd to have this API at all if under the hood it calls the safe version anyway, then says, oh well, panic. Why not just always have try_from_string?

Javagedes: It is for const definitions with a hardcoded string, that will never fail once you have it working once.

```rust
const MY_GUID: Guid = Guid::from_string("...");
```

This reduces boilerplate so that users don't have to write

```rust
const MY_GUID: Guid = {
    match Guid::try_from_string("...") {
      Ok(guid) => guid,
      Err(err) => panic!("");
    }
}
```

I should also note that `.unwrap()` is not const, so you could not do `const MY_GUID: Guid = Guid::try_from_string("...").unwrap()`

PR #969, opened 2025-10-27
[REBASE&FF] Switch to nightly toolchain

Javagedes:
## Description

This pull request switches the compiler to a nightly version dated one day past the 1.90 release date, rather then using the 1.90 release date directly with RUSTC_BOOTSTRAP=1.

This change pulls us in line with our own documentation and also makes using tools like vscode and it's plugins easier. One example of a pain point of RUSTC_BOOTSTRAP=1 is that the unit test UI in vscode would fail to compile host based unit tests unless your environment had RUSTC_BOOTSTRAP set to 1.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [x] Includes documentation?

## How This Was Tested

N/A

## Integration Instructions

N/A

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -6,7 +6,4 @@
     "--message-format=json"
   ],
   "rust-analyzer.cargo.targetDir": true,
-  "rust-analyzer.runnables.extraEnv": {
-    "RUSTC_BOOTSTRAP": "1"
-  },
 }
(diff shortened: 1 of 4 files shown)


makubacki:
I'm going to update workflows to use a GitHub app derived token. Tracked in https://github.com/OpenDevicePartnership/patina/issues/973.

For now, it's fine to merge this ignoring the "Rust Version Change Detection" failure.

PR #977, opened 2025-10-28
Cleanup Resc Desc HOB v1 Feature

os-d:
## Description

The PR to add a feature to include the Resc Desc HOB v1 was merged before all PR feedback was resolved. This resolves that feedback.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

N/A.

## Integration Instructions

N/A.

### DIFF
--- a/docs/src/integrate/dxe_core.md
+++ b/docs/src/integrate/dxe_core.md
@@ -493,26 +493,24 @@ For detailed memory allocation behavior, see [DXE Core Memory Management](../dxe
 
 ### 9.3 Resource Descriptor HOB Version Support
 
-Patina DXE Core supports two mutually exclusive formats for Resource Descriptor HOBs: V1 (legacy) and V2 (modern,
-with cache attributes). The version supported is selected at compile time using a Cargo feature flag:
+Patina DXE Core supports two mutually exclusive formats for Resource Descriptor HOBs: v1 (what EDK II uses) and v2
+(v1 + cache attribute information). The version supported is selected at compile time using a Cargo feature flag:
 
-- **Default (V2)**: Only V2 Resource Descriptor HOBs are processed. This is the default for modern platforms and
-  enables cache attribute support.
-- **Legacy (V1)**: If the `v1_resource_descriptor_support` feature is enabled, only V1 Resource Descriptor HOBs are
-  processed (for legacy platforms). V2 HOBs are ignored in this mode.
+- **Default (v2)**: Only v2 Resource Descriptor HOBs are processed. This is the default in Patina and is required for
+  ARM64 platforms.
+- **Compat (v1)**: If the `v1_resource_descriptor_support` feature is enabled, only V1 Resource Descriptor HOBs are
+  processed in order to support an easy transition from EDK II. v2 HOBs are ignored in this mode.
 
-The code paths for V1 and V2 are strictly separated at compile time for performance and maintainability. Shared GCD
-logic is reused for both modes.
+Platforms are required to move to Resource Descriptor HOB v2s to accurately describe their system memory with caching
+information.
 
-**How to enable V1 Resource Descriptor HOB support:**
-
-You can enable V1 Resource Descriptor HOB support by setting it as the default feature in your platform binary crate's `Cargo.toml`
-(e.g., `platform_patina_dxe_core/Cargo.toml`):
+In order to support easier adoption of Patina, a feature flag is provided to only process Resource Descriptor v1 HOBs.
+This is simply added for bring up purposes and is not intended to be a production feature. Resource Descriptor HOB v1
+support is enabled by setting the feature in the platform binary crate's `Cargo.toml`:
 
 ```toml
-[features]
-default = ["v1_resource_descriptor_support"]
-v1_resource_descriptor_support = []
+[dependencies]
+patina = {version = "x", features = ["v1_resource_descriptor_support"]}
 ```
 
 This will build and test the V1 code path by default, without needing to specify the feature flag on the command line.
--- a/patina_dxe_core/examples/std.rs
+++ b/patina_dxe_core/examples/std.rs
@@ -120,7 +120,7 @@ fn build_hob_list() -> *const c_void {
             },
             owner: efi::Guid::from_fields(0, 0, 0, 0, 0, &[0u8; 6]),
             resource_type: hob::EFI_RESOURCE_SYSTEM_MEMORY,
-            resource_attribute: hob::TESTED_MEMORY_ATTRIBUTES,
+            resource_attribute: hob::TESTED_MEMORY_ATTRIBUTES | hob::EFI_RESOURCE_ATTRIBUTE_WRITE_BACK_CACHEABLE,
             physical_start: mem_base + 0xE0000,
             resource_length: 0x10000,
         },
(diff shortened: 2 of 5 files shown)


docs/src/integrate/dxe_core.md
```diff
@@ -493,26 +493,22 @@ For detailed memory allocation behavior, see [DXE Core Memory Management](../dxe
 
 ### 9.3 Resource Descriptor HOB Version Support
 
-Patina DXE Core supports two mutually exclusive formats for Resource Descriptor HOBs: V1 (legacy) and V2 (modern,
-with cache attributes). The version supported is selected at compile time using a Cargo feature flag:
+Patina DXE Core supports two mutually exclusive formats for Resource Descriptor HOBs: v1 (what EDK II uses) and v2
+(cache attribute information). The version supported is selected at compile time using a Cargo feature flag:
```
Javagedes: Would it be more apt to say this is v1 + additional cache attribute information?

docs/src/integrate/dxe_core.md
```diff
@@ -493,26 +493,22 @@ For detailed memory allocation behavior, see [DXE Core Memory Management](../dxe
...
 
-- **Default (V2)**: Only V2 Resource Descriptor HOBs are processed. This is the default for modern platforms and
-  enables cache attribute support.
-- **Legacy (V1)**: If the `v1_resource_descriptor_support` feature is enabled, only V1 Resource Descriptor HOBs are
-  processed (for legacy platforms). V2 HOBs are ignored in this mode.
+- **Default (v2)**: Only v2 Resource Descriptor HOBs are processed. This is the default in Patina and is required for
```
Javagedes, line 499: Is this true? I did not pay attention to the first PR, but it would have made more sense to me that enabling the feature flag meant you additionally parsed V1 Resource Descriptor HOBs in addition to existing V2 support. That is also typically how feature flags work. They are additive in nature.

docs/src/integrate/dxe_core.md
```diff
@@ -493,26 +493,22 @@ For detailed memory allocation behavior, see [DXE Core Memory Management](../dxe
...
+  ARM64 platforms.
+- **Compat (v1)**: If the `v1_resource_descriptor_support` feature is enabled, only V1 Resource Descriptor HOBs are
+  processed in order to support an easy transition from EDK II. v2 HOBs are ignored in this mode.
 
-The code paths for V1 and V2 are strictly separated at compile time for performance and maintainability. Shared GCD
-logic is reused for both modes.
+It is recommended all platforms move to Resource Descriptor HOB v2s to accurately describe their system memory with
```
Javagedes: Feels like we need a stronger stance then "recommend" but I could be wrong.

patina_dxe_core/src/gcd.rs
```diff
@@ -330,52 +322,36 @@ pub(crate) fn activate_compatibility_mode() {
...
+/// Parse Resource Descriptor HOB v1
 ///
-/// Returns: Some((ResourceDescriptor, cache_attributes)) or None if not a V1 resource descriptor
+/// This function takes in a HOB and returns:
+/// - Some((Resource Descriptor, None))
+/// - None if not a v1 resource descriptor HOB
 #[cfg(feature = "v1_resource_descriptor_support")]
 fn parse_resource_descriptor_hob(hob: &Hob) -> Option<(hob::ResourceDescriptor, Option<u64>)> {
     match hob {
         Hob::ResourceDescriptor(v1_res_desc) => {
```
Javagedes, line 350: I'm probably way to late for this, but it really does feel like enabling this feature should have added support to also parse v1 hobs, not replaced v2 with v1.

If I'm a platform maintainer I'd probably want to convert 1 hob at a time and make sure it works, instead of all at the same time.

docs/src/integrate/dxe_core.md
os-d: This is intended to be either/or per community discussion (and my own personal opinion). It is expected that platforms will produce both v1 and v2 HOBs in order to be able to switch between Patina and edk2 to test differences. It is detrimental to try to parse both. Additional logic is needed to ensure the HOBs are the same, you aren't missing any, etc. This is all done in the readiness tool. Using resource descriptor v2 HOBs is a requirement to use Patina.

This feature flag is a recognition that it will take time for silicon code to move to v2 HOBs and provide an easier onboarding for x64 platforms (again, ARM64 this is non-negotiable, the information is required).

patina_dxe_core/src/gcd.rs
os-d: That creates a bad paradigm and extra code. See response above.

docs/src/integrate/dxe_core.md
Javagedes: I guess your second sentence still kind of leads into my point where a platform may want to start accurately describing their memory space, so they want to use v2 hobs - but their silicon code still produces v1 HOBs. Now they can't because they cannot convince silicon partners to do their part.

I would have called it an error if there was a v1 hob and a v2 hob describing the same memory space.

os-d: > I guess your second sentence still kind of leads into my point where a platform may want to start accurately describing their memory space, so they want to use v2 hobs - but their silicon code still produces v1 HOBs. Now they can't because they cannot convince silicon partners to do their part.
> 

Really, this is only an x86 feature. As noted, ARM64 is required to do this. For x86, the resource descriptor v2 HOB doesn't matter (yet) because the MTRRs are expected to be configured correctly by the time Patina is launched, which makes the caching info moot. So, it doesn't really matter if x86 platforms move forward (yet).

Silicon partners will need to be convinced to move, but it is also trivial to have a PEIM that reads v1 HOBs and MTRRs (or just makes it up) and convert them to v2. That has been current platforms approach. Silicon code can also be edited by platforms, typically.

The problem with processing both is: platforms will never move to v2 if they don't have to (eventually this feature will be removed) and resource descriptor v1 HOBs are very often inconsistent. So those HOBs need fixing up to be used in Patina anyway, regardless of silicon code or not. edk2 is lax and ignores overlapping HOBs. Patina does not, for a good reason.

> I would have called it an error if there was a v1 hob and a v2 hob describing the same memory space.

Well, then you can't switch back and forth between the C core and Patina, which is one of the main use cases of early adopters.

joschock: My two cents - if you have V1 hobs and ability to modify the code before patina launches making V2 hobs is pretty straight-forward. They are just the V1 hobs + "this region should be <whatever> cache attribute." The real case this enables is some hypothetical scenario where you have V1 hobs and some immutable pre-patina loader that you can't change but want to run patina on. I'm not aware of any environment like that.  I'm not too concerned about easing pain in a "transition" scenario where you are running half/half v1/v2, since I don't think there is much pain there to start with - assuming you know your memory map. If you don't, no amount of mixing V1/V2 is going to make that better ;)

makubacki: My suggestion was to allow V1 HOBs to be used on x86/X64 until:

1. V2 HOBs are officially released in the PI Specification
2. Major silicon vendors have shipped V2 HOBs

A V2 default is fine, but until at least (1) is done, I don't think it's reasonable to enforce V2.

patina_dxe_core/examples/std.rs
```diff
@@ -168,7 +170,9 @@ fn build_hob_list() -> *const c_void {
             },
             owner: efi::Guid::from_fields(0, 0, 0, 0, 0, &[0u8; 6]),
             resource_type: hob::EFI_RESOURCE_FIRMWARE_DEVICE,
-            resource_attribute: hob::EFI_RESOURCE_ATTRIBUTE_PRESENT | hob::EFI_RESOURCE_ATTRIBUTE_INITIALIZED,
+            resource_attribute: hob::EFI_RESOURCE_ATTRIBUTE_PRESENT
+                | hob::EFI_RESOURCE_ATTRIBUTE_INITIALIZED
+                | hob::EFI_RESOURCE_ATTRIBUTE_WRITE_BACK_CACHEABLE,
```
makubacki: If this is a memory-mapped firmware device, I think uncacheable would provide a better example.

PR #980, opened 2025-10-28
patina_smbios: add SMBIOS component

kat-perez:
## Description

Adds a new SMBIOS component for managing SMBIOS tables and records in UEFI environments with type-safe Rust APIs and EDKII compatibility.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [x] Includes documentation?

#### Core Functionality
- **SMBIOS 3.9 Support** - Implements SMBIOS 3.9 specification (configurable for 3.x versions)
- **Type-Safe Record API** - `add_record<T>()` with compile-time validation using Rust type system
- **Configuration Table Publication** - Publishes SMBIOS 3.x entry point and tables to UEFI Configuration Table
- **EDKII-Compatible Protocol** - C FFI layer implementing EDK2 SMBIOS protocol interface (Add, UpdateString, Remove, GetNext)
- **Thread-Safe Operations** - TPL_NOTIFY synchronization via `TplMutex` for timer-interrupt safety

(skip cargo release dry run)

## How This Was Tested

Unit test: `cargo test`

Integration tests: `cargo make q35`
- Boots QEMU with Q35 platform
- SMBIOS component initializes and publishes table
- C Protocol FFI tests execute and log results
- Use 'smbiosview' in UEFI Shell to verify published records

C Protocol FFI Layer tests:
  - `Add` - Raw byte-level record addition via C protocol
  - `UpdateString` - String modification via C protocol  
  - `Remove` - Record deletion via C protocol
  - `GetNext` - Record enumeration via C protocol
  - Error handling - Operations on invalid/removed handles
 
Sample:
```
        let protocol = unsafe { &*(protocol_ptr as *const SmbiosProtocol) };

        // Test 1: Add a record using the C protocol Add function
        log::info!("  Test 1: Protocol Add function...");
        let test_record = Self::create_test_type2_record();
        let mut handle: SmbiosHandle = 0;
        
        let status = (protocol.add)(
            protocol,
            core::ptr::null_mut(), // producer_handle
            &mut handle,
            test_record.as_ptr() as *const SmbiosTableHeader,
        );

```

## Integration Instructions

patina-dxe-core-qemu:
https://github.com/OpenDevicePartnership/patina-dxe-core-qemu/pull/59
Add the following to _start function in DXE core binary:
```
use patina_smbios::component::SmbiosProvider;

pub extern "efiapi" fn _start(physical_hob_list: *const c_void) -> ! {
    Core::default()
        .with_component(SmbiosProvider::new(3, 9))
        .with_component(q35_services::smbios_platform::Q35SmbiosPlatform::new())
        <other components>
}
```

patina-qemu:
https://github.com/OpenDevicePartnership/patina-qemu/pull/78

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,6 +44,7 @@ patina_macro = { version = "14.1.1", path = "sdk/patina_macro" }
 patina_mtrr = { version = "1.0.0" }
 patina_paging = { version = "10" }
 patina_performance = { version = "14.1.1", path = "components/patina_performance" }
+patina_smbios = { version = "14.1.1", path = "components/patina_smbios" }
 patina_stacktrace = { version = "14.1.1", path = "core/patina_stacktrace" }
 proc-macro2 = { version = "1" }
 quote = { version = "1" }
--- a/components/patina_samples/src/smbios_platform.rs
+++ b/components/patina_samples/src/smbios_platform.rs
@@ -1,0 +1,334 @@
+//! SMBIOS Platform Example Component
+//!
+//! This component demonstrates the recommended SMBIOS integration pattern:
+//! 1. Uses the type-safe `add_record<T>()` API for adding SMBIOS records
+//! 2. Platform component publishes the table after all records are added
+//! 3. Uses structured record types
+//! 4. Shows how to create custom vendor-specific SMBIOS records (Type 0x80-0xFF)
+//!
+//! ## Usage
+//!
+//! In your platform's DXE core initialization:
+//!
+//! ```ignore
+//! use patina_smbios::component::SmbiosProvider;
+//! use patina_samples::SmbiosExampleComponent;
+//!
+//! Core::default()
+//!     // Add SMBIOS provider component with version
+//!     .with_component(SmbiosProvider::new(3, 9))
+//!     // Add your platform SMBIOS component
+//!     .with_component(SmbiosExampleComponent::new())
+//!     // ... rest of your components
+//! ```
+//!
+//! ## License
+//!
+//! Copyright (c) Microsoft Corporation.
+//!
+//! SPDX-License-Identifier: Apache-2.0
+
+extern crate alloc;
+
+use alloc::{string::String, vec};
+use patina::{
+    component::{IntoComponent, service::Service},
+    error::Result,
+};
+use patina_macro::SmbiosRecord;
+use patina_smbios::{
+    service::{SMBIOS_HANDLE_PI_RESERVED, Smbios, SmbiosExt, SmbiosTableHeader},
+    smbios_record::{
+        SmbiosRecordStructure, Type0PlatformFirmwareInformation, Type1SystemInformation, Type2BaseboardInformation,
+        Type3SystemEnclosure, Type127EndOfTable,
+    },
+};
+
+/// Example custom vendor-specific OEM record (Type 0x80)
+///
+/// This struct contains a `string_pool: Vec<String>` which is Rust metadata and NOT
+/// part of the SMBIOS binary format. It should NOT have `#[repr(C, packed)]` because
+/// it's never directly cast to bytes - always use `to_bytes()` for serialization.
+#[derive(SmbiosRecord)]
+#[smbios(record_type = 0x80)]
+pub struct VendorOemRecord {
+    /// SMBIOS table header
+    pub header: SmbiosTableHeader,
+    /// Example 32-bit vendor-specific OEM field
+    pub oem_field: u32,
+    /// String pool containing vendor and platform data strings
+    #[string_pool]
+    pub string_pool: alloc::vec::Vec<String>,
+}
+
+/// Example Platform SMBIOS Component
+///
+/// Demonstrates a real-world platform component that adds SMBIOS records
+/// using the type-safe `add_record<T>()` API.
+///
+/// This example shows:
+/// - Type 0: BIOS/Firmware information
+/// - Type 1: System information with UUID and serial numbers
+/// - Type 2: Baseboard/motherboard information
+/// - Type 3: System chassis/enclosure information
+/// - Type 0x80: Custom vendor-specific record (OEM range)
+/// - Type 127: End-of-table marker
+/// - Publishing the complete table to UEFI Configuration Table
+#[derive(IntoComponent, Default)]
+pub struct SmbiosExampleComponent;
+
+impl SmbiosExampleComponent {
+    /// Create a new SMBIOS platform component
+    pub fn new() -> Self {
+        Self
+    }
+
+    /// Platform SMBIOS initialization entry point
+    ///
+    /// Called by the component framework during DXE phase initialization.
+    /// The framework automatically provides the SMBIOS service as a dependency.
+    fn entry_point(self, smbios: Service<dyn Smbios>) -> Result<()> {
+        log::info!("=== Example Platform SMBIOS Component ===");
+
+        // Verify SMBIOS version
+        let (major, minor) = smbios.version();
+        log::info!("SMBIOS Version: {}.{}", major, minor);
+
+        // Add platform-specific SMBIOS records
+        log::info!("Creating platform SMBIOS records...");
+
+        // Type 0: Platform firmware/BIOS information
+        if let Err(e) = Self::add_bios_information(&smbios) {
+            log::error!("Failed to add BIOS information: {:?}", e);
+            return Err(e);
+        }
+
+        // Type 1: System information
+        if let Err(e) = Self::add_system_information(&smbios) {
+            log::error!("Failed to add system information: {:?}", e);
+            return Err(e);
+        }
+
+        // Type 2: Baseboard/motherboard information
+        if let Err(e) = Self::add_baseboard_information(&smbios) {
+            log::error!("Failed to add baseboard information: {:?}", e);
+            return Err(e);
+        }
+
+        // Type 3: System chassis/enclosure information
+        if let Err(e) = Self::add_system_enclosure(&smbios) {
+            log::error!("Failed to add system enclosure: {:?}", e);
+            return Err(e);
+        }
+
+        // Type 0x80: Platform-specific vendor record
+        if let Err(e) = Self::add_vendor_oem_record(&smbios) {
+            log::error!("Failed to add vendor OEM record: {:?}", e);
+            return Err(e);
+        }
+
+        // Type 127: End-of-table marker
+        if let Err(e) = Self::add_end_of_table(&smbios) {
+            log::error!("Failed to add end-of-table marker: {:?}", e);
+            return Err(e);
+        }
+
+        log::info!("Platform SMBIOS records created successfully");
+
+        // Publish the SMBIOS table to the UEFI Configuration Table
+        log::info!("Publishing SMBIOS table to Configuration Table...");
+        match smbios.publish_table() {
+            Ok((table_addr, entry_point_addr)) => {
+                log::info!("SMBIOS table published successfully");
+                log::info!("  Entry Point: 0x{:X}", entry_point_addr);
+                log::info!("  Table Data: 0x{:X}", table_addr);
+                log::info!("Use 'smbiosview' in UEFI Shell to view records");
+            }
+            Err(e) => {
+                log::error!("Failed to publish SMBIOS table: {:?}", e);
+                // Continue even if publication fails - not critical for boot
+            }
+        }
+
+        log::info!("SMBIOS platform component initialized successfully");
+        Ok(())
+    }
+
+    /// Add Type 0 (Platform Firmware/BIOS Information)
+    ///
+    /// Demonstrates adding firmware version, vendor, and ROM size information.
+    fn add_bios_information(smbios: &Service<dyn Smbios>) -> Result<()> {
+        let bios_info = Type0PlatformFirmwareInformation {
+            header: SmbiosTableHeader::new(0, 0, SMBIOS_HANDLE_PI_RESERVED),
+            vendor: 1,
+            firmware_version: 2,
+            bios_starting_address_segment: 0xE800,
+            firmware_release_date: 3,
+            firmware_rom_size: 0xFF, // 16MB
+            characteristics: 0x08,   // PCI supported
+            characteristics_ext1: 0x03,
+            characteristics_ext2: 0x03,
+            system_bios_major_release: 1,
+            system_bios_minor_release: 0,
+            embedded_controller_major_release: 0xFF,
+            embedded_controller_minor_release: 0xFF,
+            extended_bios_rom_size: 0,
+            string_pool: vec![
+                String::from("Example Firmware Vendor"),
+                String::from("1.0.0"),
+                String::from("10/24/2025"),
+            ],
+        };
+
+        let handle = smbios.add_record(None, &bios_info).map_err(|e| {
+            log::error!("Failed to add BIOS info: {:?}", e);
+            patina::error::EfiError::DeviceError
+        })?;
+
+        log::info!("  Type 0 (BIOS Info) - Handle 0x{:04X}", handle);
+        Ok(())
+    }
+
+    /// Add Type 1 (System Information)
+    ///
+    /// Demonstrates adding system UUID, manufacturer, product name, and serial number.
+    fn add_system_information(smbios: &Service<dyn Smbios>) -> Result<()> {
+        let system_info = Type1SystemInformation {
+            header: SmbiosTableHeader::new(1, 0, SMBIOS_HANDLE_PI_RESERVED),
+            manufacturer: 1,
+            product_name: 2,
+            version: 3,
+            serial_number: 4,
+            uuid: [0x12, 0x34, 0x56, 0x78, 0x9A, 0xBC, 0xDE, 0xF0, 0x12, 0x34, 0x56, 0x78, 0x9A, 0xBC, 0xDE, 0xF0],
+            wake_up_type: 0x06, // Power switch
+            sku_number: 5,
+            family: 6,
+            string_pool: vec![
+                String::from("Example Corporation"),
+                String::from("Example Platform"),
+                String::from("1.0"),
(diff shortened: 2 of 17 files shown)


kat-perez:
this is a remake from https://github.com/OpenDevicePartnership/patina/pull/906

components/patina_smbios/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[package]
+name = "patina_smbios"
+version.workspace = true
+license.workspace = true
+edition.workspace = true
+rust-version.workspace = true
+repository.workspace = true
+description = "System Management BIOS (SMBIOS) support for UEFI components."
```
makubacki: For clarity.

```suggestion
description = "System Management BIOS (SMBIOS) support for Patina UEFI components."
```

components/patina_samples/src/smbios_platform.rs
```diff
@@ -0,0 +1,346 @@
...
+};
+use patina_smbios::{
+    service::{SMBIOS_HANDLE_PI_RESERVED, Smbios, SmbiosTableHeader},
+    smbios_record::{
+        SmbiosRecordStructure, Type0PlatformFirmwareInformation, Type1SystemInformation, Type2BaseboardInformation,
+        Type3SystemEnclosure, Type127EndOfTable,
+    },
+};
+
+// Import the SmbiosRecord derive macro
```
makubacki: I don't see this needing to be called much more than any other import.


```suggestion
```

components/patina_smbios/src/manager/core.rs
```diff
@@ -0,0 +1,1185 @@
+//! Core SMBIOS manager implementation
```
makubacki, line 1: Please add a complete module doc comment including the copyright section to each new file.

components/patina_smbios/src/manager/core.rs
```diff
@@ -0,0 +1,1185 @@
...
+            return Ok(Vec::new());
+        }
+
+        // Remove the final double-null terminator and split by null bytes
+        let data_without_terminator = &string_pool_area[..len - 2];
+
+        // Split by null bytes and convert to Strings
+        let strings: Result<Vec<String>, _> = data_without_terminator
+            .split(|&b| b == 0)
+            .map(|bytes| core::str::from_utf8(bytes).map(String::from).map_err(|_| SmbiosError::MalformedRecordHeader))
```
makubacki: Does this need to return `Result<Vec<String>, SmbiosError>` instead of `Result<Vec<&str>, SmbiosError>`? It seems like in many cases that would be a string literal and my understanding is that the heap allocation for `String` might not always be necessary.

components/patina_smbios/src/manager/core.rs
```diff
@@ -0,0 +1,1185 @@
...
+
+        // Step 5: Validate string pool format and count strings
+        let string_count = Self::validate_and_count_strings(string_pool_area)?;
+
+        // If all validation passes, allocate handle and build record
+        let smbios_handle = self.allocate_handle()?;
+
+        let mut record_header =
+            SmbiosTableHeader { record_type: header.record_type, length: header.length, handle: smbios_handle };
+        record_header.handle = smbios_handle;
```
makubacki: Is `record_header.handle = smbios_handle;` necessary if the struct is initialized with `handle: smbios_handle` above?


```suggestion
```

components/patina_smbios/src/manager/core.rs
kat-perez: done

components/patina_smbios/src/component.rs
```diff
@@ -0,0 +1,141 @@
...
+        // SAFETY: This is safe because boot_services remains valid for the entire lifetime
+        // of the system. We leak it intentionally to get a 'static reference.
+        let boot_services_static: &'static StandardBootServices = Box::leak(Box::new(boot_services));
+
+        // Wrap the manager in TplMutex for TPL-aware reentrancy protection at TPL_NOTIFY level
+        let manager_mutex = Box::leak(Box::new(TplMutex::new(boot_services_static, Tpl::NOTIFY, manager)));
+
+        // Install the C/EDKII protocol - this installs a global reference for C compatibility
+        if let Err(e) =
+            install_smbios_protocol(cfg.major_version, cfg.minor_version, manager_mutex, boot_services_static)
```
Javagedes: If you have the smbios protocol consume the service instead of the manager, then you can get away with one less leaked box. Also, you don't need to store a static reference to StandardBootServices. You can clone it and hold on to it. It's just a wrapper around the pointer. Eventually the TplMutex will consume a StandardBootServices instead of a reference to one. But that has not happened yet, so that will be a later change.

```rust
fn entry_point(
    self,
    storage: &mut Storage 
) -> Result<()> {
    let manager = SmbiosManager::new(self.major_version, cfg.minor_version).map_err(|e| {
        log::error!("Failed to create SMBIOS manager: {:?}", e);
        patina::error::EfiError::Unsupported
    })?;

    let bs = storage.boot_services();

    // John has a draft PR up to make the the TplMutex consume a StandardBootServices instead of take a
    // Reference, so this will be another removed leak eventually :)
    let boot_services_static: &'static StandardBootServices = Box::leak(Box::new(boot_services));
    
    // Don't need the leak here. Instead we will grab the service after its registered
    let manager_mutex = TplMutex::new(boot_services_static, Tpl::NOTIFY, manager);

    storage.add_service(SmBios {
        manager: manager_mutex,
        boot_services: boot_services_static.clone(), // Take your own clone instead of a reference
        major_version: self.major_version,
        minor_version: self.minor_version,
    });

    let service: Service<SmBios> = storage.get_service().expect("Service was registered immediately above");

    let let Err(e) =
        install_smbios_protocol(self.major_version, self.minor_version, service, /*Maybe you don't need to bs static anymore?*/) 
    {
        log::error!("Failed to install SMBIOS protocol: {:?}", e);
    }

    Ok(())
}
```

components/patina_smbios/src/manager/core.rs
kat-perez: ah, good catch will update to avoid unnecessary heap allocation.

components/patina_smbios/src/manager/core.rs
kat-perez: right, that's redundant. removing

components/patina_smbios/src/config.rs
```diff
@@ -0,0 +1,146 @@
...
+//! Configuration types for the SMBIOS component
+//!
+//! ## License
+//!
+//! Copyright (c) Microsoft Corporation.
+//!
+//! SPDX-License-Identifier: Apache-2.0
+//!
+
+use crate::error::SmbiosError;
```
Javagedes: Per offline conversatin with you and a bit with Michael, I think you can entirely get rid of this Config state. Just associate a `new` function with the component, where the component instantiator specifies their version.

components/patina_smbios/src/service.rs
```diff
@@ -0,0 +1,389 @@
...
+///     // Add structured records
+///     smbios.add_record(None, &bios_info)?;
+///     
+///     // Publish to configuration table
+///     smbios.publish_table()?;
+///     Ok(())
+/// }
+/// ```
+#[derive(patina::component::service::IntoService, Clone, Copy)]
+#[service(Smbios)]
```
Javagedes: I noticed you switched back to the service being the struct. If this was due to my questions in the last PR, then I think I may have just asked my questions / made my statements poorly. We definitely want to stick with the dyn interface if possible, as it allows for easy mocking.

I actually though what you did was implementing the generics in the `Service dyn <Trait>` was really cool. If it works, then I'll probably add that to the documentation for services, as a way to have generic functions for dyn trait object services. My only question regarding it was why you had to do the crazy pointer casting instead of something similar to this:

```rust

trait TraitName {
    fn add_from_bytes(&self, producer_handle: Option<r_efi::efi::Handle>, bytes: &[u8]);
}

impl Service<dyn TraitName> {
    pub fn add_record<T>(
        &self,
        producer_handle: Option<r_efi::efi::Handle>,
        record: &T,
    ) -> core::result::Result<SmbiosHandle, crate::error::SmbiosError>
    where
        T: crate::smbios_record::SmbiosRecordStructure,
    {
        let bytes = record.to_bytes();
        self.add_from_bytes(producer_handle, &bytes)
    }
}

```

components/patina_smbios/src/service.rs
```diff
@@ -0,0 +1,389 @@
...
+
+// Methods below are tested via integration (Q35 platform component)
+// They require StandardBootServices and TplMutex which aren't mockable in unit tests
+impl Smbios {
+    /// Gets the SMBIOS version information.
+    ///
+    /// # Returns
+    ///
+    /// A tuple of (major_version, minor_version).
+    #[cfg_attr(coverage_nightly, coverage(off))]
```
Javagedes: very small nit, and you can actually ignore this if you want, but you may be able to move the `#[cfg_attr(coverage_nightly, coverage(off))]` to the `impl Smbios` to ignore coverage for all implementations.

sdk/patina_macro/src/smbios_record_macro.rs
```diff
@@ -0,0 +1,282 @@
...
+
+            #validate_impl
+
+            #string_pool_impl
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
Javagedes, line 293: nit: please add some more tests regarding adding unexpected fields to the struct - maybe a field type that you don't expect, and make sure it gives a clear error message. If you do it right, you should actually be able to have the span of the error message "red squigly" the field that is wrong, and produce a clear error message stating this type cannot be turned into bytes or something like this.

components/patina_smbios/src/config.rs
kat-perez: config state is now completely removed and now just associated a new function with the component

components/patina_samples/src/smbios_platform.rs
```diff
@@ -0,0 +1,342 @@
...
+        })?;
+
+        log::info!("  Type 0x80 (Vendor OEM) - Handle 0x{:04X}", handle);
+        Ok(())
+    }
+
+    /// Add Type 127 (End of Table) marker
+    ///
+    /// Required marker to indicate the end of the SMBIOS table.
+    fn add_end_of_table(smbios: &Service<Smbios>) -> Result<()> {
```
rogurr: Not a bug, more of a question on implementation.  The original SMBIOS driver in Tianocore automatically creates the end of table marker for the caller.

[SmbiosDxe.c](https://github.com/tianocore/edk2/blob/master/MdeModulePkg/Universal/SmbiosDxe/SmbiosDxe.c#L1142)

This allows the caller to add tables in any order at any time during boot without having to know they need to provide an end table marker.  Do we want to follow the same usage model?

components/patina_smbios/src/service.rs
kat-perez: The main reason we can't use `Service<dyn Trait>` is because add_record<T>() has a generic type parameter and trait objects cannot have generic methods

If we made add_from_bytes non-generic, we can't do `add_record<T>()` and users must manually serialize:
```
trait Smbios {
    fn add_from_bytes(&self, producer_handle: Option<Handle>, bytes: &[u8]) -> Result<...>;
    // No add_record<T>() - users must manually serialize
}
```

So I'd prefer to the concrete type if we can

components/patina_smbios/src/service.rs
kat-perez: moved it up

components/patina_smbios/src/component.rs
kat-perez: I changed the boot_services_static Box::leak into an unsafe cast, which is better since there's no extra allocation, no more memory leak, it's only a lifetime cast (we're saying this lives for 'static and Storage guarantees it) and was able to get rid of the Box::leak on TplMutex

sdk/patina_macro/src/smbios_record_macro.rs
kat-perez: added some invalid tests to smbios_record_macro

components/patina_smbios/src/component.rs
kat-perez: Now we have something like this for entry_point

``` rust
    fn entry_point(self, storage: &mut Storage) -> Result<()> {
        let cfg = self.config;

        // Create the manager with configured version
        let manager = SmbiosManager::new(cfg.major_version, cfg.minor_version).map_err(|e| {
            log::error!("Failed to create SMBIOS manager: {:?}", e);
            patina::error::EfiError::Unsupported
        })?;

        // Get boot_services from storage - it already has 'static lifetime
        // We use an unsafe cast because storage.boot_services() returns a reference with
        // a shorter lifetime, but Storage guarantees boot_services lives for 'static
        let boot_services_static: &'static patina::boot_services::StandardBootServices =
            unsafe { &*(storage.boot_services() as *const _) };

        // Wrap the manager in TplMutex for TPL-aware reentrancy protection at TPL_NOTIFY level
        // The service registration will handle leaking this when the service struct is leaked
        let manager_mutex = TplMutex::new(boot_services_static, Tpl::NOTIFY, manager);

        // Register unified SMBIOS service with all operations
        // The service registration system leaks the entire service struct (including manager)
        // Version is stored directly to avoid TplMutex overhead for immutable data access
        storage.add_service(Smbios {
            manager: manager_mutex,
            boot_services: boot_services_static,
            major_version: cfg.major_version,
            minor_version: cfg.minor_version,
        });

```

components/patina_smbios/src/service.rs
Javagedes: Talked to Kat. Looks like my suggestion is not quite possible (To implement the generic function on `Service<dyn Smbios>` due to the orphan rule (`Service<T>` is defined in `patina` crate while the trait is in this new crate). I've suggested a solution similar to the ACPI solution - e.g. the wrapper type pattern where the wrapper implements the generic function.

I'm going to mark this as resolved with no action needed and create an RFC to decide on a standard way to deal with this issue (The need for generics for services).

Javagedes (approved):

components/patina_smbios/Cargo.toml
makubacki, line 1: I recently added a README to all of the crates that we publish from this repo. Can you please add one here as well?

components/patina_smbios/src/component.rs
makubacki, line 1: Does the commit "[patina_smbios: refactor service to own TplMutex and eliminate global static](https://github.com/OpenDevicePartnership/patina/pull/980/commits/86b368b8930a08a85b0c1531cc4c0cc561c2790c)" need to be singled out? Can it just be squashed into the original SMBIOS commit (asking since this is Rebase & FF). If it needs to remain separate, please add a commit message to it.

kat-perez: yes, I will squash that commit into the original SMBIOS commit, just singled it out here to call out the change

components/patina_smbios/src/service.rs
```diff
@@ -101,22 +101,82 @@ impl<'a> Iterator for SmbiosRecordsIter<'a> {
...
-/// - **Type-safe operations**: `add_record<T>()` for structured record types (recommended)
-/// - **Record management**: `update_string()`, `remove()`
-/// - **Table management**: `version()`, `publish_table()`
+/// This trait defines the core SMBIOS operations that can be used through
+/// a trait object (`Service<dyn SmbiosService>`), enabling mocking and testing.
 ///
-/// All operations are protected by a TplMutex at TPL_NOTIFY level for thread safety.
+/// Generic operations like `add_record<T>()` are provided via an extension
+/// implementation on `Service<dyn SmbiosService>`.
+pub trait SmbiosService {
```
Javagedes: Please implement `mockall::automock` on this service trait following the standard practice. It is not documented (I've taken a note, and will add it to our documentation).

Here is an example:

https://github.com/OpenDevicePartnership/patina/blob/5cd9e9d6c60d41c004272c9594a82f65b1fc7af9/components/patina_mm/src/service/platform_mm_control.rs#L19

There is also some conditional compilation regarding features / crates in the crate:

https://github.com/OpenDevicePartnership/patina/blob/5cd9e9d6c60d41c004272c9594a82f65b1fc7af9/components/patina_mm/Cargo.toml#L32

components/patina_smbios/src/service.rs
```diff
@@ -101,22 +101,82 @@ impl<'a> Iterator for SmbiosRecordsIter<'a> {
     }
 }
 
-/// SMBIOS service providing all SMBIOS operations
+/// Object-safe trait for SMBIOS service operations
```
Javagedes, line 107: Could you rename this trait to remove `Service`? I thought we had documentation (I cannot find it, I will update) we try and avoid the word "Config" in the struct name for configs, and "Service" in the trait name for services because ultimately the user requests them via `Config<MyConfig>` and `Service<dyn MyService>` which gets a bit repetitive.

My suggestion is to rename the trait to `Smbios` and the implementor to something else. But it can be somethign else. Your choice.

components/patina_smbios/Cargo.toml
kat-perez: README added to patina_smbios crate and also updated the patina_samples README to include the smbios_platform example

components/patina_smbios/src/component.rs
kat-perez: it is now rebased into the original SMBIOS commit

components/patina_smbios/src/service.rs
kat-perez: yes, sorry forgot that was the convention. will update

components/patina_smbios/src/service.rs
kat-perez: updated

Cargo.toml
makubacki, line 1: Do you have a link to the patina-qemu integration change per (4) in https://github.com/OpenDevicePartnership/patina/blob/main/docs/src/rfc/text/0015-patina-release-process.md#feature-branch-merge-criteria?

kat-perez: links added to PR body

PR #984, opened 2025-10-29
patina_internal_cpu: Update Paging API to Always Return Cache Attrs

### LINKED ISSUE #899: [Bug]: Frequent MTRR updates on QEMU Q35

### Current Behavior

After updating to Patina 5.1.0 from Patina 4.5.1, the debug log has many prints of `MpDxe updating MTRRs with APs`. Hundreds of the prints are occurring during OS load.

### Expected Behavior

Excessive `MpDxe updating MTRRs with APs` prints are not present.

### Steps To Reproduce

Boot Patina v5.1.0 on QEMU Q35.

### Build Environment

```markdown
N/A
```

### Version Information

```text
Tag: v5.1.0
```

### Urgency

Medium

Someone else needs to fix it

No maintainer feedback needed

### Anything else?

_No response_

os-d:
## Description

Currently patina_internal_cpu will not return MTRR cache attributes if a page is unmapped; it will simply return an error. However, the GCD code needs to send the cache attribute changed event to MpDxe to wake up the APs and sync MTRRs whenever the cache attributes change. The GCD took the approach of always sending the event when a page was unmapped and is getting mapped; that resulted in large numbers of the event getting fired, each having to wake up the APs.

This changes the patina_internal_cpu API to return the cache attributes if covered by MTRRs even when the page is unmapped. For ARM64, where the cache attributes are entirely contained within the page table and the MP code starts cores fresh every time with the BSP page table, so syncing is not required, cache attributes will never be returned if the page is unmapped.

This greatly reduces the number of times the event is fired to only the times it needs to.

This PR also updates to paging v10, which just included a breaking change to update the API to not use a custom Result type.

Closes #899 .

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested booting Q35, SBSA, and a physical Intel platform to Windows and observing the number of cache attribute change events decreasing (on Q35 from 403 -> 2).

## Integration Instructions

N/A.

### DIFF
--- a/core/patina_internal_cpu/src/paging.rs
+++ b/core/patina_internal_cpu/src/paging.rs
@@ -9,15 +9,95 @@
 //! SPDX-License-Identifier: Apache-2.0
 //!
 
+use patina_paging::{MemoryAttributes, PtError};
+
 cfg_if::cfg_if! {
-    if #[cfg(all(target_os = "uefi", target_arch = "x86_64"))] {
+    if #[cfg(all(target_arch = "x86_64"))] {
         mod x64;
         pub use x64::create_cpu_x64_paging as create_cpu_paging;
-    } else if #[cfg(all(target_os = "uefi", target_arch = "aarch64"))] {
+    } else if #[cfg(all(target_arch = "aarch64"))] {
         mod aarch64;
         pub use aarch64::create_cpu_aarch64_paging as create_cpu_paging;
     } else {
         mod null;
         pub use null::create_cpu_null_paging as create_cpu_paging;
     }
+}
+
+/// Enum representing the cache attribute value of a memory region if it is not maintained
+/// by the page table. On x64 platforms, this allows for unmapped pages to still reflect
+/// the cache attributes as managed by MTRRs. On ARM64 this will always be NotSupported as
+/// cache attributes are always managed by the page table.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum CacheAttributeValue {
+    /// Valid cache attributes for the memory region
+    Valid(MemoryAttributes),
+    /// The memory region is unmapped
+    Unmapped,
+    /// Cache attributes are only supported via the page table for this architecture
+    NotSupported,
+}
+
+/// The PatinaPageTable trait is Patina's abstraction layer over the PageTable trait
+/// provided by patina-paging. This trait manages architectural abstractions over the page tables.
+pub trait PatinaPageTable {
+    /// Function to map the designated memory region to with provided
+    /// attributes. The requested memory region will be mapped with the specified
+    /// attributes, regardless of the current mapping state of the region.
+    ///
+    /// ## Arguments
+    /// * `address` - The memory address to map.
+    /// * `size` - The memory size to map.
+    /// * `attributes` - The memory attributes to map. The acceptable
+    ///   input will be ExecuteProtect, ReadOnly, as well as Uncacheable,
+    ///   WriteCombining, WriteThrough, Writeback, UncacheableExport
+    ///   Compatible attributes can be "Ored"
+    ///
+    /// ## Errors
+    /// * Returns `Ok(())` if successful else `Err(PtError)` if failed
+    fn map_memory_region(&mut self, address: u64, size: u64, attributes: MemoryAttributes) -> Result<(), PtError>;
+
+    /// Function to unmap the memory region provided by the caller. The
+    /// requested memory region must be fully mapped prior to this call. The
+    /// entire region does not need to have the same mapping state in order
+    /// to unmap it.
+    ///
+    /// ## Arguments
+    /// * `address` - The memory address to map.
+    /// * `size` - The memory size to map.
+    ///
+    /// ## Errors
+    /// * Returns `Ok(())` if successful else `Err(PtError)` if failed
+    fn unmap_memory_region(&mut self, address: u64, size: u64) -> Result<(), PtError>;
+
+    /// Function to install the page table from this page table instance.
+    ///
+    /// ## Errors
+    /// * Returns `Ok(())` if successful else `Err(PtError)` if failed
+    fn install_page_table(&mut self) -> Result<(), PtError>;
+
+    /// Function to query the mapping status and return attribute of supplied
+    /// memory region if it is properly and consistently mapped. This function
+    /// returns the caching attributes for platforms where caching attributes
+    /// are managed outside of the page table (e.g., x86_64 MTRRs), even when
+    /// the page range itself is unmapped.
+    ///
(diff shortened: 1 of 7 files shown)


os-d:
Note, CI will fail until https://github.com/OpenDevicePartnership/patina-paging/pull/137 is merged and released.

patina_dxe_core/src/gcd/spin_locked_gcd.rs
Javagedes: nit: this whole thing could just be 

```rust
let region_attributes = match page_table.query_memory_region(base_address as u64, len as u64) {
    Ok(attr) => attr,
    Err((PtError::NoMapping, attr)) => {
        unmapped = true;
        attr
    },
    Err(e) => {/* current error handling I'm to lazy to re-write */}
}
```

patina_dxe_core/src/gcd/spin_locked_gcd.rs
```diff
@@ -1944,26 +1949,28 @@ impl SpinLockedGcd {
...
-                    if let Some(region_attrs) = region_attributes
-                        && (region_attrs & MemoryAttributes::CacheAttributesMask)
-                            != (paging_attrs & MemoryAttributes::CacheAttributesMask)
-                        && (paging_attrs & MemoryAttributes::CacheAttributesMask) != MemoryAttributes::empty()
-                    {
-                        log::trace!(
-                            target: "paging",
-                            "Attributes for memory region {base_address:#x?} of length {len:#x?} were updated to {paging_attrs:#x?} from {region_attrs:#x?}, sending cache attributes changed event",
-                        );
+                    if paging_attrs & MemoryAttributes::CacheAttributesMask != MemoryAttributes::empty() {
```
Javagedes: This seems like a beast of multi-conditions. Could we clean this up anyway? maybe sub-functions or early exists or something? Or maybe even a better way to evaluate these masks. Not sure. But its a lot.

Javagedes (approved):

patina_dxe_core/src/gcd/spin_locked_gcd.rs
os-d: I did make it one less condition each by pulling it to the top :). I can move some of these checks above and always do the calcuation. It would be a waste of time in some scenarios, but not that big a deal.

core/patina_internal_cpu/src/paging/aarch64.rs
```diff
@@ -117,11 +119,11 @@ mod tests {
     fn test_remap_memory_region() {
         let mut mock_page_table = MockPageTable::new();
 
-        mock_page_table.expect_remap_memory_region().returning(|_, _, _| Ok(()));
+        mock_page_table.expect_map_memory_region().returning(|_, _, _| Ok(()));
 
         let mut paging = EfiCpuPagingAArch64 { paging: mock_page_table };
 
-        let result = paging.remap_memory_region(0x1000, 0x1000, MemoryAttributes::Uncacheable);
+        let result = paging.rap_memory_region(0x1000, 0x1000, MemoryAttributes::Uncacheable);
```
makubacki: This looks like a typo.

```suggestion
        let result = paging.remap_memory_region(0x1000, 0x1000, MemoryAttributes::Uncacheable);
```

patina_dxe_core/src/gcd/spin_locked_gcd.rs
```diff
@@ -1944,26 +1949,28 @@ impl SpinLockedGcd {
...
-                        EVENT_DB.signal_group(CACHE_ATTRIBUTE_CHANGE_EVENT_GROUP);
-                    } else if unmapped
-                        && (paging_attrs & MemoryAttributes::CacheAttributesMask) != MemoryAttributes::empty()
-                    {
-                        log::trace!(
-                            target: "paging",
-                            "Memory region {base_address:#x?} of length {len:#x?} mapped, sending cache attributes changed event",
-                        );
+                            EVENT_DB.signal_group(CACHE_ATTRIBUTE_CHANGE_EVENT_GROUP);
+                        } else if unmapped && region_attributes.is_none() {
```
makubacki: Afaict, this is currently handling two cases inside the `paging_attrs & MemoryAttributes::CacheAttributesMask != MemoryAttributes::empty()` block:

1. The first case is a mapped page with attributes:

   ```rust
   let Some(region_attrs) = region_attributes
                            && (region_attrs & MemoryAttributes::CacheAttributesMask)
                                != (paging_attrs & MemoryAttributes::CacheAttributesMask)
   ```
2. The second case is an unmapped page without cache attributes:

   ```rust
   unmapped && region_attributes.is_none()
   ```

Is the case of an unmapped page with cache attributes intended to not be handled here?

os-d: The first case actually handles two cases:
- mapped page with protection + cache attributes
- unmapped page with cache attributes only

`region_attributes` gets populated in both cases:

Image

makubacki: Ah, okay. Broken down in these cases, I think the relatively complex logic is a little easier to parse some simple comments calling out the cases might be helpful.

os-d: I did have some comments in the cases, but I can improve them if needed. I did, per Joey's suggestion, make that whole conditional simpler, because it was dangerously whacky

image


Giving a quick test and I'll push that up

makubacki: Yeah, that looks more comprehensible.

os-d:
@cfernald @makubacki @Javagedes do you want to take one more look? I realized that on ARM64, where the paging attributes are always controlled by the page table (and where this event is not needed at all) we would fire the event every time we map a page because we would see `unmapped` and `None` for cache attributes.

So, I refactored the return type to be a CacheAttributeValue enum that tells us whether the attributes are Valid (and what they are), Unmapped, or NotSupported. In the case of NotSupported, we won't needlessly fire the event.

core/patina_internal_cpu/src/paging.rs
```diff
@@ -9,15 +9,95 @@
...
+    } else if #[cfg(all(target_arch = "aarch64"))] {
         mod aarch64;
         pub use aarch64::create_cpu_aarch64_paging as create_cpu_paging;
     } else {
         mod null;
         pub use null::create_cpu_null_paging as create_cpu_paging;
     }
 }
+
+/// Enum representing the cache attribute value of a memory region if it is not maintined
```
makubacki:
```suggestion
/// Enum representing the cache attribute value of a memory region if it is not maintained
```

PR #988, opened 2025-10-30
Change the behavior of unimplemented system tables functions from panic to return error.

### LINKED ISSUE #987: [Bug]: Functions in RuntimeServices and BootServices that are not available yet should not panic.

### Current Behavior

If a driver outside patina invokes a RuntimeServices or BootServices API that is not yet available (the obvious one being VariableServices, others include time, watchdog, and monotonic count APIs), the current Patina core panics with unimplemented!().

### Expected Behavior

EDK2 behavior is to return EFI_NOT_AVAILABLE_YET (0xA000000000000002) status for methods that are not yet available.

Since there are potentially drivers/modules that maybe pre-compiled or delivered in a binary-only form that rely on this behavior, it would be better for the core to return an error than panic in the case where the routine has not yet been installed. 

### Steps To Reproduce

1. Invoke any VariableServices call before the Variable Driver runs
2. Observe that it panics. 

### Build Environment

```markdown
N/A - not environment-specific.
```

### Version Information

```text
14.0.0
```

### Urgency

Low

I will fix it

No maintainer feedback needed

### Anything else?

_No response_

joschock:
## Description

Change the behavior of unimplemented system table functions from `unimplemented!()` panic to return error (or nothing, in the event that no return is specified). This will improve compatibility with existing UEFI code that relies on this behavior. 

This is consistent with PI spec language in V1.9 Volume 2 Section 9.5 (https://uefi.org/specs/PI/1.9/V2_DXE_Foundation.html#handing-control-to-dxe-dispatcher).

Closes #987

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Builds/passes unit tests. Confirmed boot on hardware. 

## Integration Instructions

No functional change except that code that (probably incorrectly) calls the un-implemented functions will now receive an error code instead of a panic.

### DIFF
--- a/patina_dxe_core/src/systemtables.rs
+++ b/patina_dxe_core/src/systemtables.rs
@@ -11,7 +11,7 @@
 use core::{ffi::c_void, mem::size_of, slice::from_raw_parts};
 
 use alloc::{alloc::Allocator, boxed::Box};
-use patina::{boot_services::BootServices, component::IntoComponent};
+use patina::{boot_services::BootServices, component::IntoComponent, pi::error_codes::EFI_NOT_AVAILABLE_YET};
 use r_efi::efi;
 
 use crate::{allocator::EFI_RUNTIME_SERVICES_DATA_ALLOCATOR, tpl_lock};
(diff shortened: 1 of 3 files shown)


patina_dxe_core/src/systemtables.rs
```diff
@@ -23,16 +23,19 @@ pub struct EfiRuntimeServicesTable {
     runtime_services: Box<efi::RuntimeServices, &'static dyn Allocator>,
 }
 
+// Defined in PI Specification version 1.9 Appendix A https://uefi.org/specs/PI/1.9/V2_DXE_Appendix_A.html
+pub const EFI_NOT_AVAILABLE_YET: usize = (0xA0 << ((core::mem::size_of::<usize>() * 8) - 8)) | 0x02;
```
makubacki: Is there a reason not to put this in `patina::pi`? I know it's primarily used for these services not being available, but I think that's a more intuitive location for the constant to be defined in.

Javagedes (approved):

patina_dxe_core/src/systemtables.rs
joschock: Agree; will move.

joschock: fixed in latest push - added an `error_codes` module to PI.

PR #989, opened 2025-10-30
docs/dev/testing.md: Update test instructions

makubacki:
## Description

Provide additional background and guidance on choosing a testing strategy.

Update unit test coverage requirements to state that 80% patch coverage is required.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [x] Includes documentation?

## How This Was Tested

- `cargo make cspell`
- markdown lint

## Integration Instructions

- N/A

### DIFF
--- a/docs/src/dev/testing.md
+++ b/docs/src/dev/testing.md
@@ -6,15 +6,64 @@ for project-specific details.
 
 Testing in Rust can be broken up into four core categories:
 
-1. **Unit Testing:** The developer has access to the internal, private state of the module to ensure the internals
-   work as expected.
-2. **Integration Testing:** Tests are written outside of the module and test the code from an external interface
-   standpoint.
-3. **Platform Testing:** Tests that run on the platform, physical or virtual.
-4. **Doc Testing:** Code snippets in inline documentation can be compiled and executed. See
-   [Rust Documentation Tests](https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html).
+1. **[Unit Testing](https://doc.rust-lang.org/rust-by-example/testing/unit_testing.html):** The developer has access to
+   the internal, private state of the module to ensure the internals work as expected.
+2. **[Integration Testing](https://doc.rust-lang.org/rust-by-example/testing/integration_testing.html):** Tests are
+   written outside of the module and test the code from an external interface standpoint.
+3. **[Platform Testing](testing/platform.md):** Tests that run on the platform, physical or virtual.
+4. **[Doc Testing](https://doc.rust-lang.org/rust-by-example/testing/doc_testing.html):** Code snippets in inline
+   documentation can be compiled and executed. See
+     - See also: [Rust Documentation Tests](https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html).
 
 For more, see [Rust Book: Testing](https://doc.rust-lang.org/rust-by-example/testing.html).
+
+## Choosing a Testing Strategy
+
+When writing tests, consider the following questions to help choose a testing strategy:
+
+- **What am I trying to test?**
+  - Internal logic -> Unit tests (nearly all logic should have associated unit tests)
+  - Public interfaces -> Integration tests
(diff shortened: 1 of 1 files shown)


docs/src/dev/testing.md
```diff
@@ -6,16 +6,64 @@ for project-specific details.
...
+    - The [patina_debugger](https://crates.io/crates/patina_debugger) crate can help with source debugging platform
+      tests.
+
+Ideally, code should be built utilizing all available strategies. Unit and integration tests are very fast and run
+frequently during development, so they assist with verifying basic functionality, are especially useful for finding
+regressions during development, and support a rich debug environment with minimal setup to execute the test on a
+host system.
+
+Unit and integration tests greatly improve the overall agility of code changes and refactoring. Code design should
+consider testability from the start so the code being written can be exercised with maximum effectiveness on the host
```
Javagedes, line 52: it may be worth mentioning (maybe here, maybe somewhere else, maybe not at all) that an effective way to write testable code is to write small functions that do one thing - even if they are only called once. This breaks up the API surface into smaller chunks for much simpler testing and allows the compiler to have more control when optimizing for size or performance.

Javagedes: This may also just be something obvious that does not need mentioning :)

cfernald:
@os-d, I generally agree that having this as an actual block is annoying. However, we do have the check that fails and the maintainers control merge, so I think it is their purgative to when to ignore that for urgency or lack of applicability. @makubacki / @Javagedes , were you thinking we actually make that a merge blocking from the GitHub side or a maintainer policy side?

os-d:
> @os-d, I generally agree that having this as an actual block is annoying. However, we do have the check that fails and the maintainers control merge, so I think it is their purgative to when to ignore that for urgency or lack of applicability. @makubacki / @Javagedes , were you thinking we actually make that a merge blocking from the GitHub side or a maintainer policy side?

@cfernald this PR is intended to change the behavior you are describing

image

cfernald:
@os-d , I suppose my question is between the stated requirements vs the "mandatory" for the github actions

Javagedes:
> @os-d, I generally agree that having this as an actual block is annoying. However, we do have the check that fails and the maintainers control merge, so I think it is their purgative to when to ignore that for urgency or lack of applicability. @makubacki / @Javagedes , were you thinking we actually make that a merge blocking from the GitHub side or a maintainer policy side?

I think marking it as "required" is good because it actually forces people to think about and possibly do it rather then them thinking "eh this is informational, I'm not gonna do it".

On the maintainer side we can just bypass it if we need to get something in and need to come back later for tests:
image

If tests are just stupid for the code being added then as the document says, they should be turning coverage off with the attribute and providing justification as to why. As an example, Kat's PR has plenty of coverage disablement, as they are just wrapper functions that are pointless to test.

os-d:
> On the maintainer side we can just bypass it if we need to get something in and need to come back later for tests

Like I said, I'm good with this, as long as maintainers can merge. Because the usual case will not be tests we want to come back later for, those should almost always be written then, its for the case where you have:

```rust
fn testable() {
...
  if (condition_that_cannot_be_met_in_tests) {
    // change single line of code here, patch coverage is 0%
  }
...
}
```

docs/src/dev/testing.md
makubacki: I called it out.

makubacki:
To clarify:

- This is going to cause some pain for people touching code that is not covered. That is partially the point of enforcing it (from GitHub). Technical debt (untested code in this case) has to be paid by someone and what's done is done.
- There is a binary choice here - enforce patch coverage or not. Not enforcing is more convenient but is too detrimental to our long-term testing strategy to ignore on a regular basis.

The goal of this is to make uncovered code actually mean something. Today, it is ambiguous if that code was simply not covered or *could* not be covered. `#[coverage(off)]` is mentioned for the latter. In a worst case, we can override the check.

PR #1006, opened 2025-11-03
dxe_core: init_paging: Ensure Correct Memory Allocation Module

os-d:
## Description

Many systems only have a single memory allocation module HOB that is produced, which is for DXE Core. However, some systems, such as those using EDK II's Standalone MM IPL, can have additional memory allocation module HOBs.

This updates the init_paging routine to ensure it is only finding the DXE Core memory allocation module HOB to map; a similar change was made to the image initialization code, but wasn't also done here.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?
- [ ] Creates a new crate?

## How This Was Tested

Booting Q35 to Windows.

## Integration Instructions

N/A.


os-d:
Note: code coverage is 0% on this patch, but if you note from the branch name, I am adding tests to this file as we speak when I encountered this bug and decided to fix it first.

Javagedes (approved):

PR #1008, opened 2025-11-04
Make `create_performance_measurement` safer for internal (Rust) usage

berlin-with0ut-return:
## Description
Removes raw pointers from `create_performance_measurement` and adds a `CallerIdentifier` enum to distinguish different perf id `caller_identifier` interpretations.

The main goal of this PR is to make internal performance logging (ie through `log_performance_measuresurement`) use Rust-compliant interfaces like Option or enums instead of raw pointers, since raw pointers aren't necessary for the operation of internal performance instrumentation. The edk2 protocol interface remains unchanged.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?
- [ ] Creates a new crate?

## How This Was Tested
Boot and gather perf records on QEMU.

## Integration Instructions
N/A.

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -155,7 +155,7 @@ description = "Run tests and collect coverage data without generating reports."
 install_crate = false
 clear = true
 command = "cargo"
-args = ["llvm-cov", "@@split(COV_FLAGS, )", "--no-report"]
+args = ["llvm-cov", "@@split(COV_FLAGS, )", "--no-report", "--doctests"]
 dependencies = ["individual-package-targets", "llvm-cov-clean"]
 
 [tasks.coverage-lcov]
--- a/components/patina_adv_logger/Cargo.toml
+++ b/components/patina_adv_logger/Cargo.toml
@@ -28,6 +28,9 @@ zerocopy-derive = { workspace = true }
 # Only used for CLI
 clap = { workspace = true, features = ['derive'], optional = true }
 
+[dev-dependencies]
+patina_dxe_core = { path = "../../patina_dxe_core" }
+
 [features]
 default = []
 std = ['clap']
--- a/components/patina_adv_logger/README.md
+++ b/components/patina_adv_logger/README.md
@@ -1,32 +1,93 @@
 # Patina Advanced Logger
 
-The Patina Advanced Logger provides the Patina core with in-memory and serial logging capabilities.
+The Patina Advanced Logger crate provides a [log::Log](https://crates.io/crates/log) implementation including in-memory
+and serial logging capabilities and a component for producing the Advanced Logger protocol. The in-memory logging is
+written to the advanced logger memory buffer, whose address is produced via the Advanced logger guided HOB described
+below. In addition to the runtime functionality described above, this crate also supplies a command line tool to parse
+parse logs pulled off of a physical device, and a public parser object for creating custom tooling.
 
-## Platform Integration
+## Memory Log Behavior
+
+The crate stores records in a shared memory buffer that begins with an `ADVANCED_LOGGER_INFO` header.
+
+- Aligned entries follow in the memory buffer after the header.
+- Each entry records the boot phase identifier, EFI debug level mask, timestamp counter, and message bytes.
+
+## Parser Support
+
+This crate includes a bare-bones log parser executable for parsing the logs produced via the logger, and a public log
+parser struct (`patina_adv_logger::parser::Parser`) that can be used to create custom tooling. Both are available via
+the `std` feature, which exposes the `parser` module.
+
+The command line executable accesses the buffer, prints header metadata, and emits log lines with optional level and
+timestamp context. This parser underpins host utilities and remains version-aligned with the memory layout implemented
+in `memory_log.rs`.
+
+## Integration Instructions
+
+### Patina DXE Core Integration instructions
+
+Below are the instructions for setting up and configuring the patina component and logger implementations inside of
+your platform's Patina DXE Core binary. Additional setup will be required for your Platform, which is discussed further
+below.
 
 1. Instantiate `AdvancedLogger` with the desired format, filters, level, and serial implementation.
    Register it with `log::set_logger` as early as possible.
 2. Call `AdvancedLogger::init` with the physical HOB list pointer.
    This allows the logger to adopt the buffer and record its address for later protocol publication.
-3. Discover or provision an Advanced Logger buffer.
-   Note: Platform firmware typically passes the Advanced Logger HOB from a prior boot phase.
-4. Dispatch the `AdvancedLoggerComponent` through Patina so it can install the Advanced Logger protocol via boot
-   services.
+3. Register the advanced logger component (`AdvancedLoggerComponent`) to be dispatched by the Patina DXE Core so it
+   can install the Advanced Logger protocol via boot services.
 
-## Memory Log Behavior
+#### Example
 
-The crate stores records in a shared buffer that begins with an `ADVANCED_LOGGER_INFO` header.
+```rust
+use patina_dxe_core::*;
+use patina::{log::Format, serial::uart::UartNull};
+use patina_adv_logger::{component::AdvancedLoggerComponent, logger::AdvancedLogger};
 
-- Aligned entries follow the header.
-- Each entry records the boot phase identifier, EFI debug level mask, timestamp counter, and message bytes.
+use log::LevelFilter;
+use core::ffi::c_void;
 
-## Parser Support
+static LOGGER: AdvancedLogger<UartNull> = AdvancedLogger::new(
+   Format::Standard, // How logs are formatted
+   &[("allocations", LevelFilter::Off)], // set custom log levels per module
+   log::LevelFilter::Info, // Default log level
+   UartNull { }, // Serial writer instance
+);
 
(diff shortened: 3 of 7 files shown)


sdk/patina/src/uefi_protocol/performance_measurement.rs
```diff
@@ -40,7 +40,7 @@ pub enum PerfAttribute {
 }
 
 /// Function to create performance record with event description and a timestamp.
-pub type CreateMeasurement = unsafe extern "efiapi" fn(
+pub type CreateMeasurementExt = unsafe extern "efiapi" fn(
```
Javagedes: I'm not sure `Ext` is the most accurate name here. Its not an extension of core functionality, it is the core functionality. I would consider renaming this. I honestly can't think of a much better name. Maybe `Uefi` instead. Interested to know if others have an opinion.

sdk/patina/src/performance/measurement.rs
```diff
@@ -273,10 +282,103 @@ pub unsafe extern "efiapi" fn create_performance_measurement(
...
+        if let CallerIdentifier::ImageHandle(h) = *self { Some(h) } else { None }
+    }
+
+    /// Returns the GUID if the `CallerIdentifier` is a GUID pointer.
+    pub fn as_guid(&self) -> Option<&efi::Guid> {
+        if let CallerIdentifier::Guid(ref g) = *self { Some(g) } else { None }
+    }
+}
+
+pub fn create_performance_measurement_internal(
```
Javagedes: I completely understand why you named this function `create_performance_measurement_internal`, as it has been our naming standard when we are naming the internal rust implementation that are the backing functionality of an edk2 protocol, however in this is one of the instances where I don't think that naming scheme makes sense.

This is because function is actually a part of the public API for measuring performance, so having a public function in a public crate (our sdk) with the name `internal` is a bit awkward. My suggestion is to keep this named `create_performance_measurement` and rename the existing function to something else. maybe `_protocol` or `_uefi` or something. I'm don't have the best suggestion on that part.

sdk/patina/src/performance/measurement.rs
Javagedes: Two things here, with your API Changes.

This function should no longer be public. It will be confusing to anyone using the `patina` crate when they look at the documentation on docs.rs and see a `create_performance_measurement` function the ugly `C` interface with pointers and then also a `create_performance_measurement_internal` function that is the nice rust interface.

The second is that now that you have this new level of indirection, you should use it. At this point, you have the rust function and UEFI compatible function pointing back at _create_performance_measurement. You can change that upside down pyramid to just a linear line. The call order would be 

1. `"efiapi" fn create_performance_measurement` function (calls the)
2. `create_performance_measurement_internal` (should rename this) function (calls the)
3. `_create_performance_measurement` function

This reduces some code duplication regarding grabbing the static state, which has already seen a split in logic. The UEFI version returns efi::Status::SUCCESS if it's not initialized whereas the inner function returns efi::Status::NOT_READY. So that bit of weirdness also needs to be investigated. If that is on purpose then you can map the result such that efi::Status::NOT_READY -> efi::Status::EFI_SUCCESS, but that will need to be documented.

sdk/patina/src/performance/logging.rs
```diff
@@ -467,10 +467,10 @@ pub fn perf_start_ex(
 /// Adds a record that records the end time of a performance measurement.
 pub fn perf_end_ex(
     handle: efi::Handle,
-    token: *const c_char,
-    module: *const c_char,
+    token: Option<&str>,
+    module: Option<&str>,
     timestamp: u64,
-    identifier: u32,
+    identifier: u16,
```
Javagedes: I understand you changed this to u16 because the core functionality takes a u16, but this all needs to be changed to a u32 to match the UEFI implementation, including `_create_performance_measurement` and the `KnownPerfId::try_from(perf_id)` logic.

sdk/patina/src/performance/measurement.rs
berlin-with0ut-return: the efiapi function does need to be public because the uefi_protocol definition is in patina/uefi_protocol (sdk) while the performance component's entry point, which installs the protocol, is in components/.

Javagedes: Seems like the "efiapi" function should be defined in the component then. Is that possible? Seems like it should be if its just going to be a wrapper around the public create_performance_measurement function now.

Especially because we really don't want protocol implementations in the SDK. what happens when a platform needs to swap it out with a different implementation?

sdk/patina/src/performance/measurement.rs
```diff
@@ -273,10 +276,104 @@ pub unsafe extern "efiapi" fn create_performance_measurement(
     }
 }
 
+/// Represents the `caller_identifier` used in performance measurements.
+/// Due to legacy reasons, this can either be an image handle or a pointer to a GUID.
+pub enum CallerIdentifier {
+    ImageHandle(efi::Handle),
```
joschock: pedantry: consider just `Handle` as a DriverBinding Handle is usually, but not always, an Image Handle.

sdk/patina/src/performance/measurement.rs
```diff
@@ -192,13 +192,14 @@ pub mod event_callback {
...
+//
+/// Skip coverage as this function is tested via the generic version, (_create_performance_measurement).
 ///
 /// # Safety
 /// `string` must be a valid C string pointer.
 /// `caller_identifier` must be a valid image handle or GUID pointer.
-pub unsafe extern "efiapi" fn create_performance_measurement(
+pub unsafe extern "efiapi" fn create_performance_measurement_protocol(
     caller_identifier: *const c_void,
     guid: Option<&efi::Guid>,
```
joschock: if the FFI interface is *const GUID, then this representation may not work even though Option<&efi::Guid> has the same layout, since C code may not provide the expected alignment on the guid pointer. I'm not sure if this is an issue, given that it works today, but something to be aware of.

sdk/patina/src/performance/measurement.rs
```diff
@@ -192,13 +192,14 @@ pub mod event_callback {
...
-// it not mockable.
+// EDKII Performance Measurement Protocol implementation.
+//
+/// Skip coverage as this function is tested via the generic version, (_create_performance_measurement).
 ///
 /// # Safety
 /// `string` must be a valid C string pointer.
 /// `caller_identifier` must be a valid image handle or GUID pointer.
-pub unsafe extern "efiapi" fn create_performance_measurement(
+pub unsafe extern "efiapi" fn create_performance_measurement_protocol(
```
Javagedes: nit: With my work removing statics, I've started prepending these functions with `efiapi` e.g. `create_performance_measurement_efiapi`. That has not hit a PR yet, but it will soon. Since you are actively changing this function name, do you mind updating it to `create_performance_measurement_efiapi`?

Javagedes (approved):

components/patina_performance/src/component/protocol.rs
makubacki, line 1: Please add a module doc comment to this file.

components/patina_performance/src/component/protocol.rs
```diff
@@ -0,0 +1,100 @@
...
+        error::Error,
+        measurement::{CallerIdentifier, create_performance_measurement},
+        record::known::{KnownPerfId, KnownPerfToken},
+    },
+    uefi_protocol::performance_measurement::PerfAttribute,
+};
+use r_efi::efi;
+
+#[coverage(off)]
+// EDKII Performance Measurement Protocol implementation.
```
makubacki: We're trying to more consistently use:

```suggestion
// EDK II Performance Measurement Protocol implementation.
```

components/patina_performance/src/component/protocol.rs
```diff
@@ -0,0 +1,100 @@
...
+    // SAFETY: The caller ensures that string is a valid C string pointer (or NULL).
+    let string = unsafe { string.as_ref().map(|s| CStr::from_ptr(s).to_string_lossy().to_string()) };
+
+    // NOTE: If the Perf is not the known Token used in the core but have same ID with the core Token, this case will
+    //       not be supported.
+    // And in current usage mode, for the unknown ID, there is a general rule:
+    //   - If it is start pref: the lower 4 bits of the ID should be 0.
+    //   - If it is end pref: the lower 4 bits of the ID should not be 0.
+    //   - If input ID doesn't follow the rule, we will adjust it.
+    let mut perf_id = identifier as u16;
```
makubacki, line 55: Since this returns a status, should it more explicitly check the `identifier` value?

For example: `if identifier > u16::MAX { return efi::Status::INVALID_PARAMETER; }`

sdk/patina/src/performance/logging.rs
```diff
@@ -7,92 +7,98 @@
...
+    } else if module.is_some() {
         module
     } else {
-        ptr::null()
-    };
-    // SAFETY: string parameter is expected to be a valid C string.
-    unsafe {
-        (create_performance_measurement)(handle, None, string, timestamp, 0, identifier, PerfAttribute::PerfStartEntry)
+        None
     };
```
makubacki: I think you could more concisely do this with:

```rust
let string = token.or(module);
```

makubacki (approved):

PR #1011, opened 2025-11-04
sdk: Prevent null pointer dereferences [Rebase & FF]

makubacki:
## Description

Based on recent CodeQL results, update non-test code findings to more explicitly handle potentially null pointers.

- `c_ptr`: Panic with a clear message if `into_original_ptr()` is called when the stored pointer is null since it will attempt to to copy from that address.
- `volume`: Update `VolumeRef::new_from_address()` to validate the incoming FV header address before attempting to convert the address to a pointer and read from it.

---

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?
- [ ] Creates a new crate?

## How This Was Tested

- `cargo make all`
- CodeQL with rust queries

## Integration Instructions

- N/A

### DIFF
--- a/sdk/patina/src/boot_services/c_ptr.rs
+++ b/sdk/patina/src/boot_services/c_ptr.rs
@@ -11,6 +11,7 @@ use core::{
     ffi::c_void,
     marker::PhantomData,
     mem::{self, ManuallyDrop},
+    num::NonZeroUsize,
     ops::Deref,
     ptr,
 };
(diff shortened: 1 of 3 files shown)


sdk/patina/src/boot_services/c_ptr.rs
```diff
@@ -34,7 +34,11 @@ impl<'a, R: CPtr<'a, Type = T>, T> PtrMetadata<'a, R> {
     ///
     /// # Safety
     /// Caller must ensure that the pointed-to memory is still valid and uphold rust pointer invariants (e.g. no aliasing)
+    ///
+    /// # Panics
+    /// Panics if the stored pointer value is null
     pub unsafe fn into_original_ptr(self) -> R {
```
cfernald: Should this instead return an option to push the error handling to the caller? To avoid hidden panics when possible.

cfernald (approved):

sdk/patina/src/boot_services/c_ptr.rs
os-d: I had wondered the same thing, but have been directed towards `try_*` implementations for that. That was to support const use cases though, and this isn't const. I haven't looked further at what this is doing.

makubacki: I was not intending to change the design just the existing implementation. However, I only see a few callers so it wouldn't be too difficult to switch to a fallible function. Afaict, it's mostly used in protocol related operations to reconstruct pointers when (1) drop its memory (2) move ownership when installing and uninstalling protocols.

I'll add a second commit with a change to return an `Option` to cleanly separate it from the existing changes.

makubacki: @cfernald & @os-d, I put that change in https://github.com/OpenDevicePartnership/patina/pull/1011/commits/d576480f127e5867021ea394b56af2d1a16537e6.

os-d (approved):
LGTM: nit that all the safety comments (not touched in this PR) say: pointer is “leak” instead of leaked

sdk/patina/src/boot_services/c_ptr.rs
cfernald: looks good. thanks!

makubacki:
> LGTM: nit that all the safety comments (not touched in this PR) say: pointer is “leak” instead of leaked

I noticed that too, erred on the side of thinking they were good enough.

PR #1013, opened 2025-11-04
RFC [Reject]: Boot Orchestration Component

kat-perez:
## Description

This RFC proposes modular components for UEFI spec compliant boot functionality in Patina firmware. Platforms can
use reference implementations for console initialization, boot option discovery, and boot orchestration,
or implement custom boot flows using the provided library functions.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [x] Includes documentation?
- [ ] Creates a new crate?

## How This Was Tested

N/A

## Integration Instructions

N/A

### DIFF
--- a/docs/src/rfc/rejected/0028-patina-boot.md
+++ b/docs/src/rfc/rejected/0028-patina-boot.md
@@ -1,0 +1,795 @@
+# RFC: Boot Orchestration
+
+This RFC proposes modular components for UEFI-compliant boot functionality in Patina firmware.
+
+Within the scope of this RFC, boot functionality includes console initialization, boot option discovery,
+and boot orchestration as defined in UEFI Specification 2.11 Chapter 3 (Boot Manager). This RFC also
+includes BDS phase event signaling (EndOfDxe and ReadyToBoot) as defined in the Platform Initialization
+(PI) Specification.
+
+## Change Log
+
+- 2025-11-04: Initial RFC draft
+- 2025-11-06:
+  - Rewritten with modular component architecture
+  - Added optional service dependencies for platform customization
+  - Added BDS phase event signaling (EndOfDxe, ReadyToBoot)
+- 2025-12-04:
+  - Updated watchdog timer reference to UEFI Specification 2.11 Section 3.1.2
+  - Added requirement to disable watchdog when boot option returns control
+- 2025-12-09:
+  - Added component rationales and decomposition justification
+- 2025-12-10:
+  - Fixed EndOfDxe timing: now signaled after PCIe enumeration AND GOP discovery complete
+  - Added Future Work section for Connect Controller Service
+  - Expanded ConsoleDiscovery prerequisites (PCI/USB bus enumeration requirements)
+  - Added custom orchestrator example
+- 2025-12-12:
+  - Added BootOptionDiscovery component for full UEFI boot variable management
+  - Reframed RFC: UEFI-compliant boot as primary case, hardcoded paths as alternative for divergent platforms
+- 2025-12-18:
+  - Added Unresolved Questions section for BDS phase dispatch mechanism
+- 2026-01-14:
+  - Resolved BDS phase dispatch mechanism: dual integration paths (native Patina and legacy bridge)
+- 2026-02-03:
+  - Added partial device path expansion support with `expand_device_path()` helper (#1280)
+  - `boot_from_device_path()` now handles both full and partial paths transparently
+- 2026-02-10:
+  - Added Implementation Phases section defining Minimal Boot as first deliverable
+  - Minimal Boot specifies two boot flows: normal (ordered fallback chain) and hotkey (direct to error on failure)
+  - Reordered Guide-Level Explanation to lead with Minimal Boot usage
+- 2026-02-10:
+  - Added Implementation Phases section defining Minimal Boot as first deliverable
+  - Minimal Boot demonstrates patina_boot booting Windows with a shorter BDS phase
+  - Reordered Guide-Level Explanation to lead with Minimal Boot usage
+- 2026-02-12:
+  - Refactored BootOrchestrator from concrete component to trait-based design
+  - Added `BootOrchestrator` trait, `BootDispatcher` component, `SimpleBootManager` reference impl
+  - Boot options passed directly to `SimpleBootManager::new()` instead of `Config<BootConfig>`
+  - `BootDispatcher` is the Patina component gated by `Service<dyn BdsPhase>`
+  - Platforms provide custom boot flows by implementing `BootOrchestrator` trait
+- 2026-03-31:
+  - Updated RFC to match implementation: `BootConfig` naming, `DxeDispatch` service,
+    BDS architectural protocol installation, import paths
+  - Removed Phase 1/Phase 2 split — core framework fully implemented
+  - `BootOptionDiscovery` implemented as `discover_boot_options()` helper function
+    instead of a separate component (simpler, no framework machinery needed)
+  - `interleave_connect_and_dispatch()` made private to `SimpleBootManager`
+    (uses `connect_all()` which is not recommended for most platforms)
+- 2026-06-16: RFC rejected:
+  - The design is being refined and tested outside the Patina project
+  - Planned for future reevaluation; moved to the rejected folder as a suspended state
+
+## Motivation
+
+Most Patina platforms require UEFI-compliant boot functionality: discovering boot options from
+`Boot####`/`BootOrder`/`BootNext` variables, managing console devices, and orchestrating the boot
+flow per UEFI Specification Chapter 3. This RFC
+provides common components that implement full UEFI Boot Manager compliance.
+
+Some platforms may diverge from UEFI compliance (e.g., embedded systems with hardcoded boot paths). The modular
+architecture supports this by allowing platforms to use individual components selectively or provide boot options
+directly via configuration.
+
+## Technology Background
+
+The UEFI Boot Manager (UEFI Specification 2.11 Chapter 3) operates through UEFI variables (`Boot####`, `BootOrder`,
+`BootNext`, `ConIn`/`ConOut`/`ErrOut`, `ConInDev`/`ConOutDev`/`ErrOutDev`) containing boot configuration
+and console device paths. Boot options use `EFI_LOAD_OPTION`
+structures with device paths and descriptions. The BDS phase signals EndOfDxe after DXE completion and ReadyToBoot
+before boot attempts to coordinate firmware initialization and security lockdown.
+
+## Goals
+
+1. Provide UEFI-compliant boot option discovery (`Boot####`/`BootOrder`/`BootNext` variable management)
+2. Provide foundational boot orchestration for executing boot options
+3. Provide a common boot flow implementation suitable for most platforms
+4. Support platform-specific boot flows through custom orchestration
+5. Maintain modular architecture where components can be used, replaced, or omitted independently
+6. Provide reusable library functions to minimize code duplication across platforms
+
+## Requirements
+
+This RFC implements specific elements from UEFI Specification 2.11 Chapter 3 (Boot Manager) and Platform
+Initialization (PI) Specification Volume 2 (DXE/BDS phase). The following requirements define the scope
+of boot functionality provided by these components.
+
+### UEFI Specification 2.11 Chapter 3 (Boot Manager)
+
+**Boot Option Discovery**:
+
+- Read and parse `Boot####` variables containing `EFI_LOAD_OPTION` structures
+- Read `BootOrder` variable to determine boot attempt sequence
+- Check `BootNext` variable to override boot sequence for a single boot (UINT16 referencing a
+  `Boot####` option; deleted after consumption per Section 3.1.2)
+- Manage boot option list (add, remove, reorder boot options)
+- Support boot option attributes (LOAD_OPTION_ACTIVE, LOAD_OPTION_HIDDEN, etc.)
+- Expand partial device paths to full device paths by matching against discovered device topology
+  (e.g., hard drive media device paths containing only partition GUID/signature)
+
+**Console Variables**:
+
+- Discover console devices via GOP (graphics) and SimpleTextInput protocols
+- Populate `ConIn`, `ConOut`, and `ErrOut` variables with console device paths (saved to NVRAM,
+  represent default/selected consoles)
+- Populate `ConInDev`, `ConOutDev`, and `ErrOutDev` volatile variables with all discovered console
+  device paths each boot (represent all available console devices found during enumeration)
+
+**Boot Orchestration**:
+
+- If `BootNext` is set, attempt that boot option first and delete the variable regardless of outcome
+- Execute boot options from `BootOrder` sequence (or platform-provided configuration)
+- Load boot images via `LoadImage()` service
+- Start boot images via `StartImage()` service
+- Enable 5-minute watchdog timer before `StartImage()` per Section 3.1.2
+- Disable watchdog timer when boot option returns control
+- Handle boot failures and attempt next option in sequence
+- Detect hotkey via `Key####` variables (UEFI-compliant) or platform-configured scancode to trigger
+  alternative boot path (e.g., F12 for secondary boot)
+- Read `Key####` variables containing `EFI_KEY_OPTION` structures to determine hotkey-to-boot-option
+  mappings per Section 3.1.6
+- Attempt `PlatformRecovery####` boot options (in `PlatformRecoveryOrder` sequence) when all normal
+  boot options and platform-configured fallback chain are exhausted, per Section 3.4
+
+### PI Specification Volume 2 (DXE/BDS Phase)
+
+**BDS Phase Event Signaling**:
+
+- Signal `gEfiEndOfDxeEventGroupGuid` after device enumeration completes
+- Signal `gEfiEventReadyToBootGuid` immediately before attempting first boot option
+- Enable event notification to C-based PI drivers (MM, security) via EventServices
+
+**Driver Dispatch Coordination**:
+
+- Coordinate driver dispatch during device connection to ensure complete device topology enumeration
+- Support interleaved connect-dispatch pattern: connect controllers, dispatch newly loaded drivers, repeat until stable
+
+### Component Integration
+
+**Patina Component Model**:
+
+- Integrate with Patina's component model using standard dependency injection patterns
+- Enable platform orchestration that controls execution order and can skip or replace steps
+- Support progressive customization from default behavior to full custom orchestration
+
+**Service Dependencies** (from RFC 0017):
+
+- `RuntimeVariableServices`: Read/write UEFI variables (boot options, console variables)
+- `ImageServices`: Load and start boot images (`LoadImage()`, `StartImage()`)
+- `TimingServices`: Configure watchdog timer
+- `ProtocolServices`: Locate protocol handles, connect controllers, query device paths
+- `DxeDispatch`: Orchestrate driver dispatch during connect-dispatch interleaving
+- `EventServices`: Signal BDS phase events (`signal_event_group()` for event group GUIDs)
+
+*Note: This RFC uses `ProtocolServices::connect_controller()` for device connection. If future
+components require additional connection abstractions (e.g., connection filtering, connection
+policies), a dedicated `ConnectionService` could be introduced. For this RFC's scope,
+`ProtocolServices` is sufficient.*
+
+**Customization Support**:
+
+- Default: Use `BootOptionDiscovery` for UEFI-compliant boot variable management
+- Alternative: Accept boot options via `BootConfig` passed directly to
+  `SimpleBootManager::new()` for platforms not requiring UEFI compliance
+- Enable platforms to use common components as-is, compose selectively, or replace entirely
+- Provide library functions for platforms implementing custom boot flows
+
+### Planned Extensions
+
+The following features are natural extensions of the current implementation and may be
+added as helper functions or orchestrator enhancements:
+
+- **Boot menu UI**: Boot timeout countdown and boot option selection UI, using
+  options from `discover_boot_options()`. Platforms with GOP consoles could display
+  a menu; headless platforms could use serial console.
+- **OS indications**: `OsIndications`/`OsIndicationsSupport` for boot-to-firmware-UI
+  requests and other OS-to-firmware communication.
+
+### Out of Scope
+
+The following features are outside the scope of boot orchestration and belong in
+separate subsystems:
+
+- Capsule processing and update capsule handling
+- Language management (`PlatformLang`, `PlatformLangCodes`)
+- Variable policy registration (`EDKII_VARIABLE_POLICY_PROTOCOL`)
+- Hardware error record variables (`HwErr####`, `HwErrRecSupport`)
+- Deferred image loading (`EFI_DEFERRED_IMAGE_LOAD_PROTOCOL`)
+- Automatic boot device provisioning UI
+
+### Planned: Connect Controller Service
+
+This RFC currently uses `connect_controller()` (the UEFI boot service) for device
+enumeration. This works but has limitations: `connect_all()` connects every controller
+on every round (inefficient for large topologies), and native Patina components cannot
+participate in the connect controller flow.
+
+A `Service<dyn ConnectController>` abstraction would allow:
+
+- Selective controller connection (e.g., PCI-only, skip USB for headless)
+- Native Patina bus drivers participating in device discovery
+- Connection filtering and policy components
+- Custom binding logic for platform-specific device handling
+- Replacing `connect_all()` inside `interleave_connect_and_dispatch()` with
+  platform-appropriate connection strategies
+
+This is the primary remaining gap between the current implementation and production
+readiness for platforms with complex device topologies. Until implemented, platforms
+requiring custom connect controller behavior must implement custom `BootOrchestrator`
+traits that call `connect_controller()` directly with platform-specific logic.
+
+## Design Decision: BDS Phase Dispatch Mechanism
+
+The PI Specification (Section II-12.2) defines the BDS Architectural Protocol as the mechanism for
+transferring control from DXE phase to BDS phase. This RFC adopts a Patina-native approach while
+providing a bridge component for platforms with existing C-based BDS implementations.
+
+### Chosen Approach: Dual Integration Paths
+
+Patina provides two integration paths, allowing platforms to choose based on their needs:
+
+#### Path 1: Native Patina (BootDispatcher + BootOrchestrator trait)
+
+Boot orchestration uses a trait-based design. `BootOrchestrator` is a plain Rust trait defining the
+boot flow interface. `BootDispatcher` is the Patina component that installs the BDS architectural protocol and
+holds a `Box<dyn BootOrchestrator>`, invoking it when the DXE core calls BDS. `SimpleBootManager` is the reference
+implementation.
+
+```rust
+/// Trait defining the boot orchestration interface.
+pub trait BootOrchestrator: Send + Sync + 'static {
+    fn execute(
+        &self,
+        boot_services: &StandardBootServices,
+        runtime_services: &StandardRuntimeServices,
+        dxe_services: &dyn DxeDispatch,
+        image_handle: efi::Handle,
+    ) -> Result<!, EfiError>;
+}
+
+/// Component that invokes the orchestrator at BDS time.
+#[component]
+impl BootDispatcher {
+    fn entry_point(
+        self,
+        boot_services: StandardBootServices,
+        runtime_services: StandardRuntimeServices,
+        dxe_services: Service<dyn DxeDispatch>,
+        image_handle: Option<Handle>,
+    ) -> Result<()> {
+        let handle = *image_handle.ok_or(EfiError::InvalidParameter)?;
+        self.orchestrator.execute(&boot_services, &runtime_services, *dxe_services, handle);
+        Ok(())
+    }
+}
+
+/// Reference implementation with configurable boot options.
+impl BootOrchestrator for SimpleBootManager {
+    fn execute(&self, boot_services: &StandardBootServices, runtime_services: &StandardRuntimeServices, dxe_services: &dyn DxeDispatch, image_handle: efi::Handle) -> Result<!, EfiError> {
+        interleave_connect_and_dispatch(boot_services, dxe_services)?;
+        signal_bds_phase_entry(boot_services)?;
+        discover_console_devices(boot_services, runtime_services)?;
+        // Hotkey detection, ReadyToBoot signal, boot attempts...
+    }
+}
+```
+
+*Benefits*: Trait enforces the contract at compile time. Plain Rust trait with `Box<dyn>` dispatch —
+no Patina service indirection needed since there is only one consumer (`BootDispatcher`). Platforms
+swap boot orchestration by providing a different `BootOrchestrator` impl. Single component
+registration: `BootDispatcher::new(impl BootOrchestrator)`.
+
+#### Path 2: Legacy/Hybrid (BdsArchProtocolBridge)
+
+For platforms with existing C-based BDS implementations (e.g., EDK II's `BdsDxe`), a minimal bridge
+component calls the BDS Architectural Protocol. This enables incremental migration and keeps Patina
+agnostic of what owns BDS.
+
+```rust
+/// Minimal component that bridges to existing BDS arch protocol
+#[component]
+impl BdsArchProtocolBridge {
+    fn entry_point(self, protocol_services: Service<dyn ProtocolServices>) -> Result<()> {
+        let bds_protocol = protocol_services.locate_protocol::<bds::Protocol>()?;
+        unsafe { ((*bds_protocol).entry)(bds_protocol) };
+        Ok(())
+    }
+}
+```
+
+*Benefits*: Works with existing C-based BDS drivers, no DXE Core modification required, clean
+migration path for platforms transitioning to native Patina boot.
+
+### Rationale
+
+The BDS Architectural Protocol is a *dispatch mechanism*, not a functional requirement. The PI spec
+requires that BDS phase enumerate boot devices, signal events, and attempt boot options—requirements
+about *what* happens, not *how* it's dispatched.
+
+Providing both paths:
+
+1. Allows native Patina platforms to use idiomatic component patterns
+2. Supports platforms with existing C-based BDS investments
+3. Keeps Patina decoupled from any specific BDS implementation
+4. Enables gradual migration without forcing immediate rewrites
+
+C drivers that register for BDS events (EndOfDxe, ReadyToBoot) via `CreateEventEx()` continue to
+work because both paths signal these events through EventServices.
+
+## Prior Art (Existing EDK II Implementation)
+
+EDK II implements boot functionality through `BdsDxe` (`MdeModulePkg/Universal/BdsDxe`), a monolithic
+driver that combines console initialization, boot option processing, and platform integration in a single
+component. Platform customization occurs through `PlatformBootManagerLib` callbacks:
+`PlatformBootManagerBeforeConsole()`, `PlatformBootManagerAfterConsole()`,
+`PlatformBootManagerWaitCallback()`, and `PlatformBootManagerUnableToBoot()`.
+
+Boot option management uses `UefiBootManagerLib`, which provides functions for parsing `EFI_LOAD_OPTION`
+structures, enumerating boot options, and managing boot variables. Console device connection and boot
+target discovery are hardcoded into the boot flow with limited platform control over execution order.
+
+This design influenced our component architecture by demonstrating the need for console initialization,
+boot discovery, and orchestration as distinct responsibilities. However, the callback pattern limits
+platform flexibility compared to platform-controlled orchestration.
+
+## Alternatives
+
+### Callback-Based Pattern (EDK II Approach)
+
+Single Boot Manager component with fixed execution flow and platform callback hooks (`before_console()`,
+`after_console()`, etc.). Platforms customize behavior through callback traits.
+
+**Limitation**: Callbacks have fixed execution order. Platforms cannot skip steps, change ordering, or
+implement custom control flow like network-before-console or interleaved fallbacks.
+
+### Monolithic Boot Manager Component
+
+Single comprehensive component handling console initialization, boot discovery, and orchestration with
+configuration options.
+
+**Limitation**: Individual pieces cannot be replaced independently of the reference implementation.
+
+### Library-Only Approach
+
+Library functions only (`parse_boot_variables()`, `discover_console_devices()`, etc.) with no reference
+components. Platforms implement orchestration.
+
+**Limitation**: Code duplication across platforms. Simple platforms implement significant boilerplate.
+Higher adoption barrier.
+
+### Chosen Approach: Modular Components with Platform Orchestration
+
+Three components (`BootOptionDiscovery`, `ConsoleDiscovery`, `BootDispatcher`) that platforms use
+as-is, compose selectively, or replace entirely. `BootDispatcher` invokes a `BootOrchestrator` trait
+implementation at BDS time; platforms provide custom boot flows by implementing the trait.
+UEFI-compliant platforms use `BootOptionDiscovery` for boot variable management; platforms
+diverging from UEFI compliance can provide boot options directly to `SimpleBootManager::new()`.
+
+**Benefits**: Full UEFI compliance for most platforms. Custom orchestration via `BootOrchestrator`
+trait (change order, skip steps, custom control flow). Independent component replacement. Platforms
+carry only used code. Escape hatch for non-UEFI-compliant platforms via direct configuration.
+
+## Design Assumptions
+
+This RFC's architecture is based on the following design considerations:
+
+1. **Platform diversity**: Platforms range from simple embedded systems with fixed boot paths to complex
+   servers with network boot, remote attestation, and custom boot policies
+2. **Customization requirement**: Platform-specific boot flows, connection policies, and orchestration
+   requirements necessitate composable components rather than monolithic implementations
+3. **Progressive disclosure**: Architecture supports both simple use cases (single composite component)
+   and complex scenarios (individual component composition with custom orchestration)
+4. **Scope**: This RFC covers boot option discovery and console initialization alongside boot
+   orchestration because all are UEFI Boot Manager responsibilities. BootOptionDiscovery is optional
+   for platforms diverging from UEFI compliance; ConsoleDiscovery is optional for headless systems
+
+## Implementation Status
+
+The core boot orchestration framework is implemented and validated on QEMU Q35 (x64) and
+QEMU SBSA (AArch64), replacing the C-based BdsDxe driver with patina_boot.
+
+### What Is Implemented
+
+- `BootDispatcher` component with BDS architectural protocol installation
+- `BootOrchestrator` trait for custom boot flows
+- `SimpleBootManager` reference implementation
+- `BootConfig` with platform-provided device paths
+- Connect-dispatch interleaving via `DxeDispatch` service
+- Console discovery (GOP + SimpleTextInput + `ConIn`/`ConOut`/`ErrOut` and
+  `ConInDev`/`ConOutDev`/`ErrOutDev` variable writing)
+- BDS phase event signaling (EndOfDxe, ReadyToBoot)
+- Boot image execution via `LoadImage()`/`StartImage()` with watchdog timer
+- Partial device path expansion
+- Hotkey detection via `Key####` variables or platform-configured scancode
+- `PlatformRecovery####` recovery boot paths after normal boot exhaustion
+- Failure handler callback
+- Boot option discovery from UEFI variables (`discover_boot_options()` helper)
+
+### Boot Flows
+
+`SimpleBootManager` defines two boot flows selected at runtime based on hotkey state. Both flows
+share the same device enumeration, event signaling, and error handling infrastructure.
+
+**Common Preamble** (runs before either flow):
+
+1. Connect-dispatch interleaving — connect all controllers and dispatch newly loaded drivers until stable
+2. Console discovery — locate GOP and SimpleTextInput handles, populate console variables
+3. Signal EndOfDxe — security components perform lockdown
+4. Check hotkey — read `Key####` variables or platform-configured scancode; select boot flow
+
+**Flow 1: Normal Boot** (no hotkey detected)
+
+The platform configures an ordered list of boot device paths. The orchestrator attempts each in
+sequence. If all fail, the error path is taken. This supports arbitrary fallback chains (e.g.,
+Windows → Linux → UEFI Shell → PXE → USB → front page).
+
+```text
+Signal ReadyToBoot
+    |
+    v
+For each device path in BootConfig (in order):
+    |
+    Try device path (LoadImage/StartImage with 5-min watchdog)
+        |
+    Success --> boot (never returns)
+        |
+    Failure --> try next device path
+    |
+All device paths exhausted
+    |
+    v
+For each PlatformRecovery#### option (in PlatformRecoveryOrder):
+    |
+    Try recovery device path
+        |
+    Success --> boot (never returns)
+        |
+    Failure --> try next recovery option
+    |
+All recovery options exhausted
+    |
+    v
+Error path: call failure_handler (demo: show error screen)
+```
+
+**Flow 2: Hotkey Boot** (hotkey detected, e.g., F12)
+
+When the platform-configured hotkey is held during boot, the orchestrator switches to the hotkey
+device list. This flow does **not** fall through to the normal boot devices — if the hotkey device
+fails, the error path is taken directly.
+
+```text
+Signal ReadyToBoot
+    |
+    v
+Try hotkey device path (LoadImage/StartImage with 5-min watchdog)
+    |
+Success --> boot (never returns)
+    |
+Failure
+    |
+    v
+Error path: call failure_handler (demo: show error screen)
+```
+
+**Error Path**: Before calling `failure_handler`, the orchestrator attempts `PlatformRecovery####`
+boot options (if any are defined via UEFI variables). The `failure_handler` is a platform-provided
+callback invoked only after all normal and recovery options are exhausted. Production platforms can
+implement diagnostics, user prompts, or system reset as needed.
+
+### Platform-Provided Configuration
+
+```rust
+use patina_boot::{BootDispatcher, SimpleBootManager};
+use patina_boot::config::BootConfig;
+
+// In ComponentInfo::components():
+add.component(BootDispatcher::new(SimpleBootManager::new(
+    BootConfig::new(nvme_esp_device_path())    // 1st: NVMe EFI System Partition
+        .with_device(nvme_recovery_path())       // 2nd: NVMe recovery partition
+        .with_device(pxe_device_path())          // 3rd: PXE network boot
+        .with_device(usb_device_path())          // 4th: USB fallback
+        .with_hotkey(0x16)                       // F12
+        .with_hotkey_device(usb_device_path())   // Hotkey: boot from USB
+        .with_failure_handler(|| show_error_screen("All boot options exhausted"))
+)));
+```
+
+### UEFI-Compliant Configuration (with discover_boot_options)
+
+Platforms requiring full UEFI compliance use `discover_boot_options()` to read boot
+options from `Boot####`/`BootOrder`/`BootNext` variables instead of hardcoding device paths.
+A custom `BootOrchestrator` calls the helper at boot time:
+
+```rust
+use patina_boot::{BootDispatcher, helpers};
+
+struct UefiCompliantBoot;
+
+impl BootOrchestrator for UefiCompliantBoot {
+    fn execute(&self, boot_services: &StandardBootServices,
+               runtime_services: &StandardRuntimeServices,
+               dxe_services: &dyn DxeDispatch,
+               image_handle: efi::Handle) -> Result<!, EfiError> {
+        let config = helpers::discover_boot_options(runtime_services)?;
+        // Use config.devices() for boot attempts...
+    }
+}
+
+add.component(BootDispatcher::new(UefiCompliantBoot));
+```
+
+## Rust Code Design
+
+### Component Architecture
+
+This RFC provides components and a trait that transition firmware from initialization to OS execution.
+Each piece operates independently and can be replaced without modifying the others.
+
+**BootOrchestrator** (trait): Defines the boot orchestration interface. Platforms provide
+implementations with custom boot flows. This is a plain Rust trait (not a Patina service) because
+there is only one consumer (`BootDispatcher`) — service indirection would add complexity without
+benefit.
+
+```rust
+pub trait BootOrchestrator: Send + Sync + 'static {
+    fn execute(
+        &self,
+        boot_services: &StandardBootServices,
+        runtime_services: &StandardRuntimeServices,
+        dxe_services: &dyn DxeDispatch,
+        image_handle: efi::Handle,
+    ) -> Result<!, EfiError>;
+}
+```
+
+**BootDispatcher** (component): The Patina component that installs the BDS architectural protocol.
+Holds a `Box<dyn BootOrchestrator>` and invokes it when the DXE core calls the BDS entry point.
+Platforms register it with their chosen orchestrator: `BootDispatcher::new(impl BootOrchestrator)`.
+
+**SimpleBootManager**: Reference implementation of `BootOrchestrator`. Constructed from `BootConfig`
+via `SimpleBootManager::new(config)`. Implements the common boot flow (connect, signal
+events, detect hotkey, boot from device paths, handle failure).
+
+*Rationale: The trait-based design enforces the boot orchestration contract at compile time. Platforms
+swap boot behavior by providing a different `BootOrchestrator` impl to `BootDispatcher::new()`,
+requiring only a single component registration. This follows the principle that a trait is better than
+a service when there is only one consumer.*
+
+**`discover_boot_options()`** (helper function): Reads `Boot####` variables containing
+`EFI_LOAD_OPTION` structures, `BootOrder` variable, and `BootNext` variable to build a `BootConfig`.
+If `BootNext` is set, it is included as the first boot option and the variable is deleted. Takes
+`RuntimeServices` as a parameter.
+
+*Rationale: Boot option discovery is a helper function rather than a component because
+it has a single consumer and no need for framework dispatch ordering. Platforms call it
+in their orchestrator when they need UEFI-compliant boot variable support. Platforms
+diverging from UEFI compliance simply provide `BootConfig` directly to `SimpleBootManager`.*
+
+**ConsoleDiscovery** (component): Locates GOP and SimpleTextInput protocol handles, creates device
+paths, and writes `ConIn`/`ConOut`/`ErrOut` variables (NVRAM, default selection) and
+`ConInDev`/`ConOutDev`/`ErrOutDev` variables (volatile, all discovered devices each boot). Depends on
+RuntimeVariableServices and ProtocolServices.
+
+*Prerequisites:*
+
+- PCI bus enumeration must complete before GOP devices are discoverable (graphics controllers are
+  typically PCI devices)
+- USB bus enumeration must complete before USB input devices are discoverable (keyboards, mice)
+- When used with `SimpleBootManager`, these prerequisites are satisfied by the `connect_all()` call
+  that runs before console discovery
+- Platforms using `ConsoleDiscovery` standalone must ensure bus enumeration completes first
+
+*Rationale: Console setup is a component because platforms may need custom console configuration
+(serial-only, specific GOP preference, headless operation). As a component, platforms can replace
+it entirely without modifying orchestration code.*
+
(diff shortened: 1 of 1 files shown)


docs/src/rfc/text/0000-boot-manager.md
```diff
@@ -0,0 +1,707 @@
+# RFC: Boot Manager
+
+This RFC proposes a modular Boot Manager for Patina as a drop-in replacement for EDK2's BdsDxe driver. The design
```
os-d:
```suggestion
This RFC proposes a modular Boot Manager for Patina as a drop-in replacement for EDK II's BdsDxe driver. The design
```
nit: We've aligned on EDK II as the way to refer to that project, please fix up other references here

docs/src/rfc/text/0000-boot-manager.md
```diff
@@ -0,0 +1,707 @@
...
+4. **Testing Difficulties**: Global state and tight coupling make unit testing nearly impossible
+5. **Platform Customization Complexity**: Library override patterns require rebuilding entire firmware stacks
+6. **Binary Bloat**: All features compiled in whether needed or not
+
+**Compatibility Definition** (consistent with other Patina RFCs):
+
+1. **Functional Equivalence**: The Rust implementation provides all functionality required by the UEFI Boot Manager
+2. **Semantic Preservation**: Boot sequence, variable handling, and platform callbacks maintain identical behavior
+3. **UEFI Compliance**: Full adherence to UEFI 2.11 and PI 1.9 specifications (December 2024)
+4. **Platform Compatibility**: Supports all 15+ EDK2 platforms currently using BdsDxe with optional feature customization
```
os-d: Many many more platforms use EDK II than 15 (that may be the number in edk2-platforms). This line probably can be changed to something like "Supports platform extensibility to meet current scenarios"

docs/src/rfc/text/0000-boot-manager.md
```diff
@@ -0,0 +1,707 @@
...
+
+### Scope
+
+The `BootManager` component implements equivalent functionality for the following EDK2 modules and protocols:
+
+- **BdsDxe.inf** - Boot Device Selection DXE driver (EDK2's boot manager)
+- **UefiBootManagerLib** - Boot option management library
+- **PlatformBootManagerLib** - Platform customization callbacks
+- **EFI_BDS_ARCH_PROTOCOL** - BDS architectural protocol
+- **Variable Policy Protocol** interactions
```
os-d: Are you implying that the BootManager component will implement variable policy? Or just that it will interact with variable policy?

We would want variable policy to live with the variable driver. If you mean the latter, we would have to design a mechanism to interact with variable policy outside of having a Rust variable driver. That is probably worthy of a separate RFC because that is not trivial and may be throwaway work without a Rust driver.

docs/src/rfc/text/0000-boot-manager.md
```diff
@@ -0,0 +1,707 @@
...
+- **Deferred Image Load Protocol** handling
+
+**Naming Rationale**: This RFC uses "Boot Manager" rather than "BDS" (Boot Device Selection) because the component's
+responsibilities extend far beyond device selection alone. The UEFI specification itself uses "Boot Manager"
+terminology (UEFI 2.11 Section 3.1.2), and the component handles console management, variable policy, platform
+callbacks, hotkeys, OS indications, and boot execution—not just device selection. The "BDS" name is a historical EDK2
+artifact that doesn't reflect the full scope of functionality.
+
+Additional scope includes:
```
os-d: I'm a bit confused here. In the introduction, you say one of the pitfalls of BdsDxe is that it implements all these different services in one component. But then seemingly here you are saying BootManager will do the same. Would it make more sense to split these up into separate components? i.e. console device management not tie to boot option management.

docs/src/rfc/text/0000-boot-manager.md
```diff
@@ -0,0 +1,707 @@
...
+- Legacy feature removal (Lang/LangCodes, CSM support)
+
+**Flexibility Note**: The modular architecture supports a spectrum of platform complexity:
+
+- **Ultra-Minimal**: Platforms can bypass the orchestrator and services entirely, implementing hardcoded boot logic
+  directly (smallest binary size)
+- **Service-Based**: Platforms use the orchestrator with core services and optional feature flags (moderate binary
+  size)
+- **Full-Featured**: Platforms leverage all services and optional features for maximum flexibility (larger binary
+  size, but still smaller than EDK2)
```
os-d: We may not want to make size claims here because Rust code can be fairly large, even with a reduced featureset.

docs/src/rfc/text/0000-boot-manager.md
```diff
@@ -0,0 +1,707 @@
...
+
+## Goals
+
+Create a modular, type-safe, and maintainable Boot Manager implementation for Patina firmware that enables vendors to
+completely replace EDK2's BdsDxe in their platform builds.
+
+## Requirements
+
+1. **Drop-in Replacement**: Enable platforms to remove BdsDxe.inf from their build and use Patina Boot Manager
+   without modifying other firmware components
```
os-d: I would say without good reason. There may be design changes we want to push in BDS that we think make the ecosystem better, so I'm not sure that we should say 100% that we won't have different requirements.

docs/src/rfc/text/0000-boot-manager.md
```diff
@@ -0,0 +1,707 @@
...
+   - **Cons**: Unpredictable behavior, hard to debug issues, may hide configuration problems, inconsistent platform
+     behavior, unclear what "partial functionality" means
+
+3. **Configurable per platform through trait implementation**
+   - **Pros**: Platform decides policy based on requirements (server vs embedded vs client), supports both strict
+     and resilient approaches, clear ownership of error handling, enables platform-specific recovery strategies
+   - **Cons**: Inconsistent behavior across platforms, more complex to test all scenarios, platforms must understand
+     error handling implications, documentation burden
+
+**Recommendation**: Option 3 - let platforms decide error handling policy through trait implementations.
```
os-d: I don't think that we want platforms to tell us how to handle failures. I think we can architect the services in a way such that we return errors and the platform can decide what to do from there, be it call the service again with different parameters, call another service, etc.

docs/src/rfc/text/0000-boot-manager.md
```diff
@@ -0,0 +1,707 @@
...
+
+### Protocol Adapter for EDK2 Compatibility
+
+**Decision**: Implement an EDK2 protocol adapter layer that wraps Patina's service architecture.
+
+**Rationale**: For drop-in replacement, platforms must be able to remove BdsDxe.inf and add a Patina equivalent without
+modifying platform DSC files beyond the INF substitution. This requires:
+
+- Production of `EFI_BDS_ARCH_PROTOCOL` with correct GUID
+- DXE_DRIVER module type for DxeCore dispatch
```
os-d: I'm a bit confused here, this would be a Patina component, not a DXE_DRIVER, right?

docs/src/rfc/text/0000-boot-manager.md
```diff
@@ -0,0 +1,707 @@
...
+BdsDxe **calls** platform library functions:
+
+- `PlatformBootManagerWaitCallback()` → Boot countdown UI
+- `PlatformBootManagerBeforeConsole()` → Pre-console platform setup
+- `PlatformBootManagerAfterConsole()` → Post-console platform setup
+- `PlatformBootManagerUnableToBoot()` → Boot failure handling
+
+**Patina Implementation**:
+
+1. **Configuration Reading**: Patina Boot Manager reads from `BootManagerConfig` struct (populated at compile-time
```
os-d: We will want to review this BootManagerConfig very closely. We will not want to match all of EDK II's configuration options. We should have a very good reason for each configuration option we come up with.

docs/src/rfc/text/0000-boot-manager.md
```diff
@@ -0,0 +1,707 @@
...
+   from build configuration)
+2. **Platform Callbacks**: `PlatformPolicy` service trait provides callback methods that platforms implement
```
os-d: Similarly, we should be crystal clear on what platform hooks we are adding and why.

docs/src/rfc/text/0000-boot-manager.md
```diff
@@ -0,0 +1,707 @@
...
+2. **Executor**: Boot attempt execution, image loading, and error handling
+3. **PlatformPolicy**: Platform-specific callback implementations (maps to PlatformBootManagerLib)
+
+**Optional Services:**
+
+1. **Console**: ConIn/ConOut/ErrOut device connection and management, ConIn-on-demand event handling
+2. **DevicePathExpander**: Device path resolution and expansion
+3. **Language**: Platform language initialization (PlatformLang, PlatformLangCodes)
+4. **OsIndications**: OS communication variable handling (OsIndications, OsIndicationsSupport)
+5. **VariablePolicy**: Boot-related variable protection using EDKII_VARIABLE_POLICY_PROTOCOL
```
kuqin12: This does not seem an optional service.

docs/src/rfc/text/0000-boot-manager.md
```diff
@@ -0,0 +1,707 @@
...
+- `PlatformBootManagerBeforeConsole()` → Pre-console platform setup
+- `PlatformBootManagerAfterConsole()` → Post-console platform setup
+- `PlatformBootManagerUnableToBoot()` → Boot failure handling
+
+**Patina Implementation**:
+
+1. **Configuration Reading**: Patina Boot Manager reads from `BootManagerConfig` struct (populated at compile-time
+   from build configuration)
+2. **Platform Callbacks**: `PlatformPolicy` service trait provides callback methods that platforms implement
+3. **Protocol Adapter Translation**: Adapter layer maps EDK2 PCD reads to Patina config field access
```
os-d: I'm also confused on this statement. There should be no PCD reads, only the component referring to its configuration.

docs/src/rfc/text/0000-boot-manager.md
```diff
@@ -0,0 +1,707 @@
...
+- Independent testing of each service
+- Platform customization of specific services without affecting others  
+- Clear dependency management and interface contracts
+- Reduced complexity for maintenance and debugging
+
+**Services Identified:**
+
+**Core Services (required):**
+
+1. **Options**: Boot####, BootOrder, Driver####, and related variable management
```
os-d: I don't think we want to support DRIVER####. That is a bad pattern, despite being spec defined, and sane platforms do not use it.

docs/src/rfc/text/0000-boot-manager.md
```diff
@@ -0,0 +1,707 @@
...
+2. **Executor**: Boot attempt execution, image loading, and error handling
+3. **PlatformPolicy**: Platform-specific callback implementations (maps to PlatformBootManagerLib)
```
os-d: Is the intention that the platform must provide the PlatformPolicy service? So that the component depends on that being available and as such just calls to it?

docs/src/rfc/text/0000-boot-manager.md
```diff
@@ -0,0 +1,707 @@
...
+
+1. **Console**: ConIn/ConOut/ErrOut device connection and management, ConIn-on-demand event handling
+2. **DevicePathExpander**: Device path resolution and expansion
+3. **Language**: Platform language initialization (PlatformLang, PlatformLangCodes)
+4. **OsIndications**: OS communication variable handling (OsIndications, OsIndicationsSupport)
+5. **VariablePolicy**: Boot-related variable protection using EDKII_VARIABLE_POLICY_PROTOCOL
+6. **Hotkeys**: Boot shortcut key handling (Key#### variables, EFI_SIMPLE_TEXT_INPUT_EX_PROTOCOL)
+7. **Recovery**: Platform recovery boot option handling (PlatformRecovery#### variables)
+8. **DeferredImage**: Deferred image load handling on ReadyToBoot event
+9. **HwErrRec**: Hardware error record support (HwErr#### variables)
```
os-d: This seems like it could be decoupled from BootManager?

docs/src/rfc/text/0000-boot-manager.md
```diff
@@ -0,0 +1,707 @@
...
+    style VariablePolicy fill:#fce4ec,stroke:#880e4f,stroke-width:2px,color:#000
+    style Hotkeys fill:#fce4ec,stroke:#880e4f,stroke-width:2px,color:#000
+    style Recovery fill:#fce4ec,stroke:#880e4f,stroke-width:2px,color:#000
+    style DeferredImage fill:#fce4ec,stroke:#880e4f,stroke-width:2px,color:#000
+    style HwErrRec fill:#fce4ec,stroke:#880e4f,stroke-width:2px,color:#000
+```
+
+### Feature Flag Architecture
+
+**Decision**: Make optional boot manager features configurable through Cargo feature flags.
```
os-d: This is an okay design point, but again we need to be crystal clear on what features we support and why. If some platform somewhere uses a feature, that does not mean we have to support it. We should consider what existing platforms do and if that fits in with our model.

docs/src/rfc/text/0000-boot-manager.md
```diff
@@ -0,0 +1,707 @@
...
+- Platform customization of specific services without affecting others  
+- Clear dependency management and interface contracts
+- Reduced complexity for maintenance and debugging
+
+**Services Identified:**
+
+**Core Services (required):**
+
+1. **Options**: Boot####, BootOrder, Driver####, and related variable management
+2. **Executor**: Boot attempt execution, image loading, and error handling
```
kuqin12: Some platforms require boot attempts in a loop (OS -> PXE -> NMVE -> USB -> OS -> PXE -> ...), some do not. How would you propose to support that in this common service?

docs/src/rfc/text/0000-boot-manager.md
```diff
@@ -0,0 +1,707 @@
...
+```
+
+**Impact**: Ultra-minimal platforms (hardcoded boot paths, no services) can achieve the smallest binary size.
+Platforms using the full service architecture with optional features disabled will see significant size reduction
+compared to EDK2 through Rust's dead code elimination. Full-featured platforms will be smaller than EDK2 due to
+removal of legacy code and more efficient Rust compilation.
+
+### Trait-Based Platform Integration
+
+**Decision**: Use Rust traits for platform-specific customization points.
```
os-d: I would weigh whether this is better as a trait or as a service. It depends on how much functionality lives behind this. It may be a trait is better here because there will be no other consumers of it. I need to think through the scenario a bit more.

docs/src/rfc/text/0000-boot-manager.md
```diff
@@ -0,0 +1,707 @@
...
+
+**Decision**: Remove support for features deprecated in UEFI 2.0 (2006) and later.
+
+**Rationale**: Several BDS features are either long-deprecated or present significant security risks that are
+incompatible with Patina's design goals:
+
+- **Lang/LangCodes variables**: Replaced by PlatformLang/PlatformLangCodes in UEFI 2.0 (2006), deprecated for 19 years
+- **Legacy BIOS (CSM) support**: Modern UEFI-only platforms don't need 16-bit compatibility, Patina is explicitly
+  UEFI-only
+- **Option ROM dispatch**: Presents unacceptable security risks (arbitrary code execution from peripherals),
```
os-d: I may be confused here, are you saying no option ROM dispatch at all? Or a certain scenario that is removed? We definitely have to support option ROMs, despite the problems they come with.

os-d (commented):
The biggest part of BDS that I did not see enumerated in this RFC is the connect management, driving binding protocols, etc. That requires careful thought as it is another form of driver dispatch that is unique to the BDS phase.

I think that before we move this forward, we need to include what that story is.

docs/src/rfc/text/0000-boot-manager.md
makubacki, line 1: The RFC makes a lot of references to `BdsDxe`. I think that should be mostly consolidated to a single section focused on that relationship so that the remainder of the document is free to describe a new design without that coupling.

docs/src/rfc/text/0000-boot-manager.md
```diff
@@ -0,0 +1,707 @@
+# RFC: Boot Manager
+
+This RFC proposes a modular Boot Manager for Patina as a drop-in replacement for EDK2's BdsDxe driver. The design
+transforms the monolithic 1200-line EDK2 BdsDxe into a service-oriented architecture with 3 required core services
```
makubacki: The line count of `BdsDxe` itself is not that relevant here. Its practical complexity is significantly extended with iibraries like `PlatformBootManagerLib` and `UefiBootManagerLib`.

docs/src/rfc/text/0000-boot-manager.md
```diff
@@ -0,0 +1,707 @@
...
+1. **Monolithic Architecture**: Single driver handles multiple distinct responsibilities from console management to
+   boot execution, variable management, hotkeys, and platform integration
+2. **Memory Safety Vulnerabilities**: Manual pointer management, unchecked array access, and type confusion risks
+3. **Legacy Baggage**: Lang/LangCodes variables deprecated since UEFI 2.0, Option ROM dispatch security risks,
+   CSM/Legacy BIOS support bloat
+4. **Testing Difficulties**: Global state and tight coupling make unit testing nearly impossible
+5. **Platform Customization Complexity**: Library override patterns require rebuilding entire firmware stacks
+6. **Binary Bloat**: All features compiled in whether needed or not
+
+**Compatibility Definition** (consistent with other Patina RFCs):
```
makubacki: What does "consistent with other Patina RFCs" mean?

makubacki: This "Compatibility Definitions" subsection seems out of place in the "Motivation" section.

docs/src/rfc/text/0000-boot-manager.md
makubacki: My understanding is that this is saying this implemenation will interact with an installed variable policy protocol. If that's the case, the "protocol" detail is not that important and I'd update it to just say it will properly set variable policies.

docs/src/rfc/text/0000-boot-manager.md
```diff
@@ -0,0 +1,707 @@
...
+**Current Problems:**
+
+1. **Monolithic Architecture**: Single driver handles multiple distinct responsibilities from console management to
+   boot execution, variable management, hotkeys, and platform integration
+2. **Memory Safety Vulnerabilities**: Manual pointer management, unchecked array access, and type confusion risks
+3. **Legacy Baggage**: Lang/LangCodes variables deprecated since UEFI 2.0, Option ROM dispatch security risks,
+   CSM/Legacy BIOS support bloat
+4. **Testing Difficulties**: Global state and tight coupling make unit testing nearly impossible
+5. **Platform Customization Complexity**: Library override patterns require rebuilding entire firmware stacks
+6. **Binary Bloat**: All features compiled in whether needed or not
```
makubacki: Can you elaborate on what is meant by "All features compiled in whether needed or not"? That may be true depending on what's being discussed but this is very ambiguous at the moment.

docs/src/rfc/text/0000-boot-manager.md
```diff
@@ -0,0 +1,707 @@
...
+
+**Current Problems:**
+
+1. **Monolithic Architecture**: Single driver handles multiple distinct responsibilities from console management to
+   boot execution, variable management, hotkeys, and platform integration
+2. **Memory Safety Vulnerabilities**: Manual pointer management, unchecked array access, and type confusion risks
+3. **Legacy Baggage**: Lang/LangCodes variables deprecated since UEFI 2.0, Option ROM dispatch security risks,
+   CSM/Legacy BIOS support bloat
+4. **Testing Difficulties**: Global state and tight coupling make unit testing nearly impossible
+5. **Platform Customization Complexity**: Library override patterns require rebuilding entire firmware stacks
```
makubacki: If this is going to be listed as a problem, can you elaborate on what is meant by "rebuilding entire firmware stacks"? It's too ambiguous at the moment.

> Library override patterns require rebuilding entire firmware stacks

docs/src/rfc/text/0000-boot-manager.md
makubacki: Yeah, I don't think it is necessary to constrain the design to be a drop-in replacement for a specific C implementation. Instead, the compatibility point should be the UEFI Specification.

docs/src/rfc/text/0000-boot-manager.md
```diff
@@ -0,0 +1,707 @@
...
+
+BdsDxe **reads** PCDs at runtime:
+
+- `PcdFirmwareVendor`, `PcdFirmwareRevision` → System table initialization
+- `PcdPlatformBootTimeOut` → Boot timeout variable initialization
+- `PcdConInConnectOnDemand` → Console connection behavior control
+- `PcdPlatformRecoverySupport` → Recovery option enablement
+- `PcdCapsuleOnDiskSupport`, `PcdCapsuleFmpSupport` → Capsule feature flags
+
+BdsDxe **calls** platform library functions:
```
kuqin12: I was going to suggest on the protocol dependencies, but my guess is that it will be just those arch protocols, we probably cannot do much better than that.

However, this `PlatformBootManager` abstract might contain a lot of content, and some are very much in common. i.e. the signal end of dxe, dispatch deferred images. Would consider redoing the abstraction with this RFC?

docs/src/rfc/text/0000-boot-manager.md
```diff
@@ -0,0 +1,707 @@
...
+
+Create a modular, type-safe, and maintainable Boot Manager implementation for Patina firmware that enables vendors to
+completely replace EDK2's BdsDxe in their platform builds.
+
+## Requirements
+
+1. **Drop-in Replacement**: Enable platforms to remove BdsDxe.inf from their build and use Patina Boot Manager
+   without modifying other firmware components
+2. **Functional Completeness**: Provide all boot manager functionality required by UEFI 2.11 specification
+3. **Memory Safety**: Eliminate buffer overflows, use-after-free, and type confusion vulnerabilities
```
makubacki: Some of these items like "memory safety" I don't think are requirements or at least as written. For example, it is implied that if the code is written in Rust these memory safety guarantees will be present.

If there are specific statements being made about something in particular like certain C FFI interfaces or something those should be called out specifically.

docs/src/rfc/text/0000-boot-manager.md
```diff
@@ -0,0 +1,707 @@
...
+9. **Testing Support**: Enable comprehensive unit testing through dependency injection and mocking
+10. **Performance**: Match or exceed EDK2 boot times while reducing memory footprint
+11. **Documentation**: Provide clear migration guides and platform integration examples
+
+## Unresolved Questions
+
+### Error Recovery Strategies
+
+**Question**: How should the Boot Manager handle partial failures (e.g., one console device fails but others
+succeed)?
```
makubacki: I think this question is too generic. It's essentially asking how any "partial failure" is handled. "Partial failure" is too ambiguous and the scenarios are too diverse without scoping this down to what error(s) are specifically relevant.

docs/src/rfc/text/0000-boot-manager.md
```diff
@@ -0,0 +1,707 @@
...
+**Legacy Features**: Includes deprecated Lang/LangCodes support (UEFI 2.0, 2006), optional ROM dispatch, and
+CSM/Legacy BIOS compatibility on some platforms.
+
+**Platform Integration**: Uses C library callbacks (PlatformBootManagerLib) for platform-specific behavior, requiring
+platforms to provide implementations at build time.
+
+This RFC preserves all required UEFI functionality while eliminating legacy features and transforming the architecture
+to leverage Rust's safety guarantees.
+
+### EDK2 BdsDxe Analysis
```
makubacki: I don't think a "BdsDxe" analysis section is needed. This should focus on a design and only reference minor implementation points necessary in the BdsDxe implementation.

docs/src/rfc/text/0000-boot-manager.md
makubacki: This might be targeting code that can be linked into a component or a standalone .efi Rust driver build. In any case, the requirements section would be a good place to clarify integration targets.

docs/src/rfc/text/0000-boot-manager.md
```diff
@@ -0,0 +1,707 @@
...
+
+### Configuration and Platform Dependencies
+
+**Decision**: Boot Manager consumes configuration and platform callbacks rather than managing dependencies.
+
+**Rationale**: EDK2's BdsDxe driver is a *consumer* of Platform Configuration Database (PCD) values and platform
+library callbacks, not a provider or dependency manager. This pattern must be preserved for drop-in replacement
+compatibility.
+
+**EDK2 Dependency Pattern**:
```
makubacki: We generally do not want to tangle an EDK II invention (PCDs) with Patina. Please clearly call out the separation of configuration for Pure Rust builds.

docs/src/rfc/text/0000-boot-manager.md
makubacki, line 1: After having read the RFC, at a high-level these are some broader changes that I think can bring focus.

1. The goal of the RFC is torn throughout by (1) trying to support Rust features and the Patina component model and (2) being a "drop-in replace for `BdsDxe`" and bringing in C-based and EDK constructs. Clearly scoping this is important and should be addressed directly. Please clearly articulate why the scope of the RFC encompasses both of these. If the goal is to support both, you have three pieces to the design that occur - a "common design", "edk2-specific considerations like PCDs and protocols", and "Rust/Patina-specific considerations like services". At nearly every point in the RFC, clarity is needed on how these different use cases are supported by what's mentioned. In some cases, it might be obvious, there might be overlap, and in other cases not.
2. Partially due to the first point, I think the deep tethering to `BdsDxe` throughout the entire document encumbers a new design based on Patina components and services. Though, the suggestion for clarification in the first point might help this. If the goal is to support an independent "drop-in for BdsDxe" (implying the rest of Patina is not present) and a component for Patina, the RFC is going to get complex.
3. A lot of information is repeated throughout the document. Please identify and remove redundant points.

I think this document should primarily focus on what an ideal UEFI "boot manager" design looks like, why, and how it is best accomplished using Rust features and the Patina integration model.

docs/src/rfc/text/0000-boot-manager.md
kat-perez: no, the BootManager component will not implement variable policy but will update to say that it will properly set variable policies

docs/src/rfc/text/0000-boot-manager.md
kat-perez: Agree that I think I should rewrite the entire RFC based on the direction towards an ideal UEFI boot manager design. Will remove the goal of being a "drop-in" replacement for BdsDxe and focus more on the Patina integration model.

docs/src/rfc/text/0000-boot-manager.md
kat-perez: I left option ROM dispatch out of the scope of this because it occurs during PCI device enumeration in the DXE phase, before Boot Manager execution. I think it's not a required feature for an ideal Boot Manager. I think we would need an OptionRomDispatcher component instead which is outside of the scope of this RFC.

docs/src/rfc/text/0000-boot-manager.md
kat-perez: I think I can do more of an analysis of this in the "Unresolved Questions" section and propose some ideas there

docs/src/rfc/text/0000-boot-manager.md
kat-perez: Listing these out here meant that these are separate components or services. BootManager won't do the same since features are coming in as opt-in only. I'm following a model where platforms have to be explicit about the features that they want to include. I have this in the Appendix A. I'll try to add more of the references to the Appendix where I'm trying to communicate the design through the code

```
impl PlatformBootManager {
    pub fn entry_point(self, storage: &mut Storage) -> Result<()> {
        // Register default services
        storage.add_service(UefiOptions);
        storage.add_service(UefiConsole);
        storage.add_service(UefiExecutor);
        storage.add_service(SimplePlatform);
        
        // Run boot orchestrator
        let orchestrator = BootOrchestrator;
        orchestrator.entry_point(/* services injected automatically */)
    }
}
```

docs/src/rfc/text/0000-boot-manager.md
kat-perez: removed PCDs from this RFC

kat-perez:
i've changed the design quite a bit, added more goals and requirements to further constrain the design to be limited to the most fundamental Patina components that can achieve a successful boot

docs/src/rfc/text/0000-boot-manager.md
kat-perez: removed that type of language since we are assuming that we get those guarantees with Rust

docs/src/rfc/text/0000-boot-manager.md
kat-perez: i agree, it's not optional but it will not be implemented as a part of this RFC. but i further clarified the relationship of VariablePolicy service to the Boot Manager Components in the latest update

docs/src/rfc/text/0000-boot-manager.md
kat-perez: removed this type of phrasing from the RFC

docs/src/rfc/text/0000-boot-manager.md
kat-perez: update: removed these services for now for the scope of this RFC but add them to the optional features so that a separate RFC be required for justifying why some of these features would need to be components/services

kat-perez: and changed the most fundamental boot manager functionality into components instead

docs/src/rfc/text/0000-boot-manager.md
kat-perez: removed any size claims, meant for justification for vendors to adopt but removed in this RFC to reserve it for technical discussion

docs/src/rfc/text/0000-boot-manager.md
kat-perez: completely removed this question since i agree but i'm not going to explicitly document it in the RFC to keep it focused to the design

docs/src/rfc/text/0000-boot-manager.md
kat-perez: moved references mostly consolidated to Prior Art instead

docs/src/rfc/text/0000-boot-manager.md
kat-perez: in this RFC, we will solve PlatformBootManagerLib differently by leaving orchestration completely up to the platforms. There's no need for callbacks and having platforms full control over the customization eliminates the need for forking their own implementation of the orchestrator and instead developing their own if they want more functionality beyond the default.

Current proposed features:
1. StandardBootManager provides complete default behavior (console init, discovery, orchestration)
2. Platform customization through configuration and not callbacks
    - Connection strategy as config parameter (closures/traits)
    - Platform controls orchestration order when needed
    - Library functions provide building blocks
3. DXE phase orchestration → Outside boot components
    - Signaling end of DXE is DXE Core responsibility, not boot-specific
    - Deferred image dispatch listed in Future Work (can be separate component)

docs/src/rfc/text/0000-boot-manager.md
kat-perez: The RFC handles this through orchestration flexibility:

  StandardBootManager / BootOrchestrator: Single pass through BootOrder (standard UEFI behavior)
  - Process BootOrder once: OS → PXE → NVMe → USB
  - Returns error if all options fail
  
  Platforms needing custom loops: Replace orchestrator component

```
  #[derive(IntoComponent)]
  struct LoopingBootOrchestrator;

  impl LoopingBootOrchestrator {
      fn entry_point(
          self,
          var_services: Service<dyn RuntimeVariableServices>,
          image_services: Service<dyn ImageServices>,
          timing_services: Service<dyn TimingServices>,
      ) -> Result<!> {
          // Platform-specific: loop indefinitely
          loop {
              let boot_options = boot_manager::parse_boot_variables(&*var_services)?;
              let parent = image_services.get_current_image_handle()?;

              for option in boot_options {
                  // Try each option using library function
                  let _ = boot_manager::boot_from_load_option(
                      &option, parent, &*image_services, &*timing_services
                  );
              }

              // All failed, loop back and try again
              log::warn!("All boot options failed, retrying...");
          }
      }
  }
  
    // Use custom looping orchestrator
  Core::default()
      .with_component(ConsoleDiscovery)
      .with_component(BootOptionDiscovery)
      .with_component(LoopingBootOrchestrator)  // Custom: loops indefinitely
      .start()
```

docs/src/rfc/text/0000-boot-manager.md
kat-perez: removed this question and then also updated the unresolved questions

docs/src/rfc/text/0000-boot-manager.md
kat-perez: removed

docs/src/rfc/text/0000-boot-manager.md
kat-perez: removed PCD reads

docs/src/rfc/text/0000-boot-manager.md
kat-perez: I want to propose that platform does their own platform customization for BootOrchestrator component instead of doing exactly what BdsDxe does with platform callbacks

docs/src/rfc/text/0000-boot-manager.md
kat-perez: removed this section

docs/src/rfc/text/0000-patina-boot.md
```diff
@@ -0,0 +1,551 @@
...
+
+Platforms have diverse boot requirements ranging from simple embedded systems with hardcoded boot paths to
+complex datacenter servers with network boot, remote attestation, and IPMI integration. Modular boot
+components enable platforms to use reference implementations as-is, compose them selectively, or replace
+them entirely.
+
+This RFC provides three reference components for console initialization, boot option discovery, and boot
+orchestration. Platforms compose these components to create standard or custom boot flows. Library
+functions support custom implementations without code duplication.
+
```
apop5, line 73: We are missing information about events that are signaled during BDS.

Bds is responsible for the End of Dxe event, and a lot of drivers rely on this signal (https://github.com/tianocore/edk2/blob/master/StandaloneMmPkg/Core/StandaloneMmCore.c#L71)

There are other events that are typically signaled during the bds phase, and current EDK2 relies on providing plaform level drivers that link into bds to signal these events.

If we are going to be able to provide a compatible rust BDS, I think these signaled events need to be covered in the RFC so that compatability can be determine by users as they look to switch from edk2

docs/src/rfc/text/0000-boot-manager.md
kat-perez: @os-d were there specific design changes that you had in mind? The new approach i'm proposing here is having platforms design their own orchestration instead of relying on platform callbacks

@makubacki i've updated this RFC to make sure we keep the design focused and not relying on comparisons to EDK II design and no drop-in replacement requirement

docs/src/rfc/text/0000-boot-manager.md
kat-perez: removed any language like this and comparisons to EDK II BdsDxe

docs/src/rfc/text/0000-boot-manager.md
kat-perez: also removed any language like this and kept the RFC discussion focused on technical design points

docs/src/rfc/text/0000-boot-manager.md
kat-perez: i've removed this mechanism since I went a different direction with BootOrchestration

docs/src/rfc/text/0000-boot-manager.md
kat-perez: at that version of the RFC, I was referring to if someone took the default BdsDxe, they would have to still customize it to remove anything that they didn't want. was more of a point to introduce the motivation for feature flags

docs/src/rfc/text/0000-boot-manager.md
os-d: Nothing specific, I just didn't want to limit our design by previously made choices

docs/src/rfc/text/0000-patina-boot.md
kat-perez: I agree here that we should be adding some form of event signaling. I'll draft up a spec-compliant minimal BDS phase

docs/src/rfc/text/0000-boot-manager.md
makubacki: Thanks @kat-perez. If you could, please notify me in this thread when that is done and I'll revisit.

docs/src/rfc/text/0000-patina-boot.md
kat-perez: included mentions of how the End of DXE event and Ready to Boot events are handled with the BootOrchestrator component. please review

docs/src/rfc/text/0000-boot-manager.md
kat-perez: @makubacki this RFC is ready for review again

makubacki: Some additional thoughts from an offline conversation:

- It would help to mention what specifications and elements of those specifications that are in scope specifically, like the boot variables, etc.
- In general, clarifying requirements would be helpful. Requirements like to how to interact with notifications in different models C PI-based code and Patina components would clarify what is necessary to be considered.
- Make the platform integration model very clear. What needs to be done by a platform and how is it done.

One idea was to consider breaking out the core "BDS requirements per specifications and existing bootloaders" as a generic library crate that can be linked into a "wrapper boot manager" crate. That could allow work to move forward with reuse and less immediate decisions that need to be made.

makubacki: That's not in direct response to your most recent updates, just high-level feedback on earlier revisions.

kat-perez: @makubacki Updated the RFC with an enhanced requirements section with what is in scope (references to UEFI and PI spec) and added an explicit "Out of Scope" subsection

docs/src/rfc/text/0000-patina-boot.md
```diff
@@ -0,0 +1,640 @@
...
+   requirements necessitate composable components rather than monolithic implementations
+3. **Progressive disclosure**: Architecture supports both simple use cases (single composite component)
+   and complex scenarios (individual component composition with custom orchestration)
+4. **Scope**: This RFC covers console initialization alongside boot functionality because both are
+   required UEFI Boot Manager responsibilities. ConsoleDiscovery is optional for headless systems but
+   included for completeness
+
+## StandardBootManager Component
+
+`StandardBootManager` is a composite component that provides complete UEFI-compliant boot functionality
```
joschock: super-nit: naming: since this is a minimal config simple platform case, consider `SimpleBootManager` instead of `Standard` - which IMHO implies "does most of the things most standard platforms want" which I don't think is in view here.

docs/src/rfc/text/0000-patina-boot.md
```diff
@@ -0,0 +1,640 @@
...
+## Rust Code Design
+
+### Component Architecture
+
+The boot components provide three independent implementations that map to UEFI specification boot phases.
+These components transition firmware from initialization to OS execution.
+
+#### Component Decomposition
+
+The three components map to distinct phases of UEFI boot. **Console initialization** populates console
```
joschock: Maybe some commentary on why this three-way decomposition is chosen. Other decompositions could be imagined with more steps (for example, you could break out **System Bus Initialization** from **Console initialization** and stick PCI enumeration in there). I'm not saying the three-way is wrong, necessarily, but would be useful to have a rationale for why that particular decomposition is preferred.

docs/src/rfc/text/0000-patina-boot.md
```diff
@@ -0,0 +1,640 @@
...
+variables (`ConIn`, `ConOut`, `ErrOut`) for user interaction during boot. **Boot configuration**
+creates `Boot####` and `BootOrder` variables through automatic device discovery. **Boot orchestration**
+executes the Chapter 3 algorithm: processing `BootOrder`, parsing `EFI_LOAD_OPTION`, calling
+`LoadImage()`/`StartImage()`, and managing the watchdog timer. Each component operates independently
+and can be replaced without modifying the others.
+
+#### Component Descriptions
+
+**ConsoleDiscovery**: Scans for GOP and SimpleTextInput protocol handles, creates device paths, and
```
joschock: There's some pre-req's here - for example, to get most GOPs and Input devices, you'll have to start/enumerate the Pci and USB roots. But you likely want to do at _least_ PCI even if you don't have a GOP. It would be helpful to expand this section a bit more and be a little more detailed about how much connect work you plan to do here, and how much the platform needs to do as a pre-req.

docs/src/rfc/text/0000-patina-boot.md
```diff
@@ -0,0 +1,640 @@
...
+writes `ConIn`/`ConOut`/`ErrOut` variables. Depends on RuntimeVariableServices and ProtocolServices.
+
+**BootOptionDiscovery**: Discovers bootable media by connecting storage controllers and scanning file
```
joschock: Is this limited to just storage controllers? Other boot media (e.g. network, ramdisk, etc.) could in theory be described by a boot option. Would be good to call out a specific constraint on what you want to allow/support in this component if it isn't "anything I can express with a legal device path"

docs/src/rfc/text/0000-patina-boot.md
```diff
@@ -0,0 +1,640 @@
...
+(with configurable connection strategy), and BootOrchestrator. Orchestrates driver dispatch
+interleaving for complete device topology enumeration. Signals BDS phase events (EndOfDxe at entry,
+ReadyToBoot before boot attempts). Provides simplified integration for platforms requiring complete
+UEFI-compliant boot functionality with minimal configuration. Depends on RuntimeVariableServices,
+ImageServices, TimingServices, ProtocolServices, DxeServices, and EventServices (from RFC 0017).
+Supports builder pattern for customization.
+
+### Connection Strategy Pattern
+
+BootOptionDiscovery accepts a connection strategy as configuration to control which BlockIo device
```
joschock: `BlockIo` is a bit limiting. I think ultimately, you might want to look at anything with a filesystem you support regardless of the underlying storage foundation. Such a files system might be exposed via means other than an underlying `BlockIo` (though that is the most traditional way for fixed boot media).

docs/src/rfc/text/0000-patina-boot.md
```diff
@@ -0,0 +1,640 @@
...
+### Event Signaling and Phase Transitions
+
+The BDS phase signals events to coordinate firmware initialization and security policies.
+StandardBootManager handles event signaling at appropriate phase boundaries.
+
+**Event Flow:**
+
+StandardBootManager signals two required events:
+
+- `EndOfDxe` - After DXE phase completion, before console/boot discovery
```
joschock: EOD is usually signaled after PCIe connect and GOP discovery, FWIW.

docs/src/rfc/text/0000-boot-manager.md
joschock: PCIe enumeration usually occurs in BDS as a result of enumerating the graphics controller. It _can_ occur in DXE (and sometimes does) usually as a result of the legacy EDK II "connect controllers created in entry points" - but Patina doesn't support that.

docs/src/rfc/text/0000-patina-boot.md
```diff
@@ -0,0 +1,640 @@
...
+
+## Technology Background
+
+The UEFI Boot Manager (UEFI Specification 2.11 Chapter 3) operates through UEFI variables (`Boot####`, `BootOrder`,
+`ConIn`/`ConOut`/`ErrOut`) containing boot configuration and console device paths. Boot options use `EFI_LOAD_OPTION`
+structures with device paths and descriptions.
+
+To discover bootable media, the Boot Manager connects storage device controllers (via `ConnectController()`) to
+enable file system drivers and scan for boot loaders. BootOptionDiscovery accepts connection strategies as
+configuration, allowing platforms to customize which devices get connected while using the shared ProtocolServices
```
os-d: For a first implementation, I think it is fine to leave the connect controller flow to the protocol, but if we look forward to how Patina components would be implemented that interact with the connect controller flow, we wouldn't want a protocol, we'd want a service and it may look very different from the protocol.

It's fine not to define that here, if that is not in scope initially, but I would at least note that that work is deferred to a future RFC and this RFC does not support native Patina components interacting with the connect controller flow.

docs/src/rfc/text/0000-patina-boot.md
```diff
@@ -0,0 +1,640 @@
...
+mechanism.
+
+The BDS phase signals EndOfDxe after DXE completion and ReadyToBoot before boot attempts. These events coordinate
+firmware initialization and security lockdown.
+
+## Goals
+
+1. Provide UEFI spec compliant boot functionality
+2. Enable platforms to boot with zero custom code using reference components
```
os-d: This statement may read as a platform does not need any custom code to boot, i.e. just taken patina as is and you are good, which is not what you intend. Consider rewriting this as something like:

`Enable platforms to use a common boot manager without customization`

docs/src/rfc/text/0000-boot-manager.md
kat-perez: Thanks for the clarification, John. Since Patina uses explicit connection, PCI enumeration happens during BDS when we connect controllers. This means option ROM dispatch is already handled by the `interleave_connect_and_dispatch()` pattern in the RFC - when we connect PIC root bridges for console/boot discovery, option ROM FVs are exposed and dispatch() loads those drivers.

I'll update the RFC to clarify that the connect-dispatch interleaving handles option ROM drivers as part of device topology enumeration, rather than treating it as a separate concern.

docs/src/rfc/text/0000-patina-boot.md
kat-perez: The RFC supports your idea of breaking out System Bus Initialization from Console Initialization by having the platform use custom orchestration. Platform code could do something like this:

``` rust
  // Custom orchestrator - platform controls the sequence
  struct MyPlatformBoot;

  impl MyPlatformBoot {
      fn entry_point(self, ...) -> Result<!> {
          // 1. Connect PCI root bridges first
          connect_pci_roots(&protocol_services)?;

          // 2. Console setup
          ConsoleDiscovery::run(&protocol_services, &variable_services)?;

          // 3. More device connection if needed
          interleave_connect_and_dispatch(&protocol_services, &dxe_services)?;

          // 4. Boot
          BootManager::new(my_boot_source).run()?;
      }
  }
```

But will still add some rationale points for each component

docs/src/rfc/text/0000-patina-boot.md
kat-perez: Goal #2 reworded like this:
> 2. Enable platforms to use a common boot manager without customization

docs/src/rfc/text/0000-patina-boot.md
kat-perez: Updated here:
  ```diff
  - `EndOfDxe` - After device enumeration (connect-dispatch loop completes). Security components
  -   (e.g., SMM/MM) register for this event and perform lockdown (install `DXE_SMM_READY_TO_LOCK`).
  -   BootOrchestrator signals the event but does not handle lockdown—that is the responsibility of
  -   security components.
  + `EndOfDxe` - After device topology is stable (PCIe enumeration and GOP discovery complete).
  +   Security components (e.g., SMM/MM) register for this event and perform lockdown (install
  +   `DXE_SMM_READY_TO_LOCK`). BootOrchestrator signals the event but does not handle lockdown—that
  +   is the responsibility of security components.
  ```

docs/src/rfc/text/0000-patina-boot.md
kat-perez: good point. I've added an explicit "Future Work: Connect Controller Service" section noting that this is deferred and native Patina components cannot participate in the connect controller flow in this RFC.

docs/src/rfc/text/0000-patina-boot.md
kat-perez: I think the "BootOptionDiscovery" component can be inclusive of all possible types of "boot options".  I'm leaving the role of Config<BootOptions> to take device paths and just bypass the whole Boot#### / BootOrder variables.  That supports a hardcoded device path directly in config

docs/src/rfc/text/0000-patina-boot.md
kat-perez: Agreed on this. I've updated the RFC so that it no longer just specifies BlockIo or any particular storage foundation. This work is deferred to BootOptionDiscovery component

docs/src/rfc/text/0000-patina-boot.md
kat-perez: Expanded the RFC to address this:

  1. ConsoleDiscovery prerequisites now explicit:
    - PCI bus enumeration required for GOP devices
    - USB bus enumeration required for input devices
    - When used with BootOrchestrator, these are satisfied by connect-dispatch interleaving

  2. BootOrchestrator enumeration behavior clarified:
    - Connects all handles recursively, dispatches until topology stabilizes
    - This is a "connect all, dispatch all" reference implementation

  3. Platform customization options documented with example:

StandardBootOrchestrator is framed as a reference implementation. Production platforms typically implement custom orchestrators for optimized boot times. Added concrete example showing selective PCI-only enumeration:

``` rust
  #[component]
  impl CustomBootOrchestrator {
      fn entry_point(self, protocol_services: Service<dyn ProtocolServices>, ...) -> Result<!> {
          // Selective: Connect only PCI roots (skip USB for headless system)
          let pci_handles = protocol_services.locate_handle_buffer(
              HandleSearchType::ByProtocol(&PCI_ROOT_BRIDGE_IO_PROTOCOL_GUID)
          )?;
          for handle in pci_handles {
              protocol_services.connect_controller(handle, true)?;
          }
          while dxe_services.dispatch()? {}
          // ...
      }
  }
 ```

File: docs/src/rfc/text/0000-patina-boot.md (hunk 30 above)
Comment (kat-perez) [reply docs/src/rfc/text/0000-patina-boot.md]: I've changed two things I'm using `BootOrchestrator` to be the full standard implementation and changed this to `SimpleBootOrchestrator` to refer to a minimal platform solution (also serves as a sample of a custom `BootOrchestrator`)

Reframed:
  - Before: "reference implementation... suitable for development and simple platforms. Production platforms typically implement custom"
  - After: "standard implementation... suitable for most platforms. Platforms with specific requirements can implement custom"

File: docs/src/rfc/text/0000-boot-manager.md (hunk 9 above)
Comment (kat-perez) [reply docs/src/rfc/text/0000-boot-manager.md]: For the first iteration, I have a minimal set of `Config<BootOptions>` with only:
  - devices - list of device paths
  - hotkey - optional scancode
  - hotkey_devices - alternate boot options
  - failure_handler - callback

File: docs/src/rfc/text/0000-boot-manager.md (hunk 18 above)
Comment (kat-perez) [reply docs/src/rfc/text/0000-boot-manager.md]: Ended up doing a new approach: 1. BootOrchestrator is a component, not a trait or a service. Platform customization happens via Config<BootOptions>. Platforms needing entirely different boot flows replace the component entirely.

File: docs/src/rfc/text/0000-patina-boot.md [hunk 38]
```diff
@@ -0,0 +1,510 @@
...
+- Connection filtering and policy components
+- Custom binding logic for platform-specific device handling
+
+Until that RFC is implemented, platforms requiring custom connect controller behavior must use
+UEFI Driver Model protocols or implement custom orchestrators that call `connect_controller()`
+directly with platform-specific logic.
+
+## Unresolved Questions
+
+### BDS Phase Dispatch Mechanism
```
kat-perez: @os-d @makubacki @joschock @Javagedes, added this new question in "Unresolved Questions" that I would like feedback on please

docs/src/rfc/text/0000-patina-boot.md
```diff
@@ -0,0 +1,510 @@
...
+
+These requirements are about *what* happens, not *how* it's dispatched. `BootOrchestrator` fulfills
+all functional requirements regardless of which dispatch mechanism invokes it.
+
+C drivers that register for BDS events (EndOfDxe, ReadyToBoot) via `CreateEventEx()` continue to
+work because `BootOrchestrator` signals these events. The BDS protocol itself is not needed for
+event signaling.
+
+For platforms transitioning from C-based BDS, `call_bds()` could be retained as a fallback that
+only executes if no Patina boot component ran.
```
os-d: I think that another way to approach this is to have an extremely minimal BDS component that just calls the BDS arch protocol. So, if a platform starts with not using the `BootOrchestrator` component, it can use the `DummyBdsComponent` or whatever which will just call the protocol to engage `BdsDxe`. Then Patina can be agnostic of what owns BDS, it simply has one way of initiating the handoff (kind of) to BDS.

docs/src/rfc/text/0000-patina-boot.md
os-d: I think option 2, or a variation on it (probably need to see what the BdsPhase service looks like and if there is another version of doing that) makes the most sense, i.e. I don't think we need the DXE core to call the BDS arch protocol. Patina should have its own mechanism that it decides is the best and idiomatic and we can adjust things as needed. I've left a comment below on a thought there.

joschock: I'm agnostic on the dispatch mechanism; but one thing I _do_ think is important is enforcing a strong separation between BDS component and the core.  In the traditional EDK2 approach, BDS is a separate driver (hence the need for the arch protocol to provide a consistent way to invoke that driver), and that means that BDS and the core cannot easily share internal state; state must be communicated via well-defined mechanisms (protocols, tables, etc). So if we do move to a component dispatch model (and I like @os-d's idea about supporting the existing arch protocol with a component wrapper that calls it), that that doesn't cause things to "leak" through the boundary between the BDS component and the core. I understand that the component arch is designed to allow that separation with a service/parameter model; but we need to be disciplined about that interface.

kat-perez: @os-d @joschock Thanks for the feedback. I like the idea of having a minimal BDS component that simply calls the BDS arch protocol - it keeps Patina agnostic of what owns BDS and provides a clean migration path. What if we did something like this?
```
  /// Minimal component that bridges to existing BDS arch protocol
  #[component]
  impl BdsArchProtocolBridge {
      fn entry_point(self, protocol_services: Service<dyn ProtocolServices>) -> Result<()> {
          let bds_protocol = protocol_services.locate_protocol::<bds::Protocol>()?;
          unsafe { ((*bds_protocol).entry)(bds_protocol) };
          Ok(())
      }
  }
```

docs/src/rfc/text/0000-patina-boot.md
kat-perez: That sounds like a good approach to me. I have a proposal in the other thread

docs/src/rfc/text/0000-patina-boot.md
os-d: I think that looks reasonable

kat-perez: Thanks @os-d. @joschock - does this approach address your concern about maintaining strong separation between BDS and core? The bridge component is stateless and communicates only through ProtocolServices - no internal state leakage.

I'll update the RFC to document both integration paths:

1. Native Patina: BootOrchestrator component (no protocol indirection)
2. Legacy/hybrid: BdsArchProtocolBridge → existing C-based BdsDxe

This keeps the BDS dispatch mechanism as a platform choice rather than a Patina core decision.

joschock: Yep - that looks reasonable. My main concern is that patina doesn't become tightly coupled with the boot-orchestrator. building one that boots legacy BDS should keep us honest :)

docs/src/rfc/text/0000-patina-boot.md
```diff
@@ -0,0 +1,735 @@
...
+    BDS architectural protocol installation, import paths
+  - Removed Phase 1/Phase 2 split — core framework fully implemented
+  - `BootOptionDiscovery` implemented as `discover_boot_options()` helper function
+    instead of a separate component (simpler, no framework machinery needed)
+  - `interleave_connect_and_dispatch()` made private to `SimpleBootManager`
+    (uses `connect_all()` which is not recommended for most platforms)
+
+## Motivation
+
+Most Patina platforms require UEFI-compliant boot functionality: discovering boot options from `Boot####`/`BootOrder`
```
rogurr: Not sure how detailed you want to be here.  There is a BootNext in addition to the Boot#### and BootOrder.  It's a UINT16 to force a specific path on the next reset.

docs/src/rfc/text/0000-patina-boot.md
```diff
@@ -0,0 +1,735 @@
...
+- Read `BootOrder` variable to determine boot attempt sequence
+- Manage boot option list (add, remove, reorder boot options)
+- Support boot option attributes (LOAD_OPTION_ACTIVE, LOAD_OPTION_HIDDEN, etc.)
+- Expand partial device paths to full device paths by matching against discovered device topology
+  (e.g., hard drive media device paths containing only partition GUID/signature)
+
+**Console Variables**:
+
+- Discover console devices via GOP (graphics) and SimpleTextInput protocols
+- Populate `ConIn`, `ConOut`, and `ErrOut` variables with console device paths
```
rogurr: If you are discussing the ConIn/ConOut/ErrOut variables, you might want to also include the ConInDev/ConOutDev/ErrOutDev variables that list all found options each boot.

https://github.com/tianocore/edk2/blob/87e486f622c5eaa0bee04f090b5c543466d6178b/MdePkg/Include/Guid/GlobalVariable.h#L75

The non 'dev' vars are usually the default selection and are saved in NVRAM.  The 'dev' vars are usually volatile and created each boot upon discovery.

docs/src/rfc/text/0000-patina-boot.md
```diff
@@ -0,0 +1,735 @@
...
+The following features are natural extensions of the current implementation and may be
+added as helper functions or orchestrator enhancements:
+
+- **Boot menu UI**: Boot timeout countdown and boot option selection UI, using
+  options from `discover_boot_options()`. Platforms with GOP consoles could display
+  a menu; headless platforms could use serial console.
+- **`Key####` hotkey variables**: Full UEFI-compliant hotkey management via `Key####`
+  variables, extending the current scancode-based `detect_hotkey()` to read hotkey
+  configuration from UEFI variables.
+- **Platform recovery options**: `PlatformRecovery####` variable support for
```
rogurr: Could this be expanded to maybe be a hierarchy mechanism with a fallover?  Try option 1, if fails, try option 2, if fails, try option 3, etc.?  We talked about the recovery option which would be where we just have 2 options, but current platforms have several options; Windows boot path, Linux boot path, UEFI shell, PXE boot, USB boot, front page, etc.

docs/src/rfc/text/0000-patina-boot.md
```diff
@@ -0,0 +1,735 @@
...
+device topology before they can be used for booting.
+
+Partial paths are detected by examining the first device path node:
+
+- **Full paths** start with hardware/ACPI root nodes (`PciRoot`, `Acpi`)
+- **Partial paths** start with media nodes (`HD`, `CDROM`), messaging nodes without root, or file paths
+
+Both `expand_device_path()` and `boot_from_device_path()` are provided to support different use cases:
+
+| Function                   | Use Case                                                                                                                                 |
```
rogurr: Minor, if you want to make tables a little more readable in edit mode, you don't need to make each line the same length.

This text:
``` text
| Title 1 Left Justified | Title 2 Centered | Title 3 Right Justified |
|---|:---:|---:|
| a | b | c |
| d | e | f |
```

Produces this:
| Title 1 Left Justified | Title 2 Centered | Title 3 Right Justified |
|---|:---:|---:|
| a | b | c |
| d | e | f |

The separator only needs 3 dashes and you can put a colon ':' on both sides to center the column or on the right to right align.

rogurr (approved):

docs/src/rfc/text/0000-patina-boot.md
kat-perez: updated

docs/src/rfc/text/0000-patina-boot.md
kat-perez: Added BootNext

docs/src/rfc/text/0000-patina-boot.md
kat-perez: Added those too

docs/src/rfc/text/0000-patina-boot.md
kat-perez: Yeah, that makes sense. I've updated it to describe a fallback chain supporting arbitrary depth

makubacki (approved):
Thank you for all of your work on this design and implementation. Approving the update to move it to rejected status.

os-d:
Thanks for all the work on this! Can you add the rejection summary section as outlined here: https://github.com/OpenDevicePartnership/patina/blob/main/docs/src/rfc_lifecycle.md#rejected-rfcs

@makubacki we may want to add this to the template, or have some example of format. I like the version in the ESRT RFC that was rejected, with a summary of why as well as future considerations if the topic is revisited. There may be more we want to standardize.

os-d:
Also, there is already an RFC 27, can you name this RFC 28?

kat-perez:
@makubacki @os-d I added the "Rejected Summary" section to the end of the RFC and bumped up the RFC number to 28. Ready for review again

os-d (approved):

makubacki:
Note: Resolved all remaining open conversations to merge the RFC in the rejected state.

PR #1014, opened 2025-11-04
patina_dxe_core/gcd: Skip setting mem attrs for V1 MMIO and Reserved HOBs

### LINKED ISSUE #1012: [Bug]: Physical platforms with V1 HOBs assert

### Current Behavior

As of Patina v14.2.0, Intel physical platforms that are using V1 resource descriptor HOBs (with the `v1_resource_descriptor_support` feature) hit the following assertion:

```
INFO - Mapping memory range 0xfeda3000..0xfeda4000 as MemoryMappedIo with attributes 0x403
ERROR - panicked at C:\Users\VssAdministrator\.cargo\registry\src\index.crates.io-1949cf8c6b5b557f\patina_dxe_core-14.2.0\src\gcd.rs:234:25:
GCD failed to set memory attributes 0x0 for base: 0xFEDA3000, length: 0x1000
```

This is suspected to be a result of the changes in commits:

- [Cleanup Resc Desc HOB v1 Feature](https://github.com/OpenDevicePartnership/patina/commit/652eb42063626ba7e5a6dd610bddd68d0bd1c981)
- [Feature : Convert ResourceDescriptor V1 HOBs to V2 format with cache attributes](https://github.com/OpenDevicePartnership/patina/commit/311368b469410e8aaddc40adf1354f5d9f3c9f04)

### Expected Behavior

MMIO resource descriptor HOBs are parsed successfully when the `v1_resource_descriptor_support` feature is enabled.

### Steps To Reproduce

1. Build Patina DXE Core with patina v14.2.0
2. Enable the `v1_resource_descriptor_support` feature
3. Produce a MMIO resource dscriptor V1 HOB

### Build Environment

```markdown
N/A
```

### Version Information

```text
Tag: v14.2.0
```

### Urgency

Medium

I will fix it

No maintainer feedback needed

### Anything else?

_No response_

makubacki:
## Description

Closes #1012 

Currently, the memory attributes for MMIO resource descriptor V1 HOBs
is being calculated as `0` (since no cache attributes are set).

This causes an assert:

```
GCD failed to set memory attributes 0x0 for base: 0xFEDA3000,
length: 0x1000
```

This change skips calling `set_memory_space_attributes()` for V1
resource descriptor HOBs with `0` memory attributes which matches
previous behavior for V1 HOBs.

---

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?
- [ ] Creates a new crate?

## How This Was Tested

- `cargo make all`
- Verified a physical platform with V1 HOBs no longer asserts on MMIO or reserved HOBs that have attributes set to `0` during HOB list processing.

## Integration Instructions

- N/A

### DIFF
--- a/patina_dxe_core/src/gcd.rs
+++ b/patina_dxe_core/src/gcd.rs
@@ -218,6 +218,17 @@ pub fn add_hob_resource_descriptors_to_gcd(hob_list: &HobList) {
                     )
                     .expect("Failed to add memory space to GCD");
                 }
+
+                // V1 HOBs don't provide cache attributes for MMIO/Reserved memory. Skip setting attributes.
+                #[cfg(feature = "v1_resource_descriptor_support")]
+                if memory_attributes == 0 {
+                    log::debug!(
+                        "Skipping setting memory attributes for {gcd_mem_type:?} range {:#x?} with 0 attributes (V1 HOB)",
+                        split_range
+                    );
+                    continue;
+                }
+
                 match GCD.set_memory_space_attributes(
                     split_range.start as usize,
                     split_range.end.saturating_sub(split_range.start) as usize,


patina_dxe_core/src/gcd.rs
```diff
@@ -218,6 +218,19 @@ pub fn add_hob_resource_descriptors_to_gcd(hob_list: &HobList) {
                     )
                     .expect("Failed to add memory space to GCD");
                 }
+
+                // V1 HOBs don't provide cache attributes for MMIO/Reserved memory. Skip setting attributes.
+                #[cfg(feature = "v1_resource_descriptor_support")]
+                if memory_attributes == 0
+                    && (gcd_mem_type == GcdMemoryType::MemoryMappedIo || gcd_mem_type == GcdMemoryType::Reserved)
```
os-d: Do we need this conditional? For v1 previously, we were always skipping the attribute setting. `memory_attributes` should always be 0, but I didn't go through the path to make sure that other GCD memory types won't hit here

makubacki: My understanding when quickly looking at the logic earlier was that this prevented the issue for the other ranges:

```rust
            if gcd_mem_type == GcdMemoryType::SystemMemory {
                // Force all system memory to be RP by default (since none is allocated yet)
                memory_attributes |= efi::MEMORY_RP;
            }
```

I haven't verified that though. But this does work as-is on physical platforms. I'll defer to your decision for what to end up doing.

os-d: Mentioned offline, but my main concern is if a platform came with a persistent memory region, we'd have zero mem attributes and then hit the same issue. So I think this check could just be if memory_attributes == 0

makubacki: Updated to just check for `memory_attributes == 0`.

PR #1022, opened 2025-11-06
BugFix: Init GCD/System Table and others before running tests

vineelko:
## Description

BugFix: Init GCD/System Table and others before running tests

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?
- [ ] Creates a new crate?

## How This Was Tested

`cargo nextest`

## Integration Instructions

NA

### DIFF
--- a/patina_dxe_core/src/misc_boot_services.rs
+++ b/patina_dxe_core/src/misc_boot_services.rs
@@ -246,7 +246,10 @@ pub fn init_misc_boot_services_support(bs: &mut efi::BootServices) {
 #[coverage(off)]
 mod tests {
     use super::*;
-    use crate::{systemtables, test_support};
+    use crate::{
+        systemtables::{self, EfiSystemTable},
+        test_support,
+    };
     use core::{ffi::c_void, ptr};
     use r_efi::efi;
     use std::cell::UnsafeCell;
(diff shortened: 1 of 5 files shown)


patina_dxe_core/src/misc_boot_services.rs
```diff
@@ -274,22 +274,35 @@ mod tests {
         unsafe { core::mem::transmute(bs) }
     }
 
+    fn with_locked_state<F: Fn() + std::panic::RefUnwindSafe>(f: F) {
```
makubacki: The `with_locked_state` implementation in image.rs calls `init_system_table()` internally. Are the disjoint implementations necessary? It seems confusing to have the function have overlapping but ultimately different behavior in different modules.

https://github.com/vineelko/patina/blob/598815231e50afa93f37df6107efa4abec3990f1/patina_dxe_core/src/image.rs#L1545-L1554

patina_dxe_core/src/runtime.rs
```diff
@@ -270,15 +282,14 @@ mod tests {
                 }
                 assert_eq!(count, 5, "Not all entries were found in the image list.");
             }
-        })
-        .unwrap_or_else(|e| panic!("Test failed with runtime allocator conflict: {:?}", e));
```
makubacki, line 274: I think the custom panic messages were relatively more useful than a generic `unwrap()`. Up to you if you want to support this in `with_locked_state()`.

patina_dxe_core/src/misc_boot_services.rs
makubacki: Since everything internal to the function is coming from the `test_support` module, can a single implementation of `with_locked_state()` just be made available from there?

makubacki: Also, in the current implementation in `misc_boot_services`, it seems unnecessary to acquire the system table lock, release it, and immediately acquire it again.

For example:

```rust
        // ...
        with_locked_state(|| {
            systemtables::init_system_table();
            let mut st_guard = systemtables::SYSTEM_TABLE.lock();
            // ...
```

Where `init_system_table()` is:

```rust
pub fn init_system_table() {
    let mut table = EfiSystemTable::init();
    table.checksum();
    _ = SYSTEM_TABLE.lock().insert(table);
}
```

So, my understanding is:

- Acquire and release in `init_system_table()` due to:
  - `SYSTEM_TABLE.lock().insert(table);`
- Acquire again the test function (`f()`), due to:
  - `let mut st_guard = systemtables::SYSTEM_TABLE.lock();`

If the system table lock and the test function are both called within the same block in `with_global_lock()`, can it simply acquire the system table lock once and pass it to the test function?

Something like this:

```rust
let mut st_guard = systemtables::SYSTEM_TABLE.lock();
let st = st_guard.as_mut().expect("System Table not initialized!");
f(st);
```

Since this is only needed for tests that actually use the system table, it could simply wrap the single instance of `with_locked_state()` (if it is created) so those tests could use a "system table variant".

Something like this:

```rust
fn with_system_table<F>(f: F) 
where 
    F: FnOnce(&mut EfiSystemTable) + std::panic::RefUnwindSafe 
{
        unsafe {
            crate::test_support::init_test_gcd(None);
            crate::test_support::init_test_protocol_db();
        }
        crate::test_support::reset_dispatcher_context();
        crate::systemtables::init_system_table();
        
        let mut st_guard = systemtables::SYSTEM_TABLE.lock();
        let st = st_guard.as_mut().expect("System Table not initialized!");
        f(st);
    })
    .unwrap();
}
```

Then they could just call that more cleanly and avoid boilerplate and have a system table that's initialized and ready to use.

For example:

```rust
#[test]
fn test_misc_calc_crc32() {
    with_system_table(|st| {
        initialize_boot_services(unsafe { get_static_boot_services(st.boot_services_mut()) });
        // ...
    });
}
```

vineelko: Currently, the `with_locked_state()` implementations are tailored to the module running the tests. So there are differences between them, and I’m not sure we can create a single consolidated version of this function in `test_support`. I did try to come up with such a function, but it is becoming more complicated than what we have hence leaving the function to stay in their respective test modules.

The system table case has been addressed.

patina_dxe_core/src/runtime.rs
vineelko: The `std::panic::catch_unwind()` inside the `with_global_lock()` callback will have the stack trace containing the failure function/line. It does not seem to add much value here. Hence, I removed them.

makubacki: Sounds good. That's reasonable.

makubacki (approved):

patina_dxe_core/src/misc_boot_services.rs
makubacki: No problem. Thanks for adding the system table function.

Javagedes (approved):

cfernald (approved):

PR #1032, opened 2025-11-07
dxe_core: Always Apply NX to UC Memory

os-d:
## Description

This commit adds a new memory rules function that is easy to audit the rules that Patina applies to memory. For now, this function consists of a single rule, which is if EFI_MEMORY_UC is set on a range, EFI_MEMORY_XP must also be set.

In DXE, we should never be executing from UC memory and on AArch64 it is architecturally defined as a programming error to have device memory (what UC maps to) that is not marked XP. EDK II's ArmCpuDxe always sets XP when UC is requested.

It has been observed that speculative execution on AArch64 platforms can attempt to do an instruction fetch from device memory (which is allowed if XP is not set) and crash the system if the region touched is MMIO that does not have backing memory (say space between device registers).

For all architectures, Patina enforces that UC memory should not be executed from.

This commit also makes the XP requirement for all added memory more explicit in adding resource descriptor HOBs, though this is not a functional change.

- [x] Impacts functionality?
- [x] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested by booting to Windows on Q35, SBSA, and physical ARM64 and Intel platforms

## Integration Instructions

N/A.

### DIFF
--- a/patina_dxe_core/src/gcd.rs
+++ b/patina_dxe_core/src/gcd.rs
@@ -193,8 +193,9 @@ pub fn add_hob_resource_descriptors_to_gcd(hob_list: &HobList) {
 
         if gcd_mem_type != GcdMemoryType::NonExistent {
             // Extract cache attributes and add ReadProtect for system memory.
-            // If we are processing V1 HOBs, the cache attributes will be 0; that information is not passed.
-            let mut memory_attributes = cache_attributes & efi::CACHE_ATTRIBUTE_MASK;
+            // Always set NX attribute for all memory regions. We will make the DXE Core code sections explicitly
+            // executable later. Any memory that needs to be executable must be explicitly set as such.
+            let mut memory_attributes = (cache_attributes & efi::CACHE_ATTRIBUTE_MASK) | efi::MEMORY_XP;
             if gcd_mem_type == GcdMemoryType::SystemMemory {
                 // Force all system memory to be RP by default (since none is allocated yet)
                 memory_attributes |= efi::MEMORY_RP;
@@ -217,16 +218,6 @@ pub fn add_hob_resource_descriptors_to_gcd(hob_list: &HobList) {
                         spin_locked_gcd::get_capabilities(gcd_mem_type, resource_attributes as u64),
                     )
                     .expect("Failed to add memory space to GCD");
-                }
-
-                // V1 HOBs don't provide cache attributes for MMIO/Reserved memory. Skip setting attributes.
-                #[cfg(feature = "v1_resource_descriptor_support")]
-                if memory_attributes == 0 {
-                    log::debug!(
-                        "Skipping setting memory attributes for {gcd_mem_type:?} range {:#x?} with 0 attributes (V1 HOB)",
-                        split_range
-                    );
-                    continue;
                 }
 
                 match GCD.set_memory_space_attributes(
--- a/patina_dxe_core/src/gcd/spin_locked_gcd.rs
+++ b/patina_dxe_core/src/gcd/spin_locked_gcd.rs
@@ -2415,6 +2415,19 @@ impl SpinLockedGcd {
         result
     }
 
+    // Apply Patina memory protection rules to the given attributes.
+    //
+    // Rules:
+    //   - If UC memory is requested, always set XP memory.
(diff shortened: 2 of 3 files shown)


patina_dxe_core/src/gcd/spin_locked_gcd.rs
```diff
@@ -2415,6 +2415,22 @@ impl SpinLockedGcd {
...
+    // Apply Patina memory protection rules to the given attributes.
+    //
+    // Rules:
+    //   - If UC memory is requested, always set XP memory.
+    fn apply_memory_protection_rules(attributes: u64) -> u64 {
+        let mut new_attributes = attributes;
+
+        // If we have UC memory, we always set XP. In DXE, we should never be executing from UC memory. On
+        // AArch64, this is defined as a programming error to have executable device memory (which UC maps to).
+        if attributes & efi::MEMORY_UC == efi::MEMORY_UC {
```
Javagedes, line 2425: Do you suspect there will be many more rules? If so, a more generic interface may be better:

``` rust
trait MemoryProtectionRule {
    fn apply(attributes: &mut u64)
}

/// A MemoryProtectionRule that will set XP memory if UC memory is requested
pub struct UcSetXp;

impl MemoryProtectionRule for UcSetXp {
    #[inline(always)]
    fn apply(attributes: &mut attributes) {
        if attributes & efi::MEMORY_UC == efi::MEMORY_UC {
            attributes |= efi::MEMORY_XP
        }
    }
}

fn apply_memory_protection_rules(mut attributes: u64) -> u64 {
    UcSetXp::apply(&mut attributes);
    AnotherRule::apply(&mut attributes);
    AndAnother::apply(&mut attributes);
    attributes
}
```

patina_dxe_core/src/gcd/spin_locked_gcd.rs
```diff
@@ -2415,6 +2415,22 @@ impl SpinLockedGcd {
         result
     }
 
+    // Apply Patina memory protection rules to the given attributes.
+    //
+    // Rules:
+    //   - If UC memory is requested, always set XP memory.
+    fn apply_memory_protection_rules(attributes: u64) -> u64 {
+        let mut new_attributes = attributes;
```
Javagedes: Compiler is probably smart enough not to create a stack variable, but just to be sure, you can just put mut on the function input, update it, and return it

```rust
fn apply_memory_protection_rules(mut attributes: u64) -> u64 {

    if attributes & efi::MEMORY_UC == efi::MEMORY_UC {
        attributes |= efi::MEMORY_XP
    }

    attributes
}
```

Javagedes (approved):

makubacki (approved):

patina_dxe_core/src/gcd/spin_locked_gcd.rs
os-d: At this point, no. There was a request to have this in a separate function for easy auditability, hence having it here. I will probably rearchitect this at some point because there are other places we apply policy, but it is not trivial to colocate them, so I need to think about how to do that more.

patina_dxe_core/src/gcd.rs
makubacki, line 1: Do you think this is worth calling out in https://opendevicepartnership.github.io/patina/integrate/patina_dxe_core_requirements.html?

This should have already been done by platforms, but explicitly mentioning NX behavior might still be helpful to include.

os-d: Yeah, good call, I will update that.

patina_dxe_core/src/gcd/spin_locked_gcd.rs
os-d: Actually, I do like the trait concept a lot, but in a different way, which perhaps you meant, but I can have one trait and then everywhere across the core that applies some memory protection policy can call a function in the trait, so that way it is easily auditable. I won’t do that in this PR but as a follow on.

patina_dxe_core/src/gcd.rs
os-d: Added a section.

PR #1033, opened 2025-11-07
GCD: Move Memory Block Init to Separate Function

os-d:
## Description

Currently, when the PHIT HOB is processed to update the GCD, it calls add_memory_space() which sees the GCD memory blocks capacity is 0, so it calls init_memory_blocks, which creates capacity in the memory blocks and then calls back into add_memory_space to add the region to the GCD, which this time sees the capacity is non-zero and so does not call init_memory_blocks again.

This commit breaks the call to init_memory_blocks out into a separately callable function that is called when the PHIT HOB is processed. This is done to simply this call stack, most importantly because it has been observed that speculative execution can mispredict the capacity == 0 branch for other add_memory_space calls and do speculative reads of MMIO (when it believes that this memory region will be the one that is added to initialize the memory blocks) that on some platforms can cause stuck transactions (e.g. if some of the MMIO isn't back by an actual device or memory).

It is the responsibility of the platform to correctly report its memory resources, but, we can be a bit more defensive here since an active issue was seen and this pattern was convoluted anyway.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on Q35, SBSA, and a physical ARM64 platform.

## Integration Instructions

N/A.

### DIFF
--- a/patina_dxe_core/src/gcd/spin_locked_gcd.rs
+++ b/patina_dxe_core/src/gcd/spin_locked_gcd.rs
@@ -337,7 +337,7 @@ impl GCD {
         self.maximum_address = 1 << processor_address_bits;
     }
 
-    unsafe fn init_memory_blocks(
+    pub(crate) unsafe fn init_memory_blocks(
         &mut self,
         memory_type: dxe_services::GcdMemoryType,
         base_address: usize,
@@ -421,6 +421,7 @@ impl GCD {
         ensure!(self.maximum_address != 0, EfiError::NotReady);
         ensure!(len > 0, EfiError::InvalidParameter);
         ensure!(base_address.checked_add(len).is_some_and(|sum| sum <= self.maximum_address), EfiError::Unsupported);
+        ensure!(self.memory_blocks.capacity() > 0, EfiError::NotReady);
 
         log::trace!(target: "allocations", "[{}] Adding memory space at {:#x}", function!(), base_address);
         log::trace!(target: "allocations", "[{}]   Length: {:#x}", function!(), len);
@@ -435,9 +436,6 @@ impl GCD {
             capabilities |= efi::MEMORY_ISA_VALID;
         }
 
-        if self.memory_blocks.capacity() == 0 {
-            return unsafe { self.init_memory_blocks(memory_type, base_address, len, capabilities) };
-        }
         let memory_blocks = &mut self.memory_blocks;
 
         log::trace!(target: "gcd_measure", "search");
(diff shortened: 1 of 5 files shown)


patina_dxe_core/src/gcd/spin_locked_gcd.rs
```diff
@@ -1853,6 +1851,23 @@ impl SpinLockedGcd {
         }
     }
 
+    /// Initializes the memory blocks in the GCD.
+    ///
+    /// # Safety
+    /// The caller must ensure that the memory region specified by `base_address` and `len` is valid and not
+    /// currently in use by any other part of the system.
```
joschock: super-nit: might be good to articulate what `valid` means (I think it means "freely usable RAM"). Also, it isn't just "currently not in use" but also "will not be used by any other part of the system at any time" - otherwise you get aliasing.

patina_dxe_core/src/gcd/spin_locked_gcd.rs
```diff
@@ -435,9 +436,6 @@ impl GCD {
             capabilities |= efi::MEMORY_ISA_VALID;
         }
```
Javagedes, line 438: the io_blocks portion of the GCD does the same thing, should you update that too?

e.g. `add_io_space` will also call `self.init_io_blocks()` if `io_blocks.capacity() == 0`

patina_dxe_core/src/gcd/spin_locked_gcd.rs
```diff
@@ -421,6 +421,7 @@ impl GCD {
         ensure!(self.maximum_address != 0, EfiError::NotReady);
         ensure!(len > 0, EfiError::InvalidParameter);
         ensure!(base_address.checked_add(len).is_some_and(|sum| sum <= self.maximum_address), EfiError::Unsupported);
+        ensure!(self.memory_blocks.capacity() > 0, EfiError::NotReady);
```
makubacki, line 424: Is a unit test already checking the `self.memory_blocks.capacity()==0` condition?

makubacki (approved):

patina_dxe_core/src/gcd/spin_locked_gcd.rs
os-d: There is in the test_add_memory_space_before_memory_blocks_instantiated(). I had to refactor it slightly for this change (to also test what it was testing before), but it tests this case.

patina_dxe_core/src/gcd/spin_locked_gcd.rs
os-d: I am not going to update the IO one for two reasons:
- it is not as easy, for the memory we initialize it with one call while processing the PHIT HOB. For IO space, we just iterate through the HOBs in a loop and call add_io_space. We'd have to have special logic to keep track of which is the first one.
- IO space is x86 only, the issue mentioned with speculative execution is AArch64 only (and also wouldn't apply to IO space if it existed). This is still not the prettiest pattern, but IO space is not heavily used by us.

PR #1036, opened 2025-11-08
Revise Atomics RFC

joschock:
## Description

Revise Atomics RFC to reflect findings from performance measurement and rescope the proposal in the RFC.

Original RFC: https://github.com/OpenDevicePartnership/patina/pull/810

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [x] Includes documentation?

## How This Was Tested

N/A - documentation.

## Integration Instructions

N/A.


docs/src/rfc/text/0021-remove-atomics.md
```diff
@@ -0,0 +1,201 @@
+# RFC: `Remove Atomic from Patina`
+
```
os-d, line 2: Nit on process: Because this is a revisement to RFC 18, albeit a substantial one, I think that we wouldn’t want to delete 18 and create a new one, but still do this as an amendment to 18.

joschock: I'm not renaming it to 20 because of the revision; but another 18 snuck in and I completed this one without noticing the duplication; so the rename is just to remove the duplication.

joschock: (Duplicated with https://github.com/OpenDevicePartnership/patina/blob/main/docs/src/rfc/text/0018-system-tests.md)

PR #1039, opened 2025-11-10
smbios: replace TplMutex with RefCell to fix deadlocks and auto-add Type 127 marker

kat-perez:
Replace TplMutex with RefCell to fix deadlocks when event handlers at elevated TPL call GetNext(). Auto-add Type 127 End-of-Table marker and maintain invariant that it always appears last. Publish table immediately during initialization so it's available to drivers during DXE phase.

## Description

This change replaces `TplMutex` with `RefCell` in the SMBIOS manager to fix deadlocks that occur when UEFI event handlers at elevated TPL (Task Priority Level) call the SMBIOS protocol's `GetNext()` function. The deadlock occurred because `TplMutex` would attempt to raise TPL while already at an elevated TPL, causing the system to hang.

  **Key Changes:**
  1. **RefCell replaces TplMutex**: Uses Rust's borrow checking instead of TPL-based locking. Methods now return `AlreadyBorrowed` error if a borrow conflict occurs instead of deadlocking.

  2. **Pre-allocated buffers**: Allocates SMBIOS table and entry point buffers upfront during component initialization to avoid
  memory allocation at elevated TPL during protocol calls.

  3. **Auto-add Type 127 End-of-Table marker**: Automatically adds the required SMBIOS Type 127 (End-of-Table) marker and maintains the invariant that it always appears last, even when records are added out of order.

  4. **Early table publication**: Publishes the SMBIOS table immediately during initialization (with the Type 127 marker) so it's available to drivers throughout the DXE phase, with dynamic updates when records are added.

  **Rationale for RefCell:**
  - UEFI event handlers can run at elevated TPL and need to query SMBIOS data
  - MfciDxe was found to call SMBIOS protocol at TPL_NOTIFY (violating UEFI spec), triggering deadlock with TplMutex
  - TplMutex's design assumes callers are at lower TPL and can raise TPL safely
  - RefCell provides interior mutability without TPL constraints
  - Error handling via `AlreadyBorrowed` allows graceful degradation instead of system hangs

  **Checklist:**
  - [x] Impacts functionality?
  - [ ] Impacts security?
  - [ ] Breaking change?
  - [x] Includes tests?
  - [x] Includes documentation?

  ## How This Was Tested

  - Unit tests for RefCell borrow checking and `AlreadyBorrowed` error handling
  - Unit tests for Type 127 auto-addition and last-position invariant
  - Unit tests for manager core functionality with pre-allocated buffers
  - Integration testing with both Rust DXE Core and TianoCore DXE Core on QEMU platform
    - Verified SMBIOS table is published and accessible during DXE phase
    - Verified no deadlocks when SMBIOS protocol is called from event handlers
    - Verified dynamic table updates work correctly when drivers add records

  ## Integration Instructions

No special integration steps required. The change is internal to the patina_smbios component and maintains API compatibility with existing SMBIOS protocol consumers.

### DIFF
--- a/components/patina_smbios/src/manager/core.rs
+++ b/components/patina_smbios/src/manager/core.rs
@@ -25,7 +25,7 @@ use r_efi::{
     efi,
     efi::{Handle, PhysicalAddress},
 };
-use zerocopy::{IntoBytes, Ref};
+use zerocopy::{FromBytes, IntoBytes, Ref};
 use zerocopy_derive::{Immutable, IntoBytes as DeriveIntoBytes};
 
 use crate::{
@@ -34,6 +34,7 @@ use crate::{
         SMBIOS_HANDLE_PI_RESERVED, SMBIOS_STRING_MAX_LENGTH, SmbiosHandle, SmbiosRecordsIter, SmbiosTableHeader,
         SmbiosType,
     },
+    smbios_record::{SmbiosRecordStructure, Type127EndOfTable},
 };
 
 use super::record::SmbiosRecord;
@@ -76,6 +77,9 @@ pub struct Smbios30EntryPoint {
 /// SMBIOS table manager
 ///
 /// Manages SMBIOS records, handles, and table generation.
+///
+/// The manager uses pre-allocated buffers for table publication to avoid
+/// repeated memory allocations and maintain a stable table address throughout DXE.
 pub struct SmbiosManager {
     pub(super) records: RefCell<Vec<SmbiosRecord>>,
     pub(super) next_handle: RefCell<SmbiosHandle>,
@@ -84,6 +88,14 @@ pub struct SmbiosManager {
     pub minor_version: u8,
     entry_point_64: RefCell<Option<Box<Smbios30EntryPoint>>>,
     table_64_address: RefCell<Option<PhysicalAddress>>,
+
+    // Pre-allocated buffers for table publication
+    /// Pre-allocated buffer for SMBIOS table data (e.g., 64KB)
+    table_buffer_addr: RefCell<Option<PhysicalAddress>>,
+    /// Maximum size of the pre-allocated table buffer
+    table_buffer_max_size: usize,
+    /// Pre-allocated buffer for entry point structure
+    ep_buffer_addr: RefCell<Option<PhysicalAddress>>,
 }
 
 impl SmbiosManager {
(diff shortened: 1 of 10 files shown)


components/patina_smbios/src/manager/core.rs
```diff
@@ -115,9 +127,67 @@ impl SmbiosManager {
...
+    /// Allocate pre-allocated buffers for SMBIOS table publication
+    ///
+    /// This allocates a large buffer for the SMBIOS table and entry point upfront
+    /// to avoid repeated allocations on every Add(). The buffers are reused for
+    /// all table publications.
+    ///
+    /// Also automatically adds the Type 127 End-of-Table marker to ensure
+    /// SMBIOS compliance and maintain the invariant that it's always last.
+    ///
+    /// Must be called at TPL_APPLICATION or lower to allow memory allocation.
```
joschock, line 146: TPL_APPLICATION is the lowest TPL.

joschock: Allocation is allowed at any TPL <= NOTIFY

components/patina_smbios/src/manager/protocol.rs
```diff
@@ -41,12 +41,16 @@ pub(super) struct SmbiosProtocolInternal {
     // The public protocol that external callers will depend on
     pub(super) protocol: SmbiosProtocol,
 
-    // Internal component access only! Does not exist in C definition
-    pub(super) manager: &'static TplMutex<'static, SmbiosManager, StandardBootServices>,
+    manager: &'static RefCell<SmbiosManager>,
```
joschock, line 44: I'm not sure this is correct. RefCell is not shareable across contexts (i.e. it doesn't implement mutual exclusion). If you were hitting deadlocks with TplMutex due to re-entrancy, I don't think RefCell fixes that; it would just hide corner cases behind UB (because, if you take an interrupt that is giving you re-entrancy, RefCell doesn't expect that, and some of it's invariants are violated).

components/patina_smbios/src/manager/protocol.rs
```diff
@@ -176,6 +186,39 @@ impl SmbiosProtocol {
...
+                    // 2. Install configuration table (WITHOUT manager lock to avoid deadlock)
+
+                    // Drop manager borrow before republishing
+                    drop(manager);
+
+                    // Rebuild table data into pre-allocated buffer
+                    // The configuration table was already installed during init and points to our
+                    // pre-allocated entry point buffer. We just need to update the data in place.
+                    let (table_addr, ep_addr, entry_point) = {
+                        let manager = match internal.manager.try_borrow() {
```
joschock, line 201: I think this should assert/fault, unless there is a legit case where you expect this to be called with the manager borrowed. If that's the case, you should describe that scenario in comments.

components/patina_smbios/src/manager/protocol.rs
```diff
@@ -176,6 +186,39 @@ impl SmbiosProtocol {
...
+                            Ok((t_addr, e_addr, ep_struct)) => (t_addr, e_addr, ep_struct),
+                            Err(_) => {
+                                return efi::Status::DEVICE_ERROR;
+                            }
+                        }
+                    }; // manager borrow dropped here
+
+                    // Store addresses in manager
+                    {
+                        let manager = match internal.manager.try_borrow() {
```
joschock, line 215: Same comment: try_borrow() should always succeed unless you are designing for a specifically re-entrant scenario (in which case you should describe that scenario in comments for posterity).

components/patina_smbios/src/manager/protocol.rs
joschock: Note: TplMutex allows you to manage re-entrancy better because it not only enforces critical sections, but it also means that you won't be interrupted by anything running at the TplMutex TPL level or lower. So the only source of re-entrancy at TPL_NOTIFY would be something calling your routines at TPL_HIGH (which would not be allowed by spec) or by re-entrant calls into this code. The latter should be handled by refactoring the code to avoid reentrancy.

kat-perez:
> In general, I think moving away from TplMutex is a mistake. In particular, I think it leaves you open to subtle data race bugs. For example, let's say I do the following twisted thing (rough pseudocode):
> 
> ```
> fn timer_event_handler() {
>    smbios_protoco.add(record);
>    smbios_protocol.remove(record);
> }
> 
> fn test_entry_point() {
>    setup_repeated_timer_event_with_handler(timer_event_handler, TPL_NOTIFY);
>    // Running at TPL_APPLICATION
>    loop {
>      smbios_protocol.add(record);
>      smbios.protocol.remove(record);
>   }
> }
> ```
> 
> `timer_event_handler()` runs periodically invoked on the timer tick. It may interrupt the loop in test_entry_point at any point, in particular in the middle of private RefCell.try_borrow(). Since RefCell is not `Sync` it is unsafe for that to happen.
> 
> Put another way, the above pseudocode illustrates how one `RefCell::try_borrow()` could happen while another `RefCell::try_borrow()` is executing on the same data; a condition that violates the safety requirements of RefCell.
> 
> TplMutex avoids this by ensuring (using the core TPL behavior) that you are not interrupted during the critical section. If you are hitting a re-entrant deadlock with TplMutex, it likely means that either your code is being called at an unexpectedly high TPL (nothing should be adding SMBIOS records at TPL_HIGH), or that this module is calling back into itself while holding the TplMutex; which in my opinion is a design issue that should be fixed or at least only done with very good justification.
> 
> I think the rust compiler would normally prevent this usage of RefCell, except it doesn't know that the two RefCell accesses are to the same location (because of how the context is reconstructed from the raw pointer in this code):
> 
> ```
>         // SAFETY: We must trust the C code was a responsible steward of this pointer
>         // Cast from protocol pointer to internal struct pointer
>         let internal = unsafe { &*(protocol as *const SmbiosProtocolInternal) };
>         let manager = internal.manager.lock();
>         let manager = match internal.manager.try_borrow() {
>             Ok(m) => m,
>             Err(_) => return efi::Status::DEVICE_ERROR,
>         };
> ```
> 
> Because the `internal` is reconstructed from the raw pointer, Rust doesn't know you are potentially accessing it in two different contexts. Otherwise it would require it to be `Sync`, and that would cause a compile fail.

thanks for explaining the issue with RefCell and now I understand. RefCell isn't `Sync` so it can be corrupted during borrow checking.

I've investigated the original TplMutex deadlock and found the root cause. The old code was calling `install_configuration_table()` while holding the TplMutex lock. This caused a re-entrant lock panic when events fired and tried to access SMBIOS. It was the "calling back into itself" issue. I'll move back to restoring TplMutex and fix that issue.

kat-perez:
moving to draft while i fix the bug

kat-perez:
closing this PR since I've root-caused that the issue was not originating from using TplMutex

PR #1041, opened 2025-11-10
RFC: Static Resolution and usage in the Core

Javagedes:
## Description

**Current Status: FCP**

This RFC proposes a re-architecture of the current usage of statics in the core

#1049 is the reference implementation for only half of this RFC. It is the reference implementation for moving the core itself to be static, which will be the first step in moving all static members into the core.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

N/A

## Integration Instructions

N/A


docs/src/rfc/text/0022-core-static.md
```diff
@@ -0,0 +1,307 @@
...
+## Goals
+
+1. Size / performance improvements
+   1. Static definition of the Core reduces initial stack size
+   2. Defferred instantiation of components reduces stack size
+   3. Static resolution of dependencies (static polymorphism) over dyn Trait interfaces improves performance
+   4. Static resolution of dyn trait objects in statics
+
+2. Easier testing for Patina maintainers:
+   1. Removal of statics reduces complexity writing tests that are properly reset
```
cfernald, line 51: I would mention improved testing perf due to better ability to parallelize

docs/src/rfc/text/0000-core-static.md
```diff
@@ -0,0 +1,307 @@
...
+   3. Static resolution of dependencies (static polymorphism) over dyn Trait interfaces improves performance
+   4. Static resolution of dyn trait objects in statics
+
+2. Easier testing for Patina maintainers:
+   1. Removal of statics reduces complexity writing tests that are properly reset
+   2. Generics for Core dependencies allows for easier mocking
+
+3. Improved code cleanliness for Patina maintaineres
+   1. Never calling "efiapi" functions
+   2. Making "efiapi" functions a light wrapper around pure-rust impelmentations
```
cfernald:
```suggestion
   2. Making "efiapi" functions a light wrapper around pure-rust implementations
```

docs/src/rfc/text/0000-core-static.md
```diff
@@ -0,0 +1,307 @@
...
+generic trait resolution. If you remember previously, we did this in the core itself, and it lead to a laundry list of
+traits with complex types specified. This interface is cleaner, and provides less complexities in the `Core`.
+
+To start, this trait will have a limited number of configuration options as shown below, but it is expected that more
+will be specified as needed. Note that this trait is now what adds components / configs / services, instead of the core.
+
+```rust
+// In patina_dxe_core/src/lib.rs
+trait Platform {
+    const PRIORITIZE_32_BIT_MEMORY: bool = false;
```
cfernald: A pattern I have seen in other projects for this type of thing is instead of doing a trait constant, they have const functions in the trait. In those projects they extend this model to even allows optional trait implementations on the struct, but for basic config values it may not matter, but curious to see if you think there is any value. Perhaps mocking the config would be useful for testing.

```rust
#[inline]
const fn priotitize_32_bit_memory() -> bool {
    false // Defaults to false
}
```

docs/src/rfc/text/0000-core-static.md
```diff
@@ -0,0 +1,307 @@
...
+
+        ...
+    }
+}
+```
+
+### Creation of the `UefiState` struct
+
+Since we will be moving all of the static state into the Core, we need a clean way to store all of this state. To do
+that, we will create a new struct called `UefiState` that stores everything that is currently static. Please note that
```
cfernald: What is the benefit of putting them in a sub-struct instead of just directly into the core?

docs/src/rfc/text/0000-core-static.md
```diff
@@ -0,0 +1,307 @@
...
+        adder.add_component(...);
+    }
+
+    fn configs(adder: &mut Adder<Config>) {
+        adder.add_config(...);
+    }
+
+    fn services(c: &mut Addr<Service>) {
+        adder.add_service(...);
+    }
```
joschock: Pedantry: I think these are inconsistent with the platform trait described previously (e.g. `Adder<Config>` vs. `Add<Config>`)

docs/src/rfc/text/0000-core-static.md
```diff
@@ -0,0 +1,307 @@
...
+
+    fn components(add: &mut Add<Component>) { }
+
+    fn configs(add: &mut Add<Config>) {  }
+
+    fn services(add: &mut Add<Service>) {  }
+
+    #[cfg(target_arch = "aarch64")]
+    fn gic_bases() -> GicBases;
+}
```
joschock: Can stuff like `section_extractor()`/`gic_bases()` somehow be moved to be a Service/Config under this model? It's weird that the core config is broken out from the component config.

cfernald (commented):
Overall, I like this. It provides a significant improvement to clarity on the platform. Particularly in the platform now has a clear list of it's responsibilities in terms of the functions it implements, the configuration that is available, etc. Some of which can have defaults, and other not, forcing the platform to make a choice. Additionally, it clarifies the core construction significantly since I think the chained start state can be a bit cluttered in implicit about the stages of the core. 

My only fear on this is the pain having a robust generic may have for testing and mocking, but that seems like something we can build infrastructure around.

docs/src/rfc/text/0000-core-static.md
```diff
@@ -0,0 +1,307 @@
...
+It is suggested you read:
+
+1. [rust-lang: Traits](https://doc.rust-lang.org/reference/items/traits.html)
+2. [nomicon: repr-rust](https://doc.rust-lang.org/nomicon/repr-rust.html)
+
+## Goals
+
+1. Size / performance improvements
+   1. Static definition of the Core reduces initial stack size
+   2. Defferred instantiation of components reduces stack size
```
makubacki:
```suggestion
   2. Deferred instantiation of components reduces stack size
```

docs/src/rfc/text/0000-core-static.md
Javagedes: What do you mean by switched? The current top-of-tree implementation has them as a service and config respectively, so we could keep them that way if you prefer.

Playing devil's advocate (I'm not trying to defend too much, I don't have a strong opinion), there are a few (minor) benefits to moving it to hardcoded settings for a platform:

1. It fully declares all platform integration requirements and is validated at compile time rather than at runtime (such as the GicBases that currently fail at compile time if not set
2. For the things like services, it allows the Core to use the functionality immediately, instead of having to add in comments (like we do now) that the service needs to be registered outside of a component so that it is available early in Core startup (e.g. for the section extractor we need the functionality before components start executing, so if someone produces it via a component, the platform has missed it's window) and we have no way to check that.
3. There is a very slightly performance benefit (hardly noticeable) because we are using static polymorphism rather than passing around dyn trait objects for services or dyn Anys for configs that must be dereferenced
4. It could also be argued that it is slightly weird / counter intuitive that we are using the dynamic polymorphism functionality for things that are only used by the core, that could be configured via static / compile time polymorphism and have compile time validation that everything was properly set for a platform (Kind of rounding back on point (1))

docs/src/rfc/text/0000-core-static.md
Javagedes: traits do not currently support const functions, so you may be getting it slightly mixed up. 

Due to this, if we do truly need something that needs to be setup in the `new` function of the `Core`, it will need to be a `const` definition. 

However, for something like `PRIORITIZE_32_BIT_MEMORY`, we do not set that until (the old named) `init_memory`, which is not `const`. Due to this, we could use an inline function defined in the Platform if you prefer.

docs/src/rfc/text/0000-core-static.md
Javagedes: Mainly it was a personal preference to separate out the data. At a rough count I saw about 20 statics spread across `patina_dxe_core` and I did not exactly want that many fields in the top-level Core struct.

It also slightly separates out the "uefi" functionality from the "component" functionality, which I like.

Javagedes: We could theoretically move all the UEFI-esque functions to this struct instead of implemented on the `Core`.

docs/src/rfc/text/0000-core-static.md
```diff
@@ -0,0 +1,307 @@
...
+
+It is suggested you read:
+
+1. [rust-lang: Traits](https://doc.rust-lang.org/reference/items/traits.html)
+2. [nomicon: repr-rust](https://doc.rust-lang.org/nomicon/repr-rust.html)
+
+## Goals
+
+1. Size / performance improvements
+   1. Static definition of the Core reduces initial stack size
```
vineelko: I think it is a good idea to find the current stack size. One way I am thinking is by using the debugger, for example, since `Core` is instantiated in `_start()` on Q35(release), it allocates `sub rsp, 358h` (856 bytes) of stack space. This does not account for the peak stack usage of sub calls.

docs/src/rfc/text/0000-core-static.md
```diff
@@ -0,0 +1,307 @@
...
+   2. Defferred instantiation of components reduces stack size
+   3. Static resolution of dependencies (static polymorphism) over dyn Trait interfaces improves performance
```
vineelko: Something to consider, switching to static polymorphism might also increase the binary size.

docs/src/rfc/text/0000-core-static.md
```diff
@@ -0,0 +1,307 @@
...
+    fn configs(adder: &mut Adder<Config>) {
+        adder.add_config(...);
+    }
+
+    fn services(c: &mut Addr<Service>) {
+        adder.add_service(...);
+    }
+}
+
+static CORE: Core<Platform> = Core::new();
```
vineelko: How is the `uefi_state` member of `Core` initialized here? May be having a rough draft PR long with the RFC would make it easier to look at the details.

docs/src/rfc/text/0000-core-static.md
```diff
@@ -0,0 +1,307 @@
...
+```
+
+### Platform Interface changes
+
+We've now seen all of the core changes. What does the the change look for the platform? The platform now manually
+implements the `Platform` trait. Pretty much all platform config has moved there. That is about it. Lets see the
+example:
+
+```rust
+struct Platform;
```
vineelko: Should we consider naming this after the actual platform being used, something like `Q35Platform`?

docs/src/rfc/text/0000-core-static.md
joschock: Mmm. I am on board with wanting them to be compile-time statics - I guess what I'm trying to articulate is that it seems like mixing levels of generalization - opaque abstractions like `fn configs()` mixed with specifics like `PRIORITIZE_32_BIT_MEMORY` and `gic_bases()`. Maybe if we encapsulated or grouped the core config into a struct it would help?

docs/src/rfc/text/0000-core-static.md
Javagedes: I'm not sure I understand the question - uefi_state is just another field of `Core` which is now initialized with `Core::new()` All of the structs in `UefiState` were already static, so they already support moving them into a different struct that is static.

I'll throw up a PR here soon :)

docs/src/rfc/text/0000-core-static.md
Javagedes: I can rename it that for the RFC is you like, but ultimately the platform can name it whatever they want :)

docs/src/rfc/text/0000-core-static.md
Javagedes: Good catch. It will definitely increase the binary size (though minimally, since most of the size is us pulling in structures that were already static). I will update :)

Javagedes: Will clarify in the PR that I mean initial stack frame size will reduce. I will also mention that overall binary size will increase

docs/src/rfc/text/0000-core-static.md
Javagedes: I will get a measurement and put it in the PR, however this change is more forward looking, so I don't suspect to see a huge difference today.

e.g. Right now, we only have about 5 components or so that this will affect. Down the line when we have 20 or 30 or 50 components / configs, the stack size effect of this change would be much larger.

Javagedes:
> Overall, I think consolidating the static usage across patina is a good idea, but for static vs dynamic polymorphism argument, I feel we need some way to measure the improvements b/w each of them.

I've marked the performance impact as a very minor result of the change, not the intent of the change, as I agree, the impact is incredibly minimal, and honestly not worth measuring. The current usage of dyn trait objects in the core (outside of components) is not in a very "hot" path for performance to be noticed.

docs/src/rfc/text/0000-core-static.md
cfernald: Yes, you are right that I hallucinated the `const`. I think the only benefit would be for testing. If we were to use `fn priotitize_32_bit_memory(&self)` and actually keep an instance of then we could use mockall to test a function with different configurations without having to define N different Platform implementors for different tests. Though whether its worth the pain of keeping a struct around or potentially losing some compile time optimization, i'm really not sure. Do you think it will be painful to get test converage with the const's for config?

docs/src/rfc/text/0000-core-static.md
os-d:
```suggestion
   2. Deferred instantiation of components reduces stack size
```

docs/src/rfc/text/0000-core-static.md
cfernald: I agree that we need some organization, though I would actually split it up more then just UEFI and Component, and pull out some of the critical kernel-like infrastructure. E.g. Memory, UEFI, interrupts, debugger, components, etc. Though honestly, I believe this is an implementation detail and so I don't think the specific substructures need to be locked now.

This did make me realize we can move the debugger into the core now if it's static.

docs/src/rfc/text/0000-core-static.md
Javagedes: Clarified size reduction is regarding initial stack frame throughout the entire RFC. Included some minor measurements of stack size in the `Guide-level Explanation` section.

docs/src/rfc/text/0000-core-static.md
os-d: I agree with John, having a hierarchy here would make sense, I think. Having some logical grouping of configurations.

docs/src/rfc/text/0022-core-static.md
```diff
@@ -0,0 +1,329 @@
...
+This will introduce changes to the core and how it consumes the trait. An example is shown below
+
+```rust
+pub struct Core<P: Platform> {
+    ...
+    _p: PhantomData<P>
+}
+
+impl <P: Platform> Core<P> {
+    /// This was originally `init_memory`. It is now `start` and the only public function.
```
os-d, line 186: I think this model is much better and not having to rely on the platform to have the right order of when to call init_memory vs start.

docs/src/rfc/text/0000-core-static.md
os-d: Yeah, same thought here, I agree with bucketizing the configs

docs/src/rfc/text/0000-core-static.md
```diff
@@ -0,0 +1,329 @@
...
+
+### Creation of the `UefiState` struct
+
+Since we will be moving all of the static state into the Core, we need a clean way to store all of this state. To do
+that, we will create a new struct called `UefiState` that stores everything that is currently static. Please note that
+to reduce the complexity of the PR, moving things into `UefiState` and out of being `static` will happen over multiple
+PRs. The first one we will do is the dispatcher context, so the example below will show the dispatcher context being
+moved into `UefiState`
+
+Another benifit with this layout is that some of the dyn traitobjects currently used in statics (because we cannot
```
os-d:
```suggestion
Another benefit with this layout is that some of the dyn trait objects currently used in statics (because we cannot
```

docs/src/rfc/text/0000-core-static.md
```diff
@@ -0,0 +1,329 @@
...
+}
+```
+
+The next final tidbit which is handy for this implementation is that `impl`'s of a struct do not have to be in the same
+module that the struct is defined. This allows us to scatter `Core<P>` impls throughout the workspace.
+
+A short and sweet example of how we will implement efiapi functions across the workspace is seen below:
+
+```rust
+impl <P: Platform> Core {
```
os-d: I think this is an unfortunate design point, but I understand the reason we ended up here. Does this allow anyone to call any efiapi function in the core? They shouldn't but, now could because we have the atomic pointer available everywhere and these are all impls of the core and so any function that is an impl of core (of which there are now a lot) could call any of these?

I also think that spreading out all the impls of Core around feels odd. And then what lives in there vs not. Maybe just the efiapis do and in the below example `trust` doesn't have to? As in `trust` is in impl Dispatcher, still, and `trust_efiapi` calls `Self::instance().dispatcher.trust()` or whatever? It feels odd to have all of our functions live in impl Core.

docs/src/rfc/text/0022-core-static.md
```diff
@@ -0,0 +1,329 @@
...
+produce an API interface for accessing this state.
+
+The next limitation with this process is that we have a `Core` struct with a generic `P` for the platform. Due to this
+the layout of the `Core` is not stable and we cannot safely convert to the `Core` without also knowing `P`. What this
+means is that we must move all "efiapi" functions into the `Core`, which allows static polymorphism to handle knowing
+what `P` is for us (e.g. we can just do `Self::instance()`).
+
+```rust
+/// in lib.rs
+static __SELF: AtomicPtr<u8> = AtomicPtr::new(core::ptr::null_mut());
```
os-d, line 289: A question for my understanding: Do we actually need an AtomicPtr here vs a *Cell? My understanding is that we can use *Cell for interior mutability, which is what we want here. Everything else that needs locking is locked underneath, so there shouldn't be concurrency issues and we don't need any atomic operations?

docs/src/rfc/text/0000-core-static.md
```diff
@@ -0,0 +1,329 @@
...
+
+```rust
+struct Platform;
+
+impl patina::Platform for Platform {
+    const PRIORITIZE_32_BIT_MEMORY: bool = false;
+
+    type Extractor: SectionExtractor = CompositeSectionExtractor;
+
+    fn section_extractor -> Self::Extractor { CompositeSectionExtractor::default() }
```
os-d: In the spirit of the above comments, if we have a hierarchy here, I think we need a name for what the section extractor (and debugger, etc) are if they are not components, configs, or services

docs/src/rfc/text/0000-core-static.md
```diff
@@ -0,0 +1,329 @@
...
+    fn services(c: &mut Add<Service>) {
+        add.add_service(...);
+    }
+}
+
+static CORE: Core<Platform> = Core::new();
+
+pub extern "efiapi" fn _start(physical_hob_list: *const c_void) -> ! {
+    ...
+    CORE.start(physical_hob_list)
```
os-d: A nit: If we are changing this anyway, I prefer the `entry_point` nomenclature to `start`, to be more consistent with firmware usage, but not a big deal

docs/src/rfc/text/0000-core-static.md
```diff
@@ -0,0 +1,329 @@
...
+}
+```
+
+## Guide-Level Explanation
+
+In this RFC, we are attempting to clean up and consolidate the static usage in the `patina_dxe_core` crate, which has
+many benifits to both patina developers and platform owners. For Patina developers, by consolidating all static usage
```
os-d:
```suggestion
many benefits to both patina developers and platform owners. For Patina developers, by consolidating all static usage
```

docs/src/rfc/text/0000-core-static.md
```diff
@@ -0,0 +1,329 @@
...
+the codebase, which enables easier testing because we can mock parts that we are not actually testing, but must exist
+to compile.
+
+For platform owners, static polymorphism allows them to fully define all of their platform configurations and
+dependencies at compile time, which will result in compilation errors when initially configuring the core for their
+platform, or as configurations evolve with Patina.
+
+Generically, the combination of both a static core and delayed instantiation of components should reduce the size of
+the initial stack frame. While the size of the initial stack frame is not currently a concern, as platforms begin to
+use more patina components, it will be. This change will sacrifice some binary size (size of the `Core` object) to
```
os-d:
```suggestion
use more patina components, it will be. This change will increase the binary size (size of the `Core` object) to
```

small clarification, usually a sacrifice is something you are losing, but we are gaining binary size (look at the positives!)

docs/src/rfc/text/0000-core-static.md
Javagedes: That is a good call out. the `mockall::automock` attribute for mocking does not support const values. Each crate would have to manually do a mock with the `mockall::mock` macro. e.g.

```rust
mock::mock! {
    MyMock {}
    
    impl Platform for A {
        const PRIORITIZE_32_BIT_MEMORY: bool = true;

       fn gic_bases() -> GicBases;
    }
}
```

That reason alone may make it worth avoiding const fields in a trait at all costs. By not doing that we can throw the `#[mockall::automock]` attribute on it.

Javagedes: Updating to suggest not using const definitions at this point in time for this reason alone.

docs/src/rfc/text/0000-core-static.md
```diff
@@ -0,0 +1,335 @@
...
+- 2025-11-11: Clarified that this change reduces the initial stack frame size by sacrificing some binary size.
+- 2025-11-11: Clarified that the performance gains are a side affect, not the intent of this change and that the
+  performance differences would be incredibly small, at least for the impacted code today.
+- 2025-11-11: Added test performance impact as tests can remove the global lock, and run in parallel.
+- 2025-11-11: Remove associated constant from Platform trait to allow usage of `mockall::automock`
+
+## Motivation
+
+1. Improvements to initial stack size
+2. Testing improvementsfor Patina maintainers
```
makubacki:
```suggestion
2. Testing improvements for Patina maintainers
```

docs/src/rfc/text/0000-core-static.md
Javagedes: I am open to exampled suggestions if you have some time.

I personally like the laundry list of methods for each individual requirement for a platform because it fully lists all requirements for a platform while also allowing us to have some requirements that have default values. **I can definitely see grouping of minor, likeminded core configs into single functions to reduce some of the boilerplate useful.**

That said to one of John's point, I agree having a `config` function for component config, then a bunch of other config options for the core is confusing. We could reduce the `components`, `configs`, and `services` function to a single function called `setup_<patina>_components` or `configure_<patina>_components` or something.

Ultimately though, I don't see a single core config function that returns the entire config at once as a good long-term solution. There are a couple negatives I will list after the short example

```rust
struct CoreConfig<E: SectionExtractor> {
    ...
}

trait Platform {
    type Extractor: SectionExtractor
    fn core_config() -> CoreConfig<Self::Extractor>;
}
```

1. As we add more generics to the core, the complexity of this CoreConfig object increases and we end up with our version revision of the "Core" instead just name "CoreConfig"
2. We either have to call `core_config` multiple times and grab the one thing we need for each use case, or we have to stash the core config as a field in the core itself. Both of which I'm not personally a fan of. Even the latter (storing it in the core itself), which seems nice at first, is not great. Here in the "now" we already will hit a snag because the section extractor is used in both the dispatcher and fv modules, which will further intertwine things that we are trying to avoid
3. Our ability to have default configurations (such as the prioritize 32-bit memory) becomes a bit more of a pain. We have to provide a user interface to allow them only configuring what they want to configure, which can be done, but is overly complex depending on the route we go. Either every field must implement `Default` (which we do not want), or we have to go a builder pattern, which seems obtuse for just a config

```rust
struct CoreConfig<E: SectionExtractor> {
    extractor: E, // required 
    config1: u32, // required 
    config2: u32, // default value w/ overide
    config3: u32, // default value w/ override
}

impl <E: SectionExtractor> CoreConfig<E> {
    fn builder(extractor: E, config1: u32) {
        Self {
            exractor,
            config1,
            config2: 52, // Our default value
           config3: 40 // Our default value
        }
    }

    fn with_config2(mut self, config2: u32) -> Self {
        self.config2 = config2
        self
    }
    
    fn with_config3(mut self, config3: u32) -> Self {
        self.config3 = config3
        self
    }
}
```

docs/src/rfc/text/0000-core-static.md
```diff
@@ -0,0 +1,335 @@
...
+this, the prime knowledge area is rust traits, generics, and the static polymorphism provided by the traits.
+
+It is suggested you read:
+
+1. [rust-lang: Traits](https://doc.rust-lang.org/reference/items/traits.html)
+2. [nomicon: repr-rust](https://doc.rust-lang.org/nomicon/repr-rust.html)
+
+## Goals
+
+1. Stack size
```
makubacki:
```suggestion
1. Reduction of stack size
```

docs/src/rfc/text/0000-core-static.md
Javagedes: Updated to not use associated generics, and also to specifically say we are initially avoiding them so that the automock attribute works.

docs/src/rfc/text/0000-core-static.md
Javagedes: This comment was a reference to the 20ish statics we have throughout the codebase, not the platform configurations themselves. 

I agree that some of the current statics do not belong in the `UefiState` structure. I will make note of that in the RFC. That also means we will not (and cannot) pull the uefi-related functions to be implemented on `UefiState` rather than the `Core` because some of these methods rely on the statics mentioned above.

docs/src/rfc/text/0000-core-static.md
joschock: Could we split the trait into separate `Platform` and `CoreConfig` traits? The first trait deals with "I'd like to add these platform components/services/configs", the second trait deals with "I'd like to configure these core behaviors"
```rust
trait Platform {
    fn components(add: &mut Add<Component>) { }

    fn configs(add: &mut Add<Config>) {  }

    fn services(add: &mut Add<Service>) {  }
 }

trait CoreConfig {
    type Extractor: SectionExtractor;

    fn prioritize_32_bit_memory() -> bool { false }

    fn section_extractor() -> Self::SectionExtractor;

    #[cfg(target_arch = "aarch64")]
    fn gic_bases() -> GicBases;
}

pub struct Core<P: Platform, C:CoreConfig> {
    ...
```

makubacki: Settings specifics aside, this interface is not very cohesive, and I also thought it could benefit from grouping/organization. Can you explain in more detail who owns this configuration? "Core" is composed of multiple subsystems, so is this expected to be a grab bag of all those subsystem's configuration options in a single interface?

docs/src/rfc/text/0000-core-static.md
Javagedes: I really do not know how to spell benefit. I think This needed correcting in like 5 spots lmao.

docs/src/rfc/text/0022-core-static.md
makubacki: Yeah, I also think this will clarify logic for starting the core as complexity increases over time.

docs/src/rfc/text/0022-core-static.md
Javagedes: I think you are right, we do not need an atomic. we can use a Cell (I'll probably use a once cell though). Will update and close this.

makubacki: The Cell also preserves the type better (`OnceCell<&'static Core<P>>` vs `*mut u8`).

docs/src/rfc/text/0022-core-static.md
```diff
@@ -0,0 +1,343 @@
...
+
+impl <P: Platform> Core {
+    fn set_instance(&'static self) {
+        let ptr = NonNull::from(self).cast::<u8>().as_ptr();
+        __SELF.store(ptr, core::sync::atomic::Ordering::SeqCst);
+    }
+
+    pub(crate) fn instance<'a>() -> &'a Self {
```
makubacki, line 297: Is the lifetime intentionally not `'static`?

docs/src/rfc/text/0000-core-static.md
Javagedes: John, I like your idea. I will probably go that route depending on the way this conversation leads. It also lends well to re-usability. If we are maintaining two platforms, they could reuse the same `Platform` trait implementation but use two different `CoreConfig` implementations (maybe).

for @makubacki, I'm going to focus your question on just the `CoreConfig` as John named it. This trait is not really meant to be cohesive. It is the trait that defines the laundry list of all configurations / custom implementations that the Platform maintainer selects to configure the Patina DXE Core to work for their platform (It has nothing to do with components). It can be thought of similar to LibraryClasses and PCDs in the UEFI context. All of the configuration for the entire core happens here. Generic selections, value selections, etc. The owner is purely the Platform, and Patina Core simply uses it generically.

I agree that there could be some grouping done, but it would be minimal because the intent is that the core just grabs and uses the data it needs, when it needs it. Examples in the RFC use it like this:

```rust
fn start(&'static mut self, phsical_hob_list: *const c_void) {
    ...

    GCD.prioritize_32_bit_memory(P::prioritize_32_bit_memory());
   
    ...

    dispatcher::register_section_extractor(P::section_extractor());
    
    ....
}
```

docs/src/rfc/text/0000-core-static.md
Javagedes: The spreading out of all the `impl`s of the Core (which you mentioned feels odd) actually helps to enforce **not** doing what you mentioned in your first paragraph.

functions (including functions in `impl`s) are only public in the module they are defined (even if the struct you are `impl`ing is defined elsewhere). Due to this, so long as you don't mark the "efiapi" functions as `pub` or `pub(crate)`, then they won't be callable anywhere else in the core. The intent is to keep the `efiapi` functions private and the non-efiapi functions `pub(crate)`.

I can't think of a good solution preventing a Patina developer from just grabbing that `SELF` pointer and doing something funky with it. But we control what gets merged, so we can make sure that never happens. the `SELF` pointer is private, so Patina consumers won't be able to access it or call the `instance()` method because it is just `pub(crate)`

Does that elevate your concern at all?

For the other part, I'm fine with not implementing everything on the `Core`. It just felt a little cleaner to be anywhere in the codebase and be able to do `self.trust(...)` rather than `self.uefi_context.dispatcher_context.trust(...)`. But thats pretty pedantic and I am fine with moving it.

As for the very last part. Yes **ALL** efiapi's simply call their pure rust form through the `instance()` function, with maybe a bit of safety checking on pointers. Everything internal will always use the non-efiapi function. The sole purpose of the efiapi functions is to be registered with the system table. They should never be used outside of that. I tried to show that with my example:

```rust
impl <P: Platform> Core {
    // This function is public inside the crate. It can be called from anywhere that has access to the core
    // (inside the crate)
    pub(crate) fn trust(handle: efi::Handle, file: &efi::Guid) -> Result<(), EfiError> {
        let dispatcher = self.uefi_state.dispatcher_context.lock();
        for driver in dispatcher.pending_drivers.iter_mut() {
            if driver.firmware_volume_handle == handle && OrdGuid(driver.file_name) == OrdGuid(*file) {
                driver.security_status = efi::Status::SUCCESS;
                return Ok(());
            }
        }
        Err(EfiError::NotFound)
    }

    // This function is private and cannot be accessed outside the module. It should never be called. It's purpose
    // is to be registered with the system table and that's it.
    extern "efiapi" fn trust_efiapi(firmware_volume_handle: efi::Handle, file_name: *const efi::Guid) -> efi::Status {
        if file_name.is_null() {
        }

        let file_name = unsafe { file_name.read_unaligned() };

        match Self::instance().trust(firmware_volume_handle, &file_name) {
            Err(status) => status.into(),
            Ok(_) => efi::Status::SUCCESS,
        }
    }

    pub(crate) fn init_dxe_services(system_table: &mut EfiSystemTable) {
        ...
        trust: Self::trust_efiapi,
    }
}
```

docs/src/rfc/text/0000-core-static.md
cfernald: @Javagedes , just note that to get the full benefit of mockall (and avoid global locks) then the config struct would have to be instanced, right? i.e. `fn some_config(&self) -> bool`. That may be a pain to pass around.

docs/src/rfc/text/0022-core-static.md
Javagedes: In general, statics cannot have generics that are not specified in the crate because of how rust compilation works. That is why John's initial implementation of the dispatcher uses a dyn SectionExtractor and not compile time generics.

The overall issue that we are facing is that the Core is generic over "P", but "P" is selected by the platform implementor in their own crate. The way rust compiles, we compile all crates one by one up the dependency tree. This means we compile the patina-dxe-core while still not knowing what "P" is, so we get a compile time error.

That is why we have to type erase to a u8. Even with a OnceCell, I'll have to do something along the lines of `OnceCell<*const ()>`

docs/src/rfc/text/0000-core-static.md
os-d: I understood that the efiapi functions are just wrappers and that is good. I was just talking about where the functions live.

For  some functions, you would have to keep as self.<former_static>.func, right because you are relying on the `TplMutex` to protect the state, e.g. `self.gcd.add_memory_space()` would have to be like that, but you could have a wrapper as `self.add_memory_space()`. One thing about the current design I dislike is in some places we are four levels deep on wrappers :). I would prefer to prevent adding more wrappers to the top level.

And if we already have some functions that have to be lower in the struct hierarchy, it would feel odd to me to call

```rust
self.gcd.something_locked();
self.something_that_is_gcd_related_but_doesn't_need_locking();
```

Feels like it would make a bit more logical sense to just group that under the former static. But I'm not hard and fast on that, just musing.

For your first point, that is good that the impls can't be referenced outside of where they are defined.

docs/src/rfc/text/0000-core-static.md
makubacki: I agree that it's subtle but a little more intuitive.

docs/src/rfc/text/0022-core-static.md
Javagedes: However regarding this, OnceCell and Cell are not `send` / `sync` which means I would have to create a wrapper type. Something like:

```rust
struct Wrapper(OnceCell<*const ()>);

unsafe impl Send on Wrapper { }
```

I'm not sure I really want to deal with all of that boilerplate when the AtomicPointer works well. But I can if yo prefer.

joschock: I'm agnostic about the raw pointer wrapper mechanism, but I think `OnceCell`/`Cell` might be UB since Core might be shared between different contexts (e.g. efiapi func_a called at TPL_APP, efiapi func_b called at TPL_NOTIFY, both reconstruct Core from raw pointer). In other words, I think we need `Core`  to be `Sync`, because the only reason the compiler isn't complaining about `Sync` here is that it can't tell we are making shared aliases of it in different contexts from the unsafe raw pointer deref.

docs/src/rfc/text/0000-core-static.md
makubacki: I understand this has nothing to do with components, but I consider the core to have boundaries and subsystems within those boundaries. And I'm not sure if a single "core" setting is going to always be the same value for all of those subsystems. My understanding is that is an implicit assumption in this design. For example, both of these picking a common `setting_a` value:

```rust
GCD.setting_a(P::setting_a);
DISPATCHER.setting_a(P::setting_a);
```

Even as a platform owner, I find it a little clearer to know how what I'm specifying is being used in the core. Something like:

```rust
trait GcdConfig {
    fn prioritize_32_bit_memory() -> bool { false }
}

trait CoreConfig {
    type PlatformGcdConfig: GcdConfig;

    fn gcd_config() -> &'static Self::PlatformGcdConfig;
    ...
}
```

Consuming in the core (quick example, might have wrong syntax):

```rust
fn start(&'static mut self, phsical_hob_list: *const c_void) {
    ...
    let gcd_config = C::gcd_config();

    GCD.prioritize_32_bit_memory(gcd_config.prioritize_32_bit_memory());
...
```

---

As a consumer, I am assured that this is only impacting the GCD. If something else comes up in the future that uses `prioritize_32_bit_memory()` in the core. I can independently control that setting if needed. If the "GCD" concept is removed from the core, I can safely drop my GCD config settings without concern if other parts of the core were using those settings.

---

This is not a big deal though; I just wanted to raise the fact that my main concern is the opaque configuration surface itself. Like I said, I'm not sure if this will be an issue in practice or not.

docs/src/rfc/text/0022-core-static.md
os-d: Right, my point though was I don't think *Cell would be UB because we are wrapping anything that cannot be safely concurrently accessed with TplMutex, which enforce our guarantees here.

But, like I said, this was for my understanding, because to me it seemed like a case to use *Cell since we didn't actually want atomic behavior here. I'm okay if it stays as AtomicPtr if that makes things easier.

makubacki: I'm also fine with the atomic pointer, but I think that there should be a code comment in the implementation explaining why it is used.

joschock: Well, I'm less concerned about how we're accessing the existing fields folded into the global static (as they already have concurrency protections), than the internal aspects of `*Cell` not being safe to alias. In other words, I think calling `Cell::get()` while another `Cell:get()` on the same Cell (constructed from the same raw pointer) is in progress in a different context is UB, regardless of whether the fields inside the cell are safe to access in multiple contexts. We have to get to the concurrent-safe fields of the core via concurrent-safe methods, is my point.

docs/src/rfc/text/0000-core-static.md
Javagedes: I was mainly envisioning that only the pure rust UEFI services table functions were the ones that were registered with the `Core`, but your example in the above comment does make sense and can be confusing.

I am OK with not having all the UEFI services directly implemented on `Core`, but on the underlying structure.

I think my only comment is that even right now we just call global functions that are spread throughout the codebase, which also don't have any indication of what its related to (e.g. `core_allocate_pages` is in the `allocator.rs` file, focused on all of the different static allocators, but also ultimately links back to the GCD static, but then we have functions like `set_memory_space_attributes` that are purely associated with the GCD, but are for some reason in `dxe_services.rs` and not gcd.rs.

I think this is something where I'll update the RFC to mention that these functions will be associated where best seems fit, such as associated directly with the static, but leave that to be an implementation detail when I put up the PR. I'll update my example using the dispatcher to associate these functions directly against the dispatcher.

docs/src/rfc/text/0022-core-static.md
Javagedes: If it helps in terms of UB, the only thing that will be accessed via the `AtomicPtr<T>` are the values that were already previously static and have their own internal mutability wrappers (mainly the TplMutex). That is specifically why the `instance()` function only returns a reference, not a mutable reference.

If it helps, I could always point the pointer at the `UefiState` struct. But that would involve moving all statics into the `UefiState` struct, which goes against what was already discussed with @cfernald regarding some things (like the alloator) are more "kernal" like and do not necessary make since in the `UefiState` struct.

joschock: Agree that AtomicPtr doesn't have UB that the single-threaded `*Cell` structs would in this context. I think my further point would be that what is rebuilt from the AtomicPtr should also be `Sync` even if that's an `unsafe impl` we put there ourselves with a safety comment to emphasize that the struct is shared between contexts.

cfernald: I think that this is such a small amount of code that can be so tightly controlled that the specifics are somewhat arbitrary so long as they are well inspected and performant, and are ultimately an implementation detail. Even a simple `UnsafeCell<Option<NonNull<u8>>>` is no less safe them AtomicPointer so long as we handle our checks appropriately with some assumptions of single-threaded-ness on the initial set, especially implementing a highly restricted OnceLock. 

In any case, unless we think this is unsurmountable, I feel the specifics of the storage are better left for the code review and the mechanism itself (getting the singelton reference back from a static function) is what should be discussed here.

docs/src/rfc/text/0000-core-static.md
vineelko: NIT: Not necessarily for this PR, May be `efi_main` instead of `entry_point` because that's the default name used for uefi targets and we could probably get rid of `#[cfg_attr(target_os = "uefi", unsafe(export_name = "efi_main"))]`

docs/src/rfc/text/0000-core-static.md
Javagedes: Updated traits to be similar to what john mentioned, with the slight modification that the component info is an associated type of the Platform trait:

```rust
/// A trait to be implemented by the platform to register additional components, configurations, and services.
///
/// Allocations are available when these callbacks are invoked.
trait ComponentInfo {
    /// An optional platform callback to register components with the core.
    fn components<'a>(_add: &mut Add<'a, Component>) { }

    /// An optional platform callback to register configurations with the core.
    fn configs<'a>(_add: &mut Add<'a, Config>) { }

    /// An optional platform callback to register services with the core.
    fn services<'a>(_add: &mut Add<'a, Service>) { }
}

/// A trait to be implemented by the platform to provide configuration values and types to be used directly by the
/// Patina DXE Core.
trait Platform {
     /// The Platform component information.
    type ComponentInfo: ComponentInfo;

    /// The platform's section extractor type, used when extracting sections from firmware volumes.
    type Extractor: SectionExtractor;

    /// If true, the platform prioritizes allocating 32-bit memory when not otherwise specified.
    fn prioritize_32_bit_memory() -> bool { false }

    /// Returns an instance of the platform's section extractor when parsing firmware volumes.
    fn section_extractor() -> Self::Extractor

    /// Specifies teh GIC base addresses for AARCH64 systems.
    #[cfg(target_arch = "aarch64")]
    fn gic_bases() -> GicBases
}
```

This allows the core to continue to be simple:

```rust
static CORE: Core<Q35> = Core::new();
```

Additionally, loosened the wording to say that we may continue to sub-trait (using type aliases similar to `ComponentInfo` should the need arise, as Michael is concerned. For now, I do not plan on subtyping the minimal options we have.

docs/src/rfc/text/0022-core-static.md
```diff
@@ -0,0 +1,438 @@
...
+trait Platform {
+     /// The Platform component information.
+    type ComponentInfo: ComponentInfo;
+
+    /// The platform's section extractor type, used when extracting sections from firmware volumes.
+    type Extractor: SectionExtractor;
+
+    /// If true, the platform prioritizes allocating 32-bit memory when not otherwise specified.
+    fn prioritize_32_bit_memory() -> bool { false }
+
```
cfernald, line 167: could we add an example of a non-optional config that doesn't have a default? Not a big deal, but I think it's a noteworthy design point.

docs/src/rfc/text/0000-core-static.md
Javagedes: Loosened the wording in the RFC. The pure rust methods will now be implemented on the owning subsystem, however the "efiapi" function must continue to be implemented on the `Core` so that it can call `Core::instance()`.

Additionally, added wording regarding moving all the `efiapi` functions into a module that is not public. This will prevent anyone in patina from calling the `efiapi` function, when they should be calling the pure rust implementation from the subsystem.

docs/src/rfc/text/0000-core-static.md
Javagedes: We split into two distinct traits. One is the Platform trait for platform customizations, and one is the `ComponentInfo` trait, which has all of the component registration logic.

Traits are documented like any other code, so I think documenting the type and functions in the trait should suffice on this. Closing this, but feel free to review what I documented them as in the RFC. Feel free to open another issue there if anything is not sufficient.

docs/src/rfc/text/0022-core-static.md
```diff
@@ -0,0 +1,438 @@
...
+/// Patina DXE Core.
+trait Platform {
+     /// The Platform component information.
+    type ComponentInfo: ComponentInfo;
+
+    /// The platform's section extractor type, used when extracting sections from firmware volumes.
+    type Extractor: SectionExtractor;
+
+    /// If true, the platform prioritizes allocating 32-bit memory when not otherwise specified.
+    fn prioritize_32_bit_memory() -> bool { false }
```
cfernald, line 166: It doesn't need to be solved here, but i'd be curious to understand the impact of a const vs a function when it comes to compiler optimizations. For example,

```rust
if P::do_complicated_thing() {
    // complicated and large logic
}
```
vs
```rust
if P::DO_COMPLICATED_THING {
    // complicated and large logic
}
```

in both cases does the compiler reliably drop this logic when the config is `false`. I would hope that practically if the function unconditionally return a Boolean, that there would be no difference, but may have some implications if so. I suppose `#[inline(always)]` would potentially help as well if the compiler appears to fail to handle this as we would want.

cfernald (approved):

docs/src/rfc/text/0000-core-static.md
Javagedes: In this comment, they meant the `start` method implemented on the `Core` struct. not `_start` that is in the platform. That is a platform choice. Feel free to update it in Q35 / SBSA if you prefer.

As for this comment, I've updated `start` to `entry_point`.

docs/src/rfc/text/0022-core-static.md
Javagedes: Because it does not necessarily need to be static. Calling `Self::isntance().do_function()` does not require a static lifetime, just a lifetime for that line. So, we let the caller choose the lifetime.

Really this only matters when you also have mutable references. Having a function that borrows and give something statically can make it impossible to call anything else on it. But regardless of it being applicable, its still good practice to do it this way.

docs/src/rfc/text/0022-core-static.md
Javagedes: Yes. `GicBases` is now in this RFC, and it is not optional for AARCH64 platforms. Also, the `section_extractor` was added, and is required.

docs/src/rfc/text/0022-core-static.md
Javagedes: The const would definitely always allow the compiler to optimize out that entire code block if it was false. I think `P::do_complicated_thing()` would be iffy, especially because we cannot put `#[inline(always)]` on a trait interface, it would have to be the implementor.

The big blocker with using a const is we cannot use `automock` to have a general mock of it. Each tester would need to use `mockall::mock! { }` to create a mock with whatever const type they want.

docs/src/rfc/text/0022-core-static.md
Javagedes: Looks like @cfernald's comment closed this thread as an implementation detail that can be discussed in the PR implemetning this RFC.

I've added additional safety comments in the RFC around the usage of `Self::instance()`. Feel free to review and create a new issue with any concerns you have with the details of the singleton reference back from a static function. That is really the meat and butter that makes this all work. In my opinion, regarding safety, this is no different then any other static that we have spread throughout this codebase (and there is about 20 of them), so if we decide this is unsound, then its all unsound lmao.

If there is an issue with this singleton practice for the sake of being accessible inside "efiapi" functions only, then we need to decide now and see if there is another option. Otherwise, we will probably need to close out this RFC as denied.

docs/src/rfc/text/0022-core-static.md
cfernald: See it would help if I read more before commenting.

docs/src/rfc/text/0022-core-static.md
cfernald: I agree, I think just encouraging the practice of `[inline(always)]` for config functions when the platform is simply providing a deterministic bool would address this. We should probably add that to all the defaults at the very least.

os-d (approved):

makubacki (approved):
Approving as is, but I think this might be worth first demonstrating on a feature branch for feedback before merging straight to main.

Javagedes:
> Approving as is, but I think this might be worth first demonstrating on a feature branch for feedback before merging straight to main.

@makubacki I have an example of the first PR that will be merged, which is #1049. It is quite large as is, but I could throw on another commit on top of that with the second PR that will be merged, which brings in the `DISPATCHER_CONTEXT` and `PRIVATE_FV_DATA` into the core.

docs/src/rfc/text/0022-core-static.md
Javagedes: Added verbage in the RFC to state that documentation will be added to encourage usage of #[inline(always)] on applicable methods. Also added `#[inline(always)]` to our default implementations in the trait.

makubacki:
> @makubacki I have an example of the first PR that will be merged, which is #1049. It is quite large as is, but I could throw on another commit on top of that with the second PR that will be merged, which brings in the `DISPATCHER_CONTEXT` and `PRIVATE_FV_DATA` into the core.

That also works. The main point being a more concrete example of the design and impact on platform integration for additional feedback. If the example can serve that purpose and not further development is needed, then going directly to `main` seems fine.

Javagedes:
Moving to FCP.

PR #1042, opened 2025-11-11
Ensure that free memory has XP attribute set to match initial free memory attributes from GCD init.

joschock:
## Description

Free memory regions created at GCD init have XP attribute set. If these are then allocated and freed back, they would have the XP attribute clear. This would cause adjacent "free" regions to have different attributes. They would be coalesced in GetMemoryMap(), but allocations to the whole region behind that descriptor would fail because of the underlying attribute mismatch.

This PR adds the XP attribute on freed space to ensure that the attributes of free memory remain consistent.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on hardware platform and confirmed that free memory is correctly coalesced after this change. 

## Integration Instructions

N/A

### DIFF
--- a/patina_dxe_core/src/gcd/spin_locked_gcd.rs
+++ b/patina_dxe_core/src/gcd/spin_locked_gcd.rs
@@ -600,7 +600,7 @@ impl GCD {
         match self.set_gcd_memory_attributes(
             base_address,
             len,
-            efi::MEMORY_RP | (desc.attributes & efi::CACHE_ATTRIBUTE_MASK),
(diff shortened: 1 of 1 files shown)


os-d:
I am planning an update to these scattered individual rules to have a single trait that we call to say apply_rule_x and that way it is easily auditable what rules are being applied and where we are applying them. So both the hob processing and this free memory would call the same apply_free_memory_rule and be guaranteed to have the same attributes.

Issue #505, opened 2025-06-12
[Feature]: Remove Duplicate Memory Type Definitions


Javagedes:
### Feature Overview

Currently, the different memory types are defined in two locations in two separate ways -

As an enum in `patina_sd::efi_types::EfiMemoryType`
As const u32 wrappers in `patina_sdk::boot_services::allocations::MemoryType`

These should be combined, most likely using the enum `EfiMemoryType`, as we can simply shift the consts into an impl section of that enum.

### Solution Overview

These should be combined, most likely using the enum `EfiMemoryType`, as we can simply shift the consts into an impl section of that enum.

### Alternatives Considered

Keep Them separated

### Urgency

Low

### Are you going to implement the feature request?

Someone else needs to implement the feature

### Do you need maintainer feedback?

No maintainer feedback needed

PR #1044, opened 2025-11-11
Remove Duplicate MemoryType Definitions

Raymond-MS:
## Description

Removed duplicate MemoryTypes definition from allocation.rs. Updated the EfiMemoryTypes in efi_types.rs. Updated all instances using the old version to use EfiMemoryTypes instead. Fixed some issues related to casting by explicitly adding the Sized trait.

Fixes: #505 

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Built and ran to UEFI shell using QEMU SBSA.

## Integration Instructions

N/A

### DIFF
--- a/sdk/patina/src/boot_services.rs
+++ b/sdk/patina/src/boot_services.rs
@@ -2085,6 +2085,25 @@ mod tests {
     }
 
     #[test]
+    fn test_allocate_pool_for_type() {
+        let boot_services = boot_services!(allocate_pool = efi_allocate_pool);
+
+        extern "efiapi" fn efi_allocate_pool(
+            mem_type: efi::MemoryType,
+            _size: usize,
+            buffer: *mut *mut c_void,
+        ) -> efi::Status {
+            let expected_mem_type: efi::MemoryType = EfiMemoryType::MemoryMappedIO.into();
+            assert_eq!(mem_type, expected_mem_type);
+            unsafe { ptr::write(buffer, 0x55AA as *mut c_void) };
+            efi::Status::SUCCESS
+        }
+
+        let status = boot_services.allocate_pool_for_type(EfiMemoryType::MemoryMappedIO);
+        assert_eq!(status, Ok(0x55AA as *mut u8));
+    }
+
+    #[test]
     fn test_free_pool() {
         let boot_services = boot_services!(free_pool = efi_free_pool);
 


makubacki:
In the future, if you put just the PR number (e.g. `#505`) instead of the URL and place it on the same line as "Fixes", it will link the issue automatically. I updated it here.

sdk/patina/src/efi_types.rs
```diff
@@ -15,41 +15,43 @@ use crate::error::EfiError;
 
 /// A wrapper for the EFI memory types.
 #[derive(Debug, Copy, Clone, PartialEq, Eq)]
+#[repr(u32)]
+
 pub enum EfiMemoryType {
     /// Reserved memory for platform uses.
-    ReservedMemoryType,
+    ReservedMemoryType = 0,
```
berlin-with0ut-return: i don't think we need these assignments as (1) Rust will autoassign the labeled values with repr(u32) and (2) we don't actually care about the exact value of the memory types. please correct me if i'm wrong

os-d: I agree that we don't need the assignments, but we do care about the exact values, these need to match the UEFI spec because we report them in the EFI_MEMORY_MAP.

os-d: To that end, having what the other implementation had, where it was an explicit wrapper of the r-efi ones wouldn't be a bad approach either, to make it clear.

sdk/patina/src/efi_types.rs
```diff
@@ -15,41 +15,43 @@ use crate::error::EfiError;
 
 /// A wrapper for the EFI memory types.
 #[derive(Debug, Copy, Clone, PartialEq, Eq)]
+#[repr(u32)]
+
 pub enum EfiMemoryType {
```
os-d, line 19: The previous one implemented a From for u32, I think that is still valuable to have.

Raymond-MS: I think this already exists from what I can tell, please correct me if I am wrong. See Line#102 in this file.

os-d: Those are good. I don't think we need the generic u32 one

sdk/patina/src/efi_types.rs
```diff
@@ -15,41 +15,43 @@ use crate::error::EfiError;
 
 /// A wrapper for the EFI memory types.
 #[derive(Debug, Copy, Clone, PartialEq, Eq)]
+#[repr(u32)]
+
```
makubacki:
```suggestion
```

sdk/patina/src/efi_types.rs
makubacki: I think that was better. It's more explicit and relies on one place (the r_efi crate) getting the spec definitions right so we just reuse them

makubacki:
@Raymond-MS, I see there are clippy errors at the moment. If you run `cargo make all` locally, that should get everything in good shape for CI in the PR.

makubacki (approved):

makubacki:
@Raymond-MS, the patch code coverage is 75% right now, you need 80% to merge. Can you please check the codecov report and see if you can cover some additional lines missing coverage to get >80%?

sdk/patina/src/efi_types.rs
```diff
@@ -15,41 +15,42 @@ use crate::error::EfiError;
 
 /// A wrapper for the EFI memory types.
 #[derive(Debug, Copy, Clone, PartialEq, Eq)]
+#[repr(u32)]
 pub enum EfiMemoryType {
     /// Reserved memory for platform uses.
-    ReservedMemoryType,
+    ReservedMemoryType = efi::RESERVED_MEMORY_TYPE,
```
cfernald, line 21: with this change, you should be able to simplify `impl From<EfiMemoryType> for efi::MemoryType` below to simply return the value in all cases. 

Similarly, from_efi could be simplified as well, but we want to keep the error checking that it is within the expected ranges.

sdk/patina/src/boot_services.rs
```diff
@@ -313,10 +313,10 @@ pub trait BootServices {
     /// Allocates pool memory.
     ///
     /// [UEFI Spec Documentation: 7.2.4. EFI_BOOT_SERVICES.AllocatePool()](https://uefi.org/specs/UEFI/2.10/07_Services_Boot_Services.html#efi-boot-services-allocatepool)
-    fn allocate_pool(&self, pool_type: MemoryType, size: usize) -> Result<*mut u8, efi::Status>;
+    fn allocate_pool(&self, pool_type: EfiMemoryType, size: usize) -> Result<*mut u8, efi::Status>;
 
     /// Allocates pool memory casted as given type.
-    fn allocate_pool_for_type<T: 'static>(&self, pool_type: MemoryType) -> Result<*mut T, efi::Status> {
+    fn allocate_pool_for_type<T: 'static + Sized>(&self, pool_type: EfiMemoryType) -> Result<*mut T, efi::Status> {
```
cfernald, line 319: Perhaps dumb question, but I thought Sized was implied and didnt need to be specified.

makubacki: Because this is not actually an issue in compilation, I've suspected it is related to rust-analyzer's type inference and explicitly including `Sized` is not harmful, so it's reasonable.

sdk/patina/src/efi_types.rs
Raymond-MS: As per our conversation, due to the custom types, it is probably best we keep this as is. If we figure a better way to handle it we can address it then. Thanks!

PR #1045, opened 2025-11-11
Update child handle validation in core_disconnect_controller() [Rebase & FF]

makubacki:
## Description

The first commit is the actual change and the second consolidates and cleans up related test code.

---

**Update child handle validation in core_disconnect_controller()**

Previously, when a specific child handle was requested but was not a
child created by a driver managing the controller, that driver was not
skipped. When the driver had no children its `Stop()` was still
invoked, and the function reported success even though nothing was
disconnected.

This change:

1. Skips a driver when a specific child handle is requested but is not
   present in that driver's child list (checked via `child_handles`
   being empty after filtering). When no managing driver owns the
   child, the call returns `NotFound`.

2. Uses `total_children == child_handles.len()` to detect that the
   requested child was the driver's only child, in which case the
   driver itself is disconnected (`Stop()` with zero children). This
   matches the UEFI spec's "if ChildHandle is the only child ... the
   driver will be disconnected" behavior.

Three unit tests are added:

- `test_disconnect_specific_child_not_managed_by_driver()`: Checks
  that a specific child not owned by the single managing driver
  returns `NotFound` and `Stop()` is never called.

- `test_disconnect_specific_child_among_multiple_children()`: Checks
  that a specific child among several is destroyed (`Stop()` called
  with that one child) while the driver-level `Stop()` is not called,
  so the driver stays connected.

- `test_disconnect_specific_child_not_owned_by_either_driver()`: Checks
  that when two drivers are managing the controller and a child owned
  by neither is requested, that both drivers are skipped and `NotFound`
  is returned.

Note: `EFI_NOT_FOUND` is not listed as a return value in for
`DisconnectController()` in the UEFI 2.11 spec (section 7.3.13), but
returning it when the requested child is unmanaged matches EDK II
behavior.

---

**patina_dxe_core: Consolidate disconnect controller test setup logic**

The child-specific `core_disconnect_controller()` tests each repeated
a lot of related logic to install a driver binding, mark the driver as
managing the controller, and register its children.

This commit adds two test helpers, `new_test_handle()` and
`setup_driver_managing_controller()` to reduce the overall size of
test functions.

---

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- Run all unit tests
- Check that `test_disconnect_specific_child_not_managed_by_driver()` fails without the change and passes with it
- QEMU boot to EFI shell

## Integration Instructions

- N/A

### DIFF
--- a/patina_dxe_core/src/driver_services.rs
+++ b/patina_dxe_core/src/driver_services.rs
@@ -531,12 +531,15 @@ pub unsafe fn core_disconnect_controller(
         child_handles.retain(|x| child_set.insert(*x));
 
         let total_children = child_handles.len();
-        let mut is_only_child = false;
         if let Some(handle) = child_handle {
-            // if the child was specified, but was the only child, then the driver should be disconnected.
-            // if the child was specified, but other children were present, then the driver should not be disconnected.
-            child_handles.retain(|x| x == &handle);
-            is_only_child = total_children == child_handles.len();
+            // A specific child was requested so keep only that child. `child_handles` was deduplicated
+            // above, so this leaves at most one entry.
+            child_handles.retain(|x| *x == handle);
+
+            // If the requested child isn't managed by this driver, skip it.
+            if child_handles.is_empty() {
+                continue;
+            }
         }
 
         // resolve the handle to the driver_binding.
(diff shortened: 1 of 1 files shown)


os-d (approved):

joschock (approved):

Javagedes (approved):

makubacki:
@Javagedes, @os-d, @joschock, I finally returned to rebase this PR and clean up some test logic. Can someone please review again?

joschock (approved):

PR #1048, opened 2025-11-12
GCD: Increase Unit Test Coverage

### LINKED ISSUE #202: [Bug]: GCD Tests Use Invalid Addresses/Lengths

### Current Behavior

The existing GCD tests use non-page aligned addresses and lengths and the tests are not always testing real scenarios. These tests break when we add better correctness to the GCD. There are a lot of instances of this, so the tests have been fixed up where possible, but need to be revisited and reworked.

### Expected Behavior

The GCD tests should use page aligned addresses and lengths and test real scenarios (i.e. bits are valid for capabilities, etc.).

### Steps To Reproduce

_No response_

### Build Environment

```markdown
- OS(s):
- Tool Chain(s):
- Targets Impacted:
```

### Version Information

```text

```

### Urgency

None

Someone else needs to fix it

None

### Anything else?

_No response_

os-d:
## Description

This commit adds more robust unit testing to the GCD code, moving the spin_locked_gcd.rs file from 68% coverage to 90% coverage.

Some functions are marked with #[coverage(off)] because they are wrapper functions or otherwise not useful/possible to meaningfully test.

Closes #202 .

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Unit tests pass.

## Integration Instructions

N/A.

### DIFF
--- a/patina_dxe_core/src/gcd/spin_locked_gcd.rs
+++ b/patina_dxe_core/src/gcd/spin_locked_gcd.rs
@@ -28,7 +28,7 @@ use crate::{
     GCD, allocator::DEFAULT_ALLOCATION_STRATEGY, ensure, error, events::EVENT_DB, protocol_db,
     protocol_db::INVALID_HANDLE, tpl_lock,
 };
-use patina_internal_cpu::paging::{CacheAttributeValue, PatinaPageTable, create_cpu_paging};
+use patina_internal_cpu::paging::{CacheAttributeValue, PatinaPageTable};
 use patina_paging::{MemoryAttributes, PtError, page_allocator::PageAllocator};
 
 use patina::pi::hob::{Hob, HobList};
@@ -181,13 +181,13 @@ type GcdFreeFn =
     fn(gcd: &mut GCD, base_address: usize, len: usize, transition: MemoryStateTransition) -> Result<(), EfiError>;
 
 #[derive(Debug)]
-struct PagingAllocator<'a> {
+pub(crate) struct PagingAllocator<'a> {
     page_pool: Vec<efi::PhysicalAddress>,
     gcd: &'a SpinLockedGcd,
 }
 
 impl<'a> PagingAllocator<'a> {
-    fn new(gcd: &'a SpinLockedGcd) -> Self {
+    pub(crate) fn new(gcd: &'a SpinLockedGcd) -> Self {
         Self { page_pool: Vec::with_capacity(PAGE_POOL_CAPACITY), gcd }
     }
 }
@@ -554,6 +554,7 @@ impl GCD {
         }
     }
 
+    #[coverage(off)]
     fn allocate_memory_space_null(
         _gcd: &mut GCD,
         _allocate_type: AllocateType,
@@ -619,6 +620,7 @@ impl GCD {
         }
     }
 
+    #[coverage(off)]
     fn free_memory_space_worker_null(
         _gcd: &mut GCD,
         _base_address: usize,
@@ -1808,6 +1810,7 @@ impl SpinLockedGcd {
     /// Creates a new uninitialized GCD. [`Self::init`] must be invoked before any other functions or they will return
     /// [`EfiError::NotReady`]. An optional callback can be provided which will be invoked whenever an operation
     /// changes the GCD map.
+    #[coverage(off)]
     pub const fn new(memory_change_callback: Option<MapChangeCallback>) -> Self {
         Self {
             memory: tpl_lock::TplMutex::new(
@@ -1868,6 +1871,7 @@ impl SpinLockedGcd {
         unsafe { self.memory.lock().init_memory_blocks(memory_type, base_address, len, capabilities) }
     }
 
+    #[coverage(off)]
     pub fn prioritize_32_bit_memory(&self, value: bool) {
         self.memory.lock().prioritize_32_bit_memory = value;
     }
@@ -2068,11 +2072,10 @@ impl SpinLockedGcd {
     // as we need to allocate memory for the page table structure.
     // This function always uses the GCD functions to map the page table so that the GCD remains in sync with the
     // changes here (setting XP)
-    pub(crate) fn init_paging(&self, hob_list: &HobList) {
+    pub(crate) fn init_paging_with(&self, hob_list: &HobList, page_table: Box<dyn PatinaPageTable>) {
         log::info!("Initializing paging for the GCD");
 
-        let page_allocator = PagingAllocator::new(&GCD);
-        *self.page_table.lock() = Some(create_cpu_paging(page_allocator).expect("Failed to create CPU page table"));
+        *self.page_table.lock() = Some(page_table);
 
         // this is before we get allocated descriptors, so we don't need to preallocate memory here
         let mut mmio_res_descs: Vec<dxe_services::MemorySpaceDescriptor> = Vec::new();
@@ -2276,6 +2279,7 @@ impl SpinLockedGcd {
     ///
     /// # Documentation
     /// UEFI Platform Initialization Specification, Release 1.8, Section II-7.2.4.4
+    #[coverage(off)]
     pub fn remove_memory_space(&self, base_address: usize, len: usize) -> Result<(), EfiError> {
         let result = self.memory.lock().remove_memory_space(base_address, len);
         if result.is_ok() {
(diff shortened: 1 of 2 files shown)


patina_dxe_core/src/gcd/spin_locked_gcd.rs
```diff
@@ -40,6 +40,188 @@ use super::{
...
+
+    Box::new(SharedMockPageTable::new(mock))
+}
+
+#[cfg(test)]
+fn get_shared_mock_page_table() -> Option<Arc<MockPageTable>> {
+    let storage = SHARED_MOCK_PAGE_TABLE.lock().unwrap();
+    storage.clone()
+}
+
```
vineelko: Should this be moved to `mod test {..}` so `#[cfg(test)]` can be removed?

os-d: I'll see if it can, the problem is the actual code has to use this in the test case. But maybe not all of this

os-d: I was able to move almost all of this

vineelko:
I think breaking

`pub(crate) fn init_paging(&self, hob_list: &HobList)`
into
`fn init_paging_with(&self, hob_list: &HobList, page_table: Box<dyn PatinaPageTable>)`

would significantly simplify the test code setup. The normal code would continue to call `init_paging()` as usual, which would invoke `create_cpu_paging()` from the `patina_internal_cpu::paging` crate and then forward the call to `init_paging_with()`. But the test code could directly call `init_paging_with(.., mocked_page_table)` with a mocked page table, eliminating the need for hacks in the `use` statements.

After that, the second problem of sharing the mocked page object between the real code and test code can also be solved without relying on `Arc`, global locks, *mut pointers, unsafe. Instead of wrapping every inner field in a `RefCell`, we can declare the inner as `inner: Rc<RefCell<..>>`, similar to what we did in the paging crate’s test code.

patina_dxe_core/src/gcd/spin_locked_gcd.rs
```diff
@@ -4399,4 +4595,661 @@ mod tests {
...
+            assert_eq!(bad_alloc, Err(PtError::InvalidParameter), "Should fail to allocate with bad alignment");
+
+            // fail to allocate a zero sized page
+            let zero_alloc = allocator.allocate_page(UEFI_PAGE_SIZE as u64, 0, false);
+            assert_eq!(zero_alloc, Err(PtError::InvalidParameter), "Should fail to allocate zero sized page");
+        });
+    }
+
+    #[test]
+    #[should_panic]
```
Javagedes, line 4673: nit:


```suggestion
    #[should_panic(expected = "<something>"]
```

patina_dxe_core/src/gcd/spin_locked_gcd.rs
```diff
@@ -4399,4 +4595,661 @@ mod tests {
...
+                    reserved: [0; 4],
+                },
+                module_name: guids::DXE_CORE,
+                entry_point: dxe_core_base as u64 + 0x1000,
+            });
+            let mut hob_list = HobList::new();
+            hob_list.push(hob);
+
+            // SAFETY: We just allocated this memory, write mock PE header data directly to memory
+            let pe_header_data = [
```
Javagedes, line 4821: even though you only use this once, this might be better as test collateral. We also already have some pe images in there. Is this one special or could you just re-use the exiting pe?

os-d:
> I think breaking
> 
> `pub(crate) fn init_paging(&self, hob_list: &HobList)` into `fn init_paging_with(&self, hob_list: &HobList, page_table: Box<dyn PatinaPageTable>)`
> 
> would significantly simplify the test code setup. The normal code would continue to call `init_paging()` as usual, which would invoke `create_cpu_paging()` from the `patina_internal_cpu::paging` crate and then forward the call to `init_paging_with()`. But the test code could directly call `init_paging_with(.., mocked_page_table)` with a mocked page table, eliminating the need for hacks in the `use` statements.
> 
> After that, the second problem of sharing the mocked page object between the real code and test code can also be solved without relying on `Arc`, global locks, *mut pointers, unsafe. Instead of wrapping every inner field in a `RefCell`, we can declare the inner as `inner: Rc<RefCell<..>>`, similar to what we did in the paging crate’s test code.

I think that's fine, it is just kicking the testing can up the stack, but in this case we have the spin locked GCD wrapper for this function that I've already marked as coverage off, so we can do this there.

patina_dxe_core/src/gcd/spin_locked_gcd.rs
os-d: Yeah, I thought about putting this somewhere else, decided I'd wait and see if anyone complained :). Nothing special except that it is DXE core's GUID and I had to have copilot do a lot of shifting of offsets to make it fit within the allocated range. I'll check and see what we already have and either move this to the collateral or use something existing

patina_dxe_core/src/gcd/spin_locked_gcd.rs
os-d: Ah thanks, I was wondering if that existed. I also just filed an issue to disable debug_asserts in test code.

os-d:
> > I think breaking
> > `pub(crate) fn init_paging(&self, hob_list: &HobList)` into `fn init_paging_with(&self, hob_list: &HobList, page_table: Box<dyn PatinaPageTable>)`
> > would significantly simplify the test code setup. The normal code would continue to call `init_paging()` as usual, which would invoke `create_cpu_paging()` from the `patina_internal_cpu::paging` crate and then forward the call to `init_paging_with()`. But the test code could directly call `init_paging_with(.., mocked_page_table)` with a mocked page table, eliminating the need for hacks in the `use` statements.
> > After that, the second problem of sharing the mocked page object between the real code and test code can also be solved without relying on `Arc`, global locks, *mut pointers, unsafe. Instead of wrapping every inner field in a `RefCell`, we can declare the inner as `inner: Rc<RefCell<..>>`, similar to what we did in the paging crate’s test code.
> 
> I think that's fine, it is just kicking the testing can up the stack, but in this case we have the spin locked GCD wrapper for this function that I've already marked as coverage off, so we can do this there.

I did this, but all we really did was drop an `A` :). I.e. we went from `Arc` to `Rc`, so I'm not sure how big of a difference that really was, but it did constrain it all to the test module and let us drop the globals, so that is good.

patina_dxe_core/src/gcd/spin_locked_gcd.rs
os-d: I decided to leave this in place because it is specific to this test, to make the init_paging manipulations of it work.

patina_dxe_core/src/gcd/spin_locked_gcd.rs
os-d: Also not doing this, because with_locked_state captures this and always has the same error

image

Let's fix this the right way and disable debug_asserts for tests and actually test production flows.

vineelko:
> > > I think breaking
> > > `pub(crate) fn init_paging(&self, hob_list: &HobList)` into `fn init_paging_with(&self, hob_list: &HobList, page_table: Box<dyn PatinaPageTable>)`
> > > would significantly simplify the test code setup. The normal code would continue to call `init_paging()` as usual, which would invoke `create_cpu_paging()` from the `patina_internal_cpu::paging` crate and then forward the call to `init_paging_with()`. But the test code could directly call `init_paging_with(.., mocked_page_table)` with a mocked page table, eliminating the need for hacks in the `use` statements.
> > > After that, the second problem of sharing the mocked page object between the real code and test code can also be solved without relying on `Arc`, global locks, *mut pointers, unsafe. Instead of wrapping every inner field in a `RefCell`, we can declare the inner as `inner: Rc<RefCell<..>>`, similar to what we did in the paging crate’s test code.
> > 
> > 
> > I think that's fine, it is just kicking the testing can up the stack, but in this case we have the spin locked GCD wrapper for this function that I've already marked as coverage off, so we can do this there.
> 
> I did this, but all we really did was drop an `A` :). I.e. we went from `Arc` to `Rc`, so I'm not sure how big of a difference that really was, but it did constrain it all to the test module and let us drop the globals, so that is good.

Sorry for the delay, Thanks for the change, Oliver, 

I think you did much more than dropping 'A' :-)
1. You got rid of the `use` hack
2. Introduced the `_with` method, 
3. Deleted `setup_shared_mock_page_table()` and `get_shared_mock_page_table()`
4. Remove the unsafe blocks around `inner_ptr`
5. Dropped the unnecessary global locks.

One other change we could have included is dropping the `RefCell` from all the fields of `MockPageTable`. Since the inner value is already behind a `RefCell`, we don’t need field level `RefCell`. 

But this is a huge improvement. Thanks.

Issue #22, opened 2024-10-03
Initial Release - Release Process - Automated release notes


makubacki:
Automated GitHub release notes generated with changes categorized.

Javagedes:
Partial completion of this task in #81 which sets up the automated semantic versioning and release notes. @makubacki has stated some changes need to be made before this can be marked complete.

makubacki:
Ready to close now

PR #1049, opened 2025-11-13
[REBASE&FF] Move the `Core` to be 'static for the platform binary

Javagedes:
## Description

This pull request implements the major breaking changes that will be introduced as a part of [RFC #22](https://github.com/OpenDevicePartnership/patina/blob/main/docs/src/rfc/text/0022-core-static.md). Overall, this pull request does the following things:

1. This commit moves platform configuration of the core into the
   `PlatformInfo` trait. This trait has multiple type associations for
   configuring the different subsystems of the core. As of this commit
   the following type associations exist:
   - `ComponentInfo` used to attach and configure patina components to
     be dispatched by the core.
   - `CpuInfo` used to configure cpu releated functionality such as the
     GIC Bases for AARCH64 systems.
   - `MemoryInfo` used to configure memory related functionality such as
     prioritizing 32 bit memory allocations.
2. It requires the core to be `static` within the platform binary and simplifies the initialization process for the platform.
3. Updates all documentation to reflect the new interface

### Side Effects

Due to the changes listed above, there are some minor side effects that have occurred:

1. Initialization of the advanced logger (To get the address) is now implemented on the Logger itself, and not the component (and renamed to `init`).
2. Patina component dispatch has been moved into its own subsystem with its own TPL Mutex
3. The `Decompress` trait is no longer a service but rather specified as an associated type by the `Plaform` trait. `SectionExtractorNull` has been added back.
4. `GicBases` is no longer a config, but a required `Platform` trait method for AARCH64 systems that will cause a compilation error if not provided
5. The timer frequency is no longer a config, but an optional `Platform` trait method.

Reference Q35 / SBSA implementation: https://github.com/OpenDevicePartnership/patina-dxe-core-qemu/pull/74

- [x] Impacts functionality?
- [ ] Impacts security?
- [x] Breaking change?
- [x] Includes tests?
- [x] Includes documentation?

## How This Was Tested

Boot to Shell on Q35, SBSA

### `PlatformInfo` trait

Platforms must now implement the `PlatformInfo` trait which describes the platform customizations for the Core. This trait has multiple type associations to provide customizations to the different subsystems of the core. They can be implemented on the same type (example below), or implemented on separate types to support re-use for different platforms. See below:

```rust
use patina_dxe_core::*;
use patina_adv_logger::components

struct Q35;

impl ComponentInfo for Q35 {
    fn components(mut add: Add<Component>) {
        add.component(q35_services::mm_config_provider::MmConfigurationProvider);
        add.component(q35_services::mm_control::QemuQ35PlatformMmControl::new());
        add.component(patina_mm::component::sw_mmi_manager::SwMmiManager::new());
    }

    fn configs(mut add: Add<Config>) {
        add.config(patina_mm::config::MmCommunicationConfiguration {
                acpi_base: patina_mm::config::AcpiBase::Mmio(0x0), // Actual ACPI base address will be set during boot
                cmd_port: patina_mm::config::MmiPort::Smi(0xB2),
                data_port: patina_mm::config::MmiPort::Smi(0xB3),
                comm_buffers: vec![],
            });
    }
}

impl MemoryInfo for Q35 { }

impl CpuInfo for Q35 { }

impl Platform for Q35 {
    type MemoryInfo = Self;
    type ComponentInfo = Self;
    type CpuInfo = Self;
    type Extractor = CompositeSectionExtractor;
}
```

### `Core` must now be 'static 

The core must now be initialized as a static. Its interface has now also changed. Building off the above example, you would add:

```rust

static CORE: Core<Q35> = Core::new(CompositeSectionExtractor::new());

#[cfg_attr(target_os = "uefi", unsafe(export_name = "efi_main"))]
pub extern "efiapi" fn _start(physical_hob_list: *const c_void) -> ! {
    CORE.entry_point(physical_hob_list)
}
```

### Advanced Logger init changes

The `init_advanced_logger` method has moved to the static logger struct and is renamed to `init`, rather than the component.

```rust
static LOGGER: AdvancedLogger<Uart16550> = AdvancedLogger::new(
    Format::Standard,
    &[
        ("goblin", log::LevelFilter::Off),
        ("gcd_measure", log::LevelFilter::Off),
        ("allocations", log::LevelFilter::Off),
        ("efi_memory_map", log::LevelFilter::Off),
    ],
    log::LevelFilter::Info,
    Uart16550::Io { base: 0x402 },
);

#[cfg_attr(target_os = "uefi", unsafe(export_name = "efi_main"))]
pub extern "efiapi" fn _start(physical_hob_list: *const c_void) -> ! {
      // SAFETY: The physical_hob_list pointer is considered valid at this point as it's provided by the core
    // to the entry point.
    unsafe { LOGGER.init(physical_hob_list).unwrap(); }
    log::set_logger(&LOGGER).map(|()| log::set_max_level(log::LevelFilter::Trace)).unwrap();

    CORE.entry_point(physical_hob_list)
}
```

### DIFF
--- a/patina_dxe_core/src/component_dispatcher.rs
+++ b/patina_dxe_core/src/component_dispatcher.rs
@@ -222,6 +222,7 @@ impl ComponentDispatcher {
     }
 
     /// Logs all components that were not dispatched, and the parameter that was not satisfied that prevented dispatch.
+    #[coverage(off)]
     pub(crate) fn display_not_dispatched(&self) {
         if !self.components.is_empty() {
             let name_len = "name".len();
--- a/patina_dxe_core/src/lib.rs
+++ b/patina_dxe_core/src/lib.rs
@@ -96,6 +96,7 @@ mod tpl_mutex;
 #[cfg(test)]
 pub use component_dispatcher::MockComponentInfo;
 pub use component_dispatcher::{Add, Component, ComponentInfo, Config, Service};
+use spin::Once;
 
 #[cfg(test)]
 #[macro_use]
@@ -104,6 +105,7 @@ pub mod test_support;
 
 use core::{
     ffi::c_void,
+    num::NonZeroUsize,
     ptr::{self, NonNull},
     str::FromStr,
 };
@@ -314,6 +316,12 @@ impl GicBases {
     }
 }
 
+/// Static reference to the DXE Core instance in the compiled binary.
+///
+/// This is set during the `entry_point` call of the DXE Core and is used to provide static access to the core for use
+/// only in efiapi functions where no reference to the core is otherwise available.
+static __SELF: Once<NonZeroUsize> = Once::new();
+
 /// Platform configured DXE Core responsible for the DXE phase of UEFI booting.
 ///
 /// This struct is generic over the [PlatformInfo] trait, which is used to provide platform-specific configuration to
@@ -387,8 +395,32 @@ impl<P: PlatformInfo> Core<P> {
         Self { component_dispatcher: ComponentDispatcher::new_locked(), hob_list: Once::new(), section_extractor }
     }
 
+    /// Sets the static DXE Core instance for global access.
+    ///
+    /// Returns true if the address of self is the same as the stored address, false otherwise.
+    #[must_use]
+    fn set_instance(&'static self) -> bool {
+        let physical_address = NonNull::from_ref(self).expose_provenance();
+        &physical_address == __SELF.call_once(|| physical_address)
+    }
+
+    /// Gets the static DXE Core instance for global access.
+    ///
+    /// This should only be used in efiapi functions where no reference to the core is otherwise available.
+    #[allow(unused)]
+    pub(crate) fn instance<'a>() -> &'a Self {
+        // SAFETY: The pointer is guaranteed to be set to a valid reference of this `Self` implementation as the atomic
+        // compare_exchange guarantees only one initialization has occurred. If the pointer was already set during the
+        // `set_instance` call of another `CORE` it would have returned a failure and then panicked before reaching this point.
+        unsafe {
+            NonNull::<Self>::with_exposed_provenance(*__SELF.get().expect("DXE Core is already initialized.")).as_ref()
+        }
+    }
+
     /// The entry point for the Patina DXE Core.
     pub fn entry_point(&'static self, physical_hob_list: *const c_void) -> ! {
+        assert!(self.set_instance(), "DXE Core instance was already set!");
+
         self.init_memory(physical_hob_list);
 
         if let Err(err) = self.start_dispatcher(physical_hob_list) {
@@ -679,6 +711,28 @@ fn call_bds() -> ! {
 #[coverage(off)]
 mod tests {
     use super::*;
+
+    #[test]
+    fn test_cannot_set_instance_twice() {
+        static CORE: Core<MockPlatformInfo> =
+            Core::<MockPlatformInfo>::new(patina_ffs_extractors::NullSectionExtractor);
+        static CORE2: Core<MockPlatformInfo> =
+            Core::<MockPlatformInfo>::new(patina_ffs_extractors::NullSectionExtractor);
+        assert!(CORE.set_instance());
+
+        if NonNull::from_ref(&CORE) != NonNull::from_ref(Core::<MockPlatformInfo>::instance()) {
+            panic!("CORE instance mismatch");
+        }
+
+        // We return true because its the same address
+        assert!(CORE.set_instance());
+        // This should fail because CORE2 is a different instance
+        assert!(!CORE2.set_instance());
+
+        if NonNull::from_ref(&CORE) != NonNull::from_ref(Core::<MockPlatformInfo>::instance()) {
+            panic!("CORE instance mismatch after second set_instance");
+        }
+    }
 
     #[test]
     fn test_trait_defaults_do_not_change() {


components/patina_adv_logger/src/logger.rs
```diff
@@ -65,6 +67,44 @@ where
...
+    /// Initializes the advanced logger memory log based on the provided physical hob
+    /// list. The physical hob list is used so this can be initialized before memory
+    /// allocations.
+    ///
+    /// ## Safety
+    ///
+    /// The caller must ensure that the provided physical hob list pointer is valid and well structured. Failure to do
+    /// so may result in unexpected memory access and undefined behavior.
+    ///
+    pub unsafe fn init_advanced_logger(&self, physical_hob_list: *const c_void) -> Result<(), EfiError> {
```
Javagedes: @cfernald any opinion if we just name this `init`?

cfernald: Go for it

docs/src/integrate/dxe_core.md
```diff
@@ -207,17 +207,23 @@ Add representative initialization (replace or augment extractors to match platfo
...
+struct ExamplePlatform;
+
+impl ComponentInfo for ExamplePlatform { }
+
+impl Platform for ExamplePlatform {
+    type ComponentInfo = Self;
+    type Extractor = BrotliSectionExtractor;
+}
+
+static CORE: Core<ExamplePlatform> = Core::new(BrotliSectionExtractor);
```
os-d: Why is BrotliSectionExtractor passed in here but also defined in `type Extractor`?

patina_dxe_core/examples/std.rs
```diff
@@ -29,20 +30,28 @@ static LOGGER: patina::log::SerialLogger<patina::serial::Terminal> = patina::log
     patina::serial::Terminal {},
 );
 
+struct StdPlatform;
```
os-d, line 32: A note in general, I wonder how useful this examples/std.rs file is. It may lead people to think it can run fw, a la EmulatorPkg (which it cannot) and I think that's really what people want to see in examples, something useable. I worry this will just fall out of date and not be read by anyone. Anyway, doesn't have to be addressed here, but I wonder if we just drop this.

patina_dxe_core/src/component_dispatcher.rs
```diff
@@ -0,0 +1,459 @@
...
+    /// A platform callback to register configurations with the core.
+    #[inline(always)]
+    fn configs<'a>(_add: Add<'a, Config>) {}
+
+    /// A platform callback to register services with the core.
+    #[inline(always)]
+    fn services<'a>(_add: Add<'a, Service>) {}
+}
+
+/// A marker trait to limit [Add] methods to only adding [Component](patina::component::Component)s
```
os-d:
```suggestion
/// A marker struct to limit [Add] methods to only adding [Component](patina::component::Component)s
```

Maybe I'm misunderstanding the nomenclature, but these three are structs, not traits, right?

patina_dxe_core/src/component_dispatcher.rs
```diff
@@ -0,0 +1,459 @@
...
+        let component = component.into_component();
+        self.dispatcher.insert_component(self.dispatcher.components.len(), component);
+    }
+}
+
+impl Add<'_, Config> {
+    /// Adds a configuration value to the core's storage.
+    #[inline(always)]
+    pub fn config<C: Default + 'static>(&mut self, _config: C) {
+        self.dispatcher.storage.add_config::<C>(C::default());
```
os-d: I'm confused here, is there no way to pass a config that doesn't use the default value? The platform is expected to impl Default for anything it passes in?

patina_dxe_core/src/lib.rs
```diff
@@ -114,20 +142,79 @@ macro_rules! error {
...
+    /// The platform's section extractor type, used when extracting sections from firmware volumes.
+    type Extractor: SectionExtractor;
+
+    /// Informs the core that it should prioritize allocating 32-bit memory when
+    /// not otherwise specified.
+    ///
+    /// This should only be used as a workaround in environments where address width
+    /// bugs exist in uncontrollable dependent software. For example, when booting
+    /// to an OS that puts any addresses from UEFI into a uint32.
+    fn prioritize_32_bit_memory() -> bool {
```
cfernald, line 185: Please add `#[inline(always)]` to the functions with default implementations.

patina_dxe_core/src/component_dispatcher.rs
```diff
@@ -0,0 +1,459 @@
...
+
+impl Add<'_, Service> {
+    /// Adds a service to the core's storage.
+    #[inline(always)]
+    pub fn service(&mut self, service: impl IntoService + 'static) {
+        self.dispatcher.storage.add_service(service);
+    }
+}
+
+pub struct ComponentDispatcher {
```
os-d: pub(crate)? Is a platform expected to create this?

patina_dxe_core/src/component_dispatcher.rs
```diff
@@ -0,0 +1,459 @@
...
+/// ```
+#[cfg_attr(test, mockall::automock)]
+pub trait ComponentInfo: Sized {
+    /// A platform callback to register components with the core.
+    #[inline(always)]
+    fn components<'a>(_add: Add<'a, Component>) {}
+
+    /// A platform callback to register configurations with the core.
+    #[inline(always)]
+    fn configs<'a>(_add: Add<'a, Config>) {}
```
os-d, line 49: To clarify, these are only configs for components, not configs that the core itself might use, right?

patina_dxe_core/src/lib.rs
os-d, line 185: Maybe I misunderstood where the RFC landed, but I thought we were going to have a hierarchy?

patina_dxe_core/src/lib.rs
```diff
@@ -145,92 +232,106 @@ impl Default for GicBases {
...
-/// Once the [init_memory](Core::init_memory) method is called, the struct transitions to the `Alloc` phase,
-/// which provides methods for adding configuration and components with the DXE core, and eventually starting the
-/// dispatching process and eventual handoff to the BDS phase.
+/// This struct is generic over two traits:
 ///
-/// ## Soft Service Dependencies
+/// 1. [ComponentInfo]: This trait exposes the platform-specific configuration options used directly by the DXE Core.
+///    Associated types and methods within this trait provide configuration values that influence the behavior that
+///    may be different between platforms, such as the prioritization of 32-bit memory allocations. It may also
+///    specify associated types that define platform-specific implementations such as a SectionExtractor
```
os-d: Aren't the 32 bit memory prioritization and the SectionExtractor in the platform trait, not ComponentInfo?

patina_dxe_core/src/lib.rs
```diff
@@ -145,92 +232,106 @@ impl Default for GicBases {
...
+///    implementation.
+/// 2. [Platform]: This trait provides callbacks for the platform to register additional patina components,
```
os-d: I think you just have the two traits swapped

patina_dxe_core/src/lib.rs
```diff
@@ -274,127 +375,10 @@ impl Core<NoAlloc> {
...
-                Ok(false) => false,
-                Err(err) => {
-                    log::error!("Dispatched: Id = [{name:?}] Status = [Failed] Error = [{err:?}]");
-                    debug_assert!(false);
-                    true // Component dispatched, even if it did fail, so remove from self.components to avoid re-dispatch.
-                }
-            }
-        });
-        len != self.components.len()
+        self.component_dispatcher.lock().add_service(cpu);
```
cfernald: Preference: consider breaking this out into a separate function and only acquiring the lock once. I imagine this will grow and it would be nice to have a simple discoverable function to find the core service installations.

patina_dxe_core/src/lib.rs
```diff
@@ -145,92 +232,106 @@ impl Default for GicBases {
...
-pub struct Core<MemoryState> {
-    physical_hob_list: *const c_void,
-    hob_list: HobList<'static>,
-    components: Vec<Box<dyn Component>>,
-    storage: Storage,
-    platform_timer: PerfTimer,
-    _memory_state: core::marker::PhantomData<MemoryState>,
+pub struct Core<P: Platform> {
+    /// A parsed and heap-allocated list of HOBs provided by [Self::entry_point].
+    hob_list: TplMutex<HobList<'static>>,
```
os-d: Do we really need a `TplMutex` over the hob_list? DXE is not allowed to modify HOBs, so this should be read only.

patina_dxe_core/src/lib.rs
```diff
@@ -520,65 +473,72 @@ impl Core<Alloc> {
...
         log::info!("Finished Dispatching Drivers");
 
-        self.display_components_not_dispatched();
+        self.component_dispatcher.lock().display_not_dispatched();
 
         core_display_missing_arch_protocols();
 
         dispatcher::display_discovered_not_dispatched();
 
-        call_bds();
```
os-d, line 577: We don't call BDS any longer?

patina_dxe_core/src/lib.rs
```diff
@@ -520,65 +473,72 @@ impl Core<Alloc> {
...
-        self.insert_component(0, hw_interrupt_protocol::HwInterruptProtocolInstaller::default().into_component());
+        dispatcher.insert_component(
+            0,
+            hw_interrupt_protocol::HwInterruptProtocolInstaller::new(P::gic_bases()).into_component(),
+        );
     }
 
     /// Starts the core, dispatching all drivers.
-    pub fn start(mut self) -> Result<()> {
+    fn start(&'static self, physical_hob_list: *const c_void) -> Result<()> {
```
os-d: Since we are changing things anyway, I think it is confusing to have a function called `start` which is not where you start :). I would suggest that this could just be part of `entry_point`, after we call `init_memory`.

patina_dxe_core/src/component_dispatcher.rs
```diff
@@ -0,0 +1,459 @@
...
+
+/// A marker trait to limit [Add] methods to only adding [Component](patina::component::Component)s
+pub struct Component;
+/// A marker trait to limit [Add] methods to only adding Configs.
+pub struct Config;
+/// A marker trait to limit [Add] methods to only adding [Service](patina::component::service::Service)s
+pub struct Service;
+
+/// A struct used to allow controlled access to the Core's storage.
+pub struct Add<'a, L> {
```
cfernald, line 64: nit: should we use something with some noun in the title to be more specific avoid import issues with the [standard Add trait](https://doc.rust-lang.org/std/ops/trait.Add.html)? Maybe something like `PlatformAdd`?

patina_dxe_core/src/lib.rs
Javagedes: This hob list is not the pointer to the hob list, it is a "rusty" heap allocated structure. Due to this, the hob_list is const initialized to a zero-length hob_list and we must parse the actual hob list to fill this structure. This requires a mutable reference.

The other option I can think of is to use a OnceCell<HobList>, then create the hob list, fill it, and set it in the OnceCell. That is a good implementation as it enforces no future changes. I can update to that if you prefer that. I'm also open to other solutions.

patina_dxe_core/src/lib.rs
Javagedes: I moved to the top level `entry_point` method where we are now properly doing infinite loop between bds entry point and the dispatcher.

patina_dxe_core/src/lib.rs
```diff
@@ -145,92 +232,132 @@ impl Default for GicBases {
...
-impl Core<NoAlloc> {
-    /// Initializes the timer frequency if a platform provides a custom implementation.
-    pub fn init_timer_frequency(self, frequency: Option<u64>) -> Core<NoAlloc> {
-        if let Some(freq_override) = frequency {
-            self.platform_timer.set_frequency(freq_override);
+    /// Sets the static DXE Core instance for global access.
+    #[allow(unused)]
+    fn set_instance(&'static self) {
+        __SELF.store(self as *const _ as *mut u8, core::sync::atomic::Ordering::SeqCst);
+
```
cfernald: nit: extra space

patina_dxe_core/src/lib.rs
Javagedes: They do call `entry_point` now. `start`'s `pub` was removed. It is now only starting up the dispatching process after the core has been initialized.

Javagedes: Will rename it to something better. Maybe `start_dispatcher` or something?

patina_dxe_core/src/lib.rs
Javagedes: woops. Thanks

patina_dxe_core/src/lib.rs
```diff
@@ -145,92 +232,132 @@ impl Default for GicBases {
...
 
-impl Core<NoAlloc> {
-    /// Initializes the timer frequency if a platform provides a custom implementation.
-    pub fn init_timer_frequency(self, frequency: Option<u64>) -> Core<NoAlloc> {
-        if let Some(freq_override) = frequency {
-            self.platform_timer.set_frequency(freq_override);
+    /// Sets the static DXE Core instance for global access.
+    #[allow(unused)]
+    fn set_instance(&'static self) {
+        __SELF.store(self as *const _ as *mut u8, core::sync::atomic::Ordering::SeqCst);
```
cfernald: consider a compare and swap, checking that it is currently null and panic if not.

patina_dxe_core/src/lib.rs
Javagedes: This is associated with comment below about the two traits description being swapped. Closing this.

patina_dxe_core/src/component_dispatcher.rs
Javagedes: `pub(crate)` means only the `patina_dxe_core` internals can call this function. It gets automatically created when you create your static core.

patina_dxe_core/src/lib.rs
os-d: I think the OnceCell more properly communicates what we are doing with the hob list and like you said gets the benefit of not being able to be updated

patina_dxe_core/src/component_dispatcher.rs
Javagedes: Yes, `Config` for components have always been this way. Similar to PCDs, they have a default value if not overwritten by the platform.

patina_dxe_core/src/lib.rs
os-d: > They do call `entry_point` now. `start`'s `pub` was removed. It is now only starting up the dispatching process after the core has been initialized.

I know, but I'm saying it is confusing that `entry_point` calls `start`, but not even as the first thing it does :)

I had thought something along the lines of `start_dispatcher` also, but there is a lot of stuff happening there. But that's fine with me.

patina_dxe_core/src/lib.rs
```diff
@@ -145,92 +232,132 @@ impl Default for GicBases {
...
-pub struct Alloc;
-
-#[doc(hidden)]
-/// A zero-sized type to gate non-allocation functions in the [Core].
-pub struct NoAlloc;
+/// Static reference to the DXE Core instance in the compiled binary.
+/// 
+/// This is set during the `entry_point` call of the DXE Core and is used to provide static access to the core for use
+/// only in efiapi functions where no reference to the core is otherwise available.
+static __SELF: AtomicPtr<u8> = AtomicPtr::new(ptr::null_mut());
```
cfernald: Resuming the conversation from the RFC. I believe this could simply be a `UnsafeCell<Option<NonNull<u8>>>` where the set/get will simply enforce a set-once semantic. if we wanted to avoid the atomic. I'm actually not concerned with the use of atomics, but I know there is some hesitation around them and frankly considering this should only ever be set by the BSP w/o interrupts they aren't needed.

patina_dxe_core/src/component_dispatcher.rs
Javagedes: Yes, misnamed on my part. Removing the word all together. Will just be `A marker to limit ...`

patina_dxe_core/src/component_dispatcher.rs
os-d: Right, that was my question though, should this be scoped to only the patina_dxe_core? I would assume no one else is creating a ComponentDispatcher?

patina_dxe_core/src/lib.rs
```diff
@@ -145,92 +232,132 @@ impl Default for GicBases {
...
+    }
+
+    /// Gets the static DXE Core instance for global access.
+    /// 
+    /// This should only be used in efiapi functions where no reference to the core is otherwise available.
+    #[allow(unused)]
+    pub(crate) fn instance<'a>() -> &'a Self {
+        let ptr = __SELF.load(core::sync::atomic::Ordering::SeqCst);
+        debug_assert!(!ptr.is_null(), "DXE Core instance is not initialized.");
+        unsafe { &*(ptr as *const Self) }
```
cfernald: safety comment. Should probably state the dependency that in order to executed to the point of calling instance, set_instance must have succeeded and locked out any future callers (using the compare and swap or a is no None check).

docs/src/integrate/dxe_core.md
Javagedes: in the type portion of the trait, we are defining the actual associated type, e.g. `BrotliSectionExtractor`, `CompositeSectionExtractor`, etc.

In the `new` method, we are passing an instance of the type. Since `BrotliSectionExtractor` is a zero sized struct, the above is actually passing an "instance" (though its not really an instance since its zero sized).

I will go ahead and implement `new` on these section extractors just to make that more clear, as it is fairly confusing.

Additionally, the `type Extractor` informs the platform what type of extractor to expect the platform to use. e.g.

```rust

trait Platform {
    type Extractor: SectionExtractor
}

struct Core<P: Platform> {
    section_extractor: P::Extractor
}

impl <P: Platform> {
     pub const fn new(extractor: P::Extractor) {
        Self {
            extractor
        }
    }
}

File: patina_dxe_core/examples/std.rs (hunk 3 above)
Comment (Javagedes) [reply patina_dxe_core/examples/std.rs:32]: I'm fine with removing it

File: patina_dxe_core/src/component_dispatcher.rs (hunk 5 above)
Comment (os-d) [reply patina_dxe_core/src/component_dispatcher.rs]: Okay, that makes sense. I was just confused by always calling default here. But I understand that default changes.

So I, as a platform owner, want to change the default, do I do something like:

```rust
impl Default for AdvLoggerConfig () {
   AdvLoggerConfig {0, 3, 45}
}
```

patina_dxe_core/src/lib.rs
Javagedes: We were only going to have a heirarchy where deemed necessary. That is why we currently have a `ComponentInfo` sub trait for specifying all component configuration.

If we had more then one configuration for memory related things, I'd be for willing to add a new trait. But as it stands, I don't see the need to create a whole new trait the platform must implement for a single function. e.g. this is to verbose in my opinon

```rust
struct Q35;

impl MemoryInfo for Q35 {
    fn prioritize_32_bit_memory() -> bool { true }
}

impl ComponentInfo for Q35 {
    fn components(...) { }
    fn configs(...) { }
    fn services(...) { }
}

impl Platform for Q35 {
    type MemoryInfo = Self;
    type ComponentInfo = Self;
    type SectionExtractor = BrotliSectionExtractor;
    
    fn gic_bases() -> GicBases { todo!() }
}
```

I'm fine with doing it as long as that is the group consensus though.

os-d: Right now we only have one memory config, but there will be others (e.g. heap guard, pool guard, etc.). My main concern is if we don't add hierarchies now, we'll have breaking changes later to add them. If we are fine with that, I'm fine to say defer this and make a breaking change later when we can better scope what these hierarchies should be.

patina_dxe_core/src/component_dispatcher.rs
Javagedes: I did this so that the interface for the PlatformInfo is nice:

```rust
impl PlatformInfo for Q35 {
    fn components(add: Add<Config>) {
        add.config(...);
   }

   fn services(add: Add<Service>) {
        add.service(...);
   }

   fn components(add: Add<Component>) {
        add.component(...);
   }
}
```

I can change the name if you prefer, to prevent naming conflicts, but it will make the implementation a big "ugglier", though not sure how important that is. Just lmk.

patina_dxe_core/src/lib.rs
Javagedes: I don't have a huge opinion. The main reason I have stuck to atomic is that it is simple.

None of these other suggested types (such as UnsafeCell) are Send + Sync, so I'm going to have to create a wrapper type and manually implement Send + Sync on it, to get it to work. Where as AtomicPtr is an out of the box solution.

I think @joschock has the biggest hate of atomics so if he tells me to switch, I will :)

patina_dxe_core/src/lib.rs
Javagedes: will do if we stick to atomics.

patina_dxe_core/src/component_dispatcher.rs
Javagedes: oh. I see your confusion. That function is a big typo. I blame auto complete on that one. It is completely wrong.

Updating to:

```rust
    pub fn config<C: Default + 'static>(&mut self, config: C) {
        self.dispatcher.storage.add_config::<C>(config);
    }
```

patina_dxe_core/src/component_dispatcher.rs
Javagedes: Oh. I get it now. Yes, it should be pub(crate). Thank you

patina_dxe_core/src/lib.rs
Javagedes: definitely worth cleaning up the entire flow once this PR is merged. I don't want to complicate this PR even further. Will create an issue and self-assign.

I'll rename to `start_dispatcher` for now.

patina_dxe_core/src/component_dispatcher.rs
Javagedes: Yes, this is only for setting public configuration of components. We moved these three functions into the `ComponentInfo` trait to try and make that clearer. If it's not very clear, I'm open to suggestions to make it clear :)

patina_dxe_core/src/lib.rs
Javagedes: I don't personally have a problem with breaking changes down the road but I totally understand doing it now if there are plans to add additional configuration. Anyone else have opinions?

patina_dxe_core/src/lib.rs
Javagedes: Its fairly hard to break out into a different function right now because everything we register is currently an instance generated from the function its in. I think our overall plan is to have the core itself be the services for most if not all of these once we remove all the statics.

Due to this, I've updated the call to lock once and add everything, but I'm holding off on breaking it out into a new function for now.

docs/src/integrate/dxe_core.md
Javagedes: Added the new functions. Closing this. Feel free to re-open.

patina_dxe_core/src/lib.rs
Javagedes: Added the comment.

patina_dxe_core/src/lib.rs
Javagedes: Added the compare exchange anyway, as it was a simple change.

patina_dxe_core/src/lib.rs
Javagedes: Renamed to `start_dispatcher` and created / self-assigned #1076 to myself.

patina_dxe_core/src/component_dispatcher.rs
cfernald: I'm not very convicted here, mostly pointing out the potential collision, though i doubt it'll come up. Feel free to resolve

patina_dxe_core/src/lib.rs
cfernald: While I would typically say "hold off until we know we need it" I do think `MemoryInfo` would be a good example of a scenario where we will certainly add more options. I'd be fine with splitting this off.

patina_dxe_core/src/lib.rs
joschock: My hate of atomics has been assuaged by actual performance data which indicates they don't seem to matter, but I also think there are better/safer primitives than AtomicPtr in many (most?) cases.

In this particular case, the AtomicPtr is Send + Sync, sure. But that doesn't get you off the hook re: the things that Rust is trying to protect you from by demanding `Sync + Send` - it just fools the compiler.

Whatever you are recreating from the atomic pointer via `unsafe { &*my_atomic_ptr.load(...) }` or equivalent also needs to be at least `Sync`. The Atomic only protects your _pointer_ from data races. It doesn't protect your _pointee_ behind that pointer from data races - and if it isn't `Sync`, I think that's UB. FWIW, a `Sync + Send` impl on a wrapper around `UnsafeCell<ptr>` has the same issue. 

Core _should_ already be `Sync` because all of its fields should be concurrent-safe (i.e. behind a TplMutex or whatever). But I think it is good to make that requirement explicit. As long as that is upheld, I'm agnostic on the actual storage mechanism for the global pointer. `spin::Once` is attractive since it guarantees other invariants (like, no one accidentally changes the Core pointer after initial creation).

joschock: Remind me again why we need to erase the Core type on __SELF? it's because it is instantiated/implemented in the bin rather than in the lib?

Javagedes: The `Core` struct is generic over the `Platform` trait, which cannot be known from inside the `patina_dxe_core` crate. The `Platform` implementation can change the layout of the `Core` struct (Right now the size of the section extractor, but in the future other things.

Due to this, the `Platform` selector / instantiate must be responsible for the type erasure and reconstruction of the type, which is all done at compile time via the generics when the platform selects the `Platform` implementation to use with the core.

In terms of actual coding that is easy to understand, I cannot do this:

```rust
static __SELF: AtomicPtr<Core<P>> = ptr::null_mut();
```

I will get a compile error because `P` must be selected.

patina_dxe_core/src/lib.rs
```diff
@@ -145,92 +236,148 @@ impl Default for GicBases {
...
-}
 
-impl Core<NoAlloc> {
-    /// Initializes the timer frequency if a platform provides a custom implementation.
-    pub fn init_timer_frequency(self, frequency: Option<u64>) -> Core<NoAlloc> {
-        if let Some(freq_override) = frequency {
-            self.platform_timer.set_frequency(freq_override);
+    /// Sets the static DXE Core instance for global access.
+    #[allow(unused)]
+    fn set_instance(&'static self) -> bool {
```
cfernald, line 402: please create tests to verify set_instance set once behavior. That is

set -> succeeds
get -> returns set pointer
set -> panics

makubacki:
Going to review soon, please leave open for a while.

patina_dxe_core/src/lib.rs
Javagedes: Added the `MemoryInfo` trait as requested!

patina_dxe_core/examples/std.rs
Javagedes: https://github.com/OpenDevicePartnership/patina/issues/1090

patina_dxe_core/src/component_dispatcher.rs
Javagedes: I think the risk of collision is highly unlikely and they can simply fully expose the path for either the `Add` trait or this `Add` struct if there is a conflict.

Javagedes: Therefore I'm going to close this

patina_dxe_core/examples/std.rs
```diff
@@ -29,20 +30,27 @@ static LOGGER: patina::log::SerialLogger<patina::serial::Terminal> = patina::log
...
+    type MemoryInfo = Self;
+    type ComponentInfo = Self;
+    type Extractor = patina_ffs_extractors::CompositeSectionExtractor;
+}
+
+impl patina_dxe_core::ComponentInfo for StdPlatform {}
+
+impl patina_dxe_core::MemoryInfo for StdPlatform {}
+
+static mut CORE: Core<StdPlatform> = Core::new(CompositeSectionExtractor::new());
```
joschock: why does it need to be `static mut`?

patina_dxe_core/src/component_dispatcher.rs
```diff
@@ -0,0 +1,467 @@
+extern crate alloc;
```
joschock, line 11: nit: header?

patina_dxe_core/src/lib.rs
Javagedes: panic is inside of one of the runtime functions so I'm not bothering with that, however I did add a test that ensures `set_instance` returns false on a second call (which leads to a panic in runtime), and that the address provided from `instance()` is the same address as the static core defined by the platform.

patina_dxe_core/src/component_dispatcher.rs
```diff
@@ -0,0 +1,467 @@
...
+        Self { components: Vec::new(), storage: Storage::new() }
+    }
+
+    pub(crate) fn apply_component_info<C: ComponentInfo>(&mut self) {
+        C::configs(Add::new(self));
+        C::services(Add::new(self));
+        C::components(Add::new(self));
+    }
+
+    /// Inserts a component at the given index. If no index is provided, the component is added to the end of the list.
```
joschock: idx arg is not an option. how does a caller not provide it?

os-d: Also, I missed this, but being able to provide an index smells suspiciously like a priori dispatch :). I get we have that anyway if you just put them in the order you want to dispatch them, but do we have a use case for the index?

patina_dxe_core/src/hw_interrupt_protocol.rs
```diff
@@ -451,20 +451,23 @@ impl HwInterruptProtocolHandler {
...
+    }
+
     fn entry_point(
         self,
         interrupt_manager: Service<dyn InterruptManager>,
-        gic_bases: Config<GicBases>,
         boot_services: StandardBootServices,
     ) -> patina::error::Result<()> {
-        log::info!("GICv3 initializing {:x?}", (gic_bases.0, gic_bases.1));
+        log::info!("GICv3 initializing {:x?}", (self.0.0, self.0.1));
```
joschock: This is starting to look like an ip address. 

Can we maybe change this to 
`pub struct GicBases {
  gic_distributor: u64,
  gic_redistributor: u64
}`
so we have some nice self-documenting names? Could be done as a separate PR if too off-topic.

patina_dxe_core/src/lib.rs
Javagedes: Talked offline with John and the main concern is that `Send` compatibility is being enforced by the compiler because if we add a non-send type, `Core` will stop being `Send` also, however we are blanket implementing `Sync` on the `Core`, when we should be letting it auto implement so that we do not accidently add a type to the struct that is not `Sync`.

I've removed this blanket implementation and fixed what was necessary for us to be `Sync` automatically.

patina_dxe_core/examples/std.rs
Javagedes: It does not. Updating. Thanks

patina_dxe_core/src/component_dispatcher.rs
Javagedes: Platform components are appended to the back of the list, so when we call this function we do `.insert_component(self.dispatcher.components.len(), component);`.

The function is like this so that the core can ensure its components are first in the list.

patina_dxe_core/src/hw_interrupt_protocol.rs
Javagedes: lmao I agree this is a bit much. I'll make an issue and do it as a follow up PR

Javagedes: #1091

patina_dxe_core/src/lib.rs
```diff
@@ -145,92 +258,146 @@ impl Default for GicBases {
...
+impl<P: Platform> Core<P> {
+    /// Creates a new instance of the DXE Core in the NoAlloc phase.
+    pub const fn new(section_extractor: P::Extractor) -> Self {
+        Self { component_dispatcher: ComponentDispatcher::new_locked(), hob_list: Once::new(), section_extractor }
+    }
+
+    /// Sets the static DXE Core instance for global access.
+    ///
+    /// Returns true if the address of self is the same as the stored address, false otherwise.
+    #[allow(unused)]
```
cfernald: This looks like a good candidate for `#[must_use]` since continuing without checking the result will lead to unexpected behavior. Alternatively, `Result` would provide the same compile time check.

patina_dxe_core/src/component_dispatcher.rs
joschock: If the index is mandatory, then the language of the comment needs to be updated.

Javagedes: Ah yes, must have been an artifact of a much older change that was never updated. When it was copied to this file from `lib.rs`. the invalid documentation stayed. Updated.

patina_dxe_core/src/lib.rs
```diff
@@ -145,92 +258,146 @@ impl Default for GicBases {
...
+        if let Err(err) = self.start_dispatcher(physical_hob_list) {
+            log::error!("DXE Core failed to start: {err:?}");
+        }
+
+        loop {
+            call_bds();
+            log::info!("Returned from BDS phase. Re-dispatching.");
+            if let Err(err) = self.core_dispatcher() {
+                log::error!("DXE Core failed to re-dispatch: {err:?}");
+            }
```
joschock: I don't think that BDS should ever return? https://uefi.org/specs/PI/1.9/V2_DXE_Architectural_Protocols.html#efi-bds-arch-protocol-entry -> "This function should never return". 

EDK Core also deadloops if BDS returns, so I think this should probably just assert on return.

patina_dxe_core/src/lib.rs
```diff
@@ -145,92 +258,146 @@ impl Default for GicBases {
...
         }
-        self
+
+        let _ = self.hob_list.call_once(|| hob_list);
+        true
+    }
+
+    /// Returns a reference to the HOB list.
+    ///
+    /// Must not be called until
```
joschock: until.....

patina_dxe_core/src/component_dispatcher.rs
```diff
@@ -0,0 +1,480 @@
...
+    }
+
+    /// Sets the Boot Services table in storage.
+    ///
+    /// ## Safety
+    ///
+    /// The caller must ensure that the provided pointer will live for the lifetime of the DXE Core.
+    #[coverage(off)]
+    #[inline(always)]
+    pub(crate) unsafe fn set_boot_services(&mut self, mut bs: NonNull<efi::BootServices>) {
```
joschock: StandardBootServices and StandardRuntimeServices are `Clone` - and the construction of them is already requiring the creator to uphold essentially these safety req.  Consider just making these
`pub(crate) fn set_boot_services(&mut self, bs: StandardBootServices)` or maybe even `pub(crate) fn set_boot_services(&mut self, bs: impl BootServices)` and leaving the unsafe in the caller?

patina_dxe_core/src/lib.rs
Javagedes: Thanks. This is good information. I've removed this infinite loop. Instead, we now panic if we fail to find BDS protocol or we return from BDS.

patina_dxe_core/src/lib.rs
Javagedes: added must_use. good call.

patina_dxe_core/src/component_dispatcher.rs
Javagedes: Updated all of this. We don't need any unsafe anywhere (Even on the caller) once we update this function to take standardbootservices / runtimeservices directly. Just an artifact of slowly changing over time. Good catch.

patina_dxe_core/Cargo.toml
```diff
@@ -52,6 +52,7 @@ patina_internal_collections = { path = "../core/patina_internal_collections" }
 env_logger = { workspace = true }
 mockall = { workspace = true }
 serial_test = { workspace = true }
+zerocopy = { workspace = true, features = ["derive"] }
```
makubacki: Should we use `zerocopy_derive` for `dev-dependencies` as well? The inconsistency makes an already confusing trap for how that dependency is managed a little more confusing.

I don't have a strong opinion though. This is fine for simplicity if others prefer that.

patina_dxe_core/src/lib.rs
```diff
@@ -114,20 +149,100 @@ macro_rules! error {
 
 pub(crate) static GCD: SpinLockedGcd = SpinLockedGcd::new(Some(events::gcd_map_change));
 
+/// A trait to be implemented by the patform to provide configuration values and types related to memory management
```
makubacki:
```suggestion
/// A trait to be implemented by the platform to provide configuration values and types related to memory management
```

patina_dxe_core/src/lib.rs
```diff
@@ -114,20 +149,100 @@ macro_rules! error {
...
+///   }
+/// }
+#[cfg_attr(test, mockall::automock)]
+pub trait MemoryInfo {
+    /// Informs the core that it should prioritize allocating 32-bit memory when
+    /// not otherwise specified.
+    ///
+    /// This should only be used as a workaround in environments where address width
+    /// bugs exist in uncontrollable dependent software. For example, when booting
+    /// to an OS that puts any addresses from UEFI into a uint32.
```
makubacki: nit: clarity

```suggestion
    /// bugs exist in uncontrollable dependent software. For example, when booting
    /// to an OS that erroneously stores UEFI allocated addresses greater than
    /// 32-bits in a 32-bit variable.
```

patina_dxe_core/src/lib.rs
makubacki: Can these traits be moved into a separate file (module)? They can be re-exported if needed to simplify their namespace, but it'd be easier to manage and focus on them outside of this lib.rs file.

patina_dxe_core/src/lib.rs
```diff
@@ -114,20 +149,100 @@ macro_rules! error {
...
+///   type MemoryInfo = Self;
+///   type ComponentInfo = Self;
+///   type Extractor = patina_ffs_extractors::NullSectionExtractor;
+/// }
+///
+/// impl ComponentInfo for ExamplePlatform {}
+/// impl MemoryInfo for ExamplePlatform {}
+/// ```
+#[cfg_attr(test, mockall::automock(type Extractor = patina_ffs_extractors::NullSectionExtractor; type ComponentInfo = MockComponentInfo; type MemoryInfo = MockMemoryInfo;))]
+pub trait Platform {
```
makubacki: Is there a reason some trait types have an "Info" postfix and ones like this (`Platform`) do not?

patina_dxe_core/src/lib.rs
```diff
@@ -114,20 +149,100 @@ macro_rules! error {
...
+    /// By default, this returns `None`, indicating that the core should attempt to determine the frequency
+    /// automatically using cpu architecture-specific methods.
+    #[inline(always)]
+    fn timer_frequency() -> Option<u64> {
+        None
+    }
+
+    /// Informs the core of the GIC base addresses for AARCH64 systems.
+    #[cfg(target_arch = "aarch64")]
+    fn gic_bases() -> GicBases;
```
makubacki: This seems a little messy when considering growth. Can arch-specific stuff be collected into a single trait and that included here with a cfg condition?

patina_dxe_core/src/lib.rs
```diff
@@ -114,20 +149,100 @@ macro_rules! error {
...
+
+    /// The platform's section extractor type, used when extracting sections from firmware volumes.
+    type Extractor: SectionExtractor;
+
+    /// Returns the performance timer frequency for the platform.
+    ///
+    /// By default, this returns `None`, indicating that the core should attempt to determine the frequency
+    /// automatically using cpu architecture-specific methods.
+    #[inline(always)]
+    fn timer_frequency() -> Option<u64> {
```
makubacki: There are several timers (TSC, HPET, ACPI, 8254, etc.) for different purposes, while the doc comment is helpful, the actual function name needs to be more self-explanatory. At a minimum, `perf_timer_frequency()`.

patina_dxe_core/src/lib.rs
```diff
@@ -622,26 +633,63 @@ fn call_bds() {
...
+            // field of the protocol is a valid function pointer that conforms to the expected calling convention.
+            unsafe {
+                ((*bds).entry)(bds);
+            }
         }
-    }
+        Err(err) => log::error!("Unable to locate BDS protocol: {err:?}"),
+    };
+
+    unreachable!("BDS protocol should be found and should never return.");
```
makubacki: nit: clarity

```suggestion
    unreachable!("BDS arch protocol should be found and should never return.");
```

patina_dxe_core/src/lib.rs
```diff
@@ -622,26 +633,63 @@ fn call_bds() {
...
+        Ok(bds) => {
+            let bds = bds as *mut bds::Protocol;
+            // SAFETY: The BDS protocol is the valid C structure as defined by the UEFI specification. The entry
+            // field of the protocol is a valid function pointer that conforms to the expected calling convention.
+            unsafe {
+                ((*bds).entry)(bds);
+            }
         }
-    }
+        Err(err) => log::error!("Unable to locate BDS protocol: {err:?}"),
```
makubacki: nit: clarity

```suggestion
        Err(err) => log::error!("Unable to locate BDS arch protocol: {err:?}"),
```

patina_dxe_core/src/hw_interrupt_protocol.rs
makubacki: Is this comment meant to have the GIC version hardcoded in it?

```suggestion
        log::info!("GIC initializing {:x?}", (self.0.0, self.0.1));
```

Applies to other references in the function.

patina_dxe_core/src/hw_interrupt_protocol.rs
```diff
@@ -451,20 +451,23 @@ impl HwInterruptProtocolHandler {
...
         let aarch64_int = unsafe {
```
makubacki, line 471: Looks there is not a safety comment here for:

```rust
let aarch64_int = unsafe {
            AArch64InterruptInitializer::new(self.0.0 as _, self.0.1 as _)
                .inspect_err(|_| log::error!("Failed to initialize GICv3"))?
        };
```

patina_dxe_core/src/lib.rs
Javagedes: No real reason other than `Platform` just sounded better than `PlatformInfo`. I think it also helps that the `Platform` trait is the main trait that consumes all other traits, so having it be slightly different makes sense.

With that said, I can change it to `PlatformInfo` if you prefer all traits are similar.

patina_dxe_core/src/lib.rs
Javagedes: I'm open to doing something different with this trait method, however I do not think grouping arch-specific things into a single trait is the right move. I think that doing so detracts from the whole point of breaking up the config into separate sub-traits for better organization and clarity to the platform maintainer.

This is because we are already grouping by subsystem. There is a potential overlap between grouping by subsystem and grouping by architecture. If we introduce a new config for a specific subsystem, but it is only for a specific architecture, it is no longer clear where that configuration should go. Even if we decide, it should go in a specific place, that is still failing the original intent of breaking these into different sub traits for better organization and clarity.

patina_dxe_core/src/lib.rs
Javagedes: I think it makes sense to move the traits to where their consumer is. I've already done so with the `ComponentInfo` trait, which is in `component_dispatcher.rs`.

I think the `Platform` trait makes sense in lib.rs because the `Core` is the direct (and only) consumer of it.

I agree with moving `MemoryInfo`. I think I want to do that in a follow up PR immediately following this, though. This is because I plan on creating a `memory.rs` module and moving all of the other memory related things off of that (allocator, gcd, etc). And that change is outside the scope of this PR

If that is good for you, I'll leave `MemoryInfo` where it is at for now, and create an issue to move it as described above, and follow up with that immediately after this.

patina_dxe_core/Cargo.toml
Javagedes: switched to zerocopy-derive. I think that is the right answer.

patina_dxe_core/src/lib.rs
Javagedes: Realistically this trait makes sense in a `CpuConfig` trait, but I'm hesitant to do so because there is only one configuration at the moment the boilerplate does not seem worth it. I was convinced by others to break out `MemoryInfo` even though there is only one configuration, but that is because they have close in the future plans to offer more configurations.

patina_dxe_core/src/hw_interrupt_protocol.rs
Javagedes: After reading the safety requirements for this, it is due to needing to ensure that the gic bases are valid. So I will bubble up that unsafe function to the trait

Javagedes: Was able to limit the unsafe to just the construction of the `GicBases` struct, not the trait method. The struct invariants satisfy the unsafe requirements of `Aarch64InterruptInitializer`.

patina_dxe_core/src/hw_interrupt_protocol.rs
Javagedes: Updated generically to GIC

patina_dxe_core/src/hw_interrupt_protocol.rs
joschock: Pedantically, they are just numbers until you use them to touch the GIC - i.e. like raw pointers until you deref them. But as long as we have safety enforcement somewhere, it's aesthetics.

joschock: Also - valid - and exclusive (i.e. they point to a live GIC, and no one else is going to talk to that GIC)

patina_dxe_core/src/lib.rs
makubacki: Yeah, I didn't want to cause too much work to split things up, but I'd also prefer these to live with the consumer, sounds good. Deferring the `MemoryInfo` move is fine.

patina_dxe_core/src/lib.rs
makubacki: I'd prefer to `PlatformInfo` because it's a little more intuitive when mentally grouping these types, it sounds like you're fine with that too, so let's do that.

patina_dxe_core/src/lib.rs
makubacki: This comment articulates a nuance I agree with:

> I'm open to doing something different with this trait method...

My main ask is to not have a monolithic config surface that might get littered with individual arch conditions over time. I think "CPU config" is going to naturally have some deviation, the architectural structures are simply different.

patina_dxe_core/src/lib.rs
Javagedes: #1095

patina_dxe_core/src/lib.rs
Javagedes: Renamed to PlatformInfo

patina_dxe_core/src/lib.rs
Javagedes: I've Added a `CpuConfig` that is used for CPU related configuration. This lives in lib.rs right now, But I will be creating an issue to consolidate the cpu related functionality (cpu_arch_protocol, hw_interrupt_protocol, and an interface for interreacting with `patina_internal_cpu` to a top level module similar to the plans for memory.rs

Javagedes: #1096

makubacki (approved):

PR #1050, opened 2025-11-13
Replace AtomicPtr with Cell in patina_internal_collections.

joschock:
## Description

Switch internal collections implementation to `Cell` from `AtomicPtr` pursuant to RFC 0021 (https://github.com/OpenDevicePartnership/patina/pull/1036)

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Boot-tested on platform hardware, measured performance delta (no regression, but no marked improvement performance-wise). 

## Integration Instructions

N/A

### DIFF
--- a/core/patina_internal_collections/src/bst.rs
+++ b/core/patina_internal_collections/src/bst.rs
@@ -8,10 +8,7 @@
 //!
 #[cfg(feature = "alloc")]
 extern crate alloc;
-use core::{
-    cmp::Ordering,
-    sync::atomic::{self, AtomicPtr},
-};
+use core::{cell::Cell, cmp::Ordering};
 
 use crate::{
     Error, Result, SliceKey,
@@ -24,7 +21,7 @@ where
     D: SliceKey,
 {
     storage: Storage<'a, D>,
-    root: AtomicPtr<Node<D>>,
+    root: Cell<*mut Node<D>>,
 }
 
 impl<'a, D> Bst<'a, D>
@@ -37,12 +34,12 @@ where
     /// [with_capacity](Self::with_capacity) to create a tree with a given slice of memory immediately. Otherwise use
     /// [resize](Self::resize) to replace the memory later.
     pub const fn new() -> Self {
-        Bst { storage: Storage::new(), root: AtomicPtr::new(core::ptr::null_mut()) }
+        Bst { storage: Storage::new(), root: Cell::new(core::ptr::null_mut()) }
     }
 
     /// Creates a new binary tree with a given slice of memory.
     pub fn with_capacity(slice: &'a mut [u8]) -> Self {
-        Self { storage: Storage::with_capacity(slice), root: AtomicPtr::default() }
+        Self { storage: Storage::with_capacity(slice), root: Cell::default() }
     }
 
     /// Returns the number of elements in the tree.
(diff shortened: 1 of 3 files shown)


core/patina_internal_collections/src/bst.rs
```diff
@@ -68,11 +65,7 @@ where
 
     /// Returns the current root of the tree.
     fn root(&self) -> Option<&Node<D>> {
-        let root_ptr = self.root.load(atomic::Ordering::SeqCst);
-        if root_ptr.is_null() {
-            return None;
-        }
-        Some(unsafe { &*root_ptr })
+        unsafe { self.root.get().as_ref() }
```
Javagedes, line 70: nit: safety comment

core/patina_internal_collections/src/bst.rs
```diff
@@ -90,13 +83,12 @@ where
     pub fn add(&mut self, data: D) -> Result<usize> {
         let (idx, node) = self.storage.add(data)?;
 
-        if self.root.load(atomic::Ordering::SeqCst).is_null() {
-            self.root.store(node.as_mut_ptr(), atomic::Ordering::SeqCst);
-            return Ok(idx);
+        if self.root.get().is_null() {
+            self.root.set(node.as_mut_ptr());
+        } else {
+            let root = unsafe { self.root.get().as_mut().expect("root is not null") };
```
Javagedes: nit: safety comment

core/patina_internal_collections/src/bst.rs
joschock: https://github.com/OpenDevicePartnership/patina/pull/1053 addresses those safety comments; I'll rebase on that once it completes to address this.

core/patina_internal_collections/src/bst.rs
joschock: https://github.com/OpenDevicePartnership/patina/pull/1053 addresses those safety comments; I'll rebase on that once it completes to address this.

core/patina_internal_collections/src/node.rs
```diff
@@ -275,86 +266,86 @@ where
...
+        self.color.get()
     }
 
     fn parent(&self) -> Option<&Node<D>> {
-        let node = self.parent.load(Ordering::SeqCst);
+        let node = self.parent.get();
         match node.is_null() {
             true => None,
-            false => Some(unsafe { &*node }),
+            false => Some(unsafe { node.as_mut().expect("node is not null") }),
```
makubacki: `&Node<D>` is a shared reference. `node.as_mut()` is returning a mutable reference. Is that necessary?

Could something like `node.as_ref()` work?

makubacki: That will also handle the null check for returning the `Option`.

joschock: Good point; I'll switch it to as_ref(). I think due to the signature of the method, the caller still receives an immutable ref.

makubacki (approved):

core/patina_internal_collections/src/bst.rs
joschock: fixed in latest push.

core/patina_internal_collections/src/bst.rs
joschock: rebased.

core/patina_internal_collections/src/node.rs
joschock: fixed in latest push.

Issue #1030, opened 2025-11-07
[Task]: Improve unit test coverage in patina_dxe_core/src/image.rs


makubacki:
### Request Description

Unit test patch coverage was overridden to merge https://github.com/OpenDevicePartnership/patina/pull/1028.

Unit test coverage in the file should be updated to reach 80%.

### Are you going to make the change?

Someone else needs to make the change

### Do you need maintainer feedback?

No maintainer feedback needed

Issue #176, opened 2024-11-22
[Bug]: Don't attempt to reload images that memory protections failed to apply to


makubacki:
### Is there an existing issue for this?

- [x] I have searched existing issues

### Current Behavior

Currently, if memory protections cannot be applied, some code paths will log an error and continue. For example:

https://github.com/pop-project/uefi-dxe-core/blob/93e7b5246ae81922cd2315cd22b02a3f29261f8e/dxe_core/src/image.rs#L343-L359

### Expected Behavior

This issue tracks not loading the image again per the note in this [comment](https://github.com/pop-project/uefi-dxe-core/pull/175/files/dd048c7dc03b273a3bec385ad0aaf477c3cb1e20#r1854605388).

### Build Environment

```markdown
- OS(s):
- Tool Chain(s):
- Targets Impacted:
```

### Version Information

```text

```

### Urgency

Low

### Are you going to fix this?

Someone else needs to fix it

### Do you need maintainer feedback?

None

PR #1051, opened 2025-11-13
Add test coverage for memory protection error handling

garybeihl:
## Description

Task #1030 requests improved unit test coverage for memory protection error handling. This commit adds comprehensive test coverage for all three error paths introduced by #176, improving overall coverage metrics.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [X] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- All 387 tests pass successfully with no regressions.
- Coverage verified with cargo llvm-cov

## Integration Instructions

N/A if nothing is required.

### DIFF
--- a/patina_dxe_core/src/image.rs
+++ b/patina_dxe_core/src/image.rs
@@ -2311,8 +2311,11 @@ mod tests {
         //
         // This scenario represents a corrupted or uninitialized GCD state where memory
         // descriptors are missing - an edge case that the error handling should catch.
+        //
+        // Note: In debug builds, this hits a debug_assert!(false) which panics.
+        // The panic is caught by with_global_lock(), so we check for that.
 
-        let _ = test_support::with_global_lock(|| {
+        let result = test_support::with_global_lock(|| {
             // SAFETY: These test initialization functions require unsafe because they
             // manipulate global state (GCD, protocol DB, system table)
             unsafe {
@@ -2389,6 +2392,14 @@ mod tests {
             assert!(result.is_err(), "Protection should fail when section address is not in GCD");
             assert_eq!(result.unwrap_err(), EfiError::NotFound, "Expected NotFound from GCD descriptor lookup");
         });
+
+        // In debug builds, debug_assert!(false) panics and with_global_lock catches it
+        #[cfg(debug_assertions)]
+        assert!(result.is_err(), "Expected panic from debug_assert! in debug build");
+
+        // In release builds, debug_assert is compiled away and function returns error normally
+        #[cfg(not(debug_assertions))]
+        assert!(result.is_ok(), "Expected successful test execution in release build");
     }
 
     #[test]
@@ -2399,8 +2410,11 @@ mod tests {
         //
         // We craft a malformed PE image with a section virtual_size that will cause
         // align_up() to overflow when aligning to section_alignment.
+        //
+        // Note: In debug builds, this hits a debug_assert!(false) which panics.
+        // The panic is caught by with_global_lock(), so we check for that.
 
-        let _ = test_support::with_global_lock(|| {
+        let result = test_support::with_global_lock(|| {
             // SAFETY: These test initialization functions require unsafe because they
             // manipulate global state (GCD, protocol DB, system table)
             unsafe {
@@ -2449,9 +2463,18 @@ mod tests {
                 Ok(_) => panic!("Image with overflowing section alignment should fail to load"),
             }
         });
+
+        // In debug builds, debug_assert!(false) panics and with_global_lock catches it
+        #[cfg(debug_assertions)]
+        assert!(result.is_err(), "Expected panic from debug_assert! in debug build");
+
+        // In release builds, debug_assert is compiled away and function returns error normally
+        #[cfg(not(debug_assertions))]
+        assert!(result.is_ok(), "Expected successful test execution in release build");
     }
 
     #[test]
+    #[cfg(debug_assertions)]
     fn load_image_should_fail_with_unaligned_section_address() {
         // This test verifies error path #3 from Fix #176: set_memory_space_attributes failure
         // (Task #1030 coverage improvement)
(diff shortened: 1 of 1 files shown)


patina_dxe_core/src/image.rs
```diff
@@ -2196,4 +2196,317 @@ mod tests {
...
+                image_base_page: 0x1000,
+                image_num_pages: 2,
+            };
+
+            // Call apply_image_memory_protections directly
+            let _result = super::apply_image_memory_protections(&pe_info, &private_info);
+
+            // Should FAIL with NotFound because the GCD RBT is empty (no descriptors),
+            // so get_closest_idx returns None and get_memory_descriptor_for_address returns NotFound
+            #[cfg(not(debug_assertions))]
```
os-d: I don't think we need to switch back and forth between debug asserts enabled or not (they always will be in test). Instead, you can do a `#[should_panic]` decorator on this function and the test will pass if it panics. That's not the most ideal because any panic will qualify as success, AFAIK, you can't specify what panic should be the right panic.

Now, something we've talked a lot about in the past is disabling debug asserts for tests and I think we should. I'll file an issue.

makubacki: Yeah, there is no need for this. If I understand what you're referring to correctly, you can check for specific panic messages `#[should_panic(expected = "Panic message")]`.

.gitignore
```diff
@@ -4,3 +4,4 @@ Cargo.lock
 docs/book/
 docs/target/
 .vscode/launch.json
+.git_credentials
```
Javagedes, line 7: nit: this should probably be a rebase & ff with this commit being separate.

patina_dxe_core/src/image.rs
```diff
@@ -2196,4 +2196,317 @@ mod tests {
             assert_eq!(get_buffer_by_file_path(true, device_path_ptr), Ok((image, false, handle, 0)));
         });
     }
+
+    // Tests for memory protection error handling (Fix #176, Task #1030)
```
Javagedes: nit: I think you should leave you the task `1030` portion of the following comment(s) throughout the tests.

The mention of these tests being specifically associated with Fix 176 make sense though.

patina_dxe_core/src/image.rs
```diff
@@ -2196,4 +2196,317 @@ mod tests {
...
+                image_info_ptr: core::ptr::null_mut(),
+                image_device_path_ptr: core::ptr::null_mut(),
+                pe_info: pe_info.clone(),
+                relocation_data: Vec::new(),
+                image_base_page: 0x1000,
+                image_num_pages: 2,
+            };
+
+            // Call apply_image_memory_protections directly
+            let _result = super::apply_image_memory_protections(&pe_info, &private_info);
```
makubacki: Why is this line:

```rust
let _result = super::apply_image_memory_protections(&pe_info, &private_info);
```

And the assertions are on `result`?

```rust
                assert!(result.is_err(), "Protection should fail when section address is not in GCD");
                assert_eq!(result.unwrap_err(), EfiError::NotFound, "Expected NotFound from GCD descriptor lookup");
```

These are two different variables.

patina_dxe_core/src/image.rs
```diff
@@ -2196,4 +2196,317 @@ mod tests {
...
+        //
+        // Normally, the first add_memory_space call creates an initial
+        // NonExistent descriptor covering [0, maximum_address), so all subsequent lookups
+        // succeed. By skipping add_memory_space entirely, we keep the RBT empty, making
+        // ANY address lookup fail with NotFound.
+        //
+        // This scenario represents a corrupted or uninitialized GCD state where memory
+        // descriptors are missing - an edge case that the error handling should catch.
+
+        let result = test_support::with_global_lock(|| unsafe {
```
makubacki: This unsafe block covers more than it needs to, it should be scoped to the actual unsafe operation(s). Please add a `SAFETY:` comment for each new unsafe block as well.

patina_dxe_core/src/image.rs
garybeihl: But the panic is inside with_global_lock which catches it and returns Err. #[should_panic] expects an uncaught panic, right? Maybe it would be better to check the result returned by with_global_lock.

.gitignore
garybeihl: done - thanks

patina_dxe_core/src/image.rs
garybeihl: done - thanks!

patina_dxe_core/src/image.rs
garybeihl: fixed - thanks!

patina_dxe_core/src/image.rs
garybeihl: SAFETY comments added - thanks!

patina_dxe_core/src/image.rs
os-d: I don't think this overriding comment is necessary, with all the comments in the tests

patina_dxe_core/src/image.rs
```diff
@@ -2207,4 +2207,310 @@ mod tests {
...
+
+            // Should FAIL with NotFound because the GCD RBT is empty (no descriptors),
+            // so get_closest_idx returns None and get_memory_descriptor_for_address returns NotFound
+            assert!(result.is_err(), "Protection should fail when section address is not in GCD");
+            assert_eq!(result.unwrap_err(), EfiError::NotFound, "Expected NotFound from GCD descriptor lookup");
+        });
+
+        // In debug builds, debug_assert! panics and with_global_lock catches it, returning Err
+        // In release builds, the function completes normally and returns Ok
+        #[cfg(debug_assertions)]
```
os-d, line 2397: We were going to drop these, right? And just handle the debug_assertions case and later clean that up

garybeihl: Done - thanks!

patina_dxe_core/src/image.rs
garybeihl: Removed it - thanks!

patina_dxe_core/src/image.rs
os-d: Aren't you expecting these tests to hit the `debug_asserts`? In which case you'll need to add `#[should_panic]` to the tests.

garybeihl: The panic is caught internally by `test_support::with_global_lock()`, which uses `std::panic::catch_unwind()` to handle panics gracefully. The `#[should_panic]` attribute only works when the **test function itself** panics at the top level. Since our panic is caught inside `with_global_lock()`, the test function returns normally with an `Err(...)` Result, and `#[should_panic]` would never trigger.

os-d:
@Javagedes @makubacki can you re-review when you have a chance?

makubacki (approved):

Issue #583, opened 2025-07-14
[Task]: Add safety comments around `unsafe` blocks


makubacki:
### Feature Overview

The 2024 edition changes in the [Update to the 2024 Edition](https://github.com/OpenDevicePartnership/patina/pull/577) PR led to a lot of code sections being explicitly marked as `unsafe` that were not before. This was often around unsafe code within an unsafe function.

These unsafe blocks need SAFETY comments. This change tracks going through those changes and adding those comments.

We should also check if the code can be updated to not use unsafe at all or better alternatives in general during this work.

### Urgency

Medium

### Do you need maintainer feedback?

No maintainer feedback needed

---

Quick analysis of the current state:

| Crate | Undocumented Unsafe Count | Owner |
|-------|--------------------------|-------|
| patina_dxe_core | 403 | @Raymond-MS  (@makubacki - test code) |
| patina | 273 | @makubacki |
| patina_stacktrace | 70 | @vineelko |
| patina_internal_cpu | 41 | @os-d |
| patina_pi | 30 | @makubacki |
| patina_internal_collections | 28 | @Javagedes |
| patina_debugger | 20 | @os-d |
| patina_ffs | 19 | @joschock |
| patina_internal_device_path | 18 | @apop5 |
| patina_mm | 13 | @makubacki |
| patina_adv_logger | 12 | @cfernald |
| patina_macro | 4 | @Javagedes |
| patina_performance | 2 | @berlin-with0ut-return |
| patina_ffs_extractors | 0 | N/A |
| patina_internal_depex | 0 | N/A |
| patina_samples | 0 | N/A |

joschock:
https://github.com/OpenDevicePartnership/patina/pull/887 <- patina_ffs safety comments PR.

berlin-with0ut-return:
#888 - patina_performance and patina/performance safety fixes

apop5:
add "-W", "clippy::undocumented_unsafe_blocks" to the Makefile

makubacki:
patina_mm: [patina_mm: Add missing safety comments](https://github.com/OpenDevicePartnership/patina/pull/892)

apop5:
patina_internal_device_path https://github.com/OpenDevicePartnership/patina/pull/895

makubacki:
patina_pi (partially done, HOB content skipped intentionally as it should be refactored): https://github.com/OpenDevicePartnership/patina/pull/903

cfernald:
[...]

PR #1053, opened 2025-11-13
Enable unsafe block lint [Rebase & FF]

### LINKED ISSUE #761: [Task]: Enable the `undocumented_safety_blocks` clippy lint
### Request Description

Tracks enabling https://rust-lang.github.io/rust-clippy/master/#undocumented_unsafe_blocks to check that `unsafe` blocks and `impl`s without `// SAFETY:` comments are found by clippy.

I will make the change

No maintainer feedback needed

### Anything else?

_No response_

makubacki:
## Description

Enables the `undocumented_unsafe_blocks` clippy lint in the workspace for all crates except `patina_dxe_core` as tracked in https://github.com/OpenDevicePartnership/patina/issues/583.

---

**Enable the undocumented_unsafe_blocks lint**

Closes #761

Checks that `unsafe` blocks have safety comments by enabling the
following clippy lint.

https://rust-lang.github.io/rust-clippy/master/index.html#undocumented_unsafe_blocks

The lint is not enabled in `patina_dxe_core` at this time as it still
has a lot of unsafe blocks that need safety comments. Those will be
addressed in a future change.

---

**patina_performance: Replace unsafe test code**

Removes the need for unsafe by leaking the Box to get a static
reference.

---

**Add missing unsafe safety comments**

Adds safety comments missing for unsafe blocks in the following crates:

- patina
- patina_dxe_core
- patina_internal_collections
- patina_internal_cpu
- patina_internal_device_path
- patina_mm
- patina_performance
- patina_stacktrace
- patina_smbios

---

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- `cargo make clippy`

## Integration Instructions

- N/A

### DIFF
--- a/components/patina_smbios/Cargo.toml
+++ b/components/patina_smbios/Cargo.toml
@@ -26,5 +26,8 @@ enable_patina_tests = []
 mockall = ["dep:mockall", "std"]
 std = []
 
+[lints.clippy]
+undocumented_unsafe_blocks = "warn"
+
 [lints.rust]
 unexpected_cfgs = { level = "warn", check-cfg = ['cfg(coverage_nightly)'] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -86,3 +86,6 @@ overflow-checks = true
 lto = false
 incremental = false
 codegen-units = 1
+
+[workspace.lints.clippy]
+undocumented_unsafe_blocks = "warn"
--- a/components/patina_adv_logger/Cargo.toml
+++ b/components/patina_adv_logger/Cargo.toml
@@ -6,7 +6,10 @@ repository.workspace = true
 license.workspace = true
 edition.workspace = true
 readme = "README.md"
-description = "Advanced logger support."
+description = "Advanced logger component."
+
+[lints]
+workspace = true
 
 [[bin]]
 name = "advlog_parser"
--- a/components/patina_mm/Cargo.toml
+++ b/components/patina_mm/Cargo.toml
@@ -8,6 +8,9 @@ edition.workspace = true
 readme = "README.md"
 description = "Management Mode (MM) support for components."
 
+[lints]
+workspace = true
+
 [dependencies]
 cfg-if = { workspace = true }
 log = { workspace = true }
--- a/components/patina_performance/Cargo.toml
+++ b/components/patina_performance/Cargo.toml
@@ -6,7 +6,10 @@ repository.workspace = true
 license.workspace = true
 edition.workspace = true
 readme = "README.md"
-description = "Performance support for the DXE Core."
+description = "Performance measurement infrastructure."
+
+[lints]
+workspace = true
 
 [dependencies]
 r-efi = { workspace = true }
--- a/components/patina_samples/Cargo.toml
+++ b/components/patina_samples/Cargo.toml
@@ -6,7 +6,10 @@ repository.workspace = true
 license.workspace = true
 edition.workspace = true
 readme = "README.md"
-description = "Sample UEFI components for the DXE Core."
+description = "Sample components for testing and reference."
+
+[lints]
+workspace = true
 
 [dependencies]
 log = { workspace = true }
--- a/core/patina_debugger/Cargo.toml
+++ b/core/patina_debugger/Cargo.toml
@@ -8,6 +8,9 @@ edition.workspace = true
 readme = "README.md"
 description = "Debugger implementation for a Patina Core."
 
+[lints]
+workspace = true
+
 [dependencies]
 cfg-if = { workspace = true }
 gdbstub = { workspace = true }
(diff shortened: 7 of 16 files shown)


sdk/patina/src/boot_services.rs
```diff
@@ -921,7 +941,8 @@ pub trait BootServices {
     ///
     /// [UEFI Spec Documentation: 7.5.7. EFI_BOOT_SERVICES.CalculateCrc32()](https://uefi.org/specs/UEFI/2.10/07_Services_Boot_Services.html#efi-boot-services-calculatecrc32)
     fn calculate_crc_32<T: 'static>(&self, data: &T) -> Result<u32, efi::Status> {
-        // the mem size of ensure that the data pointer and size are correct.
+        // SAFETY: The reference guarantees that the pointer is valid and properly aligned.
+        // size_of::<T>() provides the exact size of the data being pointed to.
```
os-d, line 945: I agree with this comment and this doesn’t need to be done for this PR, but seems like we should be able to calculate the crc on the reference (or slice created from reference) instead of the pointer. I haven’t looked deeply but I assume the pattern is: Rust code calls safe fn, which calls the unsafe fn. C code calls the unsafe fn.

I wonder if we flip that and have C code call an unsafe fn that wraps the safe fn (converting a ptr to a ref) and Rust just has all safe in its path.

Anyway, can file an issue if you agree or just close it if you don’t.

components/patina_smbios/Cargo.toml
```diff
@@ -26,5 +26,8 @@ enable_patina_tests = []
 mockall = ["dep:mockall", "std"]
 std = []
 
+[lints.clippy]
+undocumented_unsafe_blocks = "warn"
```
Javagedes, line 30:
```suggestion
workspace = true
```

sdk/patina/src/component/storage.rs
```diff
@@ -503,14 +511,23 @@ impl<'s> UnsafeStorageCell<'s> {
...
+// Safety: &mut Storage parameter provides exclusive access to the entire storage. The Param implementation
+// ensures exclusive access by tracking config reads/writes in MetaData. init_state() asserts no conflicting
+// config access exists and marks all configs as exclusively written. get_param() uses storage_mut() which
+// requires the UnsafeStorageCell was created from mutable storage access.
 unsafe impl Param for &mut Storage {
     type State = ();
     type Item<'storage, 'state> = &'storage mut Storage;
 
+    // Safety: UnsafeStorageCell was created from &mut Storage (validated by init_state). storage_mut()
+    // returns the underlying mutable reference. The Param protocol ensures this is the only active borrow
```
Javagedes: I think this is what was meant.

```suggestion
    // returns the underlying mutable reference. The Param trait ensures this is the only active borrow
```

sdk/patina/src/component/storage.rs
```diff
@@ -503,14 +511,23 @@ impl<'s> UnsafeStorageCell<'s> {
...
+    // of storage data.
     unsafe fn get_param<'storage, 'state>(
         _state: &'state Self::State,
         storage: UnsafeStorageCell<'storage>,
     ) -> Self::Item<'storage, 'state> {
+        // Safety: UnsafeStorageCell was created with exclusive access. init_state ensured no conflicting
+        // config access. storage_mut() safety requirements are met by the Param protocol.
```
Javagedes: I think this is what was meant.

```suggestion
        // config access. storage_mut() safety requirements are met by the Param trait.
```

sdk/patina/src/boot_services.rs
makubacki: I think it's worth tracking in an issue.

sdk/patina/src/component/storage.rs
```diff
@@ -432,7 +432,15 @@ impl Storage {
 #[derive(Copy, Clone)]
 pub struct UnsafeStorageCell<'s>(*mut Storage, PhantomData<(&'s Storage, &'s UnsafeCell<Storage>)>);
 
+// Safety: UnsafeStorageCell wraps a raw pointer but does not own the data. The lifetime 's ensures the pointer
```
joschock:
```suggestion
// Safety: UnsafeStorageCell wraps a raw pointer but does not own the data. The lifetime specifier ensures the pointer
```

components/patina_smbios/Cargo.toml
makubacki: I wasn't able to directly use that here. Because `[lints.rust]
unexpected_cfgs = { level = "warn", check-cfg = ['cfg(coverage_nightly)'] }` is not in the workspace root, it seemed to be an all or nothing on inheriting from the workspace. I just left it as-is.

PR #1056, opened 2025-11-13
Refactor TplMutex

joschock:
## Description

Renames the tpl_lock.rs module to tpl_mutex and refactors the logic. Key improvements:
* Remove the dependency on BootServices for TPL manipulation. TplMutex now directly calls the raise_tpl/restore_tpl APIs directly rather than through the BootServices table. This greatly simplifies the usage model and testing of TplMutex. 

Also updates Synchronization chapter of the docbook.

This change pursuant to RFC 0021 https://github.com/OpenDevicePartnership/patina/pull/1036

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [x] Includes documentation?

## How This Was Tested

Boot testing on hardware platform. 

## Integration Instructions

N/A

### DIFF
--- a/patina_dxe_core/src/tpl_mutex.rs
+++ b/patina_dxe_core/src/tpl_mutex.rs
@@ -17,6 +17,7 @@ use core::{
     cell::UnsafeCell,
     fmt,
     ops::{Deref, DerefMut},
+    sync::atomic::{AtomicBool, Ordering},
 };
 
 use r_efi::efi;
@@ -26,7 +27,7 @@ use crate::events::{raise_tpl, restore_tpl};
 /// Used to guard data with a locked MUTEX and TPL level.
 pub struct TplMutex<T: ?Sized> {
     tpl_lock_level: efi::Tpl,
-    lock: UnsafeCell<bool>,
+    lock: AtomicBool,
     name: &'static str,
     data: UnsafeCell<T>,
 }
@@ -45,7 +46,7 @@ unsafe impl<T: ?Sized + Send> Send for TplGuard<'_, T> {}
 impl<T> TplMutex<T> {
     /// Instantiates a new TplMutex with the given TPL level, data object, and name string.
     pub const fn new(tpl_lock_level: efi::Tpl, data: T, name: &'static str) -> Self {
-        Self { tpl_lock_level, lock: UnsafeCell::new(false), data: UnsafeCell::new(data), name }
+        Self { tpl_lock_level, lock: AtomicBool::new(false), data: UnsafeCell::new(data), name }
     }
 }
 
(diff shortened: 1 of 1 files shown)


joschock:
Reviewer note: to see just the functional refactor, that is broken out to a separate commit from the rename and doc updates in PR history.

patina_dxe_core/src/tpl_mutex.rs
```diff
@@ -0,0 +1,239 @@
...
+    /// # Panics
+    ///
+    /// Attempting to acquire the lock while running at a TPL level higher than the lock's TPL level will panic due to
+    /// TPL inversion.
+    pub fn try_lock(&self) -> Option<TplGuard<'_, T>> {
+        let release_tpl = raise_tpl(self.tpl_lock_level);
+        // Safety: raw pointer on lock is used for volatile semantics; pointer is valid by construction, and exclusive
+        // access is guaranteed by TPL raising.
+        if !unsafe { self.lock.get().read_volatile() } {
+            // Safety: see prior comment.
```
cfernald: I think there is still a very niche double-entry scenario, though let me know if i'm missing something. If we have a TpmMutex at less than HIGH, it's still valid to raise the TPL to the same as the current value. So, if a caller was taking TplMutex.try_lock() on less than HIGH and took an interrupt between lines 76 and 79 and at some point in the interrupt handler tries to acquire the same lock it would succeed and enter the critical section. If it failed to release the lock before returning, then this would allow double entry.

However, I think this is fine if we assume that all locks will be released before the interrupt handler returns, but it might be worth calling out that assumption in the safety comment.

joschock: in order to not release the lock, it would have to somehow not drop the guard. I think this could only happen if you did something weird like somehow move the guard into a global static so that it outlives the scope where it was created. 

I'll add a comment. I flirted with making `new()` unsafe and documenting the invariants for safe TplMutex use there; do you think that's merited here?

joschock: Also, there are potential problematic constructions that violate TplLocking (both in Patina and EDK2 locks) For example, this: https://github.com/microsoft/mu_basecore/blob/3f900bac16f21462a7589ca7a6d226057abcb3a2/MdeModulePkg/Bus/Usb/UsbBusDxe/UsbHub.c#L1003. I think we need to look closer at these and convince our selves they are not bugs.

joschock: also, I'm happy to put back the atomic flag for the additional safety, if that's the thrust of the comment.

cfernald: I don't think this assumption is likely to blow up in our faces (famous last words), so i'd be happy just adding a line in the safety comment like "There is an assumption that if this lock is taken during an interrupt while acquiring the lock that it must be freed before the interrupt returns, or else a double entry may occur." or something to that affect. Just in case it happens in the future we have a starting point for the person debugging.

cfernald: I would feel more comfortable with the atomic honestly, if there is no perf difference, but I'm not sure where we landed on that.

joschock: perf impact is negligible (at least in perf measurements of actual boot). I'll put back the atomic for safety.

joschock: I put back the atomic. measured performance delta is in the noise - less than 10ms to overall boot time which is in the boot-to-boot variance.

joschock: fixed in latest push (now that pushes are working again).

makubacki (approved):

Issue #893, opened 2025-10-16
[Feature]: config_tables: Use the zerocopy crate to Reduce Unsafe Code Blocks


VivekYadav7272:
### Feature Overview

The zerocopy crate provides safe (wrapped unsafe) memory manipulations. This offers the ability to remove a lot of unsafe blocks in Patina and rely on zerocopy to convert to and from bytes, etc.

### Solution Overview

Use zerocopy in `config_tables` module to remove unsafe usage (I'm assuming the ones that pertains to pointer-object casting?)

### Urgency

Medium

### Are you going to implement the feature request?

I will implement the feature

### Do you need maintainer feedback?

Maintainer feedback requested

### Anything else?

It'd be nice to get more clarification on exactly the kind of unsafe usage we want prevented. To be specific, for example:

In `config_tables/memory_attributes_table.rs`:
``` rust
match core_allocate_pool(efi::BOOT_SERVICES_DATA, size_of::<efi::MemoryAttributesTable>()) {
            Ok(empty_ptr) => {
                if let Some(empty_mat) = unsafe { (empty_ptr as *mut efi::MemoryAttributesTable).as_mut() } {
...
```
1. Now this sort of unsafe usage seems like a good candidate for using `zerocopy`. However, regardless, it will still require a cast to a byte slice, which would be an unsafe operation. That could probably also be eliminated if we have a `Box`-like allocation API -- is that something that's in-scope for this feature or should we let the `unsafe` needed to cast into a byte slice remain?
1. Also, looks like `zerocopy` is primarily driven by derive macros and you can't provide custom implementations for those traits. But that means we have a recursive dependency for all children members to also have it implemented. In this case specifically, `pub struct MemoryAttributesTable(*mut efi::MemoryAttributesTable);` can't implement it until `efi::MemoryAttributesTable` does, which is an external dependency. Do I duplicate the struct? Is there some other way to implement these traits that I'm missing? Sorry if that's a stupid question.

VivekYadav7272:
Sorry btw I do not understand how to add an issue as a sub-issue. The instructions online from GitHub states that there should be a link/button below the parent issue's description but I couldn't find it.

Javagedes:
You probably cannot do it with your permissions. I've done it for you :)

VivekYadav7272:
cc @os-d

os-d:
@VivekYadav7272 a few things:
[...]

Issue #604, opened 2025-07-18
[Feature]: Use the zerocopy Crate to Reduce Unsafe Code Blocks


os-d:
### Feature Overview

The zerocopy crate provides safe (wrapped unsafe) memory manipulations. This offers the ability to remove a lot of unsafe blocks in Patina and rely on zerocopy to convert to and from bytes, etc.

### Solution Overview

Use zerocopy to do memory manipulations when possible.

### Urgency

Medium

### Are you going to implement the feature request?

Someone else needs to implement the feature

### Do you need maintainer feedback?

No maintainer feedback needed

makubacki:
I'd be happy to take a few crates if we want to split this up. I don't see a reason it all needs to be done in one PR.

os-d:
> I'd be happy to take a few crates if we want to split this up. I don't see a reason it all needs to be done in one PR.

Absolutely agreed. That was my intention with marking it as paired-programming (though can be independent paired-programming :), but I could have called that out.

Happy to divvy this up amongst as many folks as want to participate.

makubacki:
@cfernald will demonstrate next Thursday (8/7).

VivekYadav7272:
I'm interested in picking up a subset if you're planning to split this up

makubacki:
> I'm interested in picking up a subset if you're planning to split this up

@VivekYadav7272, we can split this up with individuals being assigned subissues for the areas they are interested in updating to use zerocopy. Please feel free to file a subissue for the area you're interested in.

VivekYadav7272:
> Please feel free to file a subissue for the area you're interested in.

Yup! I'm currently going through the documentation and perusing the codebase for such uses. When I feel confident I've found such an area I'll ping here and create a subissue 👍

os-d:
> > Please feel free to file a subissue for the area you're interested in.
> 
> Yup! I'm currently going through the documentation and perusing the codebase for such uses. When I feel confident I've found such an area I'll ping here and create a subissue 👍

@VivekYadav7272 one area I have in my backlog to convert to zerocopy are the two config tables: https://github.com/OpenDevicePartnership/patina/tree/main/patina_dxe_core/src/config_tables. Please feel free to take this if it is a helpful starting point.

Javagedes:
Adding this here too - `patina_ffs` appears to be a good candidate for zerocopy.

VivekYadav7272:
[...]

PR #1062, opened 2025-11-15
config_tables: Use zerocopy crate to reduce unsafe code blocks

VivekYadav7272:
Please see issue #893 for _why_ this change was needed.

I'm marking this as draft because I have some cleanup to do, but to give early visibility in case I've done something obviously dumb, I'm raising the draft PR anyways.

### DIFF
--- a/patina_dxe_core/src/config_tables/memory_attributes_table.rs
+++ b/patina_dxe_core/src/config_tables/memory_attributes_table.rs
@@ -7,7 +7,8 @@
 //! SPDX-License-Identifier: Apache-2.0
 //!
 extern crate alloc;
-use alloc::vec::Vec;
+use alloc::{boxed::Box, vec, vec::Vec};
+use zerocopy::{Immutable, IntoBytes};
 
 #[cfg(not(test))]
 use spin::Once;
@@ -23,9 +24,45 @@ use crate::{
 };
 use r_efi::efi;
 
+// To be upstreamed to r_efi.
+#[repr(C)]
+#[derive(Clone, Copy, Debug, IntoBytes, Immutable)]
+pub struct MemoryDescriptor {
+    pub r#type: u32,
+    // padding, required to implement zerocopy::IntoBytes
+    reserved: u32,
+    pub physical_start: efi::PhysicalAddress,
+    pub virtual_start: efi::VirtualAddress,
+    pub number_of_pages: u64,
+    pub attribute: u64,
+}
+
+impl MemoryDescriptor {
+    pub fn new(
+        r#type: u32,
+        physical_start: efi::PhysicalAddress,
+        virtual_start: efi::VirtualAddress,
+        number_of_pages: u64,
+        attribute: u64,
+    ) -> Self {
+        Self { r#type, reserved: 0, physical_start, virtual_start, number_of_pages, attribute }
+    }
+}
+
+#[repr(C)]
+#[derive(Clone, Copy, Debug, IntoBytes, Immutable)]
+pub struct MemoryAttributesTableInternal {
+    pub version: u32,
+    pub number_of_entries: u32,
+    pub descriptor_size: u32,
+    pub reserved: u32,
+    // can't have generic N-sized array, zerocopy gets confused
+    pub entry: [MemoryDescriptor; 0],
+}
+
 // create a wrapper struct so that we can create an install method on it. That way, we can have the install function
 // be a no-op until after ReadyToBoot
-pub struct MemoryAttributesTable(*mut efi::MemoryAttributesTable);
+pub struct MemoryAttributesTable(Box<[u8]>);
 
 // this is a flag to indicate that we have passed ReadyToBoot and can install the MAT on the next runtime memory
 // allocation/deallocation.
@@ -56,12 +93,58 @@ impl MemoryAttributesTable {
             core_install_memory_attributes_table()
         }
     }
+
+    fn from_mat_desc_list(mat_desc_list: Vec<MemoryDescriptor>) -> Self {
+        let mat_internal = MemoryAttributesTableInternal {
+            version: efi::MEMORY_ATTRIBUTES_TABLE_VERSION,
+            number_of_entries: mat_desc_list.len() as u32,
+            descriptor_size: size_of::<MemoryDescriptor>() as u32,
+            reserved: 0,
+            entry: [],
+        };
+        let buffer_size =
+            mat_desc_list.len() * size_of::<MemoryDescriptor>() + size_of::<MemoryAttributesTableInternal>();
+        let mut buff = vec![0u8; buffer_size].into_boxed_slice();
+        let (mat_buff, mat_desc_list_buff) = buff.split_at_mut(size_of::<MemoryAttributesTableInternal>());
+        mat_buff.copy_from_slice(mat_internal.as_bytes());
+        mat_desc_list_buff.copy_from_slice(mat_desc_list.as_slice().as_bytes());
+
+        Self(buff)
+    }
+
+    fn as_ptr(&self) -> *const MemoryAttributesTableInternal {
+        self.0.as_ptr() as *const _
+    }
+}
+
+impl Default for MemoryAttributesTable {
+    fn default() -> Self {
+        // Self-note: BootServicesData is basically just the default allocator (see documentation for MemoryManager trait)
+        // so replace with the simpler global allocation API.
+        let empty_mat = MemoryAttributesTableInternal {
+            version: 0,
+            number_of_entries: 0,
+            descriptor_size: 0,
+            reserved: 0,
+            entry: [],
+        };
+
+        let mut empty_mat_buff = Box::new([0u8; core::mem::size_of::<MemoryAttributesTableInternal>()]);
+        empty_mat_buff.copy_from_slice(empty_mat.as_bytes());
+
+        Self(empty_mat_buff)
+    }
 }
 
 impl Debug for MemoryAttributesTable {
     fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
-        let mat = unsafe { self.0.as_ref().expect("BAD MAT PTR") };
-        let entries = unsafe { slice::from_raw_parts(mat.entry.as_ptr(), mat.number_of_entries as usize) };
+        // SAFETY: All the ways to create Self always have a valid MemoryAttributesTableInternal at the base.
+        let mat = &unsafe { *(self.0.as_ptr() as *const _ as *const MemoryAttributesTableInternal) };
+        let entries = unsafe {
+            // This awkward pointer conversion lives until the zerocopy derives are upstreamed.
+            // But this should still be good since the representation should be identical.
+            slice::from_raw_parts(mat.entry.as_ptr() as *const efi::MemoryDescriptor, mat.number_of_entries as usize)
+        };
 
         writeln!(f, "MemoryAttributesTable {{")?;
         writeln!(f, "  version: {:#X}", mat.version)?;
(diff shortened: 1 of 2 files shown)


VivekYadav7272:
`debug_image_info_table.rs`'s most of `unsafe` usage came from things I could not mutate, because either the shape was UEFI-spec mandated (like the union structure), or an operation was UEFI-spec mandated (like the `volatile_{read|write}`s on `DebugImageInfoTableHeader`). So I did not see a huge reduction in `unsafe` there.

patina_dxe_core/src/config_tables/debug_image_info_table.rs
```diff
@@ -103,7 +104,12 @@ struct DebugImageInfoTableMetadata<'a> {
     slice: Box<[EfiDebugImageInfo]>,
 }
 
-static METADATA_TABLE: AtomicPtr<DebugImageInfoTableMetadata> = AtomicPtr::new(core::ptr::null_mut());
+// SAFETY: This structure is only ever modified under the METADATA_TABLE Mutex (see below).
+// This was needed because a raw pointer to EfiDebugImageInfo was not Send, but data race on that pointer is not possible
+// as that pointer is not shared with unprotected structures, and can only be accessed through the MutexGuard here.
+unsafe impl<'a> Send for DebugImageInfoTableMetadata<'a> {}
+
+static METADATA_TABLE: Mutex<Option<DebugImageInfoTableMetadata>> = Mutex::new(None);
```
joschock: `spin::Mutex` is problematic in the UEFI context because of the kind of multi-threading that UEFI supports. Unlike most high-level operating systems, UEFI only supports single-threaded execution with interrupts. This means that there is no concurrent execution. As such, if one context acquires the lock, and then an interrupt occurs to a higher-priority context that also attempts to acquire the lock, then deadlock will occur because the lower-priority context will not regain control until the higher-priority context yields, which it will not do since it is blocked attempting to acquire the lock.

If you really need Mutex+Guard capabilities here, recommend using TplMutex instead (implemented in the `tpl_lock.rs` module). TplMutex raises the priority level of the executing context while the lock is held, preventing pre-emption by other levels that would potentially want to acquire the lock, making it a safer option for avoiding deadlock.

os-d: We don't need a lock here, this is only ever invoked when loading an image, which won't happen during an interrupt. `Cell` should be sufficient.

patina_dxe_core/src/config_tables/debug_image_info_table.rs
```diff
@@ -184,18 +194,24 @@ pub(crate) fn core_new_debug_image_info_entry(
...
     image_handle: efi::Handle,
 ) {
-    // This is a very funny check for null because it is working around an LLVM bug where checking is_null() or variations
-    // of that on a load of an atomic pointer causes improper code generation and LLVM to crash. So, this check is a workaround
-    // to check if the pointer is in the first page of memory, which is a valid check for null in this case, as we mark
-    // that entire page as invalid. LLVM issue: https://github.com/llvm/llvm-project/issues/137152.
-    let metadata_table = METADATA_TABLE.load(Ordering::SeqCst);
-    if metadata_table < UEFI_PAGE_SIZE as *mut DebugImageInfoTableMetadata {
+    // Self-note: Might need to get someone's opinion on this, but I assume that this doesn't apply for a None check
+    // only a null right? Worried because it's likely that None also has a 0x0 repr, and perhaps that's the real problem?
```
os-d: AFAIK, we should be able to drop this odd check when not using an atomic. Also though, looking at the LLVM bug, it has been merged and I believe released, so we might not need this in general

patina_dxe_core/src/config_tables/memory_attributes_table.rs
```diff
@@ -7,30 +7,68 @@
...
+    allocator::{MemoryDescriptorSlice, get_memory_map_descriptors},
     config_tables::core_install_configuration_table,
     events::EVENT_DB,
     systemtables,
 };
 use r_efi::efi;
 
 // We cache the MAT here because we need to free it in whenever we get a new runtime code/data allocation
-static MEMORY_ATTRIBUTES_TABLE: AtomicPtr<c_void> = AtomicPtr::new(core::ptr::null_mut());
+static MEMORY_ATTRIBUTES_TABLE: Mutex<Option<MemoryAttributesTable>> = Mutex::new(None);
```
os-d: This one we can use a TplMutex for. It probably isn't needed, but this does get triggered when we allocate runtime memory, which can happen during an interrupt.

os-d (commented):
Thanks for the contribution! I think the general approach is okay, just a few notes on the Mutex vs Cell/TplMutex.

patina_dxe_core/src/config_tables/memory_attributes_table.rs
VivekYadav7272: Since raising this PR, some of the issues brought up here have already been addressed. The global static no longer exists now so I've rebased my code to take advantage of that. Shouldn't be an issue now.

patina_dxe_core/src/config_tables/debug_image_info_table.rs
VivekYadav7272: Got it, I'll remove the note comments now.

patina_dxe_core/src/config_tables/debug_image_info_table.rs
VivekYadav7272: My `DebugImageInfoTable` changes seem to already have been merged in the commit that reworked atomics, although it uses `RwLock` instead of Cell. Perhaps that still exposes us to potential interrupt pre-emption deadlocks since there are a bunch of `.write()` calls.

joschock: Oh - I forgot you had this PR outstanding - if there are additional improvements here that are not part of that commit, please feel free to rebase. 

With respect to pre-emption - all the non-unit-test calls to `.write()` occur as part of an initialization flow (in `initialize_debug_image_info_table`) before the event subsystem is up, so there should be no danger of pre-emption at that point.

patina_dxe_core/src/config_tables/memory_attributes_table.rs
```diff
@@ -56,12 +93,58 @@ impl MemoryAttributesTable {
...
+    }
+
+    fn as_ptr(&self) -> *const MemoryAttributesTableInternal {
+        self.0.as_ptr() as *const _
+    }
+}
+
+impl Default for MemoryAttributesTable {
+    fn default() -> Self {
+        // Self-note: BootServicesData is basically just the default allocator (see documentation for MemoryManager trait)
```
joschock, line 122: It's not clear how this comment is related to the code at hand? I think you are noting that the `Box` created here will be in EBSD bucket, so maybe something like `// default MAT will be created in the Efi Boot Services Data memory type using the default allocator`

patina_dxe_core/src/config_tables/memory_attributes_table.rs
```diff
@@ -157,83 +225,47 @@ pub fn core_install_memory_attributes_table() {
...
-                        // free the old MAT table if we have one
-                        if let Err(err) = core_free_pool(current_ptr.as_ptr()) {
-                            log::error!("Error freeing previous MAT pointer: {err:#X?}");
-                        }
-                    }
-                    Ok(None) => (),
-                }
+        Ok(_) => {
+            // free the old MAT table if we have one
+            let current_mat = get_configuration_table(&efi::MEMORY_ATTRIBUTES_TABLE_GUID);
```
joschock, line 259: I think this is going to return the mat you just installed - i.e. I think you're going to end up freeing the new table, not the old one. The `Ok` case for `core_install_configuration_table` should return the old pointer, and that's the one I think you want to free:

```rust
Ok(old_ptr) => {
  let current_mat = get_configuration_table(&efi::MEMORY_ATTRIBUTES_TABLE_GUID);
  println!("Free this one: {:?} not this one: {:?}", old_ptr, current_mat);
}

File: patina_dxe_core/src/config_tables/memory_attributes_table.rs [hunk 6]
```diff
@@ -157,83 +225,47 @@ pub fn core_install_memory_attributes_table() {
...
-    // allocate memory for the MAT and publish it
-    let buffer_size =
-        mat_desc_list.len() * size_of::<efi::MemoryDescriptor>() + size_of::<efi::MemoryAttributesTable>();
-    match core_allocate_pool(efi::BOOT_SERVICES_DATA, buffer_size) {
-        Err(err) => {
-            log::error!("Failed to allocate memory for the MAT! Status {err:#X?}");
+    let mat = MemoryAttributesTable::from_mat_desc_list(mat_desc_list);
+    let mat_ptr = mat.as_ptr();
+
+    match core_install_configuration_table(efi::MEMORY_ATTRIBUTES_TABLE_GUID, mat_ptr as *mut c_void, st) {
```
joschock, line 251: Here you are putting `mat_ptr` into a public table (i.e. one that is visible outside of the patina core to e.g. drivers). However, because you don't leak the mat, I think it will get auto-dropped at the end of the function, meaning that the pointer you installed in the global config table is going to point to freed memory.

patina_dxe_core/src/config_tables/memory_attributes_table.rs
```diff
@@ -157,83 +225,47 @@ pub fn core_install_memory_attributes_table() {
...
-                        }
-                    }
-                    Ok(None) => (),
-                }
+        Ok(_) => {
+            // free the old MAT table if we have one
+            let current_mat = get_configuration_table(&efi::MEMORY_ATTRIBUTES_TABLE_GUID);
+            log::info!("Dumping MAT: {mat:?}");
+            if let Some(current_mat) = current_mat
+                && let Err(err) = core_free_pool(current_mat.as_ptr())
```
joschock, line 262: `current_mat` was allocated using `Box`, it is invalid to free it here using `core_free_pool`. `core_free_pool` assumes that the pointer in question was allocated with `core_allocate_pool` and that an `AllocationInfo` structure is associated with the allocation. That would not be the case if it was directly allocated from the GlobalAllocator via `Box`. If you are going to allocate with `Box`, then it needs to be freed either by dropping the box directly or by calling the global allocator `dealloc` function directly with the pointer + layout.  

I think your best bet is to create the raw pointer with `Box::into_raw` (this would also fix the issue where it gets dropped when it goes out of scope leaving a dangling pointer in the config table per my other comment), and then reclaim it here with `Box::from_raw`. However, the type info on the raw pointer is going to be erased when it goes into the config table (all pointers in there are typed `c_void`); so you need to engage in casting shenanigans to convert `*mut c_void` back to `*mut [u8]` with the right length that you need to reconstruct the box.

VivekYadav7272: Yeah that's true, I forgot about this during the rebase. Previously with a global member, this was much easier, since I effectively had something to "remember" the type for me. So, there wouldn't have been a need for unsafe casting. Freeing would have been as simple as a swap operation, hence dropping the old one elegantly. Do you think that's worth re-introducing a private global member for? After all the goal is to avoid unsafe (taken care of by TplMutex?)

patina_dxe_core/src/config_tables/memory_attributes_table.rs
VivekYadav7272: Yeah the rebase was a bit sloppy. Before I'd taken care of this because the struct was stored globally, so the lifetime was 'static.

patina_dxe_core/src/config_tables/memory_attributes_table.rs
VivekYadav7272: Ah ofcourse, that was a dumb mistake.

patina_dxe_core/src/config_tables/memory_attributes_table.rs
VivekYadav7272: Fair. Will fix that

patina_dxe_core/src/config_tables/memory_attributes_table.rs
joschock: I think it is the case that if you rely on native rust allocations (via `Box`) that it is easier to go with a design where Rust always knows about the allocation - and that implies a global static. 

The nature of the MAT table being an externally observable table in a system table structure means that it is inherently outside of the Rust memory model. For example, from a strictly API-centric viewpoint, it is entirely UEFI-spec legal (though a terrible idea and not in step with the expected design) for an external agent to delete this table from the global table and replace it with its own pointer, or even worse, modify the table contents. Neither of these are legitimate things to do with MAT specifically, but the system does not prevent them - so taking such a pointer and casting it back into the `Box` for dropping it is inherently unsafe because we rely on the good behavior of external code. And you could extend this argument to basically any approach - whatever way we handle this, it's always going to have an element of unsafety about it because of the external visibility of this table. 

The question is how to best manage this unsafety. This is a matter for opinions :) My opinion is that maybe the Rust allocation apparatus is not helpful here, because it makes a lot of things (like `drop`) implicit in an environment where maybe not all of the assumptions baked into the Rust memory model hold. If you use direct allocations (via `core_allocate_pool` and `core_free_pool` for example), it's a bit more explicit what is being done to manage the memory, but it also means opting out of the Rust memory safety guarantees. 

I can see different approaches working here; the important thing is to ensure we don't end up twisted around a broken hybrid.

makubacki:
@VivekYadav7272, are you still planning to work on this PR?

VivekYadav7272:
> @VivekYadav7272, are you still planning to work on this PR?

Hi, sorry, I wasn't able to find enough time and by that time this issue was already resolved by other PRs and commits, atleast that was my impression. https://github.com/OpenDevicePartnership/patina/issues/604#issuecomment-3826437792

makubacki:
> > @VivekYadav7272, are you still planning to work on this PR?
> 
> Hi, sorry, I wasn't able to find enough time and by that time this issue was already resolved by other PRs and commits, atleast that was my impression. [#604 (comment)](https://github.com/OpenDevicePartnership/patina/issues/604#issuecomment-3826437792)

I assume that meant a "great effort" had been made to move code to adopt zerocopy but there is still opportunity for further work. In any case, I'll close this PR for now due to its age. I think it will also help get a faster response for a new PR to be created if needed. Thanks for your contributions to the project.

makubacki:
Closing

PR #1063, opened 2025-11-16
AARCH64 GIC v3: add support for booting on a primary core that is not the first entry in the GIC Redistributor.

joschock:
## Description

The prior implementation of the gic_manager module and associated code assumed that the system boot core would be the first entry in the GIC redistributor. However, that is not always the case. This PR refactors the gic_manager module, brings it up to the most recent version of the `arm-gic` crate, and adds support for booting from a core that is not the first entry in the GIC redistributor.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Verified interrupt functionality on AARCH64 hardware platform booting from a core that isn't the first entry in the GIC redistributor. 

## Integration Instructions

N/A

### DIFF
--- a/core/patina_internal_cpu/src/interrupts/aarch64/gic_manager.rs
+++ b/core/patina_internal_cpu/src/interrupts/aarch64/gic_manager.rs
@@ -102,8 +102,8 @@ impl AArch64InterruptInitializer {
         // Convert raw GIC address pointers to appropriate types.
         // Safety: function safety requirements guarantee exclusive access to the GICR registers.
         let (gicd, gicr) = unsafe {
-            let gicd = UniqueMmioPointer::new(NonNull::new(gicd_base as _).unwrap());
-            let gicr = NonNull::new(gicr_base as _).unwrap();
+            let gicd = UniqueMmioPointer::new(NonNull::new(gicd_base as _).ok_or(EfiError::InvalidParameter)?);
+            let gicr = NonNull::new(gicr_base as _).ok_or(EfiError::InvalidParameter)?;
             (gicd, gicr)
         };
 
@@ -121,7 +121,10 @@ impl AArch64InterruptInitializer {
         }
 
         log::info!("Total Redistributors: {}, Current CPU Redistributor Index: {}", r_count, cpu_r_idx);
-        assert!(cpu_r_idx != usize::MAX, "Failed to find redistributor for current cpu");
+        if cpu_r_idx == usize::MAX {
+            log::error!("Failed to find redistributor for current cpu");
+            return Err(EfiError::DeviceError);
+        }
 
         // Initialize the GIC:
         // Enable affinity routing and non-secure group 1 interrupts.


core/patina_internal_cpu/src/interrupts/aarch64/gic_manager.rs
```diff
@@ -65,54 +68,87 @@ fn get_system_gic_version() -> GicVersion {
...
-    Ok(gic_v3)
-}
+        //Determine the linear index of the current core and the count of cores.
+        let mut cpu_r_idx = usize::MAX;
+        let mut r_count = 0;
 
-pub struct AArch64InterruptInitializer<'a> {
-    pub gic_v3: GicV3<'a>,
-}
+        let mpidr = read_sysreg!(MPIDR_EL1) & 0x0000_00ff_00ff_ffffu64;
```
os-d: nit: define the mask?

core/patina_internal_cpu/src/interrupts/aarch64/gic_manager.rs
```diff
@@ -174,25 +224,33 @@ impl AArch64InterruptInitializer<'_> {
         let int_id = self.source_to_intid(interrupt_source)?;
 
         let level = if int_id.is_private() {
-            let mut sgi = self.gic_v3.sgi_ptr(0);
-            field!(sgi, icfgr).get(index).unwrap().read() & bit != 0
+            // arm-gic does not presently provide a way to directly read the interrupt state for a given interrupt.
+            // so save the current redistributor context and read the ICFGR from there.
+            let redistributor = self.gic_v3.redistributor(self.cpu_r_idx).expect("Invalid redistributor");
```
os-d, line 236: Same logic in a few places, worth a helper func and/or upstreaming something to arm-gic?

patina_dxe_core/src/perf_timer.rs
```diff
@@ -88,13 +88,11 @@ pub(crate) fn arch_cpu_count() -> u64 {
 /// Skip coverage as any value could be valid, including 0.
 #[coverage(off)]
 pub(crate) fn arch_perf_frequency() -> u64 {
-    let mut frequency = 0u64;
```
os-d: This has now been taken care of in another pr

cspell.yml
```diff
@@ -1,194 +1,194 @@
-# @file cspell.yml
```
os-d, line 1: line endings?

core/patina_internal_cpu/src/interrupts/aarch64/gic_manager.rs
joschock: I do plan to push something to arm-gic if/when I have time; I think there should be an accessor for these regs implemented. For now I don't think a helper func is needed.

patina_dxe_core/src/perf_timer.rs
joschock: rebased, removed, and description updated.

core/patina_internal_cpu/src/interrupts/aarch64/gic_manager.rs
```diff
@@ -65,54 +72,87 @@ fn get_system_gic_version() -> GicVersion {
...
-    // Set binary point reg to 0x7 (no preemption)
-    // Safety: this is a legal value for BPR1 register.
-    // Refer to "Arm Generic Interrupt Controller Architecture Specification GIC
-    // architecture version 3 and Version 4" (Arm IHI 0069H.b ID041224)
-    // 12.2.5: "ICC_BPR1_EL1, Interrupt Controller Binary Point Register 1"
-    write_sysreg!(reg ICC_BPR1_EL1, 0x7u64);
+        // Convert raw GIC address pointers to appropriate types.
+        // Safety: function safety requirements guarantee exclusive access to the GICR registers.
+        let (gicd, gicr) = unsafe {
+            let gicd = UniqueMmioPointer::new(NonNull::new(gicd_base as _).unwrap());
```
makubacki: I know the safety comment says the caller should pass a alid `gicd_base`, but given this function alreadys returns `Result`, do you think there is value in checking for null separately and returning an error instead of panicking if `gicd_base` is null?

joschock: No preference on my part; my operating intuition is that panic makes sense because it would be a programming error to pass null (as opposed to the "current hardware is unsupported" error that is the other error return). Happy to switch it to an error return if preferred.

core/patina_internal_cpu/src/interrupts/aarch64/gic_manager.rs
```diff
@@ -65,54 +72,87 @@ fn get_system_gic_version() -> GicVersion {
...
+        // Safety: function safety requirements guarantee exclusive access to the GICR registers.
+        for (index, redistributor) in unsafe { GicRedistributorIterator::new(gicr, false) }.enumerate() {
+            r_count = index + 1;
+            if redistributor.typer().core_mpidr() == mpidr {
+                cpu_r_idx = index;
+            }
+        }
+
+        log::info!("Total Redistributors: {}, Current CPU Redistributor Index: {}", r_count, cpu_r_idx);
+        assert!(cpu_r_idx != usize::MAX, "Failed to find redistributor for current cpu");
```
makubacki: It seems like this could also return an error.


```suggestion
        if cpu_r_idx == usize::MAX {
            log::error!("Failed to find redistributor for current CPU (MPIDR: 0x{:x})", mpidr);
            return Err(EfiError::NotFound);
        }
```

core/patina_internal_cpu/src/interrupts/aarch64/gic_manager.rs
```diff
@@ -65,54 +72,87 @@ fn get_system_gic_version() -> GicVersion {
...
+    ///
+    /// * Enable affinity routing and non-secure group 1 interrupts.
+    /// * Enable gic cpu interface
+    /// * Enable gic distributor
+    ///
+    /// # Safety
+    ///
+    /// `gicd_base` must point to the GIC Distributor register space.
+    ///
+    /// `gircr_base` must point to the GIC Redistributor register space.
```
makubacki:
```suggestion
    /// `gicr_base` must point to the GIC Redistributor register space.
```

joschock: good catch. I just pasted it into cspell without thinking. will fix.

core/patina_internal_cpu/src/interrupts/aarch64/gic_manager.rs
makubacki: I don't really have a strong preference either, I'd lean toward returning "invalid parameter", but I mainly just wanted to get your thoughts based on the usage.

joschock: replaced with error returns.

core/patina_internal_cpu/src/interrupts/aarch64/gic_manager.rs
joschock: Replaced with error returns.

core/patina_internal_cpu/src/interrupts/aarch64/gic_manager.rs
joschock: replaced with error returns.

PR #1065, opened 2025-11-16
Fix a potential race condition in EventDb/Allocator.

joschock:
EventDb/Allocator: fix a circular lock dependency where gcd_map_change() callback could signal an event while downstream of an event lock.

This PR also removes the EVENT_DB_INITIALIZED static which should not be required and masks this issue in test. 

## Description

In some scenarios, an gcd_map_change() could occur as a result of an allocation expansion while the event lock was held. Since signalling an event required holding the event lock, this would cause re-entrancy and deadlock. 

An example of such a deadlock is when a runtime event (such as virtual address change) is created, it goes into `runtime::add_runtime_event()` which then allocates the event structure in the runtime allocator and pushes it on a linked list. This is a pool allocation; but if that pool allocation causes an expand in the runtime allocator, this will then trigger a GCD change which will then signal an event for memory map change resulting in a re-entrant deadlock of this nature. This can be seen in unit testing if only the first commit of this PR is taken. 

This change adds a mitigation for this by allowing events to be signaled even if the event lock is held - if a signal is attempted while the event lock is held, it is instead placed in a pending event queue; once the event lock is released, then any pending events signaled while the lock was held are drained (in signal order) into the event database. 

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Booted with aarch64 platform hardware with these changes.

## Integration Instructions

N/A

### DIFF
--- a/patina_dxe_core/src/event_db.rs
+++ b/patina_dxe_core/src/event_db.rs
@@ -1515,6 +1515,103 @@ mod tests {
     }
 
     #[test]
+    fn signal_event_under_event_lock_should_use_pending_queue() {
+        with_locked_state(|| {
+            static SPIN_LOCKED_EVENT_DB: SpinLockedEventDb = SpinLockedEventDb::new();
+            let mut events: Vec<efi::Event> = Vec::new();
+            // create some non-grouped events.
+            for _ in 0..10 {
+                events.push(
+                    SPIN_LOCKED_EVENT_DB
+                        .create_event(
+                            efi::EVT_TIMER | efi::EVT_NOTIFY_SIGNAL,
+                            efi::TPL_NOTIFY,
+                            Some(test_notify_function),
+                            None,
+                            None,
+                        )
+                        .unwrap(),
+                );
+            }
+
+            let uuid = Uuid::from_str("aefcf33c-ce02-47b4-89f6-4bacdeda3377").unwrap();
+            let group1 = efi::Guid::from_bytes(uuid.as_bytes());
+            // create some grouped events.
+            let mut grouped_events: Vec<efi::Event> = Vec::new();
+            for _ in 0..10 {
+                grouped_events.push(
+                    SPIN_LOCKED_EVENT_DB
+                        .create_event(
+                            efi::EVT_TIMER | efi::EVT_NOTIFY_SIGNAL,
+                            efi::TPL_NOTIFY,
+                            Some(test_notify_function),
+                            None,
+                            Some(group1),
+                        )
+                        .unwrap(),
+                );
+            }
+
+            let mut guard = SPIN_LOCKED_EVENT_DB.lock();
+            // signal all events while holding the lock.
+            for event in &events {
+                SPIN_LOCKED_EVENT_DB.signal_event(*event).unwrap();
+            }
+
+            //signal the group
+            SPIN_LOCKED_EVENT_DB.signal_group(group1);
+
+            // check ordering of pending_signals and ensure none are marked signaled.
+            for (event, pending_signal) in events.iter().zip(guard.pending_signals.lock().iter()) {
+                if let PendingSignals::Event(pending_event) = pending_signal {
+                    assert_eq!(event, pending_event);
+                    assert!(!guard.is_signaled(*event));
+                } else {
+                    panic!("Unexpected Group in guard.pending_signals");
+                }
+            }
+
+            // last element should be a group signal.
+            if let Some(PendingSignals::Group(pending_group)) = guard.pending_signals.lock().last() {
+                assert_eq!(group1, *pending_group);
+            } else {
+                panic!("group not in guard.pending_signals");
+            }
+
+            // size should be the number of ungrouped events plus one group signal.
+            assert_eq!(guard.pending_signals.lock().len(), events.len() + 1);
+
+            // none of the group signals should be signaled.
+            for event in &grouped_events {
+                assert!(!guard.is_signaled(*event));
+            }
+
+            // none of these should be pending while the lock is held.
+            assert!(guard.pending_notifies.is_empty());
+
+            // drop the lock. This should signal all the events.
+            drop(guard);
+
+            // now all events should be signaled.
+            for event in &events {
+                assert!(SPIN_LOCKED_EVENT_DB.is_signaled(*event));
+            }
+            for event in &grouped_events {
+                assert!(SPIN_LOCKED_EVENT_DB.is_signaled(*event));
+            }
+
+            // check the notify queue ordering. note: grouped events are expected to be notified in reverse creation order.
+            let ordered_events_iter = events.iter().chain(grouped_events.iter().rev());
+
+            let notify_iter = iter::from_fn(|| SPIN_LOCKED_EVENT_DB.consume_next_event_notify(efi::TPL_APPLICATION));
+
+            for (event, notify) in ordered_events_iter.zip(notify_iter.map(|n| n.event)) {
+                assert_eq!(*event, notify);
+            }
+        });
+    }
+
+    #[test]
     fn queue_event_notify_should_queue_event_notify() {
         with_locked_state(|| {
             static SPIN_LOCKED_EVENT_DB: SpinLockedEventDb = SpinLockedEventDb::new();


Javagedes:
This really seems like we should be reworking the way we handle how we trigger event notifies so that the event db is not locked when the notify callback is triggered. Is there a reason / limitation we cannot do that?

To trigger an event we just need the function pointer and context pointer. Pointers can be copied, So we could just pop a copy of those out of the event db, release the lock, trigger the callback, Once we return from the callback, set the event to not triggered.

joschock:
> This really seems like we should be reworking the way we handle how we trigger event notifies so that the event db is not locked when the notify callback is triggered. Is there a reason / limitation we cannot do that?
> 
> To trigger an event we just need the function pointer and context pointer. Pointers can be copied, So we could just pop a copy of those out of the event db, release the lock, trigger the callback, Once we return from the callback, set the event to not triggered.

The event lock is not held during event notify callbacks. Those are issued in the loop in `restore_tpl`, and are done so with no locks held from the core side. That's one of the reasons why restore_tpl is a bit squirrelly as it can't hold the event db lock during the calls (and so, for example, can't use a simple iterator over the event queue). 

In this case, it's not that an event notify function is _called_ with the lock held, it's that an event is _signalled_ with the lock held. The re-entrancy path looks something like:

Outside call to EventDb primitive that acquires the event lock -> heap allocation within the EventDb code -> allocator expand allocates pages from GCD -> GCD allocation calls EventDb.signal_event() which tries to take the EventDb lock (causing deadlock/assert) to queue the event within the event_db structs for eventual callback during restore_tpl once conditions are met. 

The triggering condition of having an allocator expansion during an EventDb call is actually pretty rare - I had been seeing this very occasionally during unit test runs; haven't ever seen it on real HW.

Javagedes:
> > This really seems like we should be reworking the way we handle how we trigger event notifies so that the event db is not locked when the notify callback is triggered. Is there a reason / limitation we cannot do that?
> > To trigger an event we just need the function pointer and context pointer. Pointers can be copied, So we could just pop a copy of those out of the event db, release the lock, trigger the callback, Once we return from the callback, set the event to not triggered.
> 
> The event lock is not held during event notify callbacks. Those are issued in the loop in `restore_tpl`, and are done so with no locks held from the core side. That's one of the reasons why restore_tpl is a bit squirrelly as it can't hold the event db lock during the calls (and so, for example, can't use a simple iterator over the event queue).
> 
> In this case, it's not that an event notify function is _called_ with the lock held, it's that an event is _signalled_ with the lock held. The re-entrancy path looks something like:
> 
> Outside call to EventDb primitive that acquires the event lock -> heap allocation within the EventDb code -> allocator expand allocates pages from GCD -> GCD allocation calls EventDb.signal_event() which tries to take the EventDb lock (causing deadlock/assert) to queue the event within the event_db structs for eventual callback during restore_tpl once conditions are met.
> 
> The triggering condition of having an allocator expansion during an EventDb call is actually pretty rare - I had been seeing this very occasionally during unit test runs; haven't ever seen it on real HW.

Ah I see. It's an allocation happening inside the EventDb. Did you narrow it down to what is allocating? Can we avoid allocations in general?

joschock:
> Ah I see. It's an allocation happening inside the EventDb. Did you narrow it down to what is allocating? Can we avoid allocations in general?

I did when I coded it a few weeks back (as part of the initial interrupt-based atomics RFC); but unfortunately it's hard to repro and I don't have the details to hand any more. I _think_ it might have been the clone in event_db::timer_tick(), but less than 50% confidence on that.

os-d (approved):

Javagedes:
> > Ah I see. It's an allocation happening inside the EventDb. Did you narrow it down to what is allocating? Can we avoid allocations in general?
> 
> I did when I coded it a few weeks back (as part of the initial interrupt-based atomics RFC); but unfortunately, it's hard to repro and I don't have the details to hand any more. I _think_ it might have been the clone in event_db::timer_tick(), but less than 50% confidence on that.

I don't think it would be the clone. That puts it on the stack. It is almost assuredly inserting either an Event or notify that is causing a re-allocation of the btreemap.

My opinion on this change is that this seems incredibly intrusive for such a minor thing. It's touching two entirely different subsystems. I understand the need for this fix, but I'm trying to think of a solution that does not require a push and/or drain on every touch of the event db. Not only is it intrusive but it seems like it would cause unexpected timing issues as we won't be firing events when appropriate.

In the documentation for the event DB, it is already mentioned that the BTreeMap is less performant than the C implementation for lookup. The only solution I have been able to think of is switching back to the C implementation which is a fixed-sized array of linked lists. I think we could make an elegant "rusty" solution of that and would allow for the allocation of the event to be performed prior to locking and inserting the event / notify.

I would be willing to do this work rather than getting this in, unless this is incredibly urgent.

joschock:
> > > Ah I see. It's an allocation happening inside the EventDb. Did you narrow it down to what is allocating? Can we avoid allocations in general?
> > 
> > 
> > I did when I coded it a few weeks back (as part of the initial interrupt-based atomics RFC); but unfortunately, it's hard to repro and I don't have the details to hand any more. I _think_ it might have been the clone in event_db::timer_tick(), but less than 50% confidence on that.
> 
> I don't think it would be the clone. That puts it on the stack. It is almost assuredly inserting either an Event or notify that is causing a re-allocation of the btreemap.
> 
> My opinion on this change is that this seems incredibly intrusive for such a minor thing It's touching two entirely different subsystems. I understand the need for this fix, but I'm trying to think of a solution that does not require a push and/or drain on every touch of the event db. Not only is it intrusive but it seems like it would causing some unexpected timing issues as we won't be firing events when appropriate.
> 
> In the documentation for the event DB, it is already mentioned that the BTreeMap is less performant than the C implementation for lookup. The only solution I have been able to think of is switching back to the C implementation which is a fixed-sized array of linked lists. I think we could make an elegant "rusty" solution of that and would allow for the allocation of the event to be performed prior to locking and inserting the event / notify.
> 
> I would be willing to do this work rather than getting this in, unless this is incredibly urgent.

This is not super urgent, but I think it _could_ eventually show up in a real system at which point it would probably be blocking. 

One thing I'll note is that the main change only touches the eventdb subsystem. The change in the allocator is mostly a behavioral optimization which I think is a good behavioral change to avoid allocator thrash but could be PR'd separately as an unrelated change if desired. 

I have no strong objections to a refactor along the lines you suggest (although I do prefer that we keep the data structures as simple and "mainstream" as possible - when it comes to BtreeSet vs. RollOurOwnTplPriorityQueue, I have a refutable bias towards the former).

But it's not clear that switching up the data structures for the pending queue would actually remove this re-entrancy problem in the general case. The only way to decouple the two subsystems for certain is to either forbid heap allocation downstream of any event-lock, or forbid event signaling in the allocation subsystem. If either of those are allowed, then I think you need an alternate path for signal_event() that doesn't require the main event lock. 

The current proposed change in this PR also retains a flavor of this problem - what if you `signal_event` from an allocation downstream of `signal_event`? I'm not sure that can happen, but I'm also not sure it won't - so this change may not even fully resolve the issue. It may be that `signal_event` needs to further move to a `try_lock()` on `pending_signals` and make `signal_event` a fallible call. 

The EDK2 core gets around this by dropping and re-acquiring the event lock in the middle of `CoreSignalEvent` but that would require a significant restructuring eventdb with a more complicated granular locking approach, and I'm not enthusiastic about the complexity that would introduce. In some ways, this approach of having a separately locked `pending_signals` queue to evade the main event lock is spiritually akin to that approach. 

I agree that this change is intrusive, and I'm open to alternative approaches - but I'm not sure if the "EventDb can't allocate" approach is less intrusive.

On Performance:
I don't think the performance considerations are dominant here; theoretically the BTreeSet used in pending_notifies is less performant than the C queue implementation - but `pending_notifies` is usually going to be very small (definitely less than 10 unless something really weird is happening), so it's not clear that a more complicated data structure is justified on a performance basis. 

The larger perf impact is likely the main BtreeMap holding the main eventdb, but data seems to indicate that there isn't a significant performance hit to overall boot time from that either. If there is a perf impact on the main eventdb BTree, it could probably be addressed by moving to a HashMap instance.

Javagedes:
I only mentioned the performance comment in the EventDb struct definition as an attempt at some "support" for re-working the event db :)

I guess I would need to take a deeper look at the advanced logger, but only heap allocations would trigger this path, and the only thing I saw from a brief look at the code was inserting things into the BTreeMap. The map may re-allocate to double its size. Maybe we are using vecs or other heap allocated types somewhere else that I did not see. If we are allocating more than that, I feel like it could be done differently. I do agree that enforcing no allocations in the event DB would be difficult.

My suggestion was merely to allocate the event node before locking the event db, then just adding that event to the linked list. None of that allocates.

patina_dxe_core/src/allocator/fixed_size_block_allocator.rs
```diff
@@ -1572,160 +1578,198 @@ mod tests {
...
             let stats = fsb.stats();
             assert_eq!(stats.pool_allocation_calls, 4);
             assert_eq!(stats.pool_free_calls, 2);
             assert_eq!(stats.page_allocation_calls, 3);
             assert_eq!(stats.page_free_calls, 1);
-            assert_eq!(stats.reserved_size, MIN_EXPANSION * 2);
-            assert_eq!(stats.reserved_used, MIN_EXPANSION + uefi_pages_to_size!(5));
-            assert_eq!(stats.claimed_pages, uefi_size_to_pages!(MIN_EXPANSION * 5) + 1 + 0x104);
+            assert_eq!(stats.reserved_size, RESERVED_SIZE);
+            assert_eq!(stats.reserved_size, RESERVED_SIZE);
```
makubacki:
```suggestion
```

patina_dxe_core/src/event_db.rs
```diff
@@ -559,13 +566,36 @@ impl Default for SpinLockedEventDb {
...
     fn lock(&self) -> tpl_lock::TplGuard<'_, EventDb> {
         self.inner.lock()
     }
 
+    fn try_lock(&self) -> Option<tpl_lock::TplGuard<'_, EventDb>> {
+        self.inner.try_lock()
+    }
+
+    fn drain_pending_signals(&self) {
+        let mut pending_signals = self.pending_signals.lock();
```
makubacki: I may have missed this in an earlier conversation somewhere, but the `pending_signals` lock is held until the end of this function (throughout the drain) nesting the signal event lock. Is it possible that `signal_event()` will be called again and takes its else branch to try to acquire the pending signals lock and hit a panic because of re-entrancy?

joschock: Yeah, this is what I was noting above in terms of potential re-entrancy even with this change. I'm ruminating on alternatives and digesting the larger threads above to see if there is a better approach. For this particular case, I could drain the pending queue into a local vec then release the lock, process it, and do that in a loop until drain produces no more events. but that might trade deadlock for infinite loop; so I'm not sure that's better.

joschock: I think for now, I'm going to take this back to draft and do two things:
1) create a unit test that can repro the re-entrancy to better understand the behavior, 
2) find a cleaner way to unravel the re-entrancy.

joschock:
@Javagedes 
> Ah I see. It's an allocation happening inside the EventDb. Did you narrow it down to what is allocating? Can we avoid allocations in general?

I went back to investigate my consistent repro and discovered that the issue is masked by the (I believe unnecessary) `EVENT_DB_INITIALIZED` static. The specific allocation triggering the signal that causes re-entrancy is occurring downstream of `EventDb::create_event()` for runtime events which invokes `runtime::add_runtime_event()`. This causes an expand in the Runtime allocator; which then signals the change event. 

EVENT_DB_INITIALIZED was masking this case by essentially causing the callback to never get signaled in unit tests.  This only masks the obvious reproduction in unit testing; on a real system this case can occur as an edge case after EVENT_DB_INTIALIZED is set if a runtime expansion occurs. 

This branch removes the EVENT_DB_INITIALIZED static (https://github.com/joschock/patina/tree/event_reentrancy_repro) and can repro 100% in the `test_create_event_virtual_address_change()` unit test.

patina_dxe_core/src/allocator/fixed_size_block_allocator.rs
joschock: Removed from this changeset.

patina_dxe_core/src/event_db.rs
joschock: I've revisited the repro and cleaned up the approach. Fundamentally it's the same side-band pending_signals approach, but coded it in a different, less-intrusive way. Switching this back out of draft to solicit further feedback.

joschock:
@Javagedes @makubacki @cfernald @os-d - re-posted this for round 2; please have a look when you have time. 

Repro is easy if you take just the first commit of the change; that removes the static flag that was preventing this from showing up in unit test.

makubacki:
> @Javagedes @makubacki @cfernald @os-d - re-posted this for round 2; please have a look when you have time.
> 
> Repro is easy if you take just the first commit of the change; that removes the static flag that was preventing this from showing up in unit test.

Thanks. If you can please leave it up through Friday, I'll review then.

patina_dxe_core/src/event_db.rs
```diff
@@ -608,7 +667,18 @@ impl SpinLockedEventDb {
...
+        if let Some(mut guard) = self.try_lock() {
+            guard.signal_event(event)
+        } else {
+            //unable to acquire lock; queue signal for later processing.
+            if let Some(mut pending_signals) = self.pending_signals.try_lock() {
+                pending_signals.push(PendingSignals::Event(event));
+                Ok(())
+            } else {
+                log::warn!("Could not acquire pending signals lock to queue signal.");
+                Err(EfiError::DeviceError)
```
os-d: I was going to say that this is UEFI spec defined, so we can't return arbitrary error codes (the efiapi version of this just returns whatever we return here), but the UEFI spec also doesn't believe this function can fail :)

Image

So, the strictly proper spec thing would be to only return success, but I'm okay with "breaking" the spec to return errors. I leave it here as food for thought.

patina_dxe_core/src/event_db.rs
```diff
@@ -608,7 +667,18 @@ impl SpinLockedEventDb {
...
-        self.lock().signal_event(event)
+        if let Some(mut guard) = self.try_lock() {
+            guard.signal_event(event)
+        } else {
+            //unable to acquire lock; queue signal for later processing.
+            if let Some(mut pending_signals) = self.pending_signals.try_lock() {
+                pending_signals.push(PendingSignals::Event(event));
+                Ok(())
+            } else {
+                log::warn!("Could not acquire pending signals lock to queue signal.");
```
os-d: This might be able to be dropped now?

https://github.com/OpenDevicePartnership/patina/blob/9c452663df55b0e11c5cef200f0261274d841311/patina_dxe_core/src/events.rs#L113-L114

This is the one case where we don't acquire a TPL HIGH lock, but also represents an unexpected error condition

patina_dxe_core/src/events.rs
```diff
@@ -310,19 +310,14 @@ extern "efiapi" fn timer_available_callback(event: efi::Event, _context: *mut c_
...
-            | gcd::MapChangeType::FreeMemorySpace
-            | gcd::MapChangeType::RemoveMemorySpace => EVENT_DB.signal_group(efi::EVENT_GROUP_MEMORY_MAP_CHANGE),
-            gcd::MapChangeType::SetMemoryAttributes | gcd::MapChangeType::SetMemoryCapabilities => (),
-        }
+    match map_change_type {
+        gcd::MapChangeType::AddMemorySpace
+        | gcd::MapChangeType::AllocateMemorySpace
+        | gcd::MapChangeType::FreeMemorySpace
+        | gcd::MapChangeType::RemoveMemorySpace => EVENT_DB.signal_group(efi::EVENT_GROUP_MEMORY_MAP_CHANGE),
+        gcd::MapChangeType::SetMemoryAttributes | gcd::MapChangeType::SetMemoryCapabilities => (),
```
os-d, line 314: This isn't part of this, but I will note that edk2 will signal this event on capability change: https://github.com/tianocore/edk2/blob/9ff73832ea0440b86fa0dfa42158fce1ef3b57d2/MdeModulePkg/Core/Dxe/Gcd/Gcd.c#L1015 (that ends up calling CoreAddRange(), which is where this gets signaled). Per UEFI spec, we should call this when the memory map changes (not clear which memory map, but if we assume EFI_MEMORY_MAP, then capabilities are part of that, attributes aren't and so we should signal this for capability changes).

I'm not sure if anyone even listens to this, but anyway, something to consider changing, since we are here. But doesn't need to be part of this.

os-d (approved):

patina_dxe_core/src/event_db.rs
joschock: Well, the expectation is that this can only occur if there is a signal downstream of self.pending_signals.push(). I don't think it will ever happen (rationale to follow) but I wasn't brave enough to make it an assert. I'll change it to an assert. 

Why I don't think it will happen:
The only way this can happen is if a chain like this occurs:  (1) event primitive call that takes the event lock -> (2) event primitive makes an allocation that expands -> (3) expand signals event -> (4) signal_event accquires pending_signals lock -> (5) pending_signals.push() generates an allocation -> (6) allocation expands -> (7) expand signals event -> (8) re-entrance

Since an expand has just happened in (3), it should not also happen at (6), and thus this chain should not occur. Presently that _could_ happen on some boundary cases - the pool only expands by max(MIN_EXPANSION, acutal_allocation) - which means that for some edge cases, it could allocate `actual_allocation` and have zero space left over and thus the next allocation would also expand. My plan (as in this original PR) was to change the allocator logic to always ensure that an expand results in at least MIN_EXPANSION free space in the pool - but I'll make a separate PR for that. 

TBH this is brittle in the sense that it assumes that the only async source of event calls while the event log is held is the allocation subsystem, but that's true today AFAIK.

os-d: Maybe debug_assert so we catch it in development, but just return success? I agree with your general thought process, but I am a bit squeamish on adding straight asserts unless we really can't continue without some condition being true

patina_dxe_core/src/event_db.rs
joschock: I think this was added when EventDb was still spin::Mutex and not TplMutex. I'll remove it.

patina_dxe_core/src/events.rs
joschock: For bisectability reasons, I think it should be done as a separate PR. Filed: https://github.com/OpenDevicePartnership/patina/issues/1113

PR #1069, opened 2025-11-17
PE Image Load: Include image name on failure

vineelko:
## Description

PE Image Load: Include image name on failure

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Validated on Q35

## Integration Instructions

NA


patina_dxe_core/src/image.rs
```diff
@@ -589,13 +589,15 @@ fn core_load_pe_image(
...
+    let pe_file_name = pe_info.filename.as_deref().unwrap_or("Unknown");
+
     // based on the image type, determine the correct allocator and code/data types.
     let (code_type, data_type) = match pe_info.image_type {
         EFI_IMAGE_SUBSYSTEM_EFI_APPLICATION => (efi::LOADER_CODE, efi::LOADER_DATA),
         EFI_IMAGE_SUBSYSTEM_EFI_BOOT_SERVICE_DRIVER => (efi::BOOT_SERVICES_CODE, efi::BOOT_SERVICES_DATA),
         EFI_IMAGE_SUBSYSTEM_EFI_RUNTIME_DRIVER => (efi::RUNTIME_SERVICES_CODE, efi::RUNTIME_SERVICES_DATA),
         unsupported_type => {
-            log::error!("core_load_pe_image_failed: unsupported image type: {unsupported_type:#x?}");
+            log::error!("core_load_pe_image_failed: {pe_file_name} unsupported image type: {unsupported_type:#x?}");
```
Javagedes: Is this the full path or just the binary name? If it is the full path, I suspect we may need to re-format these error messages or it may be kind of hard to view.

os-d (approved):

patina_dxe_core/src/image.rs
vineelko: It is just the file name, not the entire path.

Javagedes (approved):

patina_dxe_core/src/image.rs
makubacki: `core_load_pe_image_failed` is used in 6 places and `core_load_pe_image failed` in 2 places. I prefer `core_load_pe_image failed`, but for searchability in logs, it'd be nice if they're all consistent.

makubacki (approved):

patina_dxe_core/src/image.rs
vineelko: Resolved it.

Issue #765, opened 2025-09-24
[Bug]: memory attribute protocol inconsistent attributes


berlin-with0ut-return:
### Is there an existing issue for this?

- [x] I have searched existing issues

### Current Behavior

```
    match dxe_services::core_get_memory_space_descriptor(base_address) {
        Ok(descriptor) => {
            if descriptor.base_address != base_address || descriptor.length != length {
                log::error!(
                    "{} Inconsistent attributes for: base_address {:#x} length {:#x}",
                    function!(),
                    base_address,
                    length
                );
                return efi::Status::NO_MAPPING;
            }
            unsafe { *attributes = descriptor.attributes & efi::MEMORY_ACCESS_MASK };
            efi::Status::SUCCESS
        }
```

On some Intel boots the above error is logged. The memory attribute protocol should not check attribute consistency. 

### Expected Behavior

No error in logs in regard to memory attributes.

### Steps To Reproduce

PXE boot on Intel platforms. Although this may be reproducible in QEMU as well.

### Build Environment

```markdown
- OS(s): Windows
- Tool Chain(s): VS2022
- Targets Impacted: DEBUG
```

### Version Information

```text
latest (8.1.0)
```

### Urgency

Low

### Are you going to fix this?

Someone else needs to fix it

### Do you need maintainer feedback?

No maintainer feedback needed

os-d:
This didn't get closed when the PR fixing it was merged, closing.

PR #1072, opened 2025-11-17
[REBASE & FF] Memory Management Improvements

### LINKED ISSUE #385: [Feature]: Create common interface for setting attributes across MemorySpaceDescriptors
### Feature Overview

Currently, there are several implementations of using `core_get_memory_space_descriptor` to set attributes accross multiple descriptors. Examples at the type of this feature request:
- `set_page_attributes`
- `clear_memory_attributes`
- `set_memory_attributes`

### Solution Overview

Create a common method for doing this to avoid redundant code

### Alternatives Considered

_No response_

### Urgency

Low

No maintainer feedback needed

### Anything else?

_No response_

os-d:
## Description

This PR has a series of improvements for the memory management subsystem. The most substantial is the introduction of the `MemoryProtectionPolicy` struct where all memory protection policy decisions are centralized. Other code will call rules in implemented as associated functions (except in one case that requires an instance of the struct; the GCD owns this sole instance). This provides better auditability and ensuring consistency.

The other commits were various issues discovered when refactoring. All commits have detailed descriptions. I attempted to nicely organize the commits to allow for bisectability, but don't guarantee it.

Closes #385 and #765 .

- [x] Impacts functionality?
- [x] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested booting Q35/SBSA to Windows/Linux. Tested booting physical ARM64 and Intel platforms to Windows.

## Integration Instructions

N/A.

### DIFF
--- a/patina_dxe_core/src/gcd/spin_locked_gcd.rs
+++ b/patina_dxe_core/src/gcd/spin_locked_gcd.rs
@@ -2059,43 +2059,20 @@ impl SpinLockedGcd {
         self.io.lock().init(io_address_bits);
     }
 
-    /// Helper function to perform an operation on each GCD descriptor in the given range.
+    /// Returns an iterator over GCD descriptors in the given range.
     ///
     /// Arguments:
     /// - `base_address`: The starting address of the range.
     /// - `len`: The length of the range.
-    /// - `operation`: The operation to perform on each descriptor. It takes the descriptor, the base address of the current range,
-    ///   the length of the current range, and a context value.
-    /// - `context`: A context value to pass to the operation.
     ///
     /// Returns:
-    /// - `Ok(())` if the operation was successful for all descriptors.
-    /// - `Err(EfiError)` if the operation failed for any descriptor. The operation does not continue if a failure occurs.
-    pub(crate) fn for_each_desc_in_range<F>(
+    /// - An iterator that yields `MemorySpaceDescriptor`s without allocating memory.
+    pub(crate) fn iter(
         &self,
         base_address: usize,
         len: usize,
-        mut operation: F,
-        context: u64,
-    ) -> Result<(), EfiError>
-    where
-        F: FnMut(MemorySpaceDescriptor, usize, usize, u64) -> Result<(), EfiError>,
-    {
-        let mut current_base = base_address as u64;
-        let range_end = (base_address + len) as u64;
-        while current_base < range_end {
-            let descriptor = self.get_memory_descriptor_for_address(current_base as efi::PhysicalAddress)?;
-            let descriptor_end = descriptor.base_address + descriptor.length;
-
-            // it is still legal to split a descriptor and only operate on part of it
-            let next_base = u64::min(descriptor_end, range_end);
-            let current_len = next_base - current_base;
-            operation(descriptor, current_base as usize, current_len as usize, context)?;
-
-            current_base = next_base;
-        }
-
-        Ok(())
+    ) -> impl Iterator<Item = Result<MemorySpaceDescriptor, EfiError>> {
+        DescRangeIterator::new(self, base_address, len)
     }
 
     // Take control of our own destiny and create a page table that the GCD controls
@@ -2412,40 +2389,38 @@ impl SpinLockedGcd {
         // to coalesce the memory blocks before attempting to free them
         self.memory.lock().get_memory_block_allocation_state(base_address, len)?;
 
+        let range = base_address as u64..base_address.checked_add(len).ok_or(EfiError::InvalidParameter)? as u64;
+
         // Set the attributes before freeing the memory space so that the memory blocks are merged together and we
         // can free the range. It is valid to call free pages on memory which has different attributes. If we fail the
         // free, the memory will be unmapped, but still marked allocated in the memory blocks. This is acceptable as it
         // will not be used again, we will return a failure to the caller and they can ignore this memory (which can
         // not be used after the failed free anyway).
-        self.for_each_desc_in_range(
-            base_address,
-            len,
-            |desc, base_address, len, _| {
-                // we call the worker here because we want to ensure we are getting the caching attribute from the
-                // correct descriptor. It is possible the caching attribute is different across descriptors.
-                if let Err(e) = self.set_memory_space_attributes_worker(
-                    base_address,
-                    len,
-                    MemoryProtectionPolicy::apply_free_memory_policy(desc.attributes),
-                    desc.attributes,
-                ) && e != EfiError::NotReady
-                {
-                    // if we failed to set the attributes in the GCD, we want to catch it, but should still try to go
-                    // down and free the memory space. NotReady is ignored here because the memory bucket code will
-                    // call this before paging is initialized.
-                    log::error!(
-                        "Failed to set free memory attributes for {:#x?} of length {:#x?} Status: {:#x?}",
-                        base_address,
-                        len,
-                        e
-                    );
-                    debug_assert!(false);
-                    return Err(e);
-                }
-                Ok(())
-            },
-            0,
-        )?;
+        for desc_result in self.iter(base_address, len) {
+            let desc = desc_result?;
+            let current_range = desc.get_range_overlap_with_desc(&range);
+            // we call the worker here because we want to ensure we are getting the caching attribute from the
+            // correct descriptor. It is possible the caching attribute is different across descriptors.
+            if let Err(e) = self.set_memory_space_attributes_worker(
+                current_range.start as usize,
+                (current_range.end - current_range.start) as usize,
+                MemoryProtectionPolicy::apply_free_memory_policy(desc.attributes),
+                desc.attributes,
+            ) && e != EfiError::NotReady
+            {
+                // if we failed to set the attributes in the GCD, we want to catch it, but should still try to go
+                // down and free the memory space. NotReady is ignored here because the memory bucket code will
+                // call this before paging is initialized.
+                log::error!(
+                    "Failed to set free memory attributes for {:#x?} of length {:#x?} Status: {:#x?}",
+                    current_range.start,
+                    (current_range.end - current_range.start),
+                    e
+                );
+                debug_assert!(false);
+                return Err(e);
+            }
+        }
 
         match self.memory.lock().free_memory_space(base_address, len, transition) {
             Ok(()) => {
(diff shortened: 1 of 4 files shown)


patina_dxe_core/src/gcd/spin_locked_gcd.rs
```diff
@@ -2060,6 +2060,45 @@ impl SpinLockedGcd {
         self.io.lock().init(io_address_bits);
     }
 
+    /// Helper function to perform an operation on each GCD descriptor in the given range.
```
os-d: note: I'm likely adding another commit that changes this to an iterator. The original impetus for this design is gone, so an iterator probably makes more sense.

patina_dxe_core/src/gcd/spin_locked_gcd.rs
```diff
@@ -569,7 +581,43 @@ impl GCD {
...
+        let range_end = (base_address + len) as u64;
+
+        while current_base < range_end {
+            log::trace!(target: "gcd_measure", "search");
+            let idx = memory_blocks.get_closest_idx(&current_base).ok_or(EfiError::NotFound)?;
+            let block = memory_blocks.get_with_idx(idx).ok_or(EfiError::NotFound)?;
+
+            // Check that the block covers the current base
+            if current_base < block.start() as u64
+                || range_end > block.end() as u64 && block.end() as u64 <= current_base
```
makubacki: Can you please add grouping parentheses here?

patina_dxe_core/src/gcd.rs
```diff
@@ -25,10 +28,384 @@ use r_efi::efi;
...
+        // map it WB by default, because 0 is being used as the null page, it may not have gotten cache attributes
+        // populated
+        if let Ok(descriptor) = GCD.get_memory_descriptor_for_address(0)
+            // set_memory_space_attributes will set both the GCD and paging attributes
+            && descriptor.memory_type != GcdMemoryType::NonExistent
+            && let Err(e) = GCD.set_memory_space_attributes(0, UEFI_PAGE_SIZE, efi::MEMORY_WB)
+        {
+            log::error!("Failed to map page 0 for compat mode. Status: {e:#x?}");
+            debug_assert!(false);
+        }
```
makubacki: I personally think this is easier to read over the nested if let's, but feel free to ignore.

```rust
match GCD.get_memory_descriptor_for_address(0) {
    Ok(desc) if desc.memory_type != GcdMemoryType::NonExistent => {
        if let Err(e) = GCD.set_memory_space_attributes(0, UEFI_PAGE_SIZE, efi::MEMORY_WB) {
            log::error!("Failed to map page 0 for compat mode. Status: {e:#x?}");
            debug_assert!(false);
        }
    }
    _ => {}
}
```

patina_dxe_core/src/gcd.rs
```diff
@@ -25,10 +28,384 @@ use r_efi::efi;
...
+    }
+
+    /// Rule: All loaded image stacks must have a guard page that is unmapped. We accomplish that by setting
+    /// the RP attribute on the guard page. We also set XP to align with other unmapped pages.
+    ///
+    /// Arguments
+    /// * `attributes` - The memory attributes for the stack guard page
+    ///
+    /// Use Case: This is called when loading an image to ensure the stack guard page is protected.
+    pub(crate) fn apply_image_stack_guard_policy(attributes: u64) -> u64 {
```
makubacki: This and similar functions nearby seem like good candidates to be `const`.

```suggestion
    pub(crate) const fn apply_image_stack_guard_policy(attributes: u64) -> u64 {
```

patina_dxe_core/src/memory_attributes_protocol.rs
```diff
@@ -87,45 +114,27 @@ extern "efiapi" fn set_memory_attributes(
...
+            // descriptor first and then set the new attributes as the GCD API takes into account all attributes set or unset.
+            let new_attributes = descriptor.attributes | attributes;
+
+            dxe_services::core_set_memory_space_attributes(base_address as u64, length as u64, new_attributes)
+        },
+        attributes,
+    )
+    // only a few status codes are allowed per UEFI spec, so return unsupported
+    // we don't have a reliable mechanism to reset any previously set attributes if an earlier block succeeded
+    // because any tracking mechanism would be require memory allocations which could change the descriptors
```
makubacki:
```suggestion
    // because any tracking mechanism would require memory allocations which could change the descriptors
```

patina_dxe_core/src/memory_attributes_protocol.rs
```diff
@@ -147,45 +156,27 @@ extern "efiapi" fn clear_memory_attributes(
...
+            // descriptor first and then set the new attributes as the GCD API takes into account all attributes set or unset.
+            let new_attributes = descriptor.attributes & !attributes;
+
+            dxe_services::core_set_memory_space_attributes(base_address as u64, length as u64, new_attributes)
+        },
+        attributes,
+    )
+    // only a few status codes are allowed per UEFI spec, so return unsupported
+    // we don't have a reliable mechanism to reset any previously set attributes if an earlier block succeeded
+    // because any tracking mechanism would be require memory allocations which could change the descriptors
```
makubacki:
```suggestion
    // because any tracking mechanism would require memory allocations which could change the descriptors
```

patina_dxe_core/src/gcd.rs
```diff
@@ -25,10 +28,384 @@ use r_efi::efi;
...
 pub use spin_locked_gcd::{AllocateType, MapChangeType, SpinLockedGcd};
 
+/// The MemoryProtectionPolicy struct is the source of truth for Patina's memory protection rules.
+/// All memory protection decisions in Patina are driven by functions in this struct to have one
+/// easily auditable location.
+///
+/// All rules in this struct are associated functions (don't require an instantiation of the struct)
+/// except for the apply_default_allocated_memory_protection_policy function, because this relies on
+/// internal state. The GCD contains a MemoryProtectionPolicy instance to manage this state.
+pub(crate) struct MemoryProtectionPolicy {
```
Javagedes, line 42: This is going to be difficult for me with the move to bring statics into the core, as your activate compatability mode function makes a lot of global static GCD calls.

I feel like its better to continue to keep the `memory_allocation_default_attributes` value inside the GCD then make this struct an enum:

```rust
enum MemoryProtectionPolicy {
    Allocated(u64),
    RescDescHob(GcdMemoryType),
    NxToUc,
    MemoryAttributesTable,
    ImageStackGuard,
    Image,
    MemoryMap,
    /* I'm tired of scrolling down to find the next one in your code */
}
```

Then you just use it as:

```rust
    let attributes = MemoryProtectionPolicy::Allocated(gcd.default).apply(attributes);
    let attributes = MemoryProtectionPolicy::RescDescHob(memory_type).apply(attributes);
    let attributes = MemoryProtectionPolicy::MemoryMap.apply(attributes);
```

And if you keep `activate_compatibility_mode` as a part of `MemoryProtectionPolicy`, then that would be the one static function. e.g.

```rust
    MemoryProtectionPolicy::activate_compatability_mode(&gcd, <other parts>);
```

patina_dxe_core/src/gcd/spin_locked_gcd.rs
```diff
@@ -590,38 +638,13 @@ impl GCD {
         let idx = memory_blocks.get_closest_idx(&(base_address as u64)).ok_or(EfiError::NotFound)?;
 
         match Self::split_state_transition_at_idx(memory_blocks, idx, base_address, len, transition) {
```
joschock:
```suggestion
        Self::split_state_transition_at_idx(memory_blocks, idx, base_address, len, transition).map_err(|e|e.into())
```

patina_dxe_core/src/gcd.rs
os-d: That's why I'm hoping to get this merged before your change to move statics into the core :)

I'm not sure how I feel about the enum. Let's chat offline.

os-d: Discussed offline. Primary concern was to have `activate_compatibility_mode()` take the GCD as a param instead of as self. Changed to that.

patina_dxe_core/src/gcd/spin_locked_gcd.rs
os-d: Changed to iterator and pushing a new commit.

patina_dxe_core/src/gcd/spin_locked_gcd.rs
os-d: This actually ends up having to be the slightly funnier

```rust
Self::split_state_transition_at_idx(memory_blocks, idx, base_address, len, transition)
            .map(|_| ())
            .map_err(|e| e.into())
```

Because `split_state_transition_at_idx` returns a `Result<usize, EfiError>` whereas `free_memory_space()` returns a `Result<(), EfiError>`. I had debated between the match and the map + map_error, but don't have any real preference. I updated to double map.

joschock: ah. I misread. I though the return of split_state_transition_at_idx was a more compatible `Result<(), EfiError>`. I'm agnostic now; the new construction is fine.

os-d:
@Javagedes @joschock @cfernald I have addressed the current feedback and finished the update to move to an iterator. Please continue your review when you have a chance. I sanity checked the latest changes by booting Q35 and SBSA to Windows. Physical platform testing has also finished on Intel and ARM successfully.

patina_dxe_core/src/config_tables/memory_attributes_table.rs
```diff
@@ -166,17 +166,10 @@ pub fn core_install_memory_attributes_table() {
...
-                            // likely extra sections in the memory bins and so should not be used
-                            // Data we will mark as XP only, as likely the caching attributes were changed, which
-                            // dropped the XP attribute, so we need to set it here.
-                            0 if descriptor.r#type == efi::RUNTIME_SERVICES_CODE => mat_allowed_attrs,
-                            0 if descriptor.r#type == efi::RUNTIME_SERVICES_DATA => {
-                                efi::MEMORY_RUNTIME | efi::MEMORY_XP
-                            }
-                            _ => descriptor.attribute & mat_allowed_attrs,
-                        },
+                        attribute: MemoryProtectionPolicy::apply_memory_attributes_table_policy(
```
cfernald, line 169: Should we have a debug_assert to very no RWX memory here? It's always a pain to debug those when they slip through.

patina_dxe_core/src/gcd.rs
```diff
@@ -25,10 +28,401 @@ use r_efi::efi;
...
+    ///   - Sections w/o the write characteristic are marked as Read Only
+    ///
+    /// Arguments
+    /// * `section_base_addr` - The base address of the section being loaded
+    /// * `section_characteristics` - The PE/COFF section characteristics
+    ///
+    /// Returns a tuple of (attributes, capabilities) to be applied to the section
+    ///
+    /// Use Case: This is called when loading an image to ensure each section has the proper memory protections.
+    #[inline(always)]
```
cfernald: I think we can drop a lot of these inline directives. In the end the compiler can make the choice of when it is optimal to inline vs not. The only cases I think we should force the matter is when we think we will get potential code removal benefits by forcing a constant replacement.

patina_dxe_core/src/gcd.rs
```diff
@@ -25,10 +28,401 @@ use r_efi::efi;
...
+    ///
+    /// Arguments
+    /// - * `capabilities` - The existing capabilities for the memory region
+    ///
+    /// Returns the updated capabilities and the attributes to set
+    ///
+    /// Use Case: This is called whenever new memory is added to the GCD
+    #[inline(always)]
+    pub(crate) const fn apply_add_memory_policy(capabilities: u64) -> (u64, u64) {
+        (capabilities | efi::MEMORY_ACCESS_MASK | efi::MEMORY_RUNTIME, efi::MEMORY_RP | efi::MEMORY_XP)
```
cfernald, line 215: Why runtime in this case?

cfernald (approved):

patina_dxe_core/src/gcd.rs
os-d: Ironically, I didn't have these, was asked to make these consts, had seen recent conversation about making const funcs inlined, so thought I'd get ahead of the next comment :). I'll remove them.

patina_dxe_core/src/gcd.rs
os-d: That's the capabilities field, all memory is "capable" of being runtime. Basically we are trying to say all memory is capable of being runtime, RO, XP, and RP.

patina_dxe_core/src/config_tables/memory_attributes_table.rs
os-d: Aren't we guaranteed to not have RWX by this?

Image

if we come in with 0 (i.e. RWX) it will set one or more of RO + XP.

Or are you saying have a debug_assert for the input coming in as RWX?

PR #1074, opened 2025-11-18
Always enable GICv4 support.

joschock:
## Description

This PR always enables GICv4 support in the arm-gic crate. GicV4 support is backwards compatible with GicV3, and is necessary for properly calculating the GIC Redistributor register offsets. 

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Booted on hardware with GicV4, observed that GIC redistributor accesses are no longer incorrect.

## Integration Instructions

N/A - changes are backward compat with GicV3, though the expectation is that most systems are GicV4.

### DIFF
--- a/core/patina_internal_cpu/src/interrupts/aarch64/gic_manager.rs
+++ b/core/patina_internal_cpu/src/interrupts/aarch64/gic_manager.rs
@@ -78,7 +78,8 @@ pub struct AArch64InterruptInitializer {
 }
 
 impl AArch64InterruptInitializer {
-    /// Create AArch64InterruptInitializer from register bases and initialize GICv3 hardware for use by the current cpu.
+    /// Create AArch64InterruptInitializer from register bases and initialize GICv3/4 hardware for use by the current
+    /// cpu.
     ///
     /// * Enable affinity routing and non-secure group 1 interrupts.
     /// * Enable gic cpu interface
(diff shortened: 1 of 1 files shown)


apop5:
@joschock 

Have we documented that patina is only supporting V3/V4 GIC?

joschock:
> @joschock
> 
> Have we documented that patina is only supporting V3/V4 GIC?

AFAIK, we don't have documentation on the aarch64 hardware support in general or on the GIC specifically.

os-d:
> > @joschock
> > Have we documented that patina is only supporting V3/V4 GIC?
> 
> AFAIK, we don't have documentation on the aarch64 hardware support in general or on the GIC specifically.

No time like the present to start it :). This is something that could go here: https://github.com/OpenDevicePartnership/patina/blob/main/docs/src/integrate/patina_dxe_core_requirements.md. Perhaps under an architecture requirements section or something.

joschock:
> > > @joschock
> > > Have we documented that patina is only supporting V3/V4 GIC?
> > 
> > 
> > AFAIK, we don't have documentation on the aarch64 hardware support in general or on the GIC specifically.
> 
> No time like the present to start it :). This is something that could go here: https://github.com/OpenDevicePartnership/patina/blob/main/docs/src/integrate/patina_dxe_core_requirements.md. Perhaps under an architecture requirements section or something.

I can add an architectural requirement section with a bullet that says "we support GIC v3/v4 on AARCH64" - but it would be lonely by itself :). if we're going to do that, maybe it needs to be a bit more wholistic. For example, I think we were assuming invariant TSC on x86 (which isn't well supported by QEMU, for instance). Things like NX and PML4 are also assumed (and rightly so, given that they've been around for more than two decades). How do we calibrate the dial around what rises to a level of "you need to know this at the outset from docs" vs assumed hardware support?

makubacki:
> > > > @joschock
> > > > Have we documented that patina is only supporting V3/V4 GIC?
> > > 
> > > 
> > > AFAIK, we don't have documentation on the aarch64 hardware support in general or on the GIC specifically.
> > 
> > 
> > No time like the present to start it :). This is something that could go here: https://github.com/OpenDevicePartnership/patina/blob/main/docs/src/integrate/patina_dxe_core_requirements.md. Perhaps under an architecture requirements section or something.
> 
> I can add an architectural requirement section with a bullet that says "we support GIC v3/v4 on AARCH64" - but it would be lonely by itself :). if we're going to do that, maybe it needs to be a bit more wholistic. For example, I think we were assuming invariant TSC on x86 (which isn't well supported by QEMU, for instance). Things like NX and PML4 are also assumed (and rightly so, given that they've been around for more than two decades). How do we calibrate the dial around what rises to a level of "you need to know this at the outset from docs" vs assumed hardware support?

To help make immediate progress, I filed an issue to track independent work for each architecture and assigned the AARCH64 issue to @joschock. Feel free to add other assignees and split the work as needed.

https://github.com/OpenDevicePartnership/patina/issues/1075

PR #1084, opened 2025-11-19
Additional zerocopy_derive imports

makubacki:
## Description

Adds more imports to those in e9bcdea.

We opted to depend on `zerocopy` and `zerocopy-derive` directly instead of using the re-export of the derive macros in the `zerocopy` crate using the `derives` feature.

In this case, the documentation states:

> derive Provides derives for the core marker traits via the
> zerocopy-derive crate. These derives are re-exported from
> zerocopy, so it is not necessary to depend on zerocopy-derive
> directly.
>
> However, you may experience better compile times if you instead
> directly depend on both zerocopy and zerocopy-derive in your
> Cargo.toml, since doing so will allow Rust to compile these crates
> in parallel. To do so, do not enable the derive feature, and list
> both dependencies in your Cargo.toml with the same leading non-zero
> version number; e.g:
>
> [dependencies]
> zerocopy = "0.X"
> zerocopy-derive = "0.X"
>
> To avoid the risk of duplicate import errors if one of your
> dependencies enables zerocopy’s derive feature, import derives as
> use zerocopy_derive::* rather than by name
> (e.g., use zerocopy_derive::FromBytes).

From: https://docs.rs/zerocopy/latest/zerocopy/#cargo-features

This follows the advice there to import derives from `zerocopy_derive` using a wildcard.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- `cargo make all`
- `cargo make build` on each crate

## Integration Instructions

- N/A

### DIFF
--- a/components/patina_mm/src/config.rs
+++ b/components/patina_mm/src/config.rs
@@ -24,7 +24,7 @@ use core::{fmt, pin::Pin, ptr::NonNull};
 
 use patina::{Guid, base::UEFI_PAGE_MASK};
 use r_efi::efi;
-use zerocopy::{FromBytes, Immutable, IntoBytes, KnownLayout};
+use zerocopy_derive::*;
 
 /// MM Communication Buffer Status
 ///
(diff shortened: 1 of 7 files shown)


makubacki:
The `patina_performance` crate's enabling of "derive" allowed the workspace build to succeed. I'll push an update for the mdbook soon and then publish the PR.

PR #1087, opened 2025-11-19
patina_debugger: Switch breakpoint to be conditional

cfernald:
## Description

The breakpoint always causing an exception may not be well understood and could lead to people leaving breakpoints in production code causing unexpected exceptions. This commit changes the existing `breakpoint` function to only trigger a breakpoint if the debugger is enabled, and renames the existing `breakpoint` function to `breakpoint_unchecked` to indicate that it will always trigger a breakpoint regardless of the debugger state.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Unit tests

## Integration Instructions

N/A

### DIFF
--- a/core/patina_internal_cpu/src/cpu/aarch64/cpu.rs
+++ b/core/patina_internal_cpu/src/cpu/aarch64/cpu.rs
@@ -86,19 +86,18 @@ impl EfiCpuAarch64 {
     }
 
     fn data_cache_line_len(&self) -> u64 {
-        #[cfg(all(not(test), target_arch = "aarch64"))]
-        {
-            let ctr_el0 = unsafe {
-                let ctr_el0: u64;
-                asm!("mrs {}, ctr_el0", out(reg) ctr_el0);
-                ctr_el0
-            };
-            return 4 << ((ctr_el0 >> 16) & 0xf);
-        }
-        #[cfg(not(target_arch = "aarch64"))]
-        {
-            // For all other cases, return 64 bytes
-            64_u64
+        cfg_if::cfg_if! {
+            if #[cfg(all(not(test), target_arch = "aarch64"))]  {
+                let ctr_el0 = unsafe {
+                    let ctr_el0: u64;
+                    asm!("mrs {}, ctr_el0", out(reg) ctr_el0);
+                    ctr_el0
+                };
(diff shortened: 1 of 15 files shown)


core/patina_debugger/src/lib.rs
```diff
@@ -198,12 +200,28 @@ pub fn initialize(interrupt_manager: &mut dyn InterruptManager) {
...
+
+/// Invokes a debug break instruction unconditionally. If this routine is invoked when
+/// the debugger is not enabled, it will cause an unhandled exception.
+///
+/// ## Important
+///
+/// This should only be used in debug scenarios or when it is impossible to continue
+/// execution in the current state and an CPU exception must be raised.
+#[inline(always)]
+pub fn breakpoint_unchecked() {
```
os-d, line 221: Should we not make this pub? i.e. if you are in a scenario where you want to fire a brk without the debugger enabled, I think that maybe we should just have someone do that explicitly in their code. Easy to do and then the debugger doesn't provide any gun to shoot yourself in the foot

cfernald: Well sometimes you really do need a new hole in your foot. The way I originally envisioned this function was simply the debugger providing a convenient wrapper from the architecture breakpoint instruction. Effectively the `__debug_breakpoint()`/`DbgBreakPoint()` intrinsic. It can be useful in debug scenarios to pepper these in, even if there is no debugger to capture state through an exception. I'm open to closing it, but I don't think we need to lock up such tools so long as they are explicit that they we for debug scenarios.

os-d: Agreed on the necessity of it, I just don't think it needs to come from the debugger. If anything, it could come from the SDK's arch specific stuff. I'm not opposed to this being here, but to me this would make more sense not in the debugger. But not a huge preference. If you'd rather keep this here that's fine, too.

cfernald: I could see the SDK being more appropriate though I'm not sure I'd want to do that now just for this, I would rather keep it located with `breakpoint` for consistency for now.

Javagedes (approved):

PR #1093, opened 2025-11-20
Patina Dxe Core: Add a default environment based test logger

vineelko:
## Description

Since Patina relies on the `log::*` API for logging, not configuring a default
logger can cause `log::*` calls to appear completely uncovered in llvm-cov
reports, slightly reducing overall coverage. Adding a simple environment based
logger avoids this issue. Although this change enables coverage for `log::*`,
there are still cases that cannot be covered. For example, in the function
below, there is no way to get coverage for lines 6 and 7.

```
1   1   fn print() {
2   1       let mut i = 0;
3   11      for _ in 0..10 {
4   10          log::info!(
5   10              "This will be covered {} {}",
6                   i,
7                   "This and the above line will not be covered."
8               );
9   10          i = i + 1;
10          }
11  1   }
```

The environment logger can be enabled as shown below:
To enable logging, set the `RUST_LOG` environment variable to the desired
log level (e.g., `debug`, `info`, `warn`, `error`) before running the tests.

```bash
$ RUST_LOG=debug cargo test -p patina_dxe_core allocator::usage_tests::uefi_memory_map -- --nocapture
or
PS:> $env:RUST_LOG="debug"; cargo test -p patina_dxe_core allocator::usage_tests::uefi_memory_map -- --nocapture
```

For `cargo make coverage` a corresponding `Makefile.toml` change will be added
from patina-devops
```
[tasks.test]
description = "Run tests and collect coverage data without generating reports."
env = { "RUST_LOG" = "info" }  <---
...
```

## How This Was Tested

`cargo make all`

## Integration Instructions

NA

### DIFF
--- a/patina_dxe_core/src/allocator/usage_tests/uefi_memory_map.rs
+++ b/patina_dxe_core/src/allocator/usage_tests/uefi_memory_map.rs
@@ -159,24 +159,6 @@ mod tests {
     use serial_test::serial;
     use std::panic::RefUnwindSafe;
 
-    // Logger initialization
-    fn init_logger() {
-        use std::sync::OnceLock;
-        static INIT: OnceLock<()> = OnceLock::new();
-
-        INIT.get_or_init(|| {
-            // Default to no logging unless RUST_LOG environment variable is set
-            let mut builder = env_logger::Builder::from_default_env();
-
-            // If RUST_LOG is not set, default to Off (no logging)
-            if std::env::var("RUST_LOG").is_err() {
-                builder.filter_level(log::LevelFilter::Off);
-            }
-
-            builder.init();
-        });
-    }
-
     /// HOB configuration structures to simplify building test scenarios.
     ///
     /// These simplified configuration structs provide a convenient test API without requiring
@@ -324,6 +306,7 @@ mod tests {
         /// Execute the test scenario and validate results
         pub fn run_test(&self) {
             test_support::with_global_lock(|| {
+                test_support::init_test_logger();
                 let hob_list_ptr = self.build_custom_hob_list();
 
                 // SAFETY: This is a test environment where the initialization order is controlled per
@@ -673,7 +656,6 @@ mod tests {
     #[test]
     #[serial]
     fn test_simple_memory_scenario() {
-        init_logger();
         let scenario = MemoryMapTestScenario::new("Simple Memory Test", SIZE_64MB as u64)
             .with_resource_descriptor(ResourceDescriptorConfig {
                 resource_type: hob::EFI_RESOURCE_SYSTEM_MEMORY,
@@ -695,7 +677,6 @@ mod tests {
     #[test]
     #[serial]
     fn test_mmio_and_memory_scenario() {
-        init_logger();
         let scenario = MemoryMapTestScenario::new("MMIO and Memory Test", SIZE_128MB as u64) // 128MB
             .with_resource_descriptor(ResourceDescriptorConfig {
                 resource_type: hob::EFI_RESOURCE_SYSTEM_MEMORY,
@@ -741,7 +722,6 @@ mod tests {
     #[test]
     #[serial]
     fn test_runtime_services_memory() {
-        init_logger();
         let scenario = MemoryMapTestScenario::new("Runtime Services Test", SIZE_256MB as u64)
             .with_resource_descriptor(ResourceDescriptorConfig {
                 resource_type: hob::EFI_RESOURCE_SYSTEM_MEMORY,
@@ -781,7 +761,6 @@ mod tests {
     #[test]
     #[serial]
     fn test_complex_real_world_hob_scenario() {
-        init_logger();
         let scenario = create_complex_hob_scenario().with_validation(|descriptors| {
             MemoryMapValidation::new()
                 .expect_min_descriptors(5)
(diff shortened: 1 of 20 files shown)


os-d (approved):
By default this won't log anything, right? Only if you pass --no-capture will the log have it? I'm worried about cluttering the log and slowing down tests if we do actual logging.

patina_dxe_core/src/allocator/usage_tests/uefi_memory_map.rs
```diff
@@ -159,24 +140,6 @@ mod tests {
     use serial_test::serial;
     use std::panic::RefUnwindSafe;
```
makubacki, line 161: Right now, `test_support::init_test_logger()` is:

```rust
pub(crate) fn init_test_logger() {
    static LOGGER: ConsoleLogger = ConsoleLogger;
    let _ = log::set_logger(&LOGGER);
    log::set_max_level(log::LevelFilter::Info);
}
```

Can the `OnceLock` pattern be taken there?

patina_dxe_core/src/allocator/usage_tests/uefi_memory_map.rs
```diff
@@ -16,25 +16,6 @@
 //! memory map in different scenarios. This can also be useful to quickly see the impact of
 //! changes in patina code to the UEFI memory map.
 //!
-//! ## Logging
```
makubacki, line 19: I'd like to keep the `env_logger` support as documented here.

makubacki (changes requested):

patina_dxe_core/src/allocator/usage_tests/uefi_memory_map.rs
makubacki: Marked as changes requested for this request.

Javagedes:
Are you sure we actually need a real logger here? I think we just need to make sure the `level` conditions are correct so that the `if` statement in the expanded macro does not bail out. I kind of think that just involves setting the max_level() correctly maybe? This is what the expanded log macros come out to:

```rust
{
    let lvl = ::log::Level::Error;
    if lvl <= ::log::STATIC_MAX_LEVEL
        && lvl <= ::log::max_level()
    {
        ::log::__private_api::log(
            { ::log::__private_api::GlobalLogger },
            format_args!(
                "Error freeing previous MAT pointer: {0:#X?}",
                err,
            ),
            lvl,
            &(
                "patina_dxe_core::config_tables::memory_attributes_table",
                "patina_dxe_core::config_tables::memory_attributes_table",
                ::log::__private_api::loc(),
            ),
            (),
        );
    }
}
```

Overall, I'm not against having a real logger, but realistically it should be unnecessary with the debugger on a host based system.

vineelko:
> By default this won't log anything, right? Only if you pass --no-capture will the log have it? I'm worried about cluttering the log and slowing down tests if we do actual logging.

That's correct. No logs without `--nocapture`

patina_dxe_core/src/allocator/usage_tests/uefi_memory_map.rs
vineelko: Sure will fix that.

patina_dxe_core/src/allocator/usage_tests/uefi_memory_map.rs
vineelko: Ahh I see the difference now, it is when setting the `RUST_LOG=debug` is what makes the difference otherwise at info level both envlogger and console log results in similar output. 

But the bigger problem here is, `patina_dxe_core` crate gets compiled into a single binary(patina_dxe_core-0fa2c65cbd247021.exe), So we cannot have two loggers in the same process(uefi_memory_map logger init will panic if console logger is already enabled). 

So I think we can do something like below, 
- In `uefi_memory_map` module, inside `test_support::with_global_lock` we can set the log level based on RUST_LOG appropriately and restore it to info level towards the end of the closure. 

Any thoughts? I am pushing a fixed version of this a separate commit.

vineelko:
> Are you sure we actually need a real logger here? I think we just need to make sure the `level` conditions are correct so that the `if` statement in the expanded macro does not bail out. I kind of think that just involves setting the max_level() correctly maybe? This is what the expanded log macros come out to:
> 
> ```rust
> {
>     let lvl = ::log::Level::Error;
>     if lvl <= ::log::STATIC_MAX_LEVEL
>         && lvl <= ::log::max_level()
>     {
>         ::log::__private_api::log(
>             { ::log::__private_api::GlobalLogger },
>             format_args!(
>                 "Error freeing previous MAT pointer: {0:#X?}",
>                 err,
>             ),
>             lvl,
>             &(
>                 "patina_dxe_core::config_tables::memory_attributes_table",
>                 "patina_dxe_core::config_tables::memory_attributes_table",
>                 ::log::__private_api::loc(),
>             ),
>             (),
>         );
>     }
> }
> ```
> 
> Overall, I'm not against having a real logger, but realistically it should be unnecessary with the debugger on a host based system.

I still think there is value in have a real logger because it helps diagnose individual tests using `--nocapture` and especially since during normal test execution this does not make any different wrt to the output. For more granular control on the log level may be we can switch the real logger to be based on env_logger. That way we can set the var appropriately along with `--nocapture`

patina_dxe_core/src/allocator/usage_tests/uefi_memory_map.rs
vineelko: May be let me try and switch everything to `env_logger` that would simplify a bunch of things. Please hold off on the review of this PR for today.

vineelko: Updated the PR to completely switch to `env_logger`.

patina_dxe_core/src/allocator/usage_tests/uefi_memory_map.rs
vineelko: Resolved.

makubacki (approved):

Issue #996, opened 2025-10-31
[Documentation]: Add DXE execution performance delta for C DXE Core vs Patina on X64 and AARCH64 platform


makubacki:
### Request Description

To provide a rough estimate of performance delta in the DXE execution timeframe, update the Patina document to include the C and Patina DXE Core performance delta on a X64 and AARCH64 platform.

Note: Only a difference needs to be given. Platform details, DXE start time, and DXE execution time itself do not need to be included. The focus of this data is the *difference* in performance as a reference point, not those other details.

### Are you going to make the change?

Someone else needs to make the change

### Do you need maintainer feedback?

No maintainer feedback needed

PR #1103, opened 2025-11-20
Minor documentation update on current perf delta

berlin-with0ut-return:
## Description

As per https://github.com/OpenDevicePartnership/patina/issues/996#event-20641579881 add a note on current Patina performance on physical (Intel) platform.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [x] Includes documentation?

## How This Was Tested

N/A. 

## Integration Instructions

N/A.


berlin-with0ut-return:
@makubacki could you clarify what additional information (if any) you would like in this documentation?

docs/src/patina.md
```diff
@@ -300,3 +300,10 @@ dereference detection and stack guard are active so code (C DXE driver or a thir
 protection violations caught at runtime.
 
 For more details about mememory management in Patina see [Memory Management](./dxe_core/memory_management.md).
+
+#### Performance
+
+The performance numbers below are provided as of Patina v16, using the C implementation as a reference for
+DXE core execution times.
```
makubacki:
```suggestion
The performance numbers below are provided as of Patina v16. On these physical platform(s),
the only difference is the edk2 C DXE core versus the Patina Rust DXE Core. This is provided
solely as a rough reference for Patina DXE core performance on a physical system.
```

docs/src/patina.md
```diff
@@ -300,3 +300,10 @@ dereference detection and stack guard are active so code (C DXE driver or a thir
 protection violations caught at runtime.
 
 For more details about mememory management in Patina see [Memory Management](./dxe_core/memory_management.md).
+
+#### Performance
+
+The performance numbers below are provided as of Patina v16, using the C implementation as a reference for
+DXE core execution times.
+
+**X64 (Intel)**: +9.1%
```
makubacki: Can you please clarify in the document exactly what this percentage represents?

makubacki:
> @makubacki could you clarify what additional information (if any) you would like in this documentation?

Left a suggestion and request for clarification. Once those are addressed, I think you can publish the PR if you're ready.

docs/src/patina.md
makubacki: Was this intentionally not included?

```suggestion
The performance numbers below are provided as of Patina v16. On these physical platform(s),
the only difference is the edk2 C DXE core versus the Patina Rust DXE Core. This is provided
solely as a rough reference for Patina DXE core performance on a physical system.
```

makubacki (approved):

docs/src/patina.md
berlin-with0ut-return: no, i meant to include it but missed it

docs/src/patina.md
```diff
@@ -300,3 +300,11 @@ dereference detection and stack guard are active so code (C DXE driver or a thir
...
 
 For more details about mememory management in Patina see [Memory Management](./dxe_core/memory_management.md).
+
+#### Performance
+
+The performance numbers below are provided as of Patina v16. On these physical platform(s),
+the only difference is the edk2 C DXE core versus the Patina Rust DXE Core. This is provided
+solely as a rough reference for Patina DXE core performance on a physical system.
+
+**X64 (Intel)**: +9.1% (DXE core execution time delta - % increase from C)
```
spbrogan: sorry.  I am not sure i follow what DXE core execution time represents.  
In patina our dxe core  has more than just DxeCore in edk2.  How is that accounted for?  Are the calls into the dxe core performance measured.  IE - Does locate protocol take longer or not?  How is the entire boot time impacted?  

My takeaway is that Patina core is 10% slower but I have no idea if that is 100ms or 100us or 1 second?  I understand the idea behind only showing delta % but it needs to be clear what that means.

berlin-with0ut-return: i am measuring from the entrypoint of DxeIpl to the entrypoint of BdsDxe as a rough approximation for DXE execution time. it's not perfect but should give a rough idea of when DXE starts/ends, and using the same entrypoint perf measurement keeps things consistent across firmware builds.

the difference is within the ms range. i can add a concrete delta which would be ~80ms slower for the Rust DXE core when measured as described above

spbrogan: In the benchmarking work stream do we have any specific data we could share?  That might be more constrained.  A single measurement here is hard to extract much from. 

Given this is the Patina overview page it would make more sense to make some generic statements about performance and then back that up with specific details like what you have added here.  For example:

#### Performance

The Patina Dxe Core is authored in Rust which generally produces larger code than C.  <is there some generic well accepted statement for why and how much larger/slower on average rust code than c code.  If not here is mine.>  This is attributed to the extra safety checks and capabilities that the language builds in.  That said, general consensus is that Rust is the most performant of the memory safe languages and adds the least overhead.  Rust also has significantly more language features and types (HashMap, Vectors, iterators, etc) and Patina uses these to provide more robust implementations within the core.  Patina also enables numerous CPU capabilities differently than most C firmware.  All this to say, raw wall clock performance is not a fair comparison, but it is the most universal and relevant comparison when discussing PC firmware.  As of Patina v16.0 running on an Intel x64 system we are seeing a ~10% slower execution of the DXE phase (100s of ms).

PR #1104, opened 2025-11-21
[patina_dxe_core] Safety comments for lib.rs

Raymond-MS:
## Description

Updated lib.rs to include safety comments around unsafe blocks and added checks where necessary. Added test code for the updated function.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Built and ran with QEMU SBSA platform. Booted to UEFI shell successfully.

## Integration Instructions

N/A

### DIFF
--- a/patina_dxe_core/src/lib.rs
+++ b/patina_dxe_core/src/lib.rs
@@ -585,25 +585,35 @@ fn call_bds() -> ! {
     // Enable status code capability in Firmware Performance DXE.
     match protocols::PROTOCOL_DB.locate_protocol(status_code::PROTOCOL_GUID) {
         Ok(status_code_ptr) => {
-            let status_code_protocol = unsafe { (status_code_ptr as *mut status_code::Protocol).as_mut() }.unwrap();
-            (status_code_protocol.report_status_code)(
-                EFI_PROGRESS_CODE,
-                EFI_SOFTWARE_DXE_CORE | EFI_SW_DXE_CORE_PC_HANDOFF_TO_NEXT,
-                0,
-                &patina::guids::DXE_CORE,
-                ptr::null(),
-            );
+            if let Some(status_code_protocol_ptr) = NonNull::new(status_code_ptr) {
+                // SAFETY: Some(status_code_protocol_ptr) guarantees that the pointer is non-NULL
+                let status_code_protocol = unsafe { status_code_protocol_ptr.cast::<status_code::Protocol>().as_ref() };
+                (status_code_protocol.report_status_code)(
+                    EFI_PROGRESS_CODE,
+                    EFI_SOFTWARE_DXE_CORE | EFI_SW_DXE_CORE_PC_HANDOFF_TO_NEXT,
+                    0,
+                    &patina::guids::DXE_CORE,
+                    ptr::null(),
+                );
+            } else {
+                log::error!("status_code protocol pointer is NULL")
+            }
         }
         Err(err) => log::error!("Unable to locate status code runtime protocol: {err:?}"),
-    };
+    }
 
     match protocols::PROTOCOL_DB.locate_protocol(bds::PROTOCOL_GUID) {
-        Ok(bds) => {
-            let bds = bds as *mut bds::Protocol;
-            // SAFETY: The BDS arch protocol is the valid C structure as defined by the UEFI specification. The entry
-            // field of the protocol is a valid function pointer that conforms to the expected calling convention.
-            unsafe {
-                ((*bds).entry)(bds);
+        Ok(bds_ptr) => {
+            if let Some(bds_protocol_ptr) = NonNull::new(bds_ptr) {
+                let bds_protocol_ptr = bds_protocol_ptr.cast::<bds::Protocol>();
+                // SAFETY: The BDS arch protocol is the valid C structure as defined by the UEFI specification. The entry
+                // field of the protocol is a valid function pointer that conforms to the expected calling convention.
+                // Some(bds_protocol_ptr) guarantees that the pointer is non-NULL
+                unsafe {
+                    (bds_protocol_ptr.as_ref().entry)(bds_protocol_ptr.as_ptr());
+                }
+            } else {
+                log::error!("bds protocol pointer is NULL")
             }
         }
         Err(err) => log::error!("Unable to locate BDS arch protocol: {err:?}"),
@@ -616,6 +626,7 @@ fn call_bds() -> ! {
 #[coverage(off)]
 mod tests {
     use super::*;
+    use core::sync::atomic::AtomicBool;
 
     #[test]
     fn test_cannot_set_instance_twice() {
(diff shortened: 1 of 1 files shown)


patina_dxe_core/src/lib.rs
```diff
@@ -541,10 +541,15 @@ impl<P: PlatformInfo> Core<P> {
     /// Clippy gets unhappy if we call get_c_hob_list_size directly, because it gets confused, thinking
     /// get_c_hob_list_size is not marked unsafe, but it is
     fn get_hob_list_len(hob_list: *const c_void) -> usize {
+        // SAFETY: hob_list is checked for NULL before the call to get_hob_list_len
         unsafe { get_c_hob_list_size(hob_list) }
     }
 
     fn initialize_system_table(&self, physical_hob_list: *const c_void) -> Result<()> {
+        if physical_hob_list.is_null() {
+            return Err(InvalidParameter);
```
os-d: We typically use the full EfiError::InvalidParameter for clarity

patina_dxe_core/src/lib.rs
```diff
@@ -541,10 +541,15 @@ impl<P: PlatformInfo> Core<P> {
     /// Clippy gets unhappy if we call get_c_hob_list_size directly, because it gets confused, thinking
     /// get_c_hob_list_size is not marked unsafe, but it is
     fn get_hob_list_len(hob_list: *const c_void) -> usize {
+        // SAFETY: hob_list is checked for NULL before the call to get_hob_list_len
```
os-d: While here, you could try removing this wrapper function and call the unsafe func directly and see if clippy is still confused

patina_dxe_core/src/lib.rs
```diff
@@ -686,14 +691,17 @@ fn call_bds() -> ! {
...
-            (status_code_protocol.report_status_code)(
-                EFI_PROGRESS_CODE,
-                EFI_SOFTWARE_DXE_CORE | EFI_SW_DXE_CORE_PC_HANDOFF_TO_NEXT,
-                0,
-                &patina::guids::DXE_CORE,
-                ptr::null(),
-            );
+            if !status_code_ptr.is_null() {
+                // SAFETY: status_code_ptr is checked for NULL before it is used
+                let status_code_protocol = unsafe { (status_code_ptr as *mut status_code::Protocol).as_mut() }.unwrap();
```
os-d: This actually violates our statement that patina dxe core doesn’t use any protocols. @makubacki is there a better place for this or we ignore for now until a status code infra exists?

os-d: That being said, I don’t think you need the null check here, the Ok branch is guaranting a good value passed back. The safety comment can say something like locate_protocol guarantees this is a valid pointer to the protocol.

patina_dxe_core/src/lib.rs
os-d: But that being said, I don’t think we need to null check here, we already have in init_memory. In fact, start_dispatcher doesn’t need the physical hob list passed in at all, it should use the relocated hob list, i.e. it should call self.hob_list() to get the Rust version of the hobs. Can you make that change?

patina_dxe_core/src/lib.rs
makubacki: This is fine for now and the status code needs to be reported to listeners for compatibility.

The `Ok` branch is one thing, but `is_null()` is also redundant because `as_mut()` returns an `Option` based on a null pointer check.

I agree that you rely on the `locate_protocol()` return value as well and then either keep `as_mut()` if you'd like, but given the trust in `locate_protocol()`, you could just do `let status_code_protocol = unsafe { &mut *(status_code_ptr as *mut status_code::Protocol) };`.

patina_dxe_core/src/lib.rs
makubacki: Locally, I dropped `get_hob_list_len()` entirely and clippy passed as configured for `patina_dxe_core` right now.

patina_dxe_core/src/lib.rs
joschock: It's entirely possible for locate protocol to return a NULL interface. I think if you did the following in a driver:
```
gBS->InstallProtocolInterface(&Handle, &WhateverStatusCodeGuid, EFI_NATIVE_INTERFACE, NULL);
```
then above could return `Ok(null)` for the locate protocol.

joschock:
```rust
Ok(status_code_ptr) => {
   if let Some(protocol) = unsafe { (status_code_ptr as *mut status_code::Protocol).as_mut() {
   // do stuff
   }
```

this doesn't guard against the even worse example:
```
gBS->InstallProtocolInterface(&handle, &WhateverStatusCodeGuid, EFI_NATIVE_INTERFACE (VOID *)0xDEADBEEF);
```

but I'm not sure what you can do about that.

makubacki: The earlier comment was in response to Oliver's suggestion which is dependent on your level of trust in the protocol GUID matching a valid protocol interface structure.

To be clear, you should check for null before attempting to dereference from the protocol. Just do it one time.

patina_dxe_core/src/lib.rs
makubacki: I made a separate PR (https://github.com/OpenDevicePartnership/patina/pull/1120) related to this since it started to leave the scope of the safety comments @Raymond-MS was asked to add in this PR.

The system table cannot directly install the Rust HOBs because the HOB list it provides needs to be binary compatible with the PI-spec defined list. It is a fine solution for managing HOBs more idiomatically in Rust, but some code will still need a HOB list formatted in normal binary representation not a vector of references to HOB structures.

With the consolidation of state into the `Core` type, I think there is opportunity to simplify stack management of the HOB list pointer and draw a better distinction between these two types. That's what I did in that PR.

makubacki: Closing this thread since that PR was meant to address this conversation and it is now merged.

makubacki (approved):

patina_dxe_core/src/lib.rs
makubacki: Resolving this thread since the null check is in the latest code.

makubacki:
@Raymond-MS, the codecov patch check failure is related to your new code. Can you please try adding some tests to cover the code?

image

patina_dxe_core/src/lib.rs
```diff
@@ -585,25 +585,31 @@ fn call_bds() -> ! {
...
-            (status_code_protocol.report_status_code)(
-                EFI_PROGRESS_CODE,
-                EFI_SOFTWARE_DXE_CORE | EFI_SW_DXE_CORE_PC_HANDOFF_TO_NEXT,
-                0,
-                &patina::guids::DXE_CORE,
-                ptr::null(),
-            );
+            if !status_code_ptr.is_null() {
+                // SAFETY: status_code_ptr is checked for NULL before it is used
+                let status_code_protocol = unsafe { &mut *(status_code_ptr as *mut status_code::Protocol) };
```
Javagedes: I would use `NonNull` here. It is an easy way to null check and get a reference to the underlying protocol.

https://doc.rust-lang.org/std/ptr/struct.NonNull.html

```rust
// NonNull::new(..) returns None if the pointer is noull
if let Some(protocol) = NonNull::new(status_code_ptr) {
    // SAFETY: <comment>
    unsafe { (protocol.as_ref().report_status_code)(<params>) };
}
```

patina_dxe_core/src/lib.rs
```diff
@@ -585,25 +585,31 @@ fn call_bds() -> ! {
...
 
     match protocols::PROTOCOL_DB.locate_protocol(bds::PROTOCOL_GUID) {
-        Ok(bds) => {
-            let bds = bds as *mut bds::Protocol;
-            // SAFETY: The BDS arch protocol is the valid C structure as defined by the UEFI specification. The entry
-            // field of the protocol is a valid function pointer that conforms to the expected calling convention.
-            unsafe {
-                ((*bds).entry)(bds);
+        Ok(bds_ptr) => {
+            if !bds_ptr.is_null() {
```
Javagedes: Same here

patina_dxe_core/src/lib.rs
```diff
@@ -649,4 +659,19 @@ mod tests {
 
         assert!(!<TestPlatform as MemoryInfo>::prioritize_32_bit_memory());
     }
+
+    #[test]
+    #[should_panic = "BDS arch protocol should be found and should never return."]
+    fn test_mock_call_bds() {
```
Javagedes: This needs way more tests. Basically a full enumerated test of scenarios where `status_code_protocol` does or does not exist, and if it does exist, is or is not null and where `bds` does or does not exist, and if it does exist, is or is not null.

patina_dxe_core/src/lib.rs
```diff
@@ -585,25 +585,31 @@ fn call_bds() -> ! {
     // Enable status code capability in Firmware Performance DXE.
     match protocols::PROTOCOL_DB.locate_protocol(status_code::PROTOCOL_GUID) {
```
Javagedes, line 586: I would reduce this double match statement (once you use NonNull) to this:

```rust
    if let Ok(status_code_ptr) = protocols::PROTOCOL_DB.locate_protocol(status_code::PROTOCOL_GUID)
      && let Some(status_code_protocol) = NonNull::new(status_code_ptr) {
        <call the function>
    }
```

patina_dxe_core/src/lib.rs
makubacki: I was thinking about asking for this too, I wasn't sure how important it was though. It is definitely cleaner.

PR #1112, opened 2025-11-21
patina_smbios: Fix protocol installation and cleanup implementation

kat-perez:
## Description
Fixes SMBIOS protocol initialization to prevent allocation failures during SmbiosAdd() operations.

## What changed

**Fixed initialization order:**
- Allocate SMBIOS buffers before installing protocol (was failing when SmbiosAdd() called during install)
- Publish initial table with Type 127 (End-Of-Table) marker immediately after protocol install
- Subsequent SmbiosAdd() calls update buffer in-place

**C protocol compatibility improvements:**
- GetNext now returns pointers into the published table (not internal buffer), ensuring RecordA != RecordB for different records
- Added checksum-based table integrity verification to detect direct modifications
- Returns `TableDirectlyModified` error if published table was modified directly instead of using protocol APIs
- Error message guides users to use Remove() + Add() or UpdateString() instead

**Safety improvements:**
- Scoped manager locks to prevent TplMutex re-entrant lock panic during republish
- Use write_unaligned() for FFI pointer writes
- Narrowed unsafe scopes in protocol layer

**Code cleanup:**
- Use MemoryManager service for allocations
- Replace hard-coded sizes with SIZE_(n)KB constants
- Enhanced error handling and debug_asserts

---
- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested
- Integrated with Tiano DXE Core to test C FFI protocol functions
- Integrated with patched branch in patina-dxe-core-qemu to ensure it also works with Patina DXE Core
- Verified direct modification detection with test in PatinaSmbiosDxe driver

## Integration Instructions
N/A

### DIFF
--- a/components/patina_smbios/src/component.rs
+++ b/components/patina_smbios/src/component.rs
@@ -11,13 +11,13 @@
 extern crate alloc;
 use crate::{
     error::SmbiosError,
-    manager::{SmbiosManager, install_smbios_protocol},
-    service::SmbiosImpl,
+    manager::SmbiosManager,
+    service::{Smbios, SmbiosImpl},
 };
 use alloc::boxed::Box;
 use patina::{
     boot_services::tpl::Tpl,
-    component::{Storage, component},
+    component::{Storage, component, service::memory::MemoryManager},
     error::Result,
     tpl_mutex::TplMutex,
 };
@@ -55,8 +55,8 @@ impl SmbiosConfiguration {
 /// - Record management: `update_string()`, `remove()`
 /// - Table management: `version()`, `publish_table()`
 ///
-/// The provider creates an SMBIOS manager instance protected by a TplMutex.
-/// A global reference is maintained for C/EDKII protocol compatibility.
+/// The provider creates an SMBIOS manager instance protected by a TplMutex
+/// and installs the SMBIOS protocol for C/EDKII driver compatibility.
 ///
 /// # Example
 ///
@@ -98,20 +98,22 @@ impl SmbiosProvider {
     fn entry_point(self, storage: &mut Storage) -> Result<()> {
         let cfg = self.config;
 
-        // Create the manager with configured version
-        let manager = SmbiosManager::new(cfg.major_version, cfg.minor_version).map_err(|e| {
-            log::error!("Failed to create SMBIOS manager: {:?}", e);
-            patina::error::EfiError::Unsupported
-        })?;
+        let manager = SmbiosManager::new(cfg.major_version, cfg.minor_version)?;
 
-        // Get boot_services from storage - it already has 'static lifetime
+        // SAFETY: Storage guarantees that boot_services has a 'static lifetime.
         // We use an unsafe cast because storage.boot_services() returns a reference with
-        // a shorter lifetime, but Storage guarantees boot_services lives for 'static
+        // a shorter lifetime bound, but Storage guarantees it lives for 'static.
         let boot_services_static: &'static patina::boot_services::StandardBootServices =
-            // SAFETY: Storage guarantees that boot_services has a 'static lifetime.
             unsafe { &*(storage.boot_services() as *const _) };
 
-        // Create the SMBIOS service struct with owned TplMutex
+        // Get the MemoryManager service for memory allocations
+        let memory_manager = storage.get_service::<dyn MemoryManager>().ok_or(patina::error::EfiError::Unsupported)?;
+
+        // Allocate buffers and add Type 127 End-of-Table marker
+        // This must be done before protocol installation to avoid allocate_pages during Add()
+        manager.allocate_buffers(*memory_manager)?;
+
+        // Create TplMutex at TPL_NOTIFY for thread safety against timer interrupts
         let manager_mutex = TplMutex::new(boot_services_static, Tpl::NOTIFY, manager);
         let smbios_service = SmbiosImpl {
             manager: manager_mutex,
@@ -120,19 +122,22 @@ impl SmbiosProvider {
             minor_version: cfg.minor_version,
         };
 
-        // Leak the service to get a 'static reference
-        // This single leak gives us a reference that can be used for both protocol and service
+        // Leak the service to get a 'static reference for both Rust service and C protocol
         let smbios_static: &'static SmbiosImpl = Box::leak(Box::new(smbios_service));
 
-        // Install the C/EDKII protocol using the leaked service's manager
-        if let Err(e) =
-            install_smbios_protocol(cfg.major_version, cfg.minor_version, smbios_static.manager(), boot_services_static)
-        {
-            log::error!("Failed to install SMBIOS protocol: {:?}", e);
-        }
-
-        // Register the leaked service (the IntoService impl for &'static T handles this)
         storage.add_service(smbios_static);
+
+        // Install SMBIOS protocol for C/EDKII driver compatibility
+        crate::manager::install_smbios_protocol(
+            cfg.major_version,
+            cfg.minor_version,
+            &smbios_static.manager,
+            boot_services_static,
+        )?;
+
+        // Publish initial table (Type 127 only) to register buffer with UEFI configuration table
+        // Subsequent Add() calls will update this buffer in-place
+        smbios_static.publish_table()?;
 
         Ok(())
     }
--- a/components/patina_smbios/src/manager.rs
+++ b/components/patina_smbios/src/manager.rs
@@ -21,7 +21,7 @@ mod protocol;
 mod record;
 
 // Re-export main types and functions
-pub use core::SmbiosManager;
+pub use core::{SMBIOS_3_X_TABLE_GUID, SmbiosManager};
 pub(crate) use record::SmbiosRecord;
 
 use alloc::boxed::Box;
@@ -55,7 +55,7 @@ pub fn install_smbios_protocol(
     boot_services: &'static StandardBootServices,
 ) -> Result<efi::Handle, SmbiosError> {
     // Create the protocol instance with internal struct
-    let internal = SmbiosProtocolInternal::new(major_version, minor_version, manager_mutex);
+    let internal = SmbiosProtocolInternal::new(major_version, minor_version, manager_mutex, boot_services);
     let interface = Box::into_raw(Box::new(internal));
 
     // Install the protocol using the unchecked interface since we have a raw pointer
--- a/components/patina_smbios/src/manager/core.rs
+++ b/components/patina_smbios/src/manager/core.rs
@@ -14,11 +14,7 @@ extern crate alloc;
 
 use alloc::{boxed::Box, string::String, vec::Vec};
 use core::cell::RefCell;
-use patina::{
-    boot_services::{BootServices, allocation::AllocType},
-    efi_types::EfiMemoryType,
-    uefi_size_to_pages,
-};
+use patina::{base::SIZE_64KB, uefi_size_to_pages};
 use r_efi::{
     efi,
     efi::{Handle, PhysicalAddress},
@@ -32,6 +28,7 @@ use crate::{
         SMBIOS_HANDLE_PI_RESERVED, SMBIOS_STRING_MAX_LENGTH, SmbiosHandle, SmbiosRecordsIter, SmbiosTableHeader,
         SmbiosType,
     },
+    smbios_record::{SmbiosRecordStructure, Type127EndOfTable},
 };
 
 use super::record::SmbiosRecord;
@@ -82,6 +79,16 @@ pub struct SmbiosManager {
     pub minor_version: u8,
     entry_point_64: RefCell<Option<Box<Smbios30EntryPoint>>>,
     table_64_address: RefCell<Option<PhysicalAddress>>,
+    /// Pre-allocated buffer for SMBIOS table data
+    table_buffer_addr: RefCell<Option<PhysicalAddress>>,
+    /// Maximum size of the pre-allocated table buffer
+    table_buffer_max_size: usize,
+    /// Pre-allocated buffer for entry point structure
+    ep_buffer_addr: RefCell<Option<PhysicalAddress>>,
+    /// Checksum of published table (for detecting direct modifications)
+    published_table_checksum: RefCell<Option<u32>>,
+    /// Size of the published table (needed for checksum verification)
+    published_table_size: RefCell<usize>,
 }
 
 impl SmbiosManager {
@@ -113,7 +120,66 @@ impl SmbiosManager {
             minor_version,
             entry_point_64: RefCell::new(None),
             table_64_address: RefCell::new(None),
+            // Pre-allocated buffers start as None, set up during component init
+            table_buffer_addr: RefCell::new(None),
+            table_buffer_max_size: SIZE_64KB,
+            ep_buffer_addr: RefCell::new(None),
+            published_table_checksum: RefCell::new(None),
+            published_table_size: RefCell::new(0),
         })
+    }
+
+    /// Allocate buffers for SMBIOS table publication
+    ///
+    /// Allocates buffers for the SMBIOS table and entry point upfront to avoid
+    /// repeated allocations. The buffers are reused for all table publications.
+    ///
+    /// Also adds the Type 127 End-of-Table marker to maintain the invariant that
+    /// it's always the last record.
+    ///
+    /// # Errors
+    ///
+    /// Returns `SmbiosError::AllocationFailed` if buffer allocation fails.
+    /// Returns `SmbiosError::HandleExhausted` if unable to allocate handle for Type 127.
+    pub(crate) fn allocate_buffers(
+        &self,
+        memory_manager: &dyn patina::component::service::memory::MemoryManager,
+    ) -> Result<(), SmbiosError> {
+        // Check if already allocated
+        if self.table_buffer_addr.borrow().is_some() {
+            return Ok(());
+        }
+
+        use patina::{component::service::memory::AllocationOptions, efi_types::EfiMemoryType};
+
+        // Allocate table buffer
+        let table_pages = uefi_size_to_pages!(self.table_buffer_max_size);
+        let table_allocation = memory_manager
+            .allocate_pages(table_pages, AllocationOptions::new().with_memory_type(EfiMemoryType::ACPIReclaimMemory))
+            .map_err(|_| SmbiosError::AllocationFailed)?;
+        let table_slice = table_allocation.into_raw_slice::<u8>();
+        let table_addr = table_slice as *mut u8 as u64;
+
+        // Allocate entry point buffer (1 page is plenty)
+        let ep_allocation = memory_manager
+            .allocate_pages(1, AllocationOptions::new().with_memory_type(EfiMemoryType::ACPIReclaimMemory))
+            .map_err(|_| SmbiosError::AllocationFailed)?;
+        let ep_slice = ep_allocation.into_raw_slice::<u8>();
+        let ep_addr = ep_slice as *mut u8 as u64;
+
+        *self.table_buffer_addr.borrow_mut() = Some(table_addr);
+        *self.ep_buffer_addr.borrow_mut() = Some(ep_addr);
+
+        // Automatically add Type 127 End-of-Table marker to ensure SMBIOS compliance
+        // This is added directly to records to bypass the add_from_bytes validation
+        // which rejects Type 127 to prevent external callers from adding it
+        let type127 = Type127EndOfTable::new();
+        let bytes = type127.to_bytes();
+        let header = SmbiosTableHeader::new(127, 4, self.allocate_handle()?);
+        let record = SmbiosRecord::new(header, None, bytes, 0);
+        self.records.borrow_mut().push(record);
+
+        Ok(())
     }
 
     /// Validate a string for use in SMBIOS records
@@ -259,64 +325,45 @@ impl SmbiosManager {
         Ok(candidate)
     }
 
-    /// Builds the SMBIOS table and installs it in the UEFI Configuration Table
+    /// Build SMBIOS table data and entry point using pre-allocated buffers
     ///
-    /// This function performs the following steps:
-    /// 1. Consolidates all SMBIOS records into a contiguous memory buffer
-    /// 2. Creates an SMBIOS 3.x Entry Point Structure with proper checksum
-    /// 3. Allocates ACPI Reclaim memory for both the table and entry point
-    /// 4. Installs the entry point via the UEFI Configuration Table
+    /// Copies table data into pre-allocated buffers without calling allocate_pages.
+    /// This allows safe republishing during Add/Update/Remove operations.
     ///
-    /// # Arguments
+    /// Returns (table_address, ep_address, entry_point) but does NOT install the configuration table.
+    /// The caller must call install_configuration_table separately without holding locks.
     ///
-    /// * `boot_services` - UEFI Boot Services for memory allocation and table installation
-    ///
-    /// # Returns
-    ///
-    /// Returns a tuple of `(table_address, entry_point_address)` containing the physical
-    /// addresses where the SMBIOS table data and entry point structure were allocated.
-    ///
-    /// # Errors
-    ///
-    /// * `SmbiosError::NoRecordsAvailable` - No SMBIOS records have been added
-    /// * `SmbiosError::AllocationFailed` - Failed to allocate memory or install the configuration table
-    ///
-    /// # Safety
-    ///
-    /// This function uses unsafe code for:
-    /// - Creating mutable slices to allocated memory
-    /// - Writing the entry point structure to allocated memory
-    /// - Calling the UEFI `install_configuration_table` interface
-    ///
-    /// All memory allocations use UEFI Boot Services and are properly tracked by the firmware.
-    pub fn install_configuration_table(
-        &self,
-        boot_services: &patina::boot_services::StandardBootServices,
-    ) -> Result<(PhysicalAddress, PhysicalAddress), SmbiosError> {
+    pub fn build_table_data(&self) -> Result<(PhysicalAddress, PhysicalAddress, Smbios30EntryPoint), SmbiosError> {
+        // Get pre-allocated buffer addresses
+        let table_address = self.table_buffer_addr.borrow().ok_or(SmbiosError::AllocationFailed)?;
+        let ep_address = self.ep_buffer_addr.borrow().ok_or(SmbiosError::AllocationFailed)?;
+
+        // Borrow records
         let records = self.records.borrow();
+
+        // Verify invariant: Type 127 End-of-Table marker must be last record
+        debug_assert!(
+            records.last().map(|r| r.header.record_type) == Some(127),
+            "Type 127 End-of-Table marker must be the last record in the table"
+        );
 
         // Step 1: Calculate total table size
         let total_table_size: usize = records.iter().map(|r| r.data.len()).sum();
 
+        debug_assert!(total_table_size > 0, "Cannot build table: no SMBIOS records have been added");
         if total_table_size == 0 {
-            log::error!("Cannot install configuration table: no SMBIOS records have been added");
             return Err(SmbiosError::NoRecordsAvailable);
         }
 
-        // Step 2: Allocate memory for the table (using UEFI Boot Services memory allocation)
-        let table_pages = uefi_size_to_pages!(total_table_size);
-        let table_address = boot_services
-            .allocate_pages(
-                AllocType::AnyPage,
-                EfiMemoryType::ACPIReclaimMemory, // SMBIOS tables go in ACPI Reclaim memory
-                table_pages,
-            )
-            .map_err(|_| SmbiosError::AllocationFailed)?;
+        // Step 2: Check size fits in pre-allocated buffer
+        if total_table_size > self.table_buffer_max_size {
+            return Err(SmbiosError::AllocationFailed);
+        }
 
-        // Step 3: Copy all records to the table
-        // SAFETY: `table_address` was just successfully allocated by UEFI Boot Services for
-        // `total_table_size` bytes. The memory is valid, properly aligned, and exclusively owned
-        // by this function. The size matches our calculation, making this slice creation safe.
+        // Step 3: Copy all records to the pre-allocated buffer
+        // Records Vec maintains invariant: Type 127 is always last
+        // So we can just copy in order
+        // SAFETY: table_address points to pre-allocated buffer with size >= total_table_size
         let table_slice = unsafe { core::slice::from_raw_parts_mut(table_address as *mut u8, total_table_size) };
         let mut offset = 0;
 
@@ -326,7 +373,7 @@ impl SmbiosManager {
             offset += record_bytes.len();
         }
 
-        // Step 4: Create SMBIOS 3.0+ Entry Point Structure
+        // Step 4: Create entry point structure
         let mut entry_point = Smbios30EntryPoint {
             anchor_string: *b"_SM3_",
             checksum: 0,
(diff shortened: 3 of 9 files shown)


components/patina_smbios/src/manager/core.rs
```diff
@@ -82,6 +83,13 @@ pub struct SmbiosManager {
     pub minor_version: u8,
     entry_point_64: RefCell<Option<Box<Smbios30EntryPoint>>>,
     table_64_address: RefCell<Option<PhysicalAddress>>,
+    // Pre-allocated buffers for table publication
```
os-d:
```suggestion
```

components/patina_smbios/src/manager/core.rs
```diff
@@ -113,9 +121,63 @@ impl SmbiosManager {
             minor_version,
             entry_point_64: RefCell::new(None),
             table_64_address: RefCell::new(None),
+            // Pre-allocated buffers start as None, set up during component init
+            table_buffer_addr: RefCell::new(None),
+            table_buffer_max_size: 64 * 1024, // 64KB default max size
```
os-d:
```suggestion
            table_buffer_max_size: SIZE_64KB,
```
Also include `use patina::base::SIZE_64KB`

components/patina_smbios/src/manager/core.rs
```diff
@@ -113,9 +121,63 @@ impl SmbiosManager {
...
+    /// Also adds the Type 127 End-of-Table marker to maintain the invariant that
+    /// it's always the last record.
+    ///
+    /// Must be called at TPL_APPLICATION or lower to allow memory allocation.
+    ///
+    /// # Errors
+    ///
+    /// Returns `SmbiosError::AllocationFailed` if buffer allocation fails.
+    /// Returns `SmbiosError::HandleExhausted` if unable to allocate handle for Type 127.
+    #[cfg_attr(coverage_nightly, coverage(off))]
```
os-d:
```suggestion
    #[coverage(off)]
```

We are just using this form. Please amend everywhere

components/patina_smbios/src/manager/core.rs
```diff
@@ -113,9 +121,63 @@ impl SmbiosManager {
...
 
+    /// Allocate buffers for SMBIOS table publication
+    ///
+    /// Allocates buffers for the SMBIOS table and entry point upfront to avoid
+    /// repeated allocations. The buffers are reused for all table publications.
+    ///
+    /// Also adds the Type 127 End-of-Table marker to maintain the invariant that
+    /// it's always the last record.
+    ///
+    /// Must be called at TPL_APPLICATION or lower to allow memory allocation.
```
os-d: Why does TPL level come into play here? There is no lowever level than TPL_APPLICATION, but you can make memory allocations at any TPL.

components/patina_smbios/src/manager/core.rs
```diff
@@ -113,9 +121,63 @@ impl SmbiosManager {
...
+    ///
+    /// # Errors
+    ///
+    /// Returns `SmbiosError::AllocationFailed` if buffer allocation fails.
+    /// Returns `SmbiosError::HandleExhausted` if unable to allocate handle for Type 127.
+    #[cfg_attr(coverage_nightly, coverage(off))]
+    pub fn allocate_buffers(
```
os-d: Why is this a public fn? Seems like this is something the service would simply do as part of its initialization

components/patina_smbios/src/manager/core.rs
```diff
@@ -113,9 +121,63 @@ impl SmbiosManager {
...
+        &self,
+        boot_services: &patina::boot_services::StandardBootServices,
```
os-d: We shouldn't use boot services for allocation here, we should be using the [memory manager service](https://github.com/OpenDevicePartnership/patina/blob/main/sdk/patina/src/component/service/memory.rs)

components/patina_smbios/src/manager/core.rs
```diff
@@ -337,46 +374,31 @@ impl SmbiosManager {
...
+        Ok((table_address, ep_address, entry_point))
+    }
 
-        // Store addresses for future reference
-        drop(records); // Release borrow before mutating
+    /// Store table addresses after installation
+    pub fn store_table_addresses(
+        &self,
+        table_address: PhysicalAddress,
+        _ep_address: PhysicalAddress,
```
os-d: `_ep_address` isn't needed?

components/patina_smbios/src/manager/protocol.rs
```diff
@@ -130,7 +155,14 @@ impl SmbiosProtocol {
         // SAFETY: We must trust the C code was a responsible steward of this pointer
         // Cast from protocol pointer to internal struct pointer
         let internal = unsafe { &*(protocol as *const SmbiosProtocolInternal) };
-        let manager = internal.manager.lock();
+
+        let manager = match internal.manager.try_lock() {
+            Ok(guard) => guard,
+            Err(_) => {
+                log::error!("[SMBIOS Add] ERROR: try_lock FAILED - mutex already locked!");
```
os-d: debug_assert is helpful to catch these cases, instead of just a print that is easily missed. Same with others.

components/patina_smbios/src/manager/core.rs
```diff
@@ -566,9 +596,17 @@ impl SmbiosManager {
     /// # Returns
     ///
     /// Returns an iterator yielding tuples of `(SmbiosTableHeader, Option<Handle>)`.
+    #[allow(dead_code)]
```
os-d: Is this actually dead code? Looks like the protocol calls this?

components/patina_smbios/src/manager.rs
```diff
@@ -47,6 +47,7 @@ use self::protocol::{SmbiosProtocol, SmbiosProtocolInternal};
 ///
 /// Returns the protocol handle on success. The caller is responsible for managing the
 /// protocol lifetime (though in practice, UEFI protocols persist for the system lifetime).
+#[allow(dead_code)]
```
os-d: Is this dead code? I assume you are always installing the protocol?

You've added a lot of allow dead code in this PR, but without removing calls to these functions. What changed?

components/patina_smbios/src/service.rs
```diff
@@ -205,14 +203,29 @@ impl SmbiosImpl {
...
+    /// Rebuild and republish table data after a record mutation
+    ///
+    /// This updates the pre-allocated buffer with the latest records
+    fn republish_table(&self) -> core::result::Result<(), crate::error::SmbiosError> {
+        let (table_addr, ep_addr, entry_point) = {
+            let manager = self.manager.lock();
+            manager.build_table_data()?
+        };
+
+        let manager = self.manager.lock();
```
os-d, line 213: Seems odd to get the lock, drop it, then get it again

components/patina_smbios/src/service.rs
```diff
@@ -225,8 +238,29 @@ impl Smbios for SmbiosImpl {
...
     {
-        let manager = self.manager.lock();
-        manager.publish_table(self.boot_services)
+        let (table_addr, ep_addr, entry_point) = {
+            let manager = self.manager.lock();
+            manager.build_table_data()?
+        };
+
+        // Install configuration table without holding lock to avoid deadlock
+        // If this triggers an event that calls back into SMBIOS, it can acquire the lock
```
os-d: Is this something you've seen?

components/patina_smbios/src/manager/core.rs
kat-perez: i think i meant that it needed to be lower than what I set the TplMutex level which is at NOTIFICATION, but I'll just remove the comment since it doesn't add much

components/patina_smbios/src/manager/core.rs
kat-perez: yeah, i'll change it to pub(crate) instead

components/patina_smbios/src/manager/core.rs
kat-perez: ah yeah. removing this and the one on get_record_data() too

components/patina_smbios/src/manager.rs
kat-perez: hmm I think I added these by mistake. I'll look through again and add a comment for why the allow dead_code was needed

components/patina_smbios/src/manager/core.rs
```diff
@@ -259,64 +322,39 @@ impl SmbiosManager {
...
-    /// - Calling the UEFI `install_configuration_table` interface
-    ///
-    /// All memory allocations use UEFI Boot Services and are properly tracked by the firmware.
-    pub fn install_configuration_table(
-        &self,
-        boot_services: &patina::boot_services::StandardBootServices,
-    ) -> Result<(PhysicalAddress, PhysicalAddress), SmbiosError> {
+    /// Returns (table_address, ep_address, entry_point) but does NOT install the configuration table.
+    /// The caller must call install_configuration_table separately without holding locks.
+    #[coverage(off)]
```
joschock: Consider adding a comment describing why we can't do coverage on this fn.

components/patina_smbios/src/manager/protocol.rs
```diff
@@ -42,11 +42,15 @@ pub(super) struct SmbiosProtocolInternal {
...
     // Internal component access only! Does not exist in C definition
-    pub(super) manager: &'static TplMutex<'static, SmbiosManager, StandardBootServices>,
+    pub(super) manager: &'static TplMutex<'static, SmbiosManager>,
 
-    // Storage for header returned by C protocol's get_next
+    // Boot services needed for table republishing after Add/Update/Remove
+    pub(super) boot_services: &'static patina::boot_services::StandardBootServices,
+
+    // Storage for complete record (header + data + strings) returned by C protocol's get_next
+    // Must be large enough to hold any SMBIOS record. Max record size is ~64KB per spec.
```
joschock: if max size is 64K, why not just use that instead of 8K?

components/patina_smbios/src/manager/protocol.rs
```diff
@@ -223,7 +258,16 @@ impl SmbiosProtocol {
             };
 
             match manager.update_string(handle, str_num, rust_str) {
-                Ok(()) => efi::Status::SUCCESS,
+                Ok(()) => {
+                    drop(manager);
```
joschock: recommend an informative comment on why an explicit drop is needed here.

components/patina_smbios/src/manager/protocol.rs
```diff
@@ -260,45 +313,72 @@ impl SmbiosProtocol {
 
         // SAFETY: Cast from protocol pointer to internal struct pointer
         let internal = unsafe { &*(protocol as *const SmbiosProtocolInternal) };
-        let manager = internal.manager.lock();
+
+        let result = {
+            let manager = internal.manager.lock();
+
+            // SAFETY: C UEFI protocol caller guarantees pointers are valid
+            unsafe {
```
joschock: consider narrowing the scope of the unsafe to just the pointer derefs:
```rust
let (handle, type_filter) = unsafe {
  (*smbios_handle, if record_type.is_null() { None } else { Some(*record_type) })
};
//safe logic
```

components/patina_smbios/src/manager/protocol.rs
```diff
@@ -260,45 +313,72 @@ impl SmbiosProtocol {
...
+                    }
+                    None => (SMBIOS_HANDLE_PI_RESERVED, SmbiosTableHeader::new(0, 0, 0), None, false),
+                }
+            }
+        }; // manager borrow drops here
 
         // SAFETY: The pointers are checked for being null above. They are expected to point
         // to valid data objects due to the expectations of the protocol interface. The type_filter
         // is optionally dereferenced after a null check.
         unsafe {
```
joschock, line 356: Consider narrowing the scope of this unsafe as well so it covers as little code as possible.

components/patina_smbios/src/manager/protocol.rs
joschock, line 309: The FFI code may pass pointers that are not properly aligned, and this will cause things like:
`*record = foo` to panic for that reason.

For the best safety/compatibility, recommend using explicit `write_unaligned` and `read_unalgined` on these pointers (as well as other FFI pointers in other APIs - won't comment on them all). https://doc.rust-lang.org/std/ptr/fn.write_unaligned.html

e.g.
```rust
*record = buffer.as_mut_ptr() as *mut SmbiosTableHeader;
```
becomes
```rust
record.write_unaligned(buffer.as_mut_ptr() as *mut SmbiosTableHeader);
```

components/patina_smbios/src/component.rs
```diff
@@ -98,20 +98,22 @@ impl SmbiosProvider {
...
-            patina::error::EfiError::Unsupported
-        })?;
+        let manager = SmbiosManager::new(cfg.major_version, cfg.minor_version)?;
 
-        // Get boot_services from storage - it already has 'static lifetime
+        // SAFETY: Storage guarantees that boot_services has a 'static lifetime.
         // We use an unsafe cast because storage.boot_services() returns a reference with
-        // a shorter lifetime, but Storage guarantees boot_services lives for 'static
+        // a shorter lifetime bound, but Storage guarantees it lives for 'static.
         let boot_services_static: &'static patina::boot_services::StandardBootServices =
```
joschock, line 106: this seems weird to me; you shouldn't have to recast to workaround lifetimes. @Javagedes - is there a more appropriate way to manage boot_services?

components/patina_smbios/src/service.rs
```diff
@@ -225,8 +233,29 @@ impl Smbios for SmbiosImpl {
...
+        // SAFETY: We pass a valid GUID and a pointer to ACPI_RECLAIM_MEMORY that remains valid
+        unsafe {
+            self.boot_services
+                .install_configuration_table_unchecked(
```
joschock: can this be reworked to use the checked version of install_configuration_table?

components/patina_smbios/src/service.rs
kat-perez: yes, i agree. not sure why i had it like that. i can adjust it

components/patina_smbios/src/service.rs
kat-perez: I found that in install_configuration_table in [patina_dxe_core/src/config_tables.rs:122](https://github.com/OpenDevicePartnership/patina/blob/e360b984ec0e39f5914a9db8cab95f85cc6b26be/patina_dxe_core/src/config_tables.rs#L122) calls `EVENT_DB.signal_group(vendor_guid). So if an event handler tried to call any of the SMBIOS API while holding the lock, it would deadlock.

My solution in `publish_table()` had to be:
1. lock->`build_table_data`->drop lock
2. call `install_configuration_table`
3. re-acquire lock->store_table_addresses()->drop lock

components/patina_smbios/src/service.rs
kat-perez: yes, done

components/patina_smbios/src/service.rs
kat-perez: fixed

components/patina_smbios/src/manager/core.rs
kat-perez: updated

components/patina_smbios/src/manager/protocol.rs
kat-perez: this is a mistake. that number was for the maximum table size. i checked again and i don't think there is a limit listed in the spec for records specifically. but to reach the 8KB rationale: each string has a limit of (SMBIOS_STRING_MAX_LENGTH == 64 bytes) and in the worst case we could assume ~100 strings. if we want a some buffer over extreme cases 125 strings *  64 = 8KB
updated comment to say this instead:
```
// 8KB is sufficient for any realistic SMBIOS record (255 bytes structured data + ~7.7KB strings)
```

components/patina_smbios/src/component.rs
kat-perez: lemme think through some other options too

components/patina_smbios/src/manager/protocol.rs
kat-perez: added

components/patina_smbios/src/component.rs
kat-perez: @joschock @Javagedes here's a proposal but it would be a breaking change because the signature changes from `TplMutex<'a, T, B>` to `TplMutex<T, B>` and all code using TplMutex must remove lifetime parameters
Modify TplMutex to own BootServices instead of borrow
```
  // BEFORE:
  pub struct TplMutex<'a, T: ?Sized, B: BootServices> {
      boot_services: &'a B,  // Borrowed
      // ...
  }
```
```
  // AFTER:
  pub struct TplMutex<T: ?Sized, B: BootServices = StandardBootServices> {
      boot_services: B,  // Owned
      // ...
  }
  impl<T, B: BootServices + Clone> TplMutex<T, B> {
      pub fn new(boot_services: &B, ...) -> Self {
          Self { boot_services: boot_services.clone(), ... }
      }
  }
```
Make MockBootServices cloneable with default expectations (sdk/patina/src/boot_services.rs)
```
  impl Clone for MockBootServices {
      fn clone(&self) -> Self {
          let mut mock = Self::new();
          // Add default TPL expectations so TplMutex can use cloned mocks
          mock.expect_raise_tpl().times(..).returning(|tpl| tpl);
          mock.expect_restore_tpl().times(..).returning(|_| ());
          mock
      }
  }
```

Remove unsafe cast in component.rs
```
  // BEFORE (unsafe):
  let boot_services_static: &'static StandardBootServices =
      unsafe { &*(storage.boot_services() as *const _) };
  let manager_mutex = TplMutex::new(boot_services_static, Tpl::NOTIFY, manager);

  // AFTER (safe):
  let manager_mutex = TplMutex::new(storage.boot_services(), Tpl::NOTIFY, manager);
  let boot_services = storage.boot_services().clone();
  let smbios_service = SmbiosImpl { manager: manager_mutex, boot_services, ... };
```

components/patina_smbios/src/manager/core.rs
kat-perez: done, i've just removed it

components/patina_smbios/src/component.rs
kat-perez: Branch of how this works (tested and integrated successfully): https://github.com/kat-perez/patina/tree/smbios-experimental-boot-services-static

components/patina_smbios/src/manager/protocol.rs
```diff
@@ -260,14 +320,17 @@ impl SmbiosProtocol {
...
-        // SAFETY: The pointers are checked for being null above. They are expected to point
-        // to valid data objects due to the expectations of the protocol interface. The type_filter
-        // is optionally dereferenced after a null check.
-        unsafe {
-            let handle = *smbios_handle;
-            let type_filter = if record_type.is_null() { None } else { Some(*record_type) };
+        let result = {
+            let manager = internal.manager.lock();
+
+            // SAFETY: C UEFI protocol caller guarantees pointers are valid
```
makubacki, line 328: I'm concerned we're going to dismiss some safety scenarios with too much trust in C callers. In this case, there is not a *guarantee* that the protocol pointer is valid. It should be, but it is not guaranteed. I also think that `protocol` should be checked for null as well before converting it to `*const SmbiosProtocolInternal`.

components/patina_smbios/src/service.rs
makubacki: Is there a risk that in (2) the table data changes and (3) operates on stale data?

components/patina_smbios/src/manager/core.rs
```diff
@@ -455,7 +493,17 @@ impl SmbiosManager {
 
         let smbios_record = SmbiosRecord::new(record_header, producer_handle, data, string_count);
 
-        self.records.borrow_mut().push(smbios_record);
+        // Maintain invariant: Type 127 End-of-Table marker must always be last
+        let mut records = self.records.borrow_mut();
+
+        // Insert before Type 127 if it exists as the last element, otherwise append
+        if records.last().is_some_and(|r| r.header.record_type == 127) {
```
makubacki: Is there a situation other than a failure scenario where Type 127 would not be the last record?

components/patina_smbios/src/manager/protocol.rs
kat-perez: added null checks on the protocol pointer and also added pointer alignment validation

components/patina_smbios/src/service.rs
kat-perez: Yeah, that is true. I'm thinking about this scenario:
1. lock -> `build_table_data()` -> unlock
2. call `install_configuration_table()`
    - triggers `EVENT_DB.signal_group()`
    - event handler fires and a driver calls `SmbiosAdd()`
    - `SmbiosAdd()` calls `republish_table()`
        - lock -> `build_table_data()` > writes entry_point_B to ep_addr
        - calls `store_table_addresses(table_addr, entry_point_B)`
        - now ep_addr buffer contains entry_point_B
3. lock -> `store_table_addresses(entry_point_A)` -> unlock
    - overwrites entry_point_64 with a stale entry_point_A

New solution:
remove step (3) entirely

components/patina_smbios/src/manager/core.rs
kat-perez: no, there isn't a non-failure scenario where Type 127 would not be the last

makubacki: Can you please add a `debug_assert!()` then?

kat-perez: added a debug_assert to `build_table_data()` and `add_from_bytes()` and unit tests

components/patina_smbios/src/manager/core.rs
```diff
@@ -113,9 +116,79 @@ impl SmbiosManager {
...
+    /// Returns `SmbiosError::AllocationFailed` if buffer allocation fails.
+    /// Returns `SmbiosError::HandleExhausted` if unable to allocate handle for Type 127.
+    ///
+    /// ## Coverage Exclusion
+    ///
+    /// This function is excluded from unit test coverage (`#[coverage(off)]`) because testing it in
+    /// unit tests would require significant test infrastructure:
+    /// 1. PageAllocation requires a `&'static MemoryManager` reference (requires Box::leak or unsafe transmute)
+    /// 2. Mock MemoryManager would need to allocate real memory and handle free_pages() on drop
+    /// 3. Test would need to manage memory lifecycle without UEFI page allocator
```
joschock: nitpick: I appreciate the detail, but I think there's a danger of it becoming obsolete over time if details of memory allocation change. I think it would be sufficient to say something like "testing it in unit tests would require test infrastructure to mock memory allocations"

Also, looking at the implementation, I think you'd just instantiate a `MockMemoryManager` and drop a couple of expectations on it for two `allocate_pages()` calls (though they would have to make allocations to meet the caller expectations).

components/patina_smbios/src/manager/core.rs
```diff
@@ -259,64 +332,57 @@ impl SmbiosManager {
...
+        // Verify invariant: Type 127 End-of-Table marker must be last record
+        debug_assert!(
+            records.last().map(|r| r.header.record_type) == Some(127),
+            "Type 127 End-of-Table marker must be the last record in the table"
+        );
+
         // Step 1: Calculate total table size
         let total_table_size: usize = records.iter().map(|r| r.data.len()).sum();
 
+        debug_assert!(total_table_size > 0, "Cannot build table: no SMBIOS records have been added");
```
joschock, line 353: nit: I don't think you'll ever hit this, because if you would, the debug_assert on 364 would have blown up first. I'm okay with keeping it if you feel it is useful for documentation or "future design change protection" purposes.

components/patina_smbios/src/manager/core.rs
```diff
@@ -424,21 +462,27 @@ impl SmbiosManager {
...
+        }
+
+        // Step 4: Validate header->length is <= (record_data.length - 2) for string pool
         // The string pool needs at least 2 bytes for the double-null terminator
         if (header.length as usize + 2) > record_data.len() {
             return Err(SmbiosError::RecordTooSmall);
         }
 
-        // Step 4: Validate and count strings in a single efficient pass
+        // Step 5: Validate and count strings in a single efficient pass
```
joschock: super-nit: step 5 and step 6 seem to be doing two parts of the same thing. consider combining the comment.

components/patina_smbios/src/manager/core.rs
```diff
@@ -455,7 +499,23 @@ impl SmbiosManager {
...
+        // Maintain invariant: Type 127 End-of-Table marker must always be last
+        let mut records = self.records.borrow_mut();
+
+        // Insert before Type 127 if it exists as the last element, otherwise append
+        if records.last().is_some_and(|r| r.header.record_type == 127) {
+            let last_index = records.len() - 1;
+            records.insert(last_index, smbios_record);
+        } else {
+            records.push(smbios_record);
+        }
```
joschock: You could avoid the inspection and insert. The perf impact is basically zero, so current code is fine, but consider: 
```suggestion
        records.push(smbios_record);
        //Swap the inserted element with the one before it (if it exists) to ensure type 127 stays at the end.
        if (records.len() > 1) {
            records.swap(records.len()-1, records.len()-2);
        }
```

components/patina_smbios/src/manager/protocol.rs
```diff
@@ -169,27 +197,46 @@ impl SmbiosProtocol {
...
+                unsafe {
+                    smbios_handle.write_unaligned(handle);
+                }
 
-            // Convert handle
-            let producer_opt = if producer_handle.is_null() { None } else { Some(producer_handle) };
+                // Explicitly drop the manager guard before republish_table() to release the TplMutex lock.
+                // republish_table() needs to acquire the same lock (line 117), and TplMutex doesn't
+                // support re-entrant locking - it will panic if the lock is already held.
+                drop(manager);
```
joschock: why not make republish_table() a member of manager so that it is behind the lock? then you wouldn't need to worry about dropping the manager here, because `manager.republish_table()` behind the lock would be a thing.

components/patina_smbios/src/manager/protocol.rs
```diff
@@ -283,23 +379,58 @@ impl SmbiosProtocol {
...
+                }
+                None => (SMBIOS_HANDLE_PI_RESERVED, SmbiosTableHeader::new(0, 0, 0), None, false),
+            }
+        }; // manager borrow drops here
 
-                    if !producer_handle.is_null() {
-                        *producer_handle = prod_handle.unwrap_or(core::ptr::null_mut());
-                    }
-                    efi::Status::SUCCESS
+        let (found_handle, _header_value, prod_handle, found) = result;
```
joschock: aesthetic; consider removing `result` and making line 357 be
 `let (found_handle, _header_value, prod_handle, found) = {`

components/patina_smbios/src/manager/protocol.rs
```diff
@@ -283,23 +379,58 @@ impl SmbiosProtocol {
...
                 }
-                None => {
-                    *smbios_handle = SMBIOS_HANDLE_PI_RESERVED;
-                    efi::Status::NOT_FOUND
+
+                buffer[..record_data.len()].copy_from_slice(&record_data);
+
+                // SAFETY: record pointer is guaranteed valid by caller
+                unsafe {
+                    record.write_unaligned(buffer.as_mut_ptr() as *mut SmbiosTableHeader);
```
joschock: I don't think using an internal buffer here is viable for two reasons: 

1) Consider the following C code:

```C
EFI_SMBIOS_TABLE_HEADER *RecordA;
EFI_SMBIOS_TABLE_HEADER *RecordB;
SmbiosProto->GetNext(SmbiosProto, Handle, NULL, &RecordA, NULL); //Get a record
SmbiosProto->GetNext(SmbiosProto, Handle, NULL, &RecordB, NULL); //Get the next record, presumably different.
ASSERT(RecordA != RecordB);
ASSERT(AreTheseTheSameRecord(RecordA, RecordB) == FALSE);
```
I think both asserts will fail; because internally the same buffer is used to satisfy the caller; i.e. once you call for RecordB, RecordA is effectively overwritten with the contents of RecordB. 

2) Callers of this function in the C codebase expect to get a buffer that they can directly manipulate to change SMBIOS contents. For example (culled from real platform code, but redacted to protect the guilty):

```C
SmBiosProtocol->GetNext (SmBiosProtocol, &SmbiosHandle, &Type, (EFI_SMBIOS_TABLE_HEADER **)&Record, NULL);
//... some logic to generate VersionString...
if (VersionString[0] == 0) {
    AsciiSPrint (UnusedStr, sizeof (UnusedStr), "%d", Record->Version);
    StringNumber                  = Record->Version;
    SmbiosHandle                  = Record->Hdr.Handle;
    Record->Version = 0; //NOTE: <- Directly modifies the record and expects it to stick.
    Status                        = SmBiosProtocol->UpdateString (SmBiosProtocol, &SmbiosHandle, &StringNumber, UnusedStr);

    // Else if VersionString has data, use that in the table
  } else {
    SmbiosHandle = Record->Hdr.Handle;
    StringNumber = Record->Version;
    Status       = SmBiosProtocol->UpdateString (SmBiosProtocol, &SmbiosHandle, &StringNumber, VersionString);
  }
```

components/patina_smbios/src/component.rs
joschock: I'm okay with a refactor of the `TplMutex` along those lines; but probably should be done as a separate PR. Recommend filing an issue to track and then closing this comment here with a pointer to it.

components/patina_smbios/src/manager/protocol.rs
joschock: For (2), the answer might be "we don't allow that because it's bad behavior" - but just want to note known compat issues with existing users of the protocol.

components/patina_smbios/src/service.rs
```diff
@@ -295,7 +336,7 @@ pub trait SmbiosExt {
 }
 
 /// Implementation of extension trait for `Service<dyn Smbios>`
-#[cfg_attr(coverage_nightly, coverage(off))]
+#[coverage(off)]
```
joschock: nit: small comment on justification for no coverage.

components/patina_smbios/src/service.rs
```diff
@@ -724,4 +765,123 @@ mod tests {
         let handle = service.add_record(None, &record).unwrap();
         assert_eq!(handle, 0x1337);
     }
+
+    // Unit tests for SmbiosImpl methods
+    //
+    // NOTE: The manager() and republish_table() methods cannot be fully tested in unit tests
+    // because SmbiosImpl requires StandardBootServices (not MockBootServices). These methods
```
joschock: requiring StandardBootServices is due to TplMutex impl? If so, maybe that can be addressed when refactoring TplMutex.

components/patina_smbios/src/manager/core.rs
kat-perez: ah, you're right. I've created two unit tests and used a MockMemoryManager. I also removed #[coverage(off)] for this function and build_table_data() too. So both should have proper coverage now

components/patina_smbios/src/manager/core.rs
kat-perez: combined the comments

components/patina_smbios/src/manager/core.rs
kat-perez: i think your suggestion is cleaner. i've updated to use the swap instead of insert

components/patina_smbios/src/manager/protocol.rs
kat-perez: good catch, there's no real good reason for it to be in the internal protocol functions. I've moved it to `SmbiosManager` now

components/patina_smbios/src/manager/protocol.rs
kat-perez: gotcha, updated by inlining the tuple destructuring

components/patina_smbios/src/service.rs
kat-perez: yeah, that's what I initially thought but I was able to revise the `SmbiosImpl` to be generic over BootServices for testability. so I'm using MockBootServices now. I have this commit separated out temporarily (will squash after I confirm this works after integration tests) but for now it builds: https://github.com/OpenDevicePartnership/patina/pull/1112/commits/22a28f72bc93e0be9dc9d955f375404f3c714936

components/patina_smbios/src/service.rs
kat-perez: just realized that we can cover this. i didn't know i could use `Service::mock`. removing #[coverage(off)]

components/patina_smbios/src/manager/core.rs
kat-perez: decided to just leave it in for future design change protection (like if for some reason the type 127 as the last record gets removed)

components/patina_smbios/src/component.rs
kat-perez: new issue opened here: https://github.com/OpenDevicePartnership/patina/issues/1144

components/patina_smbios/src/manager/protocol.rs
kat-perez: @joschock I've removed the internal buffer to match the existing C protocol behavior. GetNext now returns pointers directly into the published table, so the ASSERTs in (1) will pass (RecordA != RecordB).

For (2), direct modification is not supported since it bypasses validation. The correct approach is Remove + Add for header fields, or UpdateString for strings.

To help with debugging, I've added checksum verification: when Add/Remove/UpdateString triggers a republish, we detect if the table was modified directly and return EFI_DEVICE_ERROR with a log message explaining what happened and how to fix it. This ensures developers don't silently lose their changes. They'll see a log like this if they try to directly modify the record:
```
INFO - [ERROR] patina_smbios::manager::core: [SMBIOS] Published table was modified directly (checksum mismatch: expected 00001A1E, found 00001A1F). Use Remove() + Add() to modify records, or UpdateString() for string fields.
```

components/patina_smbios/src/manager/core.rs
```diff
@@ -113,9 +120,68 @@ impl SmbiosManager {
...
+        memory_manager: &dyn patina::component::service::memory::MemoryManager,
+    ) -> Result<(), SmbiosError> {
+        // Check if already allocated
+        if self.table_buffer_addr.borrow().is_some() {
+            return Ok(());
+        }
+
+        use patina::{component::service::memory::AllocationOptions, efi_types::EfiMemoryType};
+
+        // Allocate table buffer (64KB)
```
joschock: super-nit: since the size is specified by `table_buffer_max_size`, I'd avoid putting that constant in the comment here because if it is later updated, you would have to go fix all the comments too.

components/patina_smbios/src/manager/core.rs
```diff
@@ -337,46 +384,107 @@ impl SmbiosManager {
...
-        self.table_64_address.replace(Some(table_address as u64));
+        self.table_64_address.replace(Some(table_address));
+    }
+
+    /// Calculate a simple checksum for table data
+    ///
+    /// Uses a simple sum of all bytes (wrapping). This is sufficient for
+    /// detecting accidental modifications, not cryptographic integrity.
+    fn calculate_table_checksum(data: &[u8]) -> u32 {
+        data.iter().fold(0u32, |acc, &byte| acc.wrapping_add(byte as u32))
```
joschock, line 430: This is fine for incidental checks, but would miss some edge cases (i.e. two fields change in a way that cancel each other out e.g. FieldA: 1->0, FieldB: 0->1). We have a pretty fast/simple hasher in the core that we use for hashing handles in protocol_db.rs, we could find a place to put it for more general use and use it here:
```Rust
let mut hasher = Xorshift64starHasher::default();
hasher.write(data);
cksum = hasher.finish() as u32; //truncate and keep bottom 32-bits of hash
```
This is an enhancement request, though, so feel free to defer it to a new PR/file an issue and move on with this PR without it.

components/patina_smbios/src/manager/protocol.rs
```diff
@@ -169,27 +176,41 @@ impl SmbiosProtocol {
...
-                Err(_) => efi::Status::DEVICE_ERROR,
+
+                if manager.republish_table().is_err() {
+                    log::error!("[SMBIOS Add] Failed to rebuild table");
+                    return efi::Status::DEVICE_ERROR;
+                }
+
+                efi::Status::SUCCESS
+            }
+            Err(SmbiosError::StringContainsNull) => efi::Status::INVALID_PARAMETER,
```
joschock, line 202: Is there  a reason we can't do a `From` impl on `SmbiosError` to do this conversion in a central location? that might not be possible if the same `SmbiosError` maps to different efi errors in different functions, but if it's consistent, it would be good to centralize it.

components/patina_smbios/src/manager/core.rs
kat-perez: only found two instances of that, so easy fix

components/patina_smbios/src/manager/core.rs
kat-perez: https://github.com/OpenDevicePartnership/patina/issues/1161

components/patina_smbios/src/manager/protocol.rs
kat-perez: I don't think there's any reason that we can't consolidate it. I will defer the consolidation to different PR though

kat-perez: https://github.com/OpenDevicePartnership/patina/issues/1163

PR #1117, opened 2025-11-22
Compile time component param validation and drop testing debug asserts [Rebase & FF]

### LINKED ISSUE #1054: [Tests]: Run Tests With Debug Asserts Disabled
### Feature Overview

Currently, we run tests with debug asserts enabled. However, we should test the production flows, particularly as many unit tests are testing proper error flows, which have debug_asserts in them. If we only test we hit a panic (I'm not aware if you can say only succeed if we hit a specific panic, but it may exist), we don't test the production case.

### Solution Overview

Disable debug_asserts in tests.

### Alternatives Considered

_No response_

### Urgency

Low

No maintainer feedback needed

### Anything else?

_No response_

### LINKED ISSUE #1108: [Bug]: Replace `debug_assert()` in `core_load_pe_image()` with return value

### Current Behavior

The `debug_assert!()`:

https://github.com/OpenDevicePartnership/patina/blob/9c452663df55b0e11c5cef200f0261274d841311/patina_dxe_core/src/image.rs#L595

Creates unnecessarily complex testing and inconsistent expectations for handling. The `debug_assert!()` can stay if desired but an error code should be returned as well.

### Expected Behavior

Errors are returned from the function when an error occurs.

### Steps To Reproduce

Review the code in `core_load_pe_image()` in [patina](https://github.com/OpenDevicePartnership/patina/tree/main)/[patina_dxe_core](https://github.com/OpenDevicePartnership/patina/tree/main/patina_dxe_core)/[src](https://github.com/OpenDevicePartnership/patina/tree/main/patina_dxe_core/src)
/image.rs.

### Build Environment

```markdown
N/A
```

### Version Information

```text
Tag: v16.0.0
```

### Urgency

Low

I will fix it

No maintainer feedback needed

### Anything else?

_No response_

makubacki:
## Description

In draft while still work-in-progress.

---

**patina_dxe_core/image: Remove debug asserts from unit tests**

Closes #1108 

- Return `EfiError::LoadError` for an invalid HII resource section.
- Remove complex logic in some unit tests for testing different
  conditions with debug asserts enabled or disabled.

---

**Remove `debug_assert!()` expectations in unit tests**

Debug assertions will no longer be enabled when running unit tests.

While the assertions can still be used in the code, they are removed
from unit test expectations.

---

**Add compile-time component parameter validation**

Replaces runtime parameter validation with compile-time checks
enforced with procedural macros and marker traits. This allows
the most common errors made using the pre-defined parameter types
to be detected very quickly during build.

Main Changes:

- Add `validate_params_macro` module with support for qualified type
  path normalization
- Introduce `#[component_impl]` for `impl` blocks and
  `#[component_entry_point]` for standalone functions (those outside
  of an `impl` block).
  - Note:
    - `impl` blocks that contain entry points for a component now
	  must have the `#[component_impl]` on the `impl` block.
	- Standalone component entry points (functions outside `impl`
	  blocks), must have the `#[component_entry_point]` attribute
- Update `IntoComponent` derive to enforce validation with marker
  traits
- Normalize type paths to handle qualified types
  (e.g. `Config` vs `patina::Config`)

---

**Disable debug assertions in the test profile**

Resolves #1054 

To have unit tests stay focused on production expectations and
reduce complexity in tests for managing debug assertions, debug
assertions are disabled in the `test` profile.

- Debug asserts may still be used but we're avoiding writing tests
  around them
- Non-debug asserts may be used and tested in unit tests

---

- [x] Impacts functionality?
- [ ] Impacts security?
- [x] Breaking change?
- [x] Includes tests?
- [x] Includes documentation?

## How This Was Tested

- `cargo make all`
- mdbook test

## Integration Instructions

- Todo (after taking the PR out of draft)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -82,7 +82,7 @@ opt-level = 3
 [profile.test]
 opt-level = 0
 debug = true
-debug-assertions = true
+debug-assertions = false
 overflow-checks = true
 lto = false
 incremental = false


sdk/patina_macro/src/lib.rs
```diff
@@ -77,6 +78,52 @@ pub fn component(item: proc_macro::TokenStream) -> proc_macro::TokenStream {
...
+/// For parameter validation on the entry_point function, use `#[component_entry_point]`
+/// on the entry_point function itself.
+///
+/// ## Usage
+///
+/// ```rust,ignore
+/// use patina::component::prelude::*;
+///
+/// #[component]
+/// pub struct MyComponent {
```
Javagedes, line 95: While I am interested in the interface of `#[component]` and `#[entry_point]` (rather than `component_entry_point`), I would like to see that as a replacement for `#[derive(IntoComponent)]` rather than a way to do compile time validation of a component.

I don't want to comment too much regarding this since there is no implementation of the validation part of this yet, so you may have found a cool way to do this reliably - But off the top of my head I'm not sure it's possible to implement in a safe / reliable way that works for all Parameters, even ones unknown to us.

If you go down the route you are appearing to go, which is to validate the param via a macro, then apply a trait if it is valid, Then I have some concerns: 

1. `Param` is a public trait, so anyone can create their own param implementation which may interact with any of our existing Params in unknown ways. Due to this, the above implementation is only ever going to be an optional add-on for validating components that only use Params we control
2. Since it is optional, we still have to maintain a way to validate parameters at runtime.
3. This ultimately means we have to maintain a way to validate param types two separate ways, in the trait implementation and in the macro. To my knowledge, this cannot be consolidated. And additionally, we have to keep these two validation ways in-sync over time.
4. Since we still need to maintain the validation at runtime, we have not solved the underlying issue of this PR, which is to remove the `debug_asserts` from the `initialize` function (This one is iffy, as again, this is in draft and you may not have gotten to it yet)
5. This will be a compilation cost every build even though the interface will probably never change, because its a macro running and doing validation, rather than just the compiler doing it's "compiler things"
6. This logic is easily duplicated with a single test that does `component.into_component().initialize(...)`.

_________________

ok with all of this said, these are my thoughts:

1. I really like your idea of attribute macros for `#[component]` and `#[entry_point]`. I would see this as a replacement for the `#[derive(IntoComponent)]`. The below is assuming we do that.
2. If you really want to have some sort of optional compile-time validation, I think a good way to do it is to add an extra option to `#[entry_point]` such as `validate` (e.g. `#[entry_point(validate)]` that will create a compiler error if (1) the parameters conflict, or (2) the parameter contains an unknown param type

Another big thing is that this PR is really about dropping `debug_asserts` in the code. As mentioned in (1), this implementation needs to be optional, and we need to continue to have a way to do runtime validation. Because this is optional, I think this work should be split up into two separate PRs. The first PR is removing the debug asserts, and the second is finding a valid solution to doing some type of const validation of the parameter.

I think for (1) we can very easily update the `initialize` method of the `Param` interface to return a `Result`. If there is a conflict, then `initialize` returns returns an error, and we do not register the component. Additionally, we can display the component in the `display_not_dispatched` method. Additionally, we `initialize` returning a `Result`, we can very easily have a single test for each component that validates it's interface:

```rust
#[test]
fn test_component_is_valid() {
    let mut storage = Storage::new();
    assert!(MyComponent::new().into_component().initialize(&mut storage).is_ok())
}
```

Heck, if you did not want to have any runtime cost, then we could split `initialize()` into two functions, `initialize()` and `validate()`, and not run `validate()` at runtime at all, and just say in our component docs that it is the responsibility of the component owner to ensure the component's interface is not UB. This can be done by having a single test that calls `assert!(MyComponent::new().into_component().validate(&mut storage).is_ok())`

_______

As for the const validation implementation, For it to be a complete implementation, it really needs to work on all parameters, even ones we do not know about. Since const trait methods are not a thing yet, that is probably going to have to be done via a manual implementation of static method implemented on the `Param` implementor type, but also outside of the `Param` trait. Something along the lines of:

```rust
    struct MyParam;

    // Auto generated via some `#[param]` macro or something, outside of the `Param` trait.
    pub const fn validate(BitSet) -> Result<BitSet, ()> { ... }

    impl Component for StructComponent {
        const IS_VALID: bool = {
             let bitset = BitSet::new();
             let Some(bitset) = Param1::validate(bitset) else {
                 return false;             
             }
             let Some(bitset) = Param2::validate(bitset) else {
                 return false;
             }
             true
         }

        fn initialize(&mut self, storage: &mut storage) {
           const _: () = { assert!(Self::IS_VALID) };
        }
    }
```
    
```

makubacki: @Javagedes and I spoke offline and clarified some details about the current PR and decided upon the following additional changes:

1. Replace `#[component_impl]` (for impl blocks) and `#[component_entry_point]` (for "standalone" functions) with a single `#[entry_point]` attribute for both where it is directly applied to the `entry_point()` method in an impl block.
2. Remove the ability for custom entry point function names and require the name to be `entry_point()` to simplify component analysis for humans and static analysis.
  - Attempt to eliminate the `#[entry_point]` on impl entry point methods if possible
3. Determine if marker traits can be eliminated in favor of directly causing the compilation failure in the macro with the appropriate span needed.

---

This will continue to follow the "hybrid" approach to parameter validation already implemented in the PR to statically validate known parameters at compilation time and rely upon runtime validation of custom parameter types.

---

This PR does not attempt to eliminate all `debug_assert!()`'s in the codebase but replace usages with more effective alternatives where possible and remove testing of debug assertions in unit tests. The changes for this unrelated to the component system will be broken out from this PR into a separate PR.

makubacki: Superseded by a simplified design in https://github.com/OpenDevicePartnership/patina/pull/1124. I'll leave this draft PR for reference for a few days, but the plan is to focus on that PR moving forward for compile-time component parameter validation.

makubacki:
Closing the PR. The design in https://github.com/OpenDevicePartnership/patina/pull/1124 was accepted and merged.

PR #1119, opened 2025-11-24
patina_dxe_core: Prevent installing duplicate child FVs

### LINKED ISSUE #1097: [Task]: Detect and report redundant FV HOBs
### Request Description

FVs are being dispatched twice on a platform. It is suspected that a FV and FV2 HOB are being processed independently leading to redundant dispatch. This issue tracks confirming the cause and then devising a solution in the Patina Readiness Tool and the Patina DXE Core to address the potential issue.

I will make the change

No maintainer feedback needed

### Anything else?

_No response_

makubacki:
## Description

Closes #1097 

Add logic to check if a firmware volume is already installed before
processing FV image files during DXE dispatch. This prevents
duplicate installation when both an FV HOB (describing a child FV
extracted earlier) and the parent FV containing that compressed child
are processed.

The check queries installed FVB (Firmware Volume Block) protocols to
determine if a firmware volume with matching name GUID already
exists.

`parse_hob_fvs()` installs FVB protocols for FV HOBs, so any child
FVs already described in FV HOBs should have FVB protocols installed.

---

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- `cargo make all`
- QEMU Q35 boot to a minimal Windows image
- Physical Intel platform boot to Windows

## Integration Instructions

- N/A - This changes behavior but should only prevent duplicate FVs from being processed.

### DIFF
--- a/patina_dxe_core/src/dispatcher.rs
+++ b/patina_dxe_core/src/dispatcher.rs
@@ -162,6 +162,58 @@ unsafe impl Send for DispatcherContext {}
 static DISPATCHER_CONTEXT: TplMutex<DispatcherContext> =
     TplMutex::new(efi::TPL_NOTIFY, DispatcherContext::new(), "Dispatcher Context");
 
+/// Checks if a firmware volume with the given name GUID is already installed.
+///
+/// Searches all installed firmware volume block (FVB) protocol handles to determine if
+/// any have a matching FV name GUID.
+///
+/// # Arguments
+/// * `fv_name_guid` - The firmware volume name GUID to check for
+///
+/// # Returns
+/// `true` if a firmware volume with the given name GUID is already installed
+/// `false` otherwise
+fn is_fv_already_installed(fv_name_guid: efi::Guid) -> bool {
+    // Get all handles with a FVB protocol
+    let fvb_handles = match PROTOCOL_DB.locate_handles(Some(firmware_volume_block::PROTOCOL_GUID)) {
+        Ok(handles) => handles,
+        Err(_) => return false,
+    };
+
+    // Check each FVB handle to see if it has the same FV name GUID
+    for handle in fvb_handles {
+        let Ok(ptr) = PROTOCOL_DB.get_interface_for_handle(handle, firmware_volume_block::PROTOCOL_GUID) else {
+            continue;
+        };
+        let fvb_ptr = ptr as *mut firmware_volume_block::Protocol;
+
+        // Safety: fvb_ptr is obtained from a valid handle that has a FVB protocol instance
+        // and the as_ref() call checks for null
+        let Some(fvb) = (unsafe { fvb_ptr.as_ref() }) else {
+            continue;
+        };
+
+        let mut fv_address: u64 = 0;
+        let status = (fvb.get_physical_address)(fvb_ptr, core::ptr::addr_of_mut!(fv_address));
+        if status.is_error() || fv_address == 0 {
+            continue;
+        }
+
+        // Safety: fv_address is checked for being non-zero above
+        let Ok(volume) = (unsafe { VolumeRef::new_from_address(fv_address) }) else {
+            continue;
+        };
+
+        if let Some(name) = volume.fv_name()
+            && name == fv_name_guid
+        {
+            return true;
+        }
+    }
+
+    false
+}
+
 pub fn dispatch() -> Result<bool, EfiError> {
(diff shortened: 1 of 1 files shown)


patina_dxe_core/src/dispatcher.rs
```diff
@@ -41,6 +42,8 @@ use crate::{
     tpl_mutex::TplMutex,
 };
 
+static HOB_LIST: Once<&'static hob::HobList<'static>> = Once::new();
```
joschock: General question: could you just reject any duplicate FvNameGuid without recourse to the hob list? I think that's what the EDK2 core does...

makubacki: That should work. I pushed the update and updated the PR description.

patina_dxe_core/src/dispatcher.rs
```diff
@@ -268,7 +324,37 @@ pub fn dispatch() -> Result<bool, EfiError> {
...
+
+                    if let Some(fv_name_guid) = fv_name_guid
+                        && is_fv_already_installed(fv_name_guid)
+                    {
+                        log::debug!(
+                            "Skipping FV file {:?} - FV with name GUID {:?} is already installed",
+                            guid_fmt!(candidate.file_name),
+                            guid_fmt!(fv_name_guid)
+                        );
+                        dispatch_attempted = true;
```
joschock: Is `dispatch_attempted` necessary in this case? I think it will just cause the dispatcher to go for another round, which might not be needed if nothing material has happened to the state (since rejecting the FV won't change the overall global dispatch context).

joschock (approved):

patina_dxe_core/src/dispatcher.rs
makubacki: I dropped it.

apop5 (approved):

patina_dxe_core/src/dispatcher.rs
```diff
@@ -162,6 +162,62 @@ unsafe impl Send for DispatcherContext {}
...
+fn is_fv_already_installed(fv_name_guid: efi::Guid) -> bool {
+    // Get all handles with a FVB protocol
+    let fvb_handles = match PROTOCOL_DB.locate_handles(Some(firmware_volume_block::PROTOCOL_GUID)) {
+        Ok(handles) => handles,
+        Err(_) => return false,
+    };
+
+    // Check each FVB handle to see if it has the same FV name GUID
+    for handle in fvb_handles {
+        let fvb_ptr = match PROTOCOL_DB.get_interface_for_handle(handle, firmware_volume_block::PROTOCOL_GUID) {
```
Javagedes: nit: I think the below is cleaner instead of match statements

```rust

if let Some(ptr) = PROTOCOL_DB.get_interface_for_handle(handle, firmware_volume_block::PROTOCOL_GUID) else {
    continue;
};

File: patina_dxe_core/src/dispatcher.rs [hunk 4]
```diff
@@ -162,6 +162,62 @@ unsafe impl Send for DispatcherContext {}
...
+            Ok(ptr) => ptr as *mut firmware_volume_block::Protocol,
+            Err(_) => continue,
+        };
+
+        // Safety: fvb_ptr is obtained from a valid handle that has a FVB protocol instance
+        // and the as_ref() call checks for null
+        let fvb = unsafe {
```
Javagedes: Same here - though not sure how the unsafe comes into play

```rust
if let Some(fvb) = fvb_ptr.as_ref() else {
    continue;
}

File: patina_dxe_core/src/dispatcher.rs [hunk 5]
```diff
@@ -162,6 +162,62 @@ unsafe impl Send for DispatcherContext {}
...
+        };
+
+        let mut fv_address: u64 = 0;
+        let status = (fvb.get_physical_address)(fvb_ptr, core::ptr::addr_of_mut!(fv_address));
+        if status.is_error() || fv_address == 0 {
+            continue;
+        }
+
+        // Safety: fv_address is checked for being non-zero above
+        let volume = match unsafe { VolumeRef::new_from_address(fv_address) } {
```
Javagedes:
```rust
if let Some(volume) = unsafe { VolumeRef::new_from_address(fv_address) } else {
  continue
}
```

PR #1120, opened 2025-11-25
patina_dxe_core: Clean up HOB init flow

makubacki:
## Description

The HOB list is relocated to DXE allocated memory shortly into DXE core entry. The Rust allocated HOB list is not binary compatible with the C HOB list. The Rust HOB list is a vector of references to HOB structures while the C hob list is a compact series of binary HOB structures.

While the Rust HOB list is sufficient for most use cases that simply need information from the HOBs in a more type safe and ergonomic interface, there are still use cases that require binary compatibility of the HOB list such as placement of the HOB into the system table where it is discovered by other modules, including C modules.

Today, this works because the HOB list pointer is directly passed to the system table initialization function where it is placed into the HOB list system table. This change relocates both HOB lists in the same place in the `init_memory()` function. Later DXE functionality consistently works on a relocated HOB list after `init_memory()` returns.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- `cargo make all`
- QEMU SBSA and Q35 EFI shell boot

## Integration Instructions

- N/A

### DIFF
--- a/patina_dxe_core/src/lib.rs
+++ b/patina_dxe_core/src/lib.rs
@@ -66,6 +66,8 @@
 
 extern crate alloc;
 
+use alloc::boxed::Box;
+
 mod allocator;
 mod component_dispatcher;
 mod config_tables;
@@ -110,7 +112,6 @@ use core::{
     str::FromStr,
 };
 
-use alloc::boxed::Box;
 use gcd::SpinLockedGcd;
 use memory_manager::CoreMemoryManager;
 use mu_rust_helpers::{function, guid::CALLER_ID};
@@ -123,7 +124,7 @@ use patina::{
         measurement::create_performance_measurement,
     },
     pi::{
-        hob::{HobList, get_c_hob_list_size},
+        hob::{HobList, get_pi_hob_list_size},
         protocols::{bds, status_code},
         status_code::{EFI_PROGRESS_CODE, EFI_SOFTWARE_DXE_CORE, EFI_SW_DXE_CORE_PC_HANDOFF_TO_NEXT},
     },
@@ -329,10 +330,11 @@ impl<P: PlatformInfo> Core<P> {
     /// The entry point for the Patina DXE Core.
     pub fn entry_point(&'static self, physical_hob_list: *const c_void) -> ! {
         assert!(self.set_instance(), "DXE Core instance was already set!");
+        assert!(!physical_hob_list.is_null(), "The DXE Core requires a non-null HOB list pointer.");
 
-        self.init_memory(physical_hob_list);
+        let relocated_hob_list = self.init_memory(physical_hob_list);
 
-        if let Err(err) = self.start_dispatcher(physical_hob_list) {
+        if let Err(err) = self.start_dispatcher(relocated_hob_list) {
             log::error!("DXE Core failed to start: {err:?}");
         }
 
@@ -357,7 +359,9 @@ impl<P: PlatformInfo> Core<P> {
     }
 
     /// Initializes the core with the given configuration, including GCD initialization, enabling allocations.
-    fn init_memory(&self, physical_hob_list: *const c_void) {
+    ///
+    /// Returns the relocated HOB list pointer that should be used for all subsequent operations.
+    fn init_memory(&self, physical_hob_list: *const c_void) -> *mut c_void {
         log::info!("DXE Core Crate v{}", env!("CARGO_PKG_VERSION"));
 
         GCD.prioritize_32_bit_memory(P::MemoryInfo::prioritize_32_bit_memory());
@@ -367,10 +371,6 @@ impl<P: PlatformInfo> Core<P> {
 
         // For early debugging, the "no_alloc" feature must be enabled in the debugger crate.
         // patina_debugger::initialize(&mut interrupt_manager);
-
-        if physical_hob_list.is_null() {
-            panic!("HOB list pointer is null!");
-        }
 
         gcd::init_gcd(physical_hob_list);
 
@@ -388,6 +388,20 @@ impl<P: PlatformInfo> Core<P> {
         PROTOCOL_DB.init_protocol_db();
         // Initialize full allocation support.
         allocator::init_memory_support(&hob_list);
+
+        // Relocate the PI Spec HOB list
+        //
+        // SAFETY: physical_hob_list is checked for null when it is accepted at the DXE core entry point.
+        let pi_hob_list_size = unsafe { get_pi_hob_list_size(physical_hob_list) };
+
+        // SAFETY: Creating a slice from the original PI HOB list pointer with the calculated size.
+        let pi_hob_slice = unsafe { core::slice::from_raw_parts(physical_hob_list as *const u8, pi_hob_list_size) };
+
+        // Leak a DXE allocated PI HOB list so it is available throughout the DXE phase.
+        let relocated_hob_list = Box::leak(pi_hob_slice.to_vec().into_boxed_slice()).as_mut_ptr().cast::<c_void>();
+
+        // Relocate the Rust HOB list
+        //
         // we have to relocate HOBs after memory services are initialized as we are going to allocate memory and
         // the initial free memory may not be enough to contain the HOB list. We need to relocate the HOBs because
         // the initial HOB list is not in mapped memory as passed from pre-DXE.
@@ -411,6 +425,8 @@ impl<P: PlatformInfo> Core<P> {
         component_dispatcher.add_service(CoreMemoryManager);
         component_dispatcher
             .add_service(cpu::PerfTimer::with_frequency(P::CpuInfo::perf_timer_frequency().unwrap_or(0)));
+
+        relocated_hob_list
     }
 
     /// Performs a combined dispatch of Patina components and UEFI drivers.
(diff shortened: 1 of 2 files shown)


patina_dxe_core/src/lib.rs
```diff
@@ -289,6 +291,8 @@ static __SELF: Once<NonZeroUsize> = Once::new();
 /// static CORE: Core<ExamplePlatform> = Core::new(NullSectionExtractor);
 /// ```
 pub struct Core<P: PlatformInfo> {
+    /// The original C HOB list pointer provided at entry point.
```
joschock: nitpick: this is not entirely accurate, since the hob list is relocated and most of the time is not going to be the pointer provided at entry. Maybe "Pointer to PI-spec compatible  HOB list for use in FFI" or something to that effect.

sdk/patina/src/pi/hob.rs
```diff
@@ -786,6 +786,8 @@ pub struct Capsule {
 
 /// Represents a HOB list.
 ///
+/// This is a parsed Rust representation of the HOB list that provides better type safety and ergonomics but does not
+/// have binary compatibility with the original C HOB list structure.
```
joschock:
```suggestion
/// have binary compatibility with the original PI Spec HOB list structure.
```

patina_dxe_core/src/lib.rs
```diff
@@ -388,6 +396,19 @@ impl<P: PlatformInfo> Core<P> {
         PROTOCOL_DB.init_protocol_db();
         // Initialize full allocation support.
         allocator::init_memory_support(&hob_list);
+
+        // Relocate the C HOB list
```
joschock:
```suggestion
        // Relocate the PI Spec HOB list
```

patina_dxe_core/src/lib.rs
```diff
@@ -289,6 +291,8 @@ static __SELF: Once<NonZeroUsize> = Once::new();
 /// static CORE: Core<ExamplePlatform> = Core::new(NullSectionExtractor);
 /// ```
 pub struct Core<P: PlatformInfo> {
+    /// The original C HOB list pointer provided at entry point.
+    c_hob_list: AtomicPtr<c_void>,
```
Javagedes: With my current understand of the PR, I don't agree with this change, and it is technically UB because we now have two owners of the hob list. The C DXE Core and the configuration table whereas prior to this, only the config table owned the actal relocated c hob list.

From what I can tell, the only thing this change does is prevent the need of passing the hob list to `init_memory`. and `initialize_system_table`. This is not a strong enough benefit to stash the hob list pointer in the Core as it also promotes future usage of it instead of the "rusty" hob list.

I think I need further explanation for the need of this change. Otherwise, I don't think I agree with this being merged.

makubacki: What is your actual concern? The HOB list was and still is a leaked box. The configuration table did not "own" the HOB list before, it simply exposed a global pointer to the leaked box buffer to all of DXE. The HOB list itself is read-only (not mutated in DXE) and the Rust allocated HOB list is NOT equivalent to the C HOB list because it is not binary compatible.

What is different about this that is concerning to you?

Javagedes: In a sense, the configuration table did own the hob list before, because it was the only object with any way to reference it. Now it can be referenced by the Core, which means basically anywhere in patina_dxe_core, and it continues to be accessible via the configuration table. I agree its read-only so not really UB. I redact that statement :)

As for my concern, this PR does not appear to be doing anything at all, other than storing it in the DXE Core for no discernable reason. Pre and Post this PR change, we appropriately pass the binary compatible hob list to the initialization functions that need it. All this change effectively does is stash the pointer in the core without ever using it (other then the instances it was already being used). Is there a future PR coming where you use the stashed binary compatible hob list after the Core initialization phase where we no longer have access to it?

makubacki: The change makes it obvious that the other HOB list is not binary compatible, this is important in real-world firmware usage and something that both @os-d and @Raymond-MS tried to use incorrectly in https://github.com/OpenDevicePartnership/patina/pull/1104 with the changes currently there causing actual UB because it attempts to cast a `HobList` the C HOB list.

makubacki: That said, they were rightfully confused and I think this change makes a properly relocated PI-spec compatible list more obvious which I prefer.

sdk/patina/src/pi/hob.rs
makubacki: I agree PI is better than "C". I'll rename the field and other structures to align with "pi".

patina_dxe_core/src/lib.rs
Javagedes: Personally, that feels like a lot of hand-holding. The `unsafe` keyword is required when dereferencing a pointer with the explicit requirement that the type you are de-referencing is the valid type. IMO the above is issue is mainly because @Raymond-MS is just starting to learn rust.

The fact that `HobList` is both a wrapper around a Vec and that there is no #`#[repr(C)]` should be enough to tell someone that this is not something you simply cast to. In my opinion, the right answer to this problem is to not intermingle C types and Rust types in patina::pi, as the pure rust types are technically not really PI spec compliant - or at a minimum add documentaiton

But adding an attribute to the Core just to make it clear that these two types are separate representations of the same thing, even though one type will never be used, seems heavy handed to me.

joschock: > it is technically UB because we now have two owners of the hob list

I don't think there is UB because there aren't two owners? There are two heap-side copies of the hob list. One that is a set of Rust-ish structures for use by rust consumers within the core, and one that preserves the PI spec layout for use by outside drivers that look for it in the config table. As far as I know, the two are not connected; so there shouldn't be an aliasing concerns there. 

As to whether that pointer to the traditional PI list needs to be stashed in the `Core`, I'm agnostic. But it does need to exist and be published in the config table.

makubacki: > Personally, that feels like a lot of hand-holding.

I don't think there is much hand holding here. For the code to work, it has to have a PI-spec style HOB list and a Rust style HOB list. The Rust HOB list is great but there is still required usage for the PI-spec list and that use requires it to remain valid throughout boot which means it also has to be relocated. All of that is already done, the only thing this really does is make the list more obvious in the `Core` structure. Since this is a DXE Core structure I think the benefit of this being obvious and maintained alongside the other list outweighs it being exposed.

> In my opinion, the right answer to this problem is to not intermingle C types and Rust types in patina::pi

Can you please clarify? The only change in `patina::pi` right now is a doc comment on existing code.

> But adding an attribute to the Core

Can you also clarify what is meant by "attribute" here? The change adds a new field to `Core`.

image

> just to make it clear that these two types are separate representations of the same thing

Depending on the definition of "thing" here, they are not the same thing. One is a PI-spec formatted HOB list and the other a vector of references. I assume you intend "same thing" to generically mean the same "HOB list". That's true, but what this change is clarifying is that there are separate copies and the storage format matters.

> even though one type will never be used

They're both used. The PI-spec HOB list is used to create the Rust HOB list and to serve as a PI-spec compatible HOB list throughout boot (directly referenced in the system table) and the Rust HOB list is used by Rust code.

Javagedes: > For the code to work, it has to have a PI-spec style HOB list and a Rust style HOB list.

For this part, the main thing is I don't want to use the c_hob_list in rust. Currently we don't. And with this change, we still don't. (Other then publishing the config table, which I agree absolutely must stay). But by having this field in the Core, in encourages possible future use of the PI binary compliant hob list within the rust core (just by being available), which I don't think we should do.

> Can you please clarify? The only change in patina::pi right now is a doc comment on existing code.

I just meant that if patina::pi module is meant for PI compliant structs and definitions, we maybe should not have rust representation of the HOB list in it, since it is not compliant with the PI binary layout of a HOB list.

> Can you also clarify what is meant by "attribute" here? The change adds a new field to Core.

Typo. I meant field, not attribute.

> They're both used. 

I see what you mean, But the way I see it, once we have allocations available, we should switch to the rusty vec hob list and use only that for everything else (The exception being the config table where we publish the PI binary compatible hob list). Any existence of the PI binary compatible HOB list post-allocations available encourages usage of it, which I don't think we should be doing.

makubacki: If `c_hob_list` is removed from `Core`, then I'd still like all memory relocation work to be done in `init_memory()`. 

Prior to this change, `init_memory()` returns with the two HOB lists in torn state, the PI spec HOB list returning to non-relocated buffer and the Rust HOB list in a relocated buffer.

Then, later users like `initialize_system_table()` have to know "This HOB list pointer I've been passed is actually the first time it's been handed out by the caller since memory was relocated, therefore, I need to perform the PI spec HOB list copy to a reallocated buffer". This implicitly encodes flow ordering assumptions in the code that don't need to be there.

This was primarily the type of confusion the original change was meant to simplify flow wise.

So, if we remove it, would you be okay then with `init_memory()` returning a pointer to the relocated HOB list and then that is passed to `start_dispatcher()` which will ultimately pass it to `initialize_system_table()`?

Javagedes: Yes I would be OK with that

makubacki: Pushed. I'll resolve this conversation.

os-d: I’ll add on that the relocation _has_ to occur before init_memory finishes (and even at a certain point within that) because PEI does not produce a memory allocation HOB for the HOB list itself. That memory gets reclaimed by DXE and can be allocated, meaning nothing later can point to the original location.

patina_dxe_core/src/lib.rs
```diff
@@ -388,6 +388,20 @@ impl<P: PlatformInfo> Core<P> {
         PROTOCOL_DB.init_protocol_db();
         // Initialize full allocation support.
         allocator::init_memory_support(&hob_list);
+
+        // Relocate the PI Spec HOB list
+        //
+        // SAFETY: physical_hob_list is checked for null when it is accepted at the DXE core entry point.
+        let pi_hob_list_size = unsafe { get_c_hob_list_size(physical_hob_list) };
```
joschock: nit: for consistency,

```suggestion
        let pi_hob_list_size = unsafe { get_pi_hob_list_size(physical_hob_list) };
```

makubacki: I was going to leave this to prevent modifying a public function in `patina::pi` but this is the only place it is used. I updated it.

makubacki:
Fyi: The most recent update just pushes some unit tests for `get_pi_hob_list_size()`.

PR #1121, opened 2025-11-26
patina_dxe_core: PI dispatcher subsystem / removal of statics

Javagedes:
## Description

This pull-request combines `fv.rs` and `dispatcher.rs` into a single module (`pi_dispatcher`) and removes the statics currently used in `fv.rs` and `dispatcher.rs`. This is a part of the work to remove all statics in `patina_dxe_core`.

This also begins the work of removing any `efiapi` functions from the public interface and only making them available via system tables.

There may be extra work in a separate PR to bring `image.rs` into the fold of `pi_dispatcher`.

### Reviewing This PR

This PR, while it has a lot of LoC Changes, This is mainly due to the re-organization of `dispatcher.rs` and `fv.rs`. The main things to review, that actually changed are:

1. `fv.rs` needs a complete review. This changed the internal data structure of the self-managed FV/FVB protocols and changed the interface for the Pure rust methods to use `NonNull`
2. A new struct defintion in`pi_dispatcher` called `PiDispatcher` that has some abstractions over the `DispatcherContext` and `FvData` (Now called `FvProtocolData`) code
3. Changes to all tests to create a static core and set it as the singleton instance for the test. This is unforuntely going to be necessary until we get the major systems like protocol db switched over.
4. Minor changes to `dxe_services.rs` to account for needing to call `Self::instance()` in some efiapi functions.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

CI continues to pass. Boot to shell in Q35.

## Integration Instructions

N/A

### DIFF
--- a/patina_dxe_core/src/pi_dispatcher/fv.rs
+++ b/patina_dxe_core/src/pi_dispatcher/fv.rs
@@ -174,12 +174,12 @@ impl<P: PlatformInfo> FvProtocolData<P> {
         let (lba_base_addr, block_size) = fv.lba_info(lba).map(|(addr, size, _)| (addr as usize, size as usize))?;
 
         let mut bytes_to_read = num_bytes;
-        if offset + bytes_to_read > block_size {
-            debug_assert!(offset + bytes_to_read <= block_size); // caller should not request to read beyond the block.
-            bytes_to_read = block_size - offset;
+        if offset.saturating_add(bytes_to_read) > block_size {
+            debug_assert!(offset.saturating_add(bytes_to_read) <= block_size); // caller should not request to read beyond the block.
+            bytes_to_read = block_size.saturating_sub(offset);
         }
 
-        let lba_start = (physical_address as usize + lba_base_addr + offset) as *mut u8;
+        let lba_start = (physical_address as usize).saturating_add(lba_base_addr).saturating_add(offset) as *mut u8;
         // Safety: lba_start is calculated from the base address of a valid FV, plus an offset and offset+num_bytes.
         // consistency of this data is guaranteed by checks on instantiation of the VolumeRef.
         // The FV data is expected to be 'static (i.e. permanently mapped) for the lifetime of the system.
@@ -428,7 +428,7 @@ impl<P: PlatformInfo> FvProtocolData<P> {
                         },
                         memory_type: MEMORY_MAPPED_IO,
                         starting_address: base_address,
-                        ending_address: base_address + fv.size(),
+                        ending_address: base_address.saturating_add(fv.size()),
                     },
                     end_dev_path: efi::protocols::device_path::End {
                         header: efi::protocols::device_path::Protocol {
@@ -794,9 +794,14 @@ impl<P: PlatformInfo> FvProtocolData<P> {
         //data. If so, copy to fill the destination buffer, and return
         //WARN_BUFFER_TOO_SMALL.
 
+        // We only want to copy the min(section_data.len(), local_buffer_size) bytes. If the local_buffer_size is
+        // larger than the section_data, we could inadvertently copy beyond the section_data slice.
+        let copy_size = core::cmp::min(section_data.len(), local_buffer_size);
+
         // Safety: local_buffer_ptr is either provided by the caller (and null-checked above), or allocated via allocate pool and
-        // is of sufficient size to contain the data.
-        let dest_buffer = unsafe { slice::from_raw_parts_mut(local_buffer_ptr as *mut u8, local_buffer_size) };
+        // is of sufficient size to contain the data. We copy the minimum of section_data.len() and local_buffer_size to ensure we do not
+        // copy beyond the bounds of either buffer.
+        let dest_buffer = unsafe { slice::from_raw_parts_mut(local_buffer_ptr as *mut u8, copy_size) };
         dest_buffer.copy_from_slice(&section_data[0..dest_buffer.len()]);
 
         //TODO: authentication status not yet supported.
@@ -848,7 +853,7 @@ impl<P: PlatformInfo> FvProtocolData<P> {
 
         // Safety: caller must provide valid pointers for key, file_type, name_guid, attributes, and size. They are null-checked above.
         unsafe {
-            (key as *mut usize).write_unaligned(local_key + 1);
+            (key as *mut usize).write_unaligned(local_key.saturating_add(1));
             name_guid.write_unaligned(file_name);
             if (fv_attributes & fvb::attributes::raw::fvb2::MEMORY_MAPPED) == fvb::attributes::raw::fvb2::MEMORY_MAPPED
             {
@@ -1860,5 +1865,105 @@ mod tests {
             }
         })
         .unwrap();
+    }
+
+    #[test]
+    fn test_fv_read_section_limits_read() {
+        // This test verifies that when the buffer size is larger than the section size, only the section
+        // size is copied, and the function returns SUCCESS.
+        let mut file = File::open(test_collateral!("DXEFV.Fv")).unwrap();
+        let mut fv: Vec<u8> = Vec::new();
+        file.read_to_end(&mut fv).expect("failed to read test file");
+
+        let fv = fv.leak();
+        let base_address: u64 = fv.as_ptr() as u64;
+        let parent_handle: Option<efi::Handle> = None;
+
+        test_support::with_global_lock(|| {
+            static CORE: MockCore = MockCore::new(CompositeSectionExtractor::new());
+            CORE.override_instance();
+            unsafe { test_support::init_test_gcd(None) };
+
+            let fv_interface = MockProtocolData::new_fv_protocol(parent_handle);
+            let fv_ptr = NonNull::from(&*fv_interface);
+
+            let metadata = Metadata::new_fv(fv_interface, base_address);
+            CORE.pi_dispatcher.fv_data.lock().fv_metadata.insert(fv_ptr.addr(), metadata);
+
+            let fv_ptr1 = fv_ptr.as_ptr();
+
+            // Safety: the following test code must uphold the safety expectations of the unsafe
+            // functions it calls. This unsafe section encompasses all of the logic for the remaining
+            // test since this is test code.
+            unsafe {
+                // Use a known file GUID from the test FV
+                let mut guid: efi::Guid = efi::Guid::from_fields(
+                    0x1fa1f39e,
+                    0xfeff,
+                    0x4aae,
+                    0xbd,
+                    0x7b,
+                    &[0x38, 0xa0, 0x70, 0xa3, 0xb6, 0x09],
+                );
+                let name_guid: *mut efi::Guid = &mut guid;
+
+                // First get the actual file size by passing null buffer
+                let mut actual_section_size: usize = 0;
+                let mut auth_status: u32 = 0;
+
+                let status = MockProtocolData::fv_read_section_efiapi(
+                    fv_ptr1,
+                    name_guid,
+                    19,
+                    0,
+                    &mut std::ptr::null_mut() as *mut *mut c_void,
+                    &mut actual_section_size,
+                    &mut auth_status,
+                );
+
+                assert_eq!(status, efi::Status::SUCCESS);
+                assert!(actual_section_size > 0, "Section size should be greater than 0");
+                // Test a buffer larger than the file section size
+                let larger_size = actual_section_size + 512; // 512 bytes larger than section
+
+                let layout = Layout::from_size_align(larger_size, 8).unwrap();
+                let mut buffer = alloc(layout) as *mut c_void;
+                assert!(!buffer.is_null(), "Memory allocation failed!");
+
+                // Fill the buffer with a pattern to check the copy
+                let buffer_slice = slice::from_raw_parts_mut(buffer as *mut u8, larger_size);
+                buffer_slice.fill(0xFE);
+
+                let mut buffer_size = larger_size;
+                let status = MockProtocolData::fv_read_section_efiapi(
+                    fv_ptr1,
+                    name_guid,
+                    19,
+                    0,
+                    &mut buffer,
+                    &mut buffer_size,
+                    &mut auth_status,
+                );
+
+                // 1. Status should be SUCCESS
+                assert_eq!(status, efi::Status::SUCCESS, "Expected SUCCESS when buffer is larger than file size");
+
+                // 2. buffer_size should be updated to the actual section size
+                assert_eq!(buffer_size, actual_section_size, "buffer_size should be updated to actual section size");
+
+                // 3. Verify only the section data was copied (rest should remain 0xFE)
+                let copied_data = slice::from_raw_parts(buffer as *const u8, actual_section_size);
+                let all_ff = copied_data.iter().all(|&b| b == 0xFE);
+                assert!(!all_ff, "Section data should have been copied to buffer (not all 0xFE)");
+
+                let remaining_data = slice::from_raw_parts(
+                    buffer.add(actual_section_size) as *const u8,
+                    larger_size - actual_section_size,
+                );
+                let all_ff_remaining = remaining_data.iter().all(|&b| b == 0xFE);
+                assert!(all_ff_remaining, "Remaining buffer beyond section size should remain unchanged (all 0xFE)");
+            }
+        })
+        .unwrap()
     }
 }


patina_dxe_core/src/pi_dispatcher/fv.rs
```diff
@@ -0,0 +1,1702 @@
...
+use crate::{
+    Core, PlatformInfo,
+    allocator::core_allocate_pool,
+    protocols::{PROTOCOL_DB, core_install_protocol_interface},
+    tpl_mutex,
+};
+
+/// A container for a FV or FVB protocol instance.
+///
+/// The protocols instances themselves are not used directly by rust code, but this container is used to manage the
```
joschock:
```suggestion
/// The protocol instances themselves are not used directly by rust code, but this container is used to manage the
```

patina_dxe_core/src/pi_dispatcher/fv.rs
```diff
@@ -0,0 +1,1702 @@
...
+
+    /// Creates a new Metadata instance for a FV protocol.
+    fn new_fv(protocol: Box<pi::protocols::firmware_volume::Protocol>, physical_address: u64) -> Self {
+        Self { protocol: Protocol::Fv(protocol), physical_address }
+    }
+}
+
+/// Stored protocol data for any FV/FVB protocols installed by the DXE core.
+pub(super) struct FvProtocolData<P: PlatformInfo> {
+    /// A map of installed FV/FVB protocol pointers and their physical addresses.
```
joschock:
```suggestion
    /// A map of installed FV/FVB protocol pointers (key) and the corresponding metadata (value).
```

patina_dxe_core/src/pi_dispatcher/fv.rs
```diff
@@ -0,0 +1,1702 @@
...
+impl<P: PlatformInfo> FvProtocolData<P> {
+    /// Returns the FV's physical address for the given protocol pointer, if it is in-fact a FV protocol.
+    #[inline(always)]
+    fn get_fv_address(&self, protocol: NonNull<pi::protocols::firmware_volume::Protocol>) -> Option<u64> {
+        let Some(Metadata { protocol: Protocol::Fv(_), physical_address }) =
+            self.fv_metadata.get(&protocol.cast::<c_void>().as_ptr())
+        else {
+            return None;
+        };
+        Some(*physical_address)
```
joschock:
```suggestion
        if let Some(Metadata { protocol: Protocol::Fv(_), physical_address }) = self.fv_metadata.get(&protocol.cast::<c_void>().as_ptr()) {
          Some(*physical_address)
        } else {
          None
        }
```

joschock: and similar suggestion for `get_fvb_address`

patina_dxe_core/src/pi_dispatcher/fv.rs
```diff
@@ -0,0 +1,1702 @@
...
+    }
+}
+
+// Safety: access to private global data is only through mutex guard, so safe to mark sync/send.
+unsafe impl<P: PlatformInfo> Send for FvProtocolData<P> {}
```
joschock: Consider using a more narrowly scoped unsafe impl on a newtype wrapper for `mut *c_void` in the BTree in FvProtocolData so that it's auto-`Send` so that you don't need to use one here? That's a bit more defensive, as it will catch someone adding a non-Send type to FvbProtocolData in the future.

patina_dxe_core/src/pi_dispatcher/fv.rs
```diff
@@ -0,0 +1,1702 @@
...
+        parent_handle: Option<efi::Handle>,
+        base_address: u64,
+    ) -> Result<efi::Handle, EfiError> {
+        let mut protocol = Self::new_fvb_protocol(parent_handle);
+
+        let protocol_ptr = protocol.as_mut() as *mut pi::protocols::firmware_volume_block::Protocol as *mut c_void;
+
+        let metadata = Metadata::new_fvb(protocol, base_address);
+
+        // save the protocol structure we're about to install in the private data.
```
joschock, line 334: since we have an invariant on that the fv/fvb protocols should never be removed once installed, we could consider adding a usage (e.g. via `open_protocol` or a new API on ProtocolDb) to ensure that attempting to uninstall them returns ACCESS_DENIED.

patina_dxe_core/src/pi_dispatcher/fv.rs
```diff
@@ -0,0 +1,1702 @@
...
+        _this: *const pi::protocols::firmware_volume::Protocol,
+        _information_type: *const efi::Guid,
+        _buffer_size: usize,
+        _buffer: *const c_void,
+    ) -> efi::Status {
+        efi::Status::UNSUPPORTED
+    }
+}
+
+//Firmware Volume device path structures and functions
```
joschock, line 892: should this move somewhere in sdk/patina/src/pi?

patina_dxe_core/src/pi_dispatcher/fv.rs
```diff
@@ -0,0 +1,1702 @@
...
+    type MockCore = Core<MockPlatformInfo>;
+    type MockProtocolData = FvProtocolData<MockPlatformInfo>;
+
+    #[test]
+    fn test_fv_init_core() {
+        test_support::with_global_lock(|| {
+            /* Start with Clearing Private Global Data, Please note that this is to be done only once
+             * for test_fv_functionality.
+             * In case other functions/modules are written, clear the private global data again.
+             */
```
joschock: I'm not sure what this block comment is referring to? Can you clarify where the clearing is happening?

patina_dxe_core/src/pi_dispatcher/fv.rs
```diff
@@ -0,0 +1,1702 @@
...
+                        fv_ptr1,
+                        guid_ref_invalid_ref,
+                        6,
+                        10,
+                        &mut buffer_valid3 as *mut *mut c_void,
+                        buffer_valid_size3,
+                        auth_valid_p,
+                    );
+
+                    /* Valid guid case - panicing, debug this further, for now comment*/
```
joschock, line 1513: maybe file an issue?

patina_dxe_core/src/pi_dispatcher/fv.rs
```diff
@@ -0,0 +1,1702 @@
...
+                    MockProtocolData::fv_read_file_efiapi(
+                        fv_ptr1,
+                        guid_valid_ref,
+                        std::ptr::null_mut(),
+                        buffer_valid_size3,
+                        found_type_ref,
+                        file_attributes,
+                        auth_valid_p,
+                    );
+                    /* Raise Bug for this case , case when Buffer size is 0 and buffer not NULL. last block*/
```
joschock: issue?

patina_dxe_core/src/pi_dispatcher/section_decompress.rs
```diff
@@ -0,0 +1,77 @@
...
+    FirmwareFileSystemError,
+    section::{SectionExtractor, SectionHeader},
+};
+
+/// Section extractor that provides UEFI decompression, with an optional additional [SectionExtractor] implementation.
+#[derive(Default)]
+pub(super) struct CoreExtractor<E: SectionExtractor>(E);
+
+impl<E: SectionExtractor> CoreExtractor<E> {
+    /// Creates a new [CoreExtractor] with no additional extractor.
```
joschock:
```suggestion
    /// Creates a new [CoreExtractor] with the specified additional extractor.
```

patina_dxe_core/src/dxe_services.rs
```diff
@@ -370,130 +367,138 @@ extern "efiapi" fn get_io_space_map(
...
+        let dxe_system_table_ptr = &dxe_system_table as *const dxe_services::DxeServicesTable;
+        let crc32 = unsafe {
+            crc32fast::hash(from_raw_parts(
+                dxe_system_table_ptr as *const u8,
+                mem::size_of::<dxe_services::DxeServicesTable>(),
+            ))
+        };
+        dxe_system_table.header.crc32 = crc32;
+
+        let dxe_system_table = Box::new_in(dxe_system_table, &EFI_RUNTIME_SERVICES_DATA_ALLOCATOR);
```
joschock: Can this be Boot Services Data? I'm not sure it needs to be runtime. I think it is in the global data section of DxeMain in the EDK2 core which would be BSD.

patina_dxe_core/src/pi_dispatcher.rs
```diff
@@ -69,6 +77,343 @@ const ALL_ARCH_DEPEX: &[Opcode] = &[
...
+    #[coverage(off)]
+    pub fn install_firmware_volumes_from_hoblist(
+        &self,
+        hob_list: &patina::pi::hob::HobList,
+    ) -> Result<(), efi::Status> {
+        self.fv_data.lock().install_firmware_volumes_from_hoblist(hob_list)
+    }
+
+    /// Performs a single dispatch iteration.
+    pub fn dispatch(&self) -> Result<bool, EfiError> {
```
joschock, line 136: I assume no significant functional changes to this routine? I did not review it closely.

Javagedes: Correct, no functional changes at all other then swapping `DISPATCHER_CONTEXT.lock()` for `self.dispatcher_context.lock()`

patina_dxe_core/src/pi_dispatcher/fv.rs
Javagedes: This was from the existing implementation. All tests are from the original `fv.rs` and remain unchanged except for the slightly modifications necessary to call the functions.

patina_dxe_core/src/pi_dispatcher/fv.rs
Javagedes: This was from the existing implementation. All tests are from the original `fv.rs` and remain unchanged except for the slightly modifications necessary to call the functions.

patina_dxe_core/src/pi_dispatcher/fv.rs
Javagedes: An artifact from the original `fv.rs` that I did not clean up. deleting :)

patina_dxe_core/src/pi_dispatcher/fv.rs
Javagedes: It cannot because these custom implementations continue to rely on accessing the global Core singleton for validation the `this` instance is one we manage.

To move outside of patina, we would need to stash all the necessary data behind the protocol and make it truly standalone. However, as we discussed, maintaining the FV metadata in the core is added security allowing us to validate that the protocol methods are being called on a valid instance of `this`.

Javagedes: Oh wait, you are referring to the device path code, not the protocol implementations. 

Yes, it can probably be moved. Let me look into it.

patina_dxe_core/src/pi_dispatcher/fv.rs
Javagedes: @joschock I have a working implementation to use a NonZeroUsize (pointer address) for the key rather than a c_void. Do you have any concerns with this before I push it as the solution to this issue?

joschock: I think that should work; main thrust of the comment is not to scope more into the unsafe Send than necessary to preserve the check for future modifications to `FvProtocolData`. Agnostic on the particulars of the mechanism.

patina_dxe_core/src/pi_dispatcher/fv.rs
joschock: Agreed, but if we don't have an issue tracking it, we probably should. I'm happy to file if that would help.

patina_dxe_core/src/pi_dispatcher/fv.rs
Javagedes: switched to NonZero usize for the key. Additionally, there was an issue that the FVB / FV protocols contain a handle (*mut c_void). I implemented Send on those protocols as handles are merely opaque pointers used for identification.

patina_dxe_core/src/pi_dispatcher/fv.rs
Javagedes: Just added the test back and it is no longer panicking. So I'll keep it enabled and will close this :)

patina_dxe_core/src/pi_dispatcher/fv.rs
Javagedes: I added this test back and this code does not panic. I will keep it now and close this.

patina_dxe_core/src/pi_dispatcher/fv.rs
Javagedes: This is a change greater than the scope of this PR, but I agree its valid. Mind raising an issue for it?

joschock: https://github.com/OpenDevicePartnership/patina/issues/1126

patina_dxe_core/src/pi_dispatcher/fv.rs
Javagedes: Just kidding. This test panics during CI. Will create an issue.

Javagedes: https://github.com/OpenDevicePartnership/patina/issues/1127

patina_dxe_core/src/lib.rs
```diff
@@ -616,28 +641,36 @@ fn call_bds() -> ! {
...
-        static CORE: Core<MockPlatformInfo> =
-            Core::<MockPlatformInfo>::new(patina_ffs_extractors::NullSectionExtractor);
-        static CORE2: Core<MockPlatformInfo> =
-            Core::<MockPlatformInfo>::new(patina_ffs_extractors::NullSectionExtractor);
-        assert!(CORE.set_instance());
-
-        if NonNull::from_ref(&CORE) != NonNull::from_ref(Core::<MockPlatformInfo>::instance()) {
-            panic!("CORE instance mismatch");
-        }
+        with_global_lock(|| {
```
cfernald, line 671: Since this is new consider serial_test with something like `core_instance` instead of the global lock. Or will this conflict with too many tests using global lock today?

patina_dxe_core/src/pi_dispatcher/fv.rs
```diff
@@ -0,0 +1,1705 @@
...
+    /// A map of installed FV/FVB protocol pointers (key) and the corresponding metadata (value).
+    fv_metadata: BTreeMap<NonZeroUsize, Metadata>,
+    /// A marker for accessing the singleton `Core` instance in a UEFI protocol method.
+    _platform_info: core::marker::PhantomData<P>,
+}
+
+impl<P: PlatformInfo> FvProtocolData<P> {
+    /// Returns the FV's physical address for the given protocol pointer, if it is in-fact a FV protocol.
+    #[inline(always)]
+    fn get_fv_address(&self, protocol: NonNull<pi::protocols::firmware_volume::Protocol>) -> Option<u64> {
```
cfernald, line 79: For my education, what is the reason for `get_fv_address` and `get_fvb_address` if they return the same thing?

patina_dxe_core/src/pi_dispatcher/fv.rs
```diff
@@ -0,0 +1,1705 @@
...
+        let metadata = Metadata::new_fvb(protocol, base_address);
+
+        // save the protocol structure we're about to install in the private data.
+        self.fv_metadata.insert(protocol_ptr.addr(), metadata);
+
+        // install the protocol and return status
+        core_install_protocol_interface(
+            handle,
+            pi::protocols::firmware_volume_block::PROTOCOL_GUID,
+            protocol_ptr.as_ptr(),
```
cfernald, line 341: I'm a bit hesitant to keep protocols in a `Box` because we convert it to pointers and use those pointers outside of the strict lifetime of the Box. That is, the memory safety and smart pointers work against us because we may drop the Box unintentionally while the pointer could still be held by the DB or a driver. 

I think we either need to `leak`/`into_raw` the box, or in the extreme case build/use a separate smart pointer that the DB is aware of.

patina_dxe_core/src/pi_dispatcher/fv.rs
```diff
@@ -0,0 +1,1705 @@
...
+        let metadata = Metadata::new_fv(protocol, base_address);
+
+        // save the protocol structure we're about to install in the private data.
+        self.fv_metadata.insert(protocol_ptr.addr(), metadata);
+
+        // install the protocol and return status
+        core_install_protocol_interface(handle, pi::protocols::firmware_volume::PROTOCOL_GUID, protocol_ptr.as_ptr())
+    }
+
+    /// Installs both the FVB and FV protocols for a firmware volume at the specified base address.
```
cfernald, line 365: needs a safety comment for caller

patina_dxe_core/src/pi_dispatcher/fv.rs
```diff
@@ -0,0 +1,1705 @@
...
+        &self,
+        handle: Option<efi::Handle>,
+        base_address: u64,
+    ) -> Result<efi::Handle, EfiError> {
+        // Safety: caller must ensure that base_address is valid.
+        let fv = unsafe { VolumeRef::new_from_address(base_address) }?;
+
+        let device_path_ptr = match fv.fv_name() {
+            Some(fv_name) => {
+                //Construct FvPiWgDevicePath
```
cfernald: super nit: space before comment here and elsewhere in function

patina_dxe_core/src/pi_dispatcher/fv.rs
```diff
@@ -0,0 +1,1705 @@
...
+                    mem_map_device_path: MemMapDevicePath {
+                        header: efi::protocols::device_path::Protocol {
+                            r#type: efi::protocols::device_path::TYPE_HARDWARE,
+                            sub_type: efi::protocols::device_path::Hardware::SUBTYPE_MMAP,
+                            length: [
+                                (mem::size_of::<MemMapDevicePath>() & 0xff) as u8,
+                                ((mem::size_of::<MemMapDevicePath>() >> 8) & 0xff) as u8,
+                            ],
+                        },
+                        memory_type: 11, //EfiMemoryMappedIo not defined in r_efi
```
cfernald: defined as MEMORY_MAPPED_IO here: https://github.com/r-efi/r-efi/blob/97b55bed1c2c91dcbf787674849f05337ff80b33/src/system.rs#L331

patina_dxe_core/src/pi_dispatcher/fv.rs
```diff
@@ -0,0 +1,1705 @@
...
+        // install the protocol and return status
+        core_install_protocol_interface(handle, efi::protocols::device_path::PROTOCOL_GUID, device_path_ptr)
+    }
+}
+
+// FV / FVB EFIAPI compliant protocol method implementations.
+#[coverage(off)]
+impl<P: PlatformInfo> FvProtocolData<P> {
+    /// EFIAPI compliant FVB protocol GetAttributes method.
+    extern "efiapi" fn fvb_get_attributes_efiapi(
```
cfernald, line 457: consider making all the efiapi calls here unsafe, which we should probably almost always be doing to highlight they are not intended for internal use and are not safe to do so. Especially since it no longer releases us from having to declare unsafes internally.

patina_dxe_core/src/pi_dispatcher/fv.rs
Javagedes: Part of the safety of this function is in the destructuring of the Metadata and requiring that the enum we cached for your pointer (either `Protocol::Fvb(_)` or `Protocol::Fv(_)`) matches the protocol type you give us here. You could easily accidently cast the protocol to a FV instead of an FVB (or vise versa) and without this, we would give you the address. You would then continue down the path and cast the pointer to the wrong type.

```rust
if let Some(Metadata { protocol: Protocol::Fvb(_), physical_address })

// vs

if let Some(Metadata { protocol: Protocol::Fv(_), physical_address })
```

patina_dxe_core/src/pi_dispatcher/fv.rs
Javagedes: These protocols were in a box prior to this implementation.

That said, I'm fairly agnostic, and I am happy to leak them and store them as pointers if you prefer.

As devil's advocate, I'm not truly concerned though, as there is no API we expose to remove entries from btreemap of the protocols. And other than for testing, this ends up being a field of the Core, which is static.

John actually had the opposite concern of this, which is needing to prevent the removal of the protocol from the protocol_db: https://github.com/OpenDevicePartnership/patina/issues/1126

patina_dxe_core/src/pi_dispatcher/fv.rs
Javagedes: That would require changing the interface of the protocol struct definition in `pi::protocols::firmware_volume<_block>::Protocol` to expect an unsafe function. I'm fine with doing that, but do we want to go down that route? At some point we will hit r_efi protocol definitions and won't be able to do that.

Also we have no callers of these efiapi functions in rust code, so I'm not sure how much we benefit from marking them unsafe. But again, happy to go that route.

cfernald: Not a big deal, just good form IMO since these all take raw pointers. I think where-ever we own the define we should declare unsafe. I'm not going to hold up on this though, so up to you.

patina_dxe_core/src/pi_dispatcher/fv.rs
cfernald: I think John's concern is actually in line with mine, or at least exacerbates it.

I would say leak it, because even if we dont expose such a function we are setting ourselves up for failure if we create a stored pointer on something that isn't explicitly a static lifetime.

patina_dxe_core/src/lib.rs
Javagedes: The only reason we have this as a static and calling `set_instance` is because we still rely on subsystems that are globally static (e.g. the PROTOCOL_DB). Once these statics are properly instanced in the core struct, then we won't need any of these to be static at all.

The only time that is true is if we want to test the "efiapi" functions. But almost all of those are lightweight wrappers around the instanced rust functions, so that should be far and fiew in-between. In those instances, I do think we will switch to serial_test.

makubacki:
@Javagedes, just wanted to mention that I'll review on Monday 12/1 after the Thanksgiving holiday.

patina_dxe_core/src/dxe_services.rs
makubacki: I think it can be as well. But, I suggest calling it `dxe_services_system_table`. That emphasizes the table contains DXE Services (GCD and dispatcher services) which should not be used outside the DXE Foundation during runtime.

patina_dxe_core/src/pi_dispatcher/fv.rs
joschock: I think the comment is stale now?

patina_dxe_core/src/dxe_services.rs
Javagedes: Switched to a normal `Box` and `Box::into_raw`, as the global allocator is the BOOT_SERVICES_DATA_ALLOCATOR

patina_dxe_core/src/pi_dispatcher/fv.rs
Javagedes: I'm not following through with this. If we change this protocol then realistically we should be changing every protocol's function interface to be unsafe. While I agree that should be done, it can't. Not while we use r-efi.

Feel free to create an issue to evaluate making all efiapi functions unsafe if you wish. I would vote for that. But it would mean dropping r-efi :)

patina_dxe_core/src/pi_dispatcher/fv.rs
Javagedes: Moved this to a standalone issue as the device path code is spread across multiple locations and is a bit messy. The entire usage needs to be evaluated.

#1148

patina_dxe_core/src/pi_dispatcher/fv.rs
joschock: It'd be nice to write a "missing extern safety docs" lint ala the "missing_safety_docs" lint, but I don't know that clippy supports an easy extension for that.

Javagedes: I think the only way to do that would be to replace `extern "efiapi"` with an attribute macro and parse all #[doc = "..."] lines for an appropriate `## Safety` section. That is assuming that we get acesss to the #[doc = "..."] statements. But it would look something like this:

```rust
#[efiapi]
fn scary_efi_api_function( )
```

which turns into (while also checking for safety comment)

```rust
extern "efiapi" fn scary_efi_api_function()
```

But that seems a bit heavy handed.

patina_dxe_core/src/pi_dispatcher/fv.rs
```diff
@@ -0,0 +1,1864 @@
...
+        // its invariants - like not removing fv once installed - are upheld).
+        let fv = unsafe { VolumeRef::new_from_address(physical_address) }?;
+        let Ok(lba) = lba.try_into() else {
+            return Err(EfiError::InvalidParameter);
+        };
+
+        let (lba_base_addr, block_size) = fv.lba_info(lba).map(|(addr, size, _)| (addr as usize, size as usize))?;
+
+        let mut bytes_to_read = num_bytes;
+        if offset + bytes_to_read > block_size {
```
makubacki: It'd be safer to use checked arithmetic here.

patina_dxe_core/src/pi_dispatcher/fv.rs
```diff
@@ -0,0 +1,1864 @@
...
+        }
+
+        //copy bytes to output. Caller-provided buffer may be shorter than section
+        //data. If so, copy to fill the destination buffer, and return
+        //WARN_BUFFER_TOO_SMALL.
+
+        // Safety: local_buffer_ptr is either provided by the caller (and null-checked above), or allocated via allocate pool and
+        // is of sufficient size to contain the data.
+        let dest_buffer = unsafe { slice::from_raw_parts_mut(local_buffer_ptr as *mut u8, local_buffer_size) };
+        dest_buffer.copy_from_slice(&section_data[0..dest_buffer.len()]);
```
makubacki, line 805: This creates `dest_buffer` with `local_buffer_size` but `local_buffer_size` is not updated to `section.data.len()` in the case `local_buffer_ptr` is non-null (like it is when `local_buffer_ptr` is null). In that case, `dest_buffer.len()` will be the user-provided buffer size and probably not what we want to use as the slice length for the section.

My understanding is this could be clipped at the `section_data.len()` in all cases for the copy itself. Something like:

```suggestion
        let copy_len = dest_buffer.len().min(section_data.len());
        dest_buffer[0..copy_len].copy_from_slice(&section_data[0..copy_len]);
```

makubacki (approved):

patina_dxe_core/src/pi_dispatcher/fv.rs
Javagedes: switched to saturating_add throughout the file

patina_dxe_core/src/pi_dispatcher/fv.rs
Javagedes: We cannot quite do that. `copy_from_slice` requires both buffers be the exact same size, or it will panic. If you want to make this update, then we need to adjust `local_buffer_len` to instead be `section_data.len()`. Something like this:

```rust
// We only want to copy the min(section_data.len(), local_buffer_size) bytes. If the local_buffer_size is
// larger than the section_data, we could inaddvertently copy beyond the section_data slice.
let copy_size = core::cmp::min(section_data.len(), local_buffer_size);

// Safety: local_buffer_ptr is either provided by the caller (and null-checked above), or allocated via allocate pool and
// is of sufficient size to contain the data. We copy the minimum of section_data.len() and local_buffer_size to ensure we do not
// copy beyond the bounds of either buffer.
let dest_buffer = unsafe { slice::from_raw_parts_mut(local_buffer_ptr as *mut u8, copy_size) };
dest_buffer.copy_from_slice(&section_data[0..dest_buffer.len()]);
```

makubacki: That's fine. The main point would be to take the min.

Javagedes: OK, I wrote a test that validated that we panic with a buffer overrun if someone gives us a buffer_size_len greater than the size of the section. I added the above code and validated that the test now passes. Will push it soon

Javagedes: pushed.

Issue #1122, opened 2025-11-26
[Task]: Patina component parameter validation changes


makubacki:
### Feature Overview

Today, Patina component parameters are validated in debug builds (and unit tests) with debug assertions but not in release builds.

We're making a few changes to component parameter validation, some of which build on others, so they're collectively listed here.

1. Remove debug asserts as a solution for component parameter validation error handling.
2. Due to (1) and broader changes in the codebase to eliminate testing debug asserts in unit tests, do not test panics from debug asserts in unit tests
3. Perform component parameter validation for Patina defined parameter types at compile-time
   - These are the most common types used and the most likely to be used incorrectly as new component authors will be using common, existing types, not defining new custom parameter types
4. Continue to allow components to provide runtime validation of custom component parameter types
5. Have custom component parameter type validation failures result in the component not being dispatched and the failure logged in the existing summary of components not dispatched in the log with the specific component parameter validation failure messages

### Solution Overview

See above

### Urgency

Low

### Are you going to implement the feature request?

I will implement the feature

### Do you need maintainer feedback?

No maintainer feedback needed

makubacki:
I've assigned to myself for now as I need to complete 1 to 3. Then, @Javagedes will be assigned to complete 4 & 5.

makubacki:
Changes for 3 & 4 are in https://github.com/OpenDevicePartnership/patina/pull/1124

Javagedes:
5 via #1153

Javagedes:
#1153 has been merged, closing this.

PR #1123, opened 2025-11-26
Remove debug asserts in tests [Rebase & FF[

### LINKED ISSUE #1054: [Tests]: Run Tests With Debug Asserts Disabled

### LINKED ISSUE #1108: [Bug]: Replace `debug_assert()` in `core_load_pe_image()` with return value

makubacki:
## Description

---

Note: The debug asserts in the component parameter validation code in the first phase of enforcing strong parameter validation as described in https://github.com/OpenDevicePartnership/patina/issues/1122.

---

**patina_dxe_core/image: Remove debug asserts from unit tests**

Closes #1108 

- Return `EfiError::LoadError` for an invalid HII resource section.
- Remove complex logic in some unit tests for testing different
  conditions with debug asserts enabled or disabled.

---

**Remove debug_assert!() expectations in unit tests**

Debug assertions will no longer be enabled when running unit tests.

While the assertions can still be used in the code, they are removed
from unit test expectations.

---

**Disable debug assertions in the test profile**

Closes #1054 

To have unit tests stay focused on production expectations and
reduce complexity in tests for managing debug assertions, debug
assertions are disabled in the `test` profile.

- Debug asserts may still be used but we're avoiding writing tests
  around them
- Non-debug asserts may be used and tested in unit tests

---

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- `cargo make all`

## Integration Instructions

- N/A

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -82,7 +82,7 @@ opt-level = 3
 [profile.test]
 opt-level = 0
 debug = true
-debug-assertions = true
+debug-assertions = false
 overflow-checks = true
 lto = false
 incremental = false
--- a/sdk/patina/src/component.rs
+++ b/sdk/patina/src/component.rs
@@ -48,7 +48,8 @@
 //!
 //! Note: Certain combinations of parameters may lead to undefined behavior as they can allow multiple mutable
 //! accesses to the same data. Each parameter type checks for conflicts with previously registered accesses, but
-//! **ONLY** on debug builds. In release builds, these checks are disabled for performance and size reasons.
+//! **ONLY** on debug builds (and not during testing). In release builds, these checks are disabled for performance
+//! and size reasons. In test builds, they are disabled to focus testing on functionality rather than debug assertions.
 //!
 //! ### `Param` types
 //!
--- a/sdk/patina/src/component/params.rs
+++ b/sdk/patina/src/component/params.rs
@@ -764,9 +764,6 @@ mod tests {
     use super::*;
 
     #[test]
-    #[should_panic(
-        expected = "ConfigMut<usize> in component patina::component::params::tests::test_two_mutable_config_access_to_same_type_fails::TestComponent conflicts with a previous ConfigMut<usize> access."
-    )]
     fn test_two_mutable_config_access_to_same_type_fails() {
         #[derive(IntoComponent)]
         struct TestComponent;
@@ -783,9 +780,6 @@ mod tests {
     }
 
     #[test]
-    #[should_panic(
-        expected = "Config<usize> in component patina::component::params::tests::test_mutable_and_immutable_config_access_to_same_type_fails1::TestComponent conflicts with a previous ConfigMut<usize> access."
-    )]
     fn test_mutable_and_immutable_config_access_to_same_type_fails1() {
         #[derive(IntoComponent)]
         struct TestComponent;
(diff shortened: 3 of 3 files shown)


patina_dxe_core/src/allocator.rs
```diff
@@ -986,7 +986,7 @@ fn process_hob_allocations(hob_list: &HobList) {
 
         if (stack_address == 0) || (stack_length == 0) {
```
Javagedes: if we just turn these into some asserts, We would be able to reduce the "layers" by 1, because we would remove the else statement, which tabs it over by one.

```rust
assert_ne!(stack_address, 0, "Invalid Stack Configuration: stack base address must be non-zero");
assert_ne!(stack_length, 0, "Invalid Stack Configuration: stack length must be non-zero");
```

It would show an error like this:

```
thread 'main' panicked at 'Invalid Stack Configuration: stack base address must be non-zero
assertion failed: `(left != right)`
  left: `0`,
 right: `0`', src/main.rs:10:5
```

Another option (probably reduces binary size, less logging)

```rust
assert!(stack_address != 0 && stack_length != 0,
    "Invalid Stack Configuration: Stack base address {stack_address:#X} for len {stack_length:#X}");
```

patina_dxe_core/src/image.rs
```diff
@@ -2410,11 +2411,8 @@ mod tests {
         //
         // We craft a malformed PE image with a section virtual_size that will cause
         // align_up() to overflow when aligning to section_alignment.
-        //
-        // Note: In debug builds, this hits a debug_assert!(false) which panics.
-        // The panic is caught by with_global_lock(), so we check for that.
 
-        let result = test_support::with_global_lock(|| {
+        let _ = test_support::with_global_lock(|| {
```
Javagedes: `with_global_lock` catches any panics, including test panics like `assert_eq!(a, b)`. so with you catching and not reading the result, I don't think this test can fail anymore. At a minimum I think you need to unwrap the result of `with_global_lock`

patina_dxe_core/src/image.rs
```diff
@@ -2488,7 +2470,7 @@ mod tests {
         // the debug_assert!(false) panic path. In release mode, this particular error
         // case doesn't result in a load failure.
 
-        let result = test_support::with_global_lock(|| {
+        let _ = test_support::with_global_lock(|| {
```
Javagedes: Same here.

vineelko (approved):

patina_dxe_core/src/allocator.rs
makubacki: Good suggestion. I just used a single assert.

PR #1124, opened 2025-11-26
Add compile-time component parameter validation

makubacki:
## Description

Today, component parameters are validated as follows:

- In debug build with debug asserts
- In test builds with debug asserts
- Not in release builds

This change both simplifies overall component definition to make their construction more consistent and easier to follow in the codebase while also introducing compile-time parameter validation for Patina-defined parameter types.

Patina-defined parameter (`Param`) types include types like:

- `Config<T>`
- `ConfigMut<T>`
- `Hob<T>`
- `Service<T>`
- etc.

This allows the most common errors made using the pre-defined parameter types to be detected very quickly during build. Custom parameter types will be validated at runtime. A follow up change will prevent execution of components with invalid custom parameter combinations and log an error message alongside the table of components not dispatched.

---

The [trybuild](https://crates.io/crates/trybuild) crate is used in `dev-dependencies`. It provides a clean way to test that invalid parameter combinations have the expected compilation error messages. The tests are placed in a `ui` directory following convention described in the `trybuild` docs that these tests are focused on a user's interaction with the code.

---

The main changes are:

- Entry points are now always called `entry_point` and do not require any special annotation
- "Standalone" (free standing) function entry points are no longer allowed and entry points must always be an `entry_point` method in an `impl` block for the type
- Only a single `#[component]` attribute is needed on an `impl` block for a component.

The is intended to make components easier to understand and review while taking advantage of the changes to provide quick compile-time feedback on component construction.

---

The `#[component]` attribute is applied onto the `impl` block specifically because this allows visibility into the `impl` block to check if an `entry_point` method is present which is a requirement. Similarly, it provides visibility for any potential future component special methods to be validated.

- [x] Impacts functionality?
- [ ] Impacts security?
- [x] Breaking change?
- [x] Includes tests?
- [x] Includes documentation?

## How This Was Tested

- `cargo make all`
- QEMU Q35 & SBSA boot to EFI shell

## Integration Instructions

- See the component documentation updates in the change

### DIFF
--- a/sdk/patina_macro/src/validate_params_macro.rs
+++ b/sdk/patina_macro/src/validate_params_macro.rs
@@ -36,7 +36,7 @@ use syn::{FnArg, ImplItem, ItemFn, ItemImpl, Pat, Type, TypePath, parse2, spanne
 /// ## Validation Rules
 ///
 /// - Impl block must have an `entry_point` method
-/// - Entry point must have `self`, `&self`, or `&mut self` as first parameter
+/// - Entry point must have `self`, `mut self`, `&self`, or `&mut self` as the first parameter
 /// - No duplicate `ConfigMut<T>` parameters with the same type T
 /// - Cannot have both `Config<T>` and `ConfigMut<T>` for the same type T
 /// - Cannot use `&mut Storage` with `Config<T>` or `ConfigMut<T>`
--- a/sdk/patina/src/component.rs
+++ b/sdk/patina/src/component.rs
@@ -59,10 +59,8 @@
 //!
 //! ### Entry Point Requirements
 //!
-//! The entry point function's first parameter must be `self`, **NOT** `&self` or `&mut self`. This design choice
-//! allows components to be executed once and pass ownership of the entire struct (or items within the struct) to
-//! other components or services without cloning or borrowing. The rest of the parameters must implement the
-//! [Param](params::Param) trait, which is described in more detail below.
+//! The entry point function's first parameter must be `self, `mut self`, `&self` or `&mut self`. The rest of the
+//! parameters must implement the [Param](params::Param) trait, which is described in more detail below.
 //!
 //! Note: there is an arbitrary parameter count limit of 5, but this can be changed in the future if needed. See the
 //! [params] module for more information.
--- a/sdk/patina/src/component/struct_component.rs
+++ b/sdk/patina/src/component/struct_component.rs
@@ -9,10 +9,8 @@
 //! function of `Self::entry_point` exists. This can be overridden with the `#[entry_point(path = path::to::function)]`
 //! attribute.
 //!
-//! It is important to note that the function's first parameter must be `self` or `mut self`, **NOT** `&self` or
-//! `&mut self`. This design choice was made as components are only expected to be executed once, and by consuming
-//! `self`, you are able to pass ownership of the entire struct (or items within the struct) to other "things" (for
-//! lack of a better term) without the need for cloning or borrowing.
+//! It is important to note that the function's first parameter must be `self`, `mut self`, `&self` or
+//! `&mut self`.
 //!
 //! Review [Param] implementations for all types that can be used as parameters to these functions.
 //!
--- a/sdk/patina_macro/src/lib.rs
+++ b/sdk/patina_macro/src/lib.rs
@@ -243,7 +243,7 @@ pub fn smbios_record(item: proc_macro::TokenStream) -> proc_macro::TokenStream {
 /// ## Validation Rules
 ///
 /// - Impl block must contain an `entry_point` method
-/// - Entry point must have `self`, `&self`, or `&mut self` as first parameter
+/// - Entry point must have `self`, `mut self``, `&self`, or `&mut self` as the first parameter
 /// - No duplicate `ConfigMut<T>` parameters with the same type T
 /// - Cannot have both `Config<T>` and `ConfigMut<T>` for the same type T
 /// - Cannot use `&mut Storage` with `Config<T>` or `ConfigMut<T>`


sdk/patina/tests/ui/storage_with_config_mut.stderr
```diff
@@ -0,0 +1,15 @@
+error: Parameter conflict detected: parameter '_storage' (position 1) and parameter '_config' (position 2) conflict. You cannot use &Storage together with ConfigMut<T> parameters.
```
Javagedes: When two parameters conflict, I would assume we have the span of the current parameter we are parsing. We should be able to use that span in the `to_compile_error` in the component macro. E.g. the error could look something like this:

```
error: Parameter conflict detected: parameter '_storage' (position 1) and parameter '_config' (position 2) conflict. You cannot use &Storage together with ConfigMut<T> parameters.
  --> tests/ui/storage_with_config_mut.rs:12:20
   |
12 |     fn entry_point(self, _storage: &Storage, _config: ConfigMut<u32>) -> Result<()> {
   |                                              ^^^^^^^^^^^^^^^^^^^^^^^ conflicts with `&storage` parameter.
   |
```

sdk/patina_macro/src/component_macro.rs
```diff
@@ -140,36 +85,38 @@ impl Parse for Component {
 }
 
 /// The testable version of the `component` macro that uses proc_macro2::Tokenstreams.
+// Note: Marked as dead code since it's only used by tests.
+#[allow(dead_code)]
 pub(crate) fn component2(item: proc_macro2::TokenStream) -> proc_macro2::TokenStream {
```
Javagedes: Is this allow(dead_code) old?

I assume this has to be used in the actual macro, otherwise we don't have the IntoComponent implementation and the ComponentInput implementation.

sdk/patina_macro/src/validate_params_macro.rs
```diff
@@ -0,0 +1,1220 @@
...
+}
+
+/// Validates that an impl method has `self` or `mut self` as the first parameter.
+pub(crate) fn check_impl_method_has_self(func: &ItemFn) -> Result<(), TokenStream> {
+    let first_param = func.sig.inputs.first();
+
+    match first_param {
+        Some(FnArg::Receiver(_)) => Ok(()),
+        Some(FnArg::Typed(_)) => {
+            let error_msg = "Impl method entry point must use 'self' or 'mut self' as first parameter.";
```
Javagedes: component entry_points support `self`, `mut self`, `&self`, `&mut self`

This is because the `ParamFunction` will be used for more then just components in the future. Probably multi-callable events that should not be consuming self.

sdk/patina_macro/src/validate_params_macro.rs
```diff
@@ -0,0 +1,1220 @@
...
+    }
+}
+
+/// Check for parameter conflicts and return compile error if found
+/// Checks for parameter conflicts in the function signature.
+pub(crate) fn check_param_conflicts(func: &ItemFn) -> Result<(), TokenStream> {
+    let mut params: Vec<(usize, ParamType, String)> = Vec::new();
+
+    // Collect all parameters (skip 'self')
+    for (idx, arg) in func.sig.inputs.iter().enumerate() {
```
Javagedes, line 431: We should stash the span of the params here so that we can use it in error messages to red highlight over the (second) bad parameter instead of the component attribute macro itself.

sdk/patina_macro/src/validate_params_macro.rs
```diff
@@ -0,0 +1,1220 @@
...
+    // Check for conflicts
+    for i in 0..params.len() {
+        for j in (i + 1)..params.len() {
+            let (idx1, type1, name1) = &params[i];
+            let (idx2, type2, name2) = &params[j];
+
+            match (type1, type2) {
+                // Duplicate ConfigMut<T>
+                (ParamType::ConfigMut(t1), ParamType::ConfigMut(t2)) if t1 == t2 => {
+                    let error_msg = format!(
```
Javagedes: use the span here when creating the error.

sdk/patina_macro/src/validate_params_macro.rs
```diff
@@ -0,0 +1,1220 @@
...
+                        "Duplicate parameter type detected: parameter '{}' (position {}) and parameter '{}' (position {}) both have type ConfigMut<{}>. Each ConfigMut type can only appear once in a component's entry point.",
+                        name1, idx1, name2, idx2, t1
+                    );
+                    return Err(quote! {
```
Javagedes: I'm not a huge fan of this big match statement for different errors. It also differs from how we validate it at runtime. This makes it hard to maintain as when we edit the logic (or add a new parameter), we must implement the validation (or edit) two completely separate ways.

Javagedes: If possible we should try and have the validation be the same way. We cannot re-use the code due to circular dependencies, but even copy + pasting the same logic from patina to patina_macro would be beneficial.

sdk/patina_macro/src/validate_params_macro.rs
makubacki: Updated in latest push.

New error message with param span:

```
error: Patina component parameter conflict detected: parameter '_config2' (position 2) conflicts with parameter '_config1' (position 1). You cannot have both Config<u32> and ConfigMut<u32> for the same type.
  --> tests/ui/config_and_config_mut_conflict.rs:12:49
   |
12 |     fn entry_point(self, _config1: Config<u32>, _config2: ConfigMut<u32>) -> Result<()> {
   |                                                 ^^^^^^^^^^^^^^^^^^^^^^^^

error: conflicts with '_config1' parameter here
  --> tests/ui/config_and_config_mut_conflict.rs:12:26
   |
12 |     fn entry_point(self, _config1: Config<u32>, _config2: ConfigMut<u32>) -> Result<()> {
   |                          ^^^^^^^^^^^^^^^^^^^^^
```

sdk/patina/tests/ui/storage_with_config_mut.stderr
makubacki: Updated.

sdk/patina_macro/src/component_macro.rs
makubacki: I was able to clean this up. The `IntoComponent` implementation is generated in validate_params_macro.rs now.

sdk/patina_macro/src/validate_params_macro.rs
makubacki: I updated the message and added tests specifically for `&self` (`test_check_impl_method_has_self_accepts_ref_self()`) and `&mut self` (`test_check_impl_method_has_self_accepts_ref_mut_self()`).

sdk/patina_macro/src/validate_params_macro.rs
makubacki: I broke this up with a `conflicts_with()` method. It's separated out into a commit that will be squashed later to make it easier to see - https://github.com/OpenDevicePartnership/patina/pull/1124/commits/9165e5da61e08dd78ec84b79a006a27b0990e3ea.

makubacki: My understanding is that further consolidation of logic might be a larger undertaking that we might be able to track in a separate task.

docs/src/rfc/text/0003-mm-comm.md
```diff
@@ -261,12 +261,20 @@ pub trait MmCommunication {
...
 ///
 /// #
-#[derive(IntoComponent, IntoService)]
+#[derive(IntoService)]
 #[service(dyn MmCommunication)]
 pub struct MmCommunicator {
     comm_buffers: RefCell<Vec<CommunicateBuffer>>,
     sw_mmi_trigger_service: Option<Service<dyn SwMmiTrigger>>,
 }
+
```
os-d: I’m fine if we want to update the RFCs to include the new component model, but if we do we should also update the change log section? I also think that any already completed RFCs I probably wouldn’t change, but still pending ones (like ACPI) makes sense.

sdk/patina_macro/src/validate_params_macro.rs
```diff
@@ -0,0 +1,1248 @@
...
+///     fn entry_point(self, config: Config<u32>) -> Result<()> {
+///         Ok(())
+///     }
+/// }
+/// ```
+///
+/// ## Validation Rules
+///
+/// - Impl block must have an `entry_point` method
+/// - Entry point must have `self`, `&self`, or `&mut self` as first parameter
```
os-d: In another file you say that entry_point must have self as the first param, not &self or &mut self

sdk/patina_macro/src/validate_params_macro.rs
```diff
@@ -0,0 +1,1248 @@
...
+                        #entry_point_fn,
+                        self
+                    )
+                )
+            }
+        }
+    }
+}
+
+/// Validates that an impl method has `self`, `mut self`, `&self`, or `&mut self` as the first parameter.
```
os-d, line 184: Same here, either the other file should be updated or this validation should be updated

sdk/patina_macro/src/validate_params_macro.rs
makubacki: I created a dedicated tracking issue for this that I'm looking into now - https://github.com/OpenDevicePartnership/patina/issues/1145. I'll try to have an update in time for this PR, but to unblock other work, would you be ok with me following up after the PR if needed?

docs/src/rfc/text/0003-mm-comm.md
makubacki: That's reasonable. I will drop the updates from RFCs already implemented like 0003-mm-comm.md and 0009-standardize-component-crate.md update the change log in 0005-acpi.md and 0017-uefi-services-component.md.

sdk/patina_macro/src/validate_params_macro.rs
makubacki: I'll make all the descriptions consistent.

sdk/patina_macro/src/validate_params_macro.rs
makubacki: The other file is updated. The validation logic is retained as-is.

docs/src/rfc/text/0003-mm-comm.md
makubacki: Done

sdk/patina_macro/src/validate_params_macro.rs
Javagedes: sounds good.

vineelko (approved):
:shipit:

PR #1134, opened 2025-11-27
patina_dxe_core/fv: Fix `test_fv_functionality()` panic [Rebase & FF]

### LINKED ISSUE #1127: [Bug]: panic in fv.rs test case

### Current Behavior

Currently, a portion of the `fv.rs` `test_fv_functionality` test panics when running FV protocol `ReadFile` under the scenario where the buffer is valid, but the buffer size is 0. This needs to be investigated and fixed.

The test has been commented out and apporpiately marked with this bug reports ID number.

### Expected Behavior

The test should pass and not panic.

### Steps To Reproduce

1. uncomment out the test case
2. run the test

### Build Environment

```markdown
N/A
```

### Version Information

```text
76a1ef24e28317c0b92834a020284396098b9f0d
```

### Urgency

Medium

Someone else needs to fix it

No maintainer feedback needed

### Anything else?

_No response_

makubacki:
## Description

Fixes #1127

Code is currently commented out in `test_fv_functionality()` that calls `fv_read_file()` with a non-null buffer and buffer size of zero.

This change updates `fv_read_file()` to follow the same behavior as the C DXE Core `FvReadFile()` function in FwVolRead.c that will return `BUFFER_TOO_SMALL` in this scenario.

Another difference in behavior is that if the caller provides a non-null buffer but a buffer size less than the file content size, the C implementation will truncate the copy size to the size provided and perform the copy with that size. To prevent further discrepancies between caller expectations, that change is made in a second commit.

---

**patina_dxe_core: Truncate copies in fv_read_file()**

Matches the C implementation more closely by performing truncated copies
when the provided buffer is too small.

---

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- `cargo make all`

## Integration Instructions

- N/A

### DIFF
--- a/patina_dxe_core/src/fv.rs
+++ b/patina_dxe_core/src/fv.rs
@@ -379,24 +379,38 @@ extern "efiapi" fn fv_read_file(
         _ => return efi::Status::NOT_FOUND,
     };
 
-    // update file metadata output pointers.
-    // Safety: caller must provide valid pointers for found_type, file_attributes, and buffer_size. They are null-checked above.
+    // update file metadata output pointers (buffer_size is written later).
+    // Safety: caller must provide valid pointers for found_type and file_attributes. They are null-checked above.
     unsafe {
         found_type.write_unaligned(file.file_type_raw());
         file_attributes.write_unaligned(file.fv_attributes());
         //TODO: Authentication status is not yet supported.
-        buffer_size.write_unaligned(file.content().len());
     }
 
     if buffer.is_null() {
-        //caller just wants file meta data, no need to read file data.
+        // The caller just wants file meta data, no need to read file data.
+        // Safety: The caller must provide a valid pointer for buffer_size. It is null-checked above.
+        unsafe {
+            buffer_size.write_unaligned(file.content().len());
+        }
         return efi::Status::SUCCESS;
     }
 
     // Safety: caller must provide a valid pointer for buffer. It is null-checked above.
     let mut local_buffer_ptr = unsafe { buffer.read_unaligned() };
 
-    if local_buffer_ptr.is_null() {
+    // Determine the size to copy and the return status. For compatibility with existing callers to this function,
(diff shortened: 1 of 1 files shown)


Javagedes:
Wow you did this fast. I was expecting this be worked on after #1121 was merged, as it does a lot of moving of files around, and would still prefer that #1121 goes in before this.

But I can deal with the merge conflict if you want to get this in before that. The change itself looks good to me.

makubacki:
> nit: The first commit message says it isn't going to follow the C behavior because it is unintuitive but then the second commit adds that behavior. May be worth amending the commit message.

That reflects my decision process :). I decided to just get it to match to prevent dealing with some subtle difference in the future. I'll update the commit message.

patina_dxe_core/src/fv.rs
```diff
@@ -379,32 +379,38 @@ extern "efiapi" fn fv_read_file(
...
-            return efi::Status::BUFFER_TOO_SMALL;
-        }
-        if local_buffer_ptr.is_null() {
-            return efi::Status::INVALID_PARAMETER;
-        }
-    } else {
+    // Determine the size to copy and the return status. For compatibility with existing callers to this function,
+    // C code behavior (`FvReadFile()`) is retained that does the following based on inputs:
+    //
+    // 1. If the buffer pointer provided  is null, attempt to allocate a buffer of appropriate size via allocate_pool,
```
joschock, line 405:
```suggestion
    // 1. If the buffer pointer provided is null, attempt to allocate a buffer of appropriate size via allocate_pool,
```

joschock: oh, this is merged - no worries.

makubacki: I'll note to update this along with a future change since it only impacts a comment.

PR #1135, opened 2025-12-01
patina_debugger: Implement initial breakpoint timeout

cfernald:
## Description

Implements a timeout for the initial breakpoint using a timer provided by the core. This allows for the platform to configure a timeout for the debugger to wait for a connection before proceeding with normal execution.

This change also drops the notion that force_enable may not be used with a timeout. This may change when a runtime enablement story is further developed.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Manual tests on Q35

## Integration Instructions

N/A

### DIFF
--- a/patina_dxe_core/src/fv.rs
+++ b/patina_dxe_core/src/fv.rs
@@ -379,32 +379,38 @@ extern "efiapi" fn fv_read_file(
         _ => return efi::Status::NOT_FOUND,
     };
 
-    // update file metadata output pointers.
-    // Safety: caller must provide valid pointers for found_type, file_attributes, and buffer_size. They are null-checked above.
+    // update file metadata output pointers (buffer_size is written later).
+    // Safety: caller must provide valid pointers for found_type and file_attributes. They are null-checked above.
     unsafe {
         found_type.write_unaligned(file.file_type_raw());
         file_attributes.write_unaligned(file.fv_attributes());
         //TODO: Authentication status is not yet supported.
-        buffer_size.write_unaligned(file.content().len());
     }
 
     if buffer.is_null() {
-        //caller just wants file meta data, no need to read file data.
+        // The caller just wants file meta data, no need to read file data.
+        // Safety: The caller must provide a valid pointer for buffer_size. It is null-checked above.
+        unsafe {
+            buffer_size.write_unaligned(file.content().len());
+        }
         return efi::Status::SUCCESS;
     }
 
     // Safety: caller must provide a valid pointer for buffer. It is null-checked above.
     let mut local_buffer_ptr = unsafe { buffer.read_unaligned() };
 
-    if local_buffer_size > 0 {
-        //caller indicates they have allocated a buffer to receive the file data.
-        if local_buffer_size < file.content().len() {
-            return efi::Status::BUFFER_TOO_SMALL;
-        }
-        if local_buffer_ptr.is_null() {
-            return efi::Status::INVALID_PARAMETER;
-        }
-    } else {
+    // Determine the size to copy and the return status. For compatibility with existing callers to this function,
+    // C code behavior (`FvReadFile()`) is retained that does the following based on inputs:
+    //
+    // 1. If the buffer pointer provided  is null, attempt to allocate a buffer of appropriate size via allocate_pool,
+    //    set the copy size to the file size, write full file size to buffer_size output, and return SUCCESS.
+    // 2. If the buffer pointer is non-null, but the provided buffer size is smaller than the file size,
+    //    set the copy size to the provided buffer size, write this truncated size to buffer_size output,
+    //    perform the truncated copy into the provided buffer, and return WARN_BUFFER_TOO_SMALL.
+    // 3. If the buffer pointer is non-null, and the provided buffer size is sufficient to hold the file data,
+    //    set the copy size to the file size, write full file size to buffer_size output,
+    //    perform the copy into the provided buffer, and return SUCCESS.
+    let (copy_size, status) = if local_buffer_ptr.is_null() {
         //caller indicates that they wish to receive file data, but that this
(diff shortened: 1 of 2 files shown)


core/patina_debugger/src/debugger.rs
```diff
@@ -165,6 +171,14 @@ impl<T: SerialIO> PatinaDebugger<T> {
         };
 
         let mut target = PatinaTarget::new(exception_info, &self.system_state);
+        let timer: &dyn ArchTimerFunctionality = debug.timer.ok_or(DebugError::NotInitialized)?;
```
os-d: Do we want to fail here? Or default to 0 initial break point? May be useful if debugging something timer related, e.g.

core/patina_debugger/src/lib.rs
```diff
@@ -97,6 +107,7 @@
 
 mod arch;
 mod dbg_target;
+#[coverage(off)]
```
cfernald: Coverage requires pretty significant rework and infrastructure for integration tests that I have been idly working on. Disabling until we have the infra to actually test this code.

joschock: not a hard requirement, but I find it convenient to have a comment to that effect in the code

core/patina_debugger/src/debugger.rs
```diff
@@ -198,8 +207,35 @@ impl<T: SerialIO> PatinaDebugger<T> {
...
-        while !target.is_resumed() {
+        let mut timeout_reached = false;
+        if let GdbStubStateMachine::Idle(mut inner) = gdb {
+            // Always start with a stop code if starting from idle. This may be because this is the initial breakpoint
+            // or because the initial breakpoint timed out. This is not to spec, but is a useful hint to the client
+            // that a break has occurred. This allows the debugger to reconnect on scenarios like reboots.
+            let _ = inner.borrow_conn().write_all("$T05thread:01;#07".as_bytes());
+
+            // Until some traffic is received, wait for the timeout before entering the state machine.
+            if timeout != 0 {
```
os-d: Because we are doing this here, does that affect every break in? Or can the Idle state only be reached when at an initial breakpoint?

core/patina_debugger/src/debugger.rs
cfernald: Well right now the user doesn't have the ability to not provide a timer so this indicates bad internal state which will cause a failure later anyways when we try to get the GDB buffers. 

I could make it so that the caller can provide `None` if you think that would be useful.

core/patina_debugger/src/debugger.rs
```diff
@@ -60,23 +62,16 @@ where
     log_policy: DebuggerLoggingPolicy,
     /// Whether initializing the transport should be skipped.
     no_transport_init: bool,
-    /// Internal mutable debugger config.
-    config: spin::RwLock<DebuggerConfig>,
+    /// Debugger enabled state.
+    enabled: AtomicBool,
```
Javagedes, line 69: Is there a point in this `enabled` flag? It seems like the compile time feature flag takes care of this. e.g. if it is exists, its enabled. 

I think the cleanest user interface is that if the debugger is present, it is enabled. Then have the breakout timeout be an Option of a u32 with a default of None

None => no breakpoint
Some(0) -> infiinite breakpoint
Some(time) => time (s)

core/patina_debugger/src/debugger.rs
os-d: I think that having an option to have a more minimal debugger environment is always useful, if not too much work.

core/patina_debugger/src/debugger.rs
joschock: Historically, we've had auth-token-based enabling for the debugger at runtime (i.e. provision a device with a properly signed debug token, enable debugger, otherwise don't enable debugger).  It doesn't get a _ton_ of use, but I think it is a legit use-case for runtime enable of the debugger.

core/patina_debugger/src/debugger.rs
cfernald: Technically, yes. However, only the initial breakpoint will have this set to > 0. This seemed like the cleanest way to handle the different possible configurations.

cfernald: Well also the Idle state really only happens when either this is the initial breakpoint or the initial breakpoint timed out, but the above state still stands.

core/patina_debugger/src/debugger.rs
cfernald: The purpose of this is to allow for the possibility of runtime enablement. This is quite possible for device that may have   MFCI or similar mechanisms to enable the debugger without reflashing the device. 

Additionally, it's nice to have the debugger present but not enabled for compile size consistency.

core/patina_debugger/src/debugger.rs
cfernald: Easy enough, will add.

core/patina_debugger/src/debugger.rs
```diff
@@ -198,8 +207,35 @@ impl<T: SerialIO> PatinaDebugger<T> {
...
         };
 
-        // Enter the state machine until the target is resumed.
-        while !target.is_resumed() {
+        let mut timeout_reached = false;
+        if let GdbStubStateMachine::Idle(mut inner) = gdb {
+            // Always start with a stop code if starting from idle. This may be because this is the initial breakpoint
+            // or because the initial breakpoint timed out. This is not to spec, but is a useful hint to the client
+            // that a break has occurred. This allows the debugger to reconnect on scenarios like reboots.
+            let _ = inner.borrow_conn().write_all("$T05thread:01;#07".as_bytes());
```
makubacki: nit: `"$T05thread:01;#07"` would be nice to have in a defined constant.

PR #1138, opened 2025-12-01
patina_dxe_core: Reset protocol database between lib.rs tests

Raymond-MS:
## Description

Fix lib.rs tests to include resetting the protocol database.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Ran tests locally. Ran QEMU SBSA build to UEFI shell.

## Integration Instructions

N/A

### DIFF
--- a/patina_dxe_core/src/lib.rs
+++ b/patina_dxe_core/src/lib.rs
@@ -637,7 +637,7 @@ fn call_bds() -> ! {
 #[coverage(off)]
 mod tests {
     use super::*;
-    use core::sync::atomic::AtomicBool;
+    use core::{any::Any, sync::atomic::AtomicBool};
 
     #[test]
     fn test_cannot_set_instance_twice() {
@@ -672,6 +672,21 @@ mod tests {
         assert!(!<TestPlatform as MemoryInfo>::prioritize_32_bit_memory());
     }
 
+    fn with_reset_global_state<F>(f: F) -> core::result::Result<(), Box<dyn Any + Send>>
+    where
+        F: Fn() + std::panic::RefUnwindSafe,
+    {
+        test_support::with_global_lock(|| {
+            // SAFETY: init_test_protocol_db modifies global state. It is being called within a
+            // lock to have exclusive mutable access to the protocol database.
+            unsafe {
+                test_support::init_test_protocol_db();
+            }
+
+            f()
+        })
+    }
+
     #[test]
     fn test_mock_call_bds_valid_non_null() {
         static BDS_CALLED: AtomicBool = AtomicBool::new(false);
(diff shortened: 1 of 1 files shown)


patina_dxe_core/src/lib.rs
```diff
@@ -673,6 +673,12 @@ mod tests {
 
         assert!(
             test_support::with_global_lock(|| {
+                // SAFETY: init_test_protocol_db modifies global state. It is being called within a
```
makubacki: Other modules already wrap `test_support::with_global_lock()` to make the tests cleaner and consistently do the right thing. For example: https://github.com/raymond-ms/patina/blob/personal/rdiaz/FixLibTests/patina_dxe_core/src/events.rs#L346

Then, you might end up with a function like:

```rust
fn with_clean_protocol_db<F>(f: F)
where
    F: Fn() + std::panic::RefUnwindSafe,
{
        // SAFETY: init_test_protocol_db modifies global state. It is being called within a
        // lock to have exclusive mutable access to the protocol database.
        unsafe {
            test_support::init_test_protocol_db();
        }
        f()
    })
}
```

Raymond-MS: Updated. :) Thanks!

patina_dxe_core/src/lib.rs
```diff
@@ -670,60 +670,64 @@ mod tests {
         assert!(!<TestPlatform as MemoryInfo>::prioritize_32_bit_memory());
     }
 
+    fn with_clean_protocol_db<F>(f: F) -> bool
```
os-d: nit: I would consider naming this something else, because right now all it does is clean the protocol db (although even now it also does the global lock), but in the future it may have other test setup stuff in it (like in other files). Perhaps something like with_reset_global_state or whatever

patina_dxe_core/src/lib.rs
```diff
@@ -670,60 +670,64 @@ mod tests {
...
+        test_support::with_global_lock(|| {
+            // SAFETY: init_test_protocol_db modifies global state. It is being called within a
+            // lock to have exclusive mutable access to the protocol database.
+            unsafe {
+                test_support::init_test_protocol_db();
+            }
+
+            f()
+        })
+        .is_err_and(|err| {
```
os-d: This also doesn’t seem like it will always be valid. Right now the few tests here all end with calling bds, but, they could error otherwise, say before calling bds, and this would be confusing and other tests may be written that don’t call bds at all. I would recommend leaving the error handling in each test and just let them error out.

os-d (approved):

makubacki:
@Raymond-MS, we're getting frequent failures in CI because of this. Do you need any help wrapping up the final changes so we can merge the PR?

Raymond-MS:
> @Raymond-MS, we're getting frequent failures in CI because of this. Do you need any help wrapping up the final changes so we can merge the PR?

No, I am finishing this up now.

Issue #791, opened 2025-09-29
[Bug]: events::wait_for_event needs to be updated to match EDK2 semantics.


joschock:
### Is there an existing issue for this?

- [x] I have searched existing issues

### Current Behavior

1. wait_for_event does not signal the idle loop event (https://github.com/tianocore/edk2/blob/38370cf4924205df97b8bcaf0f184983a4b34f66/MdeModulePkg/Core/Dxe/Event/Event.c#L697)
2. wait_for_event returns an error if out_index is null, but it is an optional input and should simply be ignored if null.

### Expected Behavior

function should align with edk2 semantics.

### Steps To Reproduce

By inspection. 

### Build Environment

```markdown
N/A
```

### Version Information

```text
Tag: patina-v8.3.0
```

### Urgency

Low

### Are you going to fix this?

I will fix it

### Do you need maintainer feedback?

No maintainer feedback needed

PR #1140, opened 2025-12-01
Update events::wait_for_event() semantics

joschock:
## Description

Adjusts wait_for_event semantics to better match historical behavior of core:
* make `out_index` parameter optional
* invoke CPU sleep function during idle loop.

Closes https://github.com/OpenDevicePartnership/patina/issues/791

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Boot to OS on AARCH64 hardware.

## Integration Instructions

N/A

### DIFF
--- a/patina_dxe_core/src/events.rs
+++ b/patina_dxe_core/src/events.rs
@@ -15,7 +15,7 @@ use r_efi::efi;
 
 use patina::pi::protocols::timer;
 
-use patina_internal_cpu::interrupts;
+use patina_internal_cpu::{cpu::EfiCpu, interrupts};
 
 use crate::{
     event_db::{SpinLockedEventDb, TimerDelay},
@@ -132,7 +132,7 @@ extern "efiapi" fn wait_for_event(
             match check_event(event) {
                 efi::Status::NOT_READY => (),
                 status => {
-                    // Safety: caller must ensure that out_index is a valid pointer if it is not null. It is null-checked above.
+                    // Safety: caller must ensure that out_index is a valid pointer if it is not null.
                     if !out_index.is_null() {
                         unsafe {
                             out_index.write_unaligned(index);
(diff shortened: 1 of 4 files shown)


patina_dxe_core/src/events.rs
```diff
@@ -132,16 +132,25 @@ extern "efiapi" fn wait_for_event(
             match check_event(event) {
                 efi::Status::NOT_READY => (),
                 status => {
-                    // Safety: caller must ensure that out_index is a valid pointer. It is null-checked above.
-                    unsafe {
-                        out_index.write_unaligned(index);
-                    };
+                    // Safety: caller must ensure that out_index is a valid pointer if it is not null. It is null-checked above.
```
makubacki: Technically, this would be more useful as a function safety comment, but many of these are already not marked `unsafe`. Since it's not `pub`, it also might not matter as much. Feel free to add a brief function safety comment explaining the behavior or just close.

os-d:
```suggestion
                    // Safety: caller must ensure that out_index is a valid pointer if it is not null. It is null-checked before being written to.
```

No longer above

Javagedes (approved):

patina_dxe_core/src/events.rs
joschock: I would like to mark the function unsafe, but it can't be marked unsafe without breaking changes to r_efi upstream because it's an implementation of a Boot Services member that is not marked unsafe in the upstream r_efi definition of the API.

IMHO `extern` (at least the way it used in patina) should be treated as `unsafe` for purposes of safety documentation. I'll adjust the comment.

joschock: Thinking about it more, I think I will leave the safety comment as-is, at least for this PR. Otherwise `wait_for_event` would be the only function with this kind of doc. If we decide that's worth it, I think we should do so as a focused "safety doc externs" PR.

PR #1141, opened 2025-12-01
Makefile.toml: Add doc test task to `all` [Rebase & FF]

### LINKED ISSUE #1125: [Feature]: Include `cargo test --doc` results in `cargo make all`
### Feature Overview

Doc tests should be included in the `cargo make all` command to more comprehensively verify changes locally before running PR status checks.

### Solution Overview

Include `cargo test --doc` results in the `all` cargo make task.

### Alternatives Considered

_No response_

### Urgency

Low

I will implement the feature

No maintainer feedback needed

### Anything else?

_No response_

makubacki:
## Description

Resolves #1125 

Add a new `test-doc` task that runs `cargo test --doc` and include it in the `all` task dependencies to ensure documentation tests are validated as part of the "all" flow.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- `cargo make test-doc`
- `cargo make all`

## Integration Instructions

- N/A

### DIFF
--- a/patina_dxe_core/src/cpu.rs
+++ b/patina_dxe_core/src/cpu.rs
@@ -36,21 +36,22 @@ use patina_internal_cpu::{cpu::EfiCpu, interrupts::Interrupts};
 /// struct PlatformConfig;
 /// # impl ComponentInfo for PlatformConfig {}
 /// # impl MemoryInfo for PlatformConfig {}
-/// # impl CpuInfo for PlatformConfig {}
 ///
-/// impl PlatformInfo for PlatformConfig {
-///   # type MemoryInfo = Self;
-///   # type Extractor = patina_ffs_extractors::NullSectionExtractor;
-///   # type ComponentInfo = Self;
-///   # type CpuInfo = Self;
-///
-///   # #[cfg(target_arch = "aarch64")]
+/// impl CpuInfo for PlatformConfig {
+///   #[cfg(target_arch = "aarch64")]
 ///   fn gic_bases() -> GicBases {
 ///     /// SAFETY: gicd and gicr bases correctly point to the register spaces.
 ///     /// SAFETY: Access to these registers is exclusive to this struct instance.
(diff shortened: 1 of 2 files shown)


makubacki:
@Javagedes & @vineelko, I also tried adding `--doctests` to the existing `test` task and it worked, but as mentioned [here](https://docs.rs/crate/cargo-llvm-cov/latest/source/docs/cargo-llvm-cov.txt) it is unstable:

```
warning: --doctests option is unstable
```

It is just a warning. I think that might be better from the perspective that it seems to include doc tests in coverage which would be more accurate for overall coverage.

**Before**

image

**After**

image

---

Any concerns with this approach?

makubacki:
> @Javagedes & @vineelko, I also tried adding `--doctests` to the existing `test` task and it worked,
> ...
> Any concerns with this approach?

I went ahead and pushed this update. Can you please reapprove if you have no concern?

vineelko:
Michael, I think your new commit makes sense. I was confused because there is no `--doctests` on `cargo test`. I think you meant `--doctests` to `cargo make coverage`. I don't have any concerns.

makubacki:
> Michael, I think your new commit makes sense. I was confused because there is no `--doctests` on `cargo test`. I think you meant `--doctests` to `cargo make coverage`. I don't have any concerns.

I mean the `test` task in Makefile.toml specifically which has been updated to participate in coverage.

makubacki:
Also, it's going to take a few pushes to get the aarch64 tests passing.

vineelko:
I didn't know that `[tasks.test]` is mapped to `llvm-cov`. Now I remember why it was done. Make sense now. Thanks.

vineelko (approved):

makubacki:
> Also, it's going to take a few pushes to get the aarch64 tests passing.

Done

PR #1146, opened 2025-12-02
[REBASE & FF] Add Stability Patina Tests

os-d:
## Description

This PR adds stability tests to Patina DXE Core. It also adds timer based patina test dispatch.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [x] Includes documentation?

## How This Was Tested

Tested on Q35 and SBSA.

## Integration Instructions

See docs.

### DIFF
--- a/patina_dxe_core/src/core_patina_tests/stability_tests.rs
+++ b/patina_dxe_core/src/core_patina_tests/stability_tests.rs
@@ -1,0 +1,535 @@
+//! DXE Core Patina Test Stability Tests
+//!
+//! These tests are intended to verify the stability and reliability of the Patina DXE Core.
+//!
+//! ## License
+//!
+//! Copyright (c) Microsoft Corporation.
+//!
+//! SPDX-License-Identifier: Apache-2.0
+//!
+
+use super::test_support::*;
+use crate::{GCD, gcd::AllocateType};
+use alloc::vec::Vec;
+use patina::{
+    base::{SIZE_1GB, SIZE_2MB, SIZE_4KB},
+    pi::dxe_services::GcdMemoryType,
+    test::patina_test,
+    u_assert, u_assert_eq,
+};
+use patina_paging::MemoryAttributes;
+use r_efi::efi;
+
+/// Stability Test: Split a 2MB page into 4KB pages and verify correctness
+#[patina_test]
+#[on(timer = 3_000_000)] // 300ms interval
+fn page_table_tests_2mb_split() -> patina::test::Result {
+    let mut addr_vec = Vec::new();
+    for _ in 0..19 {
+        addr_vec.push(
+            GCD.allocate_memory_space(
+                AllocateType::TopDown(None),
+                GcdMemoryType::SystemMemory,
+                21, // alignment shift for 2MB
+                SIZE_2MB,
+                0x05D as efi::Handle,
+                None,
+            )
+            .unwrap(),
+        );
+    }
+
+    for addr in addr_vec {
+        u_assert_eq!(addr % SIZE_2MB, 0);
+
+        // set the whole region as RWX first so we can adjust the attributes easily after
+        GCD.set_memory_space_attributes(addr, SIZE_2MB, efi::MEMORY_WB)
+            .map_err(|_| "Failed to set attributes on 2MB region")?;
+
+        // Make sure the page is actually a 2MB page by reading the entry in the self map
+        let mut pt = get_self_mapped_page_table(addr as u64, PageTableLevel::Level2);
+        let mut idx = get_index(addr as u64, PageTableLevel::Level2);
+        let mut entry = pt.get(idx as usize).ok_or("Index out of bounds")?;
+        let mut self_map_entry = entry;
+        let mut pte_state = get_pte_state(*entry, PageTableLevel::Level2);
+
+        u_assert!(pte_state.points_to_pa);
+        u_assert_eq!(pte_state.attributes, MemoryAttributes::empty());
+        u_assert_eq!(pte_state.next_address, addr as u64);
+        u_assert!(pte_state.present);
+
+        // if the identity mapped page table address is mapped, let's check that that matches
+        pt = get_self_mapped_page_table(addr as u64, PageTableLevel::Level3);
+        idx = get_index(addr as u64, PageTableLevel::Level3);
+        entry = pt.get(idx as usize).ok_or("Index out of bounds")?;
+        pte_state = get_pte_state(*entry, PageTableLevel::Level3);
+        let mut id_mapped_pt = pte_state.next_address;
+
+        if is_mapped(id_mapped_pt) {
+            // we are identity mapped, so check this matches the self map
+            let pt_ptr = id_mapped_pt as *const u64;
+            // No matter what level the page table itself is mapped at, we want the level 2 page table entry
+            // for the actual address, because we confirmed above that it is a 2MB page
+            idx = get_index(addr as u64, PageTableLevel::Level2);
+            // SAFETY: We have just confirmed this PT is mapped, so we can read it
+            let entry_value = unsafe { pt_ptr.add(idx as usize).read() };
+            u_assert_eq!(*self_map_entry, entry_value);
+        }
+
+        // force a split
+        GCD.set_memory_space_attributes(addr, SIZE_4KB, efi::MEMORY_XP | efi::MEMORY_WB)
+            .map_err(|_| "Failed to set attributes on 4KB region")?;
+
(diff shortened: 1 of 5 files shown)


patina_dxe_core/src/patina_test_support.rs
```diff
@@ -0,0 +1,279 @@
...
+
+mod stability_tests;
+
+use alloc::slice;
+use bitfield_struct::bitfield;
+use core::arch::asm;
+use patina::base::{SIZE_1GB, SIZE_2MB, SIZE_4KB, UEFI_PAGE_SHIFT};
+use patina_paging::MemoryAttributes;
+
+pub(crate) struct PteInfo {
```
Javagedes: Would it be worth making any of these structs public in patina-paging (assuming that's where this stuff came from) and reusing them?

sdk/patina_macro/src/test_macro.rs
```diff
@@ -131,6 +131,13 @@ fn parse_on_attr(attr: &Attribute) -> syn::Result<proc_macro2::TokenStream> {
...
+                // CASE2: $[on(timer = interval_in_100ns_units)]
+                Meta::NameValue(nv) if nv.path.is_ident("timer") => {
+                    let value = &nv.value;
+                    return Ok(quote! {
+                        patina::test::__private_api::TestTrigger::Timer(#value)
+                    });
+                }
                 // No other cases are supported right now.
                 _ => {
                     return Err(syn::Error::new(meta.span(), "Only support #[on(event = ...)]"));
```
Javagedes: This error message needs updated. Probably just to say that it is an unsupported attribute configuration or something

patina_dxe_core/src/patina_test_support.rs
os-d: I debated on this with @cfernald . In the end, I decided to not make any of this public from patina-paging for a few reasons:
- From the patina-paging side, these really shouldn't be public. Consumers of the crate should not be interacting with these functions or structures at all and it is dangerous for them to do so.
- The patina-paging structs are hidden behind abstractions which would be awkward to use here and in patina-paging we don't want the arch agnostic code to have access directly to the arch specific code. In addition, these structures are not 1:1 with the patina-paging structures (aside from the architecturally defined structures).
- The idea behind these tests is an independent verification of the system state. These are architectural definitions, not patina-paging definitions (which isn't really true of the self map checking, though if we really cared that could be made non-patina-paging specific as well). The idea being that even if patina-paging has a bug in these structs or parsing them, we should be able to catch it here. Which is why I don't use any patina-paging functions to validate state, I do it all manually.

Now, that being said I also debated putting the architectural definitions in the SDK, e.g., but that also felt inappropriate because we really don't want any actual FW code to use these; it really is just for patina_test code. But I'm open to moving somewhere else. If patina-paging were allowed to depend on the SDK, then that would be better, but currently we do not let it.

Javagedes: Nah I don't mind where it is at. Probably not worth moving things around.

patina_dxe_core/src/patina_test_support/stability_tests.rs
```diff
@@ -0,0 +1,535 @@
...
+            u_assert!(pte_state.present);
+            if i == 0 {
+                u_assert_eq!(pte_state.attributes, MemoryAttributes::ExecuteProtect);
+            } else {
+                u_assert_eq!(pte_state.attributes, MemoryAttributes::empty());
+            }
+        }
+
+        // now let's make sure that the paging crate did all the TLB invalidation it needed to
+        flush_tlb();
```
joschock: will this potentially mask issues?

patina_dxe_core/src/patina_test_support.rs
```diff
@@ -0,0 +1,279 @@
...
+//! Code to help support running Patina Tests.
+//!
+//! ## License
+//!
+//! Copyright (c) Microsoft Corporation.
+//!
+//! SPDX-License-Identifier: Apache-2.0
+//!
+
+mod stability_tests;
```
cfernald: This is a lot of code. I wonder if there is some way we can conditionally compile it only when patina_tests are enabled.

patina_dxe_core/src/core_patina_tests/test_support.rs
```diff
@@ -0,0 +1,279 @@
+//! DXE Core Patina Test Support
+//!
+//! Code to help support running Patina Tests.
```
cfernald, line 3: Stylistic, but I find it odd that the module root has "support" and the sub-modules have the tests. I wonder if this should all be something like "core_tests" with "test_support" and "stability_tests" being sub-modules.

patina_dxe_core/src/core_patina_tests/stability_tests.rs
```diff
@@ -0,0 +1,535 @@
...
+    pi::dxe_services::GcdMemoryType,
+    test::patina_test,
+    u_assert, u_assert_eq,
+};
+use patina_paging::MemoryAttributes;
+use r_efi::efi;
+
+/// Stability Test: Split a 2MB page into 4KB pages and verify correctness
+#[patina_test]
+#[on(timer = 3_000_000)] // 300ms interval
```
cfernald, line 26: What is the motivation of running this multiple times? It seems to me that point in time shouldn't make much of a difference, right?

cfernald (approved):

patina_dxe_core/src/patina_test_support.rs
os-d: Oops, I originally had the entire patina_test_support module at the top only be compiled when the feature is enabled, but I accidentally removed that. I will add it back.

patina_dxe_core/src/core_patina_tests/test_support.rs
os-d: I don't mind doing that, I was just using the module since it existed, but fine to have the top module just be a simple container module with no logic.

os-d: When I first wrote this, I was envisioning the tests going into the actual core modules themselves and not being isolated, so this module was just support. Then I decided it was a lot of code and would be nice to isolate the tests, but didn't change the name.

patina_dxe_core/src/core_patina_tests/stability_tests.rs
os-d: The original idea behind these tests was a stress test. I was envisioning having a single test that came up and did a lot of scary operations and then would continuously do that.

I ended up splitting it up into separate tests, but I still wanted to keep the idea of a stress test, especially since these test cases were around issues of speculative execution. My rationale was that the page table is continuously different throughout boot and speculative execution is obviously different at any given point in time, so for these tests I felt more comfortable running them a lot, to have more confidence.

On a given system, we've seen where a large page split may only happen once or twice, so I really wanted to force that situation to happen many more times and catch it in the test and not on a platform.

patina_dxe_core/src/core_patina_tests/test_support.rs
cfernald: Yeah, slight preference to clarifying this is not just support, but actual tests.

patina_dxe_core/src/patina_test_support.rs
cfernald: I think that feature is just for the SDK, right? I think we would have to add one for the core.

patina_dxe_core/src/core_patina_tests/stability_tests.rs
cfernald: I'm alright with it, mostly just curous if there was a specific scenario I wasn't thinking of. No need to change this.

patina_dxe_core/src/patina_test_support.rs
os-d: Actually, this wasn't an accident, I remember the conversation I had with @Javagedes now. The problem was that the enable_patina_tests feature is not a feature of patina_dxe_core, but of patina (SDK). We had a discussion about changing some of the code organization so that patina_dxe_core owns the feature and enables the feature in patina SDK as well, but that will come after this. I will file an issue to track.

os-d: https://github.com/OpenDevicePartnership/patina/issues/1159. Resolving this for now and this will be conditionally added based on that issue resolution.

patina_dxe_core/src/core_patina_tests/test_support.rs
os-d: Refactored.

kuqin12 (approved):

PR #1147, opened 2025-12-02
TplMutex: own BootServices instead of borrowing

kat-perez:
## Description

This eliminates the lifetime parameter from TplMutex, removing the need for unsafe lifetime casts when creating static TplMutex instances.

- [x] Impacts functionality?
- [ ] Impacts security?
- [x] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

`cargo make all`

Integrated this branch into patina-dxe-core-qemu and tested that it boots to UEFI shell in patina-qemu

## Integration Instructions

Update all TplMutex::new call sites to pass an owned BootServices instead of a reference:

```
// Before
let mutex = TplMutex::new(&boot_services, Tpl::NOTIFY, data);

// After
let mutex = TplMutex::new(boot_services.clone(), Tpl::NOTIFY, data);
```

Remove any 'static lifetime parameters from TplMutex type annotations:

```
// Before
TplMutex<'static, MyData, StandardBootServices>

// After
TplMutex<MyData, StandardBootServices>
```

### DIFF
--- a/components/patina_performance/src/component/performance.rs
+++ b/components/patina_performance/src/component/performance.rs
@@ -42,7 +42,7 @@ use r_efi::system::EVENT_GROUP_READY_TO_BOOT;
 pub use mu_rust_helpers::function;
 
 /// Context parameter for the Ready-to-Boot event callback that fetches MM performance records.
-type MmPerformanceEventContext<BB, B, F> = Box<(BB, &'static TplMutex<'static, F, B>, Service<dyn MmCommunication>)>;
+type MmPerformanceEventContext<BB, B, F> = Box<(BB, &'static TplMutex<F, B>, Service<dyn MmCommunication>)>;
 
 /// Performance Component.
 pub struct Performance;
@@ -89,7 +89,7 @@ impl Performance {
         runtime_services: RR,
         records_buffers_hobs: Option<P>,
         mm_comm_service: Option<Service<dyn MmCommunication>>,
-        fbpt: &'static TplMutex<'static, F, B>,
+        fbpt: &'static TplMutex<F, B>,
         timer: Service<dyn ArchTimerFunctionality>,
     ) -> Result<(), EfiError>
     where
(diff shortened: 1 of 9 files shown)


Javagedes (approved):

components/patina_performance/src/component/performance.rs
```diff
@@ -89,7 +89,7 @@ impl Performance {
         runtime_services: RR,
         records_buffers_hobs: Option<P>,
         mm_comm_service: Option<Service<dyn MmCommunication>>,
-        fbpt: &'static TplMutex<'static, F, B>,
+        fbpt: &'static TplMutex<F, B>,
```
Javagedes, line 92: Not really a part of your PR, but does this need to be a static reference to a TplMutex?

sdk/patina/src/tpl_mutex.rs
```diff
@@ -20,8 +20,11 @@ use crate::boot_services::{BootServices, StandardBootServices, tpl::Tpl};
 /// Type use for mutual exclusion of data across Tpl (task priority level)
 ///
 /// This mutex will raise the TPL to the specified level when locked, and restore it when the lock is released.
-pub struct TplMutex<'a, T: ?Sized, B: BootServices = StandardBootServices> {
-    boot_services: &'a B,
+///
+/// The mutex owns the BootServices instance, eliminating lifetime requirements and unsafe casts
```
joschock: super-nit: my personal opinion is that it is not useful to refer to prior design points in comments; they should describe what is happening now and how to use it, unless there is an intentional deviation from a spec or something like that that requires explanation.

kat-perez: Got it, will not do that for future comments. Updated the comment to only reflect the current design

components/patina_performance/src/component/performance.rs
kat-perez: It looks like fbpt is a global static and is used outside of `entry_point` so we need the static reference to TplMutex

sdk/patina/src/tpl_mutex.rs
joschock: (editorial note: frankly, I do this too, and others should call me out on it ;))

PR #1151, opened 2025-12-03
patina_debugger: Implement single register access

cfernald:
## Description

Implement the single register read and write functionality for the debugger target. Windbg uses the `P<index:value>` packets for write altered register instead of sending the full register context every time. 

This resolves an issue where register edits appear to work, until the resume where the edits are lost.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on Q35 & SBSA

## Integration Instructions

N/A

### DIFF
--- a/core/patina_debugger/src/arch/aarch64.rs
+++ b/core/patina_debugger/src/arch/aarch64.rs
@@ -295,8 +295,8 @@ pub struct Aarch64CoreRegs {
     pub sp: u64,
     /// Instruction pointer
     pub pc: u64,
-    /// Floating point control
-    pub fpcr: u64,
+    /// Floating point status
+    pub fpsr: u64,
     /// PE status
     pub cpsr: u32,
 }
@@ -323,7 +323,7 @@ impl Registers for Aarch64CoreRegs {
 
         write_bytes!(&self.sp.to_le_bytes());
         write_bytes!(&self.pc.to_le_bytes());
-        write_bytes!(&self.fpcr.to_le_bytes());
+        write_bytes!(&self.fpsr.to_le_bytes());
         write_bytes!(&self.cpsr.to_le_bytes());
     }
 
@@ -349,7 +349,7 @@ impl Registers for Aarch64CoreRegs {
 
         self.sp = read!(u64);
         self.pc = read!(u64);
-        self.fpcr = read!(u64);
+        self.fpsr = read!(u64);
         self.cpsr = read!(u32);
         Ok(())
     }
@@ -393,7 +393,7 @@ impl UefiArchRegs for Aarch64CoreRegs {
             ],
             sp: context.sp,
             pc: context.elr,
-            fpcr: context.fpsr,
+            fpsr: context.fpsr,
             cpsr: context.spsr as u32,
         }
     }
@@ -432,7 +432,7 @@ impl UefiArchRegs for Aarch64CoreRegs {
         context.lr = self.regs[30];
         context.sp = self.sp;
         context.elr = self.pc;
-        context.fpsr = self.fpcr;
+        context.fpsr = self.fpsr;
         context.spsr = self.cpsr as u64;
     }
 
--- a/core/patina_debugger/src/arch/xml/aarch64_registers.xml
+++ b/core/patina_debugger/src/arch/xml/aarch64_registers.xml
@@ -34,6 +34,6 @@
   <reg name="x30" bitsize="64" type="int64" regnum="30" />
   <reg name="sp" bitsize="64" type="data_ptr" regnum="31" />
   <reg name="pc" bitsize="64" type="code_ptr" regnum="32" />
-  <reg name="fpcr" bitsize="64" type="int64" regnum="33" />
+  <reg name="fpsr" bitsize="64" type="int64" regnum="33" />
   <reg name="cpsr" bitsize="32" type="int32" regnum="34" />
 </feature>


core/patina_debugger/src/arch/aarch64.rs
```diff
@@ -435,6 +435,146 @@ impl UefiArchRegs for Aarch64CoreRegs {
...
+                    23 => context.x23,
+                    24 => context.x24,
+                    25 => context.x25,
+                    26 => context.x26,
+                    27 => context.x27,
+                    28 => context.x28,
+                    _ => return Err(()),
+                };
+                let bytes = value.to_le_bytes();
+                buf[..8].copy_from_slice(&bytes);
```
os-d: nit: instead of 8 vs 4 in these, perhaps doing `let size = sizeof(context.<reg>)`?

core/patina_debugger/src/arch/x64.rs
```diff
@@ -409,6 +409,146 @@ impl UefiArchRegs for X64CoreRegs {
...
+                    10 => context.r10,
+                    11 => context.r11,
+                    12 => context.r12,
+                    13 => context.r13,
+                    14 => context.r14,
+                    15 => context.r15,
+                    _ => return Err(()),
+                };
+                let bytes = value.to_le_bytes();
+                buf[..8].copy_from_slice(&bytes);
```
os-d: Same nit

core/patina_debugger/src/arch/aarch64.rs
makubacki, line 1: This is in existing code, but is it expected?

```
context.fpsr = self.fpcr;
```

I'm not super familiar with the register, it just seems to go against the normal pattern.

core/patina_debugger/src/arch/aarch64.rs
```diff
@@ -435,6 +435,146 @@ impl UefiArchRegs for Aarch64CoreRegs {
...
         context.spsr = self.cpsr as u64;
     }
+
+    fn read_single_from_context(
+        context: &ExceptionContext,
+        reg_id: <super::SystemArch as gdbstub::arch::Arch>::RegId,
+        buf: &mut [u8],
+    ) -> Result<usize, ()> {
+        match reg_id {
+            Aarch64CoreRegId::Gpr(n) => {
```
makubacki: Could the GPR logic be consolidated in a helper function to avoid repetition?

core/patina_debugger/src/arch/aarch64.rs
makubacki: I assume this is unlikely, but `buf[..8].copy_from_slice(&bytes);` will panic if `buf.len()` is not 8. Since you're already returning a result, I think checking for this and returning an `Err` would be better.

core/patina_debugger/src/arch/aarch64.rs
```diff
@@ -435,6 +435,146 @@ impl UefiArchRegs for Aarch64CoreRegs {
...
+                let value = match n {
+                    0 => context.x0,
```
joschock: macro?

makubacki: I mentioned something similar in https://github.com/OpenDevicePartnership/patina/pull/1151#discussion_r2586727544. Either a function or macro would be fine with me, main purpose would be to consolidate logic.

core/patina_debugger/src/arch.rs
```diff
@@ -96,4 +96,18 @@ pub trait UefiArchRegs: Sized {
     fn read_from_context(&mut self, context: &ExceptionContext) {
         *self = Self::from_context(context);
     }
+
+    // Reads a single register from a UEFI context structure.
+    fn read_single_from_context(
```
vineelko: >single [](http://example.com/codeflow?start=12&length=6)

NIT: May be renaming it as `read_register_from_context()` would be better.

core/patina_debugger/src/arch/aarch64.rs
vineelko: Maybe `buf.get(..8)` works better.

core/patina_debugger/src/arch/aarch64.rs
cfernald: I played with this a bit, and its not really straight forward. At least that I can see. you can't do partial names in macros, at least that I found so its the same number of lines, just redirected. Additionally, a function isn't great either because the difference in mutability requires equal duplication.

@joschock @makubacki , open to suggestions if i'm missing something though,

core/patina_debugger/src/arch/aarch64.rs
cfernald: Resolving this to handle in conversation below.

core/patina_debugger/src/arch/aarch64.rs
joschock: You could use a hack based on casting it to u64 slice and indexing it. Not sure this is better? And you'd have to newtype the context structs, since they are in r_efi.

```rust
trait RegisterIndex {
  fn reg_index(&self, idx: usize) -> u64;
}

impl RegisterIndex for SystemContextAArch64 {
   fn reg_index(&self, idx: usize) -> u64 {
       //Safety: struct is repr c, so layout is well known.
       unsafe {
           slice_from_raw_parts(self.as_ptr() as *const u64, 28)[idx]
       }
   }
}

core/patina_debugger/src/arch/aarch64.rs
cfernald: I honestly cannot recall. I do vaugely remember there being some oddity here, though cannot find anytrhing. This is inherited from the C version.

https://github.com/microsoft/mu_feature_debugger/blob/567fdffeb56215ddfe0c43330865d7d4f3455dff/DebuggerFeaturePkg/Library/DebugAgent/GdbStub/GdbStubAarch64.c#L56

I will just resolve this, and if something comes up when we will know. In reality it doesnt matter since we dont meaningfully use either.

core/patina_debugger/src/arch/aarch64.rs
cfernald: I'm leaning toward just leaving it as it is. It's a little repetitious, but only slightly. The macros I have added makes it even more streamlined.

Issue #560, opened 2025-06-30
Consider running unit tests (or a subset of them) under Miri


makubacki:
This was converted to an issue to track adding miri support to nightly builds. Might entail also fixing things, etc.

---

### Discussed in https://github.com/OpenDevicePartnership/patina/discussions/546

<div type='discussions-op-text'>

<sup>Originally posted by **wesleywiser** June 19, 2025</sup>
This would help find issues like #530 and #544. Miri is a dynamic analysis (like ASAN or valgrind) so it likely won't find every instance of UB in your code but will help you with any code that is covered by your unit tests and does not run into the platform limitations the tool has. 

When investigating #544, I was able to run the tool locally by finding a nightly toolchain that corresponds to Rust 1.85 (such as `nightly-2025-01-03`), installing it and then running whichever unit tests, you want. For example:

```
> rustup toolchain install nightly-2025-01-03 --component miri
> cargo +nightly-2025-01-03 miri test -- create_fill_check_test
...
error: Undefined Behavior: attempting a write access using <326320> at alloc132202[0x50], but that tag does not exist in the borrow stack for this location
   --> components\patina_adv_logger\src\memory_log.rs:294:9
    |
294 | /         ptr::write_volatile::<Self>(
295 | |             address as *mut Self,
296 | |             Self::new(log_entry.phase, log_entry.level, log_entry.timestamp, log_entry.data.len() as u16),      
297 | |         );
    | |         ^
    | |         |
    | |_________attempting a write access using <326320> at alloc132202[0x50], but that tag does not exist in the borrow stack for this location
    |           this error occurs as part of an access at alloc132202[0x50..0x68]
    |
    = help: this indicates a potential bug in the program: it performed an invalid operation, but the Stacked Borrows rules it violated are still experimental
    = help: see https://github.com/rust-lang/unsafe-code-guidelines/blob/master/wip/stacked-borrows.md for further information
help: <326320> was created by a SharedReadOnly retag at offsets [0x18..0x7c]
   --> components\patina_adv_logger\src\memory_log.rs:181:33
    |
181 |         let address = unsafe { (self as *const AdvLoggerInfo).byte_offset(current_offset as isize) };
    |                                 ^^^^
    = note: BACKTRACE (of the first span) on thread `memory_log::tests::create_fill_check_test`:
    = note: inside `memory_log::AdvLoggerMessageEntry::init_from_memory` at components\patina_adv_logger\src\memory_log.rs:294:9: 297:10
[...]

PR #1152, opened 2025-12-03
  patina_internal_collections: Fix UB memory read in Node fields

garybeihl:
## Description
Fixes an undefined behavior issue where `Cell::set()` reads uninitialized memory during linked list creation in `Storage::resize()`.

## Root Cause
- `Cell::set()` internally uses `mem::replace()`, which reads the old value before writing the new one.
- When `Storage::resize()` allocates new nodes and calls `build_linked_list()`, the Cell fields contain uninitialized memory.
- Reading uninitialized memory is undefined behavior, even if immediately overwritten. Unwanted compiler "optimizations" could follow.

## Impact
- Any package using patina_internal_collections.
- Potential memory corruption and non-deterministic errors
- Detected by Miri testing (issue #560)

## Fix
- Initialize all Cell fields using `ptr::write()` before `build_linked_list()`
- Use `addr_of_mut!()` to read field pointers without creating references to uninitialized data

## Introduced by
- PR #1050 (Nov 13, 2025) Replace AtomicPtr with Cell in patina_internal_collections
- `AtomicPtr::store()` writes without reading the old value, but `Cell::set()` uses `mem::replace()` which reads before writing

Related to #560

- [X] Impacts functionality?
- [X] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- Tested with: `cargo +nightly-2025-09-19 miri test -p patina_dxe_core`.
- 7 tests now pass (previously 0/469 due to this UB issue).

## Integration Instructions

N/A


core/patina_internal_collections/src/node.rs
```diff
@@ -204,6 +204,31 @@ where
         // When current capacity is 0, we just need to copy the data and build the available list
         if self.capacity() == 0 {
             self.data = buffer;
+
+            // Initialize all Cell fields to prevent reading uninitialized memory.
+            // Cell::set() internally uses mem::replace() which reads the old value before writing.
+            // We use ptr::write to initialize the fields without creating references to uninitialized data.
+            for i in 0..self.data.len() {
```
os-d: Couldn't this be:

```rust
for node in buffer {
  node.color = Cell::new(BLACK);
  node.parent = Cell::new(null);
  ...
}
```

i.e. we can avoid the pointer manipulation?

joschock: Also, recommend making the values consistent with the defaults in Node::new(). i.e. `Cell::new(RED)`.

core/patina_internal_collections/src/node.rs
```diff
@@ -204,6 +204,31 @@ where
...
+                // SAFETY: node_ptr is derived from self.data which is a valid slice with length i+1.
+                // We use addr_of_mut! to get field pointers without creating intermediate references
+                // to uninitialized data, then ptr::write to initialize the fields.
+                unsafe {
```
joschock: I think this needs to be done in all cases, not just the capacity == 0 case. In fact, I think it might need to be done before the `let buffer`  at the beginning of the call.

garybeihl: Yes - I agree. We also need to update the test for this. Currently, it only takes the capacity==0 path,

core/patina_internal_collections/src/node.rs
garybeihl: > Couldn't this be:
> 
> ```rust
> for node in buffer {
>   node.color = Cell::new(BLACK);
>   node.parent = Cell::new(null);
>   ...
> }
> ```
> 
> i.e. we can avoid the pointer manipulation?

But then in that case buffer still contains uninitialized memory, right? So `for node in buffer` creates `&mut Node<D>` references to uninitialized memory, which is still UB, even if immediately overwritten. `node.color = Cell::new(BLACK)` creates a reference `&mut Node<D>` and then the assignment operator reads the old value, just like `mem::replace` reads uninitialized memory. We need to get a raw pointer to the field without creating a reference to uninitialized memory, hence the pointer manipulation. I agree it's not very intuitive...

os-d: Does miri complain on this? Maybe a case for MaybeUninit so that doesn't happen? I'm speculating.

Javagedes: @garybeihl 

I need to look at this more in depth, but if this is the solution, it should probably be completed in `Self::build_linked_list`. This function is already looping through each uninitialized node to set up the linked list of available nodes for allocations. A simple solution would just be to initialize it there while setting up the linked list. It also stops the need to writing null_ptr to the left and right nodes because we are setting them (except for the first and last node in the linked list)

os-d: I had thought the same thing, but `build_linked_list` gets called when the list is expected to be initialized, it looks like.

Javagedes: also @joschock I think this was an issue before your change. My implementation always had these nodes as only partially initialized because the logic was I only need the linked list of available nodes (so left and right ptr). Nothing else was necessary. When we go to use a node, we popped it and fully initialized it for use.

So theoretically it is safe, but if misused, would definitely be UB. So adding this initialization is for the best.

Javagedes: > I had thought the same thing, but `build_linked_list` gets called when the list is expected to be initialized, it looks like.

Nah, `build_linked_list` does not need to be initialized because `build_linked_list` is not reading any data from the initialized node. It is only setting the left and right pointers.

joschock: I think this code creates a mut ref to the slice without initializing the nodes within is technically UB already. So I don't think you can defer it to `build_linked_list` unless you defer creation of the slice to there. 
```
let buffer = unsafe {
            slice::from_raw_parts_mut::<'a, Node<D>>(
                slice as *mut [u8] as *mut Node<D>,
                slice.len() / mem::size_of::<Node<D>>(),
            )
        };
```

Javagedes: @garybeihl can you post the actual full miri test failure?

core/patina_internal_collections/src/node.rs
```diff
@@ -204,6 +204,31 @@ where
...
+                    let node_ptr = self.data.as_mut_ptr().add(i);
```
Javagedes: Another option is you could require D implements default then just do this:

```rust
for i in 0..self.data.len() {
    unsafe { self.data.as_mut_ptr().add(i).write(Node::new(D::default())); }
}
```

core/patina_internal_collections/src/node.rs
garybeihl: The full backtrace miri test output is >250K. The gist is that `resize` calls `build_linked_list` on uninitialized data (`buffer`). `build_linked_list` calls `Cell::set_right` which calls `Cell::set` on uninitialized data. `Cell::set` does a read before write which triggers the UB.

To reproduce the error, install miri with
```bash
    rustup component add miri --toolchain nightly-2025-09-19
```
 Run the test with Stacked Borrows disabled to see the definite UB:                                                                                        
```bash
MIRIFLAGS="-Zmiri-backtrace=full -Zmiri-disable-stacked-borrows" \
cargo +nightly-2025-09-19 miri test -p patina_dxe_core --lib \
allocate_deallocate_test                                                     │
```

core/patina_internal_collections/src/node.rs
garybeihl: That would work I think, but you'd be changing the API by adding the `D: Default` trait and you'd waste some CPU with all the new calls to `Default`. If that is what the consensus favors, however I can try to implement that path.

joschock: Without something like `Default`, it's not clear how to initialize `D`. Since it's generic, we can't assume that filling it with zeros is legit, for example. The UB is not just about accessing uninitialized data; I suspect it would be possible to construct a `D` such that the compiler would make poor choices based on assumptions of initialization that are not upheld, even without a direct read to the corresponding uninit memory. I think the read is just where Miri happens to detect the UB. 

If we don't do `Default`, we might have to do `data: Option<D>` so that we can make `data` the well-defined value of `None`, and then make the slice with uninitialized nodes:
```rust
        let buffer = unsafe {
            let unint_buffer = slice::from_raw_parts_mut::<<'a, Node<D>>(
                slice as *mut [u8] as *mut <'a, Node<D>,
                slice.len() / mem::size_of::<Node<D>>(),
            )
            for i in 0..unint_buffer .len() {
              // new_empty makes a Node with data = None, and everything else defaults.
              unint_buffer .as_mut_ptr().add(i).write(Node::new_empty()); } 
            }
        };
```

garybeihl: Shall I redo the PR switching to the `D: Default` trait method?

joschock: I want other opinions, but mine is that `Default` is unnecessarily constraining on `D` when what we are trying to solve is essentially an internal problem of the Node implementation. I would go with the `Option<D>` approach instead. 

I also need to study more, but I suspect we need to make the `unint_buffer` a slice of `MaybeUninit<Node<D>>` as well, and initialize it before transmuting it to a slice of `Node<D>`. I _think_ (though my confidence is not great) that simply doing a slice from raw parts on uninitialized `Node<D>` might be UB.

Javagedes: There are many options I would like to investigate before considering `Node<Option<D>>`. Using option is going to have multiple issues:

1. I'm guessing that is going to get you stuck with more `Cell` wrappers because you may not have mutable access to the node to set `D` once we've initialized it.
2. It's going to make you have to unwrap D even though we know it exists
3. It is most likely going to change the memory layout as rust will need to add a discernment for the option.

I would like to see usage of MaybeUninit at all levels investigated, e.g. `MaybeUninit<[Node<D>]>`, `&[MaybeUninit<Node<D>>]`, and even `Node<MaybeUnit<D>>`. At a bare minimum, MaybeUninit is guaranteed to keep the same memory layout and alignment.

Javagedes:
Hi @garybeihl , I noticed you've been keeping this PR up to date, but not working on it (which is completely fine). 

I just wanted to make sure you were not waiting on additional information from any of us that have commented on this PR (i.e. make sure we are not blocking you in any way)!

garybeihl:
No - not blocked - I uploaded a version of the changes that uses MaybeUninit and D: Default - just waiting for further comments or change requests. If you could have a look when you get a chance, that would be great - thanks!

joschock (approved):

core/patina_internal_collections/src/node.rs
```diff
@@ -188,17 +188,38 @@ where
...
+
+        // Initialize all nodes in the buffer to prevent undefined behavior.
+        // Cell::set() internally uses mem::replace() which reads the old value before writing.
+        // By fully initializing each Node with Node::new(D::default()), we ensure all fields
+        // (including the data field) are in a well-defined state before any operations.
+        for elem in uninit_buffer.iter_mut() {
+            // SAFETY: We're initializing uninitialized memory using MaybeUninit::write,
+            // which is the safe way to initialize MaybeUninit without reading the old value.
+            elem.write(Node::new(D::default()));
+        }
```
makubacki, line 227: Given that below all of the old nodes are going to be copied over:

```rust
for i in 0..self.len() {
    buffer[i].data = old.data;
}
```

Is the temporary initialization from the `Default` implementation needed for those nodes? Can this just be:

```rust
for elem in &mut uninit_buffer[self.len()..] {
    elem.write(Node::new(D::default()));
}
```

While this is a minor point, it does take away some variability due to the unknown complexity of the `Default` implementation.

core/patina_internal_collections/src/node.rs
makubacki, line 1: I know this is pre-existing but the code around Line 262 looks suspicious:

```rust
let idx = if !self.available.get().is_null() { self.idx(self.available.get()) } else { self.len() };
```

Is it guaranteed that `idx == self.len() == buffer.len()` will not be true?

If that's true, then these buffer accesses would be out of bounds:

```rust
        Self::build_linked_list(&buffer[idx..]);
        self.available.set(buffer[idx].as_mut_ptr());
```

core/patina_internal_collections/src/node.rs
```diff
@@ -23,7 +23,7 @@ pub const fn node_size<D: SliceKey>() -> usize {
 /// A on-stack storage container for the nodes of a red-black tree.
 pub(crate) struct Storage<'a, D>
 where
-    D: SliceKey,
+    D: SliceKey + Default,
```
makubacki: My understanding is that `resize()` is all that requires a `Default` implementation.

If that's the case, does it make sense to just place the trait bound there? That would simplify cases that don't need to resize. For example:

```rust
impl<'a, D> Storage<'a, D>
where
    D: SliceKey + Copy + Default,
{
    /// Resizes the storage container to a new slice of memory.
    ///
    /// # Trait Requirements
    ///
    /// This method requires `D: Default` to initialize new nodes. If your type
    /// doesn't implement Default, you cannot resize the storage.
    pub fn resize(&mut self, slice: &'a mut [u8]) {
        ...
    }
}
```

core/patina_internal_collections/src/node.rs
Javagedes: Overall I do not think `Default` is the right way to go. It adds an unnecessary requirement just to satisfy miri. I think the better route is to have each node's `D` wrapped in a maybeuninit.

core/patina_internal_collections/src/node.rs
Javagedes: Ooh this is a good bug.

If idx is `self.len()`, then the `Self::build_linked_list(&buffer[idx..]);` is fine, but `self.available.set(buffer[idx].as_mut_ptr());` is 100% out of bounds.

makubacki: Since this unrelated to the changes currently in this PR, I broke it out into a separate PR here - https://github.com/OpenDevicePartnership/patina/pull/1243.

core/patina_internal_collections/src/node.rs
garybeihl: yes - good point - with_capacity also requires Default - I updated the commit to reflect this.

core/patina_internal_collections/src/node.rs
garybeihl: ok - thanks for your feedback - I've reworked the code to use the maybeunint wrapper on the Node data field in place of the Default trait implementation. Thank you for looking at it.

core/patina_internal_collections/src/node.rs
garybeihl: Based on javagedes' feedback in another conversation for this PR, I've removed the Default trait implementation and replaced with a MaybeUninit wrapper approach.

garybeihl:
Is there anything more for me to do here? If not, can I get another approval?

makubacki (approved):
@Javagedes & @joschock should have a final review as well.

garybeihl:
Ping - please review so I can close this and move on to other Miri-found issues. Thanks!

core/patina_internal_collections/src/node.rs
makubacki: Does this need to go through `MaybeUnit<Node<D>>` too?

core/patina_internal_collections/src/node.rs
```diff
@@ -467,7 +486,7 @@ pub struct Node<D>
 where
     D: SliceKey,
 {
-    pub data: D,
+    pub data: MaybeUninit<D>,
```
makubacki: Does this still need to be `pub` with functions like `pub unsafe fn data()` and `pub unsafe fn data_mut()` having been added? Or could it at least be `pub(crate)`?

core/patina_internal_collections/src/node.rs
```diff
@@ -216,7 +229,13 @@ where
         for i in 0..self.capacity() {
```
makubacki: Is `for i in 0..self.capacity()` supposed to be `for i in 0..self.len()`? That's what the safety comment seems to indicate below.

core/patina_internal_collections/src/node.rs
```diff
@@ -587,12 +646,15 @@ where
...
-        self.data.key()
+        // SAFETY: This method is only called on nodes that are in use (initialized).
+        // Nodes in the available list are never accessed for their key.
+        unsafe { self.data().key() }
     }
 }
 
 #[cfg(test)]
 #[coverage(off)]
+#[allow(clippy::undocumented_unsafe_blocks)]
```
makubacki: We're generally trying to provide safety comments more granularly (even for tests) to avoid needed to apply this at the module level.

core/patina_internal_collections/src/node.rs
garybeihl: Good catch — yes, `expand()` has the same problem. It creates a `&mut [Node<D>]` from raw bytes without going through `MaybeUninit`, which is UB for the same reason. I'll update it to use `MaybeUninit<Node<D>>` as well.

core/patina_internal_collections/src/node.rs
garybeihl: You're right — nothing outside the crate accesses the field directly; all external access goes through the `data()` and `data_mut()` accessors. I'll change it to `pub(crate)`.

PR #1153, opened 2025-12-03
[REBASE&FF] Log bad components rather than panic

Javagedes:
## Description

This change updates the logic of component initialization to track a component that failed to initialize, and log it towards the end of boot, rather than panic on debug builds or possibly have UB on release builds.

Example output:

```
INFO - WARN - Components not dispatched:
INFO - WARN - ------------------------------- -------------------------------------------------
INFO - WARN - name                            error message
INFO - WARN - qemu_q35_dxe_core::BadComponent ConflictingParam Param Conflicts with everything.
```

**Note**: While tests still need to be written, this is ready for review.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

1. CI passes
2. Created a test parameter that alwys fails to initialize. Ensured it was logged during boot
3. Q35 continues to boot.

## Integration Instructions

N/A

### DIFF
--- a/sdk/patina/src/component/metadata.rs
+++ b/sdk/patina/src/component/metadata.rs
@@ -21,14 +21,14 @@ pub struct MetaData {
     access: Access,
     /// The name of the component.
     name: Cow<'static, str>,
-    /// the name of the last param that failed to be set.
-    last_failed_param: Option<Cow<'static, str>>,
+    /// The error message preventing the component from being dispatched.
+    error_message: Option<Cow<'static, str>>,
 }
 
 impl MetaData {
     /// Creates a new metadata object for a component.
     pub fn new<S>() -> Self {
-        Self { access: Access::new(), name: Cow::from(core::any::type_name::<S>()), last_failed_param: None }
+        Self { access: Access::new(), name: Cow::from(core::any::type_name::<S>()), error_message: None }
     }
 
     /// Returns the name of the component, including the module path.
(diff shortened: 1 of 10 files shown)


sdk/patina/src/component/metadata.rs
```diff
@@ -12,39 +12,41 @@
...
     pub fn new<S>() -> Self {
-        Self { access: Access::new(), name: core::any::type_name::<S>(), last_failed_param: None }
+        Self { access: Access::new(), name: Cow::from(core::any::type_name::<S>()), error_message: None }
     }
 
     /// Returns the name of the component, including the module path.
     #[inline(always)]
-    pub fn name(&self) -> &'static str {
-        self.name
+    pub fn name(&self) -> Cow<'static, str> {
```
makubacki, line 36: I haven't worked much with `Cow`, but is it necessary to always return `Cow<'static, str>` with a manual clone versus returning `&Cow<'static, str>`?

Javagedes: The main benefit of Cow (Copy on Write) is zero cost copies. A Cow is similar to your recent Guid type. It is an enum of an owned and a Borrow. Doing a <Cow>.copy() is the same as returning a reference. Only when you need to mutate the Cow will it do a new allocation, mutate it, and turn itself from a Cow::Borrowed to a Cow::owned.

PR #1154, opened 2025-12-04
patina_debugger/x64: Fix FPU and ST register index calculation

makubacki:
## Description

Correct the register ID offset calculation for FPU (registers 29-35) and ST (registers 36-44) to properly map to array indices 0-6 and 0-8 respectively.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- `cargo make all`

## Integration Instructions

- N/A


makubacki:
Noticed in pre-existing code while reviewing https://github.com/OpenDevicePartnership/patina/pull/1151.

PR #1156, opened 2025-12-04
[AArch64] Fixing a SPI interrupt ID miscalculation

kuqin12:
## Description

The SPI interrupt ID range comes after the PPI range, which starts from 32. The current logic compensates this by subtracting the the starting index, however the subtraction expression is incorrect.

This change fixed the index miscalculation.

In addition, it also removes 2 redundant steps for setting the priority as it is already covered in gic setup step.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This was tested on both QEMU SBSA and proprietary hardware platform, verified that the ARM watchdog timer can fire properly through the corresponding interrupt handler.

## Integration Instructions

N/A

### DIFF
--- a/sdk/patina/tests/compile_fail_tests.rs
+++ b/sdk/patina/tests/compile_fail_tests.rs
@@ -20,4 +20,8 @@ fn compile_fail_tests() {
     t.compile_fail("tests/ui/duplicate_commands.rs");
     t.compile_fail("tests/ui/duplicate_boot_services.rs");
     t.compile_fail("tests/ui/duplicate_runtime_services.rs");
+    t.compile_fail("tests/ui/duplicate_storage_mut.rs");
+    t.compile_fail("tests/ui/storage_and_storage_mut_conflict.rs");
+    t.compile_fail("tests/ui/storage_mut_and_storage_conflict.rs");
+    t.pass("tests/ui/multiple_storage_allowed.rs");
 }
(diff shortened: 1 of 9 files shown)


core/patina_internal_cpu/src/interrupts/aarch64/gic_manager.rs
```diff
@@ -159,12 +159,6 @@ impl AArch64InterruptInitializer {
 
         // Disable all interrupts and set priority to 0x80.
         gic_v3.enable_all_interrupts(false);
-        for i in IntId::private() {
```
joschock, line 162: is priority set not needed?

makubacki:
Fyi: @kuqin12, to help better correlate the change to affected consumers, I added a "[AArch64]" prefix to the PR description.

core/patina_internal_cpu/src/interrupts/aarch64/gic_manager.rs
kuqin12: It is needed. But this is already done as part of the `git_v3.setup` call.

kuqin12: Actually, they also disabled all interrupts as well. I will try to remove the line above and test.

kuqin12: Tested on both virtual and physical hardware. The disable all interrupts are not needed to be called once again after setup call.

PR #1157, opened 2025-12-04
[REBASE & FF] Implement reloading the core through the debugger

cfernald:
### patina_dxe_core: Add support for reloading the core from the debugger

This commit implements a new feature that allows the Patina core to be
reloaded from the debugger without needing a full system reboot or reflash.
This is particularly useful for development and debugging, as it enables
rapid iteration on core changes. This feature is intended to be coordinated
from the UefiExt debugger extension. This feature is experimental and
is likely to evolve and change.

### patina_debugger: Send nack packet on debugger reset

When a error occurs, it may because a packet was a packet was corrupted
or partially received. In this case, it is better to send a NACK packet
to the host to request a retransmission, rather than sending a stop code.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [x] Includes documentation?

## How This Was Tested

- [x] Q35
- [x] SBSA
- [x] Intel Platform

## Integration Instructions

Platforms may add the `debugger_reload` feature from `patina_dxe_core` to enable this feature.


core/patina_debugger/src/debugger.rs
```diff
@@ -211,10 +216,15 @@ impl<T: SerialIO> PatinaDebugger<T> {
...
         let mut timeout_reached = false;
         if let GdbStubStateMachine::Idle(mut inner) = gdb {
-            // Always start with a stop code if starting from idle. This may be because this is the initial breakpoint
+            // If this is a restart, send a nack to request a resend of the failing packet.
+            // Otherwise, always start with a stop code if starting from idle. This may be because this is the initial breakpoint
             // or because the initial breakpoint timed out. This is not to spec, but is a useful hint to the client
             // that a break has occurred. This allows the debugger to reconnect on scenarios like reboots.
-            let _ = inner.borrow_conn().write_all(GDB_STOP_PACKET.as_bytes());
+            if restart {
+                self.transport.write(GDB_NACK_PACKET);
```
os-d: Why `self.transport.write` here vs the `inner.borrow_con.write_all()` for the other one?

docs/src/dev/debugging/core_reload.md
```diff
@@ -0,0 +1,56 @@
+# Core Reload
+
+Debugger core reload is an experimental feature that allows a new Patina core to be loaded and executed from the initial
+breakpoint of the existing core. This it intended to be used for rapid development where flashing the device is time
+consuming or logistically difficult.
+
+## Enabling Core Reload
+
+To enable core reload, the debugger must first be setup and enabled. Additionally, the `debugger_reload` feature must
```
os-d:
```suggestion
To enable core reload, the debugger must first be set up and enabled. Additionally, the `debugger_reload` feature must
```

docs/src/dev/debugging/core_reload.md
```diff
@@ -0,0 +1,56 @@
+# Core Reload
+
+Debugger core reload is an experimental feature that allows a new Patina core to be loaded and executed from the initial
+breakpoint of the existing core. This it intended to be used for rapid development where flashing the device is time
+consuming or logistically difficult.
+
+## Enabling Core Reload
+
+To enable core reload, the debugger must first be setup and enabled. Additionally, the `debugger_reload` feature must
+be enabled for the `patina_dxe_core` crate. When both of these prerequisites are true, core reload should be functional
```
os-d:
```suggestion
be enabled for the `patina_dxe_core` crate. When both of these prerequisites are true, core reload is functional
```

Let's be confident! :)

docs/src/dev/debugging/core_reload.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Enabling Core Reload
+
+To enable core reload, the debugger must first be setup and enabled. Additionally, the `debugger_reload` feature must
+be enabled for the `patina_dxe_core` crate. When both of these prerequisites are true, core reload should be functional
+at the initial breakpoint.
+
+## Reloading the Core
+
+Reloading the core should be done using a debugger extension. Currently this is only support in windbg using the
```
os-d:
```suggestion
Reloading the core should be done using a debugger extension. Currently this is only supported in WinDbg using the
```

core/patina_debugger/src/debugger.rs
```diff
@@ -35,7 +35,8 @@ use crate::{
 const GDB_BUFF_LEN: usize = 0x2000;
 
 /// A default GDB stop packet used when entering the debugger.
-const GDB_STOP_PACKET: &str = "$T05thread:01;#07";
+const GDB_STOP_PACKET: &[u8] = "$T05thread:01;#07".as_bytes();
```
Javagedes: I think you can just do b"<thing>" instead of .as_bytes(), but I'm not positive

docs/src/dev/debugging/core_reload.md
```diff
@@ -0,0 +1,56 @@
...
+[UefiExt extension](https://github.com/microsoft/uefi_debug_tools/tree/main/UefiDbgExt). When the extension is installed,
+the feature is enabled, and the debugger is broken in on the initial breakpoint, then you can use the `!uefiext.reload`
```
os-d: nit: we can talk about this on the UefiExt PR, but I think `!reload` is a bit confusing because of the built in `.reload` in WinDbg that is about symbol reloading. I would consider another name, perhaps `!loadcore` or something more descriptive.

docs/src/dev/debugging/core_reload.md
```diff
@@ -0,0 +1,56 @@
...
+kd> !uefiext.reload D:\patina-dxe-core-qemu\target\x86_64-unknown-uefi\debug\qemu_q35_dxe_core.efi
+Loading image of size 1982464 bytes from D:\patina-dxe-core-qemu\target\x86_64-unknown-uefi\debug\qemu_q35_dxe_core.efi
+Compressed image size: 738544 bytes
+Transferred 738544 / 738544 bytes (100%) in 20 seconds (288 kbps)...
+Image loaded successfully.
+... symbols reload prints ...
+Executing new core...
+```
+
+If the debugger is enabled int he new core, you will see a new break-in on the initial breakpoint. Otherwise it will
```
os-d:
```suggestion
If the debugger is enabled in the new core, you will see a new break-in on the initial breakpoint. Otherwise it will
```

patina_dxe_core/src/debugger_reload.rs
```diff
@@ -0,0 +1,266 @@
...
+use patina::{
+    component::service::memory::{AllocationOptions, MemoryManager, PageAllocationStrategy},
+    guids,
+    pi::hob::{self},
+    uefi_size_to_pages,
+};
+
+use crate::{memory_manager::CoreMemoryManager, pecoff};
+
+static HOB_LIST_ADDRESS: core::sync::atomic::AtomicUsize = core::sync::atomic::AtomicUsize::new(0);
```
Javagedes, line 27: Any way to not use statics?

patina_dxe_core/src/debugger_reload.rs
```diff
@@ -0,0 +1,266 @@
...
+    if TEAR_DOWN.load(core::sync::atomic::Ordering::SeqCst) {
+        let _ = writeln!(out, "Only supported at initial breakpoint");
+        return;
+    }
+
+    match args.next() {
+        Some("alloc_buffer") => {
+            // get the requested length of the prep buffer.
+            let buffer_size = match args.next() {
+                Some(size_str) => size_str.parse::<usize>().unwrap_or(0),
```
Javagedes: Is this a good use for zero copy?

docs/src/dev/debugging/core_reload.md
```diff
@@ -0,0 +1,56 @@
...
+continue execution. The `/nogo` option provides the ability to load the new core but not begin execution. This can be
+useful for early core debugging, though caution should be used as debugger exception handler transitions may be unstable.
+
+## UART Considerations
+
+The core reload feature requires a large amount of rabid data transfer between the host and the device. Depending on the
```
os-d:
```suggestion
The core reload feature requires a large amount of rapid data transfer between the host and the device. Depending on the
```
It probably is rabid, but I think this is what you meant :)

cfernald:
Related: https://github.com/microsoft/uefi_debug_tools/pull/12

docs/src/dev/debugging/core_reload.md
```diff
@@ -0,0 +1,56 @@
...
+UART implementation and support for flow control, this has the potential to overflow hardware FIFO. `ComToTcpServer.py`
```
os-d: Worth giving link to the script here

patina_dxe_core/src/debugger_reload.rs
```diff
@@ -0,0 +1,266 @@
+//! Module for adding a debugger reload command.
+//!
+//! This subsystem provides the ability to reload the core DXE image via the debugger.
```
os-d:
```suggestion
//! This subsystem provides the ability to reload the DXE core image via the debugger.
```

patina_dxe_core/src/debugger_reload.rs
```diff
@@ -0,0 +1,266 @@
...
+    let decompressed_image = alloc.leak_as_slice::<u8>();
+
+    if let Err(error) =
+        decompress_into_with_algo(compressed_image, decompressed_image, DecompressionAlgorithm::UefiDecompress)
+    {
+        let _ = writeln!(out, "Failed to decompress image: {:?}", error);
+        return None;
+    }
+
+    Some((decompressed_image.as_ptr() as usize, decompressed_size))
```
Javagedes: Why not return a reference to the buffer slice? Then you don't need to send the size manually.

patina_dxe_core/src/debugger_reload.rs
```diff
@@ -0,0 +1,266 @@
...
+}
+
+unsafe fn core_reload(address: usize, size: usize, out: &mut dyn core::fmt::Write) {
```
Javagedes: And accept the slice here

patina_dxe_core/src/debugger_reload.rs
Javagedes: Or at a minimum NonNull<[u8]> which also self manages the size

patina_dxe_core/src/debugger_reload.rs
```diff
@@ -0,0 +1,266 @@
...
+        Ok(info) => info,
+        Err(err) => {
+            let _ = writeln!(out, "Failed to parse PE image: {:?}", err);
+            return;
+        }
+    };
+
+    // Step 1: allocate the image memory.
+    let image_size = pe_info.size_of_image as usize;
+    let alloc = match CoreMemoryManager.allocate_pages(
```
Javagedes, line 195: Why? This memory stopped being tracked once we start the image anyway, and you already have the bytes

patina_dxe_core/src/debugger_reload.rs
```diff
@@ -0,0 +1,266 @@
...
+                        let _ = writeln!(out, "Failed to allocate reload buffer: {:?}", err);
+                        return;
+                    }
+                };
+
+            let address = pages.into_raw_ptr::<u8>().unwrap() as usize;
+            let _ = write!(out, "{:x}", address);
+        }
+        Some("load") => {
+            // get the requested length of the prep buffer.
```
os-d:
```suggestion
            // get the address of the prep buffer.
```

patina_dxe_core/src/debugger_reload.rs
Javagedes: Or is this the loaded image size? Might be that, and I didn't realize.

patina_dxe_core/src/debugger_reload.rs
```diff
@@ -0,0 +1,266 @@
...
+
+                if module_hob.module_name == guids::DXE_CORE {
+                    module_hob.alloc_descriptor.memory_base_address = core_address as u64;
+                    module_hob.alloc_descriptor.memory_length = core_buffer_size as u64;
+                    module_hob.entry_point = entry_point as u64;
+                    fixed_up_core = true;
+                } else if module_hob.alloc_descriptor.name == guids::HOB_MEMORY_ALLOC_STACK {
+                    stack_ptr = (module_hob.alloc_descriptor.memory_base_address
+                        + module_hob.alloc_descriptor.memory_length) as usize
+                        - 0x18;
```
os-d: Worth a comment on why the - 0x18

patina_dxe_core/src/debugger_reload.rs
```diff
@@ -0,0 +1,266 @@
...
+    }
+
+    // SAFETY: The caller is responsible for ensuring the address and size are valid.
+    let compressed_image = unsafe { core::slice::from_raw_parts(address as *const u8, size) };
+
+    // Get the decompressed size. By spec if will be the second u32 in the compressed image.
+    let decompressed_size = {
+        let size_bytes = &compressed_image[4..8];
+        u32::from_le_bytes(size_bytes.try_into().unwrap()) as usize
+    };
```
makubacki, line 169: I think the code is a little more clear (and an `unwrap()` is avoided) with:

```rust
    let decompressed_size = u32::from_le_bytes([
        compressed_image[4],
        compressed_image[5], 
        compressed_image[6],
        compressed_image[7],
    ]) as usize;
```

patina_dxe_core/src/debugger_reload.rs
```diff
@@ -0,0 +1,266 @@
...
+            let pages =
+                match CoreMemoryManager.allocate_pages(uefi_size_to_pages!(buffer_size), AllocationOptions::new()) {
+                    Ok(pages) => pages,
+                    Err(err) => {
+                        let _ = writeln!(out, "Failed to allocate reload buffer: {:?}", err);
+                        return;
+                    }
+                };
+
+            let address = pages.into_raw_ptr::<u8>().unwrap() as usize;
```
makubacki: nit: A message. Something like:

```suggestion
            let address = pages.into_raw_ptr::<u8>()
                .expect("u8 size check should failed after allocate_pages()") as usize;
```

patina_dxe_core/src/debugger_reload.rs
```diff
@@ -0,0 +1,266 @@
...
+static TEAR_DOWN: core::sync::atomic::AtomicBool = core::sync::atomic::AtomicBool::new(false);
+
+#[cfg(target_arch = "x86_64")]
+const ARCH_ALLOCATION_STRATEGY: PageAllocationStrategy = PageAllocationStrategy::MaxAddress(0xFFFF_FFFF);
+#[cfg(target_arch = "aarch64")]
+const ARCH_ALLOCATION_STRATEGY: PageAllocationStrategy = PageAllocationStrategy::Any;
+
+/// Initializes the debugger reload command subsystem.
+pub fn initialize_debugger_reload(hob_list: *const c_void) {
+    HOB_LIST_ADDRESS.store(hob_list as usize, core::sync::atomic::Ordering::SeqCst);
```
makubacki: I'm just curious. Was there a particular reason for choosing `SeqCst` for this scenario?

docs/src/dev/debugging/core_reload.md
cfernald: Fair, any suggestions? `core_reload`, `exec`, `core_exec`, `core_jump`? Naming things is hard.

patina_dxe_core/src/debugger_reload.rs
```diff
@@ -0,0 +1,266 @@
...
+    if physical_hob_list == 0 {
+        return Err("Original HOB list address is zero");
+    }
+
+    let mut next_hob = physical_hob_list as *mut hob::header::Hob;
+    let mut stack_ptr = 0usize;
+    let mut fixed_up_core = false;
+    // SAFETY: The hob list should be valid. Otherwise this is an exceedingly bad idea, but this is a debugger
+    //         operation so buckle up.
+    unsafe {
```
makubacki: It would be ideal to have this `unsafe` block smaller in size. Perhaps by splitting out functionality into smaller unsafe functions with respective safety invariants called out.

docs/src/dev/debugging/core_reload.md
```diff
@@ -0,0 +1,56 @@
+# Core Reload
+
+Debugger core reload is an experimental feature that allows a new Patina core to be loaded and executed from the initial
+breakpoint of the existing core. This it intended to be used for rapid development where flashing the device is time
```
makubacki:
```suggestion
breakpoint of the existing core. This is intended to be used for rapid development where flashing the device is time
```

docs/src/dev/debugging/core_reload.md
makubacki: Personally, I like `core_reload` from current options

Javagedes: `reload_core` sounds decent to me.
 Maybe also have a !rc macro (or whatever it's called) if possible

os-d: I think any of those is fine, it was just the name aliasing that was troubling to me. Of those, probably `core_reload` is the most intuitive to me (maybe even more so in verb - noun form, e.g. `reload_core` or `loadcore`), but I'm fine with any of them.

docs/src/dev/debugging/core_reload.md
```diff
@@ -0,0 +1,56 @@
...
+If these occur, consider changing the settings in `ComToTcpServer.py` (if used) and/or implementing flow control.
+
+## How It Works
+
+The reload occurs in several steps coordinated via the debugger monitor command between the debugger
+(specifically UefiExt) and the DXE core. The steps are as follows:
+
+1. The debug extension will load the specified file and compress it for faster transfer.
+2. The extension uses core reload monitor commands to allocate a buffer in the DXE core of the
+required size to hold the compressed image. The address of this buffer is returned to the debugger.
```
makubacki: nit: For consistency with the way (6) is currently formatteed.


```suggestion
   required size to hold the compressed image. The address of this buffer is returned to the debugger.
```

docs/src/SUMMARY.md
```diff
@@ -36,6 +36,7 @@
 - [Debugging](dev/debugging.md)
   - [Windbg Debugging](dev/debugging/windbg_debugging.md)
   - [Windbg Debugging Example](dev/debugging/windbg_example.md)
+  - [Core Reload](devl/debugging/core_reload.md)
```
makubacki:
```suggestion
  - [Core Reload](dev/debugging/core_reload.md)
```

docs/src/dev/debugging/core_reload.md
Javagedes: Load_core may be better as reload core kind of makes me think we are "reloading" the same core instead of "loading" a new core. But it's incredibly minor and pedantic that I don't think it truly matters.

patina_dxe_core/src/debugger_reload.rs
makubacki: Also, it seems like a more explicit error would be nice if `size_str` parsing fails (falling back to the "Buffer size must be greater than 0" is not quite the same as a parsing failure that may have lead to defaulting to `0` here).

core/patina_debugger/src/debugger.rs
cfernald: No reason, forgot to change then when I rebased this and moved it from before the inner was available. Will fix.

patina_dxe_core/src/debugger_reload.rs
```diff
@@ -0,0 +1,266 @@
...
+    uefi_size_to_pages,
+};
+
+use crate::{memory_manager::CoreMemoryManager, pecoff};
+
+static HOB_LIST_ADDRESS: core::sync::atomic::AtomicUsize = core::sync::atomic::AtomicUsize::new(0);
+static TEAR_DOWN: core::sync::atomic::AtomicBool = core::sync::atomic::AtomicBool::new(false);
+
+#[cfg(target_arch = "x86_64")]
+const ARCH_ALLOCATION_STRATEGY: PageAllocationStrategy = PageAllocationStrategy::MaxAddress(0xFFFF_FFFF);
```
joschock, line 33: can you add a comment documenting why this is needed for x86? it will be useful for posterity

docs/src/dev/debugging/core_reload.md
cfernald: I'm going with `!uefiext.loadcore` typing underscores during commands is annoying.

patina_dxe_core/src/debugger_reload.rs
cfernald: Unfortunately, not really. The debugger calls come back asynchronously and without context/structures. We could redesign the monitor callback to accept an instanced type instead of a function pointer to carry context, but I don't think that is something I would want generally.

cfernald: Though I suppose we could make this part of the new singleton core context thing.

patina_dxe_core/src/debugger_reload.rs
cfernald: Nope, changing to relaxed

patina_dxe_core/src/debugger_reload.rs
cfernald: @Javagedes , I don't think I'm following. How is zerocopy helpful here?

patina_dxe_core/src/debugger_reload.rs
cfernald: Multiple reason. First, this is the expanded image which will be larger then the flat image. Second, the allocation for the final location of the core may have specific requirements, such as being <4gb for x64.

patina_dxe_core/src/debugger_reload.rs
Javagedes: I'd say it's not important for this PR. Can re evaluate the design which actually making the debugger as a whole not static.

patina_dxe_core/src/debugger_reload.rs
Javagedes: You are manually pulling out data into type from a buffer, which is the purpose of zero copy. 

If it is only the singular usize then it may not be worth it. But it was just a Thought.

patina_dxe_core/src/debugger_reload.rs
cfernald: Yeah, you caught me being lazy. Will fix.

patina_dxe_core/src/debugger_reload.rs
cfernald: This is converting a string of a number into a usize, so I don't think it applies, but let me know if I'm misunderstanding.

Javagedes: Ah. No it's me that is misunderstanding :)

makubacki (approved):

patina_dxe_core/src/debugger_reload.rs
```diff
@@ -0,0 +1,275 @@
...
+fn reload_monitor(args: &mut core::str::SplitWhitespace<'_>, out: &mut dyn core::fmt::Write) {
+    if TEAR_DOWN.load(core::sync::atomic::Ordering::Relaxed) {
+        let _ = writeln!(out, "Only supported at initial breakpoint");
+        return;
+    }
+
+    match args.next() {
+        Some("alloc_buffer") => {
+            // get the requested length of the prep buffer.
+            let buffer_size = match args.next() {
```
Javagedes: nit - imo you should break out the handling of the `alloc_buffer` command into a new function (same for `load`). Better organization / maintainability in the long run, and a pretty small thing to do.

Open to not doing it though. Just a small nitpick.

PR #1165, opened 2025-12-05
[AArch64] Move hardware interrupt handlers to be guarded by per-handler RwLock

kuqin12:
## Description

The current code guards all the interrupt handlers in a common TPL mutex. This will have 2 side effects:

1. All interrupt handlers will run under the TPL defined by the handler mutex, which is set to TPL_HIGH. This will prevent a lot of fundamental functionalities from core to panic;
2. All interrupt IDs will be guarded by the same mutex, which could potentially run into re-entrant mutex access if a different interrupt comes in while the first interrupt is still being processed.

This change moves the hardware interrupt handler to be guarded using per-handler RwLock to resolve above issues.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This change is tested on QEMU SBSA and physical platform.

## Integration Instructions

N/A

### DIFF
--- a/patina_dxe_core/src/cpu/hw_interrupt_protocol.rs
+++ b/patina_dxe_core/src/cpu/hw_interrupt_protocol.rs
@@ -1,5 +1,5 @@
 use crate::tpl_mutex::TplMutex;
-use alloc::{boxed::Box, vec, vec::Vec};
+use alloc::{boxed::Box, format, vec, vec::Vec};
 use core::ffi::c_void;
 use patina_internal_cpu::interrupts::{
     ExceptionContext, InterruptHandler, InterruptManager, gic_manager::AArch64InterruptInitializer,
@@ -407,7 +407,7 @@ impl InterruptHandler for HwInterruptProtocolHandler {
 
         let rw_handler = self.handlers[raw_value as usize]
             .try_read()
-            .expect("Failed to read lock in exception handler for interrupt ID {}!", raw_value);
+            .expect(&format!("Failed to read lock in exception handler for interrupt ID 0x{:x}", raw_value));
 
         if let Some(handler) = *rw_handler {
             handler(raw_value as u64, context);
@@ -436,8 +436,8 @@ impl HwInterruptProtocolHandler {
 
         let m_handler = handler as *const c_void;
 
-        // If the handler is a null pointer, return invalid parameter
-        if let Some(mut rw_handler) = self.handlers[interrupt_source as usize].try_write() {
+        if let Some(rw_handler) = self.handlers[interrupt_source].try_read() {
+            // Use read access to test the state of the handler
             if m_handler.is_null() && (*rw_handler).is_none() {
                 return efi::Status::INVALID_PARAMETER;
             }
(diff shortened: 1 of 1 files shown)


cfernald (approved):

patina_dxe_core/src/cpu/hw_interrupt_protocol.rs
```diff
@@ -418,38 +422,46 @@ impl InterruptHandler for HwInterruptProtocolHandler {
...
         }
 
         let m_handler = handler as *const c_void;
 
         // If the handler is a null pointer, return invalid parameter
-        if m_handler.is_null() & self.handlers.lock()[interrupt_source].is_none() {
-            return efi::Status::INVALID_PARAMETER;
-        }
+        if let Some(mut rw_handler) = self.handlers[interrupt_source as usize].try_write() {
+            if m_handler.is_null() & (*rw_handler).is_none() {
```
makubacki:
```suggestion
            if m_handler.is_null() && (*rw_handler).is_none() {
```

patina_dxe_core/src/cpu/hw_interrupt_protocol.rs
```diff
@@ -418,38 +422,46 @@ impl InterruptHandler for HwInterruptProtocolHandler {
...
+                return efi::Status::INVALID_PARAMETER;
+            }
 
-        if !m_handler.is_null() & self.handlers.lock()[interrupt_source].is_some() {
-            return efi::Status::ALREADY_STARTED;
-        }
+            if !m_handler.is_null() & (*rw_handler).is_some() {
```
makubacki:
```suggestion
            if !m_handler.is_null() && (*rw_handler).is_some() {
```

patina_dxe_core/src/cpu/hw_interrupt_protocol.rs
```diff
@@ -404,7 +405,10 @@ impl InterruptHandler for HwInterruptProtocolHandler {
             return;
         }
 
-        if let Some(handler) = self.handlers.lock()[raw_value as usize] {
+        let rw_handler =
+            self.handlers[raw_value as usize].try_read().expect("Failed to read lock in exception handler!");
```
makubacki: It'd probably be helpful to print the interrupt ID in the panic message.

makubacki (approved):

patina_dxe_core/src/cpu/hw_interrupt_protocol.rs
```diff
@@ -418,38 +422,46 @@ impl InterruptHandler for HwInterruptProtocolHandler {
...
-            self.aarch64_int.lock().disable_interrupt_source(interrupt_source as u64)
+            // If the interrupt handler is unregistered then disable the interrupt
+            if m_handler.is_null() {
+                *rw_handler = None;
+                if let Err(err) = self.aarch64_int.lock().disable_interrupt_source(interrupt_source as u64) {
+                    return err.into();
+                }
+            } else {
+                *rw_handler = Some(handler);
+                if let Err(err) = self.aarch64_int.lock().enable_interrupt_source(interrupt_source as u64) {
```
joschock: you might want to drop the lock before you enable the interrupt source, otherwise an interrupt could immediately occur and try and get the handler and panic on the try_read.

patina_dxe_core/src/cpu/hw_interrupt_protocol.rs
```diff
@@ -418,38 +422,46 @@ impl InterruptHandler for HwInterruptProtocolHandler {
...
+            }
 
-        // If the interrupt handler is unregistered then disable the interrupt
-        let result = if m_handler.is_null() {
-            self.handlers.lock()[interrupt_source as usize] = None;
-            self.aarch64_int.lock().disable_interrupt_source(interrupt_source as u64)
+            // If the interrupt handler is unregistered then disable the interrupt
+            if m_handler.is_null() {
+                *rw_handler = None;
+                if let Err(err) = self.aarch64_int.lock().disable_interrupt_source(interrupt_source as u64) {
```
joschock: I think you should disable the interrupt before you write the new empty handler to avoid a scenario where an interrupt occurs after the handler is `None` but before the interrupt is disabled.

kuqin12: Thanks, good point. The flow is rearranged.

patina_dxe_core/src/cpu/hw_interrupt_protocol.rs
kuqin12: Agreed, the enable is moved out of the try_write block and should be called after the lock is dropped now.

patina_dxe_core/src/cpu/hw_interrupt_protocol.rs
kuqin12: This is updated.

os-d:
Is there a similar pattern on the x86 side that needs to be fixed?

kuqin12:
> Is there a similar pattern on the x86 side that needs to be fixed?

@cfernald can better comment on this. The exception handlers already use RwLock to manage exceptions for both x86 and ARM64. This hardware interrupt protocol is a specific handler that handles IRQ exceptions specifically, but I don’t see an equivalent mechanism on the x86 side.

cfernald:
> > Is there a similar pattern on the x86 side that needs to be fixed?
> 
> @cfernald can better comment on this. The exception handlers already use RwLock to manage exceptions for both x86 and ARM64. This hardware interrupt protocol is a specific handler that handles IRQ exceptions specifically, but I don’t see an equivalent mechanism on the x86 side.

@os-d, Kun is correct. x86's interrupt differentiation happens mainly on the initial interrupt vector through it's 256 entry table. ARM's initial table is only a few (sync, async, serror) and the individual interrupt differentiation happens in this second layer. The first layer used by both architectures was already using this pattern.

PR #1166, opened 2025-12-05
Fixes for RSDP and XSDT allocation

berlin-with0ut-return:
## Description
The RSDP cannot be created with `AcpiTable::new()` due to its unique header format.

Similar, we should discontinue use of `Box::new_in` and `Vec::new_in` due to `#allocator_api` being poorly supported and unstable. 

- [x] Impacts functionality? (ACPI functionality still the same. Fixes bugs.)
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested
Boot on QEMU to Windows.

## Integration Instructions
N/A.

### DIFF
--- a/components/patina_acpi/src/acpi.rs
+++ b/components/patina_acpi/src/acpi.rs
@@ -201,7 +201,9 @@ where
         if let Some(fadt_table) = self.acpi_tables.write().get_mut(&Self::FADT_KEY) {
             // SAFETY: We verify the table's signature before calling `install_facs`.
             let facs_addr = unsafe { facs_info.as_ref::<AcpiFacs>() } as *const AcpiFacs as u64;
+            // SAFETY: The struct maintains an invariant mapping between the FADT and `Self::FADT_KEY`.
             unsafe { fadt_table.as_mut::<AcpiFadt>().set_x_firmware_ctrl(facs_addr) };
+            // SAFETY: The struct maintains an invariant mapping between the FADT and `Self::FADT_KEY`.
             unsafe { fadt_table.as_mut::<AcpiFadt>().inner._firmware_ctrl = facs_addr as u32 };
             fadt_table.update_checksum(ACPI_CHECKSUM_OFFSET)?;
         }
@@ -253,14 +255,14 @@ where
         }
 
         // Read the header to validate the XSDT signature is valid.
-        // SAFETY: `xsdt_address` has been validated to be non-null.
         let xsdt_header = rsdp.xsdt_address as *const AcpiTableHeader;
+        // SAFETY: `xsdt_address` has been validated to be non-null.
         if (unsafe { *xsdt_header }).signature != signature::XSDT {
             return Err(AcpiError::InvalidSignature);
         }
 
-        // SAFETY: We validate that the XSDT is non-null and contains the right signature.
         let xsdt_ptr = rsdp.xsdt_address as *const AcpiTableHeader;
+        // SAFETY: We validate that the XSDT is non-null and contains the right signature.
         let xsdt = unsafe { &*(xsdt_ptr) };
 
         if xsdt.length < ACPI_HEADER_LEN as u32 {
@@ -281,6 +283,7 @@ where
                 return Err(AcpiError::InvalidSignature);
             }
 
+            // SAFETY: The tables in the HOB should be valid ACPI tables.
             let facs_table = unsafe {
                 AcpiTable::new(facs_from_ptr, self.memory_manager.get().ok_or(AcpiError::ProviderNotInitialized)?)?
             };
@@ -295,6 +298,7 @@ where
                 return Err(AcpiError::InvalidSignature);
             }
 
+            // SAFETY: The tables in the HOB should be valid ACPI tables.
             let dsdt_table = unsafe {
                 AcpiTable::new(dsdt_from_ptr, self.memory_manager.get().ok_or(AcpiError::ProviderNotInitialized)?)?
             };
@@ -322,6 +326,7 @@ where
             // SAFETY: The caller must ensure that the XSDT in the HOB points to valid table entries.
             let tbl_header = unsafe { *(entry_addr as *const AcpiTableHeader) };
             // Because we are installing from raw pointers, information about the type of the table cannot be extracted.
+            // SAFETY: The tables in the hob should be valid ACPI tables.
             let mut table = unsafe {
                 AcpiTable::new(tbl_header, self.memory_manager.get().ok_or(AcpiError::ProviderNotInitialized)?)?
             };
@@ -352,16 +357,20 @@ where
         // If the FACS is already installed, update the FADT's x_firmware_ctrl field.
         // If not, it will be updated when the FACS is installed.
         if let Some(facs) = self.acpi_tables.read().get(&Self::FACS_KEY) {
+            // SAFETY: The struct maintains an invariant mapping between the FADT and `Self::FADT_KEY`.
             unsafe { fadt_info.as_mut::<AcpiFadt>() }.inner.x_firmware_ctrl = facs.as_ptr() as u64;
+
             // Set the 32-bit pointer to the FACS. (This is an ACPI 1.0 legacy field.)
             // Ideally this would not be necessary, but the current Windows OS implementation relies on this field.
             // This workaround can be removed when Windows no longer relies on these fields.
+            // SAFETY: The struct maintains an invariant mapping between the FADT and `Self::FADT_KEY`.
             unsafe { fadt_info.as_mut::<AcpiFadt>() }.inner._firmware_ctrl = facs.as_ptr() as u32;
         }
 
         // If the DSDT is already installed, update the FACP's x_dsdt field.
         // If not, it will be updated when the DSDT is installed.
         if let Some(dsdt) = self.acpi_tables.read().get(&Self::DSDT_KEY) {
+            // SAFETY: The struct maintains an invariant mapping between the FADT and `Self::FADT_KEY`.
             unsafe { fadt_info.as_mut::<AcpiFadt>() }.inner.x_dsdt = dsdt.as_ptr() as u64;
         }
 
@@ -401,6 +410,7 @@ where
         // If the FADT is already installed, update the FACP's x_dsdt field.Add commentMore actions
         // If not, it will be updated when the FACP is installed.
         if let Some(facp) = self.acpi_tables.write().get_mut(&Self::FADT_KEY) {
+            // SAFETY: The struct maintains an invariant mapping between the FADT and `Self::FADT_KEY`.
             unsafe { facp.as_mut::<AcpiFadt>() }.inner.x_dsdt = dsdt_info.as_ptr() as u64;
             facp.update_checksum(ACPI_CHECKSUM_OFFSET)?;
         };
@@ -494,7 +504,7 @@ where
                         .with_memory_type(EfiMemoryType::ACPIReclaimMemory),
                 )
                 .map_err(|_e| AcpiError::AllocationFailed)?;
-            let xsdt_ptr = xsdt_allocation.into_raw_ptr().ok_or(AcpiError::AllocationFailed)? as *mut u8;
+            let xsdt_ptr = xsdt_allocation.into_raw_ptr().ok_or(AcpiError::AllocationFailed)?;
             let xsdt_addr = xsdt_ptr as u64;
             // SAFETY: The allocated call above allocated enough memory for the new XSDT.
             let new_xsdt_slice: &'static mut [u8] = unsafe { core::slice::from_raw_parts_mut(xsdt_ptr, num_bytes_new) };
@@ -503,6 +513,16 @@ where
             new_xsdt_slice[0..num_bytes_original].copy_from_slice(&xsdt_data.slice[0..num_bytes_original]);
             // Fill in trailing space with zeros so it is accessible (Vec length != Vec capacity).
             new_xsdt_slice[num_bytes_original..].fill(0);
+
+            // Free the previous buffer.
+            // SAFETY: The address and size are correctly stored in the `AcpiXsdtMetadata` structure.
+            unsafe {
+                self.memory_manager
+                    .get()
+                    .ok_or(AcpiError::ProviderNotInitialized)?
+                    .free_pages(xsdt_data.slice.as_ptr() as usize, uefi_pages_to_size!(num_bytes_original))
+                    .map_err(|_e| AcpiError::FreeFailed)
+            }?;
 
             // Update the RSDP with the new XSDT address.
             if let Some(ref mut rsdp) = *self.rsdp.write() {
@@ -543,20 +563,27 @@ where
             signature::FACS => {
                 // Clear out the FACS pointer in the FADT.
                 if let Some(fadt_table) = self.acpi_tables.write().get_mut(&Self::FADT_KEY) {
+                    // SAFETY: The FADT signature has been verified before calling `delete_table`.
+                    // The struct maintains an invariant mapping between the FADT and `Self::FADT_KEY`.
                     unsafe { fadt_table.as_mut::<AcpiFadt>() }.set_x_firmware_ctrl(0);
+
                     // Also clear out the 32-bit pointer to the FACS. (This is an ACPI 1.0 legacy field.)
                     // Ideally this would not be necessary, but the current Windows OS implementation relies on this field.
                     // This workaround can be removed when Windows no longer relies on these fields.
+
+                    // SAFETY: The FADT signature has been verified before calling `delete_table`.
+                    // The struct maintains an invariant mapping between the FADT and `Self::FADT_KEY`.
                     unsafe { fadt_table.as_mut::<AcpiFadt>() }.inner._firmware_ctrl = 0;
                     fadt_table.update_checksum(ACPI_CHECKSUM_OFFSET)?;
                 }
 
                 // Free the FACS memory.
+                // SAFETY: The FACS was allocated in ACPI Reclaim memory using page allocation. It has a well-defined size.
                 unsafe {
                     self.memory_manager
                         .get()
                         .ok_or(AcpiError::ProviderNotInitialized)?
-                        .free_pages(physical_addr as usize, 1)
+                        .free_pages(physical_addr as usize, uefi_pages_to_size!(mem::size_of::<AcpiFacs>()))
                         .map_err(|_| AcpiError::FreeFailed)
                 }?;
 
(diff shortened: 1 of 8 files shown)


Javagedes:
While I agree with removing the allocator_api functionality as it has not been developed in years, the claim that Box::new_in or Vec::new_in causes unpredictable allocations is very, very concerning.

Can you provide a reproducible example of this? All this feature should be doing is calling our Allocator trait implementation, instead of the global allocators Allocator trait implementation. If you are seeing something wrong the it is most likely a bug in our own Allocator implementation. 

Also the MemoryManager service uses the allocator trait from the allocato_api service, so if there is a bug in that unstable feature, I'm not 100% sure containing to use it would work.

berlin-with0ut-return:
> While I agree with removing the allocator_api functionality as it has not been developed in years, the claim that Box::new_in or Vec::new_in causes unpredictable allocations is very, very concerning.
> 
> Can you provide a reproducible example of this? All this feature should be doing is calling our Allocator trait implementation, instead of the global allocators Allocator trait implementation. If you are seeing something wrong the it is most likely a bug in our own Allocator implementation.
> 
> Also the MemoryManager service uses the allocator trait from the allocato_api service, so if there is a bug in that unstable feature, I'm not 100% sure containing to use it would work.

the example that caused me to track down this behavior was as follows (i have only included the first few lines of the memory map):

1. print memory map before ACPI entry point:
```
INFO - Reserved Memory          0x0                  0x0             0x50            UC|WC|WT|WB
INFO - BootServicesData         0x50000              0x0             0x1             UC|WC|WT|WB
INFO - Conventional Memory      0x51000              0x0             0x4F            UC|WC|WT|WB
INFO - Conventional Memory      0x100000             0x0             0x740           UC|WC|WT|WB
INFO - BootServicesData         0x840000             0x0             0xDC0           UC|WC|WT|WB
INFO - Conventional Memory      0x1600000            0x0             0x799BE         UC|WC|WT|WB
INFO - BootServicesData         0x7AFBE000           0x0             0x20            UC|WC|WT|WB
INFO - Conventional Memory      0x7AFDE000           0x0             0x6             UC|WC|WT|WB
INFO - BootServicesData         0x7AFE4000           0x0             0x60            UC|WC|WT|WB|UCE|WP
INFO - Conventional Memory      0x7B044000           0x0             0x1E8C          UC|WC|WT|WB|UCE|WP
INFO - BootServicesData         0x7CED0000           0x0             0x900           UC|WC|WT|WB|UCE|WP
INFO - RuntimeServicesData      0x7D7D0000           0x0             0x700           0x800000000000101F
INFO - RuntimeServicesCode      0x7DED0000           0x0             0x100           0x800000000000101F
INFO - Reserved Memory          0x7DFD0000           0x0             0x510           UC|WC|WT|WB|UCE|WP
**INFO - ACPI Reclaim Memory      0x7E4E0000           0x0             0x20            UC|WC|WT|WB|UCE|WP**
```

2. allocate XSDT (292 bytes) (as you can see in the logs the allocator is somehow calculating that an additional 0x100 pages are required for AcpiReclaim even though this should easily fit inside the existing AcpiReclaim region at 0x7E4E0000).
```
INFO - TRACE - C:\Users\sherryfan\patina\patina_dxe_core\src\allocator\fixed_size_block_allocator.rs:783: required_pages: 0x100
INFO - TRACE - C:\Users\sherryfan\patina\patina_dxe_core\src\gcd\spin_locked_gcd.rs:1999: Memory region 0x7cdd0000 of length 0x100000 mapped with attributes MemoryAttributes(
INFO -     Writeback | ExecuteProtect,
INFO - )
INFO - TRACE - C:\Users\sherryfan\patina\patina_dxe_core\src\allocator\fixed_size_block_allocator.rs:808: Allocated additional memory at address: 0x7cdd0000
INFO - WARN - XSDT allocated: address=0x7cdd0200, capacity=292 bytes
```

3. print memory map after XSDT allocation:
```
INFO - Type                     Physical Start       Virtual Start   Number of Pages Attributes
INFO - Reserved Memory          0x0                  0x0             0x50            UC|WC|WT|WB
INFO - BootServicesData         0x50000              0x0             0x1             UC|WC|WT|WB
INFO - Conventional Memory      0x51000              0x0             0x4F            UC|WC|WT|WB
INFO - Conventional Memory      0x100000             0x0             0x740           UC|WC|WT|WB
INFO - BootServicesData         0x840000             0x0             0xDC0           UC|WC|WT|WB
INFO - Conventional Memory      0x1600000            0x0             0x799BE         UC|WC|WT|WB
INFO - BootServicesData         0x7AFBE000           0x0             0x20            UC|WC|WT|WB
INFO - Conventional Memory      0x7AFDE000           0x0             0x6             UC|WC|WT|WB
INFO - BootServicesData         0x7AFE4000           0x0             0x60            UC|WC|WT|WB|UCE|WP
INFO - Conventional Memory      0x7B044000           0x0             0x1D8C          UC|WC|WT|WB|UCE|WP
**INFO - ACPI Reclaim Memory      0x7CDD0000           0x0             0x100           UC|WC|WT|WB|UCE|WP**
INFO - BootServicesData         0x7CED0000           0x0             0x900           UC|WC|WT|WB|UCE|WP
INFO - RuntimeServicesData      0x7D7D0000           0x0             0x700           0x800000000000101F
INFO - RuntimeServicesCode      0x7DED0000           0x0             0x100           0x800000000000101F
INFO - Reserved Memory          0x7DFD0000           0x0             0x510           UC|WC|WT|WB|UCE|WP
INFO - ACPI Reclaim Memory      0x7E4E0000           0x0             0x20            UC|WC|WT|WB|UCE|WP
     
```

as you can see, a 1MB extra region was allocated for AcpiReclaim even though there should be 0x20 pages already available. this then causes the system to reset to fix up the AcpiReclaim memory bucket and it also claims ~0x150 pages for AcpiReclaim when it should be much less (around 25 total pages should be enough for all ACPI tables).

since it happens so early in boot and right after the XSDT allocation and goes away after Vec::new_in is replaced with allocate_pages, i assumed that was the issue. a similar pattern occurs for the RSDP.

Javagedes:
Did you debugger through the allocation flow and watch what happened? Or check the capacity of the Vec capacity (not size) after each call?Both the Vec logic and our allocator logic do pre-allocations. Vec, to ensure we don't have to reallocate on every push, and our Allocators, to ensure we don't fragment memory too much.

The allocator_api feature is just a facilitator. All it does is expose a new trait, implement that trait for the global allocator, and add a few methods for all the alloc types to use a custom allocator instead of our allocator. 

I still truly think there is a bug in our code if the above is happening that needs to be root caused.

makubacki:
I reproduced this with a simple unit test. It looks like the issue might be that `FixedSizeBlockAllocator::fallback_alloc()` finds `self.allocators` is `None` so `FixedSizeBlockAllocatorError::OutOfMemory` is returned and then `Allocator::allocate()` applies `MIN_EXPANSION` (1MB):

```rust
let mut allocation_size = max(additional_mem_required, MIN_EXPANSION);
```

os-d:
> I reproduced this with a simple unit test. It looks like the issue might be that `FixedSizeBlockAllocator::fallback_alloc()` finds `self.allocators` is `None` so `FixedSizeBlockAllocatorError::OutOfMemory` is returned and then `Allocator::allocate()` applies `MIN_EXPANSION` (1MB):
> 
> ```rust
> let mut allocation_size = max(additional_mem_required, MIN_EXPANSION);
> ```

I definitely could see that in the unit test case, but during boot ACPI reclaim should be one of the bucket types that has an allocator because we allocated the bucket at the very beginning. Maybe I’m confusing what you mean by what is going wrong.

As for allocator_api it isn’t deprecated, just unstable. It doesn’t appear remotely close to stabilization, but the alternative is still in RFC form and going to be years away from stabilization, so we are stuck with this for a long time.

makubacki:
> I definitely could see that in the unit test case, but during boot ACPI reclaim should be one of the bucket types that has an allocator because we allocated the bucket at the very beginning. Maybe I’m confusing what you mean by what is going wrong.
> 
> As for allocator_api it isn’t deprecated, just unstable. It doesn’t appear remotely close to stabilization, but the alternative is still in RFC form and going to be years away from stabilization, so we are stuck with this for a long time.

My thought based on the test results and the size being returned is that in the allocator this connection might need to be made - https://github.com/OpenDevicePartnership/patina/pull/1167. I thought just putting it in a draft PR is easier than explaining.

Javagedes:
This is not a memory type with a static allocator. It is dynamic and gets created on first use. I can't verify but my guess is when allocating through the memorymanager through the normal alloc flow implemented via the Allocator trait, we hit this None case and allocating the 1MB. When allocating via a direct allocate_pages call, we do not hit this None case

makubacki:
> This is not a memory type with a static allocator. It is dynamic and gets created on first use. I can't verify but my guess is when allocating through the memorymanager through the normal alloc flow implemented via the Allocator trait, we hit this None case and allocating the 1MB. When allocating via a direct allocate_pages call, we do not hit this None case

That's it and what the previously linked PR addresses.

os-d:
My point is that this shouldn’t be the first use, so the allocator should exist. But let me look at the PR, I’m on my phone currently :)

berlin-with0ut-return:
> My point is that this shouldn’t be the first use, so the allocator should exist. But let me look at the PR, I’m on my phone currently :)

if i'm understsanding your point correctly my ACPI component is the first use of AcpiReclaim. 

in the original code the first call to `get_allocator` in the entry point causes a new `FixedSizeBlockAllocator` to be created with memory type AcpiReclaim. at this point both `self.allocators` and `self..list_heads` are none/empty. on the subsequent `alloc` call triggered by Vec allocation it would run into the `fallback_alloc` case and then the None `self.allocators` case as Michael mentioned above. 

so i guess the main problem is that the FixedSizeBlockAllocator doesn't "know" about the existing 0x20 AcpiReclaim region which i believe #1167 correctly addresses

os-d:
> > My point is that this shouldn’t be the first use, so the allocator should exist. But let me look at the PR, I’m on my phone currently :)
> 
> if i'm understsanding your point correctly my ACPI component is the first use of AcpiReclaim.
> 
> in the original code the first call to `get_allocator` in the entry point causes a new `FixedSizeBlockAllocator` to be created with memory type AcpiReclaim. at this point both `self.allocators` and `self..list_heads` are none/empty. on the subsequent `alloc` call triggered by Vec allocation it would run into the `fallback_alloc` case and then the None `self.allocators` case as Michael mentioned above.
> 
> so i guess the main problem is that the FixedSizeBlockAllocator doesn't "know" about the existing 0x20 AcpiReclaim region which i believe #1167 correctly addresses

Well, there are two things:
- This shouldn't be the first use of ACPI reclaim, it should have had an initial bucket memory associated with it, which creates the allocator (Michael is looking into that, per his comment on his PR)
- I'm still not following why an existing ACPI reclaim region exists and why it should be part of the allocator. If it was from a memory allocation HOB, say, then it shouldn't be owned by the allocator, because the allocator would think it was free memory. I would like to understand what the existing region is from.

makubacki:
Sorry, I had high priority work to get done and haven't been able to return to that PR yet. I'll sync with Oliver offline.

joschock:
The FSB allocator is designed for high-traffic pool-style allocations - i.e. it expects that you are going to be lots of allocates/frees of various sizes ala traditional malloc/free. So - if it doesn't have memory to satisfy an allocation, it's gonna grab a large chunk (as discussed above) because it assumes that you're going to be using it and freeing it a ton. Grabbing a chunk from the GCD requires more work because of the RBT manipulations - which are reasonably fast, but pool allocations from a primed pool are effectively constant time. Thus the pool allocator wants to minimize the number of times GCD interactions happen - thus the large expansion size. 

For a fixed-address long term stable large allocation (like an ACPI table), the pool allocator (via Box::new_in() and friends) is not really an efficient choice. A direct page allocation is going to be better in that scenario at least in the current implementation. 

One possible approach here would be to parameterize the allocators to allow for tuning the expansion size based on memory type. Really the only one that likely has a significant perf impact is EfiBootServicesData - the rest could probably have a min expansion size of 0 and not really affect performance that much.

makubacki:
> Really the only one that likely has a significant perf impact is EfiBootServicesData - the rest could probably have a min expansion size of 0 and not really affect performance that much.

I've looked at this a bit more and I do see allocations falling in the reserved region. However, 1MB minimum expansion is grabbing a lot so I'm leaning toward this solution too.

@berlin-with0ut-return, as John mentioned, you should likely stay with direct page allocation here.

@joschock, if you're open to some adjustments to the minimum expansion size for other memory types, I'm happy to update my other PR to focus on a proposal for that.

joschock:
> @joschock, if you're open to some adjustments to the minimum expansion size for other memory types, I'm happy to update my other PR to focus on a proposal for that.

That seems reasonable to me.

components/patina_acpi/src/acpi.rs
```diff
@@ -113,8 +116,9 @@ where
     }
 
     /// Sets up tracking for the RSDP internally.
-    pub fn set_rsdp(&self, rsdp_table: AcpiTable) {
-        self.acpi_tables.write().insert(Self::RSDP_KEY, rsdp_table);
+    pub fn set_rsdp(&self, rsdp: &'static mut AcpiRsdp) {
```
joschock, line 119: RwLock is prone to deadlock if it is actually used in an async context (like; a timer interrupt). I don't see that anywhere in this implementation, so probably fine; but since it's a public routine, might be worth a doc comment for callers to be aware that it is not safe to invoke simultaneously from multiple contexts. If invocation from a callback is a supported usage model, then probably need to switch to a TplMutex.

components/patina_acpi/src/acpi.rs
```diff
@@ -480,27 +484,33 @@ where
             let num_bytes_new = ACPI_HEADER_LEN + new_capacity * ACPI_XSDT_ENTRY_SIZE;
 
             // The XSDT is always allocated in reclaim memory.
-            let allocator = self
+            let xsdt_allocation = self
                 .memory_manager
                 .get()
                 .ok_or(AcpiError::ProviderNotInitialized)?
-                .get_allocator(EfiMemoryType::ACPIReclaimMemory)
+                .allocate_pages(
```
makubacki, line 501: Is the previous buffer freed somewhere?

components/patina_acpi/src/acpi.rs
```diff
@@ -952,31 +941,38 @@ mod tests {
...
-        // Add some extra space after the XSDT so it's safe to write the entry.
-        xsdt_allocated_bytes.extend(core::iter::repeat(0u8).take(100));
-        let xsdt_metadata = AcpiXsdtMetadata {
-            n_entries: 0,
-            max_capacity: MAX_INITIAL_ENTRIES,
-            slice: xsdt_allocated_bytes.into_boxed_slice(),
-        };
+        let xsdt_ptr = xsdt_allocation.into_raw_ptr().unwrap() as *mut u8;
+        let header_bytes = xsdt_info.header.hdr_to_bytes();
+        unsafe {
```
makubacki, line 1005: Needs a safety comment.

components/patina_acpi/src/component.rs
```diff
@@ -92,20 +107,20 @@ impl AcpiProviderManager {
...
-        let xsdt_ptr = xsdt_allocated_bytes.as_mut_ptr();
-        let xsdt_addr = xsdt_ptr as u64;
-        let xsdt_metadata = AcpiXsdtMetadata {
-            n_entries: 0,
-            max_capacity: MAX_INITIAL_ENTRIES,
-            slice: xsdt_allocated_bytes.into_boxed_slice(),
-        };
+        // Convert the raw pointer into a Box<[u8], &'static dyn Allocator>
+        // We need to create a slice from the raw pointer and then box it with the allocator
+        let xsdt_slice: &'static mut [u8] = unsafe { core::slice::from_raw_parts_mut(xsdt_ptr, xsdt_size) };
```
makubacki, line 123: Needs safety comment.

components/patina_acpi/src/component.rs
```diff
@@ -92,20 +107,20 @@ impl AcpiProviderManager {
...
             },
         };
-        // Fill in XSDT data.
+
+        // Write the XSDT header to the allocated memory
         let header_bytes = xsdt_info.header.hdr_to_bytes();
-        xsdt_allocated_bytes.extend(header_bytes);
-        // Fill in trailing space with zeros so it is accessible (Vec length != Vec capacity).
-        xsdt_allocated_bytes.extend(core::iter::repeat(0u8).take(xsdt_capacity - ACPI_HEADER_LEN));
+        unsafe {
```
makubacki, line 114: Needs safety comment.

components/patina_acpi/src/component.rs
```diff
@@ -122,11 +137,31 @@ impl AcpiProviderManager {
...
+                patina::component::service::memory::AllocationOptions::new()
+                    .with_memory_type(EfiMemoryType::ACPIReclaimMemory),
+            )
+            .map_err(|_e| EfiError::OutOfResources)?;
+
+        // Get the raw pointer from the allocation
+        let rsdp_ptr = rsdp_allocation.into_raw_ptr().ok_or(EfiError::OutOfResources)? as *mut AcpiRsdp;
+
+        // Write the RSDP data to the allocated memory
+        unsafe {
```
makubacki, line 160: Needs safety comment.

components/patina_acpi/src/component.rs
```diff
@@ -122,11 +137,31 @@ impl AcpiProviderManager {
...
+            core::ptr::write(rsdp_ptr, rsdp_data);
+        }
+
+        // Convert the raw pointer into a Box<AcpiRsdp, &'static dyn Allocator>
+        let rsdp_allocated = unsafe { Box::leak(Box::from_raw(rsdp_ptr)) };
```
makubacki, line 166: Needs safety comment.

components/patina_acpi/src/service.rs
```diff
@@ -159,3 +163,41 @@ pub(crate) trait AcpiProvider {
...
+    use crate::acpi_table::{AcpiFacs, AcpiFadt};
+
+    use super::*;
+
+    #[test]
+    fn test_get_table_wrong_type() {
+        unsafe impl Send for AcpiTable {}
+        unsafe impl Sync for AcpiTable {}
+
+        let table = unsafe {
```
makubacki, line 189: Needs safety comment.

components/patina_acpi/src/acpi.rs
berlin-with0ut-return: it was not before, i have fixed it

components/patina_acpi/src/acpi.rs
berlin-with0ut-return: i was missing 
```
[lints]
workspace = true
```
it should catch all missed safety comments now.

PR #1167, opened 2025-12-05
patina_dxe_core: Initialize allocators with existing GCD regions

makubacki:
## Description

Initialize newly-created allocators with pre-existing memory region from the GCD that match the allocator's handle. This prevents `MIN_EXPANSION` from being triggered on to initialize the allocator which can be a large (~1MB) allocation.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- `cargo make all`
- QEMU testing in progress (in draft while testing further)

## Integration Instructions

- N/A

### DIFF
--- a/patina_dxe_core/src/allocator/fixed_size_block_allocator.rs
+++ b/patina_dxe_core/src/allocator/fixed_size_block_allocator.rs
@@ -1887,7 +1887,7 @@ mod tests {
                 DEFAULT_PAGE_ALLOCATION_GRANULARITY,
             );
 
-            // Pre-poulatee the allocator with a small existing region
+            // Pre-populate the allocator with a small existing region
             let existing_region_size = 0x20000;
             let existing_region_addr = GCD
                 .allocate_memory_space(


os-d:
The scenario here is that if you call allocate_pages, it wasn't creating the allocator? Or this is to cover memory allocation HOBs for the region?

Isn't it actually dangerous for the allocator to own pre-existing memory regions? Won't it think they are unallocated, but could be allocated?

For the case this was created for, the allocator should exist:

https://github.com/OpenDevicePartnership/patina/blob/328548b28e0340a35532f81c1c0dfad11831e209/patina_dxe_core/src/allocator.rs#L1127-L1132

Unless we aren't creating a bucket for it, which would be a bug that would break hibernate.

makubacki:
> The scenario here is that if you call allocate_pages, it wasn't creating the allocator? Or this is to cover memory allocation HOBs for the region?
> 
> Isn't it actually dangerous for the allocator to own pre-existing memory regions? Won't it think they are unallocated, but could be allocated?
> 
> For the case this was created for, the allocator should exist:
> 
> https://github.com/OpenDevicePartnership/patina/blob/328548b28e0340a35532f81c1c0dfad11831e209/patina_dxe_core/src/allocator.rs#L1127-L1132
> 
> Unless we aren't creating a bucket for it, which would be a bug that would break hibernate.

I need a bit more time to look through the flow.

patina_dxe_core/src/allocator.rs
```diff
@@ -328,10 +328,42 @@ impl AllocatorMap {
...
+    /// when suitable memory already exists.
+    fn initialize_allocator_with_existing_regions(allocator: &UefiAllocator, handle: efi::Handle) {
+        let mut descriptors: Vec<dxe_services::MemorySpaceDescriptor> =
+            Vec::with_capacity(GCD.memory_descriptor_count() + 10);
+        if GCD.get_memory_descriptors(&mut descriptors).is_err() {
+            // Just proceed without initializing if there is a problem retrieving descriptors
+            return;
+        }
+
+        for descriptor in descriptors.iter() {
```
joschock, line 353: I think you need to differentiate between free and allocated descriptors in the GCD. 

Might be easier to just call GCD.allocate_memory_space(). I believe that the GCD logic will already attempt to satisfy the allocation with pages that are assigned to this memory type if it can.

joschock:
Commenting here instead of ACPI PR since discussion is more scoped here.

One thing we could consider but would be a larger change would be to indeed consider dropping the `allocator_api` support. The system allocator trait (`GlobalAlloc`) that we really need for general heap support in rust is stable, and we can just assign that to the EBSD allocator.

We could drop the allocator_api but retain pretty much exactly the FSB/GCD implementation we have now; the main thing that would be impacted would be the `Box::new_in()`/`Vec::new_in()/LinkedList::new_in()` constructs which are not presently used extensively in the core (in a quick search, I only see them used to publish system tables and manage the runtime event/images lists). 

The downside is that would mean that allocations to places other than the EBSD heap would need to be done with explicit allocation calls, and things like Box/Vec other rust heap structures wouldn't work with them. The upside is that this would get rid of a chunk of our unstable.

I don't see this as urgently needed; but wanted to throw it out there as thought food.

os-d:
> Commenting here instead of ACPI PR since discussion is more scoped here.
> 
> One thing we could consider but would be a larger change would be to indeed consider dropping the `allocator_api` support. The system allocator trait (`GlobalAlloc`) that we really need for general heap support in rust is stable, and we can just assign that to the EBSD allocator.
> 
> We could drop the allocator_api but retain pretty much exactly the FSB/GCD implementation we have now; the main thing that would be impacted would be the `Box::new_in()`/`Vec::new_in()/LinkedList::new_in()` constructs which are not presently used extensively in the core (in a quick search, I only see them used to publish system tables and manage the runtime event/images lists).
> 
> The downside is that would mean that allocations to places other than the EBSD heap would need to be done with explicit allocation calls, and things like Box/Vec other rust heap structures wouldn't work with them. The upside is that this would get rid of a chunk of our unstable.
> 
> I don't see this as urgently needed; but wanted to throw it out there as thought food.

We had this discussion internally as well and I did have research. There is a newer version of allocator_api that is in RFC form (storage IIRC) and that sounds like new work will go towards.

The current thought process was we wanted to keep the *_in() functions working, because that does provide value to Rust developers and not having to create our own types or alloc pages for every Box/Vec etc.

Certainly open to further discussion.

makubacki:
I created this draft PR as a quick reference for the ACPI discussion. I’ll close it for now and I'll probably follow up with a change for the minimum expansion size. If someone wants to take things further, an RFC would probably be useful for broader discussion.

PR #1169, opened 2025-12-06
Use a lower minimum expansion size for non-BS Data memory types

makubacki:
## Description

Replace the `MIN_EXPANSION` constant with memory-type-specific expansion sizes to optimize memory usage. Boot Services Data allocators now use `HIGH_TRAFFIC_ALLOC_MIN_EXPANSION` (1MB), runtime memory types use `LOW_TRAFFIC_RUNTIME_ALLOC_MIN_EXPANSION` (uses runtime granularity), and other types use `LOW_TRAFFIC_ALLOC_MIN_EXPANSION` (page size / 4KB).

---

A few other changes are included to better separate these allocator types:

1. Do not have `UefiAllocator` directly depend on `SpinLockedFixedSizeBlockAllocator`. Instead take the parameter as a generic type to decouple the types.
2. Add a `PageAllocator` trait that defines the interface `UefiAllocator` needs to use an allocator. Have `SpinLockedFixedSizeBlockAllocator` implement this trait.
3. Because the `SpinLockedFixedSizeBlockAllocator` type is no longer fixed (e.g. `SpinLockedFixedSizeBlockAllocator<A>` vs `SpinLockedFixedSizeBlockAllocator<B>`), a single static array cannot be used to hold all of the static allocators. This change adds a `macro` module to hold macros that make iterating across the diverse allocator types more ergonomic.

---

A follow up change will be made to simplify overall allocator design and management, tracked in https://github.com/OpenDevicePartnership/patina/issues/1174.

---

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- `cargo make all`

## Integration Instructions

- N/A

### DIFF
--- a/patina_dxe_core/src/allocator.rs
+++ b/patina_dxe_core/src/allocator.rs
@@ -237,6 +237,14 @@ pub static EFI_RUNTIME_SERVICES_DATA_ALLOCATOR: UefiAllocatorWithFsb = UefiAlloc
     efi::RUNTIME_SERVICES_DATA,
 );
 
+pub static STATIC_ALLOCATORS: [(&UefiAllocatorWithFsb, efi::MemoryType); 5] = [
+    (&EFI_BOOT_SERVICES_DATA_ALLOCATOR, efi::BOOT_SERVICES_DATA),
+    (&EFI_LOADER_CODE_ALLOCATOR, efi::LOADER_CODE),
+    (&EFI_BOOT_SERVICES_CODE_ALLOCATOR, efi::BOOT_SERVICES_CODE),
+    (&EFI_RUNTIME_SERVICES_CODE_ALLOCATOR, efi::RUNTIME_SERVICES_CODE),
+    (&EFI_RUNTIME_SERVICES_DATA_ALLOCATOR, efi::RUNTIME_SERVICES_DATA),
+];
+
 fn memory_attributes_to_str(f: &mut core::fmt::Formatter<'_>, attributes: u64) -> core::fmt::Result {
     let mut attrs = Vec::new();
     let mut string_len = 0;
@@ -398,12 +406,11 @@ impl AllocatorMap {
     // Returns an iterator that checks all allocators by handle.
     fn find_memory_type_by_handle(&self, handle: efi::Handle) -> Option<efi::MemoryType> {
         // Check static allocators first, then dynamic allocators
-        let _ = for_each_static_allocator!(alloc => {
+        for (alloc, mem_type) in STATIC_ALLOCATORS.iter() {
             if alloc.handle() == handle {
-                return Some(alloc.memory_type());
+                return Some(*mem_type);
             }
-            false
-        });
+        }
         self.iter_dynamic().find(|x| x.handle() == handle).map(|x| x.memory_type())
     }
 
--- a/patina_dxe_core/src/allocator/macros.rs
+++ b/patina_dxe_core/src/allocator/macros.rs
@@ -9,30 +9,11 @@
 /// Macro to iterate over all static allocators and execute an expression for each.
 /// Returns `true` if any allocator returns `true` from the expression.
 /// The variable `$alloc` is available in the expression and represents each allocator.
-///
-/// # Example
-/// ```ignore
-/// if for_each_static_allocator!(alloc => alloc.total_allocated() > 0) {
-///     // At least one allocator has allocated memory
-/// }
-/// ```
 #[macro_export]
 macro_rules! for_each_static_allocator {
     ($alloc:ident => $action:expr) => {{
-        ({
-            let $alloc = &$crate::allocator::EFI_BOOT_SERVICES_DATA_ALLOCATOR;
-            $action
-        }) || ({
-            let $alloc = &$crate::allocator::EFI_LOADER_CODE_ALLOCATOR;
-            $action
-        }) || ({
-            let $alloc = &$crate::allocator::EFI_BOOT_SERVICES_CODE_ALLOCATOR;
-            $action
-        }) || ({
-            let $alloc = &$crate::allocator::EFI_RUNTIME_SERVICES_CODE_ALLOCATOR;
-            $action
-        }) || ({
-            let $alloc = &$crate::allocator::EFI_RUNTIME_SERVICES_DATA_ALLOCATOR;
+        $crate::allocator::STATIC_ALLOCATORS.iter().any(|(alloc_ref, _)| {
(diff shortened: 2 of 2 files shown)


joschock:
You could use a const-generic for this:
```rust
pub struct FixedSizeBlockAllocator<const MIN_EXPANSION:usize> {
...
}
...
impl<const MIN_EXPANSION:usize> FixedSizeBlockAllocator<MIN_EXPANSION> {
...
}

pub(crate) static EFI_BOOT_SERVICES_DATA_ALLOCATOR: UefiAllocator<SIZE_1MB> = UefiAllocator::new(
    &GCD,
    NonNull::from_ref(GCD.memory_type_info(efi::BOOT_SERVICES_DATA)),
    protocol_db::EFI_BOOT_SERVICES_DATA_ALLOCATOR_HANDLE,
    DEFAULT_PAGE_ALLOCATION_GRANULARITY,
);

etc.
```

patina_dxe_core/src/allocator/fixed_size_block_allocator.rs
```diff
@@ -767,9 +779,10 @@ unsafe impl Allocator for SpinLockedFixedSizeBlockAllocator {
                 const _: () = assert!(ALIGNMENT.is_multiple_of(align_of::<AllocatorListNode>()));
                 const _: () = assert!(ALIGNMENT.is_multiple_of(UEFI_PAGE_SIZE) && ALIGNMENT > 0);
 
-                // As a matter of policy, allocate at least `MIN_EXPANSION` memory and ensure the size is
-                // aligned to `ALIGNMENT`.
-                let mut allocation_size = max(additional_mem_required, MIN_EXPANSION);
+                // As a matter of policy, allocate at least the minimum expansion amount of memory and ensure the
+                // size is aligned to `ALIGNMENT`.
+                let mut allocation_size =
+                    max(additional_mem_required, min_expansion_for_memory_type(self.lock().memory_type()));
```
joschock: this will require taking the lock to get the memory type, which slightly (maybe imperceptibly) increases the runtime cost. It's in the expansion flow which is not fast path already; but noting that this isn't zero-cost. If you use a const generic to spec the MIN_EXPANSION size, you could avoid this.

makubacki: I'll update it to use a const generic on Monday.

os-d (approved):
While we are here, is there already a check (I think not based on the issue that led us to here) to make sure that we aren’t blowing buckets? It would be worth a print and debug assert when a bucketized allocator has to expand. Doesn’t need to be done here, but is related to the issue. That could go with a behavior change such that the bucketized allocators just hold onto the entire bucket memory and then service pool and pages directly from it and are never expected to expand.

As I’m typing this, this is a bigger change, so definitely not for here, but @joschock thoughts on this?

patina_dxe_core/src/allocator/fixed_size_block_allocator.rs
makubacki: Updated

joschock:
> While we are here, is there already a check (I think not based on the issue that led us to here) to make sure that we aren’t blowing buckets? It would be worth a print and debug assert when a bucketized allocator has to expand. Doesn’t need to be done here, but is related to the issue. 

On the platform side we have https://github.com/microsoft/mu_plus/blob/05554be4a2ee7cc35f401c46492e02746c323c9d/MsCorePkg/Library/MemoryTypeInformationChangeLib/MemoryTypeInformationChangeLib.c#L25 to report bucket changes in log and telemetry, and we write tasks for various projects on the regular to tune the buckets to match. 

While we definitely need tools to allow us to tune the buckets, these exist. So I'm not sure it is worth making the allocators more complex or adding further constraints on their behavior is worth it.

makubacki:
I'm going to explore simply storing the min expansion size in a field in the allocator to reduce some complexity around the static type management for the const generic in the current PR. As I added the commit for the const generic type parameter onto the original commit in the PR branch, I'll do the same for this to show the delta for the different approaches.

makubacki:
> I'm going to explore simply storing the min expansion size in a field in the allocator to reduce some complexity around the static type management for the const generic in the current PR. As I added the commit for the const generic type parameter onto the original commit in the PR branch, I'll do the same for this to show the delta for the different approaches.

Pushed in the latest commit

patina_dxe_core/src/allocator.rs
```diff
@@ -70,61 +94,150 @@ cfg_if::cfg_if! {
...
+            page_free_calls: 0,
+            reserved_size: 0,
+            reserved_used: 0,
+            claimed_pages: 0,
+        }
+    }
+}
+
+/// The interface needeed for an allocator used by UefiAllocator.
+pub trait PageAllocator: GlobalAlloc + Allocator + Display + Sync + Send {
```
joschock: naming nit: `PageAllocator` describes this trait but since the trait also requires mandatory support for `GlobalAlloc` and `Allocator` traits it also supports sub-page allocations. A casual read of the name elsewhere in the code could lead to mistaken understanding that it is really only supporting page-level allocations. Perhaps consider something like `GeneralAllocator` or `MemoryAllocator` or something that doesn't imply a size or allocation strategy might be more natural.

joschock: or you maybe could decouple it from requiring the other traits and move them up to the UefiAllocator Level e.g.
```rust
pub struct UefiAllocator<A> 
where A: PageAllocator + GlobalAlloc + Allocator + Display + Sync + Send
...
```

patina_dxe_core/src/allocator.rs
```diff
@@ -252,19 +365,156 @@ impl Debug for MemoryDescriptorSlice<'_> {
...
+        for allocator in ALLOCATORS.lock().iter_dynamic() {
+            if allocator.memory_type() == memory_type {
+                return allocator.get_memory_ranges();
+            }
+        }
+        Vec::new()
+    })
+}
+
+/// An allocator reference that represents a static allocator.
```
joschock: Since they are all the same type now that the expansion is a field, is this still needed?

patina_dxe_core/src/allocator.rs
makubacki: I moved the trait bounds to `UefiAllocator`.

patina_dxe_core/src/allocator.rs
makubacki: I was able to remove `AllocatorRef`.

PR #1173, opened 2025-12-08
Rework memory_attributes_table to remove the MEMORY_ATTRIBUTES_TABLE static

joschock:
## Description

Remove `MEMORY_ATTRIBUTES_TABLE` global static - the pointer in question is stored in the global configuration table. Instead of keeping a duplicate copy, use the entry in the global table.
* adjust `core_install_configuration_table` return value to include the old pointer if the table is removed or modified.
* add `get_configuration_table` API to return a pointer from the configuration table for the given vendor guid.
* rework MAT logic to use the config table exclusively.
* Replace POST_RTB atomic bool with Once and add a `TestOnce` capability for test support to allow resetting the Once for testing purposes.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Passes unit tests; booted on AARCH64 HW, checked MAT table as dumped in UEFI logs. 

## Integration Instructions

N/A

### DIFF
--- a/patina_dxe_core/src/config_tables/memory_attributes_table.rs
+++ b/patina_dxe_core/src/config_tables/memory_attributes_table.rs
@@ -95,9 +95,6 @@ pub fn init_memory_attributes_table_support() {
 // After this point, subsequent runtime memory allocations/deallocations will create new MAT tables
 extern "efiapi" fn core_install_memory_attributes_table_event_wrapper(event: efi::Event, _context: *mut c_void) {
     core_install_memory_attributes_table();
-    // now we want to capture any future runtime memory changes, so we will mark that ReadyToBoot has occurred
-    // and the install callback will be invoked on the next runtime memory allocation
-    POST_RTB.call_once(|| {});
 
     if let Err(status) = EVENT_DB.close_event(event) {
         log::error!("Failed to close MAT ready to boot event with status {status:#X?}. This should be okay.");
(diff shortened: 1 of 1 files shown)


patina_dxe_core/src/config_tables/memory_attributes_table.rs
```diff
@@ -8,34 +8,29 @@
...
 pub struct MemoryAttributesTable(*mut efi::MemoryAttributesTable);
 
 // this is a flag to indicate that we have passed ReadyToBoot and can install the MAT on the next runtime memory
-// allocation/deallocation
-static POST_RTB: AtomicBool = AtomicBool::new(false);
+// allocation/deallocation.
+
+// `RwLock` is used here (instead of `Once`) to facilitate unit test flows resetting the value. Outside of unit-test,
+// usage is expected to be a single write false->true during boot.
+static POST_RTB: RwLock<bool> = RwLock::new(false);
```
os-d: It seems to me it would be better to have a cfg(test) RwLock vs cfg(not(test)) Once to make sure we have the right behavior. To have consistent behavior, maybe having a struct with a single member and different impls for test/not test to get/set.

It’s not a big deal, but I also don’t like test code forcing us to have looser guarantees in prod code.

patina_dxe_core/src/config_tables/memory_attributes_table.rs
```diff
@@ -100,19 +95,15 @@ extern "efiapi" fn core_install_memory_attributes_table_event_wrapper(event: efi
...
     }
 }
 
 pub fn core_install_memory_attributes_table() {
-    let mut st_guard = systemtables::SYSTEM_TABLE.lock();
-    let st = st_guard.as_mut().expect("System table support not initialized");
-
-    let current_ptr = MEMORY_ATTRIBUTES_TABLE.load(Ordering::Relaxed);
-    if current_ptr.is_null() {
+    if get_configuration_table(&efi::MEMORY_ATTRIBUTES_TABLE_GUID).is_none() {
```
os-d: I don’t have a problem with the get routine being added, but if we do switch to a struct here, it could have a Once (or OnceCell) on whether we have installed it, so we don’t have to check all the tables every time. If we have a lot of runtime allocations post RTB, that would be wasting cycles. Probably fine, but still could be simpler. May be able to just check the post rtb flag itself.

patina_dxe_core/src/config_tables/memory_attributes_table.rs
joschock: I originally had it as `Once` until I tried to run tests. I'll try a few things and see what makes the most sense aesthetically.

patina_dxe_core/src/config_tables/memory_attributes_table.rs
joschock: I think in the normal case the number of config tables to be pretty small (less than a dozen). It's not a search of a large list; I expect that in practice the perf impact should be negligible (the cost of a half-dozen GUID read/compare operations, basically), and it should be confined to GCD map changes (i.e. page allocations) which are already fairly rare relative to pool operations.

 If you think the performance warrants it, though, I can check the POST_RTB flag.

patina_dxe_core/src/config_tables/memory_attributes_table.rs
joschock: Added a TestOnce instance.

patina_dxe_core/src/config_tables/memory_attributes_table.rs
os-d: I agree that it should be pretty small, just seems like we don’t need to do it every time since we should have the information already. I’ll leave it up to you.

joschock: Gated it on POST_RTB

makubacki (approved):

PR #1177, opened 2025-12-09
patina_dxe_core: image.rs: Improve usage of safe abstractions

Javagedes:
## Description

This pull request updates `image.rs` to use more rust abstractions over the more low-level / raw pointer usage we were using before. While this pull request will be squashed, each commit is cleanly broken out into it's own change. **It is suggested to review this PR commit by commit**.

**Commit 1:** This commit updates the hii_resource_section to be stored in an `Option<Box<[u8]>>` created via the `CoreMemoryManager` and the `MemoryManager` trait rather than a `Option<*mut [u8]>`. This allows us to remove multiple fields from `PrivateImageData` that existed purely to manually free the pages when dropped and simplify the management of this field.

**Commit 2:**: This commit updates the loaded image buffer to be stored in a `Box<[u8]>` created via the `CoreMemoryManager` and the `MemoryManager` trait for the same reasons as Commit 1.

**Commit 3:** This commit updates the `ImageStack` struct to be a `Box<[u8]>` created via the `CoreMemoryManager` and the `MemoryManager` trait for the same reasons as Commit 1.

**Commit 4:** This commit removes the unnecessary `image_info_ptr` field, which is just a pointer to the ` image_info: Box<efi::protocols::loaded_image::Protocol>,` field. We instead get the pointer directly from the Box.

**Commit 5:** This commit updates the `image_device_path_ptr` to store an `Option<Box<[u8]>>` rather than a `*mut c_void` and cleanly manages the lifetime of this data rather then leaking it. **NOTE: This commit has a bugfix to re-insert the private_image_data if we fail to uninstall either the `loaded_image` or `loaded_image_device_path` protocols for a reason other than NotFound**.

**Commit 6**: This commit simply pulls out the `exit_data` as a new struct so that we can mark `Send + Sync` on only that field. This is to eventually be able to remove the blanket `Send + Sync` implementation on the `DxeCoreGlobalImageData` struct.

**Commit 7:** Adds tests

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

All CI passes. Continue to boot to shell on Q35

## Integration Instructions

N/A

### DIFF
--- a/patina_dxe_core/src/image.rs
+++ b/patina_dxe_core/src/image.rs
@@ -61,6 +61,13 @@ pub const EFI_IMAGE_SUBSYSTEM_EFI_RUNTIME_DRIVER: u16 = 12;
 
 pub const ENTRY_POINT_STACK_SIZE: usize = 0x100000;
 
+// Compile time assert to make sure `STACK_ALIGNMENT` (which comes from uefi_corosensei) is never larger than
+// UEFI_PAGE_SIZE. This can cause issues with the stack allocation not being aligned properly. This was chosen rather
+// than updating the `AllocationOptions` alignment configuration being set to `STACK_ALIGNMENT` because we cannot
+// guarantee that the alignment will be a multiple of UEFI_PAGE_SIZE in all cases. We would rather hit a compile time
+// error then runtime error where no image is executed because we fail to allocate the stack.
+const _: () = assert!(STACK_ALIGNMENT < UEFI_PAGE_SIZE);
+
 // dummy function used to initialize PrivateImageData.entry_point.
 #[coverage(off)]
 extern "efiapi" fn unimplemented_entry_point(
@@ -264,7 +271,7 @@ impl PrivateImageData {
         let image_buffer = self.image_buffer.as_mut().ok_or(EfiError::LoadError)?;
         let physical_addr = self.image_info.image_base as usize;
 
-        // Update relocation data so it can be execution from it's current location in memory
+        // Update relocation data so if we need to relocate again later, we have the necessary info.
         self.relocation_data =
             pecoff::relocate_image(&self.pe_info, physical_addr, image_buffer, &self.relocation_data).map_err(
                 |err| {


cfernald:
@Javagedes, need to fixup the PR title.

patina_dxe_core/src/image.rs
```diff
@@ -266,21 +268,40 @@ impl PrivateImageData {
...
         }
+
+        Ok(())
+    }
+
+    fn relocate_image(&mut self) -> Result<(), EfiError> {
+        let image_buffer = self.image_buffer.as_mut().ok_or(EfiError::LoadError)?;
+        let physical_addr = self.image_info.image_base as usize;
+
+        // Update relocation data so it can be execution from it's current location in memory
```
joschock: nit: comment grammar could use some work.

patina_dxe_core/src/image.rs
```diff
@@ -73,75 +72,59 @@ extern "efiapi" fn unimplemented_entry_point(
...
-        let mut stack: efi::PhysicalAddress = 0;
         let len = align_up(size.max(MIN_STACK_SIZE), STACK_ALIGNMENT)?;
         // allocate an extra page for the stack guard page.
-        let allocated_pages = uefi_size_to_pages!(len) + 1;
+        let page_count = uefi_size_to_pages!(len) + 1;
 
-        // allocate the stack, newly allocated memory will have efi::MEMORY_XP already set, so we don't need to set it
-        // here
-        core_allocate_pages(efi::ALLOCATE_ANY_PAGES, efi::BOOT_SERVICES_DATA, allocated_pages, &mut stack, None)?;
+        let stack = CoreMemoryManager.allocate_pages(page_count, AllocationOptions::default())?.into_boxed_slice();
```
joschock, line 91: trivial and also previously existed, but: 'STACK_ALGINMENT` comes from corosensei and is usually way less than a page; but might be worth a static assert to ensure that it doesn't somehow become bigger than page alignment. I could imagine some future crazy architecture requiring the stack to be a 2MB boundary or something, in which case this code would silently violate that.

patina_dxe_core/src/image.rs
```diff
@@ -1226,23 +1224,36 @@ pub fn core_unload_image(image_handle: efi::Handle, force_unload: bool) -> Resul
...
+    if let Err(err) = core_uninstall_protocol_interface(
         image_handle,
         efi::protocols::loaded_image::PROTOCOL_GUID,
         private_image_data.image_info.as_ref() as *const efi::protocols::loaded_image::Protocol as *mut c_void,
-    );
+    ) && err != EfiError::NotFound
+    {
+        log::warn!("Failed to uninstall loaded image protocol for handle {image_handle:?}: {err:?}");
+        PRIVATE_IMAGE_DATA.lock().private_image_data.insert(image_handle, private_image_data);
+        return Ok(());
```
joschock: The spec is a bit ambiguous here, but are you sure `Ok` is the best return code?

patina_dxe_core/src/image.rs
Javagedes: Added this compile time assert:

```rust
// Compile time assert to make sure `STACK_ALIGNEMENT` (which comes from uefi_corosensei) is never larger than
// UEFI_PAGE_SIZE. This can cause issues with the stack allocation not being aligned properly. This was chosen rather
// than updating the `AllocationOptions` alignment configuration being set to `STACK_ALIGNMENT` because we cannot
// guarantee that the alignment will be a multiple of UEFI_PAGE_SIZE in all cases. We would rather hit a compile time
// error then runtime error where no image is executed because we fail to allocate the stack.
const _: () = assert!(STACK_ALIGNMENT < UEFI_PAGE_SIZE);
```

patina_dxe_core/src/image.rs
Javagedes: We were basically returning OK here before (by not checking it), and it appeared to work fine. I'm happy to change it if you prefer. That would make this a functional change however, which comes with it's own risk.

patina_dxe_core/src/image.rs
```diff
@@ -241,18 +223,38 @@ impl PrivateImageData {
...
-        &mut self,
-        alignment: usize,
-        memory_type: EfiMemoryType,
-        resource_section: &[u8],
-    ) -> Result<(), EfiError> {
+    fn load_resource_section(&mut self, image: &[u8]) -> Result<(), EfiError> {
+        let loaded_image = self.image_buffer.as_ref();
+
+        let result = pecoff::load_resource_section(&self.pe_info, image)
+            .inspect_err(|err| {
```
cfernald: Can't you do the logic in inspect_err in map_err to avoid having both?

patina_dxe_core/src/image.rs
```diff
@@ -73,75 +72,59 @@ extern "efiapi" fn unimplemented_entry_point(
...
-        let mut attributes = match dxe_services::core_get_memory_space_descriptor(stack) {
+        let mut attributes = match dxe_services::core_get_memory_space_descriptor(base_address) {
             Ok(descriptor) => descriptor.attributes,
             Err(_) => DEFAULT_CACHE_ATTR,
         };
 
         attributes = MemoryProtectionPolicy::apply_image_stack_guard_policy(attributes);
 
-        if let Err(err) = dxe_services::core_set_memory_space_attributes(stack, UEFI_PAGE_SIZE as u64, attributes) {
+        if let Err(err) =
```
cfernald, line 104: You can use `CoreMemoryManager.set_page_attributes()` for consistency

patina_dxe_core/src/image.rs
```diff
@@ -959,7 +960,7 @@ pub fn core_load_image(
     core_install_protocol_interface(
         Some(handle),
         efi::protocols::loaded_image_device_path::PROTOCOL_GUID,
-        loaded_image_device_path as *mut c_void,
+        private_info.image_device_path_ptr(),
```
cfernald: I've mentioned this before, but I get nervous about the practice of creating stored pointers from structures in boxes because we are circumventing the compilers understanding of lifetimes, etc. Do you think we should just leak this in this case? I can't tell if I'm being paranoid.

Javagedes: I really don't like using `Box::leak(...)` if I don't have to. It's basically a cop-out and would make us fail miri tests for leaking data. We are not a long-running program, so running out of memory by leaking memory is not really a concern, but I still frown upon it.

I bit my tongue in the fv code because the protocols we were leaking should never be uninstalled. However, in this scenario, these images can be uninstalled, and thus the data should also be deallocated.

I'm open to having my arm twisted into doing it if you and others have strong opinions, but I'd prefer not to :)

joschock: No strong opinions in this case, because I feel that I can reason about the memory ok and I don't expect outside code to be messing with the data in ways that break the compiler invariants (e.g. mutating it unexpectedly). 

To Chris' point, I do think that there is value in some cases to explicitly declare that you, the programmer, are now taking ownership of this memory because what you want to do (install a raw pointer in a protocol database for potentially promiscuous use outside the core) could break invariants the compiler expects. 

You could use `into_raw` instead of leaking which (in theory, for a subset of boxed types) allows you to `from_raw` to deallocate. I view the in_raw->from_raw flow to be essentially declaring that I'm removing the Box from the rust compilers purview to manage it myself - but that doesn't mean I'm leaking it forever.

Javagedes: I'm not sure I see the difference between manually doing the `into_raw` -> `from_raw` flow manually rather than letting the `Box` do the exact same thing, at least in this particular scenario. In this scenario, I would do `into_raw` when creating the object, then have to manually implement `Drop` on `PrivateImageData` where I then call `from_raw`. But that is basically just being my own box manually.

I could see it make sense if, during each individual protocol uninstall (for an image), I do:

1. Attempt to uninstall protocol
2. If success, call from_raw and deallocate
3. set the pointer for that protocol in the `PrivateImageData` to null
4. repeat for next protocol uninstall

But that seems like unnecessary complexity especially because now I have to go back to storing metadata (the PageFree struct) to call `from_raw`, when I can just attempt to uninstall all protocols and if I fail to uninstall even one, I just put the private image data back into our hashmap.

Javagedes: e.g. this is about to be the new flow for uninstall when I push my changes here in a second:

```rust
    let private_image_data = PRIVATE_IMAGE_DATA.lock().private_image_data.remove(&image_handle).unwrap();

    // If something fails to uninstall, then re-insert the private image data back into the map so the protocols
    // are not deallocated.
    if private_image_data.uninstall(image_handle).is_err() {
        PRIVATE_IMAGE_DATA.lock().private_image_data.insert(image_handle, private_image_data);
    }
```

joschock: Right; that's why in this case maybe `into_raw` isn't needed - we aren't doing things (that I can see) that would break compiler assumptions about the memory. But that's only because I don't expect external code to be scribbling into the box in a way that would break type invariants.

patina_dxe_core/src/image.rs
Javagedes: Updated all instances using both inspect_err and map_err to only use map_err. Good call.

patina_dxe_core/src/image.rs
Javagedes: as discussed, offline, this is a conversation that needs @os-d involved. Created #1185 to track the conversation.

patina_dxe_core/src/image.rs
Javagedes: Keeping OK as this is what the C core does (technically it does not even check)

patina_dxe_core/src/image.rs
Javagedes: resolving as a won't do. Please re-open if there is more conversation to be had or you really want me to do it :)

PR #1179, opened 2025-12-10
Implement `AcpiGet` (to enable custom ACPI SDT protocol split)

berlin-with0ut-return:
## Description

The ACPI SDT protocol as defined in the UEFI spec includes two main functionalities: 1) get and register_notify, which integrate directly with the AcpiTableProtocol (uninstall, install), 2) AML manipulation. Due to the complexity of reimplementing AML functionality in Rust and its unrelatedness to general ACPI functionality (install, uninstall, publish, get), an intermediate step for ACPI Rust integration is to only integrate functionality related to installation and retrieval of tables, while keeping AML functionality in C. Doing this requires splitting the ACPI SDT protocol into two parts: table functionality and AML functionality. 

- [X] Impacts functionality?
- [ ] Impacts security?
- [ ] **Breaking change?** Note: this doesn't directly break anything because the ACPI component is not yet enabled, but sets up the ACPI component to be enabled (additional changes needed in Patina and physical platform repos). Once the component is enabled in patina-dxe-core-qemu builds will break until all changes are integrated.
- [X] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- Boot Q35 to shell and Windows.
- TBD: test on physical Intel platform.

## Integration Instructions

- patina-dxe-core-qemu (or platform equivalent): turn on ACPI component
- patina-edk2: integrate new AcpiSdt split driver 
- patina-qemu (or platform equivalent): replace AcpiTableDxe with new AcpiSdt driver
- PatinaPkg: add modified AcpiSdt driver

### DIFF
--- a/components/patina_acpi/src/acpi_protocol.rs
+++ b/components/patina_acpi/src/acpi_protocol.rs
@@ -35,7 +35,7 @@ unsafe impl ProtocolInterface for AcpiTableProtocol {
         efi::Guid::from_fields(0xffe06bdd, 0x6107, 0x46a6, 0x7b, 0xb2, &[0x5a, 0x9c, 0x7e, 0xc5, 0x27, 0x5c]);
 }
 
-// C function interfaces for ACPI Table Protocol and ACPI SDT Protocol.
+// C function interfaces for ACPI Table Protocol and ACPI Get Protocol.
 type AcpiTableInstall = extern "efiapi" fn(*const AcpiTableProtocol, *const c_void, usize, *mut usize) -> efi::Status;
 type AcpiTableUninstall = extern "efiapi" fn(*const AcpiTableProtocol, usize) -> efi::Status;
 type AcpiTableGet = extern "efiapi" fn(usize, *mut *mut AcpiTableHeader, *mut u32, *mut usize) -> efi::Status;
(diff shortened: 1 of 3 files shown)


components/patina_acpi/src/acpi_protocol.rs
makubacki, line 1: Please add a brief rationale statement in the PR description - why and how it fits into a larger ACPI design plan.

components/patina_acpi/src/integration_test.rs
```diff
@@ -73,7 +73,7 @@ fn acpi_protocol_test(bs: StandardBootServices) -> patina::test::Result {
     let table_protocol =
         unsafe { bs.locate_protocol::<AcpiTableProtocol>(None) }.expect("Locate protocol should succeed.");
     // SAFETY: there is only one reference to the `AcpiSdtProtocol` during this test.
-    let sdt_protocol = unsafe { bs.locate_protocol::<AcpiSdtProtocol>(None) }.expect("Locate protocol should succeed.");
+    let sdt_protocol = unsafe { bs.locate_protocol::<AcpiGetProtocol>(None) }.expect("Locate protocol should succeed.");
```
makubacki: nit: You might want to rename `sdt_protocol` to `get_protocol` or something like that to more clearly distinguish the protocol being used here.

PR #1182, opened 2025-12-11
patina_smbios: update platform integration documentation

### LINKED ISSUE #1178: [Documentation]: add additional platform integration instructions to patina_smbios README
### Request Description

Add instructions for platforms to remove SmbiosDxe.inf from their platform, and that they can replace any platform specific smbios with your component

I will make the change

No maintainer feedback needed

### Anything else?

_No response_

kat-perez:
## Description

Updates the instructions in the patina_smbios crate README for platform integration. Closes #1178

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [x] Includes documentation?

## How This Was Tested

N/A

## Integration Instructions

N/A

### DIFF
--- a/components/patina_smbios/README.md
+++ b/components/patina_smbios/README.md
@@ -46,6 +46,95 @@ commands.add_component(SmbiosProvider::new(3, 9));  // SMBIOS 3.9
 > creation fails during `entry_point()`, the component returns
 > `EfiError::Unsupported` instead of panicking.
 
+## Platform Integration
+
+To integrate the `patina_smbios` component into your platform, remove the
+`SmbiosDxe` driver and add the `SmbiosProvider` component. This replaces
+`SmbiosDxe` as the producer of the `EFI_SMBIOS_PROTOCOL`.
+
+### Step 1: Remove the SmbiosDxe Driver
+
+Remove `SmbiosDxe.inf` from your platform's DSC file (e.g.,
+`YourPlatform.dsc` or `YourPlatformCommon.dsc.inc`):
+
+```diff
+-  #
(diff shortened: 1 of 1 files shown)


components/patina_smbios/README.md
```diff
@@ -46,6 +46,88 @@ commands.add_component(SmbiosProvider::new(3, 9));  // SMBIOS 3.9
...
 
+## Platform Integration
+
+To integrate the `patina_smbios` component into your platform, you must remove
+the existing SMBIOS drivers and replace them with the Patina SMBIOS component.
+
+### Step 1: Remove Existing SMBIOS Drivers
+
+Remove the following entries from your platform's DSC file (e.g.,
+`YourPlatform.dsc` or `YourPlatformCommon.dsc.inc`):
```
Javagedes, line 58: I'm hesitant to say that they **must** remove their platform's SmbiosPlatformDxe driver. Realistically they could keep their platforms Smbios platform driver, and everything would still work because your Patina component produces the protocol.

The above scenario will only be realistic when other components are developed that rely on the SmBios Patina service, which is not the case right now. But in that scenario, the platform would replace the EDKII `SmBiosDxe` driver and register the Patina component(s) without necessarily also replacing their platforms EDKII platform SmBios driver.

kat-perez: Yeah, I can clarify that without saying they "must" remove their platform driver. I moved the platform driver steps after the `SmbiosProvider` component add and labeled the platform SMBIOS component step optional

PR #1191, opened 2025-12-13
image: do not finish loading image if platform policy prevents it.

Javagedes:
## Description

In the EDKII implementation of the DXE Core, load_image will only continue loading an image when authentication fails if the error is a EFI_SECURITY_VIOLATION error. Any other error (and specifically EFI_ACCESS_DENIED) will immediately abort without actually loading the image.

The Patina DXE Core loads the image regardless of the result of the authentication, and only reports the result to the caller after the load has completed. This appears to be a security risk because even though the handle is returned as null, the protocols are still loaded and accessible. A call to `LocateHandleBuffer` could get you the handle and then a `StartImage` could start it.

This commit stops the loading of an image if authentication returns any error other then EFI_SECURITY_VIOLATION.

- [x] Impacts functionality?
- [x] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

CI, Boot to shell.

## Integration Instructions

N/A

### DIFF
--- a/patina_dxe_core/src/image.rs
+++ b/patina_dxe_core/src/image.rs
@@ -1138,6 +1138,7 @@ extern "efiapi" fn load_image(
         Err(ImageStatus::LoadError(err)) => return err.into(),
     };
 
+    // Safety: Caller must ensure that image_handle is a valid pointer. It is null-checked above.
     unsafe { image_handle.write_unaligned(handle) };
     status
 }


patina_dxe_core/src/image.rs
```diff
@@ -978,6 +992,18 @@ pub fn core_load_image(
     // authenticate the image
     let security_status = authenticate_image(file_path, &image_to_load, boot_policy, from_fv, authentication_status);
 
+    // If a security violation occurs, we still load the image, but will ultimately return a ImageStatus::SecurityViolation
+    if let Err(err) = security_status
+        && err != EfiError::SecurityViolation
+    {
+        // If the error is AccessDenied, we abort loading completely, as platform policy prohibits the image from being loaded
+        if err == EfiError::AccessDenied {
+            return Ok(ImageStatus::AccessDenied);
```
joschock: aesthetic nit: it's a bit weird to have `Ok(failure)` status. Consider `Result<efi::Handle, ImageStatus>`? where the `Ok` state corresponds is the current `Loaded`?

Ultimately, it isn't a functional issue, so okay if you feel like it is better this way and want to keep it.

Javagedes: My argument to stick with this implementation is that the error return case indicates an actual unexpected runtime error where as the OK case is the current status of the image, ejther loaded, loaded with a security violation, or completely blocked by platform policy. While EDKII treats them as EFI_ERRORs, I see them as valid, expected return cases.

joschock: I think about it as `Ok()` means the image loaded and I can do whatever I want with it. `Err()` means I have to take additional error handling actions on the caller side because the image is not ready for immediate use. I do see the distinction between "this can happen due to policy and is not a bug" vs "this should never happen and is probably a bug", but I don't think the former are `Ok()`. 

However, it is aesthetic, if you prefer the present implementation, I am okay with that.

Javagedes: Your reasoning makes sense, and we can still cleanly handle all the scenarios with the suggested interface by updating `ImageStatus` to

```rust
enum ImageStatus {
    LoadError(EfiError),
    SecurityViolation(efi::Handle),
    AccessDenied
};
```

I will update to that. Good feedback, thank you :)

makubacki (approved):

patina_dxe_core/src/image.rs
Javagedes: Updated!

PR #1192, opened 2025-12-15
patina_dxe_core: Additional cleanup of image.rs

Javagedes:
## Description

This commit performs some additional non-functional cleanup of image.rs to break up the larger functions into smaller chunks that only perform one operation each. This makes understanding and testing the code simpler.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

CI, Boots to shell.

## Integration Instructions

N/A

### DIFF
--- a/patina_dxe_core/src/image.rs
+++ b/patina_dxe_core/src/image.rs
@@ -3279,7 +3279,7 @@ mod tests {
                     reserved: [0; 4],
                 },
                 module_name: guids::DXE_CORE,
-                entry_point: entry_point as usize as u64,
+                entry_point: entry_point as *const () as u64,
             };
             let end_hob = patina::pi::hob::header::Hob {
                 r#type: patina::pi::hob::END_OF_HOB_LIST,


Javagedes:
I will work on unit tests today. But please feel free to review. It is a non-functional change, just doing some re-organization.

os-d (approved):

patina_dxe_core/src/image.rs
```diff
@@ -377,10 +375,53 @@ impl PrivateImageData {
         result
     }
 
-    fn image_device_path_ptr(&self) -> *mut c_void {
+    /// Calculates and sets the file path for this image, if appropriate
+    ///
+    /// If the device_handle is valid, determines the file name from that handles device path.
```
joschock:
```suggestion
    /// If the device_handle is valid, determines the file name from the device path instance installed on the device_handle.
```

patina_dxe_core/src/image.rs
```diff
@@ -377,10 +375,53 @@ impl PrivateImageData {
...
+        {
+            // strip the parent device path prefix from the full device path to leave only the file node
+            let (_, device_path_size) =
+                device_path_node_count(device_path as *mut efi::protocols::device_path::Protocol).unwrap_or((0, 0));
+            let split_idx =
+                device_path_size.saturating_sub(core::mem::size_of::<efi::protocols::device_path::Protocol>());
+            let file_path = unsafe { (file_path as *const u8).add(split_idx) };
+            Some(file_path as *mut efi::protocols::device_path::Protocol)
+        } else {
+            None
```
joschock: Is providing both a valid device_handle and a non-null file_path an error case? We return None here, but this seems more like a "You called me with unexpected args" case.

patina_dxe_core/src/image.rs
```diff
@@ -377,10 +375,53 @@ impl PrivateImageData {
...
+    fn set_file_name(&mut self, file_path: *mut efi::protocols::device_path::Protocol) -> Result<(), EfiError> {
+        let fixed_file_path = if self.image_info.device_handle == protocol_db::INVALID_HANDLE {
+            Some(file_path)
+        } else if file_path.is_null()
+            && let Ok(device_path) = PROTOCOL_DB
+                .get_interface_for_handle(self.image_info.device_handle, efi::protocols::device_path::PROTOCOL_GUID)
+        {
+            // strip the parent device path prefix from the full device path to leave only the file node
+            let (_, device_path_size) =
+                device_path_node_count(device_path as *mut efi::protocols::device_path::Protocol).unwrap_or((0, 0));
```
joschock: Consider returning an error if malformed device path count returns an error rather than masking it with `unrwap_or`

patina_dxe_core/src/image.rs
```diff
@@ -482,12 +522,132 @@ impl DxeCoreGlobalImageData {
...
+        assert_eq!(handle, protocol_db::DXE_CORE_HANDLE);
+
+        initialize_debug_image_info_table(system_table);
+        core_new_debug_image_info_entry(
+            EfiDebugImageInfoNormal::EFI_DEBUG_IMAGE_INFO_TYPE_NORMAL,
+            private_image_data.image_info.as_ref() as *const efi::protocols::loaded_image::Protocol,
+            handle,
+        );
+
+        self.private_image_data.insert(handle, private_image_data);
```
joschock: As a defensive measure, consider moving this up to right after core_install_protocol_interface. Once that succeeds, external agents can find a pointer to private_image_data, and so it needs to remain valid. This function is presently infallible but if that were to change in the future and an early exit were to occur between the install_protocol and the insertion of the private_image_data into the long-lived map, you could end up with a dangling interface. Putting them closer together is an aid to the programmer that they might be related :).

patina_dxe_core/src/image.rs
```diff
@@ -482,12 +522,132 @@ impl DxeCoreGlobalImageData {
...
+        PROTOCOL_DB.validate_handle(parent).inspect_err(|err| log::error!("Invalid parent handle {err:?}"))?;
+
+        PROTOCOL_DB.get_interface_for_handle(parent, efi::protocols::loaded_image::PROTOCOL_GUID).map_err(|err| {
+            log::error!("Failed to get loaded image interface on the parent handle: {err:?}");
+            EfiError::InvalidParameter
+        })?;
+        Ok(())
+    }
+
+    /// Returns the image metadata by it's buffer using device path protocol.
```
joschock:
```suggestion
    /// Returns a tuple of image meta-data: `(image_as_vec, from_fv, device_handle, authentication_status)`
```

patina_dxe_core/src/image.rs
```diff
@@ -482,12 +522,132 @@ impl DxeCoreGlobalImageData {
...
+        file_path: *mut efi::protocols::device_path::Protocol,
+    ) -> (Vec<u8>, bool, *mut c_void, u32) {
+        if let Ok((_, device_handle)) = core_locate_device_path(efi::protocols::device_path::PROTOCOL_GUID, file_path) {
+            (image.to_vec(), false, device_handle, 0)
+        } else {
+            (image.to_vec(), false, protocol_db::INVALID_HANDLE, 0)
+        }
+    }
+
+    /// Returns the image metadata by it's file path using simple file system or load file protocols.
```
joschock:
```suggestion
    /// Returns the image metadata by its file path using simple file system or load file protocols.
```

patina_dxe_core/src/image.rs
```diff
@@ -482,12 +522,132 @@ impl DxeCoreGlobalImageData {
...
+    ///
+    /// Note: presently none of the supported methods return `from_fv` or `authentication_status`.
```
joschock: Not sure if the `Note:` is still accurate. At least one of the branches is returning `true` for from_fv

patina_dxe_core/src/image.rs
Javagedes: makes sense. I made the change here for the patina dxe core image, but also in `core_load_image` for actual images.

patina_dxe_core/src/image.rs
Javagedes: Updated to bubble up the error instead of unwrapping to 0,0 and keeping the entire device path.

patina_dxe_core/src/image.rs
Javagedes: There are two things going on here lol. The first is that my code here had a bug in it (I'm checking that it is null, but in reality, this function should only be called only if the protocol is not null. The second was an existing issue, which is that we return None if we cannot get the interface for the handle. 

This is what the C code does:

1. If `file_path` is not null set the filepath var and continue to (2). Otherwise completely exit.
2. If we have a have a valid device handle, continue to (3). Otherwise skip to (4)
3. If the device handle has a device_path protocol, strip remove that device path from our device path and set the filepath var
4. Set the file path protocol to the filepath var.

I will update this whole bit of logic.

patina_dxe_core/src/image.rs
```diff
@@ -375,43 +382,38 @@ impl PrivateImageData {
...
-    fn set_file_name(&mut self, file_path: *mut efi::protocols::device_path::Protocol) -> Result<(), EfiError> {
-        let fixed_file_path = if self.image_info.device_handle == protocol_db::INVALID_HANDLE {
-            Some(file_path)
-        } else if file_path.is_null()
+    /// The file path is the full device path for this image and is what is set for the loaded_image_device_path protocol.
+    fn set_file_path(&mut self, file_path: NonNull<efi::protocols::device_path::Protocol>) -> Result<(), EfiError> {
+        let mut fp = file_path.as_ptr();
+
+        // If the device handle is valid, and the handle has a device path protocol, we need to adjust our file path
+        if self.image_info.device_handle != protocol_db::INVALID_HANDLE
```
joschock: `get_interface_for_handle` will check for device_handle validity, so you could eliminate the conditional here.

patina_dxe_core/src/image.rs
```diff
@@ -375,43 +382,38 @@ impl PrivateImageData {
...
             && let Ok(device_path) = PROTOCOL_DB
                 .get_interface_for_handle(self.image_info.device_handle, efi::protocols::device_path::PROTOCOL_GUID)
         {
-            // strip the parent device path prefix from the full device path to leave only the file node
             let (_, device_path_size) =
                 device_path_node_count(device_path as *mut efi::protocols::device_path::Protocol)?;
             let split_idx =
```
joschock, line 402: nit: might be worth a clarifying comment that the subtract here is removing the device path end node.

patina_dxe_core/src/image.rs
joschock: updated logic looks good to me.

patina_dxe_core/src/image.rs
Javagedes: Hi John. I just wrote a test regarding this, and I do not see the END node being subtracted. This is the test:

1. Device handle with path `PCI(0,1C)/PCI(0,0)/END`
2. child full path: `PCI(0,1C)/PCI(0,0)/EFI/BOOT/BOOTX64.EFI/END`
3. Final device path: `EFI/BOOT/BOOTX64.EFI/END` < device path end node still here

Is this expected? Do I need to make some logical changes?

I'll be pushing the test in just a second, it is called `test_set_file_path_with_a_device_handle`

joschock: The test is correct and the result is correct - the subtract is just removing the `END` from `PCI(0,1C)/PCI(0,0)/END` before calculating the prefix index into `PCI(0,1C)/PCI(0,0)/EFI/BOOT/BOOTX64.EFI/END`. Without the subtract, it would remove too much of the prefix (it would remove the node header for the `EFI/BOOT/BOOTX64.EFI` node).

The comment is to clarify this as a cognitive aid to future developers. Logic is sound.

Javagedes: OK this all makes sense. Thanks. I've updated the comment.

patina_dxe_core/src/image.rs
Javagedes: Makes sense. Updated!

vineelko (approved):

PR #1198, opened 2025-12-15
Replaced AtomicPtr with Once in boot_services and runtime_services

### LINKED ISSUE #1197: [Feature]: Evaluate removal of AtomicPtr in boot_services and runtime_services
### Feature Overview

See RFC 0021; this task is to evaluate the above two modules in SDK for removal of AtomicPtr usage. 

### Solution Overview

Replace with spin::Once

### Alternatives Considered

_No response_

### Urgency

Low

I will implement the feature

No maintainer feedback needed

### Anything else?

_No response_

joschock:
## Description

Replaces the AtomicPtr usage with Once for semantics that better align with intended usages. Also makes the `static` lifetime requirement on BootServices (which was effectively there when it was being used as a raw pointer) more explicit. 

Closes #1197

- [x] Impacts functionality?
- [ ] Impacts security?
- [x] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Passes unit testing, confirmed boot to EFI shell in QEMU Q35.

## Integration Instructions

Construction of `StandardBootServices` and `StandardRuntimeServices` will need to supply a static-lifetime reference to the appropriate structure. This should already be the case since most instantiations will be from the global boot services or runtime services pointers from the system table that already have `static` semantics (i.e. they live for the lifetime of UEFI). For test usages, leaking a Box should suffice.

### DIFF
--- a/patina_dxe_core/src/systemtables.rs
+++ b/patina_dxe_core/src/systemtables.rs
@@ -636,20 +636,36 @@ impl EfiSystemTable {
 
     #[allow(dead_code)]
     pub fn boot_services(&self) -> &'static efi::BootServices {
+        // SAFETY: boot_services pointer in the system table is valid once system table is constructed.
+        // Note: some function pointers within the boot_services table may point to stub functions until
+        // architectural drivers outside the core install the corresponding architectural interfaces; but
+        // the pointers will be valid pointers to stub functions.
         unsafe { self.system_table.boot_services.as_ref().expect("BootServices uninitialized") }
     }
 
     #[allow(dead_code)]
     pub fn boot_services_mut(&mut self) -> &mut efi::BootServices {
+        // SAFETY: boot_services pointer in the system table is valid once system table is constructed.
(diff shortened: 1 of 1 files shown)


patina_dxe_core/src/systemtables.rs
```diff
@@ -635,7 +635,7 @@ impl EfiSystemTable {
     }
 
     #[allow(dead_code)]
-    pub fn boot_services(&self) -> &efi::BootServices {
+    pub fn boot_services(&self) -> &'static efi::BootServices {
         unsafe { self.system_table.boot_services.as_ref().expect("BootServices uninitialized") }
```
Javagedes, line 643: nit: could you add safety comments here just because you happen to be touching the code? Just for this and the runtime services function

Javagedes (approved):

patina_dxe_core/src/systemtables.rs
joschock: Added, also filed an issue to make some of these functions `unsafe` (#1200 )

PR #1201, opened 2025-12-16
[feature/patina-boot] patina_boot: Add patina_boot crate

kat-perez:
## Description

Adds BootOrchestration component, simple console discovery, simple BootOption Config

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [x] Includes documentation?

## How This Was Tested

QEMU Platform Integration:
- Q35
- SBSA

## Integration Instructions

N/A

### DIFF
--- a/components/patina_boot/Cargo.toml
+++ b/components/patina_boot/Cargo.toml
@@ -1,0 +1,25 @@
+[package]
+name = "patina_boot"
+resolver = "2"
+version.workspace = true
+repository.workspace = true
+license.workspace = true
+edition.workspace = true
+description = "Boot orchestration components for Patina firmware."
+
+[lints]
+workspace = true
+
+[dependencies]
+log = { workspace = true }
+patina = { workspace = true, features = ["unstable-device-path"] }
+patina_macro = { workspace = true }
+r-efi = { workspace = true }
+
+[dev-dependencies]
+patina = { path = "../../sdk/patina", features = ["mockall", "unstable-device-path"] }
+mockall = { workspace = true }
+
+[features]
+default = []
+std = []
--- a/components/patina_boot/src/component/boot_orchestrator.rs
+++ b/components/patina_boot/src/component/boot_orchestrator.rs
@@ -1,0 +1,74 @@
+//! Boot orchestrator component.
+//!
+//! Orchestrates the boot flow with device enumeration, event signaling, and boot execution.
+//!
+//! ## Rationale
+//!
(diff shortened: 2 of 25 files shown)


components/patina_boot/src/library.rs
```diff
@@ -0,0 +1,311 @@
+//! Library functions for boot orchestration.
```
Javagedes: nit: Not a big fan of calling this module "library" considering it is a sub-module of "lib.rs". I don't really have a suggestion for a better module name but would ask that it is updated.

components/patina_boot/Cargo.toml
```diff
@@ -0,0 +1,25 @@
+[package]
+name = "patina_boot"
+resolver = "2"
+version.workspace = true
+repository.workspace = true
+license.workspace = true
+edition.workspace = true
+description = "Boot orchestration components for Patina firmware."
```
Javagedes, line 8: Please add a readme file, even if it is basically empty :)

components/patina_boot/src/component/boot_orchestrator.rs
```diff
@@ -0,0 +1,86 @@
...
+    ///
+    /// # Flow
+    ///
+    /// 1. Perform connect-dispatch interleaving for device enumeration
+    /// 2. Signal EndOfDxe (security components perform lockdown)
+    /// 3. Discover consoles
+    /// 4. Signal ReadyToBoot
+    /// 5. Execute boot options from config
+    /// 6. If all boot options fail, call failure handler
+    fn entry_point(
```
Javagedes, line 50: From the looks of this, this will be triggered immediately, well before you are ready, as there is no function parameter that is not immediately available. You will need something to signal that this component should be executed.

Additionally, How does this interact with `call_bds` ? and the BDS protocol? Might be a conversation for the RFC.

kat-perez: Added this topic to "Unresolved Questions" section, to open discussion: https://github.com/kat-perez/patina/blob/rfc-boot-manager/docs/src/rfc/text/0000-patina-boot.md#unresolved-questions

components/patina_boot/src/library.rs
kat-perez: thoughts on "helpers"?

kat-perez:
Remaking PR with different branch name

PR #1204, opened 2025-12-18
Improve memory map stability [Rebase & FF]

makubacki:
## Description

A series of commits to improve memory map stability and allocation efficiency for loader memory.

---

**allocator: Remove invalid unit test check**

A non-null memory map pointer is not necessary for getting the
memory map size with a non-null size pointer.

---

**Eliminate vector allocation in get_memory_map()**

Moves GCD related functionality for generating the EFI memory map
into the GCD. Iterates over GCD descriptors in place to avoid
allocating temporary vectors on every entry.

Also removes temporary vector allocation when get_memory_map() is
called to get the memory map size making that flow more efficient.

---

**allocator: Use high-traffic expansion for loader code**

As an efficiency improvement, use high-traffic allocation
expansion for loader code allocations, which can be
relatively large.

---

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- `cargo make all` (unit and integration tests)
- QEMU boot to Windows
- Intel physical platform boot to Windows

## Integration Instructions

- N/A

### DIFF
--- a/patina_dxe_core/src/allocator.rs
+++ b/patina_dxe_core/src/allocator.rs
@@ -200,6 +200,17 @@ pub static EFI_LOADER_CODE_ALLOCATOR: UefiAllocatorWithFsb = UefiAllocator::new(
     efi::LOADER_CODE,
 );
 
+pub static EFI_LOADER_DATA_ALLOCATOR: UefiAllocatorWithFsb = UefiAllocator::new(
+    SpinLockedFixedSizeBlockAllocator::new(
+        &GCD,
+        protocol_db::EFI_LOADER_DATA_ALLOCATOR_HANDLE,
+        NonNull::from_ref(GCD.memory_type_info(efi::LOADER_DATA)),
+        DEFAULT_PAGE_ALLOCATION_GRANULARITY,
+        HIGH_TRAFFIC_ALLOC_MIN_EXPANSION,
+    ),
+    efi::LOADER_DATA,
+);
+
 pub static EFI_BOOT_SERVICES_CODE_ALLOCATOR: UefiAllocatorWithFsb = UefiAllocator::new(
     SpinLockedFixedSizeBlockAllocator::new(
         &GCD,
@@ -237,9 +248,10 @@ pub static EFI_RUNTIME_SERVICES_DATA_ALLOCATOR: UefiAllocatorWithFsb = UefiAlloc
     efi::RUNTIME_SERVICES_DATA,
 );
 
-pub static STATIC_ALLOCATORS: [(&UefiAllocatorWithFsb, efi::MemoryType); 5] = [
+pub static STATIC_ALLOCATORS: [(&UefiAllocatorWithFsb, efi::MemoryType); 6] = [
     (&EFI_BOOT_SERVICES_DATA_ALLOCATOR, efi::BOOT_SERVICES_DATA),
     (&EFI_LOADER_CODE_ALLOCATOR, efi::LOADER_CODE),
+    (&EFI_LOADER_DATA_ALLOCATOR, efi::LOADER_DATA),
     (&EFI_BOOT_SERVICES_CODE_ALLOCATOR, efi::BOOT_SERVICES_CODE),
     (&EFI_RUNTIME_SERVICES_CODE_ALLOCATOR, efi::RUNTIME_SERVICES_CODE),
     (&EFI_RUNTIME_SERVICES_DATA_ALLOCATOR, efi::RUNTIME_SERVICES_DATA),
@@ -432,8 +444,9 @@ impl AllocatorMap {
         // Check static allocators first, then dynamic allocators
         match memory_type {
             efi::BOOT_SERVICES_DATA => Ok(&EFI_BOOT_SERVICES_DATA_ALLOCATOR),
-            efi::LOADER_CODE | efi::BOOT_SERVICES_CODE => Ok(match memory_type {
+            efi::LOADER_CODE | efi::LOADER_DATA | efi::BOOT_SERVICES_CODE => Ok(match memory_type {
                 efi::LOADER_CODE => &EFI_LOADER_CODE_ALLOCATOR,
+                efi::LOADER_DATA => &EFI_LOADER_DATA_ALLOCATOR,
                 efi::BOOT_SERVICES_CODE => &EFI_BOOT_SERVICES_CODE_ALLOCATOR,
                 _ => unreachable!(),
             }),
@@ -496,6 +509,7 @@ impl AllocatorMap {
         match memory_type {
             efi::BOOT_SERVICES_DATA => return Some(&EFI_BOOT_SERVICES_DATA_ALLOCATOR),
             efi::LOADER_CODE => return Some(&EFI_LOADER_CODE_ALLOCATOR),
+            efi::LOADER_DATA => return Some(&EFI_LOADER_DATA_ALLOCATOR),
             efi::BOOT_SERVICES_CODE => return Some(&EFI_BOOT_SERVICES_CODE_ALLOCATOR),
             efi::RUNTIME_SERVICES_CODE => return Some(&EFI_RUNTIME_SERVICES_CODE_ALLOCATOR),
             efi::RUNTIME_SERVICES_DATA => return Some(&EFI_RUNTIME_SERVICES_DATA_ALLOCATOR),
@@ -1281,7 +1295,7 @@ mod tests {
                 },
                 &[
                     // for test, pick dynamic allocators, since state is easier to clean up for those.
-                    0x02, 0x00, 0x00, 0x00, 0x00, 0x01, 0x00, 0x00, //0x0100 pages of LOADER_DATA
+                    0x0d, 0x00, 0x00, 0x00, 0x00, 0x01, 0x00, 0x00, //0x0100 pages of PAL_CODE
                     0x09, 0x00, 0x00, 0x00, 0x00, 0x02, 0x00, 0x00, //0x0200 pages of ACPI_RECLAIM_MEMORY
                     0x0a, 0x00, 0x00, 0x00, 0x00, 0x03, 0x00, 0x00, //0x0300 pages of ACPI_MEMORY_NVS
                 ],
@@ -1308,8 +1322,8 @@ mod tests {
 
             init_memory_support(&hob_list);
 
-            let loader_range = ALLOCATORS.lock().get_allocator(efi::LOADER_DATA).unwrap().reserved_range().unwrap();
-            assert_eq!(loader_range.end - loader_range.start, 0x100 * 0x1000);
+            let pal_code_range = ALLOCATORS.lock().get_allocator(efi::PAL_CODE).unwrap().reserved_range().unwrap();
+            assert_eq!(pal_code_range.end - pal_code_range.start, 0x100 * 0x1000);
 
             let reclaim_range =
                 ALLOCATORS.lock().get_allocator(efi::ACPI_RECLAIM_MEMORY).unwrap().reserved_range().unwrap();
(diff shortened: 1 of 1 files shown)


patina_dxe_core/src/config_tables/memory_attributes_table.rs
```diff
@@ -142,22 +143,32 @@ pub fn core_install_memory_attributes_table() {
...
-        Ok(descriptors) => descriptors,
+    // Reserve a buffer for the memory map descriptors
+    let descriptor_count = GCD.memory_descriptor_count();
+    let mut desc_buffer = Vec::with_capacity(descriptor_count);
+    desc_buffer.resize(
+        descriptor_count,
+        efi::MemoryDescriptor { r#type: 0, physical_start: 0, virtual_start: 0, number_of_pages: 0, attribute: 0 },
+    );
+
+    let actual_count = match GCD.populate_efi_memory_map(&mut desc_buffer, true) {
```
Javagedes: nit: i would use a let Ok statement here since you dont' care about the actual error

```rust
let Ok(actual_count) = GCD.populate_efi_memory_map(&mut desc_buffer, true) else {
  log::error!("Failed to populate memory map descriptors.");
  return;
}
```

patina_dxe_core/src/gcd/spin_locked_gcd.rs
```diff
@@ -1146,6 +1152,213 @@ impl GCD {
...
+
+        for read_idx in 0..descriptors.len() {
+            let current = descriptors[read_idx];
+
+            // Try to merge with the previous descriptor
+            if write_idx > 0 {
+                let prev = &mut descriptors[write_idx - 1];
+                if prev.r#type == current.r#type
+                    && prev.attribute == current.attribute
+                    && prev.physical_start + prev.number_of_pages * UEFI_PAGE_SIZE as u64 == current.physical_start
```
os-d: nit: uefi_pages_to_size!

patina_dxe_core/src/gcd/spin_locked_gcd.rs
```diff
@@ -1146,6 +1152,213 @@ impl GCD {
...
+    }
+
+    /// Determines if a GCD memory descriptor should be included in the EFI memory map.
+    ///
+    /// # Arguments
+    /// * `descriptor` - The GCD memory space descriptor
+    ///
+    /// Returns
+    /// * `Some(memory_type)` if the descriptor should be included, `None` otherwise.
+    fn is_efi_memory_map_descriptor(descriptor: &MemorySpaceDescriptor) -> Option<efi::MemoryType> {
```
Javagedes: nit: I understand what this is doing but I don't think the function name and description is quite accurate. Its doing more then just returning a boolean value if should be included in the memory map or not.

Also, to me, this makes more sense as an associated function of the `MemorySpaceDescriptor` struct rater then here.

patina_dxe_core/src/gcd/spin_locked_gcd.rs
```diff
@@ -1146,6 +1152,213 @@ impl GCD {
...
+    ///
+    /// ## Arguments
+    ///
+    /// * `descriptor` - The GCD memory space descriptor
+    /// * `memory_type` - The EFI memory type for this descriptor
+    /// * `active_attributes` - If true, use active attributes; if false, use capabilities
+    ///
+    /// Returns
+    /// * `u64` - The adjusted attributes for the EFI memory descriptor.
+    fn adjust_efi_memory_map_descriptor(
```
Javagedes, line 1206: @os-d will probably want this consolidated to his `gcd::MemoryProtectionPolicy` struct

patina_dxe_core/src/gcd/spin_locked_gcd.rs
```diff
@@ -1146,6 +1152,213 @@ impl GCD {
...
+
+        let mut write_idx = 0;
+
+        for read_idx in 0..descriptors.len() {
+            let current = descriptors[read_idx];
+
+            // Try to merge with the previous descriptor
+            if write_idx > 0 {
+                let prev = &mut descriptors[write_idx - 1];
+                if prev.r#type == current.r#type
```
os-d, line 1175: Check capabilities here as well? Capabilities are not really that useful, but theoretically could be used

patina_dxe_core/src/gcd/spin_locked_gcd.rs
makubacki: True. I originally did just return a boolean but had it return an `Option` for all the cases its used in. I'll move it to `MemorySpaceDescriptor`.

Javagedes (approved):

patina_dxe_core/src/gcd/spin_locked_gcd.rs
os-d: Yeah, I remember debating it at the time and I don't remember why I didn't include the runtime setting on runtime code and data to be in the policy. I think having this adjust function be separate from the policy is totally fine, there are two different use cases here and it is really just whether to apply the policy or not.

But at this point, I would say the runtime bit on runtime services code and data should just move into apply_efi_memory_map_policy(), then this function (if it stays a function) can just be if active_attributes, just return attributes, otherwise apply policy.

patina_dxe_core/src/gcd/spin_locked_gcd.rs
```diff
@@ -1146,6 +1152,213 @@ impl GCD {
...
     }
 
+    /// Merges adjacent EFI memory descriptors in place.
+    ///
+    /// # Arguments
+    /// * `descriptors` - A mutable slice of EFI memory descriptors to be merged.
+    ///
+    /// Returns
+    /// * `usize` - The new count of descriptors after merging.
+    fn merge_blocks_in_place(descriptors: &mut [efi::MemoryDescriptor]) -> usize {
```
os-d, line 1162: It would be nice to have some unit tests for these pure functions.

patina_dxe_core/src/gcd/spin_locked_gcd.rs
makubacki: This taking the logic from here and operating on memory attributes instead of GCD attributes:

https://github.com/OpenDevicePartnership/patina/blob/d9108dfb947bc8b5bd41d12629e903bd718081bb/patina_dxe_core/src/allocator.rs#L779-L782

Let me know if I misunderstood the request.

patina_dxe_core/src/gcd/spin_locked_gcd.rs
makubacki: Will do.

patina_dxe_core/src/gcd/spin_locked_gcd.rs
makubacki: Done

patina_dxe_core/src/allocator.rs
```diff
@@ -195,7 +195,7 @@ pub static EFI_LOADER_CODE_ALLOCATOR: UefiAllocatorWithFsb = UefiAllocator::new(
         protocol_db::EFI_LOADER_CODE_ALLOCATOR_HANDLE,
```
joschock, line 195: Do we need this on the EFI_LOADER_DATA_ALLOCATOR too?

patina_dxe_core/src/allocator.rs
```diff
@@ -913,18 +801,9 @@ extern "efiapi" fn get_memory_map(
...
-    assert_ne!(efi_descriptors.len(), 0);
-
-    let required_map_size = efi_descriptors.len() * mem::size_of::<efi::MemoryDescriptor>();
-
-    // Safety: caller must ensure that memory_map_size is a valid pointer. It is null-checked above.
+    let required_map_size = GCD.memory_descriptor_count_for_efi_memory_map() * mem::size_of::<efi::MemoryDescriptor>();
+    assert_ne!(required_map_size, 0);
     unsafe { memory_map_size.write_unaligned(required_map_size) };
-
     if map_size < required_map_size {
```
joschock, line 821: I think there is a subtle semantic difference between the EDK2 core and this impl; on the EDK2 core, if I call get_memory_map with memory_map_size = 0 to query the required size, I'll get a number roughly close to the size that would be returned from a call with a large enough buffer - i.e. memory_map_size will be roughly num_merged_decscriptors * descriptor size. However, on Patina, a query with memory_map_size = 0 will return num_unmerged_descriptors * descriptor size, which would likely be a much larger number.

Put another way - I think a caller of gBS->GetMemorymap() on the EDK2 core only needs to allocate roughly the number of _merged and filtered_ descriptors for the receiving buffer, whereas on Patina it needs to allocate roughly the number of _total unmerged and unfiltered_ descriptors for the receiving buffer.  

On patina the GetMemoryMap will return this larger size if BUFFER_TOO_SMALL is returned, so I can't think of how this would cause an actual problem, but I'm not sure exactly how the boot loader might use it and I wanted to note it as a potential delta vs. core behavior.

patina_dxe_core/src/gcd/spin_locked_gcd.rs
makubacki: I tried this but it got more complicated because `prev.number_of_pages` is `u64` and everything except `UEFI_PAGE_SIZE` is here. So, it stayed cleaner just leaving it with only `UEFI_PAGE_SIZE` being cast to `u64`.

patina_dxe_core/src/allocator.rs
makubacki: I'll update it too.

patina_dxe_core/src/allocator.rs
makubacki: This is something I had in mind, but I didn't think it would be a practical issue. Let me know if you think it's a problem. Otherwise, I can mention it in the patina docs.

patina_dxe_core/src/gcd/spin_locked_gcd.rs
os-d: Oh, sorry I was thinking of gcd descriptors, not memory map descriptors

patina_dxe_core/src/allocator.rs
joschock: From my side I think it should be okay? Maybe @os-d or @cfernald could comment based on their knowledge of how boot applications use this API?

os-d: Windows will allocate 2 x <return of GetMemoryMap(size = 0). Linux will allocate 8 + <return of GetMemoryMap(size = 0). Which causes EBS failures on Linux semi-frequently because we can easily get more than 8 descriptors created when doing EBS stuff.

So, I actually think this is good to lie with a bigger number because I think we'll prevent EBS issues (though with us not letting memory map changes during EBS, that *should* not be a problem). As long as this number isn't massive (and if it is we would have other problems), then I think this is fine and maybe even preferable to hedge our bets for Linux.

The only potential concern I have, that @cfernald would need to weigh in on (or a simple Windows boot test could confirm, since I believe this was 100% reproable) is that at one point recently, bootmgr was choking on patina having too many memory descriptors, it had some expectation in an array length that we were pushing past and it would die. We fixed that on the patina side by merging descriptors better. I don't recall if a bootmgr fix was made.

However, I believe this is the case where the actual returned descriptor count was very large, not in the let-me-see-what-size-buffer case, so I think that wouldn't be an issue here, but we should confirm.

makubacki: @os-d, I've booted this change to Windows on a few platforms. Since @cfernald might not be available to respond, I recall that and I believe I still have that image around. If so, I'll double check, otherwise I'll note this behavior in a doc update and merge.

makubacki: Doc update pushed. The Windows image I have from around that time completed EBS successfully:

```
Exit from WriteLogFiles.
MpInitChangeApLoopCallback() done!
INFO - Patina on-system unit-test results:
  patina_adv_logger::integration_test::adv_logger_test ... ok (1 passes)
  patina_dxe_core::core_patina_tests::stability_tests::page_table_tests_1gb_split ... ok (187 passes)
  patina_dxe_core::core_patina_tests::stability_tests::page_table_tests_1gb_unmap_2mb_remap ... ok (187 passes)
  patina_dxe_core::core_patina_tests::stability_tests::page_table_tests_1gb_unmap_4kb_remap ... ok (187 passes)
  patina_dxe_core::core_patina_tests::stability_tests::page_table_tests_2mb_split ... ok (187 passes)
  patina_dxe_core::core_patina_tests::stability_tests::page_table_tests_2mb_unmap ... ok (187 passes)
  patina_dxe_core::memory_manager::memory_manager_allocations_test ... ok (1 passes)
  patina_dxe_core::memory_manager::memory_manager_attributes_test ... ok (1 passes)
  qemu_resources::q35::component::service::smbios_test::q35_smbios_ffi_test ... ok (1 passes)

INFO - EBS completed successfully.
```

PR #1212, opened 2026-01-05
patina_dxe_core: move `image.rs` to `pi_dispatcher`

Javagedes:
## Description

This commit moves `image.rs` as a sub-module underneath `pi_dispatcher` and removes the global static variable in favor of a field of the `PiDispatcher` struct of the core.

This commit also updates all tests to not rely on the global static, and instead a non-global static core.

**NOTE: This PR changes no logic. All logical changes were handled in #1192. This PR only moves the code and replaces `STATIC` usage with `self` usage.**

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

CI, continues to boot to shell.

## Integration Instructions

N/A

### DIFF
--- a/patina_dxe_core/src/lib.rs
+++ b/patina_dxe_core/src/lib.rs
@@ -225,7 +225,7 @@ pub trait MemoryInfo {
     type MemoryInfo = MockMemoryInfo;
     type CpuInfo = MockCpuInfo;
 ))]
-pub trait PlatformInfo {
+pub trait PlatformInfo: 'static {
     /// The platform's memory information and configuration.
     type MemoryInfo: MemoryInfo;
 
@@ -482,7 +482,7 @@ impl<P: PlatformInfo> Core<P> {
     ///
     /// 1. A single iteration of dispatching Patina components, retaining those that were not dispatched.
     /// 2. A single iteration of dispatching UEFI drivers via the dispatcher module.
-    fn core_dispatcher(&self) -> Result<()> {
+    fn core_dispatcher(&'static self) -> Result<()> {
         perf_function_begin(function!(), &CALLER_ID, create_performance_measurement);
         loop {
             // Patina component dispatch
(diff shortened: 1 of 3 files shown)


patina_dxe_core/src/pi_dispatcher/image.rs
```diff
@@ -654,13 +657,490 @@ impl DxeCoreGlobalImageData {
...
+        perf_load_image_end(handle, create_performance_measurement);
+
+        match security_status {
+            Err(EfiError::SecurityViolation) => Err(ImageStatus::SecurityViolation(handle)),
+            Err(_) => unreachable!(), // other errors handled above
+            _ => Ok(handle),
+        }
+    }
+
+    // Loads the image specified by the device_path (not yet supported) or
```
joschock: I think `(not yet supported)` might be stale? At least, I think there is support for loading from a buffer located with a device path.

patina_dxe_core/src/lib.rs
```diff
@@ -226,7 +225,7 @@ pub trait MemoryInfo {
     type MemoryInfo = MockMemoryInfo;
     type CpuInfo = MockCpuInfo;
 ))]
-pub trait PlatformInfo {
+pub trait PlatformInfo: 'static {
```
joschock, line 228: for my own understanding, what is the requirement for the lifetime here?

Javagedes: It is because `PlatformInfo` is a trait that can technically be implemented by anything. It could be implemented by a concrete type, or by a type with a lifetime:

```rust
impl PlatformInfo for MyStruct { }

impl<'a> PlatformInfo for &'a MyStruct { }

impl<'a> PlatformInfo for OtherStruct<'a> { }
```

To ensure the implementor of `P` lives long enough for these extern EFIAPI functions that use `Self::instance()`, PlatformInfo must also be static. I don't necessarily to set the `'static` requirement on `PlatformInfo` trait itself. I could technically set it as a requirement on the `impl` blocks of the `Core` / other structs (e.g. `impl<P: PlatformInfo + 'static> Core<P> { ... }`. However It is much cleaner and easier to just set the requirement on the trait itself, because there is not really a scenario where it does not need to be static.

This is only just now being caught because this is the first time we are calling a static function from a `Self::instance()` getter (That is coming from the fact that `start_image` needs to be `'static self` to satisfy corosensei).

makubacki (approved):

Issue #1092, opened 2025-11-19
[Task]: Evaluate running clippy against `not(test)` code


makubacki:
### Request Description

Clippy is currently running in `cargo make clippy` with test enabled using the command defined in this task:

```toml
[tasks.clippy]
description = "Run cargo clippy."
clear = true
command = "cargo"
args = ["clippy", "--all-targets", "--all-features", "--", "-D", "warnings"]
```

`--all-targets` includes tests, so code that is conditionalized with `not(test)` is not run against clippy. This task tracks evaluating this behavior to determine if a chance should be made to include running clippy against that code.

### Are you going to make the change?

Someone else needs to make the change

### Do you need maintainer feedback?

No maintainer feedback needed

hashcatHitman:
Can you provide an example of the issue you're having? I tried looking into this but I couldn't reproduce it. Here's my setup:

In `patina_dxe_core/src/lib.rs`, I added:

```rust
#[allow(unused)]
fn normal() {
    let the_answer: u8 = 42;
    let ref_answer: &u8 = &the_answer;
    let ptr_answer: *const u8 = core::ptr::from_ref(ref_answer);
    let the_answer_2: u8 = unsafe { *ptr_answer }; // CLIPPYCHECK: NORMAL
}

#[allow(unused)]
#[cfg(not(test))]
fn not_test() {
    let the_answer: u8 = 42;
    let ref_answer: &u8 = &the_answer;
    let ptr_answer: *const u8 = core::ptr::from_ref(ref_answer);
    let the_answer_2: u8 = unsafe { *ptr_answer }; // CLIPPYCHECK: CFG NOT TEST
}

#[allow(unused)]
#[cfg(test)]
fn cfg_test() {
    let the_answer: u8 = 42;
    let ref_answer: &u8 = &the_answer;
    let ptr_answer: *const u8 = core::ptr::from_ref(ref_answer);
    let the_answer_2: u8 = unsafe { *ptr_answer }; // CLIPPYCHECK: CFG TEST
}

#[allow(unused)]
#[test]
fn test() {
    let the_answer: u8 = 42;
    let ref_answer: &u8 = &the_answer;
    let ptr_answer: *const u8 = core::ptr::from_ref(ref_answer);
    let the_answer_2: u8 = unsafe { *ptr_answer }; // CLIPPYCHECK: TEST
}

#[allow(unused)]
#[cfg(feature = "compatibility_mode_allowed")]
fn feature() {
[...]

PR #1214, opened 2026-01-06
Small cleanup of compilation blockers and warnings for `aarch64`

hashcatHitman:
## Description

As part of the investigation into and work on #1092, I moved `clippy` to the "Rust" CI job, causing it to be run for each runner. This was accomplished in [patina-devops#53](https://github.com/OpenDevicePartnership/patina-devops/pull/53), which is part of the v0.3.7 release of patina-devops.

Running `clippy` on the `aarch64` runner exposes a handful of issues that previously went unnoticed, which will cause CI to fail when we move to the v0.3.7 release of patina-devops. In preparation for that move, this PR resolves a two categories of failure:

1. Unresolved modules/crates (E0432 and E0433), which was caused by the `doc` feature enabling code which used modules/dependencies that were gated behind an architecture but not the `doc` feature.
2. Needless return warnings.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- `cargo make all` locally, with the `deny` command commented out (as it is currently blocking me on RUSTSEC-2025-0137, which is unrelated to what this PR does).
- A CI run using v0.3.7 of patina-devops ([found here](https://github.com/hashcatHitman/patina/actions/runs/20736376748)). It still fails due to the remaining issues (dead code and missing safety comments), but otherwise passes.

## Integration Instructions
N/A

### DIFF
--- a/core/patina_internal_cpu/src/cpu/aarch64/cpu.rs
+++ b/core/patina_internal_cpu/src/cpu/aarch64/cpu.rs
@@ -93,7 +93,7 @@ impl EfiCpuAarch64 {
                     asm!("mrs {}, ctr_el0", out(reg) ctr_el0);
                     ctr_el0
                 };
-                return 4 << ((ctr_el0 >> 16) & 0xf);
+                4 << ((ctr_el0 >> 16) & 0xf)
             } else {
                 // For test mode or non-aarch64 platforms, return 64 bytes
                 64_u64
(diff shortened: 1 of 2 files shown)


joschock (approved):

makubacki:
Note: I am overriding the patch coverage check for the line noted as missing coverage in this change.

PR #1216, opened 2026-01-06
[REBASE&FF] patina_debugger: allow capturing state with `add_monitor_command`

Javagedes:
## Description

This commit allows users to add the `move` keyword to capture state in a monitor command, instead of needing to store it in a global static for access inside the command

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

CI, continues to boot.

## Integration Instructions

There is no integration changes necessary. This just enables the usage of the keyword `move`:

```rust
let x: u32 = 5;
patina_debugger::add_montior_command("custom_command", "Description", move |args, writer| {
    let _ = write!(writer, "Captured state: {x}");
});
```

### DIFF
--- a/patina_dxe_core/src/config_tables/debug_image_info_table.rs
+++ b/patina_dxe_core/src/config_tables/debug_image_info_table.rs
@@ -93,7 +93,6 @@ const IMAGE_INFO_TABLE_SIZE: usize = 128; // initial size of the table
 /// Metadata structure for the DebugImageInfoTable, which contains the actual table and its size. It is only used
 /// internally to manage the table and is not part of the UEFI spec.
 struct DebugImageInfoTableMetadata<'a> {
-    dbg_system_table_pointer_address: efi::PhysicalAddress,
     actual_table_size: u32,
     table: &'a mut DebugImageInfoTableHeader,
     slice: Box<[EfiDebugImageInfo]>,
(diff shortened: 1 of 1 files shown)


core/patina_debugger/src/debugger.rs
```diff
@@ -412,13 +412,14 @@ impl<T: SerialIO> Debugger for PatinaDebugger<T> {
...
         description: &'static str,
-        callback: crate::MonitorCommandFn,
+        callback: Box<crate::MonitorCommandFn>,
     ) {
         if !self.enabled() {
             return;
         }
 
         self.system_state.lock().add_monitor_command(command, description, callback);
+        log::info!("Added debugger monitor command: {command}");
```
os-d: I don't think this is needed, you can query all the monitor commands from the debugger

Javagedes: It was an existing log that I moved here after moving where the conditional compilation for the "alloc" feature. Happy to delete it, but it is technically a loss of a log compared to prior to this change.

os-d: Let's delete it, save on binary size :)

cfernald (approved):

PR #1217, opened 2026-01-06
patina_dxe_core: Move system_table_ptr add_monitor_command.

Javagedes:
## Description

This commit moves where the monitor (and logic to produce) the EFI system
table pointer from the debug_image_info_table module to the systemtables
module, which logically makes more sense, and also allows for further
separation of the Pi Dispatcher.

blocked by #1216 

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

CI, Boot to shell

## Integration Instructions

N/A

### DIFF
--- a/patina_dxe_core/src/config_tables/debug_image_info_table.rs
+++ b/patina_dxe_core/src/config_tables/debug_image_info_table.rs
@@ -8,16 +8,11 @@
 //!
 extern crate alloc;
 use alloc::{boxed::Box, vec, vec::Vec};
-use patina::base::UEFI_PAGE_SIZE;
 use spin::RwLock;
 
-use core::{ffi::c_void, fmt::Debug, mem::size_of, ptr};
+use core::{ffi::c_void, fmt::Debug, ptr};
 
-use crate::{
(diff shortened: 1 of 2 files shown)


Javagedes:
> I somewhat disagree with this move. I actually think the system table pointer publishing makes more sense in the debug image info table. In the UEFI spec, these two live together in the same section (whereas the system table section is a totally different section). There are diagrams showing why the system table pointer structure exists and how it relates to the debug image info table. So just from a pure spec point of view, it makes sense to keep it colocated with the debug image info table. https://uefi.org/specs/UEFI/2.10_A/18_Protocols_Debugger_Support.html#efi-debug-support-table
> 
> If it truly makes something better or easier in the code (though it looks like it is roughly just a copy/paste), then I'm okay with moving it, but I think we need a justification to move it from the UEFI spec colocated spot.

I think this makes sense and I don't think I necessarily need to move it. I will close this now and continue my other work to see if it can stick around as is.

PR #1219, opened 2026-01-06
debug_image_info_table: Refactor table to own its own memory

Javagedes:
## Description

This commit re-writes debug_image_info_table in preparation for moving it underneath the PiDispatcher module. The main goal of this refactor is to allow the DebugImageInfoData structure to manage it's own memory instead of delegating to a `Box<[...]>`. `Box<[...]>` is typically only used for types that do not grow over time because adding or removing from the slice requires a new allocation. Since we add / remove entries frequently (though not a hot path), we should not be using a `Box>[...]>`.

Instead, we now manage our own allocation manually for this structure by tracking the not only the current size in the header, but also the current capacity outside of the header. This allows us to add or remove entries without needing continuously re-allocate.

To properly review this PR, the main focus should be `patina_dxe_core/src/config_tables/debug_image_info_table.rs`. I would look at this as a standalone file addition. There are really no similarities between the old and new implementation.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

1. Ran miri on all tests in `debug_image_info_table`.
2. Passes CI. Continues to boot to shell.
3. Debugger functionality still works

## Integration Instructions

N/A

### DIFF
--- a/patina_dxe_core/src/config_tables/debug_image_info_table.rs
+++ b/patina_dxe_core/src/config_tables/debug_image_info_table.rs
@@ -45,6 +45,13 @@ impl From<ImageInfoType> for u32 {
 ///
 /// - `header.table_size` and `capacity` always reflect the actual state of the allocated bytes buffer pointed to by
 ///   `header.table`, ensuring no out-of-bounds access can occur.
+///
+/// ## Warning
+///
+/// The above invariants are only upheld on the assumption that this struct is the sole modifier of the underlying
+/// table. This cannot be guaranteed due to the fact that the table pointer ([DebugImageInfoTableHeader]) is exposed
+/// publicly via the UEFI configuration table mechanism. It is expected that this table is read-only when accessed
+/// via this mechanism, but this cannot be enforced.
 pub struct DebugImageInfoData {
     /// The header of the debug image info table, which is registered as a UEFI configuration table.
     header: DebugImageInfoTableHeader,
--- a/patina_dxe_core/src/pi_dispatcher/image.rs
+++ b/patina_dxe_core/src/pi_dispatcher/image.rs
@@ -973,7 +973,15 @@ impl<P: super::PlatformInfo> super::PiDispatcher<P> {
         }
         let handles = PROTOCOL_DB.locate_handles(None).unwrap_or_default();
 
-        DEBUG_IMAGE_INFO_TABLE.write().remove_entry(image_handle);
+        // Remove the debug image info table entry for this image.
+        if let Some(mut table) = DEBUG_IMAGE_INFO_TABLE.try_write() {
+            table.remove_entry(image_handle);
+        } else {
+            debug_assert!(
+                false,
+                "Failed to remove debug image info table entry during unload_image, re-entrant lock detected."
+            );
+        }
 
         // close any protocols opened by this image.
         for handle in handles {


patina_dxe_core/src/config_tables/debug_image_info_table.rs
```diff
@@ -6,392 +6,514 @@
...
+        &mut self,
+        image_info_type: ImageInfoType,
+        protocol: NonNull<efi::protocols::loaded_image::Protocol>,
+        handle: efi::Handle,
+    ) {
+        self.lock_table();
 
-        ptr::write_volatile(&mut (*ptr).crc32, crc32);
+        if self.len() == self.capacity() && self.grow().is_err() {
+            log::error!("Failed to add Debug Image Entry for loaded image protocol @ [{:p}]", protocol);
```
joschock: nit: the `err` result from self.grow() (if it occurs) is probably more interesting for logging than the `protocol` being added.

patina_dxe_core/src/config_tables/debug_image_info_table.rs
```diff
@@ -6,392 +6,514 @@
...
 
-    patina_debugger::add_monitor_command("system_table_ptr", "Prints the system table pointer", move |_, out| {
-        let _ = write!(out, "{address:x}");
-    });
-}
+    /// Removes the first entry matching the specified handle.
+    pub fn remove_entry(&mut self, handle: efi::Handle) {
+        self.lock_table();
+        if let Some(index) = self.find(handle) {
+            let _ = self.swap_remove(index);
```
joschock: Per spec, I think this is fine, but I wonder if there are caller assumptions about ordering? In the EDK2 implementation, the table entry is just marked with a "NULL" NormalImage handle on deletion, but overall table structure/ordering of entries isn't changed: https://github.com/microsoft/mu_basecore/blob/63bb30b4398ce2ac91383040bde9c2ed62795211/MdeModulePkg/Core/Dxe/Misc/DebugImageInfo.c#L271

patina_dxe_core/src/config_tables/debug_image_info_table.rs
```diff
@@ -6,392 +6,514 @@
...
+            // SAFETY: layout is non-zero sized due to DEFAULT_CAPACITY being non-zero
+            unsafe { alloc::alloc::alloc_zeroed(layout) }
+        } else {
+            let old_layout = Layout::array::<EfiDebugImageInfo>(self.capacity)?;
+            let new_capacity = self.capacity * 2;
+            let new_layout = Layout::array::<EfiDebugImageInfo>(new_capacity)?;
+            self.capacity = new_capacity;
+            // SAFETY: layout is the same layout that was used to allocate the original buffer due to the invariants
+            //   of this struct.
+            // SAFETY: new_size is greater than zero du to the `if` branch above ensuring capacity is non-zero.
```
joschock:
```suggestion
            // SAFETY: new_size is greater than zero due to the `if` branch above ensuring capacity is non-zero.
```

patina_dxe_core/src/config_tables/debug_image_info_table.rs
```diff
@@ -6,392 +6,514 @@
...
+    /// Status flag indicating the table has been modified.
+    pub const EFI_DEBUG_IMAGE_INFO_TABLE_MODIFIED: u32 = 0x2;
+
+    /// Creates a new, empty Debug Image Info Table Header.
+    const fn new() -> Self {
+        Self { update_status: 0, table_size: 0, table: ptr::null_mut() }
+    }
+
+    /// Returns the current update status.
+    #[inline(always)]
```
joschock: minor nit/open question: as far as I can tell, the general rust idiom seems to be "trust the compiler on inlining" (see e.g. https://std-dev-guide.rust-lang.org/policy/inline.html). Is there a reason to be aggressive about `#[inline(always)]` that I'm missing?

patina_dxe_core/src/config_tables/debug_image_info_table.rs
```diff
@@ -6,392 +6,514 @@
...
-            let boxed_debug_image_info_table =
-                unsafe { Box::from_raw(debug_image_info.normal_image as *mut EfiDebugImageInfoNormal) };
-            drop(boxed_debug_image_info_table);
-
-            if i != table_size - 1 {
-                // if this is not the last entry, we need to move the last entry to this position
-                // SAFETY: This is safe because we are accessing the table and we ensure that it is initialized
-                debug_image_info.normal_image = last_normal_image_ptr;
-            }
+// SAFETY: Access to the mutable pointer is gated behind a method that requires &mut self.
```
joschock:
```suggestion
// SAFETY: Access to the mutable pointer is gated behind methods that requires &mut self.
```

patina_dxe_core/src/config_tables/debug_image_info_table.rs
```diff
@@ -6,392 +6,514 @@
...
+unsafe impl Send for DebugImageInfoTableHeader {}
+// SAFETY: Access to the mutable pointer is gated behind a method that requires &mut self.
```
joschock:
```suggestion
// SAFETY: Access to the mutable pointer is gated behind methods that requires &mut self.
```

patina_dxe_core/src/config_tables/debug_image_info_table.rs
```diff
@@ -6,392 +6,514 @@
...
-            metadata_table.table.table_size -= 1;
-            break;
+/// A union representing different types of debug image info entries.
+///
+/// each variant must start with a `u32` field representing the image info type, which is used to
+/// determine the actual type of the entry to access.
+///
+/// At present, only one type is defined: EfiDebugImageInfoNormal.
+#[repr(C)]
+union EfiDebugImageInfo {
```
joschock, line 337: is there a reason to not just use EfiDebugImageNormal and omit the union in our implementation if we are assuming no external table modifications?

patina_dxe_core/src/config_tables/debug_image_info_table.rs
```diff
@@ -6,392 +6,514 @@
...
-    unsafe {
-        let update_status = metadata_table.table.get_update_status();
-        metadata_table.table.set_update_status(
-            (update_status & !DebugImageInfoTableHeader::EFI_DEBUG_IMAGE_INFO_UPDATE_IN_PROGRESS)
-                | DebugImageInfoTableHeader::EFI_DEBUG_IMAGE_INFO_TABLE_MODIFIED,
-        )
-    };
+    /// Returns the handle associated with this debug image info entry, if the entry type has one.
+    fn handle(&self) -> Option<efi::Handle> {
+        // SAFETY: The invariants of this struct ensure that this pointer is in-fact a valid pointer to EfiDebugImageInfoNormal.
```
joschock, line 364: This safety invariant only holds if we assume this implementation is the only thing in the system which modifies the struct. I think that's reasonable, but we should probably call that out in the safety comment and potentially check the `image_info_type` is consistent with EfiDebugImageInfoNormal.

patina_dxe_core/src/config_tables/debug_image_info_table.rs
```diff
@@ -6,392 +6,514 @@
...
+        let normal = unsafe { &*self.normal };
+        Some(normal.image_handle)
+    }
+}
+
+impl Drop for EfiDebugImageInfo {
+    fn drop(&mut self) {
+        // SAFETY: The invariants of this struct ensure that this pointer was allocated via this instance of EfiDebugImageInfo.
```
joschock, line 372: Again, assuming some other agent in the system didn't stuff an entry in here.

patina_dxe_core/src/config_tables/debug_image_info_table.rs
```diff
@@ -6,392 +6,514 @@
...
-/// This structure is used to store information about a loaded image for debugging purposes.
-#[repr(C)]
-#[derive(Debug)]
-pub struct EfiDebugImageInfoNormal {
-    pub image_info_type: u32,
-    pub loaded_image_protocol_instance: *const efi::protocols::loaded_image::Protocol,
-    pub image_handle: efi::Handle,
-}
+/// The global debug image info table instance.
+pub static DEBUG_IMAGE_INFO_TABLE: RwLock<DebugImageInfoData> = DebugImageInfoData::new_locked();
```
joschock, line 21: There is a potential super edge case with RwLock that could cause issues in some weird universe. UEFI spec (https://uefi.org/specs/UEFI/2.11/07_Services_Boot_Services.html#tpl-restrictions) requires that load_image/start_image run at TPL < TPL_CALLBACK (i.e. TPL_APPLICATION), but permits unload_image to be invoked at TPL <= TPL_CALLBACK. That means, in theory, that unload_image could be interrupting either load_image or start_image and contending for the lock on the table which could lead to deadlock.  Invoking uload_image in a callback seems... unwise. But strict interpretation of the spec allows it.

patina_dxe_core/src/config_tables/debug_image_info_table.rs
```diff
@@ -6,392 +6,514 @@
...
-#[derive(Copy, Clone)]
-pub union EfiDebugImageInfo {
-    image_info_type: *const u32,
-    normal_image: *const EfiDebugImageInfoNormal,
+impl ImageInfoType {
+    /// The UEFI constant representing a normal debug image info entry.
+    const EFI_DEBUG_IMAGE_INFO_TYPE_NORMAL: u32 = 0x1;
+
+    /// Converts the ImageInfoType to its corresponding u32 representation.
+    fn to_u32(&self) -> u32 {
```
cfernald: why not do this using the Into/From trait?

patina_dxe_core/src/config_tables/debug_image_info_table.rs
```diff
@@ -6,392 +6,514 @@
...
+/// Represents a table of debug image information entries.
+///
+/// ## Invariants
+///
+/// - `length` and `capacity` always reflect the actual state of the allocated `data` buffer, ensuring no out-of-bounds
+///   access can occur.
+pub struct DebugImageInfoData {
+    /// The header of the debug image info table. This is what is exposed to UEFI clients.
+    header: DebugImageInfoTableHeader,
+    /// The total capacity of the table.
```
cfernald: nit: please clarify bytes vs entries in the capacity comment.

patina_dxe_core/src/config_tables/debug_image_info_table.rs
```diff
@@ -6,392 +6,514 @@
...
-#[derive(Debug)]
-pub struct DebugImageInfoTableHeader {
-    update_status: u32, // This is made not pub to force volatile access to the field, per UEFI spec
-    pub table_size: u32,
-    pub efi_debug_image_info_table: *const EfiDebugImageInfo,
+/// Represents a table of debug image information entries.
+///
+/// ## Invariants
+///
+/// - `length` and `capacity` always reflect the actual state of the allocated `data` buffer, ensuring no out-of-bounds
```
os-d: Do you mean the `len()` instance function? Also a bit odd to also put `data` in quotes because it isn't a field or a function.

patina_dxe_core/src/config_tables/debug_image_info_table.rs
```diff
@@ -6,392 +6,514 @@
...
+///   access can occur.
+pub struct DebugImageInfoData {
+    /// The header of the debug image info table. This is what is exposed to UEFI clients.
```
os-d: The terminology of `UEFI clients` is odd here, as is saying this is the header, because this encapsulates the header. Perhaps saying something like `Metadata for the debug image info table`

patina_dxe_core/src/config_tables/debug_image_info_table.rs
os-d: I agree with Chris's previous statement that we can let the compiler inline what it wants or not except for cases where we need something inlined to work for const initialization (there are other cases why we might say inline(never) but that's not relevant here). But I don't believe that is the case for any of these, so I would drop it and let the compiler optimize as it sees fit.

patina_dxe_core/src/config_tables/debug_image_info_table.rs
```diff
@@ -6,392 +6,514 @@
...
-                crc32: 0,
-            },
+    /// Returns the current capacity of the table.
+    #[inline(always)]
+    fn capacity(&self) -> usize {
+        self.capacity
+    }
+
+    /// Sets the update-in-progress flag.
+    fn lock_table(&mut self) {
```
os-d: nit: I think the naming of lock vs unlock table is confusing because you've also added a real lock and this has nothing to do with that. I would just keep this named as set_update_status, etc.

patina_dxe_core/src/config_tables/debug_image_info_table.rs
os-d: Load order doesn't matter. I've actually fixed the removal in edk2, there was a glaring bug, just haven't pulled it down to Mu: https://github.com/tianocore/edk2/pull/11013. There was a discussion about whether we cared about load order being correct and the answer was no. The only consumer of this is a debugger, our debugger(s) don't care about the order of this list.

patina_dxe_core/src/config_tables/debug_image_info_table.rs
```diff
@@ -6,392 +6,514 @@
...
-    metadata_table.table.table_size += 1;
+impl Drop for DebugImageInfoData {
+    fn drop(&mut self) {
+        // Call drop on each entry in the table
+        let data = self.table_mut();
+        for i in 0..self.len() {
+            // SAFETY: Invariants of this struct meet the requirements of drop_in_place. e.g.
+            //   - data[i] is owned by this struct and is valid for both reads and writes.
+            //   - data[i] is properly aligned for EfiDebugImageInfo.
+            //   - data[i] is non-noll.
```
os-d:
```suggestion
            //   - data[i] is non-null.
```

patina_dxe_core/src/config_tables/debug_image_info_table.rs
os-d: It doesn't really matter. This is destined for r-efi, which as we know is very pedantic, so if we want to use that definition, then we'll be stuck with the union because it is UEFI spec defined. Even though it is meaningless.

patina_dxe_core/src/pi_dispatcher.rs
```diff
@@ -144,6 +154,58 @@ impl<P: PlatformInfo> PiDispatcher<P> {
...
+        // This includes installing the debug image info table and the system table pointer structure.
+        if core_install_configuration_table(
+            EFI_DEBUG_IMAGE_INFO_TABLE_GUID,
+            DEBUG_IMAGE_INFO_TABLE.read().header() as *const _ as *mut c_void,
+            system_table,
+        )
+        .is_err()
+        {
+            log::error!("Failed to install configuration table for EFI_DEBUG_IMAGE_INFO_TABLE_GUID");
+            return;
```
os-d: We should continue here if we fail, the system will still boot, just be harder to debug. debug_assert! is fine. Same for the other failures here.

os-d (commented):
I think this approach looks reasonable, as noted earlier, I just want to make sure the debugger still works before approving.

patina_dxe_core/src/config_tables/debug_image_info_table.rs
```diff
@@ -6,392 +6,514 @@
...
+    fn update_status(&self) -> u32 {
+        // SAFETY: This is a field owned by this struct and is valid for reads.
+        unsafe { ptr::read_volatile(&self.update_status) }
+    }
+
+    /// Sets the update status.
+    #[inline(always)]
+    fn set_update_status(&mut self, status: u32) {
+        // SAFETY: This is a field owned by this struct and is valid for writes.
+        unsafe { ptr::write_volatile(&mut self.update_status, status) }
```
cfernald, line 287: I know this isn't new to your PR, but I wonder if the volatile here actually makes any difference. Especially now with the real lock.

patina_dxe_core/src/config_tables/debug_image_info_table.rs
```diff
@@ -6,392 +6,514 @@
...
-    // SAFETY: This is safe because we are accessing the table header and we ensure that it is initialized
-    let debug_image_info = &mut metadata_table.slice[metadata_table.table.table_size as usize];
-    let debug_image_info_table =
-        Box::new(EfiDebugImageInfoNormal { image_info_type, loaded_image_protocol_instance, image_handle });
+    /// Doubles the current capacity of the table.
+    ///
+    /// If the current capacity is zero, sets it to a default initial capacity.
+    fn grow(&mut self) -> Result<(), alloc::alloc::LayoutError> {
+        // DEFAULT_CAPACITY must always be greater than zero.
+        const DEFAULT_CAPACITY: usize = 4;
```
cfernald: Maybe make the default a bit larger. We will really never have so few modules so we can probably optimize a bit. Maybe 16?

Javagedes:
> I'll review more tomorrow, but did you test the debugger works after this change? i.e. doing something like
> 
> ```
> !uefiext.modulebreak SmbiosDxe
> g
> !findall
> 
> x SmbiosDxe!*
> ```
> 
> And it is able to load the symbols for all images?

@os-d I have a bunch of updates to make from this good feeback, but I did take a moment to check the debugger. All is working well from your requested checks. SmbiosDxe no longer exists, but I used another module :)

patina_dxe_core/src/config_tables/debug_image_info_table.rs
Javagedes: There was no real reason other than I have a tendency to always put `#[inline(always)]` on one-liner functions because I have an unreasonable fear of hitting a weird compiler quirk where we jump to a function just to immediately jump to another.

These are good points though. I've removed the `#[inline(always)]` statements :)

patina_dxe_core/src/config_tables/debug_image_info_table.rs
Javagedes: Good point. I've updated to reference `header.table_size` instead of `length`. I've also updated the documentation regarding the allocated `data` buffer to instead say "allocated bytes buffer pointed to by `header.table`"

patina_dxe_core/src/config_tables/debug_image_info_table.rs
Javagedes:
``` rust
    /// The header of the debug image info table. This is what is exposed to UEFI clients.
    header: DebugImageInfoTableHeader,
```

I agree the wording of "UEFI Clients" is weird, but this is the exact header that is published as the configuration table, so that portion of this statement seems accurate, unless I'm misunderstanding your concern.

patina_dxe_core/src/config_tables/debug_image_info_table.rs
Javagedes: I wanted to limit the scope of the `const EFI_DEBUG_IMAGE_INFO_TYPE_NORMAL` to just the enum, so I decided to make the u32 conversion a standalone function instead of the trait.

I should use the Into trait though. I will update.

patina_dxe_core/src/config_tables/debug_image_info_table.rs
os-d: The reason this is volatile is because the UEFI spec defines update_status as volatile:

image

Unlike in C, you can't mark a struct field volatile in Rust, so this was to meet UEFI spec compliance. From our debugger, we don't care about the update status field at all, so we could probably get rid of it, but again, to meet the pedantry of r-efi, I was trying to directly meet the UEFI spec.

patina_dxe_core/src/config_tables/debug_image_info_table.rs
os-d: Agreed on the header, I was reading the comment as for struct DebugImageInfoData, not the header field within it. I would just change the UEFI clients terminology.

os-d (approved):

patina_dxe_core/src/config_tables/debug_image_info_table.rs
Javagedes: Good call. Updated to 16, but I'd be willing to go hire if you prefer. I doubt any system would have modules less than 100 to be honest.

patina_dxe_core/src/config_tables/debug_image_info_table.rs
Javagedes: Clarified the total capacity as the total number of [EfiDebugImageInfo] entries able to be added to the table before a reallocation is needed!

patina_dxe_core/src/config_tables/debug_image_info_table.rs
Javagedes: I will keep this is then. Thanks for the explanation

patina_dxe_core/src/config_tables/debug_image_info_table.rs
Javagedes: Updated to log the grow error rather than the particular protocol

patina_dxe_core/src/config_tables/debug_image_info_table.rs
Javagedes: Thank you for the clarification. I will close this as my changes seem appropriate as is based off of this.

patina_dxe_core/src/config_tables/debug_image_info_table.rs
Javagedes: I will keep this as-is per above comments.

joschock: FWIW, we can't upstream it as implemented now, because it'd violate the no trait impls on foreign types rule.

patina_dxe_core/src/config_tables/debug_image_info_table.rs
Javagedes: Added remarks regarding this to the `DebugImageInfoData` struct.

patina_dxe_core/src/config_tables/debug_image_info_table.rs
Javagedes: Updated to do a `try_write()`. If it fails, we do a debug assert for re-entrant locking.

PR #1221, opened 2026-01-07
Allow consumer to pass in Smbios handle (not FFFEh) as stated in PI spec EFI_SMBIOS_PROTOCOL.Add()

AnhDLuong:
## Description

Allow consumer to pass in Smbios handle instead of always assigning a unique Smbios handle.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [X] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Boot to shell.

## Integration Instructions

N/A

### DIFF
--- a/components/patina_smbios/src/error.rs
+++ b/components/patina_smbios/src/error.rs
@@ -38,8 +38,6 @@ pub enum SmbiosError {
     // Handle management errors
     /// All available handles have been exhausted (reached 0xFFFE limit)
     HandleExhausted,
-    /// The specified handle was not found in the record table
-    HandleNotFound,
     /// String index is out of range for the specified record
     StringIndexOutOfRange,
     /// The specified handle was already used by another record
@@ -93,9 +91,7 @@ impl From<SmbiosError> for patina::error::EfiError {
             | SmbiosError::HandleOutOfRange => patina::error::EfiError::InvalidParameter,
 
             // Not found errors map to NOT_FOUND
-            SmbiosError::HandleNotFound | SmbiosError::NoRecordsAvailable | SmbiosError::RecordNotFound => {
-                patina::error::EfiError::NotFound
-            }
+            SmbiosError::NoRecordsAvailable | SmbiosError::RecordNotFound => patina::error::EfiError::NotFound,
 
             // Version and initialization errors map to UNSUPPORTED
             SmbiosError::UnsupportedVersion | SmbiosError::AlreadyInitialized | SmbiosError::NotInitialized => {
@@ -127,14 +123,17 @@ mod tests {
             SmbiosError::InvalidStringPoolTermination,
             SmbiosError::StringPoolTooSmall,
             SmbiosError::HandleExhausted,
-            SmbiosError::HandleNotFound,
+            SmbiosError::RecordNotFound,
             SmbiosError::StringIndexOutOfRange,
             SmbiosError::AllocationFailed,
+            SmbiosError::HandleExhausted,
             SmbiosError::NoRecordsAvailable,
             SmbiosError::AlreadyInitialized,
             SmbiosError::NotInitialized,
             SmbiosError::UnsupportedVersion,
             SmbiosError::Type127Managed,
+            SmbiosError::HandleInUse,
+            SmbiosError::HandleOutOfRange,
             SmbiosError::TableDirectlyModified,
         ];
 
@@ -184,8 +183,14 @@ mod tests {
         let efi_err: patina::error::EfiError = SmbiosError::Type127Managed.into();
         assert_eq!(efi_err, patina::error::EfiError::InvalidParameter);
 
+        let efi_err: patina::error::EfiError = SmbiosError::HandleInUse.into();
+        assert_eq!(efi_err, patina::error::EfiError::InvalidParameter);
+
+        let efi_err: patina::error::EfiError = SmbiosError::HandleOutOfRange.into();
+        assert_eq!(efi_err, patina::error::EfiError::InvalidParameter);
+
         // Test not found errors map to NOT_FOUND
-        let efi_err: patina::error::EfiError = SmbiosError::HandleNotFound.into();
+        let efi_err: patina::error::EfiError = SmbiosError::RecordNotFound.into();
         assert_eq!(efi_err, patina::error::EfiError::NotFound);
 
         let efi_err: patina::error::EfiError = SmbiosError::NoRecordsAvailable.into();
--- a/components/patina_smbios/src/manager/core.rs
+++ b/components/patina_smbios/src/manager/core.rs
@@ -12,7 +12,7 @@
 
 extern crate alloc;
 
-use alloc::{boxed::Box, string::String, vec::Vec};
+use alloc::{boxed::Box, collections::BTreeSet, string::String, vec::Vec};
 use core::cell::RefCell;
 use patina::{base::SIZE_64KB, uefi_size_to_pages};
 use r_efi::{
@@ -73,7 +73,7 @@ pub struct Smbios30EntryPoint {
 /// Manages SMBIOS records, handles, and table generation.
 pub struct SmbiosManager {
     pub(super) records: RefCell<Vec<SmbiosRecord>>,
-    pub(super) used_handles: RefCell<Vec<SmbiosHandle>>,
+    pub(super) used_handles: RefCell<BTreeSet<SmbiosHandle>>,
     pub major_version: u8,
     pub minor_version: u8,
     entry_point_64: RefCell<Option<Box<Smbios30EntryPoint>>>,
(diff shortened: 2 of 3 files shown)


components/patina_smbios/src/manager/core.rs
```diff
@@ -73,8 +73,7 @@ pub struct Smbios30EntryPoint {
 /// Manages SMBIOS records, handles, and table generation.
 pub struct SmbiosManager {
     pub(super) records: RefCell<Vec<SmbiosRecord>>,
-    pub(super) next_handle: RefCell<SmbiosHandle>,
-    pub(super) freed_handles: RefCell<Vec<SmbiosHandle>>,
+    pub(super) used_handles: RefCell<Vec<SmbiosHandle>>,
```
kat-perez: Consider using a HashSet to improve performance: 
`pub(super) used_handles: RefCell<HashSet<SmbiosHandle>>`. This would require adding hashbrown for no_std compatibility though.

Other option could be add a next_handle
```
pub(super) used_handles: RefCell<Vec<SmbiosHandle>>,
pub(super) next_handle: RefCell<SmbiosHandle>,
```

components/patina_smbios/src/manager/core.rs
```diff
@@ -659,14 +711,20 @@ impl SmbiosManager {
...
-        if (1..SMBIOS_HANDLE_PI_RESERVED).contains(&smbios_handle) {
-            self.freed_handles.borrow_mut().push(smbios_handle);
+        // update the self.used_handles
+        if (0..SMBIOS_HANDLE_PI_RESERVED).contains(&smbios_handle) {
+            let pos_handle = self
+                .used_handles
+                .borrow()
+                .iter()
+                .position(|h| *h == smbios_handle)
+                .ok_or(SmbiosError::HandleNotFound)?;
```
kat-perez: This lookup should never fail—if a record exists, its handle was registered in used_handles when it was added. Consider debug_assert! instead, since returning `HandleNotFound` here leaves the system inconsistent (record already removed at line 716).

components/patina_smbios/src/manager/core.rs
```diff
@@ -931,38 +989,71 @@ mod tests {
...
+        record_data.extend_from_slice(b"\0\0");
+        let handle1 = manager.add_from_bytes(None, &record_data).expect("add failed");
         assert_eq!(handle1, 1);
-        let handle2 = manager.allocate_handle().expect("allocation failed");
-        assert_eq!(handle2, 2);
     }
 
     #[test]
-    fn test_allocate_handle_wraps_at_max() {
+    fn test_add_request_handle_error_handlealreadystarted() {
```
kat-perez: nit: test naming - `test_add_request_handle_error_handle_already_started`

components/patina_smbios/src/manager/core.rs
```diff
@@ -297,32 +295,82 @@ impl SmbiosManager {
...
+    /// already in use. If it is not, then use the requested handle as is.
+    ///
+    /// # Arguments
+    ///
+    /// * `&SmbiosHandle` - reference to the requested handle
+    ///
+    /// # Returns
+    ///
+    /// Returns `Ok(SmbiosHandle)` with the assigned handle, or an error if the handle is already
+    /// in used.
```
kat-perez: "in used" -> "in use"

components/patina_smbios/src/manager/core.rs
```diff
@@ -659,14 +711,20 @@ impl SmbiosManager {
...
+            .ok_or(SmbiosError::RecordNotFound)?;
 
         self.records.borrow_mut().remove(pos);
 
-        // Add the freed handle to the free list for reuse
-        // Only add valid handles (1..0xFFFE) to the free list
-        if (1..SMBIOS_HANDLE_PI_RESERVED).contains(&smbios_handle) {
-            self.freed_handles.borrow_mut().push(smbios_handle);
+        // update the self.used_handles
+        if (0..SMBIOS_HANDLE_PI_RESERVED).contains(&smbios_handle) {
```
kat-perez: Could remove this redundant check since `if (0..SMBIOS_HANDLE_PI_RESERVED).contains(&smbios_handle)` is always true since we already found the record and valid records can't have reserved handles.

components/patina_smbios/src/manager/core.rs
```diff
@@ -516,6 +564,7 @@ impl SmbiosManager {
     ) -> Result<SmbiosHandle, SmbiosError> {
         // Step 1: Validate minimum size for header (at least 4 bytes)
         if record_data.len() < core::mem::size_of::<SmbiosTableHeader>() {
+            log::error!("Error - minimum size for header is too small");
```
kat-perez: nit: log message format is inconsistent—most use function_name - Error but line 567 uses Error - description. Consider making them consistent.

components/patina_smbios/src/manager/core.rs
joschock: FWIW, BtreeSet is available in no-std.

components/patina_smbios/src/error.rs
```diff
@@ -40,6 +42,10 @@ pub enum SmbiosError {
     HandleNotFound,
     /// String index is out of range for the specified record
     StringIndexOutOfRange,
+    /// The specified handle was already used by another record
+    HandleAlreadyStarted,
```
joschock: consider `DuplicateHandle`?

joschock: or `HandleInUse`

joschock (approved):

components/patina_smbios/src/manager/core.rs
AnhDLuong: Agree, will move this check to the beginning of the function, before the code that find record

components/patina_smbios/src/manager/core.rs
AnhDLuong: Does the following work? 

        let pos_handle = match self.used_handles.borrow().iter().position(|h| *h == smbios_handle) {
            Some(p) => p,
            None => {
                debug_assert!(
                    false,
                    "If a SMBIOS record exists, there must be a SMBIOS handle associated with that record"
                );
                return Err(SmbiosError::HandleNotFound);
            }
        };

components/patina_smbios/src/error.rs
AnhDLuong: I like HandleInUse

components/patina_smbios/src/manager/core.rs
AnhDLuong: Ack

components/patina_smbios/src/manager/core.rs
AnhDLuong: Ack

components/patina_smbios/src/manager/core.rs
AnhDLuong: Ack

components/patina_smbios/src/manager/core.rs
kat-perez: If we return Err in production builds, we'll still have an inconsistent state. So we'll need to do something like this instead (added comments just for explanation but you can remove):

```
      // Remove from used_handles tracking
      match self.used_handles.borrow().iter().position(|h| *h == smbios_handle) {
          Some(pos_handle) => {
              self.used_handles.borrow_mut().swap_remove(pos_handle);
          }
          None => {
              // This should never happen - if a record exists, its handle was registered
              // in used_handles when it was added via get_smbios_handle().
              debug_assert!(
                  false,
                  "If an SMBIOS record exists, there must be a SMBIOS handle associated with that record"
              );
              // In release: continue gracefully since the record is already removed
          }
      }

      Ok(())
```

That way if there's an error in debug builds, it will panic and we can catch the bug. Release builds won't leave the system in a broken state

AnhDLuong: Can you please explain what's the inconsistent state we will see if we return Err vs return Ok in production build ? IMO, we should not assert in production if it's not critical, but we should not drop the error.

kat-perez: I'm thinking about this in terms of expecting success or failure of the remove. If the caller receives an error, they might think the remove failed when it actually succeeded. If the caller's code did this
```
if let Err(e) = manager.remove(handle) {
    log::error!("Failed to remove record, will retry...");
    // This is inconsistent because the record IS gone
 }
 ```
 
 The error at the top `SmbiosError::RecordNotFound` definitely should be kept and `debug_assert!` will be compiled out in a production build so it will not assert at all in a release

AnhDLuong: I updated with BTreeSet implementation

makubacki:
Since all approvals are in with no open conversations I've rebased again and set autocomplete.

PR #1225, opened 2026-01-08
[feature/patina-boot] patina_boot: Add boot orchestration crate

kat-perez:
## Description

Adds BootOrchestration component, simple console discovery, simple BootOption Config

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [x] Includes documentation?

## How This Was Tested

QEMU Platform Integration:
- Q35
- SBSA

## Integration Instructions

N/A

### DIFF
--- a/components/patina_boot/README.md
+++ b/components/patina_boot/README.md
@@ -1,0 +1,33 @@
+# Patina Boot
+
+Boot orchestration component for Patina-based firmware implementing UEFI boot manager functionality.
+
+## Components
+
+- **BootOrchestrator**: Orchestrates device enumeration, BDS phase events, and boot option execution.
+- **ConsoleDiscovery**: Discovers console devices and populates UEFI console variables.
+
+## Usage
+
+```rust
+use patina_boot::{component::BootOrchestrator, config::BootOptions};
+
+// Configure boot options (devices are tried in order)
+let config = BootOptions::new()
+    .with_device(primary_boot_path)
+    .with_device(fallback_boot_path)
+    .with_failure_handler(|| { /* handle boot failure */ });
+
+// Add BootOrchestrator as a platform component
+add.component(BootOrchestrator);
+```
+
+## Helper Functions
+
+For custom boot flows, use the helper functions in the `helpers` module:
+
+- `connect_all()` - Connect all controllers for device enumeration
+- `signal_bds_phase_entry()` - Signal EndOfDxe event
+- `signal_ready_to_boot()` - Signal ReadyToBoot event
+- `discover_console_devices()` - Populate console variables
+- `boot_from_device_path()` - Load and start a boot image
--- a/components/patina_boot/src/component/boot_orchestrator.rs
+++ b/components/patina_boot/src/component/boot_orchestrator.rs
@@ -1,0 +1,104 @@
+//! Boot orchestrator component.
+//!
+//! Orchestrates the boot flow with device enumeration, event signaling, and boot execution.
+//!
+//! ## Rationale
+//!
+//! Boot orchestration is a component that represents the primary platform customization
+//! point. Platforms provide boot options as configuration data, allowing different boot policies
+//! without changing orchestration logic. Platforms needing entirely different boot flows can
+//! replace the component.
+//!
+//! ## License
+//!
+//! Copyright (c) Microsoft Corporation.
+//!
+//! SPDX-License-Identifier: Apache-2.0
+//!
+use patina::{
+    boot_services::{BootServices, StandardBootServices, protocol_handler::HandleSearchType},
+    component::{
+        component,
+        params::{Config, Handle},
+    },
+    error::{EfiError, Result},
+    runtime_services::StandardRuntimeServices,
+};
+use r_efi::efi;
+
+use crate::{config::BootOptions, helpers};
+
+/// Boot orchestrator component.
+///
+/// Orchestrates boot execution using boot options provided via [`Config<BootOptions>`].
+/// Connects all controllers for device topology enumeration, signals BDS phase events
+/// (EndOfDxe, ReadyToBoot), and executes boot options via `LoadImage()`/`StartImage()`
+/// with 5-minute watchdog per UEFI Section 3.1.2.
+///
+/// Handles boot failures by attempting subsequent options. Never returns on success.
+pub struct BootOrchestrator;
+
+#[component]
+impl BootOrchestrator {
+    /// Entry point for the boot orchestrator component.
+    ///
+    /// # Flow
+    ///
+    /// 1. Connect all controllers for device enumeration
+    /// 2. Signal EndOfDxe (security components perform lockdown)
+    /// 3. Discover consoles
+    /// 4. Signal ReadyToBoot
+    /// 5. Execute boot options from config
+    /// 6. If all boot options fail, call failure handler
+    #[coverage(off)] // Component integration - tested via integration tests
+    fn entry_point(
+        self,
+        boot_services: StandardBootServices,
+        runtime_services: StandardRuntimeServices,
+        boot_options: Config<BootOptions>,
+        image_handle: Option<Handle>,
+    ) -> Result<()> {
+        helpers::connect_all(boot_services.as_ref())?;
+        helpers::signal_bds_phase_entry(boot_services.as_ref())?;
+        helpers::discover_console_devices(boot_services.as_ref(), runtime_services.as_ref())?;
+        helpers::signal_ready_to_boot(boot_services.as_ref())?;
+
+        // Use the provided image handle, or fall back to finding one via LocateHandleBuffer.
+        // Per UEFI spec, the parent handle must be a valid image handle (has LoadedImage protocol).
+        let parent_handle = if let Some(handle) = image_handle {
+            *handle
+        } else {
+            log::warn!("Handle not provided, falling back to LocateHandleBuffer");
+            let image_handles = boot_services
+                .locate_handle_buffer(HandleSearchType::ByProtocol(&efi::protocols::loaded_image::PROTOCOL_GUID))
+                .map_err(|status| {
+                    log::error!("Failed to locate image handles: {:?}", status);
+                    EfiError::from(status)
+                })?;
+
+            image_handles.first().copied().ok_or_else(|| {
+                log::error!("No image handles available for parent handle");
+                EfiError::NotReady
+            })?
+        };
+
+        for device_path in boot_options.devices() {
+            match helpers::boot_from_device_path(boot_services.as_ref(), parent_handle, device_path) {
+                Ok(()) => {
+                    // Boot image returned control (e.g., EFI application exited).
+                    // Continue to try next boot option.
+                    log::warn!("Boot option returned, trying next...");
+                }
+                Err(_) => {
+                    log::warn!("Boot option failed, trying next...");
+                }
+            }
+        }
+
+        boot_options.handle_failure();
+        log::error!("All boot options exhausted and failure handler returned");
+        loop {
+            core::hint::spin_loop();
+        }
+    }
+}
--- a/components/patina_boot/src/config.rs
+++ b/components/patina_boot/src/config.rs
@@ -1,0 +1,174 @@
+//! Boot configuration types.
+//!
+//! This module provides configuration types for boot orchestration, including
+//! [`BootOptions`] which specifies platform-provided boot paths.
+//!
+//! ## License
+//!
+//! Copyright (c) Microsoft Corporation.
+//!
+//! SPDX-License-Identifier: Apache-2.0
+//!
+extern crate alloc;
+
+use alloc::{boxed::Box, vec::Vec};
+use patina::uefi_protocol::device_path::DevicePathBuf;
+
+/// Boot options provided by the platform.
+///
+/// Platforms configure boot behavior by providing this configuration to the
+/// [`BootOrchestrator`](crate::component::BootOrchestrator) component.
+///
+/// ## Example
+///
+/// ```rust,ignore
+/// use patina_boot::config::BootOptions;
+///
+/// let options = BootOptions::new()
+///     .with_device(nvme_device_path)
+///     .with_device(usb_device_path)
+///     .with_failure_handler(|| show_error_screen());
+/// ```
+#[derive(Default)]
+pub struct BootOptions {
+    /// Boot device paths in priority order.
+    devices: Vec<DevicePathBuf>,
+    /// Optional hotkey for boot override (e.g., F12 for boot menu).
+    hotkey: Option<u16>,
+    /// Handler called when all boot options fail.
+    failure_handler: Option<Box<dyn Fn() + Send + Sync>>,
+}
+
+impl BootOptions {
+    /// Create new empty boot options.
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Add a boot device path.
+    ///
+    /// Device paths are tried in the order they are added.
+    pub fn with_device(mut self, device: DevicePathBuf) -> Self {
+        self.devices.push(device);
+        self
+    }
+
+    /// Add a hotkey scancode for boot override (not yet implemented).
+    ///
+    /// This field is reserved for future boot menu functionality.
+    /// Currently, the hotkey is stored but not acted upon.
+    pub fn with_hotkey(mut self, scancode: u16) -> Self {
+        self.hotkey = Some(scancode);
+        self
+    }
+
+    /// Add a failure handler called when all boot options fail.
+    pub fn with_failure_handler<F>(mut self, handler: F) -> Self
+    where
+        F: Fn() + Send + Sync + 'static,
+    {
+        self.failure_handler = Some(Box::new(handler));
+        self
+    }
+
+    /// Get the hotkey scancode, if configured.
+    pub fn hotkey(&self) -> Option<u16> {
+        self.hotkey
+    }
+
+    /// Returns an iterator over all configured boot device paths.
+    pub fn devices(&self) -> impl Iterator<Item = &DevicePathBuf> {
+        self.devices.iter()
+    }
+
+    /// Call the failure handler if configured.
+    ///
+    /// This is called when all boot options have been exhausted.
+    pub fn handle_failure(&self) {
+        if let Some(handler) = &self.failure_handler {
+            handler();
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    extern crate alloc;
+    extern crate std;
+
+    use super::*;
+    use alloc::sync::Arc;
+    use core::sync::atomic::{AtomicBool, Ordering};
+    use patina::uefi_protocol::device_path::nodes::EndEntire;
+
+    fn create_test_device_path() -> DevicePathBuf {
+        DevicePathBuf::from_device_path_node_iter(core::iter::once(EndEntire))
+    }
+
+    #[test]
+    fn test_default_boot_options() {
+        let options = BootOptions::default();
+        assert!(options.hotkey().is_none());
+        assert_eq!(options.devices().count(), 0);
+    }
+
+    #[test]
+    fn test_new_boot_options() {
+        let options = BootOptions::new();
+        assert_eq!(options.devices().count(), 0);
+    }
+
+    #[test]
+    fn test_with_single_device() {
+        let device = create_test_device_path();
+        let options = BootOptions::new().with_device(device);
+        assert_eq!(options.devices().count(), 1);
+    }
+
+    #[test]
+    fn test_with_multiple_devices() {
+        let device1 = create_test_device_path();
+        let device2 = create_test_device_path();
+        let device3 = create_test_device_path();
+        let options = BootOptions::new().with_device(device1).with_device(device2).with_device(device3);
+        assert_eq!(options.devices().count(), 3);
+    }
+
+    #[test]
+    fn test_with_hotkey() {
+        let options = BootOptions::new().with_hotkey(0x86); // F12
+        assert_eq!(options.hotkey(), Some(0x86));
+    }
+
+    #[test]
+    fn test_failure_handler_called() {
+        let called = Arc::new(AtomicBool::new(false));
+        let called_clone = called.clone();
+
+        let device = create_test_device_path();
+        let options = BootOptions::new().with_device(device).with_failure_handler(move || {
+            called_clone.store(true, Ordering::SeqCst);
+        });
+
+        assert!(!called.load(Ordering::SeqCst));
+        options.handle_failure();
+        assert!(called.load(Ordering::SeqCst));
+    }
+
+    #[test]
+    fn test_failure_handler_not_configured() {
+        let options = BootOptions::default();
+        // Should not panic when no handler is configured
+        options.handle_failure();
+    }
+
+    #[test]
+    fn test_devices_iterator_order() {
+        let device1 = create_test_device_path();
+        let device2 = create_test_device_path();
+        let options = BootOptions::new().with_device(device1).with_device(device2);
+
+        let devices: Vec<_> = options.devices().collect();
+        assert_eq!(devices.len(), 2);
+    }
+}
--- a/components/patina_boot/src/helpers.rs
+++ b/components/patina_boot/src/helpers.rs
@@ -1,0 +1,430 @@
+//! Helper functions for boot orchestration.
+//!
+//! This module provides helper functions for platforms implementing custom boot flows.
+//! The [`BootOrchestrator`](crate::component::BootOrchestrator) component uses these
+//! internally, and platforms can use them directly for custom orchestration.
+//!
+//! ## License
+//!
+//! Copyright (c) Microsoft Corporation.
+//!
+//! SPDX-License-Identifier: Apache-2.0
+//!
+extern crate alloc;
+
+use alloc::vec::Vec;
+use core::ptr;
+
+use patina::{
+    boot_services::{BootServices, event::EventType, protocol_handler::HandleSearchType, tpl::Tpl},
+    error::{EfiError, Result},
+    guids::EVENT_GROUP_END_OF_DXE,
+    runtime_services::RuntimeServices,
+    uefi_protocol::device_path::DevicePathBuf,
+};
+use r_efi::{efi, system::EVENT_GROUP_READY_TO_BOOT};
+
+/// Watchdog timeout in seconds per UEFI Specification Section 3.1.2.
+const WATCHDOG_TIMEOUT_SECONDS: usize = 300; // 5 minutes
+
+/// Load and start a boot image with UEFI spec compliance.
+///
+/// Enables a 5-minute watchdog timer before `StartImage()` per UEFI Specification
+/// Section 3.1.2. Disables watchdog when boot returns control.
+///
+/// # Arguments
+///
+/// * `boot_services` - Boot services interface
+/// * `parent_handle` - Parent image handle for the loaded image (typically the calling image's handle)
+/// * `device_path` - Device path to the boot image
+///
+/// # Returns
+///
+/// Returns `Ok(())` if the boot image was successfully started (which typically
+/// means it returned control). Returns an error if loading or starting fails.
+pub fn boot_from_device_path<B: BootServices>(
+    boot_services: &B,
+    parent_handle: efi::Handle,
+    device_path: &DevicePathBuf,
+) -> Result<()> {
+    // Load the image
+    let device_path_ptr = device_path.as_ref() as *const _ as *mut efi::protocols::device_path::Protocol;
+    let image_handle = match boot_services.load_image(true, parent_handle, device_path_ptr, None) {
+        Ok(handle) => handle,
+        Err(status) => {
+            log::error!("LoadImage failed with status: {:?}", status);
+            return Err(EfiError::from(status));
+        }
+    };
+
+    // Enable 5-minute watchdog timer per UEFI spec Section 3.1.2
+    boot_services.set_watchdog_timer(WATCHDOG_TIMEOUT_SECONDS).map_err(EfiError::from)?;
+
+    // Start the image
+    let result = boot_services.start_image(image_handle);
+
+    // Disable watchdog timer when boot option returns control
+    let _ = boot_services.set_watchdog_timer(0);
+
+    match result {
+        Ok(()) => Ok(()),
+        Err((status, _exit_data)) => Err(EfiError::from(status)),
+    }
+}
+
+/// Connect all controllers recursively for device enumeration.
+///
+/// Connects all handles in the system recursively until the device topology
+/// stabilizes (no new handles are created).
+///
+/// # Arguments
+///
+/// * `boot_services` - Boot services interface
+///
+/// # Returns
+///
+/// Returns `Ok(())` when device topology enumeration is complete.
+pub fn connect_all<B: BootServices>(boot_services: &B) -> Result<()> {
+    // Loop until the number of handles stabilizes, indicating device topology is complete.
+    // This is needed because connecting a PCI bus creates new handles for PCI devices,
+    // which then need to be connected to bind drivers like NVMe, which creates namespace
+    // handles, etc.
+    const MAX_ITERATIONS: usize = 10;
+    let mut prev_handle_count = 0;
+
+    for _iteration in 0..MAX_ITERATIONS {
+        // Get all handles in the system
+        let handles = boot_services.locate_handle_buffer(HandleSearchType::AllHandle).map_err(EfiError::from)?;
+        let current_handle_count = handles.len();
+
+        // Connect each handle recursively
+        for &handle in handles.iter() {
+            // SAFETY: Empty driver handle list and null device path are valid per UEFI spec
+            let _ = unsafe { boot_services.connect_controller(handle, Vec::new(), ptr::null_mut(), true) };
+        }
+
+        // Check if handle count has stabilized
+        if current_handle_count == prev_handle_count {
+            break;
+        }
+
+        prev_handle_count = current_handle_count;
+    }
+
+    Ok(())
+}
+
+/// Signal EndOfDxe event for platforms implementing custom orchestration.
+///
+/// Signals `gEfiEndOfDxeEventGroupGuid` to notify security components that
+/// DXE phase initialization is complete. Security components (e.g., SMM/MM)
+/// register for this event and perform lockdown.
+///
+/// # Arguments
+///
+/// * `boot_services` - Boot services interface
+pub fn signal_bds_phase_entry<B: BootServices>(boot_services: &B) -> Result<()> {
+    // Create and signal EndOfDxe event
+    // SAFETY: Null context is valid for signal-only events
+    let event = unsafe {
+        boot_services.create_event_ex_unchecked::<()>(
+            EventType::NOTIFY_SIGNAL,
+            Tpl::CALLBACK,
+            signal_event_noop,
+            ptr::null_mut(),
+            &EVENT_GROUP_END_OF_DXE,
+        )
+    }
+    .map_err(EfiError::from)?;
+
+    let signal_result = boot_services.signal_event(event);
+    // Always close the event, even if signal failed
+    let close_result = boot_services.close_event(event);
+
+    signal_result.map_err(EfiError::from)?;
+    close_result.map_err(EfiError::from)?;
+
+    Ok(())
+}
+
+/// Signal ReadyToBoot event for platforms implementing custom orchestration.
+///
+/// Signals `gEfiEventReadyToBootGuid` immediately before attempting the first
+/// boot option. This event notifies drivers that boot is imminent.
+///
+/// # Arguments
+///
+/// * `boot_services` - Boot services interface
+pub fn signal_ready_to_boot<B: BootServices>(boot_services: &B) -> Result<()> {
+    // Create and signal ReadyToBoot event
+    // SAFETY: Null context is valid for signal-only events
+    let event = unsafe {
+        boot_services.create_event_ex_unchecked::<()>(
+            EventType::NOTIFY_SIGNAL,
+            Tpl::CALLBACK,
+            signal_event_noop,
+            ptr::null_mut(),
+            &EVENT_GROUP_READY_TO_BOOT,
+        )
+    }
+    .map_err(EfiError::from)?;
+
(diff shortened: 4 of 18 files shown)


kat-perez:
@Javagedes remade my PR but I still addressed all your comments from the previous PR (#1201)

components/patina_boot/src/config.rs
```diff
@@ -0,0 +1,186 @@
...
+/// use patina_boot::config::BootOptions;
+///
+/// let options = BootOptions::new(primary_device_path)
+///     .with_secondary(secondary_device_path)
+///     .with_failure_handler(|| show_error_screen());
+/// ```
+#[derive(Default)]
+pub struct BootOptions {
+    /// Primary boot device path.
+    primary: Option<DevicePathBuf>,
```
joschock: why not just make this a `Vec<DevicePathBuf>` so you can have as many as you like?

components/patina_boot/src/helpers.rs
```diff
@@ -0,0 +1,448 @@
...
+/// topology stabilizes (no new drivers are dispatched).
+///
+/// # Arguments
+///
+/// * `boot_services` - Boot services interface
+///
+/// # Returns
+///
+/// Returns `Ok(())` when device topology enumeration is complete.
+pub fn interleave_connect_and_dispatch<B: BootServices>(boot_services: &B) -> Result<()> {
```
joschock: the function name is a bit strange here, since there is no 'dispatch' that occurs (i.e. the dispatcher is not re-inovked). Consider something like `connect_all`

patina_dxe_core/src/pi_dispatcher/fv.rs
```diff
@@ -103,8 +103,11 @@ impl<P: PlatformInfo> FvProtocolData<P> {
     }
 
     /// Creates a new [TplMutex] wrapping a new [FvProtocolData] instance.
+    ///
+    /// Uses TPL_CALLBACK to match SYSTEM_TABLE and allow both locks to be held
+    /// simultaneously without TPL ordering violations during boot orchestration.
     pub const fn new_locked() -> tpl_mutex::TplMutex<Self> {
-        tpl_mutex::TplMutex::new(efi::TPL_NOTIFY, Self::new(), "FvData")
+        tpl_mutex::TplMutex::new(efi::TPL_CALLBACK, Self::new(), "FvData")
```
joschock: I don't think you can lower this. Making this TPL_CALLBACK means that event handlers running at TPL_NOTIFY would assert (due to TPL inversion) if they invoked any routine that needs this lock. Per the PI spec several of the firmware volume APIs are permitted to be invoked at TPL_NOTIFY (see, for example: https://uefi.org/specs/PI/1.9/V3_Code_Definitions.html#efi-firmware-volume2-protocol-getvolumeattributes), and this change would make them crash.

patina_dxe_core/src/component_dispatcher.rs
```diff
@@ -120,9 +120,13 @@ unsafe impl Send for ComponentDispatcher {}
...
 impl ComponentDispatcher {
     /// Creates a new locked ComponentDispatcher.
+    ///
+    /// Uses TPL_CALLBACK to allow drivers connected during component dispatch
+    /// (e.g., via boot orchestration's ConnectController) to call
+    /// RaiseTPL(TPL_CALLBACK) without triggering TPL violations.
     #[inline(always)]
     pub(crate) const fn new_locked() -> TplMutex<Self> {
-        TplMutex::new(efi::TPL_NOTIFY, Self::new(), "ComponentDispatcher")
+        TplMutex::new(efi::TPL_CALLBACK, Self::new(), "ComponentDispatcher")
```
joschock: This one might be okay to lower. On a cursory search, I don't see attempts to lock this at TPL higher than TPL_APP. In fact, as designed today, this means that component entry points are invoked at TPL_NOTIFY which is a pretty high priority level for an entry point. You could make the argument that this should be even lower (i.e. `TplMutex::new(efi::TPL_APPLICATION` - I'm not sure you can go that low without some experimenting tho. 

@Javagedes to comment as well, as he is more familiar with the component arch.

patina_dxe_core/src/events.rs
```diff
@@ -215,13 +226,13 @@ pub extern "efiapi" fn set_timer(event: efi::Event, timer_type: efi::TimerDelay,
...
     assert!(new_tpl <= efi::TPL_HIGH_LEVEL, "Invalid attempt to raise TPL above TPL_HIGH_LEVEL");
 
+    // fetch_max keeps the higher value, so if current TPL > new_tpl, TPL stays unchanged.
+    // Per UEFI spec, behavior is "undefined" when new_tpl <= current TPL.
+    // We interpret this as a no-op: if already at a higher TPL, we're already protected
+    // from lower-TPL interruption, so there's nothing to do. This allows event notification
+    // callbacks (at TPL_NOTIFY) to call boot services that use locks at lower TPL levels.
     let prev_tpl = CURRENT_TPL.fetch_max(new_tpl, Ordering::SeqCst);
 
-    assert!(
```
joschock, line 220: This is a key assert. Please leave it in. 

"Raising" the TPL to a lower level is not allowed because it breaks invariants that the system relies on for correct behavior. For example - let's say that I have a TplMutex<> guarding some critical section:

```rust
let my_mutex = TplMutex::new<efi::TPL_CALLBACK, (), "My Callback-level TPL Lock">;
```

By picking the TPL_CALLBACK I'm expecting two things:
1) My code (that uses this guard) may be interrupted by TPL_NOTIFY (or higher) but that
2) Code running at TPL_NOTIFY or higher _will not interact with my critical section or data_. 

This implies that my critical section is effectively isolated at TPL_CALLBACK _and only at TPL_CALLBACK_. The lock can be acquired by code running at _lower_ TPL - because - by dint of being at the lower TPL than the Mutex, it can't possibly be in the critical section (if it was, the TPL would have been raised by the Mutex). But it cannot be acquired from _higher_ TPL because code running in higher TPL might have interrupted the critical section being protected. 

To say it a different way, I'm expecting that the TPL gives me mutual exclusion over the data/code that I'm protecting with TPL raises. If something running at a higher TPL level than my TPL interrupts me and is allowed to access the same data (by acquiring the same TPL lock, for example), then we have no actual mutual exclusion guarantee on the critical section, and this is all just window dressing.

The reason we have different TPL levels is to allow for higher-priority events to supersede lower priority events for system responsiveness. But the tradeoff for that to work without data races is that the data accessed by those higher-priority events cannot be also being accessed at lower priority in the interrupted task. 

This same exact invariant is enforced in the EDK2 core for the same reasons:

https://github.com/microsoft/mu_basecore/blob/6f0bf2def19b6f34c9b27824cdc3f30601a47827/MdeModulePkg/Core/Dxe/Event/Tpl.c#L65

patina_dxe_core/src/component_dispatcher.rs
joschock: This might be the main source of your TPL woes. I'm not sure. But what I see is that this lock is held by the component dispatcher while running the entry point of Components, which means that Component entry points are running at TPL_NOTIFY. There is a _whole bunch_ of things that are not permitted at that TPL level (see https://uefi.org/specs/UEFI/2.11/07_Services_Boot_Services.html#tpl-restrictions) - so that severely limits what boot services any component is allowed to interact with in its entry point.

patina_dxe_core/src/lib.rs
```diff
@@ -484,18 +484,32 @@ impl<P: PlatformInfo> Core<P> {
...
-            // Patina component dispatch
-            let dispatched = self.component_dispatcher.lock().dispatch();
 
-            // UEFI driver dispatch
-            let dispatched = dispatched
-                || self
-                    .pi_dispatcher
-                    .dispatch()
-                    .inspect_err(|err| log::error!("UEFI Driver Dispatch error: {err:?}"))?;
+        // First, do an initial component dispatch to install core protocols
```
joschock: I think this is unnecessary. The existing `loop` construct will continue to dispatch over and over again until all drivers and components are dispatched.

patina_dxe_core/src/lib.rs
```diff
@@ -539,7 +553,10 @@ impl<P: PlatformInfo> Core<P> {
 
         memory_attributes_table::init_memory_attributes_table_support();
 
-        self.component_dispatcher.lock().set_boot_services(StandardBootServices::new(st.boot_services()));
+        self.component_dispatcher.lock().set_boot_services(StandardBootServices::new_with_dxe_core_handle(
```
joschock: I don't think it is a good idea to expose the DXE core handle as part of the StandardBootServices interface. There are other approaches that I think make more sense:
1) Use existing BS APIs to retrieve it (e.g. locate handle buffer on LoadedImage, search for DXE core GUID).
2) update component entry point invocation to pass the image handle as an arg on the entrypoint (this is how trad UEFI drivers work).

patina_dxe_core/src/pi_dispatcher.rs
```diff
@@ -587,8 +587,10 @@ impl DispatcherContext {
         }
     }
 
+    /// Uses TPL_CALLBACK to match SYSTEM_TABLE and allow both locks to be held
+    /// simultaneously without TPL ordering violations during boot orchestration.
     const fn new_locked() -> TplMutex<Self> {
-        TplMutex::new(efi::TPL_NOTIFY, Self::new(), "Dispatcher Context")
+        TplMutex::new(efi::TPL_CALLBACK, Self::new(), "Dispatcher Context")
```
joschock: I think this will also cause problems - some of the primitives for dispatcher config could be invoked from TPL_NOTIFY.

patina_dxe_core/src/protocol_db.rs
```diff
@@ -598,8 +598,11 @@ impl Default for SpinLockedProtocolDb {
 
 impl SpinLockedProtocolDb {
     /// Creates a new instance of SpinLockedProtocolDb.
+    ///
+    /// Uses TPL_CALLBACK to match SYSTEM_TABLE and allow both locks to be held
+    /// simultaneously without TPL ordering violations during boot orchestration.
     pub const fn new() -> Self {
-        SpinLockedProtocolDb { inner: tpl_mutex::TplMutex::new(efi::TPL_NOTIFY, ProtocolDb::new(), "ProtocolLock") }
+        SpinLockedProtocolDb { inner: tpl_mutex::TplMutex::new(efi::TPL_CALLBACK, ProtocolDb::new(), "ProtocolLock") }
```
joschock: Many of the protocol APIs are permitted at TPL_NOTIFY - so this must stay at least that high.

patina_dxe_core/src/protocols.rs
```diff
@@ -504,9 +504,11 @@ unsafe extern "C" fn install_multiple_protocol_interfaces(handle: *mut efi::Hand
...
     // after all protocols are installed. Despite the spec ambiguity, the reference EDK2 C implementation does raise to
     // TPL_NOTIFY prior to installing any of the interfaces, which has the effect of deferring any protocol notify
-    // callbacks until after all protocols are installed. This code matches those semantics by using a TPL guard here
-    // to ensure the logic of this function is conducted at TPL_NOTIFY.
-    let tpl_mutex = TplMutex::new(efi::TPL_NOTIFY, (), "atomic_protocol_install");
+    // callbacks until after all protocols are installed.
+    //
+    // However, we use TPL_CALLBACK here to match SYSTEM_TABLE and allow protocol notify callbacks to access
+    // boot services without TPL ordering violations during boot orchestration.
+    let tpl_mutex = TplMutex::new(efi::TPL_CALLBACK, (), "atomic_protocol_install");
```
joschock: I think this should also stay at TPL_NOTIFY, otherwise the atomic "all installed or none installed" behavior expected from install_multiple_protocol_interfaces would break.

patina_dxe_core/src/systemtables.rs
```diff
@@ -16,8 +16,12 @@ use r_efi::efi;
 
 use crate::{allocator::EFI_RUNTIME_SERVICES_DATA_ALLOCATOR, tpl_mutex};
 
+/// Global system table protected by a TPL mutex.
+///
+/// Uses TPL_CALLBACK instead of TPL_NOTIFY to allow drivers connected during
+/// boot orchestration to call RaiseTPL(TPL_CALLBACK) without TPL violations.
 pub static SYSTEM_TABLE: tpl_mutex::TplMutex<Option<EfiSystemTable>> =
-    tpl_mutex::TplMutex::new(efi::TPL_NOTIFY, None, "StLock");
+    tpl_mutex::TplMutex::new(efi::TPL_CALLBACK, None, "StLock");
```
joschock: System table needs to be protected at TPL_NOTIFY, since APIs that access it (e.g. install configuration table) are permitted at that callback level.

joschock (changes requested):
I don't think this can go in until the TPL issues are addressed.  I think the main issue here is that the component entry points are invoked at TPL_NOTIFY, which is too high. If component entry points were invoked at TPL_APPLICATION, then I think none of the other TPL changes would be needed.

I also think that component entry points should be given a valid image handle (dxe core or otherwise) to use for things like load_image calls - I think exposing it via StandardBootServices is not the right approach.

components/patina_boot/src/config.rs
kat-perez: yeah, I was going to change it in the next iteration but I can just change it now since that's an easy update

components/patina_boot/src/helpers.rs
kat-perez: done. renamed it to `connect_all`

kat-perez:
@joschock Thanks for the feedback. This PR is ready for re-review when you have a chance. Changes summarized below:

TPL:
- Lowered `ComponentDispatcher` TPL from `TPL_NOTIFY` to `TPL_APPLICATION` so component entry points now run at the appropriate level
- Reverted all other TPL changes (protocol_db, protocols, systemtables)
- Kept the assert in `raise_tpl` that prevents raising to a lower TPL level

ImageHandle:
- Added `ImageHandle` param type that components can request as a dependency
- `BootOrchestrator` now takes `Option<ImageHandle>` and uses it as the parent handle for `LoadImage` calls
- Falls back to `LocateHandleBuffer` if not provided (for now, until DXE core image handle creation is implemented)

BootOptions:
- Changed from fixed primary/secondary slots to `Vec<DevicePathBuf>` with `with_device()` builder pattern

patina_dxe_core/src/lib.rs
kat-perez: reverted

patina_dxe_core/src/events.rs
kat-perez: Thanks for the detailed explanation. I've kept the assert in place

patina_dxe_core/src/component_dispatcher.rs
```diff
@@ -120,9 +120,12 @@ unsafe impl Send for ComponentDispatcher {}
 
 impl ComponentDispatcher {
     /// Creates a new locked ComponentDispatcher.
+    ///
+    /// Uses TPL_APPLICATION so that component entry points can use boot services
+    /// that are restricted at higher TPL levels (per UEFI spec Section 7.1).
     #[inline(always)]
     pub(crate) const fn new_locked() -> TplMutex<Self> {
-        TplMutex::new(efi::TPL_NOTIFY, Self::new(), "ComponentDispatcher")
+        TplMutex::new(efi::TPL_APPLICATION, Self::new(), "ComponentDispatcher")
```
joschock, line 128: I think TplMutex at TPL_APPLICATION is effectively a spin::mutex with the added invariant that you _must_ be at TPL_APPLICATION to acquire it. That's not a bad thing, IMHO, but it is a bit strange to `RaiseTpl(TPL_APPLICATION)`. No action needed unless this bothers anyone else tho :).

Javagedes:
> I don't think this can go in until the TPL issues are addressed. I think the main issue here is that the component entry points are invoked at TPL_NOTIFY, which is too high. If component entry points were invoked at TPL_APPLICATION, then I think none of the other TPL changes would be needed.
> 
> I also think that component entry points should be given a valid image handle (dxe core or otherwise) to use for things like load_image calls - I think exposing it via StandardBootServices is not the right approach.

I will add a new parameter type that can be requested in the patina component interface that will give you a valid handle. I have no problem with changing what TPL patina components are executed at.

joschock:
I am good with the changes from the TPL/core side. Will defer to @Javagedes  on the component/handle interface stuff.

Javagedes:
> > I don't think this can go in until the TPL issues are addressed. I think the main issue here is that the component entry points are invoked at TPL_NOTIFY, which is too high. If component entry points were invoked at TPL_APPLICATION, then I think none of the other TPL changes would be needed.
> > I also think that component entry points should be given a valid image handle (dxe core or otherwise) to use for things like load_image calls - I think exposing it via StandardBootServices is not the right approach.
> 
> I will add a new parameter type that can be requested in the patina component interface that will give you a valid handle. I have no problem with changing what TPL patina components are executed at.

#1244 for an initial implementation open for thoughts.

components/patina_boot/src/component/boot_orchestrator.rs
```diff
@@ -0,0 +1,104 @@
...
+    /// 6. If all boot options fail, call failure handler
+    #[coverage(off)] // Component integration - tested via integration tests
+    fn entry_point(
+        self,
+        boot_services: StandardBootServices,
+        runtime_services: StandardRuntimeServices,
+        boot_options: Config<BootOptions>,
+        image_handle: Option<ImageHandle>,
+    ) -> Result<()> {
+        helpers::connect_all(boot_services.as_ref())?;
```
os-d, line 61: There's a lot of debate about whether to do connect all or not and for many platforms this is not the right answer. It may be reasonable to have a simple component that does this, it may also need to be configurable.

I don't have enough context on this @kuqin12 can you comment?

components/patina_boot/src/component/boot_orchestrator.rs
```diff
@@ -0,0 +1,104 @@
...
+        helpers::signal_bds_phase_entry(boot_services.as_ref())?;
+        helpers::discover_console_devices(boot_services.as_ref(), runtime_services.as_ref())?;
+        helpers::signal_ready_to_boot(boot_services.as_ref())?;
+
+        // Use the provided image handle, or fall back to finding one via LocateHandleBuffer.
+        // Per UEFI spec, the parent handle must be a valid image handle (has LoadedImage protocol).
```
os-d, line 67: If the UEFI spec says this must be valid, should it even be an option or rather an ImageHandle and if invalid error?

components/patina_boot/src/component/boot_orchestrator.rs
```diff
@@ -0,0 +1,104 @@
...
+                EfiError::NotReady
+            })?
+        };
+
+        for device_path in boot_options.devices() {
+            match helpers::boot_from_device_path(boot_services.as_ref(), parent_handle, device_path) {
+                Ok(()) => {
+                    // Boot image returned control (e.g., EFI application exited).
+                    // Continue to try next boot option.
+                    log::warn!("Boot option returned, trying next...");
```
os-d, line 90: here and below name which boot option?

components/patina_boot/src/config.rs
```diff
@@ -0,0 +1,174 @@
...
+        self.devices.push(device);
+        self
+    }
+
+    /// Add a hotkey scancode for boot override (not yet implemented).
+    ///
+    /// This field is reserved for future boot menu functionality.
+    /// Currently, the hotkey is stored but not acted upon.
+    pub fn with_hotkey(mut self, scancode: u16) -> Self {
+        self.hotkey = Some(scancode);
```
os-d, line 61: Maybe make this a `todo!()` instead so we don't forget and it doesn't accidentally get used thinking it does something (yes this is just to the feature branch, but still).

components/patina_boot/src/config.rs
```diff
@@ -0,0 +1,174 @@
...
+impl BootOptions {
+    /// Create new empty boot options.
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Add a boot device path.
+    ///
+    /// Device paths are tried in the order they are added.
+    pub fn with_device(mut self, device: DevicePathBuf) -> Self {
```
os-d, line 51: Would this be a hardcoded list of device paths? Do platforms not need a dynamic way of adding device paths?

components/patina_boot/src/helpers.rs
```diff
@@ -0,0 +1,430 @@
...
+///
+/// Returns `Ok(())` if the boot image was successfully started (which typically
+/// means it returned control). Returns an error if loading or starting fails.
+pub fn boot_from_device_path<B: BootServices>(
+    boot_services: &B,
+    parent_handle: efi::Handle,
+    device_path: &DevicePathBuf,
+) -> Result<()> {
+    // Load the image
+    let device_path_ptr = device_path.as_ref() as *const _ as *mut efi::protocols::device_path::Protocol;
```
os-d, line 51: I think this is UB where you have an immutable reference and are casting it as a mutable pointer. I think you need to do that from a mutable reference.

components/patina_boot/src/helpers.rs
```diff
@@ -0,0 +1,430 @@
...
+            break;
+        }
+
+        prev_handle_count = current_handle_count;
+    }
+
+    Ok(())
+}
+
+/// Signal EndOfDxe event for platforms implementing custom orchestration.
```
os-d, line 117: nit: this is more than just for platforms implementing custom orchestration, this has implications for SMM locking, etc. Same for other event signaling comments below

components/patina_boot/src/helpers.rs
```diff
@@ -0,0 +1,430 @@
...
+/// not yet write the `ConIn`, `ConOut`, and `ErrOut` UEFI variables.
+///
+/// Platforms requiring full console variable support should implement this
+/// functionality or use platform-specific console initialization.
+///
+/// # Arguments
+///
+/// * `boot_services` - Boot services interface
+/// * `runtime_services` - Runtime services interface (unused in stub)
+#[allow(unused_variables)]
```
os-d, line 194: nit: easy to forget these when many iterations occur, I would just do let _ in the fn below and drop this

patina_dxe_core/src/component_dispatcher.rs
os-d: @Javagedes I assume you wrote this part, any concern with changing TPL level here?

sdk/patina/src/component/params.rs
```diff
@@ -711,6 +711,87 @@ unsafe impl Param for StandardRuntimeServices {
...
+///             device_path,
+///             None,
+///             0,
+///         )?;
+///         Ok(())
+///     }
+/// }
+/// ```
+#[derive(Debug, Clone, Copy)]
+pub struct ImageHandle(r_efi::efi::Handle);
```
os-d: @Javagedes any comments on the new param? I think you are adding a Handle param as well, for this PR?

Javagedes: Yes. A few. Mainly it is that I have a PR open that I'm working on, which she is aware of as the thumbs upped my comment about it above :) (#1244). However generically - 

1. In my opinion we want each patina component to have its own handle, not share a handle as is implemented here
2. We want the handle to be more generic then specifically an EfiHandle. I can see a use case later down the road when every component is given a handle, which can be used as a handle for working with EFI system tables, and for working with patina "things"
3. I don't see (though I may have missed it), `set_handle(...)` ever being called
4. `mock` probably needs the ability to specify a handle value, not just always being null.

patina_dxe_core/src/component_dispatcher.rs
Javagedes: I do not personally have an issue with this. The other option is to put this behind a non-TPL aware lock though.

sdk/patina/src/component/params.rs
Javagedes: After talking to a few other people, I think we have come to the conclusion that to keep EFI and EFI drivers separated from patina components, that it is actually best to use the DXE Core handle. But think I'm going to go at this from a different angle. I'll update my PR and we can go from there.

components/patina_boot/src/component/boot_orchestrator.rs
kat-perez: Added boot option index to the log messages. Now logs "Boot option {} returned, trying next..." and "Boot option {} failed trying next..." with the index

components/patina_boot/src/config.rs
kat-perez: Done. `with_hotkey()` now panics with `todo!("Hotkey boot override not yet implemented")`

components/patina_boot/src/helpers.rs
kat-perez: Added documentation explaining the FFI safety. Per UEFI spec, LoadImage's DevicePath parameter is IN-only and not mutated. The cast to *mut is required for r-efi compatibility but the data is never modified. Used explicit `.cast_mut()` to make the intent clear

components/patina_boot/src/helpers.rs
kat-perez: Update the doc comments for `signal_bds_phase_entry` and `signal_ready_to_boot` to cover the broader implications including SMM/MM lockdown, security components, and firmware volume finalization.

components/patina_boot/src/helpers.rs
kat-perez: Simplified to use `let _ = boot_services.close_event(event);` pattern, removing the redundant comment

components/patina_boot/src/component/boot_orchestrator.rs
kat-perez: Changed `image_handle` from `Option<ImageHandle>` to required `ImageHandle`. Per UEFI Spec Section 7.4, `LoadImage()` requires `ParentImageHandle` to be non-NULL (returns `EFI_INVALID_PARAMETER` otherwise). Making it required enforces this at the type level.

components/patina_boot/src/component/boot_orchestrator.rs
kuqin12: For posterity, we would not want to do connect all for the boot flow in general. Especially not before the EndOfDxe. End of DXE is treated as the security boundary by a lot of platforms and connecting all devices will involve indeterministic software stack to be involved (USB, network, etc.) and bring in vulnerabilities to the system. Also, it would make the system boot more slowly compared to only connecting the disk.

So it would be better to limit the amount of controller connected and defer most of the connection to after console is up.

components/patina_boot/README.md
```diff
@@ -0,0 +1,33 @@
+# Patina Boot
```
makubacki, line 1: Can you consider more precise names than "Patina Boot"? I understand how it would have been chosen but it's also generic enough to point of being confusing.

components/patina_boot/src/component/boot_orchestrator.rs
makubacki: Adding on to @kuqin12's response, I'd resist from adding connect all to this design at all and instead provide guidance documentation on how to more meaningfully connect devices.

sdk/patina/src/uefi_protocol/device_path/nodes.rs
makubacki, line 1: @kat-perez, it seems like all of these device path enhancements could go into `main` independently of this PR? Is there a reason to keep them here?

sdk/patina/src/component/params.rs
makubacki: Thanks for the update. I'm also concerned about handles representing EFI drivers and Patina components. If the change remains at that level of scope, while relatively small, I think an RFC would be helpful to talk about and agree upon the design.

patina_dxe_core/src/component_dispatcher.rs
makubacki: @Javagedes, what was the reason(s) for originally selecting `TPL_NOTIFY`?

sdk/patina/src/component/params.rs
```diff
@@ -816,12 +897,15 @@ mod tests {
     }
 
     #[test]
-    fn test_config_can_be_accessed_while_unlocked() {
+    fn test_config_can_be_accessed_when_locked() {
```
makubacki, line 896: Do these updates to component tests need to be coupled into the larger boot orchestration PR?

components/patina_boot/src/component/boot_orchestrator.rs
kat-perez: @os-d @kuqin12 @makubacki I agree with you all here that it's the most inefficient and least secure method, so i'll be removing connect_all() from the default boot flow. I think having the simple component design with configuration makes sense. I have a proposed design for a `ConnectConfig` and a `DeviceConnector` component. But since this is going into the feature branch and I'm just trying to get an initial boot crate in, I was planning on having a separate PR for further discussion that's more focused on those components instead of grouping it with this PR at the moment.

Will be removing it in the next one when the design for a `DeviceConnector` component is introduced and `connect_all` is removed for that reason because the better design is preferred. How does that sound?

sdk/patina/src/component/params.rs
kat-perez: For this PR, ImageHandle is scoped narrowly to what LoadImage() needs - a valid parent image handle. Happy to align with whatever comes out of #1244 / the RFC discussion. Should I remove this and wait for #1244, or keep it as a temporary solution?

patina_dxe_core/src/component_dispatcher.rs
Javagedes: I was mirroring the TPL level used for the TplMutex around the UEFI driver dispatcher.

https://github.com/OpenDevicePartnership/patina/blob/19b22bff6aea395888452caabac6623af803bbda/patina_dxe_core/src/image.rs#L662-L663

In hindsight its not the same however. We release the lock before executing a UEFI driver. That is why I also suggested we don't use a Tpl aware mutex.

sdk/patina/src/component/params.rs
Javagedes: After more thought and discussion, I think this implementation works for me. I would ask a few minor nitpicks. I will leave them in the PR itself.

Javagedes: Please rename this to `Handle` and make it a struct instead of a tuple struct. I have future plans to re-use this as a generic handle ID in other contexts, while keeping the ability to use this as a handle in the UEFI sense.

sdk/patina/src/component/params.rs
```diff
@@ -711,6 +711,87 @@ unsafe impl Param for StandardRuntimeServices {
...
+    }
+
+    /// Creates a mock ImageHandle for testing purposes.
+    ///
+    /// # Safety
+    ///
+    /// The returned handle is a null pointer and should only be used in tests
+    /// where the handle is not actually dereferenced.
+    #[cfg(test)]
+    pub fn mock() -> Self {
```
Javagedes: Please update `mock` to allow passing in a valid handle. Trying to do any real mocking within a UEFI context will probably not work with a invalid handle. Someone really wanting to test with it, will need a valid handle.

sdk/patina/src/component/params.rs
```diff
@@ -711,6 +711,87 @@ unsafe impl Param for StandardRuntimeServices {
...
+///         Ok(())
+///     }
+/// }
+/// ```
+#[derive(Debug, Clone, Copy)]
+pub struct ImageHandle(r_efi::efi::Handle);
+
+impl ImageHandle {
+    /// Creates a new ImageHandle wrapping the given EFI handle.
+    pub fn new(handle: r_efi::efi::Handle) -> Self {
```
Javagedes: Please delete this, we don't want people to be able to create their own handle manually. We only want it to be create-able from the `Param` implementation (other then `mock` which is used for testing)

sdk/patina/src/component/params.rs
```diff
@@ -711,6 +711,87 @@ unsafe impl Param for StandardRuntimeServices {
...
+        Self(handle)
+    }
+
+    /// Returns the underlying EFI handle.
+    pub fn get(&self) -> r_efi::efi::Handle {
```
Javagedes: Instead of this, please implement the Deref trait to deref to a efi::Handle. I plan on this type being used for more then just a handle for images in the future.

components/patina_boot/src/config.rs
kat-perez: The with_device() builder supports both static and dynamic paths - platforms construct DevicePathBuf at runtime however they need (hardcoded topology, UEFI variables, hardware probing, etc.). For example:
```
for path in platform_discover_boot_devices() {
    options = options.with_device(path);
}
```
The config takes explicit paths to keep the orchestrator predictable. Automatic discovery (e.g., 'find all bootable NVMe namespaces') would be a separate helper or component to be implemented in another PR

components/patina_boot/README.md
kat-perez: If we're thinking about the broader BDS architecture here, we could possibly rename this crate to `patina_boot_orchestration` or `patina_bds_orchestration`? Alternate option is to add a `patina_boot` or `patina_bds` umbrella crate later that re-exports from the specific crates. Thoughts?

sdk/patina/src/uefi_protocol/device_path/nodes.rs
kat-perez: Split out the device path enhancements to #1258. Once that merges to main, I'll rebase this PR to remove the duplicate.

sdk/patina/src/component/params.rs
kat-perez: Good catch - yeah, these can be separated out. I'll open a new PR for this

kat-perez: #1259 created to address this

sdk/patina/src/component/params.rs
kat-perez: Done - renamed to `Handle` with a named field struct

sdk/patina/src/component/params.rs
kat-perez: Done - mock() now takes a handle parameter

sdk/patina/src/component/params.rs
kat-perez: Done - removed new() entirely. Handle can only be created via the Param implementation or mock() for testing.

sdk/patina/src/component/params.rs
kat-perez: Done - implemented Deref<Target = r_efi::efi::Handle>

sdk/patina/src/component/params.rs
```diff
@@ -711,6 +711,82 @@ unsafe impl Param for StandardRuntimeServices {
...
+    type State = ();
+    type Item<'storage, 'state> = Self;
+
+    unsafe fn get_param<'state>(
+        _state: &'state Self::State,
+        storage: UnsafeStorageCell<'_>,
+    ) -> Self::Item<'static, 'state> {
+        // SAFETY: Image handle is immutably borrowed from storage.
+        // validate() ensures the handle is set before get_param is called.
+        Handle { handle: unsafe { storage.storage().image_handle().unwrap() } }
```
Javagedes: nit: should use expect*"reasoning") here instead of an unwrap, which is basically exactly what the above comment is :)

sdk/patina/src/component/storage.rs
```diff
@@ -277,25 +280,44 @@ impl Storage {
...
 
     /// Adds a config datum to the storage, overwriting an existing value if it exists.
+    ///
+    /// Configs are created unlocked so that components with `ConfigMut<T>` can modify them
+    /// during the first dispatch phase. Call `lock_configs()` to lock all configs before
+    /// dispatching components that require `Config<T>` (immutable access).
     pub fn add_config<C: Default + 'static>(&mut self, config: C) {
         let id = self.register_config::<C>();
-        self.configs.insert(id, RefCell::new(ConfigRaw::new(true, Box::new(config))));
+        self.configs.insert(id, RefCell::new(ConfigRaw::new(false, Box::new(config))));
```
Javagedes, line 320: This needs to be true or pure rust components will not execute until the end of dispatch.

I'm guessing you do this so your component needs to execute last, but this is also causing all components to execute last. You need to find another way to make your component execute last.

sdk/patina/src/component/storage.rs
```diff
@@ -277,25 +280,44 @@ impl Storage {
...
     }
 
     /// Adds a default valued config datum to the storage if it does not exist.
+    ///
+    /// Default configs are created unlocked so that `ConfigMut<T>` can modify them.
     pub(crate) fn add_config_default_if_not_present<C: Default + 'static>(&mut self) -> usize {
         let idx = self.register_config::<C>();
         if !self.configs.contains(idx) {
-            self.configs.insert(idx, RefCell::new(ConfigRaw::new(true, Box::<C>::default())));
+            self.configs.insert(idx, RefCell::new(ConfigRaw::new(false, Box::<C>::default())));
```
Javagedes, line 308: same here. This is a fairly drastic change in the execution flow of components, moving all components with Config Dependencies to being one of the last things to execute. You need to find a different way to make your boot orchestrator component boot last.

sdk/patina/src/component/storage.rs
Javagedes: If you want to keep this for now, to get this PR merged in as is, since it is going to a feature branch, I'm cool with that. But you will need a different solution by the time we merge it into main.

sdk/patina/src/component/storage.rs
Javagedes: If you want to keep this for now, to get this PR merged in as is, since it is going to a feature branch, I'm cool with that. But you will need a different solution by the time we merge it into main.

sdk/patina/src/component/params.rs
kat-perez: Done - changed to .expect("image_handle validated as Some in validate()")

sdk/patina/src/component/storage.rs
kat-perez: Understood - this change was to allow the boot orchestrator to receive Config<BootOptions> after configs are locked. I'll keep it for the feature branch and find a better solution (e.g., component priority/ordering, explicit dependencies) before merging to main

sdk/patina/src/component/storage.rs
kat-perez: (response same as above)

sdk/patina/src/component/storage.rs
Javagedes: sounds good to me. will close these

makubacki (approved):
Approving for this to get into the feature branch.

PR #1227, opened 2026-01-08
patina_debugger: Add module count monitor command

cfernald:
## Description

UefiExt currently has to parse the in memory debug table structure to determine if there are new modules that are available since the last check. This adds a simple monitor command that can be used to indicate if a reload is required.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on Q35

## Integration Instructions

N/A

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,7 +4,7 @@ resolver = "2"
 members = ["components/*", "core/*", "sdk/*", "patina_dxe_core"]
 
 [workspace.package]
-version = "19.0.1"
+version = "19.0.2"
 license = "Apache-2.0"
 edition = "2024"
 rust-version = "1.89"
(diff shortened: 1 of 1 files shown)


cfernald:
> There is an edge case here, of course, where a module was loaded when we were last broken in, we continue, the module gets unloaded, we load a new module and break, count is the same. This does happen with real platforms (modules getting unloaded), but since the user can always just rerun !findall, it's not that big of a deal. If you cared, you could run a checksum of the module names or something. Although I haven't checked if we remove the module from state.modules when something gets unloaded, if we don't then this wouldn't be an issue.

@os-d, true, however not a problem here since we don't currently track unloads anyways. Generally, i'm thinking about dropping the module list in the debugger since it is redundant with the standard UEFI debug table. I haven't fully decided, but if so, I will just make this a monotonic counter for the number of module load calls, which should be indifferent to the load/unload issue.

os-d:
> Approved - but also, the debug table header has status flags that allow you know whether it's been modified since last you looked. If the problem statement is "I have to parse the whole list every time," you could consult those flags to determine whether a whole table parse is needed.

@cfernald I would probably just change UefiExt to read and clear the modified flag. We completely ignore them today. Probably simpler than adding this monitor cmd. It's a little more complex because you could be broken in in the middle of modifying the table, but there's a flag for that, too, so not that bad.

cfernald:
> Approved - but also, the debug table header has status flags that allow you know whether it's been modified since last you looked. If the problem statement is "I have to parse the whole list every time," you could consult those flags to determine whether a whole table parse is needed.

@joschock, true, though I'm hesitant to mess with debugee memory automatically like that to clear the flag for that to be useful. Though perhaps i'm being paranoid.

os-d:
> > Approved - but also, the debug table header has status flags that allow you know whether it's been modified since last you looked. If the problem statement is "I have to parse the whole list every time," you could consult those flags to determine whether a whole table parse is needed.
> 
> @joschock, true, though I'm hesitant to mess with debugee memory automatically like that to clear the flag for that to be useful. Though perhaps i'm being paranoid.

I mean, that's what the UEFI spec envisions:

image

os-d:
And, that's exactly why the spec defines (and so we do) this to be a volatile field, we expect that the value may change

cfernald:
> > Approved - but also, the debug table header has status flags that allow you know whether it's been modified since last you looked. If the problem statement is "I have to parse the whole list every time," you could consult those flags to determine whether a whole table parse is needed.
> 
> @cfernald I would probably just change UefiExt to read and clear the modified flag. We completely ignore them today. Probably simpler than adding this monitor cmd. It's a little more complex because you could be broken in in the middle of modifying the table, but there's a flag for that, too, so not that bad.

This is fair enough, and definitely more standard.

cfernald:
> And, that's exactly why the spec defines (and so we do) this to be a volatile field, we expect that the value may change

My paranoia is more around the fact that this assumes a single consumer. So if there is some software that is currently using this, unknown to us, then this would be a compatibility issue.

makubacki:
> My paranoia is more around the fact that this assumes a single consumer. So if there is some software that is currently using this, unknown to us, then this would be a compatibility issue.

My understanding is also that this is an expected use case. @cfernald, what is the scenario that you're concerned about?

cfernald:
No, I think i'm being overly paranoid. I will go that route instead. Thanks eveyone!

As a side note, I will probably just remove the `mod list` monitor command. I don't think anyone is using it anymore and is redundant with the debug table. @joschock, let me know if any of your LLDB scripts were using that, but I assume not.

joschock:
@cfernald I do use the list monitor command - I find it convenient in LLDB and I have built a bit of primitive scripting around it as well as invoking it manually. I agree it could be replaced with scripts to parse the debug table, but those don't exist yet on the LLDB side.

PR #1232, opened 2026-01-09
[patina_dxe_core] Add Missing Safety Comments

Raymond-MS:
## Description

Added missing Safety comments to the following files in patina_dxe_core:

allocator.rs
config_tables.rs
cpu.rs
debugger_reload.rs
driver_services.rs

Relates to: https://github.com/OpenDevicePartnership/patina/issues/583

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Built SBSA successfully in patina-dxe-core-qemu.

## Integration Instructions

N/A

### DIFF
--- a/patina_dxe_core/src/driver_services.rs
+++ b/patina_dxe_core/src/driver_services.rs
@@ -47,6 +47,7 @@ fn get_platform_driver_override_bindings(
     {
         Err(_) => return Vec::new(),
         Ok(protocol) => unsafe {
+            // SAFETY: locate_protocol guarantees that if `Ok` is returned, a valid pointer is encapsulated in it.
             (protocol as *mut efi::protocols::platform_driver_override::Protocol).as_mut().expect("bad protocol ptr")
         },
     };
@@ -80,6 +81,7 @@ fn get_family_override_bindings() -> Vec<*mut efi::protocols::driver_binding::Pr
     for handle in driver_binding_handles {
         match PROTOCOL_DB.get_interface_for_handle(handle, efi::protocols::driver_family_override::PROTOCOL_GUID) {
             Ok(protocol) => {
+                // SAFETY: get_interface_for_handle guarantees that if `Ok` is returned, a valid pointer is encapsulated in it.
                 let driver_override_protocol = unsafe {
                     (protocol as *mut efi::protocols::driver_family_override::Protocol)
                         .as_mut()
@@ -103,6 +105,7 @@ fn get_bus_specific_override_bindings(
         .get_interface_for_handle(controller_handle, efi::protocols::bus_specific_driver_override::PROTOCOL_GUID)
     {
         Err(_) => return Vec::new(),
+        // SAFETY: get_interface_for_handle guarantees that if `Ok` is returned, a valid pointer is encapsulated in it.
         Ok(protocol) => unsafe {
             (protocol as *mut efi::protocols::bus_specific_driver_override::Protocol)
                 .as_mut()
@@ -133,6 +136,9 @@ fn get_all_driver_bindings() -> Vec<*mut efi::protocols::driver_binding::Protoco
         Ok(handles) => get_bindings_for_handles(handles),
     };
 
+    // SAFETY: driver_bindings must contain valid pointers/handles (a & b as well as *a & *b)
+    // when sort_unstable_by() is called. If .locate_handles() returns 'Ok' and handles is
+    // not empty, we rely on get_bindings_for_handles to return a valid Vec.
     driver_bindings.sort_unstable_by(|a, b| unsafe { (*(*b)).version.cmp(&(*(*a)).version) });
 
     driver_bindings
@@ -162,6 +168,7 @@ fn authenticate_connect(
             };
 
             if let Ok(mut file_path) = file_path {
+                // SAFETY: Pointer is validated using .expect(), will panic if .as_ref() returns a NULL pointer
                 let security2 = unsafe {
                     (security2_ptr as *mut patina::pi::protocols::security2::Protocol)
                         .as_ref()
@@ -222,6 +229,8 @@ fn core_connect_single_controller(
     loop {
         let mut started_drivers = Vec::new();
         for driver_binding_interface in driver_candidates.clone() {
+            // SAFETY: driver_binding_interface is a clone of driver_candidates which is created above.
+            // The pointer should be valid as long as driver_candidates is successfully allocated.
             let driver_binding = unsafe { &mut *(driver_binding_interface) };
             let device_path = remaining_device_path.or(Some(core::ptr::null_mut())).expect("must be some");
 
@@ -323,7 +332,9 @@ pub unsafe fn core_connect_controller(
 
     if recursive {
         for child in PROTOCOL_DB.get_child_handles(handle) {
-            //ignore the return value to match behavior of edk2 reference.
+            // ignore the return value to match behavior of edk2 reference.
+            // SAFETY: We do not remove any of these handles during this call, though it is
+            // possible for a different entity at another TPL to do so.
             _ = unsafe { core_connect_controller(child, Vec::new(), None, true) };
         }
     }
@@ -491,6 +502,8 @@ pub unsafe fn core_disconnect_controller(
             .get_interface_for_handle(driver_handle, efi::protocols::driver_binding::PROTOCOL_GUID)
             .or(Err(EfiError::InvalidParameter))?;
         let driver_binding_interface = driver_binding_interface as *mut efi::protocols::driver_binding::Protocol;
+        // SAFETY: driver_binding_interface is validated above, would have been caught with the .or and ? above with
+        // the loop being terminated.
         let driver_binding = unsafe { &mut *(driver_binding_interface) };
 
         let mut status = efi::Status::SUCCESS;
@@ -519,8 +532,17 @@ extern "efiapi" fn disconnect_controller(
     driver_image_handle: efi::Handle,
     child_handle: efi::Handle,
 ) -> efi::Status {
+    if controller_handle.is_null() {
+        return efi::Status::INVALID_PARAMETER;
+    }
+
     let driver_image_handle = NonNull::new(driver_image_handle).map(|x| x.as_ptr());
     let child_handle = NonNull::new(child_handle).map(|x| x.as_ptr());
+    // SAFETY: Caller must ensure controller_handle is valid for the duration of the call.
+    // driver_image_handle and child_handle are both created above using NonNull which should
+    // guarantee they are non-null pointers. controller_handle is validated above. We do not
+    // remove any of these handles during this call, though it is possible for a different
+    // entity at another TPL to do so.
     unsafe {
         match core_disconnect_controller(controller_handle, driver_image_handle, child_handle) {
             Err(err) => err.into(),
@@ -554,6 +576,8 @@ mod tests {
     fn with_locked_state<F: Fn() + std::panic::RefUnwindSafe>(f: F) {
         test_support::with_global_lock(|| {
             test_support::init_test_logger();
+            // SAFETY: init_test_protocol_db modifies global state. It is being called within a
+            // lock to have exclusive mutable access to the protocol database.
             unsafe {
                 test_support::init_test_protocol_db();
             }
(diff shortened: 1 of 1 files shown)


patina_dxe_core/src/allocator.rs
```diff
@@ -608,7 +610,7 @@ pub fn core_allocate_pool(pool_type: efi::MemoryType, size: usize) -> Result<*mu
     match ALLOCATORS.lock().get_or_create_allocator(pool_type, handle) {
         Ok(allocator) => {
             let mut buffer: *mut c_void = core::ptr::null_mut();
-
+            // Safety: caller must ensure that buffer is a valid pointer. It is created above.
```
os-d:
```suggestion
            // Safety: buffer is a stack variable we are passing the address of, so we know it is a valid pointer.
```
This isn't the caller's responsibility, since we have it on the stack (I know you meant caller of allocator.allocate_pool, but generally in safety comments caller means the caller of this function).

patina_dxe_core/src/allocator.rs
```diff
@@ -627,6 +629,7 @@ pub fn core_free_pool(buffer: *mut c_void) -> Result<(), EfiError> {
         return Err(EfiError::InvalidParameter);
     }
     let allocators = ALLOCATORS.lock();
+    // Safety: caller must ensure that buffer is a valid pointer. It is null-checked above.
```
os-d: This one I think we should bubble up the unsafe to the function (and the safety comment with caller expectations) because the caller must provide a buffer that points to a valid buffer allocated with core_allocate_pool (or in general the pool allocator). The lower level code takes a hard assumption on that and will dereference the pointer and read from it, which is only safe if that invariant is met.

patina_dxe_core/src/allocator.rs
```diff
@@ -746,6 +749,7 @@ pub fn core_free_pages(memory: efi::PhysicalAddress, pages: usize) -> Result<(),
 
     let mut memory_type = efi::CONVENTIONAL_MEMORY;
 
+    // Safety: caller must ensure that memory is a valid address. It is validated above.
```
os-d: I actually think there is an argument that all levels of `free_pages` shouldn't be unsafe functions because the allocators actually do have a way of verifying whether they own this memory or not (well, so do the pool functions, but they don't check today, whereas we already check in the pages functions: https://github.com/OpenDevicePartnership/patina/blob/1b8e961dc7d8b3977d321f874711ec3fea6f3f79/patina_dxe_core/src/allocator/fixed_size_block_allocator.rs#L607).

So the caller doesn't have an expectation to actually pass the right thing, i.e. memory safety I don't think will be violated if the caller passes garbage. I guess if they pass an address they don't own that the allocator does still own, then it could cause problems. So...eh, unsafe can stay in the function. In that case, same here, I would bubble unsafe up to the function level.

patina_dxe_core/src/allocator.rs
```diff
@@ -817,6 +821,7 @@ extern "efiapi" fn get_memory_map(
 
     let required_map_size = GCD.memory_descriptor_count_for_efi_memory_map() * mem::size_of::<efi::MemoryDescriptor>();
     assert_ne!(required_map_size, 0);
+    // Safety: caller must ensure that memory_map_size is a valid pointer. It is null-checked above.
```
os-d: efiapi in general I think can be marked as unsafe fns, probably the same arg here that because this is a caller req, move this to be an unsafe fn and put caller expectations at the top (there are others, essentially all the input args).

patina_dxe_core/src/config_tables.rs
```diff
@@ -57,6 +57,7 @@ pub fn core_install_configuration_table(
         assert_eq!(system_table.number_of_table_entries, 0);
         None
     } else {
+        // Safety: caller must ensure that the pointer is valid. It is null-checked above.
```
os-d:
```suggestion
        // Safety: efi_system_table is an EfiSystemTable as enforced by the input argument. We have guaranteed that system_table.configuration_table is non-null and this logic enforces that if it is non-null it points to a valid slice.
```

patina_dxe_core/src/allocator.rs
```diff
@@ -568,6 +568,8 @@ impl AllocatorMap {
 
     // resets the ALLOCATOR map to empty and resets the static allocators.
     #[cfg(test)]
+    // Safety: Caller must ensure that no allocations are active and that no other
```
os-d: IIRC, the lint for each of these checks for `SAFETY:` as a preceding comment, case-sensitive. You can enable the lint and run against this and see if they fail, maybe I am misremembering.

patina_dxe_core/src/driver_services.rs
```diff
@@ -47,6 +47,7 @@ fn get_platform_driver_override_bindings(
     {
         Err(_) => return Vec::new(),
         Ok(protocol) => unsafe {
+            // Safety: Pointer is validated using .expect(), will panic if .as_mut() returns a NULL pointer
```
os-d:
```suggestion
            // Safety: locate_protocol guarantees that if `Ok` is returned, a valid pointer is encapsulated in it 
```

os-d: Same for below with the different functions. The unsafe part isn't if we return a NULL pointer here, its us trying to get a reference from a pointer, which is only safe if the pointer is valid, which we are getting as a return value from these funcs.

patina_dxe_core/src/driver_services.rs
```diff
@@ -133,6 +136,8 @@ fn get_all_driver_bindings() -> Vec<*mut efi::protocols::driver_binding::Protoco
         Ok(handles) => get_bindings_for_handles(handles),
     };
 
+    // Safety: driver_bindings must contain valid pointers/handles (a & b as well as *a & *b)
+    // when sort_unstable_by() is called. Handles are either acquired or created above.
```
os-d: This one also we are relying on get_bindings_for_handles to provide us with a valid Vec.

patina_dxe_core/src/driver_services.rs
```diff
@@ -323,7 +331,8 @@ pub unsafe fn core_connect_controller(
 
     if recursive {
         for child in PROTOCOL_DB.get_child_handles(handle) {
-            //ignore the return value to match behavior of edk2 reference.
+            // ignore the return value to match behavior of edk2 reference.
+            // Safety: Caller must ensure that the driver binding handle remains valid for the duration of the call.
```
os-d:
```suggestion
            // Safety: We do not remove any of these handles during this call, though it is possible for a different entity at another TPL to do so.
```

patina_dxe_core/src/driver_services.rs
```diff
@@ -491,6 +500,8 @@ pub unsafe fn core_disconnect_controller(
             .get_interface_for_handle(driver_handle, efi::protocols::driver_binding::PROTOCOL_GUID)
             .or(Err(EfiError::InvalidParameter))?;
         let driver_binding_interface = driver_binding_interface as *mut efi::protocols::driver_binding::Protocol;
+        // Safety: driver_binding_interface is validated above, would have been caught with the .or and ? above with
+        // the loop being terminated.
```
os-d, line 506: This is again relying on get_interface_for_handle to give us a valid pointer back

patina_dxe_core/src/driver_services.rs
```diff
@@ -519,8 +530,15 @@ extern "efiapi" fn disconnect_controller(
...
 ) -> efi::Status {
+    if controller_handle.is_null() {
+        return efi::Status::INVALID_PARAMETER;
+    }
+
     let driver_image_handle = NonNull::new(driver_image_handle).map(|x| x.as_ptr());
     let child_handle = NonNull::new(child_handle).map(|x| x.as_ptr());
+    // Safety: Caller must ensure controller_handle is valid for the duration of the call.
+    // driver_image_handle and child_handle are both created above using NonNull which should
+    // guarantee they are non-null pointers. controller_handle is validated above.
```
os-d: Same comment as core_connect_controller above

patina_dxe_core/src/allocator.rs
makubacki: `Safety:` has been used in many other places in the codebase, but we should start using `SAFETY:` for consistency and to match the [lint documentation](https://github.com/rust-lang/rust-clippy/blob/0a83962264f12f66011f5d0f62b2d7dd1a4c5781/clippy_lints/src/undocumented_unsafe_blocks.rs#L18).

I filed an issue to track that: https://github.com/OpenDevicePartnership/patina/issues/1234

makubacki: @Raymond-MS, for your PR, please use `SAFETY:`.

patina_dxe_core/src/allocator.rs
makubacki: Agree. @Raymond-MS, building off Oliver's comment there are two key requirements (1) non-null which is currently covered and (2) the buffer has a valid pool header which is only possible if it was allocated as a pool buffer. Both should be called out in a function-level safety comment.

patina_dxe_core/src/allocator.rs
makubacki: @Raymond-MS, since you mentioned offline you don't have much time to work on this, I filed the following issue to track marking the functions unsafe: https://github.com/OpenDevicePartnership/patina/issues/1235

You can either work on it in this PR or I'm fine with addressing the immediate safety comment issue so progress can be made on the lint being enabled in `patina_dxe_core` and then use the issue to track updating the function.

patina_dxe_core/src/allocator.rs
```diff
@@ -1231,6 +1236,9 @@ pub fn install_memory_services(bs: &mut efi::BootServices) {
 // Resets the ALLOCATOR map to empty and resets the static allocators for test purposes.
 #[cfg(test)]
 pub(crate) unsafe fn reset_allocators() {
```
makubacki, line 1242: Since the function is marked unsafe, can you add safety requirements in its documentation?

patina_dxe_core/src/driver_services.rs
```diff
@@ -80,6 +81,7 @@ fn get_family_override_bindings() -> Vec<*mut efi::protocols::driver_binding::Pr
     for handle in driver_binding_handles {
         match PROTOCOL_DB.get_interface_for_handle(handle, efi::protocols::driver_family_override::PROTOCOL_GUID) {
             Ok(protocol) => {
+                // Safety: Pointer is validated using .expect(), will panic if .as_mut() returns a NULL pointer
```
makubacki: I know we've been relatively loose with this overall, but there is more than just the pointer being null that needs to be considered. Such as:

- Aligned for the type
- Points to a properly initialized type that is not dangling
- Aliasing - Since this is mutable no other references may violate aliasing expectations to the address
- Lifetime - This reference won't outlive the region

---

Most of these are usually handled and accounted for. I just want to make sure they're actually being considered equally to whether the pointer is null and getting called out in comments as needed.

patina_dxe_core/src/allocator.rs
Raymond-MS: I was just following what the file was using for consistency. Do you want me to change the other Safety comments to SAFETY?

makubacki: I don't think that is necessary here. It can be done in a standalone PR and close the issue.

patina_dxe_core/src/allocator.rs
Raymond-MS: I will leave this to the other task and focus on just the safety comments in this PR.

patina_dxe_core/src/allocator.rs
Raymond-MS: This should get addressed in the task Michael set up.

patina_dxe_core/src/driver_services.rs
Raymond-MS: I am sorry, I am a little confused, is that not what is mentioned? driver_binding_interface is set twice. Once with the .get_interface_for_handle and then again right after as a mutable pointer. We validate the pointer with the .or, right? Is there a different way you wanted me to word this?

patina_dxe_core/src/driver_services.rs
Raymond-MS: I am sorry Michael, was this a general comment when investigating unsafe blocks or does this relate to this line specifically? In this situation, we don't need to validate all of those things, right?

makubacki: Just a general comment to make sure that's being taken into account for each case since a lot end up only describing the null pointer aspect.

makubacki (approved):

Issue #806, opened 2025-10-01
[rustc]: Tracking Issue: `c_variadic`


Javagedes:
### Unstable rustc feature name

c_variadic

### Unstable rustc feature issue link

https://github.com/rust-lang/rust/issues/44930

### Current progress of the feature

- [x] Tracking issue created
- [x] RFC created
- [x] RFC accepted
- [x] Implementation merged
- [ ] Feature stabilized

### RFC number and PR link

N/A - Used before current process

### Usage justification

This allows patina to implement the `install_multiple_protocol_interfaces` and `uninstall_multiple_protocol_interfaces` boot services methods.

### Crate Usage

- `patina_dxe_core`

joschock:
note: this would require an upstream change in r-efi to redefine these functions in variadic terms:
https://github.com/r-efi/r-efi/blob/97b55bed1c2c91dcbf787674849f05337ff80b33/src/system.rs#L992

Javagedes:
This is merely a tracking issue on our usage of this unstable rustc feature. No action on your part is needed. I only assigned it to you as I thought you did the initial implementation. If you did not, then I'll re-assign.

Again, no action needed regardless

joschock:
It looks like the upstream c_variadic is thru final comment period and will stabilize soon: https://github.com/rust-lang/rust/issues/44930. There's a bit of discussion of "efiapi" in the stabilization thread, but it's not clear to me whether "efiapi" will support it once it stabilizes.  Will need to do a bit more digging on that.

The targets we support (x64 and aarch64 uefi targets) empirically have correct behavior with extern "c" variadics in the environments we use - so I'd propose that we move forward with that as "stable" once the upstream c_variadics stablizes (which looks soon as noted). 

If we a) start supporting new targets (e.g. 32-bit arm) where extern "c" variadics produce incorrect code, or b) "efiapi" also stablizes variadics, then we can revisit and take appropriate action (moving to "efiapi" preferable if it supports variadics - moving to hand built-asm as needed if not).

os-d:
The feature has been stabilized and is being tracked for Rust 1.99.0: https://github.com/rust-lang/rust/milestone/147.

PR #1236, opened 2026-01-09
[Rebase & FF] Refactor System Tables implementation to add resilience around external modification.

### LINKED ISSUE #1206: [Bug]: System Table structures are not guaranteed to be externally visible.

### Current Behavior

Under the present design point the system tables and config tables implementations do not use `volatile` access when producing, consuming, or modifying system tables (the main system table, runtime services table, boot services table, and various configuration tables). This means that changes to those tables from within patina core are not guaranteed to be externally observable (i.e. the compiler can make assumptions about those accesses that might mean they don't actually get written to system memory), and changes to the system tables from outside patina (such as installation of new function pointers in the runtime services table by e.g. VariableServices) are not guaranteed to be observed by patina.

Empirically, this works at the moment, but it is essentially UB in the sense that it could _not_ work under future compiler revisions or with changes in optimizations enabled etc. 

### Expected Behavior

External driver should observe all changes that are made to externally observable structures, and all changes made by external drivers to externally observable structures should be observed by Patina.  Present code does not guarantee this, although it works in practice. 

### Steps To Reproduce

By inspection.

### Build Environment

```markdown
N/A
```

### Version Information

```text
Commit: c28a32ab2435e58eb6b332bfbab03a476396910f
```

### Urgency

Medium

I will fix it

No maintainer feedback needed

### Anything else?

_No response_

joschock:
## Description

In the present design, the system table stores `&'static` references to the System Table, Boot Services Table, and Runtime Services table. These refs imply to the rust compiler that these tables are immutable, but in fact, they are mutated under these references by drivers external to patina (e.g. when the variable service driver installs its APIs into the runtime services table). 

This refactor changes the underlying static immutable references to these tables into raw pointers to better reflect the reality of the table usage and minimize UB due to immutable ref covering a data structure that is potentially mutated. Note that it doesn't entirely eliminate the UB, since to do so would place requirements for synchronization on those external agents, none of which currently implement those requirements.  

This is a breaking change since it changes the instantiation of SDK boot services and runtime services wrapper to take raw pointers rather than references.

This PR also brings in a secondary refactor of `core_install_configuration_table` to avoid an unnecessary "Box::from_raw" conversion on the table (this is a separate commit). This routine was being touched by the above refactor anyway so was convenient for the refactor. Rebase and FF to include these as two separate changes.

Closes #1206.

- [x] Impacts functionality?
- [ ] Impacts security?
- [x] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Unit tests pass; live testing boots to OS without issues. 

## Integration Instructions

`StandardBootServices` and `StandardRuntimeServices` instantiations now take a raw pointer to the corresponding tables.

### DIFF
--- a/patina_dxe_core/src/config_tables.rs
+++ b/patina_dxe_core/src/config_tables.rs
@@ -12,7 +12,7 @@ use alloc::{boxed::Box, vec};
 use core::{
     ffi::c_void,
     ptr::{NonNull, slice_from_raw_parts_mut},
-    slice::from_raw_parts,
+    slice::{from_raw_parts, from_raw_parts_mut},
 };
 use patina::error::EfiError;
 use r_efi::efi;
@@ -52,79 +52,69 @@ pub fn core_install_configuration_table(
 ) -> Result<Option<NonNull<c_void>>, EfiError> {
     let mut system_table = efi_system_table.get();
 
-    //if a table is already present, reconstruct it from the pointer and length in the st.
-    let old_cfg_table = if system_table.configuration_table.is_null() {
-        assert_eq!(system_table.number_of_table_entries, 0);
-        None
-    } else {
-        // SAFETY: efi_system_table is an EfiSystemTable as enforced by the input argument.
-        // We have guaranteed that system_table.configuration_table is non-null and this logic
-        // enforces that if it is non-null it points to a valid slice.
-        let ct_slice_box = unsafe {
-            Box::from_raw_in(
-                slice_from_raw_parts_mut(system_table.configuration_table, system_table.number_of_table_entries),
-                &EFI_RUNTIME_SERVICES_DATA_ALLOCATOR,
-            )
-        };
-        Some(ct_slice_box)
-    };
-
-    let mut old_vendor_table_ptr = None;
-    // construct the new table contents as a vector.
-    let new_table = match old_cfg_table {
-        Some(cfg_table) => {
-            // a configuration table list is already present.
-            let mut current_table = cfg_table.to_vec();
-            let existing_entry = current_table.iter_mut().find(|x| x.vendor_guid == vendor_guid);
-            if !vendor_table.is_null() {
-                // vendor_table is not null; we are adding or modifying an entry.
-                if let Some(entry) = existing_entry {
-                    //entry exists, modify it.
-                    old_vendor_table_ptr = NonNull::new(entry.vendor_table);
-                    entry.vendor_table = vendor_table;
-                } else {
-                    //entry doesn't exist, add it.
-                    current_table.push(efi::ConfigurationTable { vendor_guid, vendor_table });
-                }
+    let (updated_table, old_vendor_table_ptr) = match system_table.configuration_table {
+        existing_tbl_ptr if existing_tbl_ptr.is_null() => {
+            // existing table is empty.
+            if vendor_table.is_null() {
+                // trying to delete a non-existing table
+                return Err(EfiError::NotFound);
             } else {
-                //vendor_table is none; we are deleting an entry.
+                // adding a new table to an empty configuration table list
+                (vec![efi::ConfigurationTable { vendor_guid, vendor_table }], None)
+            }
+        }
+        existing_table_ptr => {
+            // existing table is present. Make a copy of it as a Vec to process the updates.
+            // SAFETY: existing_table_ptr is non-null, and number_of_table_entries is valid.
+            let mut updated_table =
+                unsafe { from_raw_parts_mut(existing_table_ptr, system_table.number_of_table_entries).to_vec() };
+            let existing_entry = updated_table.iter_mut().find(|x| x.vendor_guid == vendor_guid);
+            if vendor_table.is_null() {
+                // deleting an entry.
                 if let Some(entry) = existing_entry {
                     //entry exists, we can delete it
-                    old_vendor_table_ptr = NonNull::new(entry.vendor_table);
-                    current_table.retain(|x| x.vendor_guid != vendor_guid);
+                    let old_vendor_table_ptr = NonNull::new(entry.vendor_table);
+                    updated_table.retain(|x| x.vendor_guid != vendor_guid);
+                    (updated_table, old_vendor_table_ptr)
                 } else {
-                    // Entry does not exist, so we can't delete it. Thus we leave the system table unmodified, but
-                    // since we reconstructed the Box with the config table pointer we got from the system table,
-                    // we need to forget it here to avoid dropping it and freeing the memory while it is still used.
-                    core::mem::forget(cfg_table);
+                    //entry doesn't exist, can't delete it.
                     return Err(EfiError::NotFound);
                 }
-            }
-            current_table
-        }
-        None => {
-            // config table list doesn't exist.
-            if !vendor_table.is_null() {
-                // table is some, meaning we should create the list and add this as the new entry.
-                vec![efi::ConfigurationTable { vendor_guid, vendor_table }]
             } else {
-                // table is null, but can't delete a table entry in a list that doesn't exist.
-                //since the list doesn't exist, we can leave the (null) pointer in the st alone.
-                return Err(EfiError::NotFound);
+                // adding or modifying an entry.
+                if let Some(entry) = existing_entry {
+                    //entry exists, modify it.
+                    let old_vendor_table_ptr = NonNull::new(entry.vendor_table);
+                    entry.vendor_table = vendor_table;
+                    (updated_table, old_vendor_table_ptr)
+                } else {
+                    //entry doesn't exist, add it.
+                    updated_table.push(efi::ConfigurationTable { vendor_guid, vendor_table });
+                    (updated_table, None)
+                }
             }
         }
     };
 
-    if new_table.is_empty() {
-        // if empty, just set config table ptr to null
+    // Updating the table. Reclaim the old table (if present) so it'll get dropped by the runtime allocator.
+    if !system_table.configuration_table.is_null() {
+        unsafe {
+            let _old_boxed_table = Box::from_raw_in(
+                slice_from_raw_parts_mut(system_table.configuration_table, system_table.number_of_table_entries),
+                &EFI_RUNTIME_SERVICES_DATA_ALLOCATOR,
+            );
+        }
+    }
+
+    if updated_table.is_empty() {
         system_table.number_of_table_entries = 0;
         system_table.configuration_table = core::ptr::null_mut();
     } else {
-        //Box up the new table and put it in the system table. The old table (if any) will be dropped
-        //when old_cfg_table goes out of scope at the end of the function.
-        system_table.number_of_table_entries = new_table.len();
-        let new_table = new_table.to_vec_in(&EFI_RUNTIME_SERVICES_DATA_ALLOCATOR).into_boxed_slice();
-        system_table.configuration_table = Box::into_raw_with_allocator(new_table).0 as *mut efi::ConfigurationTable;
+        // Move the updated table into an allocation in the runtime services data allocator.
+        system_table.number_of_table_entries = updated_table.len();
+        let updated_table = updated_table.to_vec_in(&EFI_RUNTIME_SERVICES_DATA_ALLOCATOR).into_boxed_slice();
+        system_table.configuration_table =
+            Box::into_raw_with_allocator(updated_table).0 as *mut efi::ConfigurationTable;
     }
 
     efi_system_table.set(system_table);


patina_dxe_core/src/protocols.rs
```diff
@@ -797,13 +798,14 @@ extern "efiapi" fn locate_device_path(
...
+
     //This bit of trickery is needed because r_efi definition of (Un)InstallMultipleProtocolInterfaces
     //is not variadic, due to rust only supporting variadic for "unsafe extern C" and not "efiapi"
-    //until very recently. For x86_64 "efiapi" and "extern C" match, so we can get away with a
-    //transmute here. Fixing it for other architectures more generally would require an upstream
-    //change in r_efi to pick up. There is also a bug in r_efi definition for
-    //uninstall_multiple_program_interfaces - per spec, the first argument is a handle, but
+    //until rust 1.91. For our purposes "efiapi" and "extern C" match, so we can get away with a
+    //transmute here. Fixing it properly would require an upstream change in r_efi to pick up. There is also a bug in
+    //the r_efi definition for uninstall_multiple_protocol_interfaces - per spec, the first argument is a handle, but
```
os-d, line 808: Do we have a task for fixing this? Seems like it should be pretty non-controversial to fix and we should...can you file a task if there isn't one?

patina_dxe_core/src/runtime.rs
```diff
@@ -68,7 +70,7 @@ impl RuntimeData {
     }
 }
 
-pub fn init_runtime_support(_rt: &mut efi::RuntimeServices) {
+pub fn init_runtime_support(_st: &mut EfiSystemTable) {
```
os-d: Just drop the param since it is unused?

patina_dxe_core/src/systemtables.rs
```diff
@@ -20,708 +20,848 @@ pub static SYSTEM_TABLE: tpl_mutex::TplMutex<Option<EfiSystemTable>> =
...
     #[allow(dead_code)]
-    pub fn system_table(&self) -> &efi::SystemTable {
-        self.system_table.as_ref()
+    /// Creates a new EFI System Table instance from the given raw pointer.
+    /// # Safety
+    /// The pointer must be valid and point to a properly initialized efi::SystemTable structure.
+    pub unsafe fn from_raw_pointer(ptr: *mut efi::SystemTable) -> Self {
+        unsafe {
+            assert!(
+                !(*ptr).boot_services.is_null(),
```
os-d, line 720: Check if ptr is null before dereferencing?

patina_dxe_core/src/systemtables.rs
```diff
@@ -20,708 +20,848 @@ pub static SYSTEM_TABLE: tpl_mutex::TplMutex<Option<EfiSystemTable>> =
...
     }
 
-    pub fn runtime_services_mut(&mut self) -> &mut efi::RuntimeServices {
-        // SAFETY: runtime_services pointer in the system table is valid once system table is constructed.
-        // Note: this creates a mutable reference to the runtime services pointer. The function should be marked unsafe
-        // since the caller needs to ensure that no other references to runtime services exist when this is called.
-        // Issue: #1200
-        unsafe { self.system_table.runtime_services.as_mut().expect("RuntimeServices uninitialized") }
+    /// Returns the raw pointer to the System Table.
+    pub fn as_ptr(&self) -> *mut efi::SystemTable {
```
os-d: nit: this seems like it would be a bit more canonical and useful to be `as_mut_ptr`

patina_dxe_core/src/systemtables.rs
```diff
@@ -20,708 +20,848 @@ pub static SYSTEM_TABLE: tpl_mutex::TplMutex<Option<EfiSystemTable>> =
...
+
+        st.boot_services = core::ptr::null_mut();
+        st.con_in = core::ptr::null_mut();
+        st.console_in_handle = core::ptr::null_mut();
+        st.con_out = core::ptr::null_mut();
+        st.console_out_handle = core::ptr::null_mut();
+        st.std_err = core::ptr::null_mut();
+        st.standard_error_handle = core::ptr::null_mut();
+
+        // SAFETY: this
```
os-d: didn't finish comment

sdk/patina/src/boot_services.rs
```diff
@@ -79,8 +79,21 @@ impl StandardBootServices {
         self.efi_boot_services.is_completed()
     }
 
-    fn efi_boot_services(&self) -> &efi::BootServices {
-        self.efi_boot_services.get().expect("Standard Boot Services is not initialized!")
+    // Returns the boot services pointer if initialized, panics otherwise.
+    fn as_ptr(&self) -> *mut efi::BootServices {
```
os-d: Same comment on naming

sdk/patina/src/boot_services.rs
```diff
@@ -79,8 +79,21 @@ impl StandardBootServices {
...
+        *self.efi_boot_services.get().expect("Standard Boot Services is not initialized!")
+    }
+
+    // returns the boot services as a static reference, panics if uninitialized.
+    // SAFETY: This is unsafe because the boot services structure may be subject to external mutation.
+    // In most use cases, this should not be an issue as external code does not typically mutate the boot services
+    // struct except in certain well-defined cases (e.g. MonotonicCount and CalculateCrc32 API installations into the
+    // boot services struct), and it is expected that those modifications will be targeted and not affect the usage of
+    // thestatic reference. If mutation does occur, however, it is technically UB, and this is marked unsafe to
```
os-d: // the static

sdk/patina/src/boot_services.rs
```diff
@@ -79,8 +79,21 @@ impl StandardBootServices {
...
 
-    fn efi_boot_services(&self) -> &efi::BootServices {
-        self.efi_boot_services.get().expect("Standard Boot Services is not initialized!")
+    // Returns the boot services pointer if initialized, panics otherwise.
+    fn as_ptr(&self) -> *mut efi::BootServices {
+        *self.efi_boot_services.get().expect("Standard Boot Services is not initialized!")
+    }
+
+    // returns the boot services as a static reference, panics if uninitialized.
+    // SAFETY: This is unsafe because the boot services structure may be subject to external mutation.
```
os-d: I'm not sure I'm quite following along with marking this unsafe or why it is UB if another entity updates the table. Technically, that could happen with any of our memory (and we of course have many patterns of sharing memory with other non-Patina components). But in this case, I don't feel like the caller of the function has anything to uphold here, so I don't feel it makes sense to mark the function unsafe. The caller of the function can't ensure anything about the boot services table, it is simply asking the pointer it has for a mutable reference. That's legitimate to do, even if the table has been mutated by another entity, it would just be one of the fields has a different value now, but we've made the pointer and reference mutable, so that should be anticipated.

That being said, I don't feel strongly on this, just reading through the comment it didn't seem to meet the standard for marking a function unsafe.

sdk/patina/src/boot_services.rs
```diff
@@ -106,57 +119,62 @@ impl Debug for StandardBootServices {
...
-            .field("get_next_monotonic_count", &(self.efi_boot_services().get_next_monotonic_count))
-            .field("install_configuration_table", &(self.efi_boot_services().install_configuration_table))
-            .field("calculate_crc32", &(self.efi_boot_services().calculate_crc32))
-            .finish()
+        // SAFETY: efi_boot_services() is unsafe to mark it as potentially subject to external mutation. If that were
+        // to happen during the debug formatting, the worst that could happen is that the debug output would be
+        // incorrect.
+        unsafe {
+            f.debug_struct("StandardBootServices")
+                .field("create_event", &(self.efi_boot_services().create_event))
```
os-d: Just call efi_boot_services once?

sdk/patina/src/boot_services.rs
```diff
@@ -1421,7 +1469,8 @@ impl BootServices for StandardBootServices {
     ) -> Result<BootServicesBox<'_, [&'static efi::Guid], Self>, efi::Status> {
         let mut protocol_buffer = ptr::null_mut();
         let mut protocol_buffer_count = 0;
-        match efi_boot_services_fn!(self.efi_boot_services(), protocols_per_handle)(
+        // SAFETY: assume that boot services struct is not being externally modified while in use.
```
os-d: All of these go back to my point that there is nothing for a caller to do but say, well, I hope no one else is modifying the table, but again, that's true of all structures, to some degree. Doesn't seem like it adds value because there is nothing actionable or even to make sure the caller thinks about when calling this.

sdk/patina/src/runtime_services.rs
```diff
@@ -63,9 +63,21 @@ impl StandardRuntimeServices {
         self.efi_runtime_services.is_completed()
     }
 
-    fn efi_runtime_services(&self) -> &efi::RuntimeServices {
-        // SAFETY: Runtime services lifetime is expected to live long enough.
-        self.efi_runtime_services.get().expect("Standard Runtime Services is not initialized!")
+    /// Returns the Runtime Services pointer. Panics if uninitialized.
+    pub fn as_ptr(&self) -> *mut efi::RuntimeServices {
```
os-d: Same naming comment

sdk/patina/src/runtime_services.rs
```diff
@@ -63,9 +63,21 @@ impl StandardRuntimeServices {
...
+        *self.efi_runtime_services.get().expect("Standard Runtime Services is not initialized!")
+    }
+
+    // returns the runtime services as a static reference, panics if uninitialized.
+    // SAFETY: This is unsafe because the runtime services structure may be subject to external mutation.
```
os-d: Same unsafe comment

sdk/patina/src/runtime_services.rs
```diff
@@ -93,22 +105,26 @@ impl Debug for StandardRuntimeServices {
...
-            .field("reset_system", &(self.efi_runtime_services().reset_system))
-            .field("get_next_high_mono_count", &(self.efi_runtime_services().get_next_high_mono_count))
-            .field("update_capsule", &(self.efi_runtime_services().update_capsule))
-            .finish()
+        // SAFETY: efi_runtime_services() is unsafe to mark it as potentially subject to external mutation. If that were
+        // to happen during the debug formatting, the worst that could happen is that the debug output would be
+        // incorrect.
+        unsafe {
+            f.debug_struct("StandardRuntimeServices")
+                .field("get_variable", &(self.efi_runtime_services().get_variable))
```
os-d: Call efi_runtime_services once?

patina_dxe_core/src/protocols.rs
joschock: I think #806 covers it.

patina_dxe_core/src/runtime.rs
joschock: happy to do so, following prior art here.

patina_dxe_core/src/systemtables.rs
joschock: I actually spent some time looking for prior art before picking the name. It's rather ambiguous: https://internals.rust-lang.org/t/as-ptr-vs-as-mut-ptr/9940, so I picked the one that seemed most consistent with other usages. But frankly, I'm happy to go either way, my preference strength on the name here is basically zero.

sdk/patina/src/boot_services.rs
joschock: The reason it is unsafe IMHO is because it is UB for memory to mutate behind a shared reference: https://doc.rust-lang.org/reference/behavior-considered-undefined.html#r-undefined.immutable. It's also UB because non-patina drivers writing the shared struct without synchronization technically cause a data race (another bullet on same list). I think both of those apply even if the "other party" doing the mutating is not rust - the compiler is still entitled to make potentially wrong assumptions based on the existence of the shared ref and the implied immutability of the struct that it refs.  

It is _not_ UB for other drivers to mutate the boot services table if no shared references to it exist (one of the main points of this PR is to get rid of long-lived shared references and replace them with raw pointers precisely to avoid that). But this function creates such a shared reference and thus opens a window to UB. 

As far as the caller not being able to do anything about that - the caller _could_ ensure that no mutation could occur on the external table e.g. by raising TPL to TPL_HIGH to kill concurrency.  IMHO doing so is an unnecessary level of paranoia (and also would make using these APIs really a pain), but I don't think it changes the unsafety of the function here, and I do want to make consumers of this function think about the fact that the table could (in theory) change out from under them. 

So - since the function opens a potential for UB that the caller should be cognizant of, I think unsafe is justified here.

sdk/patina/src/boot_services.rs
joschock: Here the safety comment is referring to the fact that we only need the table to not be modified while the shared ref exists, which is for the short lifetime of the wrapper function so we can get the function pointer to the call to the actual API. 

Truly paranoid code would look like:
```rust
let old_tpl = core_raise_tpl(TPL_HIGH)
let my_function_ptr = self.efi_boot_services().protocols_per_handle;
core_restore_tpl(old_tpl);
my_function_ptr(<whatever>)
```

I think that's unnecessary and pedantic because all current external writes I could find wouldn't break that assumption.  But the safety comment is essentially documenting that assumption that's required for correct code. 

To your point, I could move all this assumption documentation into the `efi_boot_services()` function that generates the ref; but I didn't do that because I think dangerous usages of that function are possible - for example, grabbing a 'static shared ref from that function and caching it in a global static is probably not a good idea.

joschock: I also had an iteration where I removed `efi_boot_services` altogether and forced all callers to deal directly with the *mut efi::BootServices. But that results in the same amount of unsafe blocks, tho IMHO it removes one of the two unsafety issues (the compiler no longer has a shared ref that allows it to assume immutability). I'm still not fully convinced that the shared ref does much but make things more unsafe here, so I could be convinced to remove it.

sdk/patina/src/boot_services.rs
joschock: A toy example of what I'm worried about. This doesn't exist today, but demos my concerns.

```rust
let my_bs_ref = my_standard_boot_serices.efi_boot_services();
my_bs_ref.start_image(runtime_dxe); // installs a new pointer in gBS->CalculateCrc32
let my_calc32 = my_bs_ref.calculate_crc32;
my_calc32 (buffer);
```

The compiler believes `my_bs_ref` is immutable. So it could reorder the read of the function pointer for calculate_crc32() - not the call, but the function pointer read from `my_bs_ref` - _before_ the `start_image` call that changes the function pointer. If it did such a re-ordering the call to calculate_crc32 would go to the old crc function rather than the one that was installed by runtime_dxe. I speculate that it might even be able to re-order the whole `my_calc32` call before `start_image`, if it is convinced that `start_image` won't modify `my_bs_ref` or `buffer`.  I don't believe that's the case because those calls are both externs, and I don't think you can re-order externs. But I'm not certain about that.

Again, I don't think this can happen with current implementations that I'm aware of, and I also can't imagine why such a reordering would occur - but the point is, _it could_ and it would not violate the language spec as I understand it (which also might not be the correct understanding - but this PR is a good place for enlightenment on that point - so if I've got it wrong, please help me understand what I'm missing here).

sdk/patina/src/boot_services.rs
os-d: Yeah, I mean from your description I would lean towards the path of removing the shared immutable ref at all (or make it mutable, though obviously that becomes a pain to use). Especially as patina should really not use the boot services table except to set it up (i.e. mutate it). Runtime services we are stuck with for longer, but still probably not long term, in terms of patina calling usage. But I think if we are concerned about the UB here, we should just eliminate the possibility, from the provided API anyway.

patina_dxe_core/src/systemtables.rs
os-d: Eh, yeah, reading that, either way I guess. To me personally it doesn’t make sense to say the mut in the name is whether you have &self or &mut self, at least in the form of as: mut ptr. That to me is what you are returning. If you had mut: as ptr, that would read as &mut self. But its all semantics. Feel free to close it or change it if anyone else has an opinion. Part of the weirdness, is it is clear cut if you also have a as_ptr, that returns the immutable version. So if you have as_ptr return a mut, later need to add an as_ptr that return an immutable ptr, you’ve just broken all consumers.

sdk/patina/src/boot_services.rs
os-d: No, I think your point is fair, but as mentioned above, I think the answer is to drop the shared ref and prevent this (from the API side, a caller can do whatever they want).

My main thing about making the function unsafe is that to me is if the caller needs to validate the invariants of calling the function, which in this case they don’t. Calling the function is always safe. The unsafety part here is something downstream could happen you don’t expect. But the function itself is not unsafe. So it is semantics, but then again what do we do but semantics :)

The [Rust docs](https://doc.rust-lang.org/std/keyword.unsafe.html) say:

unsafe fn: calling this function means abiding by a contract the
compiler cannot enforce.

Which to that definition, this fn is not unsafe. Calling the fn does not require any contract the compiler can’t enforce, i.e. the function itself has no UB. It’s later the UB occurs.

sdk/patina/src/boot_services.rs
joschock: Ok, I'll see about removing it. This is in the SDK, so it's more of an abstraction for components or external crates. Patina itself doesn't use it and generally just directly calls the internal boot services APIs. This is private to the impl, though, so removing it only affects the private code in these modules.

sdk/patina/src/boot_services.rs
joschock: The contract is "you have to ensure no one else modifies boot services".  This is technically achievable. But it also makes the shared ref basically useless and defeats the point of the function, I guess. I'm swayed to just remove it and let the caller deal with the raw pointer.

patina_dxe_core/src/config_tables.rs
```diff
@@ -50,81 +50,74 @@ pub fn core_install_configuration_table(
...
-                if let Some(entry) = existing_entry {
-                    //entry exists, modify it.
-                    old_vendor_table_ptr = NonNull::new(entry.vendor_table);
-                    entry.vendor_table = vendor_table;
-                } else {
-                    //entry doesn't exist, add it.
-                    current_table.push(efi::ConfigurationTable { vendor_guid, vendor_table });
-                }
+    let (updated_table, old_vendor_table_ptr) = match system_table.configuration_table {
+        x if x.is_null() => {
```
Javagedes: nit: I think we can use something other than `x` for the variable name :)

patina_dxe_core/src/systemtables.rs
```diff
@@ -20,708 +20,851 @@ pub static SYSTEM_TABLE: tpl_mutex::TplMutex<Option<EfiSystemTable>> =
...
-            set_timer: Self::set_timer_unimplemented,
-            wait_for_event: Self::wait_for_event_unimplemented,
-            signal_event: Self::signal_event_unimplemented,
-            close_event: Self::close_event_unimplemented,
-            check_event: Self::check_event_unimplemented,
-            install_protocol_interface: Self::install_protocol_interface_unimplemented,
-            reinstall_protocol_interface: Self::reinstall_protocol_interface_unimplemented,
-            uninstall_protocol_interface: Self::uninstall_protocol_interface_unimplemented,
-            handle_protocol: Self::handle_protocol_unimplemented,
+            raise_tpl: raise_tpl_unimplemented,
```
Javagedes, line 632: nit: why move these EFIAPI functions out of the impl? While I'm not against it - it is basically a no-op, it feels out of place in this PR.

sdk/patina/src/boot_services.rs
```diff
@@ -45,19 +45,19 @@ use tpl::{Tpl, TplGuard};
 
 /// This is the boot services used in the UEFI.
 pub struct StandardBootServices {
-    efi_boot_services: Once<&'static efi::BootServices>,
+    efi_boot_services: Once<*mut efi::BootServices>,
```
Javagedes, line 48: This seems like a safety degradation and more likely to lead to UB. Now we can pass it any pointer with any lifetime and cannot guarantee it will be valid for the lifetime of this StandardBootServices struct.

The underlying data could be reclaimed and reused and we would never know.

Javagedes (commented):
Overall this looks fine for trying to tell the compiler this is mutable, to prevent updates being optimized away.

sdk/patina/src/boot_services.rs
Javagedes: `Box::leak` returns a &'static mut. I feel like we could update the interface to be `Once<&'static mut efi::BootServices>` or something with a variable lifetime for tests like

```rust
pub struct StandardBootServices<'a> {
  efi_boot_services: Once<&'a mut efi::BootServices>
}
```

patina_dxe_core/src/systemtables.rs
joschock: primary motivation is to make them as private as possible. I don't think they should be part of the impl because they are never intended to be directly called except thru the system table that is being constructed by the function I moved them into. Mostly done as part of general cleanup since I'm refactoring most of the file anyway.

sdk/patina/src/boot_services.rs
joschock: The main point of this PR is to avoid UB caused by lying to the compiler by giving it a long-lived reference to these tables. 

That ref means that the compiler can assume things about the backing table i.e. - for shared ref it can assume it never changes (so it can reorder/reads/writes/etc). For mut ref, more subtle, but it can assume that no changes happen to the table that it (the rust compiler) is not aware of.

That's simply not true because it's a spec-defined global table with existing art that is mutating it outside Patina. In practice it works, because our guesses/intuition (informed by c pointer behavior) about how the compiler handles these refs tend to align with the implementation-specific reality today, and external mutation of the tables is fairly limited in scope. But it's still UB. 

If the tables were completely under Patina control (by which I mean Patina was the only thing that was allowed, by spec, to read/write them), then it would be a different story. That would require some pretty significant arch changes to the UEFI spec tho.

sdk/patina/src/boot_services.rs
joschock: I've removed the ref generation functions; I do agree this is cleaner.

sdk/patina/src/boot_services.rs
joschock: function's removed.

components/patina_mm/src/component/communicator/comm_buffer_update.rs
```diff
@@ -298,10 +298,10 @@ mod tests {
         communicator_ptr: *const MmCommunicator,
     ) -> Box<ProtocolNotifyContext> {
         let mock_bs = Box::leak(Box::new([0u8; core::mem::size_of::<r_efi::system::BootServices>()]));
-        let bs_ptr = mock_bs.as_ptr() as *const r_efi::system::BootServices;
+        let bs_ptr = mock_bs.as_mut_ptr() as *mut r_efi::system::BootServices;
         // SAFETY: bs_ptr points to an aligned mock buffer that was just created.
```
cfernald: Safety comment shouldnt be needed anymore.

patina_dxe_core/src/systemtables.rs
```diff
@@ -20,708 +20,851 @@ pub static SYSTEM_TABLE: tpl_mutex::TplMutex<Option<EfiSystemTable>> =
...
+        unsafe { self.runtime_services.write(table_copy) }
     }
 
-    #[coverage(off)]
-    extern "efiapi" fn convert_pointer_unimplemented(_: usize, _: *mut *mut c_void) -> efi::Status {
-        efi::Status::from_usize(EFI_NOT_AVAILABLE_YET)
+    // Used in tests and included for API completness; may be used in future.
+    #[allow(dead_code)]
+    /// Returns a copy of the Runtime Services table.
+    pub fn get(&self) -> efi::RuntimeServices {
```
cfernald, line 61: Should we provide getters/setters for individual fields? 

Or perhaps there is some `Deref`/`DerefMut` trait fanciness that can be done to avoid compiler assumptions without having to copy the whole table every access.

joschock: I thought about this; but in general where this is used is typically a one-and-done scenario, so not really a perf hot path. Where it _is_ potentially a hot path (e.g. in StandardBootServices->callsomething called in a tight loop or something)  then we use a raw dereference. 

So not opposed to something fancy, but didn't really see the need.

Issue #1071, opened 2025-11-17
[Bug]: Some core tests fail when run on an arm64 host


cfernald:
### Is there an existing issue for this?

- [x] I have searched existing issues

### Current Behavior

Some tests fail when run on an arm64 host in the following files:
 
- core\patina_internal_cpu\src\paging\aarch64.rs
- patina_dxe_core\src\allocator.rs
- patina_dxe_core\src\config_tables\memory_attributes_table.rs
- patina_dxe_core\src\gcd\spin_locked_gcd.rs

### Expected Behavior

Tests should pass on arm64 host

### Steps To Reproduce

Run `cargo make test` on arm64 host

### Build Environment

```markdown
- OS(s): Ubuntu 24.04 (but should repro on windows)
- Tool Chain(s): N/A
- Targets Impacted: Patina tests
```

### Version Information

```text
14.4.2
```

### Urgency

Low

### Are you going to fix this?

Someone else needs to fix it

### Do you need maintainer feedback?

No maintainer feedback needed

cfernald:
Tagging lines with `Issue #1071`

PR #1241, opened 2026-01-14
patina_dxe_core: Fix memory allocation tests for aarch64 runtime types

cfernald:
## Description

Current implementations of several tests and hob construction logic use a single page for runtime memory types. On aarch64, the runtime page allocation granularity is 64kb, so all of these memory regions are dropped causing the tests to fail.

Contributes to #1071 

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Unit tests

## Integration Instructions

N/A

### DIFF
--- a/patina_dxe_core/src/test_support.rs
+++ b/patina_dxe_core/src/test_support.rs
@@ -288,6 +288,7 @@ pub(crate) fn build_test_hob_list(mem_size: u64) -> *const c_void {
     // SAFETY: Test code - allocates memory for the test HOB list.
     let mem = unsafe { get_memory(mem_size as usize) };
     let mem_base = mem.as_mut_ptr() as u64;
+    assert!(mem_size >= 0x1B0000);
 
     // Build a test HOB list that describes memory layout as follows:
     //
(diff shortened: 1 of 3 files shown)


patina_dxe_core/src/test_support.rs
```diff
@@ -211,13 +211,16 @@ pub(crate) fn with_global_lock<F: Fn() + std::panic::RefUnwindSafe>(f: F) -> Res
 
 /// Allocates a chunk of memory of the specified size from the system allocator.
 ///
+/// The memory allocated with be 64Kb aligned to simplify alignment requirements such
+/// as AArch64 runtime memory.
```
makubacki:
```suggestion
/// The memory allocated will be 64Kb aligned to simplify alignment requirements such
/// as AArch64 runtime memory.
```

cfernald:
> It would be nice to have some of these values reflect a minimum size for 64kb AARCH64 runtime granularity, but it's also just test code.

Are you referring to the hob alignments and memory usages?

makubacki:
> > It would be nice to have some of these values reflect a minimum size for 64kb AARCH64 runtime granularity, but it's also just test code.
> 
> Are you referring to the hob alignments and memory usages?

Yes. Basically, if someone looks at this code for reference, they won't think these are arbitrary and somewhat excessive values. The code would reflect there is a minimum threshold plus some degree of arbitrariness.

cfernald:
> > > It would be nice to have some of these values reflect a minimum size for 64kb AARCH64 runtime granularity, but it's also just test code.
> > 
> > 
> > Are you referring to the hob alignments and memory usages?
> 
> Yes. Basically, if someone looks at this code for reference, they won't think these are arbitrary and somewhat excessive values. The code would reflect there is a minimum threshold plus some degree of arbitrariness.

I will tighten this up a bit and document the expectation in comments.

cfernald:
@makubacki, looked a bit. I could tighten it up a bit, but over-doing it is likely to make the tests pretty sensistive to change as it relies on calculating the alignment jumps in the buffer. I've decided to mostly just document the layout of the memory layout instead. Might tighten it a bit though.

PR #1242, opened 2026-01-14
patina_internal_cpu: remove page table creation test

cfernald:
## Description

This removes a test that causes exception on aarch64 targets. This test attempts to create an offline page table from user-mode which is not currently supported by the patina-paging crate as it will access system registers during table initialization. No equivalent test exists for x64.

Contributes to #1071 

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Unit tests

## Integration Instructions

N/A

### DIFF
--- a/core/patina_internal_cpu/src/paging/aarch64.rs
+++ b/core/patina_internal_cpu/src/paging/aarch64.rs
@@ -64,7 +64,6 @@ pub fn create_cpu_aarch64_paging<A: PageAllocator + 'static>(
 
 #[cfg(test)]
 #[coverage(off)]
-#[cfg(target_arch = "x86_64")] // Issue #1071
 mod tests {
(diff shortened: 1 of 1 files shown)


Javagedes (commented):
You could add a #[cfg_attr(target_arch = "x86_64", patina_test)] so that it only compiles as a test for x86_64 targets.

Or #[cfg_attr(target_arch = "aarch64", skip)] if you want it to show up in the test list, but as skipped

os-d:
> You could add
> 
> #[cfg_attr(target_arch = "aarch64", skip)] if you want it to show up in the test list, but as skipped on aarch64 host based unit tests

This is an aarch64 only test, so I agree about just removing it

makubacki (approved):

PR #1243, opened 2026-01-14
patina_internal_collections: Prevent out-of-bounds access in resize()

makubacki:
## Description

Add bounds checking before accessing buffer[idx] in Storage::resize().

When resizing to a buffer with no free space (idx >= buffer.len()), set available to null instead of attempting to access out-of-bounds memory.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- Unit tests (`cargo make all`)
- patina-qemu boot to EFI shell

## Integration Instructions

- N/A

### DIFF
--- a/core/patina_internal_collections/src/node.rs
+++ b/core/patina_internal_collections/src/node.rs
@@ -240,8 +240,12 @@ where
 
         let idx = if !self.available.get().is_null() { self.idx(self.available.get()) } else { self.len() };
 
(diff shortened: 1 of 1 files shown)


garybeihl:
Good catch - there are a couple more out-of-bounds checks needed - I opened a new PR #1247

PR #1244, opened 2026-01-14
Add new `Handle` param for patina components

Javagedes:
## Description

This commit adds a new `Handle` param that can be requested by patina components. This component is currently only an on-demand wrapper around a r_efi::efi::Handle, which can be used by components when interacting with UEFI services. However future enhancements may include additional metadata useful from within the patina-component context.

This commit required that the system tables were setup before components added, so that protocols can be added to the protocol db without breaking the current API barrier.

I plan on adding unit tests if this is the route we wish to go.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Continued boot to shell.

## Integration Instructions

Users may use the provided `Handle`.

### DIFF
--- a/sdk/patina/src/component/params.rs
+++ b/sdk/patina/src/component/params.rs
@@ -105,7 +105,7 @@ use core::{
 use alloc::{borrow::Cow, boxed::Box};
 
 use crate::{
-    boot_services::StandardBootServices,
+    boot_services::{BootServices, StandardBootServices},
     component::{
         metadata::MetaData,
         service::IntoService,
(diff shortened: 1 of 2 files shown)


sdk/patina/src/component/params.rs
```diff
@@ -711,6 +711,80 @@ unsafe impl Param for StandardRuntimeServices {
...
+/// A handle for self-identification of a component.
+///
+/// Currently only contains an EFI handle used for direct interaction with UEFI services.
+#[derive(Debug, Clone, Copy)]
+pub struct Handle(r_efi::efi::Handle);
+
+impl Handle {
+    /// Creates an instance of Handle from the given EFI handle.
+    ///
+    /// This function is intended for testing purposes only.
```
cfernald, line 723: Is this intended for unit tests or patina-tests? Seems odd to leave test function without any feature wrapper or otherwise.

sdk/patina/src/component/params.rs
```diff
@@ -711,6 +711,80 @@ unsafe impl Param for StandardRuntimeServices {
...
+
+    /// Returns Self as the EFI handle associated with the component.
+    ///
+    /// ## Example
+    ///
+    /// ```rust
+    /// use patina::component::params::Handle;
+    /// use patina::boot_services::{BootServices, StandardBootServices};
+    ///
+    /// fn my_component(handle: Handle, bs: StandardBootServices) {
```
cfernald, line 736: I'm curious the motivation for adding this, is there a particular use case you had in mind?

sdk/patina/src/component/params.rs
```diff
@@ -711,6 +711,80 @@ unsafe impl Param for StandardRuntimeServices {
...
+    fn validate(_state: &Self::State, _storage: UnsafeStorageCell) -> bool {
+        true
+    }
+
+    fn init_state(storage: &mut Storage, _meta: &mut MetaData) -> Result<Self::State, Cow<'static, str>> {
+        const PATINA_COMPONENT_GUID: crate::BinaryGuid =
+            crate::BinaryGuid::from_string("3346877B-C962-4A34-A42D-50B8437B827D");
+
+        // SAFETY: We are installing a marker protocol, so the interface is expected to be null.
+        let result = unsafe {
```
cfernald, line 771: Is there a way for us to add a log that can help us associate a handle with a component here?

sdk/patina/src/component/params.rs
```diff
@@ -711,6 +711,80 @@ unsafe impl Param for StandardRuntimeServices {
...
+        let state = *state;
+        Self(state as r_efi::efi::Handle)
+    }
+
+    fn validate(_state: &Self::State, _storage: UnsafeStorageCell) -> bool {
+        true
+    }
+
+    fn init_state(storage: &mut Storage, _meta: &mut MetaData) -> Result<Self::State, Cow<'static, str>> {
+        const PATINA_COMPONENT_GUID: crate::BinaryGuid =
```
joschock, line 767: I'm not sure a marker is sufficient, at least for the purposes of the boot orchestrator that @kat-perez  is working on. That wants to use this handle as the `parent_handle` in the call to `load_image`, and that means it should be an image handle (i.e. a handle with LoadedImage/LoadedImageDevicePath on it).

sdk/patina/src/component/params.rs
Javagedes: It is intended for unit tests. All of the different `Param` implementations have one. I'm happy to put it behind a feature if you wish. I'd probably do it in a separate PR and do it for all of them though.

sdk/patina/src/component/params.rs
Javagedes: IIRC it was requested in the past, but the main motivator this time is for `patina_boot` (#1225), which loads images and needs a valid image handle.

sdk/patina/src/component/params.rs
Javagedes: ah. Got it. I'll look into our ability to add the two protocols like we do for a normal image.

Javagedes:
Closing this. After talking to people offline, I think it makes the most amount of sense to re-use the patina DXE_CORE handle, which means the original implementation suggested in #1225 makes the most amount of sense

PR #1250, opened 2026-01-16
LocateHandle() search type ByRegisterNotify update

### LINKED ISSUE #1251: [Bug]: EFI_BOOT_SERVICES.LocateHandle() SearchType ByRegisterNotify is missing handles prior to registration

### Current Behavior

Registering a callback then using LocateHandle with the search type ByRegisterNotify will result in the callbacks not having access to any of the handles that were present in the core prior to the registration event.

### Expected Behavior

EDK II code will return handles to any protocol installed prior to performing the LocateHandle calls.

### Steps To Reproduce

Example:
1) Protocol 1234 is added to core
2) Register for a callback on 1234 with search key so ByRegisterNotify will properly index returned value
3) In callback, LocateHandle with search key will report no protocols available
4) Add Protocol 1234 again
5) In callback, LocateHandle with search key will return just one handle, the second addition of 1234

Running the same experiment on EDK II code will result in step 3 returning a handle to the first installation of protocol 1234.

### Build Environment

```markdown
N/A
```

### Version Information

```text
19.03
```

### Urgency

Medium

I will fix it

No maintainer feedback needed

### Anything else?

I have a fix already and a PR will be added today.

rogurr:
## Description

The current EDK II implementation performs an immediate callback when registering through LocateHandle using ByRegisterNotify and will provide each handle normally for protocols installed prior to the registration.  The Patina code was written to start caching handles only after the registration is performed.  This change will seed the cache with prior registered protocols at the registration to match the EDK II implementation.

During debug, found a location where locate_handle() was calling the underlying layer first for buffer size, but not handling the error paths.  Updated that call to return an error if encountered.

- [X] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [X] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Performing a port of a private UEFI to use Patina which requires the LocateHandle call to have knowledge of previously installed protocols.  Tested with my current code and tested with the added test functions.

## Integration Instructions

N/A

### DIFF
--- a/patina_dxe_core/src/protocol_db.rs
+++ b/patina_dxe_core/src/protocol_db.rs
@@ -2038,7 +2038,7 @@ mod tests {
             let existing_handle =
                 SPIN_LOCKED_PROTOCOL_DB.install_protocol_interface(None, guid1, interface1).unwrap().0;
 
-            // Register for notificaiton
+            // Register for notification
             let event = 0x8765 as *mut c_void;
             let reg = SPIN_LOCKED_PROTOCOL_DB.register_protocol_notify(guid1, event).unwrap();
 
--- a/patina_dxe_core/src/protocol_db.rs
+++ b/patina_dxe_core/src/protocol_db.rs
@@ -2038,6 +2038,7 @@ mod tests {
             let existing_handle =
                 SPIN_LOCKED_PROTOCOL_DB.install_protocol_interface(None, guid1, interface1).unwrap().0;
 
+            // Register for notificaiton
             let event = 0x8765 as *mut c_void;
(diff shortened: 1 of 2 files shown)


os-d (approved):
Don't need to change for this PR, but for future reference, when we have multiple unrelated changes that make sense to go in the same PR, we mark the PR title as [REBASE & FF] <title> and then have the branch have nice clean, independently reviewable commits in it and when we go to merge it, we choose the rebase option instead of squash, so we keep the commit history clear for separate changes. Again, don't need to change this for this one, but for the future.

joschock (approved):

PR #1259, opened 2026-01-22
Clean up component params tests

kat-perez:
## Summary
- Rename `test_config_can_be_accessed_while_unlocked` to `test_config_can_be_accessed_when_locked` (configs are locked by default when using `Config::init_state()`)
- Simplify `test_boot_services_can_be_retrieved` and `test_runtime_services_can_be_retrieved` to use `MaybeUninit::zeroed().as_mut_ptr()` for consistency with the rest of the codebase

### DIFF
--- a/sdk/patina/src/component/params.rs
+++ b/sdk/patina/src/component/params.rs
@@ -816,10 +816,11 @@ mod tests {
     }
 
     #[test]
-    fn test_config_can_be_accessed_while_unlocked() {
+    fn test_config_can_be_accessed_when_locked() {
         let mut storage = Storage::new();
         let mut mock_metadata = MetaData::new::<i32>();
 
+        // Config::init_state adds config which is locked by default
         let id = Config::<i32>::init_state(&mut storage, &mut mock_metadata).unwrap();
 
         assert!(Config::<i32>::try_validate(&id, (&storage).into()).is_ok());
@@ -849,7 +850,9 @@ mod tests {
         let mut storage = Storage::new();
         let mut mock_metadata = MetaData::new::<i32>();
 
+        // Config::init_state adds config which is locked by default, so ConfigMut cannot access it
         let id = Config::<i32>::init_state(&mut storage, &mut mock_metadata).unwrap();
+
         assert!(
             ConfigMut::<i32>::try_validate(&id, (&storage).into())
                 .is_err_and(|err| err == "patina::component::params::ConfigMut<'_, i32> not available.")
@@ -931,16 +934,8 @@ mod tests {
         let mut storage = Storage::default();
         let mut mock_metadata = MetaData::new::<i32>();
 
-        // OOF, this is bad. But I don't wan't to write dummy functions for all the boot service functions. So we do this
-        // instead, so that the pointer to the boot services table is not null.
-        #[allow(invalid_value)]
-        let mut efi_bs = core::mem::MaybeUninit::<r_efi::efi::BootServices>::zeroed();
-
-        // SAFETY: Test code - Creating StandardBootServices from a zeroed BootServices struct for testing.
-        // This is acceptable in test code as we're only checking parameter validation logic.
-        let bs = StandardBootServices::new(efi_bs.as_mut_ptr());
-
-        storage.set_boot_services(bs);
+        let mut mock_bs = core::mem::MaybeUninit::<r_efi::efi::BootServices>::zeroed();
+        storage.set_boot_services(StandardBootServices::new(mock_bs.as_mut_ptr()));
 
         <StandardBootServices as Param>::init_state(&mut storage, &mut mock_metadata).unwrap();
         assert!(<StandardBootServices as Param>::try_validate(&(), (&storage).into()).is_ok());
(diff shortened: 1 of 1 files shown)


sdk/patina/src/component/params.rs
```diff
@@ -816,12 +816,15 @@ mod tests {
...
 
     #[test]
-    fn test_config_can_be_accessed_while_unlocked() {
+    fn test_config_can_be_accessed_when_locked() {
         let mut storage = Storage::new();
         let mut mock_metadata = MetaData::new::<i32>();
 
         let id = Config::<i32>::init_state(&mut storage, &mut mock_metadata).unwrap();
 
+        // Config<T> requires configs to be locked
```
Javagedes: While I agree with the name change, manually calling `storage.lock_configs()` should not be necessary. All configs are locked by default unless a ConfigMut is registered, then the config is unlocked for that particular inner type.

sdk/patina/src/component/params.rs
```diff
@@ -850,6 +853,10 @@ mod tests {
         let mut mock_metadata = MetaData::new::<i32>();
 
         let id = Config::<i32>::init_state(&mut storage, &mut mock_metadata).unwrap();
+
+        // Lock configs so ConfigMut cannot access
+        storage.lock_configs();
```
Javagedes: Same here. The config is locked by default, so this is not necessary.

sdk/patina/src/component/params.rs
```diff
@@ -931,14 +938,13 @@ mod tests {
...
-        // SAFETY: Test code - Creating StandardBootServices from a zeroed BootServices struct for testing.
-        // This is acceptable in test code as we're only checking parameter validation logic.
-        let bs = StandardBootServices::new(efi_bs.as_mut_ptr());
+        // Create an uninitialized BootServices struct. We don't call any functions on it - we just
+        // need a valid non-null pointer for parameter validation testing.
+        let efi_bs: core::mem::MaybeUninit<r_efi::efi::BootServices> = core::mem::MaybeUninit::uninit();
+        // SAFETY: The struct is leaked (never dropped or read), and we only use it to verify
+        // that the Param trait correctly validates and retrieves StandardBootServices.
+        let efi_bs_ptr = Box::leak(Box::new(unsafe { efi_bs.assume_init() }));
+        let bs = StandardBootServices::new(efi_bs_ptr);
```
Javagedes: `StandardBootServices` now has a `new_uninit()` function.

Does this test still work if you only do:

```rust
let bs = StandardBootServes::new_uninit()
```

I'm not confident, this works, but could possibly clean this up if it does.

sdk/patina/src/component/params.rs
kat-perez: Removed

sdk/patina/src/component/params.rs
kat-perez: Removed

sdk/patina/src/component/params.rs
kat-perez: If we used `new_uninit()` it could look like this:

``` rust
  let bs = StandardBootServices::new_uninit();
  // Use a dangling pointer - we never dereference it, just check is_init()
  bs.init(core::ptr::NonNull::<r_efi::efi::BootServices>::dangling().as_ptr());
  storage.set_boot_services(bs);
```

joschock: you could also consider making `default_boot_services_table` (https://github.com/OpenDevicePartnership/patina/blob/db4636d3b39f31cd509442ba9d190c37ba8526c3/patina_dxe_core/src/systemtables.rs#L285) public for tests

makubacki: In the past, I thought we've used `MaybeUninit` in places like this.

```rust
let mut mock_bs = core::mem::MaybeUninit::<r_efi::efi::BootServices>::zeroed();
core.storage.set_boot_services(StandardBootServices::new(mock_bs.as_mut_ptr()));
```

makubacki: I see that was done here before. I think John's suggestion is fine there is also [a macro](https://github.com/kat-perez/patina/blob/66ca9f014ec1d9eaf6812760cb05afa4434bb9c0/sdk/patina/src/boot_services.rs#L1684) that allows individual functions to be initialized in boot_services.rs

Javagedes: > If we used `new_uninit()` it could look like this:
> 
> ```rust
>   let bs = StandardBootServices::new_uninit();
>   // Use a dangling pointer - we never dereference it, just check is_init()
>   bs.init(core::ptr::NonNull::<r_efi::efi::BootServices>::dangling().as_ptr());
>   storage.set_boot_services(bs);
> ```

In regard to my portion of this request. The current way is better then my suggested way. If you wish to do John or Michael's suggestion, then I'm cool with that. Bu no reason to use `StandardBootServices::new_uninit()`

Javagedes (approved):

sdk/patina/src/component/params.rs
makubacki: I understand we're creating uninitialized services and that's bad if we try to use them. Can someone please clarify what exactly happened here? Is that happening?

kat-perez: I'll go with `MaybeUninit` to keep it consistent

Javagedes: > I understand we're creating uninitialized services and that's bad if we try to use them. Can someone please clarify what exactly happened here? Is that happening?

These are just a simple unit test to ensure `StandardRuntimeServices` and `StandardBootServices` can be retrieved from storage so long as the underlying pointer has been set. No underlying functions are ever called, so we don't need it to be valid. We just need to make sure that the pointer is not null. So this is completely fine.

This comment I made was really just regarding the pedantics and possibly not needing to do a leak.

makubacki: Thanks, that all makes sense. Just wanted to double check something else didn't come up unexpected that this is addressing.

makubacki:
@kat-perez, can you please update the PR title and description? At this point, the PR is mostly just cleanup.

Issue #1235, opened 2026-01-09
[Task]: Make `core_free_pool()` and `core_free_pages()` unsafe


makubacki:
### Request Description

There are requirements on the `core_free_pool()` and `core_free_pages()` functions that caller's must take into account beyond simple pointer expections.

```rust
pub fn core_free_pages(memory: efi::PhysicalAddress, pages: usize) -> Result<(), EfiError>
```

```rust
pub fn core_free_pool(buffer: *mut c_void) -> Result<(), EfiError>
```

This issue tracks marking the functions as unsafe with the caller expectations that the memory buffer being given to the function is allocated per the expected APIs (in addition to not being null).

---

The goal is to clearly communicate to `free()` callers the safety requirements. Therefore, this will extend up the call stack to functions the directly call the free APIs.

### Are you going to make the change?

Someone else needs to make the change

### Do you need maintainer feedback?

No maintainer feedback needed

makubacki:
Moved to backlog for now. Waiting on resolution from https://github.com/r-efi/r-efi/pull/90.

PR #1260, opened 2026-01-23
Make `core_free_pool()` and `core_free_pages()` unsafe

berlin-with0ut-return:
## Description

Addresses https://github.com/OpenDevicePartnership/patina/issues/1235.

Mark `core_free_pages` and `core_free_pool` as unsafe as they work with raw pointers/addresses.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [X] Includes documentation? - comments document new safety requirements

## How This Was Tested
N/A.

## Integration Instructions
N/A.

### DIFF
--- a/patina_dxe_core/src/config_tables/memory_attributes_table.rs
+++ b/patina_dxe_core/src/config_tables/memory_attributes_table.rs
@@ -202,7 +202,7 @@ pub fn core_install_memory_attributes_table() {
             }
 
             // SAFETY: `mat_ptr` is non-null and was successfully allocated with sufficient size above.
-            let mat = &mut unsafe { *mat_ptr };
+            let mat = unsafe { &mut *mat_ptr };
             mat.version = efi::MEMORY_ATTRIBUTES_TABLE_VERSION;
             mat.number_of_entries = mat_desc_list.len() as u32;
             mat.descriptor_size = size_of::<efi::MemoryDescriptor>() as u32;


patina_dxe_core/src/allocator.rs
```diff
@@ -617,14 +617,35 @@ pub fn core_allocate_pool(pool_type: efi::MemoryType, size: usize) -> Result<*mu
     }
 }
 
+// Frees a pool allocated with `core_allocate_pool`.
+//
+//  # Safety
+// Freeing from a raw pointer is inherently unsafe.
+// For this function to operate safely, the caller must ensure that:
+// 1. The buffer pointer is valid: non-null and correctly aligned.
```
os-d, line 625: I don’t think there is an alignment requirement here. This is pool memory and can be anywhere. It should be right after the pool signature, but that’s not part of the safety, we’ll gracefully fail.

patina_dxe_core/src/allocator.rs
```diff
@@ -617,14 +617,35 @@ pub fn core_allocate_pool(pool_type: efi::MemoryType, size: usize) -> Result<*mu
...
     }
 }
 
-pub fn core_free_pool(buffer: *mut c_void) -> Result<(), EfiError> {
+// Frees a pool allocated with `core_allocate_pool`.
+//
+//  # Safety
+// Freeing from a raw pointer is inherently unsafe.
+// For this function to operate safely, the caller must ensure that:
+// 1. The buffer pointer is valid: non-null and correctly aligned.
```
os-d, line 643: Same

patina_dxe_core/src/allocator.rs
os-d: I think non-null is not part of the safety contract, a caller can pass null in and we should return successfully. It is part of the non-safety part of the contract

patina_dxe_core/src/allocator.rs
```diff
@@ -725,14 +746,37 @@ pub fn memory_type_for_handle(handle: efi::Handle) -> Option<efi::MemoryType> {
     ALLOCATORS.lock().memory_type_for_handle(handle)
 }
 
+// Frees a pool allocated with `allocate_pages`.
+//
+//  # Safety
+// Freeing from a raw address of unknown allocation size is inherently unsafe.
+// For this function to operate safely, the caller must ensure that:
+// 1. The address is valid: non-null and correctly page-aligned.
```
os-d, line 754: Same

patina_dxe_core/src/allocator.rs
```diff
@@ -725,14 +746,37 @@ pub fn memory_type_for_handle(handle: efi::Handle) -> Option<efi::MemoryType> {
...
     }
 }
 
-pub fn core_free_pages(memory: efi::PhysicalAddress, pages: usize) -> Result<(), EfiError> {
+// Frees a pool allocated with `core_allocate_pages`.
+//
+//  # Safety
+// Freeing from a raw address of unknown allocation size is inherently unsafe.
+// For this function to operate safely, the caller must ensure that:
+// 1. The address is valid: non-null and correctly page-aligned.
```
os-d, line 773: Same

patina_dxe_core/src/memory_manager.rs
```diff
@@ -76,8 +76,19 @@ impl MemoryManager for CoreMemoryManager {
         }
     }
 
+    // Frees a pool allocated with `allocate_pages`.
+    //
+    //  # Safety
+    // Freeing from a raw address of unknown allocation size is inherently unsafe.
+    // For this function to operate safely, the caller must ensure that:
+    // 1. The address is valid: non-null and correctly page-aligned.
```
os-d, line 84: Same

patina_dxe_core/src/allocator.rs
```diff
@@ -617,14 +617,35 @@ pub fn core_allocate_pool(pool_type: efi::MemoryType, size: usize) -> Result<*mu
...
+//
+//  # Safety
+// Freeing from a raw pointer is inherently unsafe.
+// For this function to operate safely, the caller must ensure that:
+// 1. The buffer pointer is valid: non-null and correctly aligned.
+// 2. The buffer was originally allocated by `core_allocate_pool` and was returned by one of the static or dynamic allocators.
+// 3. The buffer must refer to the start of the allocation, not an offset within it.
+// 4. The buffer must not have been freed previously.
+// 5. No aliased references to the buffer should exist after this call.
 extern "efiapi" fn free_pool(buffer: *mut c_void) -> efi::Status {
```
makubacki, line 630: This was not as clear as it could have been in the issue but one reason this was filed as an issue instead of implemented right away was because the real spirit of the change is to mark the primary caller invoked function `free_pool()` unsafe. That is what the issue meant by:

> The goal is to clearly communicate to free() callers the safety requirements. Therefore, this will extend up the call stack to functions the directly call the free APIs.

Key problem being that this is in the `BootServices` structure we use from r-efi. Though, that was discussed here https://github.com/r-efi/r-efi/pull/90.

I think some of the additional info is nice, but for the limited call sites to `core_free_pool()` I'm actually concerned that treating a C FFI wrapper as safe to an unsafe function just makes this more confusing.

patina_dxe_core/src/config_tables/memory_attributes_table.rs
```diff
@@ -201,42 +201,43 @@ pub fn core_install_memory_attributes_table() {
                 return;
             }
 
-            // this ends up being a large unsafe block because we have to dereference the raw pointer core_allocate_pool
-            // gave us and convert it to a real type and back in order to install it
+            // SAFETY: `mat_ptr` is non-null and was successfully allocated with sufficient size above.
+            let mat = unsafe { &mut *mat_ptr };
```
makubacki, line 205: I think something like this is more idiomatic.

```rust
 // SAFETY: `mat_ptr` was successfully allocated above and checked for null below.
let mat = unsafe {
    if let Some(mat) = mat_ptr.as_mut() {
        mat
    } else {
        log::error!("Got a null ptr in successful return from allocate_pool. Failed to create MAT.");
        return;
    }
};
```

Or a variation leverages `.as_mut()`.

patina_dxe_core/src/allocator.rs
```diff
@@ -725,14 +746,37 @@ pub fn memory_type_for_handle(handle: efi::Handle) -> Option<efi::MemoryType> {
     ALLOCATORS.lock().memory_type_for_handle(handle)
 }
 
+// Frees a pool allocated with `allocate_pages`.
+//
+//  # Safety
+// Freeing from a raw address of unknown allocation size is inherently unsafe.
+// For this function to operate safely, the caller must ensure that:
+// 1. The address is valid: non-null and correctly page-aligned.
+// 2. The buffer at `memory` was originally allocated by `core_allocate_pages`.
```
makubacki, line 755: For cases where this is essentially inheriting the unsafe requirements as a simple wrapper, I think it'd be more maintainable and readable to just refer to the safety comment in the wrapped function and only call out relevant deltas here.

Like this: https://doc.rust-lang.org/src/alloc/alloc.rs.html#270

berlin-with0ut-return:
i'm moving this to draft for now so that we can decide what to do wrt r_efi unsafe PR. if that is merged quickly i will continue this PR by making the "efiapi" functions unsafe, otherwise i will pull out the comment fixes without the unsafe

berlin-with0ut-return:
i am going to close this for now as it's blocked by the r_efi unsafe issue and there doesn't seem to be much movement on it. will return when that issue is closed

makubacki:
> i am going to close this for now as it's blocked by the r_efi unsafe issue and there doesn't seem to be much movement on it. will return when that issue is closed

Thanks for taking the first pass at adding this.

Issue #1233, opened 2026-01-09
[Task]: Replace Assert/Panic/Etc with Proper Error Handling and debug_assert!


os-d:
### Is there an existing issue for this?

- [x] I have searched existing issues

### Current Behavior

There have been passes before to ensure there is proper error handling in Patina coupled with debug_asserts, but some instances have slipped through, e.g. use of assert*! is still in various, e.g. https://github.com/OpenDevicePartnership/patina/blob/1b8e961dc7d8b3977d321f874711ec3fea6f3f79/patina_dxe_core/src/config_tables.rs#L56 and https://github.com/OpenDevicePartnership/patina/blob/1b8e961dc7d8b3977d321f874711ec3fea6f3f79/patina_dxe_core/src/allocator.rs#L819.

There are likely other usage patterns that will result in panics in production code instead of safe error handling. One large area is overflow/underflow on every arithmetic operation, that doesn't need to be tackled here, most cases should be fine, but obviously some are at risk for it.

### Expected Behavior

Patina should only have raw panics, in any form, in cases where it explicitly cannot continue if a given condition arises. Otherwise, safe error handling should be in place, coupled with a debug_assert!() to catch unexpected conditions at runtime, if appropriate.

### Steps To Reproduce

n/a

### Build Environment

```markdown
- OS(s):
- Tool Chain(s):
- Targets Impacted:
```

### Version Information

```text
top of tree
```

### Urgency

Medium

### Are you going to fix this?

Someone else needs to fix it

### Do you need maintainer feedback?

No maintainer feedback needed

berlin-with0ut-return:
i self-assigned because i did similar work about a year ago but since it's marked good first issue anyone newer to patina/Rust is welcome to pair-program with me

makubacki:
@berlin-with0ut-return, can you summarize what work, if any, remains?

berlin-with0ut-return:
> [@berlin-with0ut-return](https://github.com/berlin-with0ut-return), can you summarize what work, if any, remains?

closed by https://github.com/OpenDevicePartnership/patina/pull/1264, but we should keep an eye out for this pattern in future PRs to limit the amount of cleanup necessary

PR #1264, opened 2026-01-23
Improve panic handling

berlin-with0ut-return:
## Description
Addresses https://github.com/OpenDevicePartnership/patina/issues/1233.

This PR improves panic handling and logging for error cases. Some panics have been converted to errors, while others have simply been improved in clarity.

- [X] Impacts functionality? -> instead of panic in some cases, returned error.
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested
`cargo make all`

## Integration Instructions
I would consider this more of a first pass. There may have been some panics that I missed, and possibly some panics that have been kept that would be better as Result/Options.

### DIFF
--- a/patina_dxe_core/src/gcd.rs
+++ b/patina_dxe_core/src/gcd.rs
@@ -312,7 +312,6 @@ impl MemoryProtectionPolicy {
                     UEFI_PAGE_SIZE,
                     descriptor.attributes & efi::CACHE_ATTRIBUTE_MASK,
                 ) {
-
                     log_debug_assert!("Failed to map page 0 for compat mode. Status: {e:#x?}");
                 }
             }
@@ -374,7 +373,6 @@ impl MemoryProtectionPolicy {
                         (addr, len) = match align_range(addr, len, UEFI_PAGE_SIZE as u64) {
                             Ok((aligned_addr, aligned_len)) => (aligned_addr, aligned_len),
                             Err(_) => {
-
                                 log_debug_assert!(
                                     "Failed to align address {addr:#x?} + {len:#x?} to page size, compatibility mode may fail",
                                 );
@@ -414,7 +412,6 @@ impl MemoryProtectionPolicy {
             .is_err()
         {
             // if we failed to map this image RWX, we should still attempt to execute it, it may succeed
-
 
             log_debug_assert!("Failed to set GCD attributes for image {}", filename);
         }


components/patina_adv_logger/src/memory_log.rs
```diff
@@ -190,7 +190,7 @@ impl<'a> AdvancedLog<'a> {
         // alignment padding for 8 byte alignment.
         let data_offset = size_of::<AdvLoggerMessageEntry>() as u16;
         let unaligned_size = data_offset as u32 + log_entry.data.len() as u32;
-        let message_size = align_up(unaligned_size, 8).unwrap() as u32;
+        let message_size = align_up(unaligned_size, 8)? as u32;
```
joschock: I think the `as 32` is not necessary since message_size is already u32.

patina_dxe_core/src/lib.rs
```diff
@@ -438,7 +448,10 @@ impl<P: PlatformInfo> Core<P> {
         // the initial free memory may not be enough to contain the HOB list. We need to relocate the HOBs because
         // the initial HOB list is not in mapped memory as passed from pre-DXE.
         hob_list.relocate_hobs();
-        assert!(self.set_hob_list(hob_list).is_ok());
+        if self.set_hob_list(hob_list).is_err() {
+            log::error!("HOB list was already set!");
+            debug_assert!(false, "HOB list was already set!");
```
joschock: is it safe to proceed if this fails? I'm not sure how to reason about the state of the core if this fails do to some kind of weird reentrancy.

sdk/patina_ffs/src/volume.rs
```diff
@@ -603,7 +603,11 @@ impl Volume {
 
                 //Per spec, max required_content_alignment is pad files is 16M (2^24). That means that pad file size
```
joschock:
```suggestion
                //Per spec, max required_content_alignment of pad files is 16M (2^24). That means that pad file size
```

patina_dxe_core/src/allocator.rs
```diff
@@ -822,7 +822,10 @@ extern "efiapi" fn get_memory_map(
     let map_size = unsafe { memory_map_size.read_unaligned() };
 
     let required_map_size = GCD.memory_descriptor_count_for_efi_memory_map() * mem::size_of::<efi::MemoryDescriptor>();
-    assert_ne!(required_map_size, 0);
+    debug_assert!(required_map_size != 0);
+    if required_map_size == 0 {
```
cfernald, line 832: Maybe NOT_FOUND is more appropriate. Technically neither is spec defined for this function, but I'm not concerned about that.

patina_dxe_core/src/lib.rs
cfernald: Yeah, I think this should panic

patina_dxe_core/src/allocator/fixed_size_block_allocator.rs
```diff
@@ -293,8 +293,13 @@ impl FixedSizeBlockAllocator {
             Some(index) => {
                 let new_node = BlockListNode { next: self.list_heads[index].take() };
                 // verify that block has size and alignment required for storing node
-                assert!(size_of::<BlockListNode>() <= BLOCK_SIZES[index]);
-                assert!(align_of::<BlockListNode>() <= BLOCK_SIZES[index]);
+                if size_of::<BlockListNode>() > BLOCK_SIZES[index] || align_of::<BlockListNode>() > BLOCK_SIZES[index] {
+                    panic!(
```
cfernald: I wonder if, as a rule, we should keep the panic short (e.g. "FSB deallocating block too small to store BlockListNode.") and have the reasoning kept to a comment before the panic. If nothing else, to reduce compiled space used for panic rationale.

patina_dxe_core/src/allocator.rs
```diff
@@ -1184,7 +1186,9 @@ pub fn init_memory_support(hob_list: &HobList) {
                 // SAFETY: this structure comes from the hob list, so it must be 8-byte aligned (meets alignment
                 // requirement for EfiMemoryTypeInformation), and length is calculated above to fit within the
                 // Guid HOB data. Assert if alignment is not as expected.
-                assert_eq!(memory_type_slice_ptr.align_offset(mem::align_of::<EFiMemoryTypeInformation>()), 0);
+                if memory_type_slice_ptr.align_offset(mem::align_of::<EFiMemoryTypeInformation>()) != 0 {
```
cfernald: This seems like this should be done as part of a const expression for compile time enforcement instead of in each invocation of this function, considering its not specific to any input here.

patina_dxe_core/src/misc_boot_services.rs
```diff
@@ -36,7 +36,10 @@ impl<T> ArchProtocolPtr<T> {
 
     // SAFETY: ptr must be a valid pointer to T and init must only be called once.
     unsafe fn init(&self, ptr: *mut c_void) {
-        assert!(!self.0.is_completed(), "Attempted to set ArchProtocolPtr more than once.");
+        if self.0.is_completed() {
+            log::error!("Attempted to initialize ArchProtocolPtr more than once.");
```
cfernald: I'd keep a debug_assert here as well, since this indicates a software bug elsewhere.

sdk/patina/src/boot_services.rs
```diff
@@ -790,8 +790,13 @@ pub trait BootServices {
         let interface_ptr = unsafe {
             self.locate_protocol_unchecked(protocol_guid, registration.map_or(ptr::null_mut(), |r| r.as_ptr()))?
         };
-        assert_eq!(ptr::null_mut(), interface_ptr);
-        Ok(())
+        if !interface_ptr.is_null() {
+            debug_assert!(false, "Marker protocol has no data; interface should be null {:?}", protocol_guid);
```
cfernald: Maybe creating a `log_debug_assert!` macro may be nice to help with the `debug_assert!` follow by a `log::error!` pattern, to avoid having to have the duplicate strings.

PR #1265, opened 2026-01-26
Cargo.toml: Update ruint to 1.17.1

makubacki:
## Description

Ensure a version is used that patches a Rust security vulnerability.

```
│ ruint 1.16.0 registry+https://github.com/rust-lang/crates.io-index
    │ ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ security vulnerability detected
    │
    ├ ID: RUSTSEC-2025-0137
    ├ Advisory: https://rustsec.org/advisories/RUSTSEC-2025-0137
    ├ The function `reciprocal_mg10` is marked as safe but can trigger
      undefined behavior (out-of-bounds access) because it relies on
      `debug_assert!` for safety checks instead of `assert!`.

      When compiled in release mode, the `debug_assert!` is optimized
      out, potentially allowing invalid inputs to cause memory
      corruption.
    ├ Announcement: https://github.com/recmo/uint/issues/550
    ├ Solution: Upgrade to >=1.17.1 (try `cargo update -p ruint`)
    ├ ruint v1.16.0
      └── (dev) patina_internal_collections v19.0.5
          └── patina_dxe_core v19.0.5
              ├── (dev) patina_adv_logger v19.0.5
              └── (dev) patina_mm v19.0.5
                  └── patina_performance v19.0.5
                      └── patina_dxe_core v19.0.5 (*)
```

This will use ruint version 1.17.1 or later (up to 2.0.0), so the fix is included in future compatible updates.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- `cargo make all`

## Integration Instructions

- N/A

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -65,7 +65,7 @@ criterion = { version = "^0.5" }
 env_logger = "0.11"
 mockall = { version = "0.13.0" }
 rand = { version = "0.8" }
-ruint = { version = "1" }
+ruint = { version = "1.17.1" }
 serial_test = { version = "3.0" }
 winapi = { version = "0.3" }
 serde = { version = "1.0", default-features = false, features = ["derive"] }


Cargo.toml
```diff
@@ -65,7 +65,7 @@ criterion = { version = "^0.5" }
 env_logger = "0.11"
 mockall = { version = "0.13.0" }
 rand = { version = "0.8" }
-ruint = { version = "1" }
+ruint = { version = "1.17.1" }
```
cfernald: Should we use the greater then option, or is there a reason we want to pin to a specific patch?

Javagedes: cargo all supports the ">=1.17" functionality if we wanted to do it that way

makubacki: I know we've talked about this a few times, but I think the `>=` syntax is just most intuitive. I'll update it to use that.

makubacki: Done

makubacki: I just went back to the original since it is simple and has the matching behavior I intend [[1]](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#version-requirement-syntax):

```
1.2.3  :=  >=1.2.3, <2.0.0
1.2    :=  >=1.2.0, <2.0.0
1      :=  >=1.0.0, <2.0.0
0.2.3  :=  >=0.2.3, <0.3.0
0.2    :=  >=0.2.0, <0.3.0
0.0.3  :=  >=0.0.3, <0.0.4
0.0    :=  >=0.0.0, <0.1.0
0      :=  >=0.0.0, <1.0.0
```

I'd like to have this update up to the next major version (`<2.0.0`). Given this a default version requirement, and already the case, I don't see the need to make it more complicated.

PR #1268, opened 2026-01-26
Fix AArch64 NUM_EXCEPTION_TYPES off-by-one bug

### LINKED ISSUE #1267: [Bug]: Arch64 NUM_EXCEPTION_TYPES off-by-one

### Current Behavior

`NUM_EXCEPTION_TYPES` for AArch64 is `3` but should be `4`.

The assembly (`exception_handler.asm`) defines 4 exception types (Synchronous=0, IRQ=1, FIQ=2, SError=3), but the Rust constant only allocates 3 handler slots.

When an SError occurs, the system panics:
`index out of bounds: the len is 3 but the index is 3`

Observed during warm boot when UFS hardware triggered an SError.

### Expected Behavior

SError exceptions (index 3) should be handled without panic. The handler array should have 4 slots to match the 4 AArch64 exception types.

### Steps To Reproduce

1. Build and run on AArch64 hardware
2. Perform a warm boot (reboot without full power cycle)
3. Wait for UFS/storage driver initialization
4. Observe panic when hardware triggers an SError exception

Alternatively:
- Attempt to register an SError handler (exception type 3) - it will return InvalidParameter

### Build Environment

```markdown
- Hardware: AArch64 platform with UFS storage
```

### Version Information

```text
tested on - patina workspace: 19.0.5
```

### Urgency

Low

I will fix it

No maintainer feedback needed

### Anything else?

_No response_

Nir3613:
Fixed an off-by-one bug in `NUM_EXCEPTION_TYPES` for AArch64. The constant was set to `3` but should be `4` to match the 4 exception types defined in `exception_handler.asm`:

| Index | Exception Type |
|-------|----------------|
| 0 | Synchronous |
| 1 | IRQ |
| 2 | FIQ |
| 3 | SError |

With the value of `3`, SError exceptions (index 3) caused an array out-of-bounds panic:
`index out of bounds: the len is 3 but the index is 3`

This was observed during warm boot when UFS hardware triggered an SError.

Fixes #1267 

- [x] Impacts functionality

## How This Was Tested

Verified that the constant now matches the assembly definition in `core/patina_internal_cpu/src/interrupts/aarch64/exception_handler.asm` (lines 109-112).

## Integration Instructions

N/A

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -65,7 +65,7 @@ criterion = { version = "^0.5" }
 env_logger = "0.11"
 mockall = { version = "0.13.0" }
 rand = { version = "0.8" }
-ruint = { version = "1" }
+ruint = { version = "1.17.1" }
 serial_test = { version = "3.0" }
 winapi = { version = "0.3" }
 serde = { version = "1.0", default-features = false, features = ["derive"] }
--- a/patina_dxe_core/src/allocator.rs
+++ b/patina_dxe_core/src/allocator.rs
@@ -1261,18 +1261,59 @@ mod tests {
     use patina::pi::hob::{GUID_EXTENSION, GuidHob, Hob, header};
     use r_efi::efi;
 
-    fn with_locked_state<F: Fn() + std::panic::RefUnwindSafe>(gcd_size: usize, f: F) {
+    enum GcdInit {
+        /// Initializes a simple test GCD (via init_test_gcd()) with the given size.
+        WithSize(usize),
+        /// Initializes a GCD with the given HOB list size (via build_test_hob_list()).
+        WithHobList(usize),
+    }
+
+    /// Initializes global state with either a test GCD or a HOB list, then runs the given
+    /// closure `f` with the physical HOB list pointer (or null if a test GCD was used).
+    ///
+    /// Cleans up global state after `f` returns.
+    fn with_locked_state<F: Fn(*const c_void) + std::panic::RefUnwindSafe>(gcd_init: GcdInit, f: F) {
         test_support::with_global_lock(|| {
-            // SAFETY: multiple functions modify global state. Functions are
-            // called within a global lock to ensure exclusive access during
-            // initialization.
-            unsafe {
-                test_support::init_test_logger();
-                test_support::init_test_gcd(Some(gcd_size));
-                test_support::init_test_protocol_db();
-                test_support::reset_allocators();
-            }
-            f();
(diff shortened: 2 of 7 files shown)


core/patina_internal_cpu/src/interrupts/exception_handling.rs
```diff
@@ -19,7 +19,7 @@ use super::{EfiSystemContextFactory, ExceptionContext, ExceptionType, HandlerTyp
 const NUM_EXCEPTION_TYPES: ExceptionType = if cfg!(test) || cfg!(target_arch = "x86_64") {
     256
 } else if cfg!(target_arch = "aarch64") {
-    3
+    4
```
Javagedes: The name of this const is a bit of a misnomer, though I'm not sure if it is worth renaming or reducing to 255 / 3 and accounting for that where it is used. @os-d @vineelko opinions?

makubacki: As simple as this is, I suggest there be a test to verify the expected number of types per architecture. We should not have this value change without a test failing that verifies the expectation.

cfernald: How is it a misnomer? It represents the number of potential exception types for a given architecture and isn't used as a max.

Javagedes: You are right. I misread the table as only 3 entries. Closing. Thanks.

core/patina_internal_cpu/src/interrupts/exception_handling.rs
```diff
@@ -15,11 +15,17 @@ use crate::interrupts::EfiExceptionStackTrace;
 
 use super::{EfiSystemContextFactory, ExceptionContext, ExceptionType, HandlerType};
 
+// Architecture-specific exception type counts.
+// x86_64: 256 IDT entries
+// AArch64: 4 types (Synchronous, IRQ, FIQ, SError) - see exception_handler.asm:109-112
```
makubacki: I think the constant part of the comment is sufficient. The file and line details may become stale over time.

```suggestion
// AArch64: 4 types (Synchronous, IRQ, FIQ, SError)
```

core/patina_internal_cpu/src/interrupts/exception_handling.rs
makubacki: Thanks for adding the `test_num_exception_types()` unit test. Resolving this thread.

Issue #1148, opened 2025-12-03
[Feature]: Evaluate Device Path Protocol Organization


Javagedes:
### Feature Overview

Many definitions related to the Device Path protocol are spread across the repository. This includes (but may not be limited to) multiple Device path types defined in `fv.rs`, `Pathbuf` parsing functionality in `patina/uefi_protocol/device _path.rs` and additional device_path parsing as an entire utility crate via `patina_internal_device_path`.

On top of this, the `patina_performance` component crate is using the `patina_internal_device_path` crate when `patina_internal_*` crates are only to be used by `patina_dxe_core` or it's subsidiary crates. The fact that `patina_performance` is using this crate indicates that it probably should not be an "internal" crate and may be better in the SDK.

### Solution Overview

An evaluation and possible consolidation of the Device Path functionality needs to be performed to clean up the the code that has grown over time into a tangled mess.

### Urgency

Medium

### Are you going to implement the feature request?

Someone else needs to implement the feature

### Do you need maintainer feedback?

No maintainer feedback needed

PR #1269, opened 2026-01-28
[RFC - FCP]: Reorganize Device Path Libraries

berlin-with0ut-return:
## Description
See RFC for more details. Addresses https://github.com/OpenDevicePartnership/patina/issues/1148 by consolidating and reorganizing Device Path functionality.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [X] Includes documentation?

## How This Was Tested
N/A.

## Integration Instructions
N/A.


docs/src/rfc/text/0024-device-path.md
```diff
@@ -0,0 +1,123 @@
...
+- Parsing functionality split across multiple crates (e.g. `patina_internal_device_path`, `uefi_protocol`)
+- The usage of crates labeled `internal` by external crates (e.g. `patina_performance` relies on `patina_internal_device_path`)
+
+This RFC presents a strategy for integrating all Device-Path-related functionality into a single crate as part of the Patina SDK.
+
+## Technology Background
+
+For basic information on the Device Path Protocol, see the [UEFI Spec Vol 10](https://uefi.org/specs/UEFI/2.10/10_Protocols_Device_Path_Protocol.html). This RFC focuses more on its organization within Patina than specifics of Deivce Path functionality.
+
+The Patina SDK provides shared primitives used in the rest of the Patina core. For more information, see the [README](/sdk/patina/README.md). As many core implementations across several crates depend on Device Path functionality, `sdk` is the best consolidated location to place these struct definitions and parsing functions.
```
os-d:
```suggestion
The Patina SDK provides shared primitives used in the rest of the Patina core and in external callers. For more information, see the [README](/sdk/patina/README.md). As many core implementations across several crates depend on Device Path functionality, `sdk` is the best consolidated location to place these struct definitions and parsing functions.
```

docs/src/rfc/text/0024-device-path.md
```diff
@@ -0,0 +1,123 @@
...
+## Goals
+
+1. Consolidate all Device Path functionality into the Patina SDK crate
+2. Clearly delineate internal vs. external functionality within the SDK
+
+## Requirements
+
+1. Organize all Device Path struct definitions into a `device_path` module inside `sdk`
+2. Organize all Device Path parsing functions into a `device_path` module inside `sdk`
+3. Preserve internal `core` Device Path functionality, ensuring it is not used outside `core`
```
os-d: What would be the internal functionality?

docs/src/rfc/text/0024-device-path.md
```diff
@@ -0,0 +1,123 @@
...
+            ├── node.rs
+            ├── traversal.rs
+```
+
+`node.rs` includes all Device Path struct variations, such as `MediaFwVolDevicePath`. `traversal.rs` includes all Device Path traversal functionality, such as `DevicePathWalker`.
+
+The crates that use `patina_internal_device_path` will have to use `patina` as a dependency instead. This includes `patina_performance` and `patina_dxe_core`. This should not present any issues as both these crates already reference other functionality inside `patina`.
+
+### Node Construction Improvements
+By putting functionality into the Patina SDK, we are also able to control usage and initialization of Device Path structs. A major improvement to the current scheme of initializing raw fields in Device Path nodes would be to fix internal fields and only allow crates to construct nodes through `pub` functions.
```
joschock: I support the spirit of "only allow construct thru pub functions" - but there will still need to be a requirement to construct device path instances from raw byte buffers which kinda lets you get around that. I think that should be managed with an "unsafe: caller guarantees this buffer is a device path".

makubacki:
@berlin-with0ut-return, can you please specify the RFC state in the PR description? Right now, it should be "draft" since it hasn't gone to "FCP" yet.

docs/src/rfc/text/0024-device-path.md
```diff
@@ -0,0 +1,123 @@
...
+- `patina_performance`
+
+## Alternatives
+
+1. Keep as is.
+2. Put Device Path functionality somewhere other than `sdk`, such as its own crate.
+
+## Rust Code Design
+
+All device path functionality will be moved to the Patina SDK.
```
makubacki, line 80: This RFC seems to indicate all of the device path functionality it is consolidating nicely fits together into a `device_path` module. Is that true? Are there no conflicts that need design tradeoffs to be evaluated or anything like that?

docs/src/rfc/text/0024-device-path.md
berlin-with0ut-return: as of now i don't think there is any as all of the functionality inside patina_internal_device_path is generic to device path functionality in general. i'll amend this point

docs/src/rfc/text/0024-device-path.md
berlin-with0ut-return: the `DevicePathNode::new()` function can serve this purpose as it already takes in a raw pointer and is marked unsafe

docs/src/rfc/text/0024-device-path.md
berlin-with0ut-return: i've done a test run moving parts of fv.rs and patina_internal_device_path into the sdk. i don't forsee any major drawbacks based on these initial experiments

PR #1272, opened 2026-01-29
[feature/patina-boot] patina_boot: Add hotkey detection and alternate boot options

### LINKED ISSUE #1228: [Task] patina_boot: Implement hotkey detection for boot override
## Description

Implement hotkey detection in `BootOrchestrator` for boot override functionality (e.g., F12 for alternate boot).

## Requirements

- Detect platform-configured hotkey during boot
- If hotkey pressed, use alternate boot options from `Config<BootOptions>`
- Requires console input to be available

## Current State

`Config<BootOptions>` has a `hotkey` field that is stored but not acted upon. The field is documented as "reserved for future boot menu functionality."

## Related

kat-perez:
## Description

Add hotkey detection support to the boot orchestrator, allowing platforms to configure
alternate boot options that are used when a hotkey (e.g., F12) is pressed during boot.

Changes:
- Add `detect_hotkey()` helper function to check for hotkey press via SimpleTextInput protocol
- Add `hotkey_devices` field and `with_hotkey_device()` builder to `BootOptions`
- Update `BootOrchestrator` to use alternate boot options when hotkey is detected

---

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- Unit tests for `detect_hotkey()` (no input handles case)
- Unit tests for `hotkey_devices` config (single device, multiple devices, combined with hotkey)
- `cargo test -p patina_boot` passes

## Integration Instructions

Platforms can configure hotkey boot options:

```rust
BootOptions::new()
    .with_device(primary_device)
    .with_hotkey(0x0C) // F12 scancode
    .with_hotkey_device(alternate_device)
```

Closes #1228

### DIFF
--- a/components/patina_boot/src/component/boot_orchestrator.rs
+++ b/components/patina_boot/src/component/boot_orchestrator.rs
@@ -15,8 +15,10 @@
 //!
 //! SPDX-License-Identifier: Apache-2.0
 //!
+extern crate alloc;
+
 use patina::{
-    boot_services::{BootServices, StandardBootServices, protocol_handler::HandleSearchType},
+    boot_services::StandardBootServices,
     component::{
         component,
         params::{Config, Handle},
@@ -24,7 +26,6 @@ use patina::{
     error::{EfiError, Result},
     runtime_services::StandardRuntimeServices,
 };
-use r_efi::efi;
 
 use crate::{config::BootOptions, helpers};
 
@@ -47,9 +48,10 @@ impl BootOrchestrator {
     /// 1. Connect all controllers for device enumeration
     /// 2. Signal EndOfDxe (security components perform lockdown)
     /// 3. Discover consoles
-    /// 4. Signal ReadyToBoot
-    /// 5. Execute boot options from config
-    /// 6. If all boot options fail, call failure handler
+    /// 4. Check for hotkey press; if detected, use alternate boot options
+    /// 5. Signal ReadyToBoot
+    /// 6. Execute boot options from config (or hotkey_devices if hotkey detected)
+    /// 7. If all boot options fail, call failure handler
     #[coverage(off)] // Component integration - tested via integration tests
     fn entry_point(
         self,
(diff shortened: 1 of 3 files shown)


components/patina_boot/src/config.rs
```diff
@@ -53,15 +57,25 @@ impl BootOptions {
...
         self.hotkey = Some(scancode);
         self
     }
 
+    /// Add an alternate boot device path used when the hotkey is detected.
+    ///
+    /// Hotkey devices are tried in the order they are added, but only when
+    /// the configured hotkey is detected during boot.
+    pub fn with_hotkey_device(mut self, device: DevicePathBuf) -> Self {
+        self.hotkey_devices.push(device);
```
os-d, line 79: This has the same issue Aaron brought up with platforms wanting to not provide an entire device path for this. I think it is okay to start here, but the search functionality would be useful.

components/patina_boot/src/helpers.rs
```diff
@@ -22,11 +22,80 @@ use patina::{
...
+    unsafe { detect_hotkey_from_handles(boot_services, &handles, hotkey_scancode) }
+}
+
+/// Inner hotkey detection loop over handles.
+///
+/// This function is separated from `detect_hotkey` because it uses raw protocol
+/// function pointers that cannot be unit tested with mocks. Integration tests
+/// verify this code path on real hardware/emulators.
+#[coverage(off)] // Uses raw protocol function pointers - tested via integration tests
+unsafe fn detect_hotkey_from_handles<B: BootServices>(
```
os-d, line 68: SAFETY comment for unsafe function. Then you can update the safety comments in the function to reference the expectations at the function level

components/patina_boot/src/helpers.rs
```diff
@@ -22,11 +22,80 @@ use patina::{
...
+        // Read any pending keystrokes (non-blocking)
+        // The protocol will return NOT_READY if no key is available
+        loop {
+            let mut key = simple_text_input::InputKey::default();
+            // protocol_ptr is valid from handle_protocol, key is a valid output buffer
+            let status = (protocol_ptr.read_key_stroke)(protocol_ptr, &mut key);
+
+            if status == efi::Status::SUCCESS {
+                if key.scan_code == hotkey_scancode {
+                    log::info!("Hotkey detected: scancode 0x{:04X}", hotkey_scancode);
```
os-d: You can remove this, you log in the outer function that the hotkey was detected

components/patina_boot/src/component/boot_orchestrator.rs
makubacki: I know this code was not directly changed in this PR, but are you sure just falling back to the first handle with a loaded image protocol installed is appropriate? I don't think this code should make assumptions about the ordering of the handle database.

components/patina_boot/src/config.rs
```diff
@@ -53,15 +57,25 @@ impl BootOptions {
         self
     }
 
-    /// Add a hotkey scancode for boot override (not yet implemented).
+    /// Add a hotkey scancode for boot override.
     ///
-    /// This field is reserved for future boot menu functionality.
-    /// Currently, the hotkey is stored but not acted upon.
+    /// When this hotkey is detected during boot, the orchestrator will use
```
makubacki, line 62: I don't think that it is a big deal to drain pending key presses in the hot key check.

But, if you haven't already, can please mention that in a function like this that a platform would directly use? Some other feature might similarly look back at key presses and find the buffer is drained and this documentation could help connect the dots.

I know this is mentioned in `detect_hotkey()`, but it would be more visible to a platform developer here in the config.

makubacki (approved):

components/patina_boot/src/helpers.rs
kat-perez: Done - added `# Safety` section to the function.

components/patina_boot/src/helpers.rs
kat-perez: Done - removed the duplicate log.

components/patina_boot/src/config.rs
kat-perez: Done - added note to `with_hotkey` docs about keyboard buffer being drained.

components/patina_boot/src/component/boot_orchestrator.rs
kat-perez: Fixed - now returns `EfiError::InvalidParameter` if ImageHandle is not provided instead of guessing from the handle database.

components/patina_boot/src/config.rs
makubacki: Thanks

components/patina_boot/src/config.rs
kat-perez: Will address partial device path search in a future PR

PR #1273, opened 2026-01-30
Align EDKII PcdPerformanceLibraryPropertyMask with Patina enabled_measurements

liqiqiii:
## Description

Align EDKII PcdPerformanceLibraryPropertyMask with Patina enabled_measurements
Currently there's a mismatch between the PcdPerformanceLibraryPropertyMask and Patina's enabled_measurements.
For example if we set the pcd to 1, then perf component will only log the StartImage.
Add a parser function to align the two.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested with real hardware, before this test with pcd 1 we can get 60 entries, after this we can get 5000 which aligned with using the static patina config.

### DIFF
--- a/components/patina_performance/src/component/performance_config_provider.rs
+++ b/components/patina_performance/src/component/performance_config_provider.rs
@@ -37,7 +37,7 @@ pub struct PerformanceConfigurationProvider;
 /// - BIT3: DISABLE Binding Support logging
 /// - BIT4: DISABLE Binding Start logging
 /// - BIT5: DISABLE Binding Stop logging
-/// - BIT6: DISABLE all other general Perfs
+/// - BIT6: DISABLE all other general Performance Measurements (not used here)
 ///
 /// ## Patina Format (enabled_measurements)
 /// - BIT0 (1): ENABLE StartImage


components/patina_performance/src/component/performance_config_provider.rs
```diff
@@ -73,7 +125,8 @@ impl PerformanceConfigurationProvider {
             log::trace!("The Patina Performance component is disabled per HOB configuration.");
         } else {
             log::trace!("The Patina Performance component is enabled per HOB configuration.");
-            config_mut.enabled_measurements = perf_config_hob.enabled_measurements;
+            // Convert EDK II "disable bits" format to Patina "enable bits" format
+            config_mut.enabled_measurements = convert_edk2_mask_to_patina(perf_config_hob.enabled_measurements);
```
os-d, line 129: We shouldn't directly reference a PCD encoding here. The Patina perf component should read the HOB. Whatever is in the HOB it should be able to understand. If a translation is needed from what the HOB says to what the component wants to use, that's fine, but the translation is from the HOB, not from a PCD. The HOB can be generated from anything.

If the HOB is not-patina specific, then this is true of the C code as well, it needs to be able to read the HOB and parse that, regardless of a PCD. If the HOB is patina specific, then we should not translate values out of it and use what Patina expects.

makubacki: The HOB is Patina-specific and the value in the HOB is expected to be the Patina encoding [[1](https://github.com/OpenDevicePartnership/patina-edk2/blob/4d139422cf8e062e775312a089b395436ee44dc5/PatinaPkg/Include/Guid/PatinaPerformanceConfig.h#L33-L49)]:

```c
#pragma pack(push, 1)
typedef struct {
  ///
  /// Whether the Patina Performance component should be enabled.
  ///
  BOOLEAN    Enabled;
  ///
  /// A mask for selecting measurements enabled in the Patina Performance component.
  ///
  /// See the Patina Performance Component for the most recent configuration details
  /// including measurement bit definitions:
  ///
  /// https://github.com/OpenDevicePartnership/patina/tree/main/components/patina_performance
  ///
  UINT32    EnabledMeasurements;
} PATINA_PERFORMANCE_CONFIG_HOB;
#pragma pack(pop)
```

@liqiqiii, this would need to be updated in the HOB producer code on the platform side.

liqiqiii: sure, will do that. Thanks Michael and Oliver.

liqiqiii:
Per discussion this should be done on platform side, PR closed.

PR #1277, opened 2026-02-03
dxe_core: Ensure Stack Guard Is Enabled When Init'ing Paging

os-d:
## Description

Currently, the init_paging routine doesn't know about the stack guard, which is set up earlier. It ends up wiping this.

This commit moves the stack guard setup to init_paging to take it into account.

It was considered to make this more generic, to take already set attributes, but the number of cases where we expect access attributes to be set should just be the null page and stack guard page, so handle them explicitly.

- [x] Impacts functionality?
- [x] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on Q35 and SBSA with the memory protections test failing and then succeeding.

## Integration Instructions

N/A.


patina_dxe_core/src/allocator.rs
makubacki, line 1: `process_hob_allocations_should_handle_stack_attribute_set_failure()` needs to be moved or removed to account for this change.

makubacki (approved):

patina_dxe_core/src/allocator.rs
os-d: Fixed that and other tests.

cfernald (approved):

os-d:
@makubacki can we force merge this? It’s actually more code coverage than it was before, but I only got it to 78% patch coverage. I could force the attribute setting to fail on the stack hob, but that’s not useful to test, it’s just a panic.

PR #1285, opened 2026-02-05
patina_dxe_core: Provide DXE Core handle to component storage

kat-perez:
## Description

Expose `set_image_handle()` on ComponentDispatcher to allow setting the parent image handle in component storage. Initialize component storage with `DXE_CORE_HANDLE` during system table initialization.

Components that call `LoadImage()` require a parent image handle parameter. The component storage already has `set_image_handle()` but it was never called, leaving the handle as `None`.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Built and verified on QEMU Q35 platform.

## Integration Instructions

N/A

### DIFF
--- a/patina_dxe_core/src/component_dispatcher.rs
+++ b/patina_dxe_core/src/component_dispatcher.rs
@@ -174,6 +174,16 @@ impl ComponentDispatcher {
         self.storage.set_runtime_services(rs);
     }
 
+    /// Sets the image handle in storage.
+    ///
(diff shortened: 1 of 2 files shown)


kat-perez:
Closing - this change depends on `set_image_handle()` which is only in feature/patina-boot. Reopening against that branch.

PR #1286, opened 2026-02-05
[feature/patina-boot] patina_dxe_core: Provide DXE Core handle to component storage

kat-perez:
## Description

Expose `set_image_handle()` on ComponentDispatcher to allow setting the parent image handle in component storage. Initialize component storage with `DXE_CORE_HANDLE` during system table initialization.

Components that call `LoadImage()` require a parent image handle parameter. The component storage already has `set_image_handle()` but it was never called, leaving the handle as `None`.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Built and verified on QEMU Q35 platform.

## Integration Instructions

N/A

### DIFF
--- a/patina_dxe_core/src/lib.rs
+++ b/patina_dxe_core/src/lib.rs
@@ -501,41 +501,48 @@ impl<P: PlatformInfo> Core<P> {
         // Instantiate system table.
         systemtables::init_system_table();
 
-        let mut st = systemtables::SYSTEM_TABLE.lock();
-        let st = st.as_mut().expect("System Table not initialized!");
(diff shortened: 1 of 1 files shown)


os-d (approved):
This I think we should also bring to main as a separate PR. If you think there will be more churn, we can get it into your branch first, but once things settle, we can bring it over independently.

kat-perez:
@os-d Yeah, these are some changes that I discovered that I needed while developing. I'll pull out the changes that go together that can target main instead of only the feature branch in a future PR

PR #1289, opened 2026-02-05
patina_dxe_core: Prevent loading images for wrong machine type

### LINKED ISSUE #1288: [Bug]: Patina will launch binaries that are the incorrect machine type.

### Current Behavior

I observed, by pure accident, that patina will allow you to launch a binary with a mismatched architecture on a platform. This is a bad experience as well and can lead to undefined behavior. For example, this command succeeds on SBSA but causes an immediate fault, as you would expect.

Image

### Expected Behavior

The loader should return an error when trying to load an image with a mismatched machine type.

### Steps To Reproduce

1. Launch SBSA with a x64 OS image
2. Launch the boot application.

### Build Environment

```markdown
N/A
```

### Version Information

```text
N/A
```

### Urgency

High

Someone else needs to fix it

No maintainer feedback needed

### Anything else?

_No response_

makubacki:
## Description

Fixes #1288 

Check the machine type in the PE/COFF header against the architecture of the firmware. If they don't match, skip loading the image.

The same check is performed in the EDK II C DXE Core.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- `cargo make all` (unit tests)
- Attempt to load a AARCH64 EFI application in Q35 (X64)

  image

- Image rejected:

```
ERROR - core_load_pe_image failed: AdvancedLogDumper.efi unsupported machine type 0xaa64 (expected 0x8664)
```

## Integration Instructions

- N/A

### DIFF
--- a/patina_dxe_core/src/pi_dispatcher/image.rs
+++ b/patina_dxe_core/src/pi_dispatcher/image.rs
@@ -63,6 +63,14 @@ pub const EFI_IMAGE_SUBSYSTEM_EFI_APPLICATION: u16 = 10;
 pub const EFI_IMAGE_SUBSYSTEM_EFI_BOOT_SERVICE_DRIVER: u16 = 11;
 pub const EFI_IMAGE_SUBSYSTEM_EFI_RUNTIME_DRIVER: u16 = 12;
 
+/// PE/COFF Specification Machine Types
+#[cfg(target_arch = "x86_64")]
+const EXPECTED_IMAGE_MACHINE: u16 = pecoff::IMAGE_MACHINE_TYPE_X64;
+#[cfg(target_arch = "aarch64")]
+const EXPECTED_IMAGE_MACHINE: u16 = pecoff::IMAGE_MACHINE_TYPE_AARCH64;
+#[cfg(not(any(target_arch = "x86_64", target_arch = "aarch64")))]
+compile_error!("Unsupported target_arch for PE/COFF image loading");
+
 pub const ENTRY_POINT_STACK_SIZE: usize = 0x100000;
 
 // Compile time assert to make sure `STACK_ALIGNMENT` (which comes from uefi_corosensei) is never larger than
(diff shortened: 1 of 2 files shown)


makubacki:
I'll push an update to address the fact that image unit tests were operating on x64 test binaries on aarch64 hosts.

patina_dxe_core/src/pi_dispatcher/image.rs
```diff
@@ -63,6 +63,16 @@ pub const EFI_IMAGE_SUBSYSTEM_EFI_APPLICATION: u16 = 10;
 pub const EFI_IMAGE_SUBSYSTEM_EFI_BOOT_SERVICE_DRIVER: u16 = 11;
 pub const EFI_IMAGE_SUBSYSTEM_EFI_RUNTIME_DRIVER: u16 = 12;
 
+/// PE/COFF Specification Machine Types
+#[cfg(target_arch = "x86_64")]
+const EXPECTED_IMAGE_MACHINE: u16 = 0x8664;
+#[cfg(target_arch = "x86")]
```
os-d: We aren't planning on supporting patina on IA32, drop this?

patina_dxe_core/src/pi_dispatcher/image.rs
```diff
@@ -63,6 +63,16 @@ pub const EFI_IMAGE_SUBSYSTEM_EFI_APPLICATION: u16 = 10;
 pub const EFI_IMAGE_SUBSYSTEM_EFI_BOOT_SERVICE_DRIVER: u16 = 11;
 pub const EFI_IMAGE_SUBSYSTEM_EFI_RUNTIME_DRIVER: u16 = 12;
 
+/// PE/COFF Specification Machine Types
+#[cfg(target_arch = "x86_64")]
+const EXPECTED_IMAGE_MACHINE: u16 = 0x8664;
```
cfernald: I think we should probably have a const def for the objective values that get reused in `EXPECTED_IMAGE_MACHINE` and `MISMATCH_MACHINE`. e.g. `IMAGE_MACHINE_TYPE_X64: u16 = 0x8664`

patina_dxe_core/src/pi_dispatcher/image.rs
makubacki: Agree. Dropped it.

makubacki:
> I'll push an update to address the fact that image unit tests were operating on x64 test binaries on aarch64 hosts.

Added another commit that to fix the pre-existing issue of wrong machine type being used for the image tests before this change.

PR #1290, opened 2026-02-05
[feature/patina-boot] patina_boot: Support partial device path expansion

### LINKED ISSUE #1280: [Task] patina_boot: Support partial device path expansion
## Description

Implement support for expanding partial device paths to full device paths during boot option processing.

## Background

UEFI `Boot####` variables can contain partial device paths (e.g., just a hard drive partition GUID without the full PCI path). These partial paths must be expanded/matched against the actual device topology at boot time to find the corresponding full device path.

## Requirements

- Expand partial device paths to full device paths by matching against discovered device topology
- Support common partial path types:
  - Hard drive media device paths (partition GUID/signature matching)
  - File path-only device paths
  - Short-form device paths
- Integrate with `BootOrchestrator` boot flow or provide as library helper function
- Handle cases where partial path matches multiple devices (prioritization strategy)

## Implementation Options

1. **Library helper function** - `expand_partial_device_path()` in the helpers module
2. **Integrated into boot_from_device_path()** - Automatic expansion when attempting boot
3. **Separate component** - `DevicePathResolver` component

## Related

- Task: BootOptionDiscovery component (#1226)

kat-perez:
## Description

Add support for expanding partial (short-form) device paths to full device paths by matching against the device topology.

- Add `is_partial_device_path()` to detect partial paths (start with Media/Messaging nodes instead of Hardware/ACPI)
- Add `expand_device_path()` to find matching full paths by enumerating device handles
- Wire expansion into `boot_from_device_path()` for transparent handling
- Currently supports HardDrive nodes with GPT partition signature matching

This enables booting from Boot#### variables containing partial device paths like `HD(1,GPT,<GUID>)\EFI\BOOT\BOOTX64.EFI`.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- Unit tests for `is_partial_device_path()`, `expand_device_path()`, and signature matching
- `cargo test -p patina_boot` passes (32 tests)
- QEMU Q35 platform test passes

## Integration Instructions

N/A

Closes #1280

### DIFF
--- a/components/patina_boot/src/helpers.rs
+++ b/components/patina_boot/src/helpers.rs
@@ -20,7 +20,7 @@ use patina::{
     error::{EfiError, Result},
     guids::EVENT_GROUP_END_OF_DXE,
     runtime_services::RuntimeServices,
-    uefi_protocol::device_path::DevicePathBuf,
+    uefi_protocol::device_path::{DevicePath, DevicePathBuf, nodes::DevicePathType},
 };
 use r_efi::{efi, protocols::simple_text_input, system::EVENT_GROUP_READY_TO_BOOT};
 
@@ -119,8 +119,15 @@ pub fn boot_from_device_path<B: BootServices>(
     parent_handle: efi::Handle,
     device_path: &DevicePathBuf,
 ) -> Result<()> {
+    // Expand partial device paths to full paths
+    let full_path = if is_partial_device_path(device_path.as_ref()) {
+        expand_device_path(boot_services, device_path.as_ref())?
+    } else {
+        device_path.clone()
+    };
+
     // Load the image
-    let device_path_ptr = device_path.as_ref() as *const _ as *mut efi::protocols::device_path::Protocol;
+    let device_path_ptr = full_path.as_ref() as *const _ as *mut efi::protocols::device_path::Protocol;
     let image_handle = match boot_services.load_image(true, parent_handle, device_path_ptr, None) {
         Ok(handle) => handle,
         Err(status) => {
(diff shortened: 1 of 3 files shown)


components/patina_boot/src/helpers.rs
```diff
@@ -293,17 +306,161 @@ pub fn discover_console_devices<B: BootServices, R: RuntimeServices>(
...
+    // Return unchanged if already a full path
+    if !is_partial_device_path(partial_path) {
+        return Ok(partial_path.into());
+    }
+
+    // Get partial path's first node for matching
+    let partial_first = partial_path.iter().next().ok_or(EfiError::InvalidParameter)?;
+
+    // Enumerate all handles with device path protocol
+    let handles = boot_services
```
apop5: instead of locating all handles and parsing through device paths, I think you might be able to use locate_device_path and have the system return the handle directly https://github.com/OpenDevicePartnership/patina/blob/c7eaf05893f9536358999a0335c7e17153c8cf15/patina_dxe_core/src/protocols.rs#L773

kat-perez:
Updated to use `locate_device_path` as suggested by @apop5. This is more efficient than iterating all handles manually since UEFI will find the handle with the best matching device path prefix.

components/patina_boot/src/helpers.rs
```diff
@@ -119,8 +119,16 @@ pub fn boot_from_device_path<B: BootServices>(
     parent_handle: efi::Handle,
     device_path: &DevicePathBuf,
 ) -> Result<()> {
+    // Expand partial device paths to full paths
+    let full_path = if is_partial_device_path(device_path.as_ref()) {
+        log::info!("Expanding partial device path...");
```
os-d: I would drop this log, useful for developing, but not on a platform

components/patina_boot/src/helpers.rs
```diff
@@ -293,17 +301,127 @@ pub fn discover_console_devices<B: BootServices, R: RuntimeServices>(
...
+
+    // Combine the full path prefix with the remaining partial path.
+    // The remaining path (after the matched portion) needs to be appended.
+    let mut result = DevicePathBuf::from(full_path);
+
+    // SAFETY: device_path_ptr was updated by locate_device_path to point to the remaining path.
+    let remaining_path = unsafe { DevicePath::try_from_ptr(device_path_ptr as *const u8) };
+    if let Ok(remaining) = remaining_path {
+        // Only append if there's a meaningful remaining path (not just EndEntire)
+        if remaining.iter().any(|node| node.header.r#type != 0x7F) {
```
os-d: Use const here for 7F?

components/patina_boot/Cargo.toml
```diff
@@ -15,6 +15,7 @@ log = { workspace = true }
 patina = { workspace = true, features = ["unstable-device-path"] }
 patina_macro = { workspace = true }
 r-efi = { workspace = true }
+scroll = { workspace = true }
```
os-d: Is scroll actually used?

os-d (approved):
Couple minor comments. Should this go into the (soon to be created) device_path crate as generically useful helper functions?

components/patina_boot/src/helpers.rs
kat-perez: Good call, removed.

components/patina_boot/src/helpers.rs
kat-perez: Done, using `DevicePathType::End as u8` to match the rest of the codebase.

components/patina_boot/Cargo.toml
kat-perez: Nope, leftover from an earlier iteration. Removed.

kat-perez:
@os-d Good idea — filed #1299 to track moving these into the device_path crate when it's created.

kat-perez:
@os-d Good idea — filed #1299 to track moving these into the device_path crate when it's created.

components/patina_boot/src/helpers.rs
```diff
@@ -293,17 +300,127 @@ pub fn discover_console_devices<B: BootServices, R: RuntimeServices>(
...
+/// `false` if it's a full device path or empty.
+pub fn is_partial_device_path(device_path: &DevicePath) -> bool {
+    let Some(first_node) = device_path.iter().next() else {
+        return false;
+    };
+
+    // Full paths start with Hardware (1) or ACPI (2) nodes
+    // Partial paths start with Media (4), Messaging (3), or other nodes
+    let node_type = first_node.header.r#type;
+    node_type != DevicePathType::Hardware as u8 && node_type != DevicePathType::Acpi as u8
```
makubacki: Would a `device_path` of only the "end node" return true right now? My understanding is there iterator would still return `Some(EndEntire)`. Should this check for that case to return false?


```suggestion
    node_type != DevicePathType::Hardware as u8 && node_type != DevicePathType::Acpi as u8 && node_type != DevicePathType::End as u8
```

components/patina_boot/src/helpers.rs
```diff
@@ -293,17 +300,127 @@ pub fn discover_console_devices<B: BootServices, R: RuntimeServices>(
...
+/// * `Err(EfiError::InvalidParameter)` - If the partial path is empty
+/// * `Err(EfiError::NotFound)` - If no matching device was found in the topology
+///
+/// # Ambiguous Matches
+///
+/// If multiple devices match the partial path (e.g., two disks with partitions having
+/// the same signature type but different GUIDs wouldn't match, but identical GUIDs would),
+/// the first match is returned and a warning is logged. Callers needing deterministic
+/// behavior for ambiguous cases should use `expand_device_path()` directly and implement
+/// custom prioritization logic.
```
makubacki: I may have missed it but where is the warning logged?

components/patina_boot/src/lib.rs
```diff
@@ -28,3 +28,5 @@
 pub mod component;
 pub mod config;
 pub mod helpers;
+
+pub use helpers::{expand_device_path, is_partial_device_path};
```
makubacki: Given `expand_device_path()` still has some future work, if this weren't targeting the feature branch, it would probably be best not to make it publicly available. Is the plan to finish that within the feature branch?

components/patina_boot/src/helpers.rs
```diff
@@ -293,17 +300,127 @@ pub fn discover_console_devices<B: BootServices, R: RuntimeServices>(
...
+        return Ok(partial_path.into());
+    }
+
+    // Use LocateDevicePath to find the handle with the best matching device path.
+    // This is more efficient than enumerating all handles manually.
+    let mut device_path_ptr =
+        partial_path as *const DevicePath as *const u8 as *mut efi::protocols::device_path::Protocol;
+    // SAFETY: device_path_ptr points to a valid device path from partial_path.
+    let handle =
+        unsafe { boot_services.locate_device_path(&efi::protocols::device_path::PROTOCOL_GUID, &mut device_path_ptr) }
```
makubacki, line 374: Will this algorithm work with a non-mock `locate_device_path()` (not using `expect_locate_device_path()`)? It looks like some of the node advancement logic there might allow the tests to pass but fail in the normal implementation. Is that true?

components/patina_boot/src/helpers.rs
kat-perez: Good catch — an end-only device path shouldn't be treated as partial. Added `DevicePathType::End as u8` to the exclusion check and updated the test.

components/patina_boot/src/helpers.rs
kat-perez: You're right, that was stale from the previous iteration that enumerated all handles. The `locate_device_path` approach returns a single handle so there's no ambiguity handling. Removed the section.

components/patina_boot/src/lib.rs
kat-perez: Good point — removed the re-export. These are only used internally by `boot_from_device_path` so there's no need to expose them in the crate's public API. If they move to a `device_path` crate later (#1299), we can reconsider the public surface then.

components/patina_boot/src/helpers.rs
kat-perez: You're right — `remaining_device_path` does strict prefix matching so a partial path like `HD(...)` won't match a handle with `ACPI/PCI/HD(...)`. The mock masks this. I'll rework to use handle enumeration with node-by-node signature matching instead.

makubacki (approved):

PR #1292, opened 2026-02-06
[REBASE & FF] patina_dxe_core: Fix inconsistent attributes and capabilities on initial memory

cfernald:
## Description

This commit modified the memory block init code to use the memory policy attributes for the initial free memory block rather than explicitly only RP. This ensures that it will have consistent attributes when it is split up into allocated and freed blocks. Without this change those blocks cannot be coalesced leading to fragmented free memory which can cause failures in Windows allocations.

Additionally, this change uses the resource descriptor HOB to get the attributes and capabilities to ensure consistency and accuracy with the rest of the memory.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- [x] SBSA boot to OS
- [x] Q35 boot to OS
- [x] Unit tests

## Integration Instructions

N/A

### DIFF
--- a/patina_dxe_core/src/gcd.rs
+++ b/patina_dxe_core/src/gcd.rs
@@ -425,6 +425,8 @@ pub fn init_gcd(physical_hob_list: *const c_void) {
     let mut free_memory_size: u64 = 0;
     let mut memory_start: u64 = 0;
     let mut memory_end: u64 = 0;
+    let mut free_memory_attributes: u64 = 0;
+    let mut free_memory_capabilities: u64 = 0;
 
     let hob_list = Hob::Handoff(unsafe {
         (physical_hob_list as *const PhaseHandoffInformationTable)
@@ -443,6 +445,30 @@ pub fn init_gcd(physical_hob_list: *const c_void) {
             Hob::Cpu(cpu) => {
                 GCD.init(cpu.size_of_memory_space as u32, cpu.size_of_io_space as u32);
             }
+            Hob::ResourceDescriptorV2(_) | Hob::ResourceDescriptor(_) => {
+                debug_assert!(
+                    free_memory_start != 0,
+                    "The handoff HOB should come before any resource descriptor HOBs."
+                );
+
+                // Check if this is the resource descriptor for the free memory region, so that it can be used to initialize
+                // the GCD. The handoff HOB should always come first, so the free memory should always be found before the
+                // resource descriptor HOB.
+                if free_memory_start != 0
+                    && free_memory_attributes == 0
+                    && let Some((res_desc, cache_attributes)) = parse_resource_descriptor_hob(&hob)
+                    && res_desc.resource_type == hob::EFI_RESOURCE_SYSTEM_MEMORY
+                    && res_desc.physical_start <= free_memory_start
+                    && res_desc.physical_start.saturating_add(res_desc.resource_length)
+                        >= free_memory_start.saturating_add(free_memory_size)
+                {
+                    free_memory_attributes = cache_attributes.unwrap_or(0);
+                    free_memory_capabilities = spin_locked_gcd::get_capabilities(
+                        GcdMemoryType::SystemMemory,
+                        res_desc.resource_attribute as u64,
+                    );
+                }
+            }
             _ => (),
         }
     }
@@ -452,6 +478,8 @@ pub fn init_gcd(physical_hob_list: *const c_void) {
     log::info!("free_memory_start: {free_memory_start:#x?}");
     log::info!("free_memory_size: {free_memory_size:#x?}");
     log::info!("physical_hob_list: {:#x?}", physical_hob_list as u64);
+    log::info!("free_memory_attributes: {free_memory_attributes:#x?}");
+    log::info!("free_memory_capabilities: {free_memory_capabilities:#x?}");
 
     // make sure the PHIT is present and it was reasonable.
     assert!(free_memory_size > 0, "Not enough free memory for DXE core to start");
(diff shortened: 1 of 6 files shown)


patina_dxe_core/src/gcd/spin_locked_gcd.rs
```diff
@@ -411,12 +411,12 @@ impl GCD {
...
         )?;
 
-        // remove the XP and add RP on the remaining free block.
+        // Apply free memory policy on the remaining free block.
         if len > MEMORY_BLOCK_SLICE_SIZE {
             match self.set_memory_space_attributes(
                 base_address + MEMORY_BLOCK_SLICE_SIZE,
                 len - MEMORY_BLOCK_SLICE_SIZE,
-                (MemoryAttributes::Writeback | MemoryAttributes::ReadProtect).bits(),
+                MemoryProtectionPolicy::apply_free_memory_policy(efi::MEMORY_WB),
```
os-d: We should use this here: https://github.com/OpenDevicePartnership/patina/blob/8cb599725ce9f2b11dd481e7d261f4559684eced/sdk/patina/src/base.rs#L148.

Or, better get it from the HOB if we have it. Same with the calls above. I mean, realistically, system memory will always be WB, but still.

makubacki (approved):

makubacki:
@cfernald, this is currently marked "non-functional", I think that should be checked for this change.

patina_dxe_core/src/gcd/spin_locked_gcd.rs
cfernald: Yeah, even the capabilities passed in are hard coded, I'll add some logic to find the init memory descriptor and use those.

cfernald:
Re-requesting review since there is significant changes.

Javagedes:
I didn't any use case where the new field was not `WB`. Is it worth / necessary to add the field to the function over just always setting it `WB`? I may have missed a conversation where this was requested so feel free to just point me that way or ignore me :)

patina_dxe_core/src/gcd/spin_locked_gcd.rs
```diff
@@ -386,22 +388,18 @@ impl GCD {
...
         let idx = unsafe { self.add_memory_space(memory_type, base_address, len, capabilities) }?;
 
-        //initialize attributes on the first block to WB + XP
-        match self.set_memory_space_attributes(
-            base_address,
-            len,
-            (MemoryAttributes::Writeback | MemoryAttributes::ExecuteProtect).bits(),
-        ) {
+        // Initialize attributes on the first block to WB + XP
+        match self.set_memory_space_attributes(base_address, len, attributes | efi::MEMORY_XP) {
```
os-d: We should also use the allocated memory policy here

os-d (approved):
I think we should put the hob.rs change in a separate commit, easier to see in the future why that changed.

patina_dxe_core/src/gcd/spin_locked_gcd.rs
cfernald: I almost did, but it requires an instanced policy, which I dont think we have here. We could create one just for this purpose, but I wasn't sure if that was the intended pattern. Thoughts?

os-d: Can't you just do `GCD.memory_protection_policy.apply_allocated_memory_protection_policy()`? That's how it is used elsewhere.

cfernald: Oh, its in the GCD. Will use that.

patina_dxe_core/src/gcd.rs
```diff
@@ -443,7 +445,25 @@ pub fn init_gcd(physical_hob_list: *const c_void) {
             Hob::Cpu(cpu) => {
                 GCD.init(cpu.size_of_memory_space as u32, cpu.size_of_io_space as u32);
             }
-            _ => (),
+            _ => {
+                // Check if this is the resource descriptor for the free memory region, so that is can be used to initialize
```
makubacki:
```suggestion
                // Check if this is the resource descriptor for the free memory region, so that it can be used to initialize
```

patina_dxe_core/src/gcd.rs
```diff
@@ -443,7 +445,25 @@ pub fn init_gcd(physical_hob_list: *const c_void) {
             Hob::Cpu(cpu) => {
                 GCD.init(cpu.size_of_memory_space as u32, cpu.size_of_io_space as u32);
             }
-            _ => (),
+            _ => {
```
joschock: Seems a bit weird to pass all the hobs to parse_resource_descriptor_hob(). Filter at the match level? 
```
Hob::ResourceDescriptorV2(_) | HobResourceDescriptorV1(_) => {
..logic..
},
_ => ()

File: patina_dxe_core/src/gcd.rs [hunk 5]
```diff
@@ -443,7 +445,25 @@ pub fn init_gcd(physical_hob_list: *const c_void) {
...
+                // Check if this is the resource descriptor for the free memory region, so that is can be used to initialize
+                // the GCD. The handoff HOB should always come first, so the free memory should always be found before the
+                // resource descriptor HOB.
+                if free_memory_start != 0
```
joschock, line 457: if the assumption for correct behavior is that PHIT hob must be first, consider a debug assert to check/enforce that?

joschock (approved):

patina_dxe_core/src/gcd.rs
makubacki: I thought the same, but technically that is a pre-existing requirement per PI spec. If it not already verified, I think it should be done in a standalone commit.

cfernald:
> I didn't any use case where the new field was not `WB`. Is it worth / necessary to add the field to the function over just always setting it `WB`? I may have missed a conversation where this was requested so feel free to just point me that way or ignore me :)

Honestly, I was just preserving the existing functionality with the new param. In all cases before it was using WB just hidden from the caller I could add other fields, but unless there is some specific intention I'm not sure it has much testing value.

PR #1294, opened 2026-02-06
dxe_core: Add get_existent_memory_descriptor_for_address() Helper Function

os-d:
## Description

Currently, get_memory_descriptor_for_address() returns any descriptor found, including NonExistent descriptors. Because this function is used by the PI spec GetMemorySpaceDescriptor(), it needs to do that.

However, for the majority of Patina usage, we only want existent descriptors. Most of the cases using it is to apply attributes and we cannot do that on non-existent memory. We also would miss cases where we'd search for something, e.g. the stack region, and just move on if we found any descriptor, not specifically an existent one.

This commit adds a new helper function to return NotFound if get_memory_descriptor_for_address() returns a NonExistent descriptor. Usage across the tree is updated.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Booting Q35 and SBSA to Windows. Observed a unit test not failing when it should have because the stack memory region was not present in the GCD. After this fix the test fails as expected.

## Integration Instructions

N/A.

### DIFF
--- a/patina_dxe_core/src/allocator.rs
+++ b/patina_dxe_core/src/allocator.rs
@@ -926,7 +926,7 @@ fn process_hob_allocations(hob_list: &HobList) {
                 }
 
                 let mut address = desc.memory_base_address;
-                match GCD.get_memory_descriptor_for_address(address) {
+                match GCD.get_existent_memory_descriptor_for_address(address) {
                     // we found the region in the GCD, so we can allocate it
                     Ok(gcd_desc) => {
                         if gcd_desc.base_address == desc.memory_base_address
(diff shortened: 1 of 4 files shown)


os-d:
Changing this, I see now this is the underlying layer for the PI spec GetMemorySpaceDescriptor, which should be able to return NonExistent descs. Will write a wrapper function.

os-d:
> Changing this, I see now this is the underlying layer for the PI spec GetMemorySpaceDescriptor, which should be able to return NonExistent descs. Will write a wrapper function.

Updated with the wrapper fn.

PR #1295, opened 2026-02-06
RFC [FCP]: Propose Downstream Build Options Solution

os-d:
## Description

This adds an RFC to propose a solution to the problem of unifying platforms' toolchain configurations when building Patina.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [x] Includes documentation?

## How This Was Tested

N/A.

## Integration Instructions

N/A.

### DIFF
--- a/docs/src/rfc/text/0025-downstream-build-options.md
+++ b/docs/src/rfc/text/0025-downstream-build-options.md
@@ -1,0 +1,181 @@
+# RFC: Downstream Build Options
+
+This RFC intends to define the process for ensuring Patina crates are built with a standard set of build options,
+from RUSTFLAGS to compiler and linker flags.
+
+## Change Log
+
+- 2026-02-06: Initial RFC created.
+
+## Motivation
+
+Patina is heavily dependent on its build options, from stack cookies, image section size, debug asserts enabled in
+debug builds, etc. However, because Patina crates are not binary crates, we cannot control any build options
+for a platform. This can lead to a lot of noise and problems with having to document everything that Patina needs,
+platforms having issues when they don't use exactly that, and time spent debugging issues that should never have
+occurred in the first place.
+
+To avoid all of this and put platforms in the best position, Patina should do its best to ensure the correct set of
+toolchain options are being used.
+
+## Technology Background
+
+The edk2 [tools_def.template](https://github.com/tianocore/edk2/blob/HEAD/BaseTools/Conf/tools_def.template) is the
+inspiration for the idea of having platforms use a canonical set of build options. edk2 achieves this through a
+custom maintained set of tools that platforms must use to build with.
+
+Information about Rust options can be found [in the Rust docs](https://doc.rust-lang.org/cargo/reference/config.html),
+amongst other places.
+
+## Goals
+
+1. Create an easy to consume toolchain configuration for platforms
+2. Ensure the correct toolchain configuration is used when building Patina
+
+## Requirements
+
+1. Do not maintain custom build tools
+2. Do not allow Patina to be built with incorrect build options
+3. Do not put burden on the platform to figure out the right build options
+
+## Unresolved Questions
+
+- Are any of the alternatives preferable to the proposed approach? Many of these could work, but have cons that have
+  led to a different recommendation.
+- Are there any alternatives not listed that are more attractive?
+- If this approach is taken, should Patina just advertise that `config.toml` should be copied or should it create
+  a template repo for platforms to fork?
+- Should Patina check specific toolchain flags are set/unset in `build.rs` or should it only check the version?
+
+## Prior Art (Existing PI C Implementation)
+
+As noted above, edk2's tools_def.template, as used by its BaseTools, is the prior art. It gives edk2 full control
+(by default, it is overrideable per platform and per module) to what it is built with. It requires extensive custom
+build tools to maintain and is not a separateable concept from edk2's BaseTools.
+
+## Rust Code Design
+
+The proposed solution consists of two parts:
+
+### Configuration
+
+Patina will have a `config.toml` file with standard [profiles](https://doc.rust-lang.org/cargo/reference/profiles.html)
+defined for different Patina uses cases, i.e. `dev`, `release`, and `test`. Custom profiles may be defined in the
+future; this is outside the scope of this RFC. Each profile will define the set of `RUSTFLAGS`, compiler flags,
+linker flags, and any other relevant toolchain options. This will be done in conjunction with `target.\<triple\>`
+sections that are global to all profiles.
+
+Patina will document that platforms must copy this `config.toml` to their platform bin repo to build Patina. This
+puts some work on the platform to do the copy and place it correctly. It also is not guaranteed that platforms will
+follow this step. But this can be easily documented and validated (see below).
+
+### Validation
+
+Patina will have a `build.rs` file for each core (e.g. one for patina_dxe_core, one for a future patina_mm_core, etc.)
+that validates the expected profile version used by the platform. This will not check all toolchain options, but only
+enforce that the version matches what Patina expects. This is intended to be a guide rail, not an absolute requirement;
+platforms may choose to implement non-recommended toolchain options.
+
+`build.rs` files are automatically built when a crate is being built, including as a dependency. This allows us to
+automatically enforce our config version expectation. See the
+[cargo build scripts documentation](https://doc.rust-lang.org/cargo/reference/build-scripts.html) for more details.
+
+An example to ensure that Patina is using config version 5:
+
+```rust
+// config.toml
+[env]
+PATINA_CONFIG_VERSION = "5"
+
+[profile.dev]
+// whatever flags this version sets
+```
+
+```rust
+// build.rs
+use std::{env, process};
+
+fn main() {
+    let version = env::var_os("PATINA_CONFIG_VERSION")
+        .unwrap_or_default();
+
+    if version != "5" {
+        eprintln!("error: Incorrect PATINA_CONFIG_VERSION, expected version "5", got version {version}");
+        eprintln!("Use Patina's latest config.toml.");
+        process::exit(1);
+    }
+
+    // Only rerun this when the rustflags or the config version changes
+    println!("cargo:rerun-if-env-changed=CARGO_ENCODED_RUSTFLAGS");
+    println!("cargo:rerun-if-env-changed=PATINA_CONFIG_VERSION");
+}
+```
+
+This allows us to fail to build if we don't have the proper config version. It is not much overhead because Cargo will
+only rerun this when the `RUSTFLAGS` change or the `PATINA_CONFIG_VERSION` changes.
+
+> **Note:** build.rs is very powerful and flexible. It can be used to set `RUSTFLAGS` as well as check them, but that
+> is not proposed here because it only applies to the crate being built, not dependencies. This is merely for
+> validation, not application.
+
+## Guide-Level Explanation
+
+Patina uses a `config.toml` at `patina/.config/config.toml`. Platforms must copy this file to the same location in
+their platform binary repo, e.g. `cp patina/.config/config.toml plat_bin/.config/config.toml`. This will ensure the
+platform is built with the correct toolchain configuration. Patina enforces a set of toolchain configurations it
+must be built with in `patina/build.rs`. The full list is documented in the
+[Patina docs](https://opendevicepartnership.github.io/patina/).
+
+## Alternatives
+
+### Maintain a Cargo Subcommand
+
+In order to enforce platforms use the correct options, we could maintain our own `Cargo subcommand` or other custom
+build tooling.
+
+Upsides:
+
+- Complete control of what platform sets, as long as they use our tool
+
+Downsides:
+
+- Heavy maintenance burden in owning our own tools
+- Higher cost to entry to learn custom tools
+
+### Maintain Custom Target Triples
+
+The relevant toolchain configuration can all be set in custom targets.
+
+Upsides:
+
+- If Patina only can be built with the custom targets, we completely control toolchain configuration.
+
+Downsides:
+
+- Maintenance burden of maintaining our own targets
+- Not compatible with existing UEFI crates
+
+### Have Submodule or External Dependency
+
+A repo could be set up with a `.config/config.toml` in it and a custom tool could copy that to the platform's `.config`
+directory.
+
+Upsides:
+
+- Could guarantee a platform does the copy of the config file.
+
+Downsides:
+
+- Have to maintain custom script.
+- Introduce submodules/external dependencies to Patina which it currently avoids
+- Complicated, easy to envision it failing under different development scenarios
+
+### Various Non-Working Ideas
+
+Various other ideas were explored that seemed promising but are not able to work in the Cargo environment. Briefly,
+these are:
+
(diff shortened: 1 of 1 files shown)


docs/src/rfc/text/0025-downstream-build-options.md
```diff
@@ -0,0 +1,187 @@
...
+
+## Requirements
+
+1. Do not maintain custom build tools
+2. Do not allow Patina to be built with incorrect build options
+3. Do not put burden on the platform to figure out the right build options
+
+## Unresolved Questions
+
+- Are any of the alternatives preferable to the proposed approach? Many of these could work, but have cons that have
```
Javagedes, line 43: An unresolved question we should ask ourselves is if there is any decent way to detect an outdated config.toml compared to the expected config.toml for any given patina version, and any side-effects of that, if we are able to successfully accomplish that.

Javagedes: And if we even want to worry about that, since we care more about the flags than the config toml file itself.

Javagedes (approved):

docs/src/rfc/text/0000-downstream-build-options.md
```diff
@@ -0,0 +1,187 @@
...
+> **Note:** build.rs is very powerful and flexible. It can be used to set `RUSTFLAGS` as well as check them, but that
+> is not proposed here because it only applies to the crate being built, not dependencies. This is merely for
+> validation, not application.
+
+## Guide-Level Explanation
+
+Patina uses a `config.toml` at `patina/.config/config.toml`. Platforms must copy this file to the same location in
+their platform binary repo, e.g. `cp patina/.config/config.toml plat_bin/.config/config.toml`. This will ensure the
+platform is built with the correct toolchain configuration. Patina enforces a set of toolchain configurations it
+must be built with in `patina/build.rs`. The full list is documented in the Patina docs.
```
makubacki: Can you link to the docs?

docs/src/rfc/text/0025-downstream-build-options.md
```diff
@@ -0,0 +1,187 @@
...
+- Introduce submodules/external dependencies to Patina which it currently avoids
+- Complicated, easy to envision it failing under different development scenarios
+
+### Various Non-Working Ideas
+
+Various other ideas were explored that seemed promising but are not able to work in the Cargo environment. Briefly,
+these are:
+
+- Creating Patina profiles that can be enforced
+- Having a `config.toml` that lives in Patina and can apply to consumers without copying
```
makubacki, line 180: Since these are noted, for completeness, can add the reasons why these were rejected?

docs/src/rfc/text/0000-downstream-build-options.md
```diff
@@ -0,0 +1,187 @@
...
+
+Patina will document that platforms must copy this `config.toml` to their platform bin repo to build Patina. This
+puts some work on the platform to do the copy and place it correctly. It also is not guaranteed that platforms will
+follow this step. But this can be easily documented and validated (see below).
+
+### Validation
+
+Patina will have a `build.rs` file for each core (e.g. one for patina_dxe_core, one for a future patina_mm_core, etc.)
+that validates the expected toolchain parameters are being used to build Patina. This will check each and every
+required option, not just that the profile is being used, because the profile does not have versioning, so we won't
```
cfernald: The profile itself wont have versioning, but doesn't the config.toml allow setting environment variables? Could a simple version be set there to keep build.rs simple/consistent?

docs/src/rfc/text/0025-downstream-build-options.md
os-d: I think that we will catch that in build.rs and it will fail, saying hey, you don’t have whatever option set.

We may want to also do the inverse, if we used to set flag X and now don’t, we should check it isn’t set. But, also see response to Chris below.

os-d:
> I think this is a reasonable starting point. Do you plan to prototype this while the PR is open? I think that might lead to some additional points to cover in the RFC.

Yeah, that was my plan, agreed that we’ll probably learn some more while implementing it.

docs/src/rfc/text/0000-downstream-build-options.md
os-d: That was a future looking statement, as if I was writing the doc for it. Currently we don’t have the build flags defined (as a full set anyway).

docs/src/rfc/text/0000-downstream-build-options.md
os-d: That’s where I started with the proposal, in talking with Joey I went to the individual flag approach. I am open to attempting to keep a version, instead of env vars you could also go by profile name, i.e. embed the version in the name, patina-dev-v1.

I am partial to the versioning to keep build.rs simple, we would have to decide if we are doing semver, are there some flags it is acceptable to not break the build for, etc. I would start off by saying no, just a single version and you have to match.

One downside of going off env var or name is that of course can just be set by a platform without any of our flags. I.e. if it fails and they see, oh, it is looking for patina-dev-v2, create a profile with that name. But, I think that can be chalked up to misconfiguration and easy to solve and preventable with good error messages.

cfernald: Yeah, I'm not too concerned with folks circumventing the checks. We can have clear comments and warnings from build.rs on the proper way to address this. I think that would be sufficient for folks to do the right thing, unless choosing not to intentionally for some reason.

For that same reason, I am not concerned about env variables, they can be set in the environment sure, but the force option will override that for cargo processes and if they are setting it just to circumvent the build.rs check then that is their prerogative.

os-d: I agree. I will change this to a version check, with just a single version number incremented for every change, unless @makubacki or @Javagedes has different thoughts.

Javagedes: I'm fine with a simple version

makubacki: Sounds fine to me too.

docs/src/rfc/text/0025-downstream-build-options.md
os-d: Added to unresolved q's.

docs/src/rfc/text/0000-downstream-build-options.md
os-d: Added a link to the main docs.

docs/src/rfc/text/0000-downstream-build-options.md
```diff
@@ -0,0 +1,176 @@
...
+that validates the expected profile version used by the platform. This will not check all toolchain options, but only
+enforce that the version matches what Patina expects. This is intended to be a guide rail, not an absolute requirement;
+platforms may choose to implement non-recommended toolchain options.
+
+An example to ensure that Patina is using config version 5:
+
+```rust
+// config.toml
+[env]
+PATINA_CONFIG_VERSION = 5
```
cfernald: Not critical for the RFC, but `{ value = 5, force = true }` could be done to avoid confusing scenarios or persist variables.

docs/src/rfc/text/0000-downstream-build-options.md
```diff
@@ -0,0 +1,176 @@
...
+
+```rust
+// build.rs
+use std::{env, process};
+
+fn main() {
+    let version = env::var_os("PATINA_CONFIG_VERSION")
+        .unwrap_or_default();
+
+    if version != 5 {
```
cfernald: Doesn't have to be solved here, but I wonder if we will want two versions: a major breaking a minor. Where major_version == expected && minor_version >= expected. To avoid forcing consumers to have _exactly_ the right version in scenarios where we add things that should not break compilation on slightly older versions of Patina.

cfernald (approved):

docs/src/rfc/text/0000-downstream-build-options.md
os-d: I thought about adding that here, but I thought it could set the version when something else is being used, too. If you had another config.toml or something. I was thinking that we would not set our version if something else is setting it, that should be fixed.

docs/src/rfc/text/0000-downstream-build-options.md
os-d: Yeah, I also was debating if we want to do minor versions. I was thinking to leave it as a single version, currently, to keep it simple. This is the set of rules to use. Use this. We have a new rule. Use that. To a certain extent, edk2 does that, if your version is different at all, it sleeps 30 seconds to encourage you to update. Now, they do change flags without changing the version.

I would propose whenever something is changed, change the version. Then one version means one set of flags.

For the minor, I can image scenarios where we'd change a flag and it isn't required to be set, e.g. optimizing for size, but the whole point of this is to centralize on matching what Patina needs, expects, and is validated against. I think we will have the least amount of noise having a single version. But I'm not opposed to a minor, that was just my thought process.

docs/src/rfc/text/0000-downstream-build-options.md
cfernald: Well presumably the force one is paired with the actual settings being used, so it seems that should be what build.rs should be enforcing since. That is, better to ignore something that is weird and not a problem then to not be aware there is a problem.

docs/src/rfc/text/0000-downstream-build-options.md
cfernald: The "it sleeps 30 seconds to encourage you to update" doesn't work so well here IMO because it is persistent in the repo, so a project may have an issue for a long time and not notice because nobody noticed 30 seconds in pipelines.

The purpose of the minor IMO would be to allow a newer version of the file. That is, if we just added some improvements to out optimization in top of tree, a project should be able to use that even if they are using a patina release version that doesnt know that minor version yet.

os-d: I agree the sleep isn't a good pattern, just saying that's edk2's version of having a single version.

I agree with the use case for the minor version, my thought process is just that it is simpler to have only a single version and it avoids the complications of having to decide if something is a major or minor update. That sounds trivial, but with toolchain flags it can be hard to reason and we may not realize that something in patina that was changed since we did minor version X may not behave well in minor version W, or whatever.

So, my point is, I'm open to having a minor version if we think it is useful, but I would honestly prefer to have a single version and say, this is what this version of Patina has been built with. If you want guaranteed success, use this. Add other flags at your own risk. Again, with this model, a platform can add whatever flags they want, as long as they set the version number.

docs/src/rfc/text/0000-downstream-build-options.md
os-d: I guess I was thinking that not setting force was going to force a consumer to deal with the weirdness.

For example, if they have copied our config.toml, but have another one at the next level and build at that next level, it would pick theirs first and set version 4, e.g., then it would find ours and not set 5 because it was already set.

Of course, the reverse could be true (where they copied it) but that would go against the guidance.

docs/src/rfc/text/0000-downstream-build-options.md
cfernald: We can definitely start there and adapt if the syncing becomes a pain, but interesting to think through. This can be an implementation detail left for later/adaptation. I'll close this thread.

docs/src/rfc/text/0000-downstream-build-options.md
cfernald: The reverse is what scares me, since it would mask an error if the fake one what the correct version, but the real was wrong. But similar to the other thread, I just wanted to think it through some so any decision here can be an implementation detail.

docs/src/rfc/text/0000-downstream-build-options.md
os-d: Yeah, that was my thought process, that basically we would amend this RFC if we decide a minor version is needed.

docs/src/rfc/text/0000-downstream-build-options.md
os-d: Yeah, happy to decide that in the actual PR.

My understanding is that in both cases we would have wrong config. I just think that without `force` you'll catch it in one of the cases, the case that our docs guide you to (copying our config.toml to the root).

image

Any array (like rustflags) will be wrong. The deeper config takes precedence for non-arrays, so in the first case, the platform config would take precedence, in which case we wouldn't want to force our version to be set and would want to fail the version check.

In the second case our config would take precedence, but arrays would still be wrong. That's the case we won't catch in either case, but again is against the guidance.

docs/src/rfc/text/0025-downstream-build-options.md
```diff
@@ -0,0 +1,176 @@
...
+### Configuration
+
+Patina will have a `config.toml` file with [profiles](https://doc.rust-lang.org/cargo/reference/profiles.html) defined
+for different Patina uses cases, e.g. `patina-dev`, `patina-release`, etc. The full set of profiles is outside the
+scope of this RFC. Each profile will define the set of `RUSTFLAGS`, compiler flags, linker flags, and any other
+relevant toolchain options.
+
+Patina will document that platforms must copy this `config.toml` to their platform bin repo to build Patina. This
+puts some work on the platform to do the copy and place it correctly. It also is not guaranteed that platforms will
+follow this step. But this can be easily documented and validated (see below).
```
makubacki, line 70: Have you explored verifying that the config.toml file is being used in a/the patina library crate?

For example, if config.toml had an entry like:

```toml
check_patina_build_config = true
```

My understanding is that could be propagated in build.rs with something like:

```rust
println!("cargo:rustc-env=PATINA_BUILD_CONFIG={}", toml.contains("check_patina_build_config = true"));
```

Then in the patina code:

```rust
pub fn is_patina_build_config_present() -> bool {
    option_env!("PATINA_BUILD_CONFIG") == Some("true")
}
```

Then, we could log a message or something:

```rust
if !is_patina_build_config_present() {
  log::info!("Consider using the Patina build config at...");
}
```

Perhaps with an option to disable the banner if the config is not used.

---

Not sure if all that would work. I was just looking through https://doc.rust-lang.org/cargo/reference/build-scripts.html for a path to connect the config.toml presence to patina code from the crates the platform is using.

os-d: I think this is essentially what I am proposing in the Validation section below this. Does that match what you are thinking about or is this proposing something else?

os-d: Oh, you want to wire it into the actual Rust code, not just build.rs?

build.rs can fail the build with a message, which is what I propose below. I think that covers the scenario, can't build with the wrong config, get a message telling you what we expect and how to update it.

Would having it in the code provide something else?

makubacki: I think build.rs is fine. I've used them so little, I didn't realize they would be automatically used from the library crate versus manual action. That appears to be the case based on docs I read. Can you make that connection more obvious around this section?

os-d: Added.

os-d:
Draft PR added here with implementation: https://github.com/OpenDevicePartnership/patina/pull/1324

os-d:
Pushed a couple of small tweaks based on the draft PR

os-d:
Moving to FCP as consensus has been met.

PR #1296, opened 2026-02-07
patina_adv_logger: Add logger info v6 support

makubacki:
## Description

Support using the v6 logger info structure for the advanced logger.

Previously, only v5 was supported, in this change only v6 is supported. A future change may add support for both.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- Verify advanced logger HOB is loaded during initialization
- Dump memory log at EFI shell to verify log contents

## Integration Instructions

- This must be paired with the V6 logger changes in https://github.com/microsoft/mu_plus/pull/826

### DIFF
--- a/components/patina_adv_logger/src/logger.rs
+++ b/components/patina_adv_logger/src/logger.rs
@@ -20,7 +20,7 @@ use patina::{
     serial::SerialIO,
 };
 use r_efi::efi;
-use spin::Once;
+use spin::RwLock;
 
 // Exists for the debugger to find the log buffer.
 #[used]
@@ -35,7 +35,7 @@ where
     target_filters: &'a [(&'a str, log::LevelFilter)],
     max_level: log::LevelFilter,
     format: Format,
-    memory_log: Once<AdvancedLog<'static>>,
+    memory_log: RwLock<Option<AdvancedLog<'static>>>,
     pub(crate) timer: Service<dyn ArchTimerFunctionality>,
 }
 
@@ -58,7 +58,14 @@ where
         max_level: log::LevelFilter,
         hardware_port: S,
     ) -> Self {
-        Self { hardware_port, target_filters, max_level, format, memory_log: Once::new(), timer: Service::new_uninit() }
+        Self {
+            hardware_port,
+            target_filters,
+            max_level,
+            format,
+            memory_log: RwLock::new(None),
+            timer: Service::new_uninit(),
+        }
     }
 
     /// Initializes the performance timer service for timestamping log entries.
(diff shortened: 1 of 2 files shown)


makubacki:
I think it'd be best to support both v5 and v6 to not have adv logger restrict how patina versions can be matched to platforms. I'm leaving this in draft until I add the backward compatibility support for v5.

makubacki:
> I think it'd be best to support both v5 and v6 to not have adv logger restrict how patina versions can be matched to platforms. I'm leaving this in draft until I add the backward compatibility support for v5.

Pushed the update to support both v5 and v6 adv logger info structures. I've verified both on QEMU. I am finishing testing on an Intel physical platform (against both v5 and v6). I'll report back when that is done but we can start reviews now.

components/patina_adv_logger/src/logger.rs
```diff
@@ -126,16 +135,35 @@ where
...
+            if log_guard.as_ref().is_some_and(|log| log.get_address() == address) {
+                return;
+            }
+        }
+
         // SAFETY: The caller must ensure the address is valid for an AdvancedLog.
         if let Some(log) = unsafe { AdvancedLog::adopt_memory_log(address) } {
-            let memory_log = self.memory_log.call_once(|| log);
-            log::info!("Advanced logger buffer initialized. Address = {:#x}", memory_log.get_address());
+            let log_address = log.get_address();
```
os-d: nit: looks like this is just used to print the address, can't we just print `address`? Or can `adopt_memory_log` not use that address in some way?

kuqin12 (approved):

components/patina_adv_logger/src/memory_log.rs
```diff
@@ -326,32 +319,211 @@ impl AdvLoggerInfo {
...
+                return None;
+            }
+
+            (*log_info).version
+        };
+
+        let header_size = Self::header_size_for_version(version)? as u32;
+
+        // SAFETY: Version is confirmed as supported, verify offsets are consistent.
+        if unsafe { (*log_info).log_buffer_offset < header_size } {
```
os-d: Also validate `log_buffer_offset` is < `log_buffer_size`? Validate `log_current_offset` is the same two ways?

components/patina_adv_logger/src/memory_log.rs
```diff
@@ -326,32 +319,211 @@ impl AdvLoggerInfo {
...
+            },
+            new_logger_info_address: 0,
         }
     }
 
     /// Returns the size of the log buffer, which is the size of the header plus
     /// the size of the data buffer.
+    #[allow(dead_code)]
     pub fn full_size(&self) -> u32 {
-        self.log_buffer_offset + self.log_buffer_size
```
os-d, line 354: Not needed anymore since you have it on the AdvLoggerInfoRef?

components/patina_adv_logger/src/logger.rs
makubacki: `address` should be fine. This was solving a problem related to locking and logging in an earlier implementation.

components/patina_adv_logger/src/memory_log.rs
makubacki: Updated validation checks.

components/patina_adv_logger/src/memory_log.rs
makubacki: It could be dropped, thanks.

PR #1303, opened 2026-02-10
Update logger integration check to allow for messages that are not utf8-decodable.

joschock:
## Description

Update logger integration check to allow for messages that are not utf8-decodable.

Test will now pass as long as the expected messages generated by the test are present in the log, in the right order, and those messages (ignoring others) are utf8-decodable. 

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

AdvLogger integration test now passes in the presence of non-UTF8-decodable entries.

## Integration Instructions

N/A


cfernald:
You can probably just throw a coverage disable tag for the integration test file.

PR #1304, opened 2026-02-10
Add Handle component parameter type

kat-perez:
## Description

Add `Handle` component parameter type for accessing the DXE Core image  handle in component entry points (e.g., as parent handle for `LoadImage()`).

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [x] Includes documentation?

## How This Was Tested

`cargo check -p patina --features mockall` passes

## Integration Instructions

Components use `Handle` or `Option<Handle>` as entry point parameters. DXE core must call `storage.set_image_handle(handle)` for it to resolve.

### DIFF
--- a/sdk/patina/src/component/params.rs
+++ b/sdk/patina/src/component/params.rs
@@ -717,6 +717,11 @@ unsafe impl Param for StandardRuntimeServices {
 /// boot applications. Per the UEFI specification, the parent image handle must be a valid
 /// image handle (one that has the LoadedImage protocol installed).
 ///
+/// **Note:** This handle is the DXE Core's image handle, shared across all components. It
+/// should not be used as a `DriverBindingHandle` in `EFI_DRIVER_BINDING_PROTOCOL`, as
+/// multiple components sharing the same agent handle will conflict with protocol open/close
+/// tracking in the UEFI driver model.
+///
 /// ## Example
 ///
 /// ```rust,ignore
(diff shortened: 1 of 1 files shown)


kat-perez:
@joschock @os-d @Javagedes @makubacki Just a note -- this is PR coming from splitting out the changes I made in the feature/patina-boot branch that could target main instead

Javagedes (approved):

sdk/patina/src/component/params.rs
```diff
@@ -711,6 +711,83 @@ unsafe impl Param for StandardRuntimeServices {
     }
 }
 
+/// A UEFI handle that can be used in various UEFI boot service calls.
```
makubacki, line 714: The image handle is often used within the UEFI driver model as the `DriverBindingHandle` in `EFI_DRIVER_BINDING_PROTOCOL`.

This is in turn often used as the "agent handle" when opening protocols. There are specific restrictions around agent handles connected to protocols and the attributes used when opening the protocol.

All that to say, while we don't use the UEFI driver model in components now, I think guidance should be provided here to not use the image handle as the driver binding handle since so many unrelated pieces of code will have the same (DXE Core) handle and that will impact their ability to follow traditional methods of assigning agent handles on those opened protocols.

kat-perez: Gotcha, I'll update the doc comment to be explicit about that usage

PR #1305, opened 2026-02-11
Fix ComponentDispatcher TPL level to TPL_APPLICATION

kat-perez:
## Description

ComponentDispatcher was using TPL_NOTIFY, which prevents component entry points from calling boot services restricted at higher TPL levels. Changed to TPL_APPLICATION.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Existing tests pass. Verified with QEMU Q35 boot.

## Integration Instructions

N/A

### DIFF
--- a/patina_dxe_core/src/component_dispatcher.rs
+++ b/patina_dxe_core/src/component_dispatcher.rs
@@ -120,9 +120,12 @@ unsafe impl Send for ComponentDispatcher {}
 
 impl ComponentDispatcher {
     /// Creates a new locked ComponentDispatcher.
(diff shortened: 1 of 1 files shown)


kat-perez:
@joschock @os-d @Javagedes @makubacki Just a note -- this is PR coming from splitting out the changes I made in the feature/patina-boot branch that could target main instead

os-d (approved):

PR #1307, opened 2026-02-11
Remove resolver = 2

os-d:
## Description

We are currently setting resolver = 2 in most of our Cargo.tomls, however, Rust edition 2024's default resolver version is 3. We appear to have moved to resolver version 2 in the older Rust edition in order to pick up resolver.incompatible-rust-version=fallback support, but this is also the default in resolver version 3. That line will be removed from config.toml in a patina-devops PR.

This drops the explicit resolver downgrade in all Cargo.tomls except for the top level virtual workspace. This must specify the resolver version because it does not have a package section that specifies the Rust edition to get the default from. This is bumped to 3.

See https://doc.rust-lang.org/cargo/reference/resolver.html#resolver-versions for details.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

N/A.

## Integration Instructions

N/A.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,5 @@
 [workspace]
-resolver = "2"
-
+resolver = "3"
 members = ["components/*", "core/*", "sdk/*", "patina_dxe_core"]
 
 [workspace.package]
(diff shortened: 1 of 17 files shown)


cfernald (approved):

os-d:
I have to explicitly set the resolver version in the virtual workspace because it does not have a package section with a Rust version to get the default from. That only affects the top level Cargo.toml. Fixing.

PR #1312, opened 2026-02-12
dxe_core: Fix TPL Inversion in initialize_system_table

### LINKED ISSUE #1311: [Bug]: Patina Crashes in Invalid TPL Raise (0x10 -> 0x4) in Component Dispatcher

### Current Behavior

This occurs on Q35

Image

When e013eee9c4471d13f50c34c901141cb8bc44092e is reverted, the issue does not occur.

### Expected Behavior

We shouldn't panic raising TPL for the component dispatcher.

### Steps To Reproduce

Run Q35 with top of tree Patina.

### Build Environment

```markdown
- OS(s):
- Tool Chain(s):
- Targets Impacted:
```

### Version Information

```text
top of tree
```

### Urgency

High

I will fix it

No maintainer feedback needed

### Anything else?

@kat-perez for reference

os-d:
## Description

Commit e013eee9c4471d13f50c34c901141cb8bc44092e changed the TPLMutex level of the component dispatcher to TPL_APPLICATION instead of TPL_NOTIFY. However, the component dispatcher is invoked when the TPL has been raised to TPL_NOTIFY by the system table lock in patina_dxe_core\lib.rs. This causes a TPL inversion and panics.

This fixes that by dropping the system table lock (lowering the TPL) before attempting to use the component dispatcher.

Fixes #1311 

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested by observing the failure on Q35, applying the fix, then booting to Windows.

## Integration Instructions

N/A.

### DIFF
--- a/patina_dxe_core/src/lib.rs
+++ b/patina_dxe_core/src/lib.rs
@@ -501,8 +501,8 @@ impl<P: PlatformInfo> Core<P> {
         // Instantiate system table.
         systemtables::init_system_table();
 
-        let mut st = systemtables::SYSTEM_TABLE.lock();
-        let st = st.as_mut().expect("System Table not initialized!");
+        let mut st_guard = systemtables::SYSTEM_TABLE.lock();
+        let st = st_guard.as_mut().expect("System Table not initialized!");
 
         allocator::install_memory_services(st);
         gcd::init_paging(self.hob_list());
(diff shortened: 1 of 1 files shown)


makubacki:
I think it'd be a little cleaner to move the code holding the `st_guard` lock to a different function or something so it is naturally dropped versus in the middle of a large function but not a big deal to have the immediate issue resolved.

os-d:
> I think it'd be a little cleaner to move the code holding the `st_guard` lock to a different function or something so it is naturally dropped versus in the middle of a large function but not a big deal to have the immediate issue resolved.

Sorry, I had this on automerge while I was at lunch

PR #1315, opened 2026-02-12
patina_dxe_core: Add debug assert when merge free memory in memory map.

cfernald:
## Description

This commit adds an assert when merging memory descriptors in the memory map for free memory. Conventional memory should already be coalesced and consistent, and if this is not the case then contiguous free memory cannot be allocated contiguously, which can cause issues when allocated memory, especially by address. This is specifically problematic in Windows where almost all allocations are manually selected and allocated by address because it will cause a failure to allocate what looks to be perfectly fine free memory ranges.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Boot to OS on Q35

## Integration Instructions

N/A

### DIFF
--- a/patina_dxe_core/src/gcd/spin_locked_gcd.rs
+++ b/patina_dxe_core/src/gcd/spin_locked_gcd.rs
@@ -1178,6 +1178,14 @@ impl GCD {
                     && prev.attribute == current.attribute
                     && prev.physical_start + (prev.number_of_pages * UEFI_PAGE_SIZE as u64) == current.physical_start
                 {
+                    // Free memory shouldn't even need to be merged because it should already be consistent and coalesced.
+                    // If this fails to be true it can cause odd behavior if applications try to allocate blocks of free
+                    // memory by address, which is a common pattern for OS loaders.
+                    debug_assert!(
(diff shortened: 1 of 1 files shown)


os-d (approved):
I looked at this recently, but just want to double check, the unallocated bin memory will still show as the bin type, right? Or does it show as conventional memory with the image handle not being 0?

cfernald:
> I looked at this recently, but just want to double check, the unallocated bin memory will still show as the bin type, right? Or does it show as conventional memory with the image handle not being 0?

My expectation from the OS side is that it should show up as the bin type. Otherwise, the bin isn't servicing its purpose because it's not hiding the memory variance.

cfernald:
Requesting a force merge, overriding codecov, since the debug_assert is not coverable and i'd rather not add coverage tag spam.

os-d:
> Requesting a force merge, overriding codecov, since the debug_assert is not coverable and i'd rather not add coverage tag spam.

No no, I mean at the memory block layer. But we may already have done that convergence.

Issue #16, opened 2024-10-03
Initial Release - Documentation - Architecture


makubacki:
Architecture description in the main mdbook documentation.

Key areas:

- Architectural objectives
- Components & dispatch
- Operating environment assumption
  - Input data and how it is handled
  - System state that should be configured/expected at DXE Core load

---

Consider whether the below information belongs in the same or a different document.

- Guidance for transitioning from the TianoCore C DXE Core
  - Summarize design delta
  - Simplify platform integration considerations and steps

Javagedes:
Remvoing "Theory of operation" from this issue, as it is coverd by #19

makubacki:
Closing as these have all been documented in Patina documentation - https://github.com/OpenDevicePartnership/patina/blob/main/docs/src/introduction.md.

PR #1317, opened 2026-02-13
StackTrace: AArch64: Add Frame pointer based stack trace support for GCC compiled PE binaries

vineelko:
## Description

Dumps the stack trace by walking the FP/LR registers, without relying on unwind information. This is an AArch64-only fallback mechanism.

For GCC built PE images, .pdata/.xdata sections are not generated, causing stack trace dumping to fail. In this case, we attempt to dump the stack trace using an FP/LR register walk with the following limitations:

1. Patina binaries produced with LLVM almost always do not save FP/LR register pairs as part of the function prologue for non-leaf functions, even though the ABI mandates it. https://github.com/ARM-software/abi-aa/blob/main/aapcs64/aapcs64.rst#646the-frame-pointer

2. Forcing this with the `-C force-frame-pointers=yes` compiler flag can produce strange results. In some cases, instead of saving fp/lr using `stp x29, x30, [sp, #16]!`, it saves lr/fp using `stp x30, x29, [sp, #16]!`, completely breaking the stack walk. https://godbolt.org/z/7s9fG9vWe

3. Due to the above reasons, the stack walk cannot be reliably terminated.

The only reason this is being introduced is to identify the driver/app causing the exception. For example, a Shell app built with GCC that triggers an assertion can still produce a reasonable stack trace.

```text
Dumping stack trace with PC: 000001007AB72ED0, SP: 0000010078885D50, FP: 0000010078885D50
    # Child-SP              Return Address         Call Site
    0 0000010078885D50      000001007AB12770       Shell+66ED0
    1 0000010078885E90      0000010007B98DCC       Shell+6770
    2 0000010078885FF0      0000010007B98E54       qemu_sbsa_dxe_core+18DCC
    3 0000010007FFF4C0      0000010007B98F48       qemu_sbsa_dxe_core+18E54
    4 0000010007FFF800      000001007AF54D08       qemu_sbsa_dxe_core+18F48
    5 0000010007FFFA90      0000010007BAC388       BdsDxe+8D08
    6 0000010007FFFF80      0000000010008878       qemu_sbsa_dxe_core+2C388 --.
                                                                              |
    0:000> u qemu_sbsa_dxe_core!patina_dxe_core::call_bds                     |
    00000000`1002c1b0 f81f0ff3 str x19,[sp,#-0x10]!                           |
    00000000`1002c1b4 f90007fe str lr,[sp,#8]     <---------------------------'
    00000000`1002c1b8 d10183ff sub sp,sp,#0x60

    The FP is not saved, so the return address in frame #6 is garbage.
```
Symbol to source file resolution(Resolving frame 2): Since some modules in the stack trace are built with GCC and do not generate PDB files, their symbols must be resolved manually as shown below.
```text
$ addr2line -e Shell.debug -f -C 0x6770
UefiMain
~/repos/patina-qemu/MU_BASECORE/ShellPkg/Application/Shell/Shell.c:372

   371:  ASSERT (FALSE);
 > 372:  Status = gST->ConOut->ClearScreen (gST->ConOut);
   373:  if (EFI_ERROR (Status)) {
   374:     return (Status);
   375:  }
```

---

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Validated on SBSA Q35.

## Integration Instructions

The change is already hooked into the exception handler for AArch64.

### DIFF
--- a/core/patina_stacktrace/src/stacktrace.rs
+++ b/core/patina_stacktrace/src/stacktrace.rs
@@ -7,6 +7,7 @@ use core::{
 cfg_if::cfg_if! {
     if #[cfg(all(target_os = "uefi", target_arch = "aarch64"))] {
         use crate::aarch64::runtime_function::RuntimeFunction;
+        use crate::byte_reader::read_pointer64;
     } else {
         use crate::x64::runtime_function::RuntimeFunction;
     }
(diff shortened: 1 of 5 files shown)


os-d (approved):

core/patina_stacktrace/src/stacktrace.rs
```diff
@@ -186,6 +187,107 @@ impl StackTrace {
...
+                    log::debug!("{image}");
+
+                    let pc_rva = pc - image.base_address;
+                    let image_name = image.image_name.unwrap_or(no_name);
+
+                    // SAFETY: The assumption here is that the stack frame
+                    // layout is as follows: [fp] [lr] are saved on the stack in
+                    // that order, typically via stp x29, x30, [sp, #-16]!
+                    let prev_pc = unsafe { read_pointer64(fp + 8).unwrap() }; // deref lr
+                    let prev_fp = unsafe { read_pointer64(fp).unwrap() };
```
makubacki: Are these intentionally meant panic instead of propagating the error? My understanding is that this code is likely to already be running in an exception handler so panicking again would not be great.

core/patina_stacktrace/src/stacktrace.rs
```diff
@@ -186,6 +187,107 @@ impl StackTrace {
...
+                    log::warn!("     {i:>2} {:016X}      {:016X}       {image_name}+{pc_rva:X}", fp, prev_pc);
+
+                    pc = prev_pc;
+                    fp = prev_fp;
+
+                    if fp == 0 {
```
makubacki, line 283: Should there be some max number of iterations allowed to break the loop if the frame pointer chain is corrupt?

core/patina_stacktrace/src/stacktrace.rs
```diff
@@ -186,6 +187,107 @@ impl StackTrace {
...
+
+                    // SAFETY: The assumption here is that the stack frame
+                    // layout is as follows: [fp] [lr] are saved on the stack in
+                    // that order, typically via stp x29, x30, [sp, #-16]!
+                    let prev_pc = unsafe { read_pointer64(fp + 8).unwrap() }; // deref lr
+                    let prev_fp = unsafe { read_pointer64(fp).unwrap() };
+                    log::warn!("     {i:>2} {:016X}      {:016X}       {image_name}+{pc_rva:X}", fp, prev_pc);
+
+                    pc = prev_pc;
+                    fp = prev_fp;
```
makubacki, line 281: Is `prev_fp` guaranteed to not be `fp` (`prev_fp` == `fp`)?

makubacki (approved):
Not super familiar with all of the registers involved but reviewed from a language standpoint.

core/patina_stacktrace/src/stacktrace.rs
vineelko: Not intentional. Fixed it to propagate the error.

core/patina_stacktrace/src/stacktrace.rs
vineelko: Not super reliable in my experience in the past, but added a kill switch after 20 iterations.

core/patina_stacktrace/src/stacktrace.rs
vineelko: Added a check for this to bail out.

PR #1319, opened 2026-02-13
Consolidate acpi service

berlin-with0ut-return:
## Description

The old ACPI design included a wrapper service, AcpiTableManager, to enable generic installation of any table type T. This new implementation uses an `impl` on `Service<dyn AcpiProvider>` to avoid having two separate services performing the same underlying task. 

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested
All existing tests pass.

## Integration Instructions
N/A.


berlin-with0ut-return:
@Javagedes i'll leave it up to you if you want to incorporate this change.

Issue #1318, opened 2026-02-13
[Feature]: Core & SDK crates should have `alloc` use gated behind a feature


cfernald:
### Feature Overview

Many core and SDK crates will assume the availability of a global allocator. While this is true for the DXE environment, it may not be true for supervisor or other more-minimal environments. These crates should instead gate such assumptions behind a `alloc` feature which is standard practice for other no_atd crates,

### Solution Overview

Add `alloc` feature for certain crates to prevent general use of alloc not gated by this feature.

### Urgency

Low

### Are you going to implement the feature request?

I will implement the feature

### Do you need maintainer feedback?

No maintainer feedback needed

cfernald:
Initial targets:

patina
patina_adv_logger
patina_internal_cpu (already relatively contained)

PR #1323, opened 2026-02-17
patina_adv_logger: Refactor core logic from component logic

cfernald:
## Description

Currently the adv logger crate exposes all of the component and integration test, etc. by default. However, for consumers wanting to log to the advanced from more minimal environments, this is not ideal. This commit refactors this base support to be the minimal core logic, and then a component feature on top used by DXE.

Base: Exposes the core logger logic, and nothing else. Does no use alloc.

Component: Exposes the component, protocol, and integrations test logic.

To achieve this, this refactors the write and reader into separate modules and simplifies to use a read-only and write-only paradigm to simplify the wrapper logic.

With this change, consumers can use patina_adv_logger without default features to get a minimal advanced logger implementation.

Issue #1318 

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- [x] Q35 w/ integration tests
- [x] Unit tests

## Integration Instructions

N/A

### DIFF
--- a/components/patina_adv_logger/src/lib.rs
+++ b/components/patina_adv_logger/src/lib.rs
@@ -7,14 +7,23 @@
 #![cfg_attr(all(not(feature = "std"), not(doc)), no_std)]
 #![feature(coverage_attribute)]
 
+#[cfg(any(feature = "alloc", test, doc))]
 extern crate alloc;
 
-pub mod component;
+mod memory_log;
+mod writer;
+
 pub mod logger;
+
+#[cfg(any(doc, feature = "reader"))]
+pub mod reader;
+
+#[cfg(any(doc, feature = "component"))]
+pub mod component;
+#[cfg(feature = "component")]
+mod integration_test;
+#[cfg(any(doc, feature = "component"))]
 pub mod protocol;
 
 #[cfg(any(doc, feature = "std"))]
 pub mod parser;
-
-mod integration_test;
-mod memory_log;
--- a/components/patina_adv_logger/src/memory_log.rs
+++ b/components/patina_adv_logger/src/memory_log.rs
@@ -1,7 +1,6 @@
 //! UEFI Advanced Logger Memory Log Support
 //!
-//! This module provides a definitions and routines to access a Advanced Logger
-//! memory log structure.
+//! This module provides definitions for core Advanced Logger memory log structures.
 //!
 //! ## License
 //!
@@ -9,10 +8,13 @@
 //!
 //! SPDX-License-Identifier: Apache-2.0
 //!
+
+// This file contains core structures used by different modules for interacting with the advanced logger. Different feature
+// sets will use different functions. For this reason, allow unused code in this module.
+#![allow(dead_code)]
+
 use core::{
-    cell::UnsafeCell,
     mem::size_of,
-    ptr, slice,
     sync::atomic::{AtomicU32, AtomicU64, Ordering},
 };
 use patina::{
@@ -20,7 +22,6 @@ use patina::{
     error::{EfiError, Result},
 };
 use r_efi::efi;
-use zerocopy::{FromBytes, IntoBytes};
 use zerocopy_derive::*;
 
 // { 0x4d60cfb5, 0xf481, 0x4a98, {0x9c, 0x81, 0xbf, 0xf8, 0x64, 0x60, 0xc4, 0x3e }}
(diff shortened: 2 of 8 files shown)


components/patina_adv_logger/src/reader.rs
```diff
@@ -0,0 +1,310 @@
+//! UEFI Advanced Logger Reader Support
```
os-d: What's the use case for the reader? Tools/tests? In FW, other than tools and the FileLogger, we only ever write, right?

components/patina_adv_logger/src/lib.rs
```diff
@@ -7,14 +7,23 @@
...
 extern crate alloc;
 
-pub mod component;
+mod memory_log;
+mod writer;
+
 pub mod logger;
+
+#[cfg(feature = "reader")]
+pub mod reader;
```
makubacki, line 19: `cargo doc --no-default-features` fails:

```
error[E0432]: unresolved import `crate::reader`                                                                   
  --> components\patina_adv_logger\src\parser.rs:11:12
   |
11 | use crate::reader::AdvancedLogReader;
   |            ^^^^^^ could not find `reader` in the crate root
   |
note: found an item that was configured out
  --> components\patina_adv_logger\src\lib.rs:19:9
   |
18 | #[cfg(feature = "reader")]
   |       ------------------ the item is gated behind the `reader` feature
19 | pub mod reader;
   |         ^^^^^^

error[E0432]: unresolved import `alloc`
  --> components\patina_adv_logger\src\parser.rs:12:5
   |
12 | use alloc::format;
   |     ^^^^^ help: a similar path exists: `std::alloc`
```

The `parser` module depends on `reader::AdvancedLogReader` but `reader` is gated here behind the `reader` feature. This fails without the `reader` feature being enabled. You can change how you like, but this should run successfully.

components/patina_adv_logger/src/memory_log.rs
```diff
@@ -57,252 +58,61 @@ impl<'a> LogEntry<'a> {
...
-
 /// Implementation of the C struct ADVANCED_LOGGER_INFO for tracking in-memory
 /// logging structure for Advanced Logger.
 #[derive(Debug)]
 #[repr(C)]
-pub(crate) struct AdvLoggerInfoV5 {
+pub struct AdvLoggerInfoV5 {
     /// Signature 'ALOG'
-    signature: u32,
+    pub signature: u32,
```
makubacki: I have not fully read through the PR but does every field really need to be public versus `pub(crate)`?

components/patina_adv_logger/src/reader.rs
```diff
@@ -0,0 +1,310 @@
...
+
+        // SAFETY: The buffer was allocated with sufficient size (header_size + TEST_DATA_SIZE).
+        unsafe {
+            ptr::write(buffer.as_mut_ptr().cast::<AdvLoggerInfoV5>(), header);
+        }
+
+        buffer
+    }
+
+    #[allow(dead_code)]
```
makubacki: Is this really considered dead code given it is in the `tests` module and called by tests?

components/patina_adv_logger/src/writer.rs
```diff
@@ -0,0 +1,435 @@
...
+        sync::atomic::{AtomicU32, AtomicU64, Ordering},
+    };
+    use efi::PhysicalAddress;
+
+    use super::*;
+    use crate::memory_log::*;
+
+    const TEST_DATA_SIZE: usize = 128;
+
+    fn create_buffer_v5(timer_frequency: u64, hw_port_disabled: bool) -> Box<[u8]> {
```
makubacki: `create_buffer_v5()` and `create_buffer_v6()` exist in both reader.rs and writer.rs. Can a common test module share code for both?

components/patina_adv_logger/src/reader.rs
cfernald: There are two use cases at the moment. The integration test, which uses it to verify the `log!` and protocol logs are both making it to the memory log, and the parser which is available in the std build, and is used to build the executable parser in this crate.

The reason I'm exposing it is basically as you mentioned, giving the option for object level parsing in external consumers, but it seems appropriate to expose IMO.

components/patina_adv_logger/src/memory_log.rs
cfernald: It's a stylistic choice really. memory_log.rs is not a public module and so and `pub` in this file is still only public to the crate anyways. I'm happy to switch to `pub(crate)` to be more explicit if desired.

components/patina_adv_logger/src/reader.rs
cfernald: Yep, fixed

components/patina_adv_logger/src/lib.rs
cfernald: Good point, fixed by adding all top level conditions check docs to make sure all features appears in docs at all times.

components/patina_adv_logger/src/memory_log.rs
makubacki: I'd prefer the switch to `pub(crate)` for a couple of reasons:

1. It sets a precedent to more tightly restrict access for those that look at the code as "sample code"
2. My understanding is that even if the containing module is private, the `pub` versus `pub(crate)` will still technically allow the item to be re-exported based on its visibility and this would more accurately reflect what should be allowed to be re-exported.

cfernald: That makes sense. I can also just remove most of the pubs completely now since I moved the test init code back to this module.

cfernald: Alright, I have updated this, let me know if you have concerns with the updated verison.

makubacki: Looks reasonable, thanks

PR #1328, opened 2026-02-19
patina: Only include the `alloc` crate when the `alloc` feature is en…

cfernald:
## Description

There already exists an `alloc` feature, but it is not currently used for anything. This commit uses the `alloc` feature to gate the inclusion of the `alloc` crate and the modules that use it.

The current approach here is to just remove all modules that cannot compile as-is without the `alloc` crate and making 'alloc' a dependency for top-level features that require this functionality. In the future better top-level classification of categories of support could be created that may require the alloc feature (e.g. DXE), but this is currently pre-mature so the alloc feature alone will be used for now.

Issue #1318 

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Local build and tests

## Integration Instructions

N/A

### DIFF
--- a/sdk/patina/Cargo.toml
+++ b/sdk/patina/Cargo.toml
@@ -76,7 +76,7 @@ default = ['alloc']
 # with the `#[patina_test]` attribute. Otherwise, a linker crash or failure will
 # occur!
 enable_patina_tests = ["patina_macro/enable_patina_tests"]
-serde = ["dep:serde", "dep:serde_json"]
+serde = ["alloc", "dep:serde", "dep:serde_json"]
 serde-with-yaml = ["serde", "dep:serde_yaml"]
 
 unstable = ["unstable-device-path"]
--- a/sdk/patina/src/lib.rs
+++ b/sdk/patina/src/lib.rs
@@ -40,6 +40,7 @@ pub mod base;
 pub mod boot_services;
 #[cfg(any(test, feature = "alloc"))]
 pub mod component;
+#[cfg(any(test, feature = "alloc"))]
 pub mod device_path;
 #[cfg(any(test, feature = "alloc"))]
 pub mod driver_binding;


sdk/patina/Cargo.toml
makubacki: Should `serde` also be gated?

```suggestion
serde = ["alloc", "dep:serde", "dep:serde_json"]
```

sdk/patina/src/lib.rs
```diff
@@ -35,18 +36,25 @@ pub mod macros;
 
 pub mod arch;
 pub mod base;
+#[cfg(any(test, feature = "alloc"))]
 pub mod boot_services;
+#[cfg(any(test, feature = "alloc"))]
 pub mod component;
 pub mod device_path;
```
makubacki, line 44: For things like `device_path` that use things like `Vec` and `Box`, should it be gated too?

makubacki: How were the currently marked modules identified?

cfernald: Ah, i missed that. No special method to identify, just removed the modules that broken when alloc was removed, but as you pointed out this is missing some uses that were gated behind other features.

os-d (approved):

makubacki (approved):
I think it'd be helpful to put the methodology used to identify what to gate behind `alloc` in the PR/commit description. Simply because if an issue is found in the future, it would make it easier to determine if an alternative approach could identify additional instances missed in this PR.

cfernald:
> I think it'd be helpful to put the methodology used to identify what to gate behind `alloc` in the PR/commit description. Simply because if an issue is found in the future, it would make it easier to determine if an alternative approach could identify additional instances missed in this PR.

Added. I also clarified that this is likely not the ideal end-structure. IMO It would be cleaner to subdivide the SDK into features that reflect their environment. E.g. `DXE`, `MM`, etc, and those would enable `alloc` implicitly. However, I find it too early to be really evaluating that structure so I'm just going to leave these under alloc for now, which is default and transparent to consumers for the time being.

PR #1331, opened 2026-02-20
[REBASE & FF] patina_debugger: Fix use of crate without "alloc" and fix associated issues

cfernald:
## Description

Currently the alloc feature attempts to prevent use of the alloc features but does not actually remove the use of the alloc crate. This means that it is quite easy to accidentally use the alloc crate and break this use case.

Additionally, it makes it too easy to accidentally add heap allocations where they shouldn't be, such as by using `format!` in a monitor command. This feature also helps increase awareness of when heap allocations are used which is important for the debugger which runs in limited environments.

The commit removes use of the `alloc` crate when the `alloc` feature is not enabled, and removes all code leveraging it.

Issue #1318 

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on local build and Q35

## Integration Instructions

N/A

### DIFF
--- a/core/patina_debugger/src/debugger.rs
+++ b/core/patina_debugger/src/debugger.rs
@@ -11,7 +11,10 @@
 //! SPDX-License-Identifier: Apache-2.0
 //!
 
-use core::sync::atomic::{AtomicBool, Ordering};
+use core::{
+    ptr::NonNull,
+    sync::atomic::{AtomicBool, Ordering},
+};
 
 #[cfg(feature = "alloc")]
 use alloc::boxed::Box;
@@ -38,8 +41,18 @@ const GDB_BUFF_LEN: usize = 0x2000;
 const GDB_STOP_PACKET: &[u8] = b"$T05thread:01;#07";
 const GDB_NACK_PACKET: &[u8] = b"-";
 
-#[cfg(not(feature = "alloc"))]
-static GDB_BUFFER: [u8; GDB_BUFF_LEN] = [0; GDB_BUFF_LEN];
+cfg_if::cfg_if! {
+    if #[cfg(not(feature = "alloc"))] {
+        /// Static buffer for GDB communication when the `alloc` feature is not enabled.
+        static GDB_BUFFER: StaticGdbBuffer = StaticGdbBuffer(core::cell::UnsafeCell::new([0; GDB_BUFF_LEN]));
+
+        /// Wrapper for the static struct for mutability.
+        struct StaticGdbBuffer(core::cell::UnsafeCell<[u8; GDB_BUFF_LEN]>);
+
+        /// Safety: The buffer is just memory, the use of which is controlled by the debugger internal state lock.
+        unsafe impl Sync for StaticGdbBuffer {}
+    }
+}
 
 // SAFETY: The exception info is not actually stored globally, but this is needed to satisfy
 // the compiler as it will be a contained within the target struct which the GdbStub
@@ -80,6 +93,9 @@ where
     connection_timed_out: AtomicBool,
 }
 
+/// Safety: Send is safe by default for all but the gdb_buffer, but this is just a raw buffer and is safe to send between threads.
+unsafe impl<T: SerialIO> Send for DebuggerInternal<'static, T> {}
+
 /// Internal Debugger State
 ///
 /// contains the internal configuration and state for the debugger. This will
@@ -90,7 +106,7 @@ where
     T: SerialIO,
 {
     gdb: Option<GdbStubStateMachine<'a, PatinaTarget, SerialConnection<'a, T>>>,
-    gdb_buffer: Option<&'a [u8; GDB_BUFF_LEN]>,
+    gdb_buffer: Option<NonNull<[u8; GDB_BUFF_LEN]>>,
     timer: Option<&'a dyn ArchTimerFunctionality>,
     initial_breakpoint: bool,
 }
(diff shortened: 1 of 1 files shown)


core/patina_debugger/src/dbg_target/monitor.rs
```diff
@@ -140,6 +140,7 @@ impl PatinaTarget {
                 state.modules.clear_module_breakpoints();
                 let _ = out.write_str("Cleared module breaks!");
             }
+            #[cfg(feature = "alloc")]
```
makubacki: The PR is adding around 20 new instances of `#[cfg(feature = "alloc")]`. While the end goal is understandable, it makes it more difficult to reason about `alloc` and non-`alloc` paths.

For example, it looks like all of the tests are not running in the non-`alloc` path right now?

core/patina_debugger/src/system.rs
```diff
@@ -35,25 +43,42 @@ impl SystemState {
...
+    /// Attempts to handle an external monitor command. Returns `true` if the command
     /// was recognized, and `false` if it was not found.
     pub fn handle_monitor_command(
         &self,
-        command: &str,
-        args: &mut core::str::SplitWhitespace<'_>,
-        out: &mut dyn core::fmt::Write,
+        _command: &str,
+        _args: &mut core::str::SplitWhitespace<'_>,
+        _out: &mut dyn core::fmt::Write,
```
makubacki: This introduces some confusion in that the parameters are actually used behind the `alloc` feature. If it could be refactored to make the API a little more cohesive with what's happening in the implementation, I think that'd be clearer about what is actually happening.

core/patina_debugger/src/system.rs
```diff
@@ -8,23 +8,31 @@
 //! SPDX-License-Identifier: Apache-2.0
 //!
 
+#[cfg(feature = "alloc")]
 use alloc::{boxed::Box, string::String, vec::Vec};
 
+#[cfg(feature = "alloc")]
```
makubacki: I think a larger refactor can make it clearer and using `cfg_if!` to group things as needed behind that refactor will make it less cluttered.

Have you considered that? Something like:

```rust
cfg_if::cfg_if! {
    if #[cfg(feature = "alloc")] {
        // alloc mod and re-export
    } else {
        // non-alloc mod and re-export
    }
}
```

Having the same implementations in separate modules would make the behavior differences clearer imo.

cfernald: That is fair. I went back and forth a few times, and I think in the end I over-prioritized trying to avoid some duplication and it ended up becoming harder to read. Let me split some things up more clearly.

core/patina_debugger/src/dbg_target/monitor.rs
cfernald: Addressed this my splitting into modules and having both modules compile on test.

core/patina_debugger/src/system.rs
```diff
@@ -49,40 +65,24 @@ impl SystemState {
...
+        false
     }
 
+    /// Adds a module to the tracked list. Without alloc, this is a no-op.
     pub fn add_module(&mut self, name: &str, base: usize, size: usize) {
         self.modules.push(ModuleInfo { name: String::from(name), base, size });
     }
 
+    /// Checks if a module breakpoint matches the given name.
+    /// Without alloc, only the break-all flag is checked.
```
makubacki: Looks like this might be stale from `system_no_alloc.rs`?

```suggestion
```

core/patina_debugger/src/system_no_alloc.rs
```diff
@@ -0,0 +1,136 @@
...
+//! and monitor callbacks.
+//!
+//! ## License
+//!
+//! Copyright (C) Microsoft Corporation.
+//!
+//! SPDX-License-Identifier: Apache-2.0
+//!
+
+pub(crate) struct SystemState {
```
makubacki, line 13: Does a shared trait for `SystemState` not make sense because there are expected to be API differences (like `add_monitor_command()` only existing in the alloc module) or are there other reasons? That could help catch differences when compiling each individually.

core/patina_debugger/src/lib.rs
```diff
@@ -141,9 +153,10 @@ static DEBUGGER: spin::Once<&dyn Debugger> = spin::Once::new();
 /// For example, if the command is `my_command arg1 arg2`, then `arg1` and `arg2` will
 /// be the first and second elements of the iterator respectively.
 ///
-/// the second argument is a writer that should be used to write the output of the
+/// The second argument is a writer that should be used to write the output of the
 /// command. This can be done by directly invoking the [core::fmt::Write] trait methods
-/// or using the `write!` macro.
+/// or using the `write!` macro. `format!` should be avoided as it will allocate memory
+/// which should shoundn't be done in debugger when possible.
```
makubacki:
```suggestion
/// which shouldn't be done in the debugger when possible.
```

core/patina_debugger/src/system_no_alloc.rs
cfernald: Originally I didn't because I wasn't compiling both generally anyways, so it wouldn't have helped anyways without the compiling both features, but now with the change to compile both in test I suppose that is no longer true. I will add this.

cfernald: I added this. I also ended up moving the generic layer in system.rs and splitting the current system*.rs into submodules, this was just because with the trait it was starting to clutter lib.rs

Javagedes:
@codecov re-run

PR #1333, opened 2026-02-20
[feature/patina-boot] patina_boot: Refactor to trait-based boot orchestration design

kat-perez:
## Description

Refactors `patina_boot` from a monolithic `BootOrchestrator` component + `Config<BootOptions>` pattern to a trait-based design:

- **`BootOrchestrator` trait** — defines the boot flow interface with `execute() -> Result<!, EfiError>`, enforcing at the type level that successful boot never returns
- **`BootDispatcher`** — the Patina component that installs the BDS architectural protocol and delegates to a `BootOrchestrator` implementation
- **`SimpleBootManager`** — a default `BootOrchestrator` implementation for platforms with straightforward boot topologies
- **`BootConfig`** — unified boot configuration (replaces previous `BootOptions` + `SimpleBootConfig` split), requires at least one device at construction (compile-time enforcement)

Also updates `helpers.rs` imports from removed `uefi_protocol::device_path` to `device_path::paths`/`device_path::node_defs`.

`patina_dxe_core` changes (image handle plumbing) split out to #1374.

- [x] Impacts functionality?
- [ ] Impacts security?
- [x] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

1. Unit tests: 35 pass (`cargo test -p patina_boot`)
2. Integration tested on QEMU Q35 — all components dispatched, BDS phase ran, boot options attempted, failure handler fired correctly
3. CI: fmt, clippy, all platforms pass

## Integration Instructions

Update boot orchestration usage from:
```rust
use patina_boot::{component::BootOrchestrator, config::BootOptions};
// In configs():
add.config(BootOptions::new()...);
// In components():
add.component(BootOrchestrator);
```

To:
```rust
use patina_boot::{BootDispatcher, SimpleBootManager, config::BootConfig};

add.component(BootDispatcher::new(SimpleBootManager::new(
    BootConfig::new(primary_device_path())
        .with_device(fallback_device_path())
        .with_hotkey(0x16)
        .with_hotkey_device(usb_device_path())
        .with_failure_handler(|| { /* ... */ }),
)));
```


kat-perez:
@makubacki @os-d @Javagedes @joschock This is ready for review now. Thanks!

components/patina_boot/src/orchestrators.rs
```diff
@@ -0,0 +1,12 @@
+//! Built-in [`BootOrchestrator`](crate::BootOrchestrator) implementations.
```
os-d, line 1: mod.rs is the old hierarchy style, please use the new https://doc.rust-lang.org/reference/items/modules.html#module-source-filenames

components/patina_boot/src/config.rs
```diff
@@ -114,6 +118,102 @@ impl BootOptions {
...
+/// ## Example
+///
+/// ```rust,ignore
+/// use patina_boot::config::SimpleBootConfig;
+///
+/// let config = SimpleBootConfig::new(nvme_device_path, usb_device_path)
+///     .with_hotkey(0x16) // F12
+///     .with_failure_handler(|| show_error_screen());
+/// ```
+pub struct SimpleBootConfig {
```
os-d: I think that there should be config for the BootOrchestrator trait but at the trait level, we shouldn't know anything about the impl. So we should have SimpleBootConfig vs ComplexBootConfig, there should simply be a BootConfig (or whatever). Then SimpleBootManager should validate its config and if possible throw a build error because the config is too large, not structured correctly, whatever.

components/patina_boot/src/orchestrators.rs
kat-perez: Changed from mod.rs -> impls.rs

components/patina_boot/src/config.rs
kat-perez: BootOptions has been changed to BootConfig and SimpleBootConfig deleted now

components/patina_boot/src/boot_orchestrator.rs
```diff
@@ -0,0 +1,67 @@
...
+    ///
+    /// This method should not return under normal operation — a successful boot
+    /// transfers control to the boot image. If all boot options are exhausted,
+    /// the implementation should invoke its failure handler and/or spin forever.
+    fn execute(
+        &self,
+        boot_services: &StandardBootServices,
+        runtime_services: &StandardRuntimeServices,
+        image_handle: efi::Handle,
+    );
```
makubacki: It would be nice to have a little compiler enforcement around the caller/callee. Is a return type like `Result<!, EfiError>` too restrictive?

components/patina_boot/src/boot_dispatcher.rs
```diff
@@ -0,0 +1,173 @@
...
+/// BDS architectural protocol entry point.
+///
+/// Called by the DXE core after all architectural protocols are installed and
+/// all DXE drivers have been dispatched. Retrieves the stored context and
+/// delegates to the orchestrator.
+#[coverage(off)] // Extern "efiapi" callback — tested via integration tests
+extern "efiapi" fn bds_entry_point(_this: *mut bds::Protocol) {
+    let Some(context) = BDS_CONTEXT.get() else {
+        log::error!("BDS context not initialized — BootDispatcher entry_point was not called");
+        loop {
```
os-d: Why spin loop instead of panic? I would convert all spin loops to panics (in the cases we truly can do nothing else) so that the platform panic handler gets invoked, the debugger can be entered, etc.

patina_dxe_core/src/component_dispatcher.rs
```diff
@@ -178,6 +178,13 @@ impl ComponentDispatcher {
         self.storage.set_runtime_services(rs);
     }
 
+    /// Sets the image handle in storage (used as parent for LoadImage calls).
+    #[coverage(off)]
+    #[inline(always)]
```
os-d:
```suggestion
```
I would just let the compiler decide what is best for inlining, except when there is a specific reason we must have it inlined.

patina_dxe_core/src/component_dispatcher.rs
```diff
@@ -178,6 +178,13 @@ impl ComponentDispatcher {
...
+    pub(crate) fn set_image_handle(&mut self, handle: efi::Handle) {
+        self.storage.set_image_handle(handle);
```
os-d: This and the other patina_dxe_core change I would take directly to the main branch (or at a minimum do a separate PR for).

components/patina_boot/src/boot_orchestrator.rs
kat-perez: I think it's fine to return `Result<!, EfiError>`, i've updated this

components/patina_boot/src/boot_dispatcher.rs
kat-perez: I'm fine with that, updated to panic instead

patina_dxe_core/src/component_dispatcher.rs
kat-perez: Sounds good, removed this change and the other patina_dxe_core changes

patina_dxe_core/src/component_dispatcher.rs
kat-perez: Removed it here and will file a separate PR for it

components/patina_boot/src/boot_orchestrator.rs
makubacki: Thanks. I think that best for enforcing the successful case not returning and allowing error handling for unexpected scenarios to be under the ownership of the caller.

components/patina_boot/src/boot_dispatcher.rs
```diff
@@ -0,0 +1,166 @@
...
+// Access is single-threaded (BSP at TPL_APPLICATION during BDS phase).
+unsafe impl Sync for BdsContext {}
+
+/// Static storage for the BDS context. Initialized once during component dispatch,
+/// consumed once when the DXE core invokes the BDS protocol.
+static BDS_CONTEXT: Once<BdsContext> = Once::new();
+
+/// Boot dispatcher component.
+///
+/// This is the single Patina component for boot orchestration. It:
```
makubacki: This sentence is a little confusing given `BootDispatcher ` has a dependency on `BootOrchestrator`.

Was something like this intended?

```suggestion
/// This is the single Patina component for driving boot orchestration. It:
```

components/patina_boot/src/orchestrators.rs
makubacki, line 1: nit: I know you just changed this from `mod.rs` to `impls.rs`. But that seems a little weird to me in the namespace. Would `orchestrators` work?

makubacki (approved):

components/patina_boot/src/boot_dispatcher.rs
kat-perez: sure, I can update that comment

components/patina_boot/src/orchestrators.rs
kat-perez: yeah, I can change this to orchestrators instead

PR #1334, opened 2026-02-23
patina_smbios: Add Type 16, Type 17, and Type 19 SMBIOS record types

kat-perez:
## Description

Add `Type16PhysicalMemoryArray` (Physical Memory Array), `Type17MemoryDevice` (Memory Device), and `Type19MemoryArrayMappedAddress` (Memory Array Mapped Address) structs to the `patina_smbios` crate. Field layouts verified against EDK2 `SMBIOS_TABLE_TYPE16/17/19`, targeting SMBIOS 3.9 (includes 3.7+ PMIC/RCD fields on Type 17).

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Integrated into SBSA `patina-dxe-core-qemu` with a local patch and verified with patina-qemu smbiosview

## Integration Instructions

use `patina_smbios::smbios_record::{Type16PhysicalMemoryArray, Type17MemoryDevice, Type19MemoryArrayMappedAddress};`

### DIFF
--- a/components/patina_smbios/src/smbios_record.rs
+++ b/components/patina_smbios/src/smbios_record.rs
@@ -558,6 +558,179 @@ pub struct Type7CacheInformation {
     pub string_pool: Vec<String>,
 }
 
+/// Type 16: Physical Memory Array
+///
+/// Describes the attributes of a physical memory array, including the number
+/// of memory devices it contains, the maximum capacity, and the error correction type.
+///
+/// # Important: Not C-Compatible
+///
+/// This struct contains a `string_pool: Vec<String>` field which is Rust metadata and
+/// **NOT** part of the SMBIOS table binary format. Never cast this struct to bytes directly.
+/// Always use `to_bytes()` to convert to proper SMBIOS format.
+///
+/// See [`Type0PlatformFirmwareInformation`] for detailed documentation on proper usage.
+#[derive(patina_macro::SmbiosRecord)]
+#[smbios(record_type = 16)]
+pub struct Type16PhysicalMemoryArray {
+    /// SMBIOS table header
+    pub header: SmbiosTableHeader,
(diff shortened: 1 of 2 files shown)


components/patina_smbios/src/smbios_record.rs
```diff
@@ -558,6 +558,146 @@ pub struct Type7CacheInformation {
     pub string_pool: Vec<String>,
 }
 
+/// Type 16: Physical Memory Array
+#[derive(patina_macro::SmbiosRecord)]
```
makubacki, line 573: I noticed the tests manually set `length` values but the derive macro doesn't seem to use those:

https://github.com/OpenDevicePartnership/patina/blob/30bb607c989d63153a84afa3dbbd26ebb2af9db7/sdk/patina_macro/src/smbios_record_macro.rs#L267-L273

Can you clarify when the macro calculated values and the manually set values in the tests that are set and asserted on are used?

kat-perez: Right so header.length values are never used during serialization. `to_bytes()` always writes the length calculated by the derive macro.  The bytes[1] assertion in the `to_bytes` test should verify the macro-calculated value. I'll update the tests to just use 0 for that field and normalize it for Type 3 as well so that every type is following the same convention.

kat-perez: Ended up normalizing all of the other types that also had that inconsistency. Updated now

makubacki: Thanks

kat-perez:
@codecov re-run

kat-perez:
https://github.com/codecov re-run

PR #1339, opened 2026-02-24
Add Patina QEMU PR Validation [Rebase & FF]

makubacki:
## Description

Two commits to add Patina QEMU testing on PRs in this repository.

See https://github.com/OpenDevicePartnership/patina-devops/pull/81 for more details about the workflows themselves.

---

**.github: And Patina QEMU validation workflows**

Adds:

- `.github/workflows/patina-qemu-pr-validation.yml`: Validates the
  patina PR changes on QEMU.
- `.github/workflows/patina-qemu-pr-validation-post.yml`: Posts the
  results from the QEMU run back to the patina PR.

Temporarily sets the patina-devops branch to `patina_e2e_plat_validation`
in OpenDevicePartnership/patina-devops during a "rollout" phase of the
new workflow so any adjustments can be quickly made. Once the workflow
is stable, workflows in the patina-devops release will be referenced
instead.

---

**.github/ci-workflow.yml: Prevent concurrent runs**

Create a concurrency group to enforce only a single CI workflow is run
per PR at a time.

---

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- Various PR scenarios on fork:
  - Successful run
  - Compilation failure
  - Boot failure
  - Cancellation of workflow due to a new PR push
  - Log artifacts uploaded successfully
  - Creation of comment in a new PR
  - Amendment of comments in an existing PR

## Integration Instructions

- N/A - Only used for CI in this repo

### DIFF
--- a/.github/workflows/patina-qemu-pr-validation.yml
+++ b/.github/workflows/patina-qemu-pr-validation.yml
@@ -1,0 +1,72 @@
+# @file patina-qemu-pr-validation.yml
+#
+# A workflow that runs QEMU-based validation after CI succeeds for a pull request.
+#
+# Copyright (c) Microsoft Corporation.
+# SPDX-License-Identifier: Apache-2.0
+##
+name: Patina QEMU PR Validation
+
+on:
+  workflow_run:
+    workflows: ["CI"]
+    types: ["completed"]
+
+# Do not allow more than one instance of this workflow to run at the same
+# time for the same PR branch.
+concurrency:
+  group: qemu-pr-${{ github.event.workflow_run.head_branch }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+  actions: read
+  pull-requests: read
+
+jobs:
+  prepare:
+    runs-on: ubuntu-latest
+    name: Gather Incoming PR Metadata
+    if: |
+      github.event.workflow_run.conclusion == 'success' &&
+      github.event.workflow_run.event == 'pull_request'
+    outputs:
+      pr_number: ${{ steps.pr-info.outputs.pr_number }}
+    steps:
+      - name: Get PR Information
+        id: pr-info
+        shell: bash
+        env:
+          GH_TOKEN: ${{ github.token }}
+          HEAD_SHA: ${{ github.event.workflow_run.head_sha }}
+          HEAD_REF: ${{ github.event.workflow_run.head_branch }}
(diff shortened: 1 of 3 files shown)


makubacki:
The QEMU validation workflow will not run until this PR is merged into `main`.

The branch for workflows related to the Patina QEMU flow are set to [`patina_e2e_plat_validation`](https://github.com/OpenDevicePartnership/patina-devops/tree/patina_e2e_plat_validation) in OpenDevicePartnership/patina-devops. This is being done during a "rollout period" of the new workflow. The scenarios that need to be handled are fairly diverse and complex. This will allow a period of time where it can be tweaked quickly based on actual PR scenarios.

.github/workflows/patina-qemu-pr-validation.yml
```diff
@@ -0,0 +1,78 @@
...
+          import os
+          import pathlib
+
+          event_path = pathlib.Path(os.environ['GITHUB_EVENT_PATH'])
+          payload = json.loads(event_path.read_text(encoding='utf-8'))
+          prs = payload.get('workflow_run', {}).get('pull_requests', [])
+          if not prs:
+              raise SystemExit('No pull request found in the workflow_run payload.')
+
+          pr = prs[0]
```
cfernald: For my education, why is the python needed here? It seems there is `Github.event.workflow_run.pull_requests`, but admittedly it seems this isn't reliable for all use cases so my assumption is it is something to do with that.

makubacki: I think that would work for simply reading that value, but I prefer to:

- Parse JSON from a file ([`GITHUB_EVENT_PATH`](https://docs.github.com/en/actions/reference/workflows-and-actions/variables))
- Print using f-strings
- Append the PR number to the step output using Python libraries

This could have been done in bash or something else. I just preferred using Python here.

makubacki: I had previously tested against PRs from branches on my fork to the default branch on my fork. After testing with a PR from another fork to my fork, I realized that the event payload would not have `pull_requests` available, so I ended up switching to using the GitHub REST API with GitHub CLI using the head branch. The CLI picked up enough of the code that I think bash is simple enough now while passing in default environment variables available.

PR #1347, opened 2026-02-26
patina: Add `With<P: Param>` component parameter

Javagedes:
## Description

Due to the nature of parameters, some parameters conflict with eachother to prevent multiple mutable aliases to the same data. However there are some scenarios (mainly referring to the `Storage` component) that conflict with all other parameters and prevent a component writer from being able to properly control when their component executes.

As an example, a component that consumes a `&mut Storage` cannot use dependency injection to ensure the component does not run before a specific configuration value is done being mutated.

To combat this, the `With` component is added, that allows a component's interface to validate something exists without actually retrieving the data. For the `Storage` scenario, this allows a component to only execute once the underlying param is available, then manually retrieve that parameter from storage.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [x] Includes documentation?

## How This Was Tested

Created unit tests pass.

## Integration Instructions

Developers may begin using the `With` parameter. Example below

```rust
/// prevents the component from executing until the `i32` config is done being updated, and the `MyService` service has been installed.
fn entry_point(&self, storage: &mut Storage, _: With<Config<i32>>) ->Result<()> {
    let cfg = storage.get_config::<i32>>().unwrap();
}
```

### DIFF
--- a/sdk/patina/src/component.rs
+++ b/sdk/patina/src/component.rs
@@ -203,10 +203,11 @@ pub trait IntoComponent<Input> {
 pub mod prelude {
     pub use crate::{
         component::{
-            IntoComponent,
+            IntoComponent, component,
             hob::{FromHob, Hob},
-            params::{Commands, Config, ConfigMut, Handle},
+            params::{Commands, Config, ConfigMut, Handle, With},
             service::{IntoService, Service},
+            storage::Storage,
         },
         error::{EfiError, Result},
     };
--- a/sdk/patina/src/component/params.rs
+++ b/sdk/patina/src/component/params.rs
@@ -793,6 +793,60 @@ unsafe impl Param for Handle {
     }
 }
 
+/// A parameter that specifies `P` must exist for the component to execute. The parameter itself is not retrieved.
+///
+/// This is useful for components who need to verify the presence of a parameter, but cannot actually retrieve it due
+/// to borrowing rules. The most common example of this is a component that needs access to [Storage], which conflicts
+/// with almost any other parameter.
+///
+/// Some common examples related to above is ensuring a config value is finished being mutable (with [ConfigMut]) before
+/// allowing a component to run that uses [Storage] and manually retrieves the config value, or ensuring a service is
+/// registered before allowing a component to run that uses [Storage] to retrieve that service.
+///
+/// ## Example
+/// ```rust
+/// use patina::component::prelude::*;
+///
+/// pub struct ExampleComponent;
+///
+/// #[component]
+/// impl ExampleComponent {
+///   // An entry point that will not run until Config<i32> is available. This ensures any Config mutations occur
+///   // before this component runs.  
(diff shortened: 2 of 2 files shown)


os-d:
I'm not following the problem statement when you are saying params conflict with each other.

For the `With` statement, my understanding is this forces the config and service to be available before the component executes, which is great. But my question is, why would you ever not want that? We already have the Option<> to say I don't care if this exists or not, should the `With` case be the default and only option otherwise? To me, without fully comprehending your problem statement, it seems like it should be. Having the other case be the default seems error prone and like no one would want to use it?

Javagedes:
> I'm not following the problem statement when you are saying params conflict with each other.
> 
> For the `With` statement, my understanding is this forces the config and service to be available before the component executes, which is great. But my question is, why would you ever not want that? We already have the Option<> to say I don't care if this exists or not, should the `With` case be the default and only option otherwise? To me, without fully comprehending your problem statement, it seems like it should be. Having the other case be the default seems error prone and like no one would want to use it?

The crux of the issue is that `&mut Storage` parameter conflicts with `Config` and `ConfigMut` params, which means if a component is reliant on `Storage` and a configuration value, we cannot control if the component executes when the configuration is in a mutable state or in its final mutated state. i.e. ordering of components registered matters, which is not a good thing.

## Scenario 1
1. Default config value is 0
2. `component2(config: ConfigMut<u32>);` // Mutates config to 5;
3. `component1(storage: &mut Storage);`

`component1` runs and consumes config with a value of 5.

## Scenario 2
1. Default config value is 0
2. `component1(storage: &mut Storage);`
3. `component2(config: ConfigMut<u32>);` // Mutates config to 5;

`component1` runs and consumes config with a value of 0.

_____________________

However I'm going to close this soon as the `Commands` parameter solves this issue as it defers storage mutation to after the component executes, allowing for the above scenario to be supported, e.g. a component with this: `fn component(commands: &mut Commands, config: Config<u32>);

cfernald:
>However I'm going to close this soon as the Commands parameter solves this issue as it defers storage mutation to after the component executes, allowing for the above scenario to be supported, e.g. a component with this: `fn component(commands: &mut Commands, config: Config);

@Javagedes, This last part lost me. Are you saying that the `with` parameter would be effectively deprecated if components use `commands` instead of directly using `storage`? Or is there a long term purpose behind `With` where components will continue to require direct access to storage?

Javagedes:
> >However I'm going to close this soon as the Commands parameter solves this issue as it defers storage mutation to after the component executes, allowing for the above scenario to be supported, e.g. a component with this: `fn component(commands: &mut Commands, config: Config);
> 
> 
> 
> @Javagedes, This last part lost me. Are you saying that the `with` parameter would be effectively deprecated if components use `commands` instead of directly using `storage`? Or is there a long term purpose behind `With` where components will continue to require direct access to storage?

I just meant I'm going to close the PR soon once others had time to read my comment, as the `Commands` type can effectively do what I was trying to accomplish here, and I had only thought about it after I created this PR.

Issue #1341, opened 2026-02-25
[Bug]: patina_performance Should Not Guard FBPT with TplMutex


os-d:
### Is there an existing issue for this?

- [x] I have searched existing issues

### Current Behavior

Currently, patina_performance (and the relevant code in the sdk) guard the FBPT with a TplMutex at TPL_NOTIFY. This causes two issues:
- Code running higher than TPL_NOTIFY cannot have performance measured because starting/stopping perf measurements will cause a TPL inversion panic
- When creating/stopping a perf event, the TPL will get lowered and all pending events at the intermediate TPLs will get processed and run. This will throw off the perf results by artificially injecting more code into the path where perf is being tested.

### Expected Behavior

Perf measurements should be able to occur at any TPL and not add significant code into the measurement path.

This may be achieved by dropping the TplMutex and uses atomic instructions to get the next index of the FBPT to work on.

### Steps To Reproduce

Run perf on the FSB

### Build Environment

```markdown
- OS(s):
- Tool Chain(s):
- Targets Impacted:
```

### Version Information

```text
top of tree
```

### Urgency

Medium

### Are you going to fix this?

Someone else needs to fix it

### Do you need maintainer feedback?

No maintainer feedback needed

### Anything else?

@berlin-with0ut-return @magravel @makubacki for reference

makubacki:
How much code is trying to write performance measurements greater than TPL_NOTIFY?

os-d:
No current code, but we don't have patina instrumented for perf measurements except for dispatch related things. E.g. no perf measurements in the memory mgmt code, which is where I hit this issue.

makubacki:
Okay. I don't have any preference on implementation, but I don't think we should allow performance measurements greater than TPL_NOTIFY as that is a violation of the UEFI Specification (and I could imagine complicates perf infrastructure).

<img width="733" height="40" alt="Image" src="https://github.com/user-attachments/assets/a72d8af3-201d-404c-aec4-323b9377a001" />

os-d:
> Okay. I don't have any preference on implementation, but I don't think we should allow performance measurements greater than TPL_NOTIFY as that is a violation of the UEFI Specification (and I could imagine complicates perf infrastructure).
> 
[...]

PR #1358, opened 2026-03-02
Change `TplMutex` in perf component to `RefCell`

berlin-with0ut-return:
## Description

As noted in https://github.com/OpenDevicePartnership/patina/issues/1341, it is problematic to use TplMutex to guard the FBPT during performance measurements as raising and lowering the TPL causes queued event notifies to be executed during performance instrumentation. This PR changes the component to use RefCell. This prevents reentrancy (panics when reentrant) and allows more accurate perf measurements.

- [X] Impacts functionality? 
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested
Existing tests pass. Boot to shell in Q35 + collect FBPT records.

## Integration Instructions
N/A.

### DIFF
--- a/components/patina_performance/src/component/performance.rs
+++ b/components/patina_performance/src/component/performance.rs
@@ -11,7 +11,7 @@
 
 use crate::{component::protocol::create_performance_measurement_efiapi, config, mm};
 use alloc::{boxed::Box, string::String, vec::Vec};
-use core::{clone::Clone, convert::AsRef};
+use core::{cell::RefCell, clone::Clone, convert::AsRef};
 use patina::{
     boot_services::{BootServices, StandardBootServices, event::EventType, tpl::Tpl},
     component::{
@@ -33,7 +33,6 @@ use patina::{
         table::FirmwareBasicBootPerfTable,
     },
     runtime_services::{RuntimeServices, StandardRuntimeServices},
-    tpl_mutex::TplMutex,
     uefi_protocol::performance_measurement::EdkiiPerformanceMeasurement,
 };
 use patina_mm::component::communicator::MmCommunication;
@@ -42,7 +41,7 @@ use r_efi::system::EVENT_GROUP_READY_TO_BOOT;
 pub use mu_rust_helpers::function;
 
 /// Context parameter for the Ready-to-Boot event callback that fetches MM performance records.
-type MmPerformanceEventContext<BB, B, F> = Box<(BB, &'static TplMutex<F, B>, Service<dyn MmCommunication>)>;
+type MmPerformanceEventContext<BB, F> = Box<(BB, &'static RefCell<F>, Service<dyn MmCommunication>)>;
 
 /// Performance Component.
 pub struct Performance;
@@ -89,7 +88,7 @@ impl Performance {
         runtime_services: RR,
         records_buffers_hobs: Option<P>,
         mm_comm_service: Option<Service<dyn MmCommunication>>,
-        fbpt: &'static TplMutex<F, B>,
+        fbpt: &'static RefCell<F>,
         timer: Service<dyn ArchTimerFunctionality>,
     ) -> Result<(), EfiError>
     where
@@ -126,7 +125,7 @@ impl Performance {
             // Initialize perf data from hob values.
 
             set_load_image_count(hob_load_image_count);
-            fbpt.lock().set_perf_records(hob_perf_records);
+            fbpt.borrow_mut().set_perf_records(hob_perf_records);
         } else {
             log::info!("Performance: No Hob performance records provided.");
         }
@@ -345,13 +344,12 @@ impl<'a> Iterator for PerformanceRecordIterator<'a> {
 }
 
 /// Processes MM performance records and adds them to the FBPT
-fn process_mm_performance_records<F, B>(
+fn process_mm_performance_records<F>(
     comm_service: &Service<dyn MmCommunication>,
-    fbpt: &TplMutex<F, B>,
+    fbpt: &RefCell<F>,
 ) -> Result<(), MmPerformanceError>
 where
     F: FirmwareBasicBootPerfTable,
-    B: BootServices + 'static,
 {
     let record_data = fetch_all_mm_record_data(comm_service)?;
 
@@ -383,7 +381,7 @@ where
                 // Print detailed record information based on type
                 print_record_details(record.record_type, record_count, record.data);
 
-                if let Err(e) = fbpt.lock().add_record(record) {
+                if let Err(e) = fbpt.borrow_mut().add_record(record) {
                     error_count += 1;
                     log::error!("Performance: Failed adding MM record #{}: {:?}", record_count, e);
                 } else {
@@ -410,7 +408,7 @@ where
 /// Adds MM performance records to the FBPT.
 pub extern "efiapi" fn fetch_and_add_mm_performance_records<BB, B, F>(
     event: r_efi::efi::Event,
-    ctx: MmPerformanceEventContext<BB, B, F>,
+    ctx: MmPerformanceEventContext<BB, F>,
 ) where
     BB: AsRef<B> + Clone,
     B: BootServices + 'static,
@@ -534,7 +532,7 @@ mod tests {
             .expect_create_event_ex::<Box<(
                 Rc<MockBootServices>,
                 Rc<MockRuntimeServices>,
-                &TplMutex<MockFirmwareBasicBootPerfTable, MockBootServices>,
+                &RefCell<MockFirmwareBasicBootPerfTable>,
             )>>()
             .once()
             .withf_st(|event_type, notify_tpl, notify_function, _notify_context, event_group| {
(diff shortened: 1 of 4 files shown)


sdk/patina/src/boot_services/tpl.rs
```diff
@@ -47,6 +47,10 @@ impl Tpl {
     /// If code requires more processing, it needs to signal an event to wait to obtain control again at whatever level it requires.
     /// This level is typically used to process low level IO to or from a device.
     pub const NOTIFY: Tpl = Tpl(efi::TPL_NOTIFY);
+
+    /// The highest task priority level.
+    /// At this level, all interrupts and event notifications are masked.
+    pub const HIGH_LEVEL: Tpl = Tpl(efi::TPL_HIGH_LEVEL);
```
makubacki, line 53: Can adding this constant go into its own commit?

sdk/patina/src/performance/globals.rs
makubacki, line 81: It looks like this safety message is no longer accurate?

components/patina_performance/src/component/performance.rs
makubacki, line 1: Can you please update the PR description to explain the soundness of this change?

For example, here are some high-level scenarios I'm initially concerned about. Can you confirm if these are still safe?

1. EDK II Performance Measurement Protocol is used by a driver at `TPL_APPLICATION` to invoke `create_performance_measurement_efiapi()`. This calls `create_performance_measurement()` which calls `_create_performance_measurement()` which calls `fbpt.borrow_mut().add_record(record)?`.
   - Can't a timer event now interrupt that code at a TPL higher than `TPL_APPLICATION`?
   - For example, a `TPL_CALLBACK` event that calls `create_performance_measurement()`.
     - My understanding, is that in this case, `fbpt.borrow_mut()` will panic because it is already borrowed which is undesirable.
2. `report_fbpt_record_buffer()` is registered at `TPL_CALLBACK`:

   ```rust
        // Register EndOfDxe event to allocate the boot performance table and report the table address through status code.
        boot_services.as_ref().create_event_ex(
            EventType::NOTIFY_SIGNAL,
            Tpl::CALLBACK,
            Some(event_callback::report_fbpt_record_buffer),
            Box::new((BB::clone(&boot_services), RR::clone(&runtime_services), fbpt)),
            &EVENT_GROUP_END_OF_DXE,
        )?
   ```
   - A similar situation exists where `let Ok(fbpt_address) = fbpt.borrow_mut().report_table(...)` is held at `TPL_CALLBACK` which can be preempted by a `TPL_NOTIFY` callback.
3. You also have:

   ```rust
            boot_services.as_ref().create_event_ex(
                EventType::NOTIFY_SIGNAL,
                Tpl::CALLBACK,
                Some(fetch_and_add_mm_performance_records::<BB, B, F>),
                Box::new((BB::clone(&boot_services), fbpt, mm_comm_service)),
                &EVENT_GROUP_READY_TO_BOOT,
            )?;
   ```

   This adds records with:

   ```rust
   if let Err(e) = fbpt.borrow_mut().add_record(record) {
   ```

   Which we also don't want interrupted with something at `TPL_NOTIFY`.

---

Please let me know if I misunderstood the change.

berlin-with0ut-return: it seems that the general issue is reentrancy when a lower-TPL perf recording event is preempted by a higher-TPL perf recording event. the behavior of EDKII in this case is to drop the reentrant (higher-TPL) call -- it is protected with a simple boolean. if we want to model this behavior, we can change all the borrow_mut() to try_borrow_mut(). however, if we want something more preemption-proof, let's discuss offline how we can do this without also encountering the event queueing issues of TplMutex

makubacki: Can you clarify what is meant by "event queueing issues of TplMutex"?

Events being queued for dispatch is expected behavior, so more detail is needed to describe a problem with that.

---

Having code that is inherently dependent on TPL (that's how code is scheduled) for its execution encounter a situation in which it preempts code and cannot acquire a lock causing it to manually handle the error and return is not ideal and has its own overhead.

What TPL helps with is that the performance code can prevent preemption by acquiring a lock at the maximum level the performance interface can be called at (it's caller restricted level). Then, no callers can interrupt performance critical sections. They are deferred in the TPL notify queue until their calling restriction is met. In that model, the shared resource is protected, events are not lost, code paths are not entered that will ultimately fail, and the priority level of code execution is consistent. Code should not be queued for execution on TPL restorations unless there is a reason.

While the TPL system is not perfect and probably not being used correctly everywhere, it is what is used and must be for compatibility with the UEFI Specification and C code.

---

It's also worth noting that Patina can be configured to do things like set up a bunch of [tests on timer events](https://github.com/OpenDevicePartnership/patina/blob/f7c43cd9fcf60e480b89055e586b070134cb1663/patina_dxe_core/src/core_patina_tests/stability_tests.rs#L25) -> [`run_tests()`](https://github.com/OpenDevicePartnership/patina/blob/8d26951ec6c8e86d1128acfa532363345e85f43b/sdk/patina/src/test.rs#L368C8-L368C17) that, if enabled on a platform, is going to cause a lot of noise in the event database:

https://github.com/OpenDevicePartnership/patina/blob/8d26951ec6c8e86d1128acfa532363345e85f43b/sdk/patina/src/test.rs#L422-L452

I'm happy to talk further offline on details.

os-d: We should chat further offline, but there are two problems that get hit today:
1. When using TplMutex, we process all pending events when dropping the lock, this means our perf data is skewed by whatever pending events there are. edk2 does not raise/lower TPL when starting/stopping perf measurements.
2. We get into an issue when a dependency of the performance component is attempted to be used, e.g. profiling memory allocation perf (outside of the TplMutex there to not violate the spec). The memory alloc code calls start measurement, the perf component gets called and calls allocate, allocate calls get start measurement, deadlock on perf component TplMutex.

So, that's the two sides of Sherry and I talking this out. Not manipulating the TPL to avoid skewing perf results and doing the try_ version of interior mutability (same thing edk2 does, in effect) so that we ignore the recursive case.

makubacki: > When using TplMutex, we process all pending events when dropping the lock, this means our perf data is skewed by whatever pending events there are. edk2 does not raise/lower TPL when starting/stopping perf measurements.

edk2 code can modify TPL when making measurements. For example:

[CreatePerformanceMeasurement()](https://github.com/tianocore/edk2/blob/a39c3835aaea8757854b2b0890d14fe16cdcbb21/MdeModulePkg/Library/DxeCorePerformanceLib/DxeCorePerformanceLib.c#L1602) calls [InsertFpdtRecord()](https://github.com/tianocore/edk2/blob/a39c3835aaea8757854b2b0890d14fe16cdcbb21/MdeModulePkg/Library/DxeCorePerformanceLib/DxeCorePerformanceLib.c#L991) which, for some measurements, calls [GetModuleInfoFromHandle](https://github.com/tianocore/edk2/blob/a39c3835aaea8757854b2b0890d14fe16cdcbb21/MdeModulePkg/Library/DxeCorePerformanceLib/DxeCorePerformanceLib.c#L476) which calls services like `HandleProtocol()` and `CloseProtocol()` which will modify TPL through `gProtocolDatabaseLock`.

Even calls to [StartPerformanceMeasurement()/StartPeformanceMeasurementEx()](https://github.com/tianocore/edk2/blob/a39c3835aaea8757854b2b0890d14fe16cdcbb21/MdeModulePkg/Library/DxePerformanceLib/DxePerformanceLib.c#L86) and [EndPerformanceMeasurement()/EndPerformanceMeasurementEx()](https://github.com/tianocore/edk2/blob/a39c3835aaea8757854b2b0890d14fe16cdcbb21/MdeModulePkg/Library/DxePerformanceLib/DxePerformanceLib.c#L145) in DxePerformanceLib are going to call [GetPerformanceMeasurementProtocol()](https://github.com/tianocore/edk2/blob/a39c3835aaea8757854b2b0890d14fe16cdcbb21/MdeModulePkg/Library/DxePerformanceLib/DxePerformanceLib.c#L38) which through [gBS->LocateProtocol()](https://github.com/tianocore/edk2/blob/a39c3835aaea8757854b2b0890d14fe16cdcbb21/MdeModulePkg/Core/Dxe/Hand/Locate.c#L603) is going to acquire [gProtocolDatabaseLock](https://github.com/tianocore/edk2/blob/a39c3835aaea8757854b2b0890d14fe16cdcbb21/MdeModulePkg/Core/Dxe/Hand/Handle.c#L20) which is going to raise the TPL to `TPL_NOTIFY`.

Helpers like [GetDeviceInfoFromHandleAndUpdateLength()](https://github.com/tianocore/edk2/blob/a39c3835aaea8757854b2b0890d14fe16cdcbb21/MdeModulePkg/Library/DxeCorePerformanceLib/DxeCorePerformanceLib.c#L858C1-L858C40) are going to modify TPL through the protocol lock, the UEFI variable lock, and likely the memory allocator lock.

Anyway, I'm just trying to demonstrate why these problem statements are confusing.

---

I understand there are real problems like reentrancy on perf code dependencies (e.g. memory allocation) and that data exists (or can be gathered) showing real problems with the current code. But I think the solution needs to have a bit more of that data shown so the problem/solution is obvious and a bit more analysis on how it works in the different scenarios that are expected to be supported.

berlin-with0ut-return: it seems overall from the discussion above that converting to RefCell alone is not a sufficient solution, and that further discussion/maybe an RFC is necessary. as such i'm closing this PR for now

PR #1360, opened 2026-03-02
patina_test: Better support multiple TestRunners

Javagedes:
## Description

`patina_test` has always supported multiple TestRunner components to easily group tests together and have them executed. With the addition of different test triggers (such as test execution on specific UEFI events), the existing architecture has broken down. While it works, tests are currently executed more than expected (e.g. each `TestRunner` will run the test if there is an overlap in test cases). Additionally test result reporting is also logged per `TestRunner` but reports the total culmination of all tests executed.

This commit represents a large rework to resolve these issues by moving test management and execution to a private service while each `TestRunner` component is only responsible for registering tests with the single underlying service.

This impact of this commit is that each test is now only registered once, even if multiple `TestRunner` components specify the test case. Additionally, test reporting is now done by the underlying service, which prevents test results being reported multiple times. Additionally, `Immediate` (now `Manual`) tests were originally run during component execution. This has been pushed off to `Ready to Boot` and `Exit Boot Services` to allow all components to execute. Finally, each component may register its own test error handler. With these changes, if a test failure occurs, any `TestRunner` that registered that test to run, will have its handler executed.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [x] Includes documentation?

## How This Was Tested

Registered multiple TestRunner components with various scenarios and verified that:

1. All test handler failures are executed
2. Each test is only executed once regarldess of how many `TestRunner` components include the test in it's scope.
3. Tests based off of events or timers only execute once per trigger, rather than once per `TestRunner` per trigger.

## Integration Instructions

N/A

### DIFF
--- a/sdk/patina/src/test.rs
+++ b/sdk/patina/src/test.rs
@@ -192,7 +192,7 @@ impl Recorder {
         storage.boot_services().create_event_ex(
             EventType::NOTIFY_SIGNAL,
             Tpl::CALLBACK,
-            Some(Self::log_test_results),
+            Some(Self::run_tests_and_report),
             NonNull::from_ref(storage),
             &EVENT_GROUP_READY_TO_BOOT,
         )?;
@@ -201,7 +201,7 @@ impl Recorder {
         storage.boot_services().create_event(
             EventType::SIGNAL_EXIT_BOOT_SERVICES,
             Tpl::CALLBACK,
-            Some(Self::log_test_results),
+            Some(Self::run_tests_and_report),
             NonNull::from_ref(storage),
         )?;
 
@@ -238,7 +238,7 @@ impl Recorder {
     }
 
     /// An EFIAPI compatible event callback to run the manually triggered tests and log the current results of patina-test
-    extern "efiapi" fn log_test_results(event: r_efi::efi::Event, mut storage: NonNull<Storage>) {
+    extern "efiapi" fn run_tests_and_report(event: r_efi::efi::Event, mut storage: NonNull<Storage>) {
         // SAFETY: event callbacks are executed in series, so there exists no other mutable access to storage.
         let storage = unsafe { storage.as_mut() };
 
@@ -444,7 +444,6 @@ impl TestRunner {
     #[coverage(off)]
     fn entry_point(self, storage: &mut Storage) -> patina::error::Result<()> {
         let test_list: &'static [__private_api::TestCase] = __private_api::test_cases();
-        log::error!("Registering {} tests with TestRunner", test_list.len());
         self.register_tests(test_list, storage)
     }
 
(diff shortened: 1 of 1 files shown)


os-d (approved):

joschock (approved):

makubacki (approved):

sdk/patina/src/test.rs
```diff
@@ -361,132 +444,43 @@ impl TestRunner {
     #[coverage(off)]
     fn entry_point(self, storage: &mut Storage) -> patina::error::Result<()> {
         let test_list: &'static [__private_api::TestCase] = __private_api::test_cases();
-        self.run_tests(test_list, storage)
+        log::error!("Registering {} tests with TestRunner", test_list.len());
```
makubacki: Is this intentionally at error level?

makubacki: Similarly for `log::error!("Scheduling test {} with TestRunner", record.test_case.name);` further down.

sdk/patina/src/test.rs
```diff
@@ -202,121 +170,236 @@ macro_rules! u_assert_ne {
...
+                .for_each(|record| record.run(storage));
+        });
     }
 
-    /// Registers a test name with an empty record, so that it shows up in the final results even if not triggered.
-    fn empty_record(&self, name: &'static str) {
-        // SAFETY: This is safe due to the invariance of this struct so long as it is only accessed via the component system.
-        let data = unsafe { self.results.get().as_mut().expect("Pointer is not null.") };
+    /// An EFIAPI compatible event callback to run the manually triggered tests and log the current results of patina-test
+    extern "efiapi" fn log_test_results(event: r_efi::efi::Event, mut storage: NonNull<Storage>) {
```
makubacki: It seems something like `run_and_log_test_results()` would technically be more accurate. I don't have a strong preference though.

sdk/patina/src/test.rs
Javagedes: No. Those were two last minute statements I added when debugging. I will remove them.

sdk/patina/src/test.rs
```diff
@@ -202,121 +170,236 @@ macro_rules! u_assert_ne {
...
+    /// Registers UEFI event callbacks to log the test results at specific points in the boot process.
+    fn initialize(&self, storage: &mut Storage) -> patina::error::Result<()> {
+        // Log results at ready to boot
+        storage.boot_services().create_event_ex(
+            EventType::NOTIFY_SIGNAL,
+            Tpl::CALLBACK,
+            Some(Self::log_test_results),
+            NonNull::from_ref(storage),
+            &EVENT_GROUP_READY_TO_BOOT,
+        )?;
```
makubacki, line 198: Can you confirm if this understanding is correct: This will run these tests for the first time at Ready to Boot? If so, was that done before? If not, that might be worth calling out in the PR description.

makubacki (approved):

makubacki:
Had a few lingering thoughts I was going to ignore and just approve earlier. But decided to go ahead leave. No further comments.

sdk/patina/src/test.rs
Javagedes: You are correct, and it was not like that before, so I will update the PR / commit. Good call out

sdk/patina/src/test.rs
Javagedes: Updated the name. I agree being as clear as possible is best.

PR #1368, opened 2026-03-04
Makefile.toml: Add test-asan task for AddressSanitizer testing

makubacki:
## Description

Adds a `test-asan` task that runs tests  with AddressSanitizer (`-Zsanitizer=address`) enabled. The task supports Windows x64 and Linux hosts.

Full task command: `cargo make test-asan`

On Windows [1], ASan-instrumented binaries require the MSVC ASan runtime DLL (`clang_rt.asan_dynamic-x86_64.dll`) to be on PATH at runtime so that's done as part of the task setup. The DLL location is resolved using the VC toolset installation path from `vswhere` [2].

On Linux, no additional setup is needed as the ASan runtime is resolved automatically by the linker.

[1] https://learn.microsoft.com/cpp/sanitizers/asan
[2] https://github.com/microsoft/vswhere/wiki/Find-VC

A `--print-dll-path` argument is supported on Windows to print the resolved ASan DLL path and exit, which can be useful for adding that to the system PATH when running tests outside of `cargo make`.

---

Note: I've already fixed an initial set of issues other than (mostly intentional) memory leaks. I can add this to CI in a follow up PR if there is interest in doing so. This PR is to raise awareness in the patina repo about the change. It will be added to Makefile.toml in patina-devops as well, if accepted.

---

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- `cargo make test-asan` on Windows (with Visual Studio installed) and Ubuntu 24.04

## Integration Instructions

- N/A - New task available

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -158,6 +158,58 @@ command = "cargo"
 args = ["llvm-cov", "@@split(COV_FLAGS, )", "--no-report", "--doctests"]
 dependencies = ["individual-package-targets", "llvm-cov-clean"]
 
+[tasks.test-asan]
+description = "Run tests with AddressSanitizer enabled. Example `cargo make test-asan`. Use `cargo make test-asan --print-dll-path` to print the ASan runtime DLL directory."
+clear = true
+script_runner = "@duckscript"
+script = '''
+os = os_family
+
+if eq ${os} "windows"
+    # ASan testing is only supported on Windows x64.
+    # https://learn.microsoft.com/cpp/sanitizers/asan
+    arch = get_env PROCESSOR_ARCHITECTURE
+    if not eq ${arch} "AMD64"
+        echo "test-asan is only supported on Windows x64. Detected arch: ${arch}. Skipping."
+        exit 0
+    end
+
+    # The ASan runtime DLL (clang_rt.asan_dynamic-x86_64.dll) from the
+    # MSVC toolchain must be on PATH for ASan-instrumented test binaries to run.
+    #
+    # Resolve the VC toolset path per vswhere docs:
+    # https://github.com/microsoft/vswhere/wiki/Find-VC
+    program_files_x86 = get_env "ProgramFiles(x86)"
+    vswhere = set "${program_files_x86}/Microsoft Visual Studio/Installer/vswhere.exe"
+    output = exec "${vswhere}" -latest -products * -requires Microsoft.VisualStudio.Component.VC.Tools.x86.x64 -property installationPath
+    install_dir = trim ${output.stdout}
+    version = readfile "${install_dir}/VC/Auxiliary/Build/Microsoft.VCToolsVersion.default.txt"
+    version = trim ${version}
+    asan_dll_path = set "${install_dir}/VC/Tools/MSVC/${version}/bin/HostX64/x64"
+
+    args = get_env CARGO_MAKE_TASK_ARGS
+    if contains "${args}" "--print-dll-path"
+        asan_dll_path = replace ${asan_dll_path} / \\
+        echo "${asan_dll_path}"
+        exit 0
+    end
+
+    current_path = get_env PATH
+    set_env PATH "${asan_dll_path};${current_path}"
+elseif eq ${os} "linux"
+    # On Linux, the ASan runtime is resolved automatically by the linker.
+    noop
+else
+    echo "test-asan is not supported on ${os}. Skipping."
+    exit 0
+end
+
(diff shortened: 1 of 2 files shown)


Makefile.toml
```diff
@@ -158,6 +158,49 @@ command = "cargo"
...
+    # On Linux, the ASan runtime is resolved automatically by the linker.
+    noop
+else
+    echo "test-asan is not supported on ${os}. Skipping."
+    exit 0
+end
+
+set_env RUSTFLAGS "-Zsanitizer=address"
+target = get_env CARGO_MAKE_RUST_TARGET_TRIPLE
+exec --fail-on-error cargo test --target ${target} --workspace --features std
```
Javagedes, line 210: have you tried setting the RUSTFLAGS variable here then just calling cargo make test? I know we use llvm-cov backed by cargo test, so I'm not positive it will work, but may make long term maintenance simpler if we can just call the existing makefile command then maintain this custom call.

If it does not work, then I'm fine with this as-is.

os-d (approved):
Is this sync’d from patina-devops?

makubacki:
> Is this sync’d from patina-devops?

That's mentioned in the PR description:

> Note: I've already fixed an initial set of issues other than (mostly intentional) memory leaks. I can add this to CI in a follow up PR if there is interest in doing so. This PR is to raise awareness in the patina repo about the change. It will be added to Makefile.toml in patina-devops as well, if accepted.

Before doing that, we can address the CI question.

After my recent PRs, Windows output is clean. Linux fails with "SUMMARY: AddressSanitizer: 235784 byte(s) leaked in 269 allocation(s).".

Would anyone be opposed to me adding this to Windows CI?

makubacki:
> I like the general idea of being able to run code with sanitizers and below comment is not particularly related to the current PR.
> 
> My only concern is the way we are bloating the `Makefile.toml` with multiple test related tasks.
> 
> ```
> [tasks.check_tests] <-- This is marked private and only checks test but do not run them!
> [tasks.patina-test] <-- This one actually builds with `enable_patina_tests` feature but do not run tests!
> [tasks.test]      <-- Not the cargo test runner task! Instead hijacked for use with code cov
> [tasks.coverage]  <-- Runs the test behind the scenes
> [tasks.test-asan]  <-- This seems to run vanila `cargo test` but do not consider package path. 
> ```
> 
> I find it difficult to identify the actual task which is responsible for running tests with a given package(`pi::fw_fs::unit_tests`). I vaguely remember this has been discussed multiple times with `INDIVIDUAL_PACKAGE_TARGETS` or something. But the point is it is hard to discover the usage.
> 
> Please correct me if I am missing something.

For unit tests, what's wrong with just using `cargo make test`? The others are more advanced cases.

Makefile.toml
makubacki: I have not, but I'll try it.

vineelko:
> cargo make test

Again, not specifically for this PR. 

I think we should preserve the basic workflow from `cargo` even with `cargo make`
cargo test => cargo make test 
cargo test package => cargo make test package

I consider everything else as advanced use cases. Unfortunately, `cargo make test package` do not honor the package and instead runs on all packages.

Makefile.toml
makubacki: That appears to functionally work but it's still running >30 mins versus a couple of minutes before.

makubacki:
> > cargo make test
> 
> Again, not specifically for this PR.
> 
> I think we should preserve the basic workflow from `cargo` even with `cargo make` cargo test => cargo make test cargo test package => cargo make test package
> 
> I consider everything else as advanced use cases. Unfortunately, `cargo make test package` do not honor the package and instead runs on all packages.

I also wanted that to work and made it work in https://github.com/OpenDevicePartnership/patina/commit/e17f7d853da5fba211ab11b93f19e1a9ff21f550. It looks like that was unfortunately dropped later in https://github.com/OpenDevicePartnership/patina/commit/47cf46bb392cc867a8dda62040b03a44cd7bfeed. I didn't realize that happened and I can look into making that possible again.

However, we can follow up somewhere else since it's off topic.

Makefile.toml
makubacki: It's still running. It doesn't seem practical to run with coverage.

makubacki: Going to close this thread based on that experience.

vineelko (approved):

PR #1369, opened 2026-03-04
Patina: Fix RAW FFS file handling and harden section parsing

vineelko:
## Description

1. Handle RAW File types:
    - The current implementation tries to produce sections irrespective of the file
      type. This do not hold true for RAW file types. As per PI spec below,

        _The file type EFI_FV_FILETYPE_RAW denotes a file that does not contain
        sections and is treated as a raw data file. The consumer of this type of
        file must have a priori knowledge of its format and content. Because there
        are no sections, there are no construction rules._

    - Trying to decode the body of these files as Section header can result in
      incorrect parsing.

    Fixes the following bug:
    ```
    [ INFO]: \patina\sdk\patina\src\pi\serializable\serializable_fv.rs@078: Parsing FV: d2c29ba7-3809-480f-9c3d-de389c61425a
    [ INFO]: \patina\sdk\patina\src\pi\serializable\serializable_fv.rs@089:     Parsing file: 197db236-f856-4924-90f8-cdf12fb875f3
    [ INFO]: \patina\sdk\patina\src\pi\fw_fs.rs@801: Section type 0x0 unrecognized 4 1, treating as opaque with no metadata
    [ERROR]: dxe_readiness_capture\src\lib.rs@031: panicked at \patina\sdk\patina\src\pi\fw_fs.rs:805:56:
    [ERROR]: dxe_readiness_capture\src\lib.rs@031: slice index starts at 4 but ends at 1
    ```

3. Improve safety in FFS section parsing by replacing direct slice indexing with
   bounds checked `get()` calls and proper error propagation when buffers are
   malformed. General code cleanup.

---

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Booted to UEFI Shell on Q35

## Integration Instructions

NA


sdk/patina/src/pi/fw_fs.rs
```diff
@@ -715,91 +715,91 @@ impl Section {
...
-                if buffer.len() < content_offset + compression_header_size {
-                    Err(efi::Status::VOLUME_CORRUPTED)?;
-                }
+                let compression_header = buffer.get(content_offset..).ok_or(efi::Status::VOLUME_CORRUPTED)?;
                 // SAFETY: Size was validated to contain a Compression header and slice bounds is checked
                 let compression_header =
-                    unsafe { &*(buffer[content_offset..].as_ptr() as *const section::header::Compression) };
-                let data: Box<[u8]> = Box::from(&buffer[content_offset + compression_header_size..section_size]);
-                (SectionMetaData::Compression(*compression_header), data)
+                    unsafe { *(compression_header.as_ptr() as *const section::header::Compression) };
```
Javagedes: This seems like we should use zero copy to convert this byte buffer to a struct.

Javagedes: And you should have the type be a DST with the content coming after it. Then just generate the returned struct from the DST, allocating / copying where necessary.

sdk/patina/src/pi/fw_fs.rs
```diff
@@ -715,91 +715,91 @@ impl Section {
...
-                let guid_defined =
-                    unsafe { &*(buffer[content_offset..].as_ptr() as *const section::header::GuidDefined) };
-
-                //verify that buffer has enough storage for guid-specific fields.
-                let data_offset = guid_defined.data_offset as usize;
-                if buffer.len() < data_offset {
-                    Err(efi::Status::VOLUME_CORRUPTED)?;
-                }
+                let guid_defined_header =
+                    unsafe { *(guid_defined_header.as_ptr() as *const section::header::GuidDefined) };
```
Javagedes: It also looks like all of this here can be simplified with zerocopy and some DSTs

sdk/patina/src/pi/fw_fs.rs
```diff
@@ -715,91 +715,91 @@ impl Section {
...
-                if buffer.len() < content_offset + version_header_size {
-                    Err(efi::Status::VOLUME_CORRUPTED)?;
-                }
+                let version_header = buffer.get(content_offset..).ok_or(efi::Status::VOLUME_CORRUPTED)?;
                 // SAFETY: Size wass validated to contain a Version header and slice bounds is checked
-                let version_header =
-                    unsafe { &*(buffer[content_offset..].as_ptr() as *const section::header::Version) };
-                let data: Box<[u8]> = Box::from(&buffer[content_offset + version_header_size..section_size]);
-                (SectionMetaData::Version(*version_header), data)
+                let version_header = unsafe { *(version_header.as_ptr() as *const section::header::Version) };
```
Javagedes: and here :)

Javagedes: OK I see many places where zerocopy can be used. I won't comment anymore :)

sdk/patina/src/pi/fw_fs.rs
os-d: I was also wondering if we could leverage zerocopy here

sdk/patina/src/pi/fw_fs.rs
```diff
@@ -715,91 +715,91 @@ impl Section {
...
         };
 
         let (meta_data, data) = match section_header.section_type {
             FfsSectionRawType::encapsulated::COMPRESSION => {
-                let compression_header_size = mem::size_of::<section::header::Compression>();
-                //verify that buffer has enough storage for a compression header.
-                if buffer.len() < content_offset + compression_header_size {
-                    Err(efi::Status::VOLUME_CORRUPTED)?;
-                }
+                let compression_header = buffer.get(content_offset..).ok_or(efi::Status::VOLUME_CORRUPTED)?;
```
joschock: this construction doesn't check that the compression header is big enough to be an actual compression header. shouldn't it be something like `let compression_header = buffer.get(content_offset..content_offset + mem::size_of::<section::header::Compression>())`?

sdk/patina/src/pi/fw_fs.rs
```diff
@@ -715,91 +715,91 @@ impl Section {
...
+                    .ok_or(efi::Status::VOLUME_CORRUPTED)?;
+                (SectionMetaData::Compression(compression_header), Box::from(data))
             }
             FfsSectionRawType::encapsulated::GUID_DEFINED => {
-                let guid_defined_header_size = mem::size_of::<section::header::GuidDefined>();
-                //verify that buffer has enough storage for a guid_defined header.
-                if buffer.len() < content_offset + guid_defined_header_size {
-                    Err(efi::Status::VOLUME_CORRUPTED)?;
-                }
+                let guid_defined_header = buffer.get(content_offset..).ok_or(efi::Status::VOLUME_CORRUPTED)?;
```
joschock: same - content_offset might be legal, but the full extent of guid header might not be.

sdk/patina/src/pi/fw_fs.rs
joschock: Agree. I think much of this module predates our discovery/adoption of zerocopy, and is probably a ripe case for its usage.

makubacki: @vineelko , this expands your work quite a bit, but would you be willing to update this code to use zerocopy? This would move the codebase closer to our overall safety goals.

vineelko: Sure Michael, I can do that. However, the problem I’m encountering is that some of these FFS structs use r-efi’s `Guid` fields, which do not have the ZeroCopy traits implemented. This prevents us from using ZeroCopy on those FFS structures.

makubacki: We'll probably need to write wrappers that use Patina's GUID types.

vineelko:
Please hold on this PR, Got the fv from IDC folks, It looks like there is more to the actual bug. I will notify once I finished root causing the issue.

sdk/patina/src/pi/fw_fs.rs
vineelko: Fixed in the latest iteration.

sdk/patina/src/pi/fw_fs.rs
vineelko: Thanks for advising zerocopy. It eliminated some of the unsafe usage in the latest iteration of the PR. But some of the other FFS structures cannot be marked zero copiable because of Guid usage from r-efi. I will try to tackle them and other unsafe usages in a separate PR.

sdk/patina/src/pi/fw_fs.rs
vineelko: Thanks for advising zerocopy. It eliminated some of the unsafe usage in the latest iteration of the PR. But some of the other FFS structures cannot be marked zero copiable because of Guid usage from r-efi. I will try to tackle them and other unsafe usages in a separate PR.

sdk/patina/src/pi/fw_fs.rs
vineelko: Fixed it latest iteration.

vineelko:
As mentioned before, the section parsing failed with below error(Section Type 0x0).

```
[ INFO]: E:\repos\patina\sdk\patina\src\pi\serializable\serializable_fv.rs@078: Parsing FV: d2c29ba7-3809-480f-9c3d-de389c61425a
[ INFO]: E:\repos\patina\sdk\patina\src\pi\serializable\serializable_fv.rs@089:     Parsing file: 197db236-f856-4924-90f8-cdf12fb875f3
[ INFO]: E:\repos\patina\sdk\patina\src\pi\fw_fs.rs@801: Section type 0x0 unrecognized 4 1, treating as opaque with no metadata
[ERROR]: dxe_readiness_capture\src\lib.rs@031: panicked at E:\repos\patina\sdk\patina\src\pi\fw_fs.rs:805:56:
[ERROR]: dxe_readiness_capture\src\lib.rs@031: slice index starts at 4 but ends at 1
```
The parsing of sections in the section iterator required few fixes, especially around handling SectionType::ALL to ensure parity with the C UEFI implementation, where the entire body is returned.

However, after looking at the FV, The current implementation attempts to produce sections regardless of the file type. This assumption does not hold true for RAW file types. According to the PI specification below,

image

image

image

Trying to decode the body of these files as Section header can result in incorrect parsing. This is addressed in the latest iteration. 

PR is up for review.

sdk/patina/src/pi/fw_fs.rs
```diff
@@ -715,91 +715,91 @@ impl Section {
...
-                (SectionMetaData::None, data)
+                // These section types do not have a defined header. So set metadata to none, and set data to the entire section buffer.
+                (SectionMetaData::None, Box::from(buffer))
+            }
+            FfsSectionRawType::ALL => {
+                // When section type is ALL, the spec requires that the entire
+                // section be treated as raw data with no header, so set
+                // metadata to none, and set data to the entire section buffer.
+                section_size = buffer.len();
+                (SectionMetaData::None, Box::from(buffer))
```
makubacki: Can you link to where the spec says this?

I've only heard of `ALL` being used as a filtering type.

vineelko: My bad, I am actually referring to the C implementation for this.

https://github.com/microsoft/mu_basecore/blob/release/202511/MdeModulePkg/Core/Dxe/FwVol/FwVolRead.c#L505-L507

```
  //
  // If SectionType == 0 We need the whole section stream
  //
```

sdk/patina/src/pi/fw_fs.rs
makubacki: Until zerocopy is used, because `GuidDefined` is:

```rust
pub struct GuidDefined {
    pub section_definition_guid: Guid,
    pub data_offset: u16,
    pub attributes: u16,
}
```

And r-efi defines `Guid` on 4-byte alignment:

```rust
/// The individual fields are encoded as little-endian. Accessors are provided for the Guid
/// structure allowing access to these fields in native endian byte order.
#[repr(C, align(4))]
#[derive(Clone, Copy, Debug)]
#[derive(Eq, Hash, Ord, PartialEq, PartialOrd)]
pub struct Guid {
    time_low: [u8; 4],
    time_mid: [u8; 2],
    time_hi_and_version: [u8; 2],
    clk_seq_hi_res: u8,
    clk_seq_low: u8,
    node: [u8; 6],
}
```

I think that this should use `core::ptr::read_unaligned()`.

```suggestion
                let guid_defined_header =
    unsafe { core::ptr::read_unaligned(guid_defined_header.as_ptr() as *const section::header::GuidDefined) };
```

sdk/patina/src/pi/fw_fs.rs
```diff
@@ -636,7 +637,7 @@ impl<'a> File<'a> {
         &'b self,
         extractor: &'b dyn SectionExtractor,
     ) -> impl Iterator<Item = Result<Section, efi::Status>> + 'b {
-        FileSectionIterator::new(&self.data[self.header_size..self.size as usize], extractor)
+        FileSectionIterator::new(&self.data[self.header_size..self.size as usize], extractor, self.file_type)
```
makubacki: I think it would be cleaner to just return an empty iterator for the `RAW` type. Then, there doesn't need to be special file type logic for it in `FileSectionIterator`.

Something like:


```suggestion
    let buffer = if self.file_type == FfsFileRawType::RAW {
        &[] as &[u8]  // RAW files have no sections
    } else {
        &self.data[self.header_size..self.size as usize]
    };
    FileSectionIterator::new(buffer, extractor)
```

sdk/patina/src/pi/fw_fs.rs
makubacki: Is this referring to the actual type in the FFS file?

If so, I think we should flag that as an error. (1) I don't see any evidence that's a valid file type versus a pseudo-type for filtering and (2) since `SectionType:ALL` is `0`, it's likely that this would be a corrupt/invalid buffer.

The comment led me to believe that this is referring to the file value versus a filtering value. Please clarify if that's not the case.

sdk/patina/src/pi/fw_fs.rs
vineelko: Fixed it.

sdk/patina/src/pi/fw_fs.rs
vineelko: Fixed it.

sdk/patina/src/pi/fw_fs.rs
vineelko: I am reevaluating this with the IDC team to see if handling of RAW file type is sufficient enough to not encounter this case as you called out, the filtering vs section type actually being `ALL`.  In any case, I will mark it as error.

vineelko: Michael, this is up for review again with the suggested changes.

makubacki (approved):

makubacki:
@vineelko, the PR title gives the impression this is a broad FFS section parsing issue and I think it really comes down to RAW files and some other cleanup. If that's the case, could you consider updating the title?

vineelko:
Fixed the title Michael.

os-d (approved):

PR #1372, opened 2026-03-05
patina_debugger: Cleanup features and feature usage

cfernald:
## Description

This commit addresses two issues with the patina_debugger features:

1. Removes the `windbg_workarounds` feature. Windbg is the primary supported scenario for patina_debugger, and these workaround do not inhibit other debuggers from working. Removing the feature reduces complexity and removes the ability for users to accidentally break windbg support.

2. Removes the default features from patina_debugger for the workspace crate. This allows the features to be enabled for the appropriate core and by end platforms if applicable.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Regression tested with windbg on Q35

## Integration Instructions

N/A

### DIFF
--- a/core/patina_debugger/src/dbg_target.rs
+++ b/core/patina_debugger/src/dbg_target.rs
@@ -34,8 +34,8 @@ use crate::{
 };
 
 /// Addresses that windbg will attempt to read in a loop, reads from these addresses
-/// will just return 0 to avoid long retry delays.
-#[cfg(feature = "windbg_workarounds")]
+/// will just return 0 to avoid long retry delays. This has been fixed in windbg, but leaving this
+/// here for a while to support older versions of windbg. Will remove in the future.
 const WINDBG_MOCK_ADDRESSES: [u64; 3] = [0xfffff78000000268, 0, 0x34c00];
 
 /// Patina target for GDB.
@@ -82,21 +82,21 @@ impl Target for PatinaTarget {
         gdbstub::target::ext::base::BaseOps::SingleThread(self)
     }
 
-    #[cfg(feature = "windbg_workarounds")]
     #[inline(always)]
     fn use_no_ack_mode(&self) -> bool {
+        // Windbg doesn't support no-ack mode.
         false
     }
 
-    #[cfg(feature = "windbg_workarounds")]
     #[inline(always)]
     fn use_rle(&self) -> bool {
+        // Windbg doesn't support RLE.
         false
     }
 
-    #[cfg(feature = "windbg_workarounds")]
     #[inline(always)]
     fn use_x_upcase_packet(&self) -> bool {
+        // Windbg doesn't support x-upcase packet.
         false
     }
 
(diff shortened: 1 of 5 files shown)


patina_dxe_core/Cargo.toml
```diff
@@ -31,7 +31,7 @@ patina = { workspace = true, features = ["core"] }
 patina_ffs = { workspace = true }
 patina_internal_collections = { workspace = true  }
 patina_internal_cpu = { workspace = true }
-patina_debugger = { workspace = true }
+patina_debugger = { workspace = true, features = ["alloc"] }
```
os-d, line 34: Does this work for disabling alloc from the platform?

cfernald: No. The thinking being that the core implementation actually determines whether `alloc` should be used, not the platform.

core/patina_debugger/src/dbg_target.rs
```diff
@@ -34,8 +34,8 @@ use crate::{
 };
 
 /// Addresses that windbg will attempt to read in a loop, reads from these addresses
-/// will just return 0 to avoid long retry delays.
-#[cfg(feature = "windbg_workarounds")]
+/// will just return 0 to avoid long retry delays. This has been fixed in windbg, but leaving this
+/// here for a while to support older versions of windbg. Will remove in the future.
 const WINDBG_MOCK_ADDRESSES: [u64; 3] = [0xfffff78000000268, 0, 0x34c00];
```
makubacki, line 39: Technically, this will make non-windbg users now get zeroes back instead of whatever was happening before (e.g. page fault) for null pointer dereferences, right?

core/patina_debugger/src/dbg_target.rs
makubacki, line 1: The `windbg_workarounds` feature is still mentioned in lib.rs -https://github.com/OpenDevicePartnership/patina/blob/087ebd8f33c46b482a29c5be1374258d7950be1d/core/patina_debugger/src/lib.rs#L94

core/patina_debugger/src/dbg_target.rs
cfernald: Correct, I'm treating this as acceptable for the time being, and will remove this workaround all-up after the next windbgx release. The behavior observed is that memory reads that previously would return an error (wouldn't fault the debugger or anything since it checks mappings) would return 0 instead.

patina_dxe_core/Cargo.toml
cfernald: I clarified the intent in the readme.

core/patina_debugger/src/dbg_target.rs
cfernald: Though its worth mentioning there is no actual difference, since the way the feature was handled meant that platforms couldn't actually disable it always, since this feature was included by the core.

PR #1375, opened 2026-03-06
[feature/patina-boot] patina_boot: Implement full console discovery

### LINKED ISSUE #1230: [Task] patina_boot: Complete ConsoleDiscovery implementation
## Description

Complete the `ConsoleDiscovery` component implementation. Currently a stub that locates handles but doesn't write variables.

## Requirements

- Locate GOP protocol handles and create device paths
- Locate SimpleTextInput protocol handles and create device paths
- Write `ConIn`, `ConOut`, and `ErrOut` UEFI variables via RuntimeVariableServices
- Support multi-instance device paths for multiple consoles

## Current State

The current implementation in `helpers::discover_console_devices()` locates handles but does not write the console variables. This is documented as a stub.

## Related

kat-perez:
## Description

Rewrite `discover_console_devices()` from a stub into a full implementation that enumerates console protocol handles, builds multi-instance device paths, and writes `ConIn`, `ConOut`, and `ErrOut` UEFI global variables via `SetVariable`.

- Adds `EFI_GLOBAL_VARIABLE` GUID to `patina::guids`
- Adds `build_multi_instance_device_path()` helper for constructing multi-instance device paths from protocol handles
- Updates `is_partial_device_path()` to recognize FV/FvFile paths as non-partial
- Includes get_variable readback verification with device path display logging

Depends on #1333. Closes #1230 

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Verified on QEMU Q35 with VGA enabled. Console variables written and read back successfully:
- ConIn: 24 bytes (SimpleTextInput)
- ConOut: 60 bytes (SimpleTextOutput + GOP)
- ErrOut: 30 bytes (SimpleTextOutput)

## Integration Instructions

N/A

### DIFF
--- a/components/patina_boot/src/helpers.rs
+++ b/components/patina_boot/src/helpers.rs
@@ -18,14 +18,18 @@ use core::ptr;
 use patina::{
     boot_services::{BootServices, event::EventType, protocol_handler::HandleSearchType, tpl::Tpl},
     device_path::{
-        node_defs::DevicePathType,
+        node_defs::{DevicePathType, HardDrive, MediaSubType},
         paths::{DevicePath, DevicePathBuf},
     },
     error::{EfiError, Result},
-    guids::EVENT_GROUP_END_OF_DXE,
+    guids::{EFI_GLOBAL_VARIABLE, EVENT_GROUP_END_OF_DXE},
     runtime_services::RuntimeServices,
 };
-use r_efi::{efi, protocols::simple_text_input, system::EVENT_GROUP_READY_TO_BOOT};
+use r_efi::{
+    efi,
+    protocols::simple_text_input,
+    system::{EVENT_GROUP_READY_TO_BOOT, VARIABLE_BOOTSERVICE_ACCESS, VARIABLE_NON_VOLATILE, VARIABLE_RUNTIME_ACCESS},
+};
 
 /// Watchdog timeout in seconds per UEFI Specification Section 3.1.2.
 const WATCHDOG_TIMEOUT_SECONDS: usize = 300; // 5 minutes
@@ -67,7 +71,6 @@ pub fn detect_hotkey<B: BootServices>(boot_services: &B, hotkey_scancode: u16) -
 ///
 /// - `handles` must contain valid handles obtained from `locate_handle_buffer`
 /// - Each handle must support the `SimpleTextInput` protocol for `handle_protocol` to succeed
-#[coverage(off)] // Uses raw protocol function pointers - tested via integration tests
 unsafe fn detect_hotkey_from_handles<B: BootServices>(
     boot_services: &B,
     handles: &[efi::Handle],
@@ -167,12 +170,6 @@ pub fn boot_from_device_path<B: BootServices>(
 ///
 /// Returns `Ok(())` when device topology enumeration is complete.
 ///
-/// # Coverage
-///
-/// This function is marked `#[coverage(off)]` because `BootServicesBox` return
-/// values from `locate_handle_buffer` cannot be created in unit tests. The
-/// function is tested via integration tests on real hardware/emulators.
-#[coverage(off)] // BootServicesBox return type cannot be mocked - tested via integration tests
 pub fn connect_all<B: BootServices>(boot_services: &B) -> Result<()> {
     // Loop until the number of handles stabilizes, indicating device topology is complete.
     // This is needed because connecting a PCI bus creates new handles for PCI devices,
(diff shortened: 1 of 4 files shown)


makubacki:
@kat-perez, should https://github.com/OpenDevicePartnership/patina/issues/1230 be linked to this PR?

kat-perez:
> @kat-perez, should #1230 be linked to this PR?

yes, that's the issue I was working on

components/patina_boot/src/helpers.rs
```diff
@@ -268,37 +272,130 @@ pub fn signal_ready_to_boot<B: BootServices>(boot_services: &B) -> Result<()> {
...
+
+        if let Some(dp) = device_path {
+            let dp_ref: &DevicePath = dp.as_ref();
+            // SAFETY: DevicePath is #[repr(transparent)] over [u8], so a pointer to
+            // DevicePath is a pointer to the underlying byte buffer.
+            let bytes = unsafe { core::slice::from_raw_parts(dp_ref as *const DevicePath as *const u8, dp_ref.size()) };
+
+            match runtime_services.set_variable(name, &EFI_GLOBAL_VARIABLE, attrs, &bytes) {
+                Ok(()) => {
+                    log::info!("{label}: set ({} bytes)", bytes.len());
```
os-d: nit: not needed

components/patina_boot/src/helpers.rs
```diff
@@ -268,37 +272,130 @@ pub fn signal_ready_to_boot<B: BootServices>(boot_services: &B) -> Result<()> {
...
+
+                    // Readback verification: read the variable back and parse as device path
+                    match runtime_services.get_variable::<Vec<u8>>(name, &EFI_GLOBAL_VARIABLE, None) {
```
os-d: Should we just trust the variable services to report correctly on whether they did this correctly or not? Perhaps this was some testing code left in? If we have a legitimate concern about this, I would consider writing a patina_test instead of doing it in the actual code.

components/patina_boot/src/helpers.rs
```diff
@@ -268,37 +272,130 @@ pub fn signal_ready_to_boot<B: BootServices>(boot_services: &B) -> Result<()> {
...
+/// Build a multi-instance device path from all handles supporting the given protocols.
+///
+/// For each protocol GUID, locates all handles via `locate_handle_buffer` and extracts
+/// their device paths, combining them into a single multi-instance device path separated
+/// by `EndInstance` nodes.
+///
+/// # Safety
+///
+/// - Handles from `locate_handle_buffer` must be valid
+/// - `handle_protocol` must return valid device path protocol pointers
```
os-d: safety comment should be the contract with the caller of this function, not the safety of operations within the function. Which, it looks like the function has safe inputs and handles failures, it could be a safe fn?

components/patina_boot/src/helpers.rs
```diff
@@ -368,39 +475,98 @@ pub fn expand_device_path<B: BootServices>(boot_services: &B, partial_path: &Dev
...
+    const SIG_LEN: usize = 16;
+    const SIG_TYPE_OFFSET: usize = SIG_OFFSET + SIG_LEN + 1; // 37
+
+    let target_sig = partial_path.iter().find_map(|node| {
+        if node.header.r#type == DevicePathType::Media as u8
+            && node.header.sub_type == MediaSubType::HardDrive as u8
+            && node.data.len() > SIG_TYPE_OFFSET
+        {
+            let sig = &node.data[SIG_OFFSET..SIG_OFFSET + SIG_LEN];
+            let sig_type = node.data[SIG_TYPE_OFFSET];
```
os-d: The device path reorg has happened, this should go in that crate? Can the different device path types have safer operations to get signature, etc.? e.g. hw_drive_node.get_signature() which then abstracts this away? Only thinking about if the device path is malformed that SIG_OFFSET+SIG_LEN walks off the end of the slice, we have the size info so we could be safely handling that?

os-d (approved):

components/patina_boot/src/helpers.rs
kat-perez: removed

components/patina_boot/src/helpers.rs
kat-perez: Yeah, that was some testing code left in. Removed it

components/patina_boot/src/helpers.rs
kat-perez: I changed this to a safe fn instead with `handle_protocol` and `DevicePath::try_from_ptr` scoped in local unsafe blocks.

components/patina_boot/src/helpers.rs
kat-perez: Replaced the raw offset-based parsing with scroll::Pread to deserialize into the typed HardDrive struct, and use HardDrive::is_type() for node matching.

Did you mean we should also add something like a try_from_node() accessor directly on the device path types in the device path crate? Can do that as a follow-up if so.

os-d: Yeah, either that or the node has a new function that validates this is a valid type and so when we do offset dereferencing, it is safe (maybe this is already there, I didn't check). Even better than doing offset dereferencing would be for each type of device path to have accessors for the given part, e.g. HwNode.signature() gets the caller the signature without having to know about the offsets, etc.

kat-perez: So I added `HardDrive::try_from_node()` and then dropped the `scroll` dep too

components/patina_boot/src/helpers.rs
```diff
@@ -368,39 +448,75 @@ pub fn expand_device_path<B: BootServices>(boot_services: &B, partial_path: &Dev
...
+        for node in handle_path.iter() {
+            if let Some(hd) = HardDrive::try_from_node(&node)
+                && hd.partition_signature == target_sig.as_slice()
+                && hd.signature_type == target_sig_type
+            {
+                // Found a match. Build the full path from the handle's device path
+                // and append any remaining nodes from the partial path (e.g., FilePath).
+                let mut result = DevicePathBuf::from(handle_path);
+                let remaining_dp = DevicePathBuf::from_device_path_node_iter(remaining_nodes.into_iter());
+                result.append_device_path(&remaining_dp);
```
makubacki: Can this potentially append full path that ends in a file with another file path? Can `DevicePathBuf::from(handle_path)` return a path that ends in FilePath, for example, and `remaining_dp` append a file path?

Asking because, it's a little difficult to trace resulting paths for sure.

components/patina_boot/src/helpers.rs
```diff
@@ -268,37 +272,103 @@ pub fn signal_ready_to_boot<B: BootServices>(boot_services: &B) -> Result<()> {
...
 }
 
+/// Build a multi-instance device path from all handles supporting the given protocols.
+///
+/// For each protocol GUID, locates all handles via `locate_handle_buffer` and extracts
+/// their device paths, combining them into a single multi-instance device path separated
+/// by `EndInstance` nodes.
+///
+#[coverage(off)] // Uses raw protocol pointers - tested via integration tests
+fn build_multi_instance_device_path<B: BootServices>(
```
makubacki, line 324: Does this function handle the case when a single handle supports multiple protocols given? For example, `ConOut` might support both simple text output and GOP. In that case, we wouldn't the `ConOut` device path redundantly returned, right?

components/patina_boot/src/helpers.rs
```diff
@@ -268,37 +272,103 @@ pub fn signal_ready_to_boot<B: BootServices>(boot_services: &B) -> Result<()> {
...
+    ];
+
+    for &(label, name, guids) in console_vars {
+        let device_path = build_multi_instance_device_path(boot_services, guids);
+
+        if let Some(dp) = device_path {
+            let dp_ref: &DevicePath = dp.as_ref();
+            // SAFETY: DevicePath is #[repr(transparent)] over [u8], so a pointer to
+            // DevicePath is a pointer to the underlying byte buffer.
+            let bytes = unsafe { core::slice::from_raw_parts(dp_ref as *const DevicePath as *const u8, dp_ref.size()) };
```
makubacki: Since `DevicePath` is transparent over `[u8]`, did you try `.as_bytes()` (possibly eliminating the unsafe block)?

components/patina_boot/src/helpers.rs
```diff
@@ -268,37 +272,103 @@ pub fn signal_ready_to_boot<B: BootServices>(boot_services: &B) -> Result<()> {
...
+/// continues with the remaining variables.
 ///
 /// # Arguments
 ///
-/// * `boot_services` - Boot services interface
-/// * `runtime_services` - Runtime services interface (unused in stub)
-#[allow(unused_variables)]
+/// * `boot_services` - Boot services interface for handle enumeration
+/// * `runtime_services` - Runtime services interface for writing UEFI variables
+#[coverage(off)] // Uses raw protocol pointers - tested via integration tests
```
makubacki: Can you please update these messages justifying turning off test coverage to be a little more precise? For example, I don't think raw pointers are so much the problem as mocking limitations, right?

I don't want to set a false precedent in the codebase that simply using raw pointers gets code out of requiring unit tests.

components/patina_boot/src/helpers.rs
kat-perez: This won't produce a double `FilePath`. Handle paths come from partition handles whose device paths end in `HD()` node. Remaining nodes are the nodes after `HD()` like: `FilePath(\EFI\Boot\BOOTX64.efi)`. Should I add a doc comment in the code to explain/trace the paths better?

components/patina_boot/src/helpers.rs
kat-perez: Updated to avoid duplication by tracking seen handles

components/patina_boot/src/helpers.rs
kat-perez: I didn't, but updated it to `.as_bytes()` now and removed the unsafe block

components/patina_boot/src/helpers.rs
makubacki: Okay. As far as messages go, maybe just a message that clarifies where this is guaranteed would help:

> "Handle paths come from partition handles whose device paths end in HD() node"

---

My understanding was that the code under `// Enumerate all handles with DevicePath protocol` would return all device paths in the system including filesystem device paths that include but do not end in a `HD()` node.

In which case, code under `// Search for a handle whose device path contains a matching HardDrive node` would find a matching HardDrive node and this:

```rust
let mut result = DevicePathBuf::from(handle_path);
```

Would have `result` set to that filesystem path so the append operation could potentially append more filesystem nodes from `remaining_nodes`.

kat-perez: I see, yeah you're right. I assumed that `locate_handle_buffer` would only return partition handle buffers and not filesystem handles. But since the search is `ByProtocol(device_path::PROTOCOL_GUID)`, that would return every handle with the device path and not just partition handles.

Fixed this by truncating the handle's device path at the matching node before appending remaining_nodes. Also added a unit test that generically tests for this issue and that expansion should truncate at the matched node. This should work for future partial device paths too

components/patina_boot/src/helpers.rs
kat-perez: Removed these incorrect `coverage_off` and replaced with the correct mocking mechanisms

PR #1377, opened 2026-03-06
[Major Branch][REBASE&FF] Patina Test framework changes 

Javagedes:
## Description

This commit chain makes two breaking changes to the Patina Test framework:

1. Updates the `.with_filter` function to accept a `Filter` enum instead of a string, to better enable filtering multiple ways (right now only allow list and deny list)
2. Moves `patina_test` framework out of the `patina` crate and into it's own patina component crate. As a part of this change, the `build-lib` makefile command has been updated to pass `-all-features`. This change has already been pushed on main in patina-devops.

<details>
<summary> Commit 1 (Click to expand commit message) </summary>
Add exclude filter support to TestRunner
Introduce a `Filter` enum with `Include` and `Exclude` variants rather
than a generic string that is only an inclusion filter.

Exclude filters take priority over include filters. If no include
filters are specified, all non-excluded tests run. The match arms in
`should_run` are exhaustive to ensure new variants cause a compile break.
</details>

<details>
<summary> Commit 2 (Click to expand commit message) </summary>
move patina_test framework to patina_test crate
The patina_test framework was originally developed inside the SDK crate
to keep `patina_macro` as only a dependency for the `patina` crate. As
we have evolved, other patina related crates have taken a dependency on
`patina_macro`, which means the patina_test framework can be it's own
crate.

This has one large benefit: It simplfies feature logic related to
actually enable or disable the tests. There is now a single feature flag
that is off by default (`component`), and this is how test writers will
consume the crate. When a platform wants to register a test runner with
the Core, it will enable the `component` feature flag, which does two
things: (1) Make the `TestRunner` component public and (2) enable a
feature flag in `patina_macro` to actually register all found patina
tests globally.

The second, smaller benefit, is that we are moving some of the non-sdk
logic out of our SDK crate, and this framework makes sense as a
standalone crate.
</details>

- [ ] Impacts functionality?
- [ ] Impacts security?
- [x] Breaking change?
- [x] Includes tests?
- [x] Includes documentation?

## How This Was Tested

CI, Platform verification

### `.with_filter` update

```rust
// Previous
TestRunner::default()
  .with_filter("whitelisted::filter")

// New
TestRunner::default()
  .with_filter(Filter::include("include::this"))
  .with_filter(Filter::exclude("exclude::this"))
```

### New crate update

```rust
// Previous
use patina::test::{TestRunner, patina_test, u_assert, u_assert_eq, u_assert_ne, Result};

// New
use patina_test::{component::TestRunner, error::Result, u_assert, u_assert_eq, u_asssert_ne, patina_test};
```

```toml
# Old
[dependencies]
patina = { features = "enable_patina_tests"] }

# new
[dependencies]
patina_test = { features = ["component"] }
```


components/patina_test/src/component.rs
```diff
@@ -0,0 +1,401 @@
+//! Pastina Test Test Runner Component
```
makubacki:
```suggestion
//! Patina Test Test Runner Component
```

components/patina_test/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+[dependencies]
+linkme = { workspace = true }
+log = { workspace = true }
+patina = { workspace = true }
+patina_macro = { workspace = true }
+r-efi = { workspace = true }
+spin = { workspace = true }
+
+[features]
+component = ["patina_macro/enable_patina_tests"]
```
makubacki: Do you think `component` is a better name for this feature than something like `runner` or `test-runner`?

components/patina_test/src/component.rs
```diff
@@ -0,0 +1,401 @@
...
+///
+/// # Example
+///
+/// ```rust
+/// use patina_test::{TestRunner, Filter};
+///
+/// let runner = TestRunner::default()
+///     .with_filter(Filter::include("x64"))
+///     .with_filter(Filter::exclude("aarch64"));
+/// ```
```
makubacki, line 32: `cargo test --doc -p patina_test --features component` fails with:


```
test components\patina_test\src\component.rs - component::Filter (line 26) ... FAILED
test components\patina_test\src\lib.rs - (line 43) ... ok

failures:

---- components\patina_test\src\component.rs - component::Filter (line 26) stdout ----
error[E0432]: unresolved imports `patina_test::TestRunner`, `patina_test::Filter`
  --> components\patina_test\src\component.rs:28:19
   |
28 | use patina_test::{TestRunner, Filter};
   |                   ^^^^^^^^^^  ^^^^^^ no `Filter` in the root
   |                   |
   |                   no `TestRunner` in the root
   |
   = help: consider importing this struct instead:
           patina_test::component::TestRunner
   = help: consider importing one of these items instead:
           std::iter::Filter
           patina_test::component::Filter
```


```suggestion
/// ```rust
/// use patina_test::component::{TestRunner, Filter};
///
/// let runner = TestRunner::default()
///     .with_filter(Filter::include("x64"))
///     .with_filter(Filter::exclude("aarch64"));
/// ```
```

Makefile.toml
```diff
@@ -85,7 +85,7 @@ Example:
 """
 clear = true
 command = "cargo"
-args = ["build", "@@split(INDIVIDUAL_PACKAGE_TARGETS, )", "@@split(STD_FLAGS, )"]
+args = ["build", "--all-features", "@@split(INDIVIDUAL_PACKAGE_TARGETS, )", "@@split(STD_FLAGS, )"]
```
makubacki, line 88: I may have missed it, but I didn't see this `build-lib` task change mentioned in the PR/commit description. It would be good to call out.

components/patina_test/src/service.rs
```diff
@@ -0,0 +1,384 @@
...
+
+        if let Some(recorder) = storage.get_service::<Recorder>() {
+            let _ = recorder.with_mut(|records| records.get_mut(test).map(|record| record.run(storage)));
+        }
+    }
+
+    #[coverage(off)]
+    /// An EFIAPI compatible event callback to disable a timer event at ReadyToBoot
+    extern "efiapi" fn disable_timer(rtb_event: r_efi::efi::Event, context: *mut core::ffi::c_void) {
+        // SAFETY: We set up the context pointer in `run_tests` to point to a valid tuple of (Event, &mut Storage).
```
makubacki: I know that this is being copied from pre-existing code, but the tuple in the comment is wrong.

```suggestion
        // SAFETY: We set up the context pointer in `run_tests` to point to a valid tuple of (Event, StandardBootServices).
```

components/patina_test/src/component.rs
Javagedes: lol. Thanks

components/patina_test/Cargo.toml
Javagedes: I named it `component` since it makes the `component` module (which is the naming scheme for component crates) public.

However, I'm not against changing it if you prefer. If you want it changed and I had to select from the two above, I will lean towards test-runner.

makubacki: I thought `component` made sense from the internal perspective of the crate, but `component` is already be overloaded enough that something different might be clearer as a feature name. I don't have a strong opinion though.

Javagedes: Updated the feature flag to `test-runner`!

PR #1378, opened 2026-03-06
patina_dxe_core/pecoff: Add parse_mapped()

makubacki:
## Description

Resolves https://github.com/OpenDevicePartnership/patina-dxe-core-qemu/issues/134

Right now, we have `resolve_rva` set to true when parsing the DXE Core using `UefiPeInfo::parse()`.

https://github.com/OpenDevicePartnership/patina/blob/dc285f1dffa57087a72f6bfd956d36db93e8b31f/patina_dxe_core/src/gcd/spin_locked_gcd.rs#L2317-L2323

Which calls with `goblin::pe::PE::parse(bytes)` to parse the image.

By default, goblin will attempt to resolve RVAs to file offsets when parsing a PE file. However, since the DXE Core is already loaded into memory, we want to parse it, in this case, without resolving RVAs using PointerToRawData offsets.

Depending on the content at the resolved address, this can lead to parsing errors like:

```
Failed to parse PE info for DXE Core:
Goblin(Malformed("ImageDebugDirectory size of data seems wrong: 0"))
```

It appears that this has been present for a long time, likely since the decision (around c821bf9c) to start parsing the DXE Core's own loaded PE image and a recent shift in linked contents has exposed the problem.

It is possible to disable RVA resolution by using `ParseOptions` with `resolve_rva` set to false.

A new method `parse_mapped()` is added to `UefiPeInfo` that allows sets this option and is used when parsing the DXE Core in `spin_locked_gcd.rs`.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- Q35 and SBSA boot to EFI shell

Before the change, SBSA failed with the goblin parsing error shown in the description.

## Integration Instructions

- N/A

### DIFF
--- a/patina_dxe_core/src/gcd/spin_locked_gcd.rs
+++ b/patina_dxe_core/src/gcd/spin_locked_gcd.rs
@@ -2315,7 +2315,7 @@ impl SpinLockedGcd {
             .expect("Did not find MemoryAllocationModule Hob for DxeCore. Use patina::guid::DXE_CORE as FFS GUID.");
 
         let pe_info = unsafe {
-            UefiPeInfo::parse(core::slice::from_raw_parts(
+            UefiPeInfo::parse_mapped(core::slice::from_raw_parts(
                 dxe_core_hob.alloc_descriptor.memory_base_address as *const u8,
                 dxe_core_hob.alloc_descriptor.memory_length as usize,
             ))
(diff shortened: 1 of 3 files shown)


makubacki:
Note: The QEMU validation hang issue is unrelated and tracked in https://github.com/OpenDevicePartnership/patina-dxe-core-qemu/issues/135.

PR #1383, opened 2026-03-10
patina_smbios: Add SMBIOS types

ansleythompson:
## Description
Define types for SMBIOS records according the SMBIOS specs 3.0+ and update records to use types. These structs/enums/bitfields force typing when creating SMBIOS records

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Integrated into patina-dxe-core-qemu with a local patch and verified with patina-qemu smbiosview

## Integration Instructions
Use Smbios Types in place of magic numbers to create SMBIOS records.

### DIFF
--- a/components/patina_smbios/Cargo.toml
+++ b/components/patina_smbios/Cargo.toml
@@ -16,6 +16,7 @@ r-efi = { workspace = true }
 spin = { workspace = true }
 zerocopy = { workspace = true }
 zerocopy-derive = { workspace = true }
+bitfield = "0.17.0"
 
 [dev-dependencies]
 mockall = { workspace = true }
--- a/components/patina_smbios/src/smbios_types.rs
+++ b/components/patina_smbios/src/smbios_types.rs
@@ -1,0 +1,1681 @@
+//! SMBIOS Types
+//!
+//! Defines the types used in SMBIOS Records 
+//!
+//! ## License
+//!
+//! Copyright (c) Microsoft Corporation.
+//!
+//! SPDX-License-Identifier: Apache-2.0
+
+use bitfield::bitfield;
+extern crate alloc;
+
+
+bitfield! {
+    ///
+    /// BIOS Characteristics
+    /// 
+    pub struct BiosCharacteristics (u64);
+    impl Debug;
+    pub reserved, set_reserved: 1, 0;
+    pub unknown, set_unknown: 2;
+    pub bios_characteristics_unsupported, set_bios_characteristics_unsupported: 3;
+    pub isa_supported, set_isa_supported: 4;
+    pub mca_supported, set_mca_supported: 5;
+    pub eisa_supported, set_eisa_supported: 6;
+    pub pci_supported, set_pci_supported: 7;
+    pub pcmcia_supported, set_pcmcia_supported: 8;
+    pub plug_play_supported, set_plug_play_supported: 9;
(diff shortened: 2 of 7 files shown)


components/patina_smbios/src/smbios_types.rs
```diff
@@ -0,0 +1,1681 @@
...
+    ProcessorMemoryModule,
+    ProcessorIoModule,
+    InterconnectBoard
+}
+
+///
+/// Boot Up State
+///
+#[derive(Copy, Clone, Debug)]
+pub enum BootUpState {
```
kat-perez, line 299: These values are Chassis Types (SMBIOS spec Table 17 — System Enclosure or Chassis Types), not boot-up states. Boot-up states should be
```
Other = 0x01,
Unknown,
Safe,
Warning,
Critical,
...
```

components/patina_smbios/src/smbios_types.rs
```diff
@@ -0,0 +1,1681 @@
...
+    EmbeddedPc,
+    MiniPc,
+    StickPc
+}
+
+///
+/// Power Supply State
+///
+#[derive(Copy, Clone, Debug)]
+pub enum PowerSupplyState {
```
kat-perez, line 342: I think these values look like ProcessorType values instead of PowerSupplyState Values

sdk/patina_macro/src/smbios_record_macro.rs
```diff
@@ -142,14 +142,60 @@ pub(crate) fn smbios_record_derive(item: TokenStream) -> TokenStream {
                         matches!(
                             segment.ident.to_string().as_str(),
                             "u8" | "u16" | "u32" | "u64" | "i8" | "i16" | "i32" | "i64"
+                            | "BiosCharacteristics"
```
kat-perez, line 145: I think we could use `zerocopy::IntoBytes` here instead. So we could have something like:

```
#[bitfield(u8)]
#[derive(zerocopy::IntoBytes)]
pub struct BiosCharacteristicsExt1 {
    pub acpi_supported: bool,
    pub usb_legacy_supported: bool,
    // ...
```

I can refactor the smbios_record_macro to use zerocopy::IntoBytes::as_bytes() instead of the type name string list in a separate PR though. Will update you here when that's created

components/patina_smbios/Cargo.toml
```diff
@@ -16,6 +16,7 @@ r-efi = { workspace = true }
 spin = { workspace = true }
 zerocopy = { workspace = true }
 zerocopy-derive = { workspace = true }
+bitfield = "0.17.0"
```
kat-perez, line 19: Can you use bitfield-struct from the workspace? If not, can you add it to the workspace and then use `bitfield = { workspace = true }` here?

sdk/patina_macro/src/smbios_record_macro.rs
ansleythompson: Sounds good, I can update the bitfields.

kat-perez: https://github.com/OpenDevicePartnership/patina/commit/0f014bc88881106713a164cdf22d5e045beadc5c PR is on main now

kat-perez:
New PR here #1501

PR #1388, opened 2026-03-11
Use Xorshift64starHasher for SMBIOS table modification detection

### LINKED ISSUE #1161: [Task] Use Xorshift64starHasher for SMBIOS table modification detection
# Summary

Replace the current simple checksum in SmbiosManager with Xorshift64starHasher for detecting direct table modifications.

# Background

The current implementation uses a basic checksum to detect if the SMBIOS table was modified directly (bypassing the manager). While this works for most cases, it can miss edge cases where multiple fields change in ways that cancel each other out (e.g., FieldA: 1→0 and FieldB: 0→1 would produce the same checksum).

# Proposed Change

Use the existing Xorshift64starHasher from protocol_db.rs:

```
let mut hasher = Xorshift64starHasher::default();
hasher.write(data);
cksum = hasher.finish() as u32; // truncate and keep bottom 32-bits of hash
```

Tasks

  - Move Xorshift64starHasher to a shared location for general use (e.g., patina::base or a new patina::hash module)
  - Update SmbiosManager to use the hasher instead of simple checksum
  - Update any tests that verify checksum behavior

  Related

  This enhancement was identified during PR #1147 (TplMutex ownership refactor).

kat-perez:
## Description

Move `Xorshift64starHasher` from `patina_dxe_core` to a shared `patina::hash` module and replace the simple wrapping-add checksum in `SmbiosManager` with the hasher. This avoids false negatives when multiple byte changes cancel each other out.

Closes #1161

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Existing tests in `patina_smbios` (134 passed) and `patina_dxe_core` (541 passed) continue to pass. New unit tests added in `patina::hash` covering seed behavior, determinism, and byte-swap detection.

## Integration Instructions

N/A

### DIFF
--- a/components/patina_smbios/src/manager/core.rs
+++ b/components/patina_smbios/src/manager/core.rs
@@ -85,7 +85,7 @@ pub struct SmbiosManager {
     /// Pre-allocated buffer for entry point structure
     ep_buffer_addr: RefCell<Option<PhysicalAddress>>,
     /// Checksum of published table (for detecting direct modifications)
-    published_table_checksum: RefCell<Option<u32>>,
+    published_table_checksum: RefCell<Option<u64>>,
     /// Size of the published table (needed for checksum verification)
     published_table_size: RefCell<usize>,
 }
@@ -470,12 +470,15 @@ impl SmbiosManager {
         self.table_64_address.replace(Some(table_address));
     }
 
-    /// Calculate a simple checksum for table data
+    /// Calculate a hash for table data using Xorshift64*
     ///
-    /// Uses a simple sum of all bytes (wrapping). This is sufficient for
-    /// detecting accidental modifications, not cryptographic integrity.
-    fn calculate_table_checksum(data: &[u8]) -> u32 {
-        data.iter().fold(0u32, |acc, &byte| acc.wrapping_add(byte as u32))
+    /// Uses Xorshift64starHasher to detect modifications including byte swaps
+    /// that a simple checksum would miss. Not for cryptographic integrity.
+    fn calculate_table_checksum(data: &[u8]) -> u64 {
+        use core::hash::Hasher;
+        let mut hasher = patina::hash::Xorshift64starHasher::default();
+        hasher.write(data);
+        hasher.finish()
     }
 
     /// Verify that the published table has not been modified directly
--- a/sdk/patina/src/hash.rs
+++ b/sdk/patina/src/hash.rs
@@ -1,0 +1,121 @@
+//! Hash utilities for Patina
+//!
+//! Provides hash implementations for general use across Patina components.
+//!
+//! ## License
+//!
+//! Copyright (c) Microsoft Corporation.
+//!
+//! SPDX-License-Identifier: Apache-2.0
+
+use core::hash::Hasher;
+
+/// A hasher that uses the Xorshift64* algorithm to generate a random number to xor with the input bytes.
+///
+/// **Note:** This hasher is not cryptographically secure. It is intended for fast, non-adversarial
+/// hash-based change detection only.
+///
+/// [Xorshift64*](https://en.wikipedia.org/wiki/Xorshift#xorshift*)
+pub struct Xorshift64starHasher {
+    state: u64,
+}
+
+impl Xorshift64starHasher {
+    /// Initialize the hasher with a seed.
+    pub fn new(seed: u64) -> Self {
+        Xorshift64starHasher { state: seed }
+    }
+
+    /// Generate a new random state.
+    fn next_state(&mut self) -> u64 {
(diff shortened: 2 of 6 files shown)


sdk/patina/src/hash.rs
```diff
@@ -0,0 +1,115 @@
+//! Hash utilities for Patina
+//!
+//! Provides hash implementations for general use across Patina components.
+//!
+//! ## License
+//!
+//! Copyright (C) Microsoft Corporation.
```
os-d:
```suggestion
//! Copyright (c) Microsoft Corporation.
```

components/patina_smbios/src/manager/core.rs
```diff
@@ -470,12 +470,15 @@ impl SmbiosManager {
...
-    /// Uses a simple sum of all bytes (wrapping). This is sufficient for
-    /// detecting accidental modifications, not cryptographic integrity.
+    /// Uses Xorshift64starHasher to detect modifications including byte swaps
+    /// that a simple checksum would miss. Not for cryptographic integrity.
     fn calculate_table_checksum(data: &[u8]) -> u32 {
-        data.iter().fold(0u32, |acc, &byte| acc.wrapping_add(byte as u32))
+        use core::hash::Hasher;
+        let mut hasher = patina::hash::Xorshift64starHasher::default();
+        hasher.write(data);
+        hasher.finish() as u32
```
os-d: Is this okay to do? Is this an internal only hash? Or is this defined in the SMBIOS table structure? If so, and that specifies a checksum, we can't change that, right?

sdk/patina/src/hash.rs
makubacki, line 1: Long-term if this module is going to accumulate various hash implementations, it might be cleaner to put each of those in their own file and re-export from here.

sdk/patina/src/hash.rs
```diff
@@ -0,0 +1,115 @@
...
+        self.state ^= self.state << 25;
+        self.state ^= self.state >> 27;
+        self.state = self.state.wrapping_mul(0x2545F4914F6CDD1D);
+        self.state
+    }
+}
+
+impl Default for Xorshift64starHasher {
+    fn default() -> Self {
+        Xorshift64starHasher::new(compile_time::unix!())
```
makubacki, line 44: My understanding is that this will result in a single seed used for all `Xorshift64startHasher` instances in the binary.

Can you please update the doc comments for this module/type to call this out?

sdk/patina/src/hash.rs
```diff
@@ -0,0 +1,115 @@
...
+//!
+//! Copyright (C) Microsoft Corporation.
+//!
+//! SPDX-License-Identifier: Apache-2.0
+
+use core::hash::Hasher;
+
+/// A hasher that uses the Xorshift64* algorithm to generate a random number to xor with the input bytes.
+///
+/// https://en.wikipedia.org/wiki/Xorshift#xorshift*
```
makubacki: Can you please mention in the doc comment here that this is not safe for cryptographic usage so that disclaimer is closer to the type?

sdk/patina/src/hash.rs
kat-perez: Added

sdk/patina/src/hash.rs
kat-perez: Yep, added to doc comments

components/patina_smbios/src/manager/core.rs
kat-perez: This is for internal use only and is never written to the SMBIOS table. This hash is purely a runtime check to see if someone modified the table outside of the published `Remove()` and `Add()` APIs

sdk/patina/src/hash.rs
kat-perez: Right now there's only the one implementation so I'll leave it as-is, but if we add more I'll split them into separate files under a hash/ module directory and re-export from mod.rs.

sdk/patina/src/hash.rs
kat-perez: Updated

sdk/patina/src/hash.rs
makubacki: Sounds good. Should be obvious enough then.

components/patina_smbios/src/manager/core.rs
makubacki: Do you really need to truncate the u64 bit hash returned to u32 here? I know this hash doesn't matter "that much", but it still seems unnecessary.

kat-perez: Not really required to truncate it -- changed it to u64

PR #1393, opened 2026-03-12
Minor Patina ACPI test fixes

berlin-with0ut-return:
## Description
Patina tests now run after installation of ACPI tables. As such, we cannot use core tables (such as FADT, FACS) in integration tests as they will override critical OS tables.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [X] Includes tests?
- [ ] Includes documentation?

## How This Was Tested
Tests now pass + boots correctly when `enable_patina_tests` is turned on.

## Integration Instructions
N/A.

### DIFF
--- a/components/patina_acpi/src/integration_test.rs
+++ b/components/patina_acpi/src/integration_test.rs
@@ -139,13 +139,13 @@ fn acpi_protocol_test(bs: StandardBootServices) -> patina::test::Result {
             &mut get_table_key,
         );
         // We should be able to find our installed table.
-        // SAFETY: `table_buf` is valid and directly constructed from the dummy table.
         if get_result != efi::Status::SUCCESS {
             // If fails, either hit an error on a previous table or reached the end of the list without finding the installed table.
             // Both are error cases.
             u_assert!(false, "Get table should succeed for installed table");
         }
 
+        // SAFETY: `table_buf` is valid and directly constructed from the dummy table.
         if unsafe { (*table_buf).signature } == 0x12341234 {
             break;
         }


makubacki:
@berlin-with0ut-return, thank you for making this change! Please hold the PR for me. I will review it on Friday.

components/patina_acpi/src/integration_test.rs
```diff
@@ -18,57 +18,86 @@ use patina::{
...
-    let facs = AcpiFacs { signature: signature::FACS, length: mem::size_of::<AcpiFacs>() as u32, ..Default::default() };
     // SAFETY: The constructed table is a valid ACPI table.
-    assert!(unsafe { table_manager.install_acpi_table(facs) }.is_ok(), "Should install FACS table.");
+    let key2 = unsafe { table_manager.install_acpi_table(mock_table2) }.expect("Should install table.");
+
+    // Install an invalid ACPI table (too small).
+    let invalid_table =
+        AcpiTableHeader { signature: signature::MADT, length: (signature::MADT_SIZE - 2) as u32, ..Default::default() };
+    // SAFETY: The constructed table is not a valid ACPI table, so this should return an error.
+    assert!(unsafe { table_manager.install_acpi_table(invalid_table) }.is_err(), "Should not install invalid table.");
```
os-d: I think we want to use the u_assert family of macros here so that the patina_test framework records things correctly.

components/patina_acpi/src/integration_test.rs
```diff
@@ -81,45 +110,58 @@ fn acpi_protocol_test(bs: StandardBootServices) -> patina::test::Result {
...
+            &mut get_table_key,
+        );
+        // We should be able to find our installed table.
+        // SAFETY: `table_buf` is valid and directly constructed from the dummy table.
+        if unsafe { (*table_buf).signature } == 0x12341234 {
+            break;
+        } else if get_result != efi::Status::SUCCESS {
+            // If fails, either hit an error on a previous table or reached the end of the list without finding the installed table.
+            // Both are error cases.
+            panic!("Get table should succeed for installed table");
```
os-d: Same here, I’m not sure if we directly want to panic, @Javagedes ?

components/patina_acpi/src/integration_test.rs
```diff
@@ -18,57 +18,86 @@ use patina::{
...
+        AcpiTableHeader { signature: signature::MADT, length: (signature::MADT_SIZE - 2) as u32, ..Default::default() };
+    // SAFETY: The constructed table is not a valid ACPI table, so this should return an error.
+    assert!(unsafe { table_manager.install_acpi_table(invalid_table) }.is_err(), "Should not install invalid table.");
 
-    // Verify only the FADT is in the iterator.
+    // Verify only valid tables are in the iterator.
     let tables = table_manager.iter_tables();
-    assert_eq!(tables.len(), 1);
-    assert_eq!(tables[0].signature(), signature::FADT);
+    assert!(tables.len() == original_length + 2, "Should have two more tables than original.");
```
Javagedes: please use the `u_assert!` and `u_assert_eq!` macros provided by the patina crate; they are a drop-in replacement. This is so that failing tests do not panic and halt the run but are instead tracked and reported as test failures.

Javagedes: It's slightly outside the scope of the PR since it was already using `assert`, but `patina_test`s should be using `u_assert`, and I'm sorry I missed that in the original PR :)

Javagedes: whoops closing this as os-d mentioend the same thing.

components/patina_acpi/src/integration_test.rs
Javagedes: yep! we should use `u_assert` and `u_assert_eq!` here. They are drop in replacements from the patina crate (they sit right at the top of the crate).

We use `u_assert` for two reasons - (1) it won't panic and block runtime with a dead loop and (2) It allows us to record the errors correctly as @os-d mentioned.

components/patina_acpi/src/integration_test.rs
```diff
@@ -18,57 +18,86 @@ use patina::{
...
-    // Install a FACS table (special case — not iterated over).
-    let facs = AcpiFacs { signature: signature::FACS, length: mem::size_of::<AcpiFacs>() as u32, ..Default::default() };
     // SAFETY: The constructed table is a valid ACPI table.
-    assert!(unsafe { table_manager.install_acpi_table(facs) }.is_ok(), "Should install FACS table.");
+    let key2 = unsafe { table_manager.install_acpi_table(mock_table2) }.expect("Should install table.");
+
+    // Install an invalid ACPI table (too small).
+    let invalid_table =
+        AcpiTableHeader { signature: signature::MADT, length: (signature::MADT_SIZE - 2) as u32, ..Default::default() };
+    // SAFETY: The constructed table is not a valid ACPI table, so this should return an error.
```
makubacki: nit: The safety comment should focus more on the unsafe call's preconditions as opposed to the expected outcome. For example, the table still has a C-compatible layout, with a valid ACPI table header, it's just that the length has been modified to an incorrect value for test purposes. That makes this more of a data validation issue instead of a memory safety violation.

components/patina_acpi/src/integration_test.rs
```diff
@@ -81,45 +110,58 @@ fn acpi_protocol_test(bs: StandardBootServices) -> patina::test::Result {
     // Install a dummy FADT using the ACPI Table Protocol.
```
makubacki: This comment looks stale (old -> FADT, new -> MockLargeTable).

components/patina_acpi/src/integration_test.rs
```diff
@@ -81,45 +110,58 @@ fn acpi_protocol_test(bs: StandardBootServices) -> patina::test::Result {
...
+
+    // SAFETY: `table_buf` is valid and directly constructed from the dummy table.
     let retrieved_table = unsafe { &*table_buf };
-    assert_eq!(retrieved_table.signature(), signature::FADT, "Signature should match installed FADT");
+    assert_eq!(retrieved_table.signature(), 0x12341234, "Signature should match installed table.");
     assert_eq!(get_supported_table_versions, ACPI_VERSIONS_GTE_2, "Should support ACPI version 2.0+");
     assert_eq!(get_table_key, table_key_buf, "Table key should match installed key");
 
     // We should be able to access the normal FADT fields.
     // SAFETY: We know that the table_buf points to an AcpiFadt (constructed above).
```
makubacki: A couple more stale comments.

components/patina_acpi/src/integration_test.rs
```diff
@@ -81,45 +110,58 @@ fn acpi_protocol_test(bs: StandardBootServices) -> patina::test::Result {
...
     #[allow(invalid_reference_casting)]
-    let retrieved_fadt = unsafe { &*(table_buf as *const AcpiFadt) };
+    let large_table = unsafe { &*(table_buf as *const MockLargeTable) };
     // We haven't installed a FACS, so this should be zero, but still accessible.
```
makubacki: FACS seems irrelevant now.

components/patina_acpi/src/integration_test.rs
```diff
@@ -81,45 +110,58 @@ fn acpi_protocol_test(bs: StandardBootServices) -> patina::test::Result {
...
+    // SAFETY: `table_buf` is valid and directly constructed from the dummy table.
     let retrieved_table = unsafe { &*table_buf };
-    assert_eq!(retrieved_table.signature(), signature::FADT, "Signature should match installed FADT");
+    assert_eq!(retrieved_table.signature(), 0x12341234, "Signature should match installed table.");
     assert_eq!(get_supported_table_versions, ACPI_VERSIONS_GTE_2, "Should support ACPI version 2.0+");
     assert_eq!(get_table_key, table_key_buf, "Table key should match installed key");
 
     // We should be able to access the normal FADT fields.
     // SAFETY: We know that the table_buf points to an AcpiFadt (constructed above).
     #[allow(invalid_reference_casting)]
```
makubacki: Is `#[allow(invalid_reference_casting)]` really needed versus an alternate approach?

components/patina_acpi/src/integration_test.rs
makubacki:
```suggestion
    // SAFETY: invalid_table has a valid layout, but an invalid length value, so this should return an error.
```

components/patina_acpi/src/integration_test.rs
```diff
@@ -78,52 +108,63 @@ fn acpi_protocol_test(bs: StandardBootServices) -> patina::test::Result {
...
+        );
+        // We should be able to find our installed table.
+        // SAFETY: `table_buf` is valid and directly constructed from the dummy table.
+        if unsafe { (*table_buf).signature } == 0x12341234 {
+            break;
+        } else if get_result != efi::Status::SUCCESS {
+            // If fails, either hit an error on a previous table or reached the end of the list without finding the installed table.
+            // Both are error cases.
+            u_assert!(false, "Get table should succeed for installed table");
+        }
```
makubacki, line 146: Sorry, something I noticed on the second pass.  It looks like this should check the status before dereferencing the point to check the signature, right?

PR #1398, opened 2026-03-13
Extended x64 exception handler to parse supervisor bit

kuqin12:
## Description

In the programmers' manual the exception data of page fault also defined the level of access that tripped the exception. This change adds the print message when page fault is hit based on the bit status.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This was tested in a kernel-user mode isolated environment and printed the exception as expected.

## Integration Instructions

N/A

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -266,6 +266,18 @@ clear = true
 command = "cargo"
 args = ["fmt", "--all"]
 
+[tasks.fmt-check]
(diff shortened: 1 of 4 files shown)


kuqin12:
The spec snapshot is documented here:

image

PR #1404, opened 2026-03-14
Check and test without default features [Rebase & FF]

makubacki:
## Description

Two changes. One to fix a pre-existing issue in `patina_adv_logger` tests. One to check and test without default features.

I'll take the Makefile.toml changes to patina-devops once they're accepted here.

---

**patina_adv_logger: gate writer tests on the "reader" feature**

writer.rs tests import `crate::reader::AdvancedLogReader`, but
the reader module is conditionally compiled behind
`#[cfg(any(doc, feature = "reader"))]`.

This works when default features are enabled since the "component"
default feature transitively enables "reader".

This change gates `tests` on `#[cfg(all(test, feature = "reader"))]`
so it compiles cleanly when default features are disabled.

---

**Makefile.toml: Check and test without default features**

Without `--no-default-features` testing, feature-gate regressions go
undetected. Code that accidentally uses a gated module or dependency
outside its `cfg` fence compiles fine under default features but
breaks for consumers that disable defaults.

Since workspace crates like `patina_ffs_extractors` gate optional
compression backends (brotli, crc32, lzma) behind features, and
`patina/patina_internal_collections` gates alloc-dependent code
behind an "alloc" default, verifying the `--no-default-features` build
is useful to catch accidental dependencies on gated code.

Add a higher level `check-no-default-features` task with two parallel
subtasks:

  - `check-no-default-features-code`:
    - `cargo check --no-default-features`
  - `check-no-default-features-tests`:
    - `cargo test --no-run --no-default-features`

Run the task in `cargo make all`.

---

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- `cargo make all`
- `cargo make check`
- `cargo make check-no-default-features`

## Integration Instructions

- N/A

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -125,6 +125,23 @@ private = true
 command = "cargo"
 args = ["test", "--no-run", "--all-targets", "--all-features", "@@split(CARGO_MAKE_TASK_ARGS, )"]
 
+[tasks.check-no-default-features-code]
+description = "Checks rust code compiles without default features."
+private = true
+command = "cargo"
+args = ["check", "--no-default-features", "@@split(CARGO_MAKE_TASK_ARGS, )"]
+
+[tasks.check-no-default-features-tests]
+description = "Checks rust test code compiles without default features."
+private = true
+command = "cargo"
+args = ["test", "--no-run", "--no-default-features", "@@split(CARGO_MAKE_TASK_ARGS, )"]
+
+[tasks.check-no-default-features]
+description = "Checks rust code and tests compile without default features to catch feature-gate regressions."
+clear = true
+run_task = [{ name = ["check-no-default-features-code", "check-no-default-features-tests"], parallel = true }]
+
 [tasks.check]
 description = "Checks rust code for errors. Example `cargo make check`"
 clear = true
@@ -311,6 +328,7 @@ dependencies = [
     "deny",
     "cspell",
     "clippy",
+    "check-no-default-features",
     "build",
     "build-x64",
     "build-aarch64",


Javagedes:
Is there an urgency to get this in here first, rather than patina-devops?

Makefile.toml
```diff
@@ -125,10 +125,27 @@ private = true
...
+
+[tasks.check-no-default-features]
+description = "Checks rust code and tests compile without default features to catch feature-gate regressions."
+clear = true
+run_task = [{ name = ["check-no-default-features-code", "check-no-default-features-tests"], parallel = true }]
+
 [tasks.check]
 description = "Checks rust code for errors. Example `cargo make check`"
 clear = true
-run_task = [{ name = ["check_code", "check_tests"], parallel = true }]
```
Javagedes, line 131: I am concerned with doing this during cargo check. It's already kind of slow.

I'm all for a CI check, but a bit cautious for during-development.

makubacki:
> Is there an urgency to get this in here first, rather than patina-devops?

I'm fine to take there. I personally like to see the patina repo pass with the change and have a better explanation than what comes with file sync. Unless there is a strong concern with that, I'd rather Rebase & FF this and let the patina-devops sync resolve to nothing.

Makefile.toml
makubacki: I can drop it from check and keep it in all.

Javagedes:
> > Is there an urgency to get this in here first, rather than patina-devops?
> 
> 
> 
> I'm fine to take there. I personally like to see the patina repo pass with the change and have a better explanation than what comes with file sync. Unless there is a strong concern with that, I'd rather Rebase & FF this and let the patina-devops sync resolve to nothing.

I'm fine with the sync resolving to nothing

Makefile.toml
makubacki: Done

kuqin12 (approved):

PR #1408, opened 2026-03-16
Ingest patina-paging 11.0.2 and use new open_active interface

cfernald:
## Description

Patina-paging 11.0.2 introduces support for reading/editing existing 5-level page tables on AArch64. This is required for EDK2 20511 based system that have FEAT_LPA2 support. To use this change, this commit switches to use the new `open_active` interface for both AArch64 and x64, which also simplifies the consumer by handling level detection in the paging library.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- [x] Boot test w/ debugger on Q35
- [x] Boot test w/ debugger on SBSA

## Integration Instructions

N/A

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,7 +42,7 @@ patina_lzma_rs = { version = "0.3.1", default-features = false }
 patina_macro = { version = "20.1.1", path = "sdk/patina_macro" }
 patina_mm = { version = "20.1.1", path = "components/patina_mm" }
 patina_mtrr = { version = "^1.1.4" }
-patina_paging = { version = "11" }
+patina_paging = { version = "11.0.2" }
 patina_performance = { version = "20.1.1", path = "components/patina_performance" }
 patina_smbios = { version = "20.1.1", path = "components/patina_smbios" }
 patina_stacktrace = { version = "20.1.1", path = "core/patina_stacktrace" }
--- a/core/patina_debugger/src/arch/aarch64.rs
+++ b/core/patina_debugger/src/arch/aarch64.rs
@@ -216,19 +216,11 @@ impl DebuggerArch for Aarch64Arch {
     }
 
     fn get_page_table() -> Result<Self::PageTable, ()> {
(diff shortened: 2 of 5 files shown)


Cargo.toml
```diff
@@ -42,7 +42,7 @@ patina_lzma_rs = { version = "0.3.1", default-features = false }
 patina_macro = { version = "20.1.1", path = "sdk/patina_macro" }
 patina_mm = { version = "20.1.1", path = "components/patina_mm" }
 patina_mtrr = { version = "^1.1.4" }
-patina_paging = { version = "11" }
+patina_paging = { version = "^11.0.2" }
```
os-d:
```suggestion
patina_paging = { version = "11.0.2" }
```

nit: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#caret-requirements

core/patina_debugger/src/arch/aarch64.rs
```diff
@@ -216,19 +216,11 @@ impl DebuggerArch for Aarch64Arch {
     }
 
     fn get_page_table() -> Result<Self::PageTable, ()> {
-        // TODO: Check for EL1?
-        let ttbr0_el2 = read_sysreg!(ttbr0_el2);
-
-        // SAFETY: We are creating from the existing page table root, so the
-        // page tables should be a valid structure. Using PageAllocatorStub
-        // ensures that no allocations are attempted.
+        // SAFETY: We are operating in an exception context with interrupts disabled. No other entity should be altering
```
Javagedes: The word "Should" seems a bit dangerous in this context. Is it that they cannot or should not? If they should not be, is there a way we can enforce they are not?

cfernald: This is just poor wording on my part. No other code is running. Will clarify

PR #1409, opened 2026-03-16
[Rebase&FF] pecoff: add `get_section` functionality

Javagedes:
## Description

This commit pulls the functionality for getting a sub-slice that represents a particular PE/COFF section from an unloaded image out of `load_resource_section` and makes it a re-usable function. It also removes an unnecessary allocation when comparing section names.

**NOTE:** The git diff does not do this change justice due to each line's "depth" being adjusted. There is no change in logic, except for the second commit, which removes the allocation.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

CI continues to work. Boot on Q35.

## Integration Instructions

N/A

### DIFF
--- a/patina_dxe_core/src/pecoff.rs
+++ b/patina_dxe_core/src/pecoff.rs
@@ -505,8 +505,10 @@ pub fn load_resource_section(pe_info: &UefiPeInfo, image: &[u8]) -> error::Resul
 /// Returns [Goblin(Malformed)](error::Error::Goblin) error if the section size is invalid
 /// Returns [Goblin(BufferTooShort)](error::Error::Goblin) error if the section is not fully contained in the image
 pub fn get_section<'a>(target: &str, pe_info: &UefiPeInfo, image: &'a [u8]) -> error::Result<Option<&'a [u8]>> {
+    let target_bytes = target.as_bytes();
     for section in &pe_info.sections {
-        if String::from_utf8_lossy(&section.name).trim_end_matches('\0') == target {
+        let section_name_bytes = section.name.split(|&b| b == 0).next().unwrap_or_default();
+        if section_name_bytes == target_bytes {
             let mut size = section.virtual_size;
             if size == 0 || size > section.size_of_raw_data {
                 size = section.size_of_raw_data;


patina_dxe_core/src/pecoff.rs
```diff
@@ -406,119 +406,120 @@ pub fn flatten_runtime_relocation_data(relocation_data: &[RelocationBlock]) -> &
...
             }
 
-            let mut directory_entry: DirectoryEntry = resource_section.pread(core::mem::size_of::<Directory>())?;
+            let resource_directory_string =
+                resource_section.pread::<DirectoryString>(directory_entry.name_offset() as usize)?;
 
-            for _ in 0..directory.number_of_named_entries {
-                if directory_entry.name_is_string() {
-                    if directory_entry.name_offset() >= size {
+            let name_start_offset = (directory_entry.name_offset() + 1) as usize;
```
makubacki: I know this was pre-existing, but it seems a little strange for the name offset to start a byte (`+ 1`) in when a UTF-16 character is 2 bytes.

In the comment below: `// L"HII" = [0x0, 0x48, 0x0, 0x49, 0x0, 0x49]`, the low-byte would typically have the ASCII character for UTF-16 little endian:

```
// L"HII" = [0x48, 0x0, 0x49, 0x0, 0x49, 0x0]
```

patina_dxe_core/src/pecoff.rs
```diff
@@ -406,119 +406,120 @@ pub fn flatten_runtime_relocation_data(relocation_data: &[RelocationBlock]) -> &
...
+                    return Ok(Some((resource_data_entry.offset_to_data as usize, resource_data_entry.size as usize)));
+                }
+            }
+        }
+    }
+
+    Ok(None)
+}
+
+/// Returns the a slice from the image, containing the data in the requested section, if it exists.
```
makubacki:
```suggestion
/// Returns a slice from the image, containing the data in the requested section, if it exists.
```

patina_dxe_core/src/pecoff.rs
```diff
@@ -874,4 +875,15 @@ mod tests {
             Err(e) => panic!("Expected BufferTooShort error, got {e:?}"),
         }
     }
+
+    #[test]
+    fn test_load_sbom_section() {
```
makubacki: It looks like a few other test cases like testing for a section that is not found could be added.

Javagedes: I added a test for searching for a section that does not exist, as it was simple. However, the other scenarios revolving around poorly formatted data is already covered by other tests (This function has 100% line / branch coverage per codecov).

patina_dxe_core/src/pecoff.rs
makubacki: If you want, I can also take this up separately since it is unrelated to your PR.

Javagedes: @joschock may have a better answer to this, but I assume that it has to do with the name_offset() being tied to the ID, and thus needing to do an offset by 1, similar to 0 or 1 based indexing. But I'm not familiar with this enough to answer confidently when it is working as-is.

joschock: I have no memory of this place :) - I do not know what the +1 is. Maybe it is trying to skip the "length" field? But the length field is 2 bytes...

Javagedes: I'm going to resolve this. @makubacki would you like me to create an issue to have this investigated?

makubacki: Sure. I'm also interested to look in more detail if you want to assign me.

makubacki: @Javagedes, I fixed this along with a couple of other issues I noticed in https://github.com/OpenDevicePartnership/patina/pull/1431.

Let's get that in first to address those in the existing code. Then, you can rebase and proceed with the refactor including the updated code.

Issue #1410, opened 2026-03-16
[Bug]: Remove alloc reliance from stacktrace


kuqin12:
### Is there an existing issue for this?

- [x] I have searched existing issues

### Current Behavior

The stacktrace crate is including alloc in the source code directly, which imposes an implicit dependency on global allocator.

### Expected Behavior

Stacktrace provides the basic functionality of stack unwinding during exception handling, which should not rely on allocator.

### Steps To Reproduce

Search for `extern crate alloc` inside stacktrace crate.

### Build Environment

```markdown
- OS(s): Windows
- Tool Chain(s): prod-1.92
- Targets Impacted: AArch64
```

### Version Information

```text
Top of main
```

### Urgency

Medium

### Are you going to fix this?

I will fix it

### Do you need maintainer feedback?

No maintainer feedback needed

PR #1411, opened 2026-03-16
[painta_stacktrace] remove alloc usage

kuqin12:
## Description

Current patina_stacktrace is pulling in alloc from source code. However, as basic functionality as the crate provides, the dependency is improperly included.

This change removes the dependency.

Resolves https://github.com/OpenDevicePartnership/patina/issues/1410

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

The change is not a functional change and was tested with local build.

## Integration Instructions

N/A

### DIFF
--- a/core/patina_stacktrace/src/aarch64/runtime_function.rs
+++ b/core/patina_stacktrace/src/aarch64/runtime_function.rs
@@ -262,6 +262,7 @@ impl<'a> RuntimeFunction<'a> {
 #[coverage(off)]
 mod tests {
     use super::*;
+    use std::vec::Vec;
 
     const IMAGE_SIZE: usize = 0x2000;
     const PE_POINTER_OFFSET: usize = 0x3C;
@@ -272,7 +273,7 @@ mod tests {
     const EXCEPTION_TABLE_POINTER_OFFSET: usize = 0xA0;
 
     // Creates a fake PE image containing the supplied .pdata entries and additional sections.
-    fn build_pe_bytes(entries: &[(u32, u32)], extra_sections: &[(u32, &[u8])]) -> std::vec::Vec<u8> {
+    fn build_pe_bytes(entries: &[(u32, u32)], extra_sections: &[(u32, &[u8])]) -> Vec<u8> {
         let mut bytes = vec![0u8; IMAGE_SIZE];
 
         bytes[0..2].copy_from_slice(&0x5A4Du16.to_le_bytes());


core/patina_stacktrace/src/aarch64/runtime_function.rs
```diff
@@ -1,4 +1,3 @@
-extern crate alloc;
```
Javagedes, line 1: I would just update L266 in this file from `use alloc::vec::Vec` to

```rust
extern crate std;
use std::vec::Vector;
```

core/patina_stacktrace/src/lib.rs
```diff
@@ -169,6 +169,7 @@
 #![cfg_attr(all(not(feature = "std"), not(test)), no_std)]
 #![feature(coverage_attribute)]
 
+#[cfg(test)]
```
Javagedes: I'd remove this extern crate alloc instead of putting it behind a #[cfg(test)] since no code actually uses it. Just do what I mentioned above and you should be good to go.

kuqin12: Ah, yes. will update!

core/patina_stacktrace/src/aarch64/runtime_function.rs
kuqin12: So.. I ended up inlining the std into the one instance that is used. Please let me know if you prefer your original suggestion.

Javagedes: I think that should work since we don't mark `no_std` when compiling for tests.

core/patina_stacktrace/src/aarch64/runtime_function.rs
```diff
@@ -274,7 +272,7 @@ mod tests {
     const EXCEPTION_TABLE_POINTER_OFFSET: usize = 0xA0;
 
     // Creates a fake PE image containing the supplied .pdata entries and additional sections.
-    fn build_pe_bytes(entries: &[(u32, u32)], extra_sections: &[(u32, &[u8])]) -> Vec<u8> {
+    fn build_pe_bytes(entries: &[(u32, u32)], extra_sections: &[(u32, &[u8])]) -> std::vec::Vec<u8> {
```
makubacki: Our normal convention would be to keep this line of code unchanged but import `Vec<u8>` from `std::vec`. I slightly prefer that because it cleans up this type and makes it clear within the overall `test` module where `Vec` is and should be coming from. But this is also fine, so feel free to ignore.

kuqin12: Okay, converging to that style as @Javagedes also suggested so.

kuqin12: It did not occur to me that an "extern" statement is needed when I test the build locally. We will see if the build break on the pipeline.

PR #1385, opened 2026-03-11
Initial implementation of supervisor

kuqin12:
## Description

This is the initial implementation of MM supervisor and user core in Rust.

- [x] Impacts functionality?
- [x] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This was tested on QEMU Q35 platform and booted to OS desktop as well as passed supervisor test app.

## Integration Instructions

The integration guide is listed in: https://github.com/kuqin12/mu_feature_mm_supv/blob/personal/kuqin/supv_init/SeaPkg/Docs/PlatformIntegration/PlatformIntegrationSteps.md#integraion-guide-for-rust-based-supervisor. Because it provides the implementation of MM supervisor init module.

PR #1413, opened 2026-03-17
[patina_mm] refactor current folder structure to allow external consumption

kuqin12:
## Description

Current patina_mm carries some definitions that can be shared by external users (mm communication header, buffer status, etc).

This change moves those files to a sub-module that does not require alloc.

It also separates the supervisor related definitions into its own file.

Resolves https://github.com/OpenDevicePartnership/patina/issues/1437.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This was tested with local builds and booted to UEFI shell.

## Integration Instructions

N/A

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,7 +3,7 @@ resolver = "3"
 members = ["components/*", "core/*", "sdk/*", "patina_dxe_core"]
 
 [workspace.package]
-version = "20.1.3"
+version = "21.0.0"
 license = "Apache-2.0"
 edition = "2024"
 rust-version = "1.89"
@@ -31,22 +31,22 @@ log = { version = "0.4", default-features = false }
 memoffset = {version = "0.9.1" }
 mu_rust_helpers = { version = "3.0.2" }
 num-traits = { version = "0.2", default-features = false }
-patina = { version = "20.1.3", path = "sdk/patina" }
-patina_debugger = { version = "20.1.3", path = "core/patina_debugger" }
-patina_ffs = { version = "20.1.3", path = "sdk/patina_ffs" }
-patina_ffs_extractors = { version = "20.1.3", path = "sdk/patina_ffs_extractors" }
-patina_internal_collections = { version = "20.1.3", path = "core/patina_internal_collections", default-features = false }
-patina_internal_cpu = { version = "20.1.3", path = "core/patina_internal_cpu" }
-patina_internal_depex = { version = "20.1.3", path = "core/patina_internal_depex" }
+patina = { version = "21.0.0", path = "sdk/patina" }
+patina_debugger = { version = "21.0.0", path = "core/patina_debugger" }
+patina_ffs = { version = "21.0.0", path = "sdk/patina_ffs" }
+patina_ffs_extractors = { version = "21.0.0", path = "sdk/patina_ffs_extractors" }
+patina_internal_collections = { version = "21.0.0", path = "core/patina_internal_collections", default-features = false }
+patina_internal_cpu = { version = "21.0.0", path = "core/patina_internal_cpu" }
+patina_internal_depex = { version = "21.0.0", path = "core/patina_internal_depex" }
 patina_lzma_rs = { version = "0.3.1", default-features = false }
-patina_macro = { version = "20.1.3", path = "sdk/patina_macro" }
-patina_mm = { version = "20.1.3", path = "components/patina_mm" }
+patina_macro = { version = "21.0.0", path = "sdk/patina_macro" }
+patina_mm = { version = "21.0.0", path = "components/patina_mm" }
 patina_mtrr = { version = "^1.1.4" }
 patina_paging = { version = "11.0.2" }
-patina_performance = { version = "20.1.3", path = "components/patina_performance" }
-patina_smbios = { version = "20.1.3", path = "components/patina_smbios" }
-patina_stacktrace = { version = "20.1.3", path = "core/patina_stacktrace" }
-patina_test = { version = "20.1.3", path = "components/patina_test" }
+patina_performance = { version = "21.0.0", path = "components/patina_performance" }
+patina_smbios = { version = "21.0.0", path = "components/patina_smbios" }
+patina_stacktrace = { version = "21.0.0", path = "core/patina_stacktrace" }
+patina_test = { version = "21.0.0", path = "components/patina_test" }
 proc-macro2 = { version = "1" }
 quote = { version = "1" }
 r-efi = { version = "5.0.0", default-features = false }
--- a/components/patina_adv_logger/README.md
+++ b/components/patina_adv_logger/README.md
@@ -43,14 +43,14 @@ below.
 ```rust
 use patina_dxe_core::*;
 use patina::{log::Format, serial::uart::UartNull};
-use patina_adv_logger::{component::AdvancedLoggerComponent, logger::AdvancedLogger};
+use patina_adv_logger::{component::AdvancedLoggerComponent, logger::{AdvancedLogger, TargetFilter}};
 
 use log::LevelFilter;
 use core::ffi::c_void;
 
 static LOGGER: AdvancedLogger<UartNull> = AdvancedLogger::new(
    Format::Standard, // How logs are formatted
-   &[("allocations", LevelFilter::Off)], // set custom log levels per module
+   &[TargetFilter { target: "allocations", log_level: LevelFilter::Off, hw_filter_override: None }], // set custom log levels per module
    log::LevelFilter::Info, // Default log level
    UartNull { }, // Serial writer instance
 );
--- a/components/patina_adv_logger/src/component.rs
+++ b/components/patina_adv_logger/src/component.rs
@@ -69,7 +69,7 @@ where
         // SAFETY: We must trust the C code was a responsible steward of this buffer.
         let internal = unsafe { &*(this as *const AdvancedLoggerProtocolInternal<S>) };
 
-        internal.adv_logger.log_write(error_level, data);
+        internal.adv_logger.log_write(error_level, None, data);
         efi::Status::SUCCESS
     }
 
--- a/components/patina_adv_logger/src/logger.rs
+++ b/components/patina_adv_logger/src/logger.rs
@@ -29,13 +29,27 @@ use spin::RwLock;
 #[used]
 static mut DBG_ADV_LOG_BUFFER: u64 = 0;
 
+/// Per-target filter that binds a target name prefix with its log level and optional hardware print level override.
+pub struct TargetFilter<'a> {
+    /// Target name prefix to match.
+    pub target: &'a str,
+    /// Maximum log level for this target. Messages above this are dropped entirely.
+    pub log_level: log::LevelFilter,
+    /// Optional override for the hardware print level for this target. Messages above this level will not be printed
+    /// to the hardware port, but may still be logged to the memory log based on log_level and the overall max_level.
+    /// - `None` = use global `hw_print_level` from memory log header.
+    /// - `Some(level_filter)` Use the provided level filter to control hardware printing for this target, instead
+    ///   of the global `hw_print_level`.
+    pub hw_filter_override: Option<log::LevelFilter>,
+}
+
 /// The logger for memory/hardware port logging.
 pub struct AdvancedLogger<'a, S>
 where
     S: SerialIO + Send,
 {
     hardware_port: S,
-    target_filters: &'a [(&'a str, log::LevelFilter)],
+    target_filters: &'a [TargetFilter<'a>],
     max_level: log::LevelFilter,
     format: Format,
     memory_log: RwLock<Option<AdvancedLogWriter>>,
@@ -51,13 +65,13 @@ where
     /// ## Arguments
     ///
     /// * `format` - The format to use for logging.
-    /// * `target_filters` - A list of target filters to apply to the logger.
+    /// * `target_filters` - Per-target filters that control log level and optionally the hardware print filter.
     /// * `max_level` - The maximum log level to log.
     /// * `hardware_port` - The hardware port to write logs to.
     ///
     pub const fn new(
         format: Format,
-        target_filters: &'a [(&'a str, log::LevelFilter)],
+        target_filters: &'a [TargetFilter<'a>],
         max_level: log::LevelFilter,
         hardware_port: S,
     ) -> Self {
@@ -116,12 +130,18 @@ where
     }
 
     /// Writes a log entry to the hardware port and memory log if available.
-    pub(crate) fn log_write(&self, error_level: u32, data: &[u8]) {
+    ///
+    /// `hw_print_mask_override` optionally overrides the global hw_print_level
+    /// from the memory log header, enabling per-target hardware print filtering.
+    pub(crate) fn log_write(&self, error_level: u32, hw_print_mask_override: Option<u32>, data: &[u8]) {
         self.refresh_log_info_address();
         let mut hw_write = true;
         let log_guard = self.memory_log.read();
         if let Some(memory_log) = log_guard.as_ref() {
-            hw_write = memory_log.hardware_write_enabled(error_level);
+            hw_write = match hw_print_mask_override {
+                Some(mask) => memory_log.hardware_write_enabled_with_mask(error_level, mask),
+                None => memory_log.hardware_write_enabled(error_level),
+            };
             let timestamp = self.timer.map_or(0, |timer| timer.cpu_count());
             let _ = memory_log.add_log_entry(LogEntry {
                 phase: memory_log::ADVANCED_LOGGER_PHASE_DXE,
@@ -198,6 +218,11 @@ where
             self.set_log_info_address(new_address);
         }
     }
+
+    /// Returns the matching target filter for the given target name, if any.
+    fn target_filter(&self, target: &str) -> Option<&TargetFilter<'a>> {
+        self.target_filters.iter().find(|f| target.starts_with(f.target))
+    }
 }
 
 impl<S> log::Log for AdvancedLogger<'_, S>
@@ -205,19 +230,18 @@ where
     S: SerialIO + Send,
 {
     fn enabled(&self, metadata: &log::Metadata) -> bool {
-        metadata.level().to_level_filter()
-            <= *self
-                .target_filters
-                .iter()
-                .find(|(name, _)| metadata.target().starts_with(name))
-                .map(|(_, level)| level)
-                .unwrap_or(&self.max_level)
+        let max_level = self.target_filter(metadata.target()).map(|f| f.log_level).unwrap_or(self.max_level);
+        metadata.level().to_level_filter() <= max_level
     }
 
     fn log(&self, record: &log::Record) {
-        if self.enabled(record.metadata()) {
+        let filter = self.target_filter(record.target());
+        let max_level = filter.map(|f| f.log_level).unwrap_or(self.max_level);
+
+        if record.metadata().level().to_level_filter() <= max_level {
             let level = log_level_to_debug_level(record.metadata().level());
-            let mut writer = BufferedWriter::new(level, self);
+            let hw_print_mask_override = filter.and_then(|f| f.hw_filter_override).map(log_level_filter_to_debug_mask);
+            let mut writer = BufferedWriter::new(level, hw_print_mask_override, self);
             self.format.write(&mut writer, record);
             writer.flush();
         }
@@ -239,6 +263,24 @@ const fn log_level_to_debug_level(level: Level) -> u32 {
     }
 }
 
+/// Converts a `log::LevelFilter` to a hardware print mask.
+const fn log_level_filter_to_debug_mask(level_filter: log::LevelFilter) -> u32 {
+    match level_filter {
+        log::LevelFilter::Error => memory_log::DEBUG_LEVEL_ERROR,
+        log::LevelFilter::Warn => memory_log::DEBUG_LEVEL_ERROR | memory_log::DEBUG_LEVEL_WARNING,
+        log::LevelFilter::Info => {
+            memory_log::DEBUG_LEVEL_ERROR | memory_log::DEBUG_LEVEL_WARNING | memory_log::DEBUG_LEVEL_INFO
+        }
+        log::LevelFilter::Debug | log::LevelFilter::Trace => {
+            memory_log::DEBUG_LEVEL_ERROR
+                | memory_log::DEBUG_LEVEL_WARNING
+                | memory_log::DEBUG_LEVEL_INFO
+                | memory_log::DEBUG_LEVEL_VERBOSE
+        }
+        log::LevelFilter::Off => 0,
+    }
+}
+
 /// Size of the buffer for the buffered writer.
 const WRITER_BUFFER_SIZE: usize = 128;
 
@@ -248,6 +290,7 @@ where
     S: SerialIO + Send,
 {
     level: u32,
+    hw_print_mask_override: Option<u32>,
     writer: &'a AdvancedLogger<'a, S>,
     buffer: [u8; WRITER_BUFFER_SIZE],
     buffer_size: usize,
@@ -257,9 +300,9 @@ impl<'a, S> BufferedWriter<'a, S>
 where
     S: SerialIO + Send,
 {
-    /// Creates a new BufferedWriter with the specified log level and writer.
-    const fn new(level: u32, writer: &'a AdvancedLogger<'a, S>) -> Self {
-        Self { level, writer, buffer: [0; WRITER_BUFFER_SIZE], buffer_size: 0 }
+    /// Creates a new BufferedWriter with the specified log level, optional hardware print mask override, and writer.
+    const fn new(level: u32, hw_print_mask_override: Option<u32>, writer: &'a AdvancedLogger<'a, S>) -> Self {
+        Self { level, hw_print_mask_override, writer, buffer: [0; WRITER_BUFFER_SIZE], buffer_size: 0 }
     }
 
     /// Flushes the current buffer to the underlying writer.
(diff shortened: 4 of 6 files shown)


components/patina_mm/src/protocol/mm_supervisor_request.rs
```diff
@@ -0,0 +1,226 @@
...
+//!
+//! ## License
+//!
+//! Copyright (C) Microsoft Corporation.
+//!
+//! SPDX-License-Identifier: Apache-2.0
+
+use zerocopy::FromBytes;
+use zerocopy_derive::{FromBytes, Immutable, IntoBytes, KnownLayout};
+use r_efi::efi::Guid;
```
kuqin12: Old style

components/patina_mm/src/protocol/mm_supervisor_request.rs
```diff
@@ -0,0 +1,226 @@
...
+    /// Validates the header signature and revision.
+    pub fn is_valid(&self) -> bool {
+        self.signature == SIGNATURE && self.revision <= REVISION
+    }
+
+    /// Reads a header from a byte slice.
+    ///
+    /// Returns `None` if the slice is too small or misaligned.
+    pub fn from_bytes(bytes: &[u8]) -> Option<Self> {
+        if bytes.len() < Self::SIZE {
```
kuqin12: See https://github.com/OpenDevicePartnership/patina/pull/1385#discussion_r2919452895

components/patina_mm/src/protocol/mm_supervisor_request.rs
```diff
@@ -0,0 +1,226 @@
...
+            0x0004 => Ok(Self::CommUpdate),
+            other => Err(other),
+        }
+    }
+}
+
+/// Deprecated module — use [`RequestType`] enum variants instead.
+///
+/// Kept temporarily for backward compatibility.
+pub mod requests {
```
kuqin12: See this: https://github.com/OpenDevicePartnership/patina/pull/1385#discussion_r2919471740

components/patina_mm/src/protocol/mm_supervisor_request.rs
```diff
@@ -0,0 +1,226 @@
...
+pub struct MmSupervisorUnblockMemoryParams {
+    /// Memory descriptor identifying the region to unblock.
+    pub memory_descriptor: efi::MemoryDescriptor,
+    /// GUID identifying the requesting driver/module.
+    pub identifier_guid: Guid,
+}
+
+impl MmSupervisorUnblockMemoryParams {
+    /// Size of this structure in bytes (56).
+    pub const SIZE: usize = core::mem::size_of::<Self>();
```
kuqin12: See this: https://github.com/OpenDevicePartnership/patina/pull/1385#discussion_r2919483560

components/patina_mm/src/protocol/mm_supervisor_request.rs
kuqin12: See this: https://github.com/OpenDevicePartnership/patina/pull/1385#discussion_r2919491099

components/patina_mm/src/protocol/mm_supervisor_request.rs
```diff
@@ -0,0 +1,226 @@
...
+        Self::read_from_bytes(&bytes[..Self::SIZE]).ok()
+    }
+}
+
+// ============================================================================
+// Protocol Constants
+// ============================================================================
+
+/// The expected signature value ('MSUP' as little-endian u32).
+pub const SIGNATURE: u32 = 0x5055534D;
```
kuqin12: See this: https://github.com/OpenDevicePartnership/patina/pull/1385#discussion_r2919514134

components/patina_mm/src/comm_buffer_hob.rs
```diff
@@ -0,0 +1,187 @@
...
+use zerocopy_derive::{FromBytes, Immutable, KnownLayout};
+
+// =============================================================================
+// GUIDs
+// =============================================================================
+
+/// GUID for the MM communication buffer HOB (`gMmCommBufferHobGuid`).
+///
+/// `{ 0x6c2a2520, 0x0131, 0x4aee, { 0xa7, 0x50, 0xcc, 0x38, 0x4a, 0xac, 0xe8, 0xc6 } }`
+pub const MM_COMM_BUFFER_HOB_GUID: efi::Guid = efi::Guid::from_fields(
```
kuqin12: Old style

components/patina_mm/src/comm_buffer_hob.rs
```diff
@@ -0,0 +1,187 @@
...
+//! SPDX-License-Identifier: Apache-2.0
+//!
+
+use r_efi::efi;
+use patina::Guid;
+use zerocopy_derive::{FromBytes, Immutable, KnownLayout};
+
+// =============================================================================
+// GUIDs
+// =============================================================================
```
kuqin12: See this: https://github.com/OpenDevicePartnership/patina/pull/1385#discussion_r2919544291

components/patina_mm/tests/patina_mm_integration/mm_supervisor/communication_tests.rs
```diff
@@ -315,14 +315,14 @@ fn test_mm_supervisor_unblock_mem_request() {
...
+        signature: mm_supv::SIGNATURE,
         revision: mm_supv::REVISION,
-        request: mm_supv::requests::UNBLOCK_MEM, // This uses the constant!
+        request: mm_supv::RequestType::UnblockMem as u32,
         reserved: 0,
         result: 0,
     };
 
-    let request_bytes = unblock_mem_request.to_bytes();
+    let request_bytes = unblock_mem_request.as_bytes().to_vec();
```
kuqin12: ??

sdk/patina/src/mm_services.rs
```diff
@@ -0,0 +1,612 @@
...
+/// This is the MM equivalent of
+/// [`StandardBootServices`](crate::boot_services::StandardBootServices).
+pub struct StandardMmServices {
+    efi_mm_system_table: Once<*mut EfiMmSystemTable>,
+}
+
+// SAFETY: The raw pointer is only written once (protected by `Once`) and the
+// underlying table is not expected to change after initialisation.
+unsafe impl Sync for StandardMmServices {}
+unsafe impl Send for StandardMmServices {}
```
kuqin12: ??

components/patina_mm/src/comm_buffer_hob.rs
```diff
@@ -0,0 +1,172 @@
...
+//!
+//! ## License
+//!
+//! Copyright (C) Microsoft Corporation.
+//!
+//! SPDX-License-Identifier: Apache-2.0
+//!
+
+use patina::{Guid, BinaryGuid};
+use zerocopy_derive::{FromBytes, Immutable, KnownLayout};
```
kuqin12: Wonder if these are so fundamental that should live at the root of this lib.rs.

kuqin12: Or at least export as the root level.

sdk/patina/src/mm_services.rs
kuqin12, line 1: Should do this in a separate PR.

components/patina_mm/src/protocol/mm_supervisor_request.rs
kuqin12: Fixed.

components/patina_mm/src/protocol/mm_supervisor_request.rs
kuqin12: Fixed 2 instances in this PR.

components/patina_mm/src/protocol/mm_supervisor_request.rs
kuqin12: Moved to enum for both requests and responses.

kuqin12: Actually, one moment while I move the enum to be error only.

kuqin12: Updated to only do conversion between response type into efi_status

components/patina_mm/src/protocol/mm_supervisor_request.rs
kuqin12: Removing the magic number

components/patina_mm/src/comm_buffer_hob.rs
kuqin12: Fixed.

components/patina_mm/src/comm_buffer_hob.rs
kuqin12: Removed

sdk/patina/src/mm_services.rs
kuqin12: Removed for a future PR

components/patina_mm/src/comm_buffer_hob.rs
kuqin12: That will still be a breaking change, if there are any external users. Will not do that for now.

sdk/patina/src/mm_services.rs
kuqin12: Removed.

components/patina_mm/src/protocol/mm_supervisor_request.rs
kuqin12: Updated.

components/patina_mm/tests/patina_mm_integration/mm_supervisor/communication_tests.rs
kuqin12: not sure what is going on with this one...

Javagedes:
@makubacki Curious to get your thoughts here but seeing as this particular crate (`patina_mm`) is a component, it does not make sense to carry public definitions that will be used by others. The intent of component crates is to only really be consumed by platforms when registering components or by other components when they use services provided by the component.

I'm wondering if we should move common definitions for patina_mm related things into the `patina` crate rather than here. With the current implementation this will create a weird dependency where the supervisor depends on this component.

makubacki:
> @makubacki Curious to get your thoughts here...

I haven't looked at the PR changes yet, but `patina_mm` was originally written to provide components and supporting configs and services. That code had to proceed conservatively until the broader looming MM changes were ready. There might be some code in there that makes sense to refactor based on those larger MM changes.

I agree with your reasoning. Components provide functionality, not public definitions. Yes, common definitions should be moved to the sdk/patina crate.

kuqin12:
> I'm wondering if we should move common definitions for patina_mm related things into the `patina` crate rather than here. With the current implementation this will create a weird dependency where the supervisor depends on this component.

> I agree with your reasoning. Components provide functionality, not public definitions. Yes, common definitions should be moved to the sdk/patina crate.

Sounds good. Will update accordingly.

components/patina_performance/Cargo.toml
```diff
@@ -17,7 +17,7 @@ uuid = { workspace = true }
 
 mu_rust_helpers = { workspace = true }
 patina = { workspace = true }
-patina_mm = { workspace = true }
+patina_mm = { workspace = true, features = ["alloc"] }
```
kuqin12: revert

components/patina_mm/Cargo.toml
```diff
@@ -23,14 +23,15 @@ zerocopy-derive = { workspace = true }
...
+patina_dxe_core = { path = "../../patina_dxe_core" }
 
 [target.'cfg(target_arch="x86_64")'.dependencies]
 x86_64 = { workspace = true, features = [
   "instructions",
 ] }
 
 [features]
-doc = []
+doc = ['alloc']
```
kuqin12: no need

components/patina_performance/Cargo.toml
kuqin12: Removed

components/patina_mm/Cargo.toml
kuqin12: removed

kuqin12:
Okay, moved the non-alloc related definitions (header structures, constants, protocols) to patina and updated the consumers to use the new definitions.

kuqin12:
@Javagedes @makubacki would you mind reviewing once more? It is updated to move the definitions to patina. Thanks!

kuqin12:
The validation failed due to `--toolchain GCC5` in unrecognized :)

makubacki:
> The validation failed due to `--toolchain GCC5` in unrecognized :)

For the record, the workflow also needed to not pass `--toolchain` after the CLANGPDB transition in patina-qemu removed the parameter.

sdk/patina/src/management_mode/comm_buffer_hob.rs
```diff
@@ -0,0 +1,171 @@
...
+/// MM Communication Buffer Status
+///
+/// Shared structure between DXE and MM environments to communicate the status
+/// of MM communication operations. This structure is written by DXE before
+/// triggering an MMI and read/written by MM during MMI processing.
+///
+/// This is a structure currently used in some MM Supervisor MM implementations.
+#[derive(Debug, Clone, Copy, FromBytes, Immutable, KnownLayout)]
+#[repr(C)]
+pub struct MmCommBufferStatus {
```
makubacki, line 53: This is a little concerning because it doesn't match:

- [mu_basecore definition](https://github.com/microsoft/mu_basecore/blob/6dcd36fba6e3dc6ecfae53a020ef0f6270b81915/MdeModulePkg/Include/Guid/MmCommBuffer.h#L44-L70)

```rust
typedef struct {
  ///
  /// Whether the data in the fixed MM communication buffer is valid when entering from non-MM to MM.
  ///
  BOOLEAN    IsCommBufferValid;

  /// MU_CHANGE Starts
  /// For padding purpose
  // UINT8      Reserved[7];
  UINT8      Reserved[6];

  ///
  /// The channel used to communicate with MM.
  ///
  BOOLEAN    TalkToSupervisor;
  /// MU_CHANGE Ends

  ///
  /// The return status when returning from MM to non-MM.
  ///
  UINT64     ReturnStatus;

  ///
  /// The size in bytes of the output buffer when returning from MM to non-MM.
  ///
  UINT64     ReturnBufferSize;
} MM_COMM_BUFFER_STATUS;
```
- [edk2 definition](https://github.com/tianocore/edk2/blob/c4d29cb62187060493a1f595083ddfb6dd346f39/MdeModulePkg/Include/Guid/MmCommBuffer.h#L43-L61)

```rust
typedef struct {
  ///
  /// Whether the data in the fixed MM communication buffer is valid when entering from non-MM to MM.
  ///
  BOOLEAN    IsCommBufferValid;

  /// For padding purpose
  UINT8      Reserved[7];

  ///
  /// The return status when returning from MM to non-MM.
  ///
  UINT64     ReturnStatus;

  ///
  /// The size in bytes of the output buffer when returning from MM to non-MM.
  ///
  UINT64     ReturnBufferSize;
} MM_COMM_BUFFER_STATUS;
```

Also, none of these are packed. That is okay given the alignment that was added in the other repos (but missing here).

Given all of this, what's the holistic plan for everything to line up?

kuqin12: Yeah, that is our current approach to bifurcate the target communication channel.

I think at some point we will need to update the SMM control implementation to consolidate the register values, i.e. rdx being 0xdeadbeeffeedf00d to talk to user mode, anything else will be talking to the user.

The originally packed structure in rust actually caused the system to crash :) So I removed the pack when I was doing the testing. But I agree that we should add the unit8 here to match one of the c implementations.

makubacki: Packed should generally be used for HOBs because they are data shared between separately compiled binaries and environments (e.g. PEI arch vs DXE arch). It should have been in the original C structure. But, yes, given the padding came into the C definition a few months ago, it should be reflected here now.

components/patina_mm/tests/patina_mm_integration/common/handlers.rs
```diff
@@ -17,11 +17,16 @@
...
 use crate::patina_mm_integration::common::constants::*;
+use r_efi::efi;
 
 extern crate alloc;
 use alloc::{string::String, vec::Vec};
-use zerocopy::{FromBytes, IntoBytes};
-use zerocopy_derive::*;
+pub use zerocopy::IntoBytes;
+
+// Import shared protocol types from patina_mm
```
makubacki: I don't think this comment is really needed. It appears to be a routine import.

```suggestion
```

components/patina_mm/tests/patina_mm_integration/common/handlers.rs
```diff
@@ -180,40 +127,40 @@ impl MmSupervisorHandler {
 
     fn handle_get_info_request(&self) -> MmHandlerResult<Vec<u8>> {
         let response_header = MmSupervisorRequestHeader {
-            signature: mm_supv::REQUEST_SIGNATURE,
-            revision: mm_supv::REVISION,
-            request: mm_supv::requests::VERSION_INFO,
+            signature: SIGNATURE,
```
makubacki: For `SIGNATURE` and `REVISION`, can you please include `mm_supervisor_request` like `mm_supervisor_request::SIGNATURE`? That will algin with normal convention for clarity.

components/patina_mm/tests/patina_mm_integration/mm_supervisor/communication_tests.rs
makubacki: Have you tried dropping `.to_vec()`? `.communicate()` is:

```rust
    pub fn communicate(
        &self,
        guid: &BinaryGuid,
        data: &[u8],
    ) -> Result<Vec<u8>, patina_mm::component::communicator::Status>
```

It seems like that would work.

sdk/patina/src/management_mode/comm_buffer_hob.rs
```diff
@@ -0,0 +1,171 @@
...
+/// UEFI MM Communicate Header
+///
+/// A standard header that must be present at the beginning of any MM communication buffer.
+///
+/// ## Notes
+///
+/// - This only supports V1 and V2 of the MM Communicate header format.
+#[derive(Debug, Clone, Copy)]
+#[repr(C)]
+pub struct EfiMmCommunicateHeader {
```
makubacki: Do you think this should go in a different module than `comm_buffer_hob`? It seems that this and potentially `MmCommBufferStatus` and `EfiMmEntryContext` would be more intuitive in a file like mm_comm.rs or comm_buffer.rs. Then, they could be accessed like `mm_comm::Header` and `mm_comm::Status`.

sdk/patina/src/management_mode.rs
```diff
@@ -0,0 +1,21 @@
+//! Management Mode (MM) SDK for Patina
+//!
+//! This crate provides the Management Mode (MM) related definitions for Patina.
+//!
+//! ## Features
+//!
+//! - `comm_buffer_hob`: Provides data and HOB definitions to support the MM communication.
+//! - `protocol`: Contains MM-related protocol structures and definitions.
+//!
```
makubacki: I think you can drop this. These are more "modules" than "features" but they're also easily discoverable without being listed here.

sdk/patina/src/management_mode/comm_buffer_hob.rs
kuqin12: Okay, the padding is added to the structure.

kuqin12: I will file a separate issue to track the MM control update.

components/patina_mm/tests/patina_mm_integration/common/handlers.rs
kuqin12: This is removed

components/patina_mm/tests/patina_mm_integration/common/handlers.rs
kuqin12: This is fixed up. Along with other places that have the same pattern.

sdk/patina/src/management_mode.rs
kuqin12: This is dropped.

sdk/patina/src/management_mode/comm_buffer_hob.rs
kuqin12: This is now moved to mm_cis under pi folder. To the same spirit, I also added the structure definition for mmst in that same file, because that is where c version defines it as well. This will be used later when user core is in pulled in.

sdk/patina/src/management_mode/comm_buffer_hob.rs
kuqin12: An issue was filed on the c side because it will likely touch c code side: https://github.com/microsoft/mu_feature_mm_supv/issues/630

components/patina_mm/tests/patina_mm_integration/mm_supervisor/communication_tests.rs
kuqin12: I dropped `to_vec` and it needs me to change all the consumer to drop the `&`. I am mostly perplexed about why the `to_bytes` are failing.

sdk/patina/src/management_mode/comm_buffer_hob.rs
kuqin12: And the `fiMmCommunicateHeader` is removed because it is already defined in the communication module.

sdk/patina/src/management_mode.rs
```diff
@@ -0,0 +1,21 @@
+//! Management Mode (MM) SDK for Patina
+//!
+//! This crate provides the Management Mode (MM) related definitions for Patina.
+//!
+//! ## Features
```
cfernald: Isn't "features" typically used for crate features? Maybe "modules"

sdk/patina/src/pi/spec_version.rs
```diff
@@ -0,0 +1,16 @@
+//! Platform Initialization Specification MM Core Interface
```
cfernald: MM Core Interface? It seems this is generic, right?

sdk/patina/src/pi/spec_version.rs
```diff
@@ -0,0 +1,16 @@
...
+//!
+//! This module contains definitions related to the MM Core Interface as defined
```
cfernald: description also seems wrong

sdk/patina/src/pi/protocols/communication.rs
```diff
@@ -91,18 +92,55 @@ pub struct Protocol {
...
+    }
+
+    /// Returns the size of the header in bytes.
+    pub const fn size() -> usize {
+        core::mem::size_of::<Self>()
+    }
+
+    /// Get the header GUID from the communication buffer.
+    ///
+    /// Returns `Some(guid)` if the buffer has been properly initialized with a GUID,
```
cfernald: outdated, doesnt return an option

sdk/patina/src/pi/mm_cis.rs
```diff
@@ -0,0 +1,398 @@
+//! Platform Initialization Specification MM Core Interface
```
cfernald: Maybe a better module title

sdk/patina/src/pi/mm_cis.rs
```diff
@@ -0,0 +1,398 @@
...
+pub const MM_MMST_REVISION_MAJOR: u32 = spec_version::PI_SPECIFICATION_MAJOR_REVISION;
+
+/// MMST minor revision, the same as the PI Specification minor revision.
+pub const MM_MMST_REVISION_MINOR: u32 = spec_version::PI_SPECIFICATION_MINOR_REVISION;
+
+/// PI Specification version encoded as `(major << 16) | minor`.
+pub const MM_SYSTEM_TABLE_REVISION: u32 = (MM_MMST_REVISION_MAJOR << 16) | MM_MMST_REVISION_MINOR;
+
+//
+// This gnarly POS of EFI_MM_CPU_IO_PROTOCOL is embedded in MMST,
```
cfernald: Jut making sure you meant for this comment to be checked in haha

sdk/patina/src/management_mode.rs
kuqin12: I thought I removed these already. Maybe it was copied multiple times. Will remove.

sdk/patina/src/pi/mm_cis.rs
kuqin12: I guess so :) This original header shocked me because there is a protocol inclusion inside. Jokes aside, will remove.

sdk/patina/src/management_mode.rs
kuqin12: This is removed.

sdk/patina/src/pi/spec_version.rs
kuqin12: Nope, this was copied from the cis file. This is now updated to PI spec version.

sdk/patina/src/pi/protocols/communication.rs
kuqin12: Updated. It just returns a guid now.

sdk/patina/src/pi/mm_cis.rs
kuqin12: Updated.

sdk/patina/src/pi/mm_cis.rs
kuqin12: Removed.

sdk/patina/src/pi/spec_version.rs
kuqin12: Updated.

components/patina_mm/tests/patina_mm_integration/mm_supervisor/communication_tests.rs
kuqin12: Resolving for now as this only impacts the tests at this point. Will see if the pattern needs tuning as the usage picks up.

PR #1419, opened 2026-03-18
[patina_dxe_core] Add info to assert on fragmented free memory in spin_locked_gcd

thomashinds:
## Description

Minor update to the assert in spin_locked_gcd.rs for the case where free memory is fragmented to include the address and attributes of offending regions in the error message for ease of debugging. Also converts the assert with message to an error print and assert so the issue can be logged in release builds as well, rather than silently ignored.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on my physical ARM platform in debug & release.

## Integration Instructions

n/a

### DIFF
--- a/patina_dxe_core/src/gcd/spin_locked_gcd.rs
+++ b/patina_dxe_core/src/gcd/spin_locked_gcd.rs
@@ -1174,16 +1174,24 @@ impl GCD {
                 let prev = &mut descriptors[write_idx - 1];
                 if prev.r#type == current.r#type
                     && prev.attribute == current.attribute
-                    && prev.physical_start + (prev.number_of_pages * UEFI_PAGE_SIZE as u64) == current.physical_start
+                    && prev.physical_start + uefi_pages_to_size!(prev.number_of_pages as usize) as u64
+                        == current.physical_start
                 {
                     // Free memory shouldn't even need to be merged because it should already be consistent and coalesced.
                     // If this fails to be true it can cause odd behavior if applications try to allocate blocks of free
                     // memory by address, which is a common pattern for OS loaders.
-                    debug_assert!(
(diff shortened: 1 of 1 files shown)


cfernald (approved):

patina_dxe_core/src/gcd/spin_locked_gcd.rs
```diff
@@ -1179,11 +1179,18 @@ impl GCD {
...
-                    debug_assert!(
-                        prev.r#type != efi::CONVENTIONAL_MEMORY,
-                        "Free memory is fragmented in memory descriptors!"
-                    );
-
+                    if prev.r#type == efi::CONVENTIONAL_MEMORY {
+                        log::error!(
+                            "Free memory is fragmented in memory descriptors! prev: {:#x}-{:#x} (attr: {:#x}), current: {:#x}-{:#x} (attr: {:#x})",
+                            prev.physical_start,
+                            prev.physical_start + (prev.number_of_pages * UEFI_PAGE_SIZE as u64),
```
makubacki: This might be a place to use [uefi_pages_to_size!](https://github.com/OpenDevicePartnership/patina/blob/4e014bd3d49bc194c4dd94c55797aa0c75988134/sdk/patina/src/macros.rs#L67)

thomashinds: Thanks - I've updated this to use that macro, though it does require an extra cast since the types in the descriptors are u64 rather than usize.

makubacki: No problem

PR #1421, opened 2026-03-19
patina_dxe_core: Add DxeDispatch service for driver dispatch

kat-perez:
## Description

Add a `DxeDispatch` service trait in the SDK and a `CoreDxeDispatch`
implementation in patina_dxe_core that delegates to the PI dispatcher.
The service is registered alongside other core services (MemoryManager,
PerfTimer, etc.) and consumed via dependency injection by components
that need to trigger driver dispatch passes.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- Built SBSA DXE core binary with `BootDispatcher` + `SimpleBootManager` consuming the service via DI
- Booted Windows ARM64 under QEMU SBSA-ref with Patina BDS handling the full boot flow
- Connect-dispatch interleaving discovered AHCI device, expanded partial device path, loaded Windows bootloader, ExitBootServices completed
- 106 on-system unit tests passed (0 fails)

## Integration Instructions

The `DxeDispatch` service is registered automatically by the DXE core.
Components consume it via dependency injection:

```rust
fn entry_point(self, dxe_dispatch: Service<dyn DxeDispatch>) -> Result<()> {
    dxe_dispatch.dispatch()?;
    Ok(())
}
```

### DIFF
--- a/patina_dxe_core/src/dxe_dispatch_service.rs
+++ b/patina_dxe_core/src/dxe_dispatch_service.rs
@@ -1,0 +1,38 @@
+//! DXE Core Dispatch Service
+//!
+//! Provides the [`CoreDxeDispatch`] service implementation, which exposes
+//! the PI dispatcher to components via dependency injection. This allows
+//! components to trigger DXE driver dispatch passes (e.g., to interleave
+//! controller connection with driver dispatch during boot).
+//!
+//! ## License
+//!
+//! Copyright (c) Microsoft Corporation.
+//!
+//! SPDX-License-Identifier: Apache-2.0
+//!
+use patina::{
+    component::service::{IntoService, dxe_dispatch::DxeDispatch},
+    error::Result,
+};
+
+use crate::{Core, PlatformInfo};
+
+/// DXE dispatch service backed by the PI dispatcher.
+#[derive(IntoService)]
+#[service(dyn DxeDispatch)]
+pub(crate) struct CoreDxeDispatch<P: PlatformInfo>(&'static Core<P>);
+
+#[coverage(off)]
+impl<P: PlatformInfo> CoreDxeDispatch<P> {
+    pub(crate) fn new(core: &'static Core<P>) -> Self {
+        Self(core)
+    }
+}
+
+#[coverage(off)]
+impl<P: PlatformInfo> DxeDispatch for CoreDxeDispatch<P> {
+    fn dispatch(&self) -> Result<bool> {
+        self.0.pi_dispatcher.dispatch()
+    }
+}
--- a/patina_dxe_core/src/lib.rs
+++ b/patina_dxe_core/src/lib.rs
@@ -77,6 +77,7 @@ mod cpu;
 mod debugger_reload;
 mod decompress;
 mod driver_services;
+mod dxe_dispatch_service;
 mod dxe_services;
 mod event_db;
 mod events;
@@ -401,7 +402,7 @@ impl<P: PlatformInfo> Core<P> {
     /// Initializes the core with the given configuration, including GCD initialization, enabling allocations.
     ///
     /// Returns the relocated HOB list pointer that should be used for all subsequent operations.
-    fn init_memory(&self, physical_hob_list: *const c_void) -> *mut c_void {
+    fn init_memory(&'static self, physical_hob_list: *const c_void) -> *mut c_void {
         log::info!("DXE Core Crate v{DXE_CORE_VERSION}");
 
         GCD.prioritize_32_bit_memory(P::MemoryInfo::prioritize_32_bit_memory());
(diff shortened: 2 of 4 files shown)


patina_dxe_core/src/lib.rs
```diff
@@ -362,6 +362,13 @@ impl<P: PlatformInfo> Core<P> {
         }
     }
 
+    /// Performs a single DXE driver dispatch pass.
+    ///
+    /// Returns `true` if any drivers were dispatched, `false` if no drivers remain.
+    pub fn dispatch(&'static self) -> Result<bool> {
```
os-d: I don't think making this public makes sense. @Javagedes do you have thoughts on this? Feels like a different paradigm is needed.

kat-perez: I could move this to the SDK instead?

Like this at sdk/patina/src/pi/dxe_services.rs?
```
/// Interface for dispatching DXE drivers.
///
/// Boot orchestrators use this to interleave controller connection
/// with driver dispatch.
pub trait DxeServices: Send + Sync + 'static {
    /// Run a single dispatch pass. Returns `true` if any drivers were dispatched.
    fn dispatch(&self) -> crate::error::Result<bool>;
}
```

Javagedes: Yeah, I don't think we want to make this public outside of the core. It leaks containment of functionality that the platform's top level `main()` function should not have access to, which could result in platforms doing bad things.

Based off of the current usage:

```rust
impl DxeServices for CoreDxeServices {
    fn dispatch(&self) -> patina::error::Result<bool> {
        CORE.dispatch()
    }
}

// Boot orchestration with connect-dispatch interleaving
add.component(BootDispatcher::new(
  SimpleBootManager::new(
    BootConfig::new(create_boot_path())
      .with_failure_handler(|| log::error!("Boot failed: all boot options exhausted")),
    ),
    CoreDxeServices,
));
```

I think one of these two options better lean into the patina component model via dependency injection:

1. Create a new service trait defined in the `patina` crate for either (1) only dispatch or (2) the entire dxe services table, and implemented in the core (Similar to the [CoreMemoryManager](https://github.com/OpenDevicePartnership/patina/blob/904c49e1b1d1906f8041857670ded27d3e5e4322/patina_dxe_core/src/lib.rs#L476)), then consume it through dependency injection
2. Do the same as 1, but instead do it for configuration table functionality, then you can grab the dxe services configuration table and call dispatch from [dxe_services](https://github.com/OpenDevicePartnership/patina/blob/904c49e1b1d1906f8041857670ded27d3e5e4322/patina_dxe_core/src/dxe_services.rs#L409)

kat-perez: Thanks, I'll go with (1.) by making a new service trait instead. Would be something like `CoreDxeDispatch`, implemented like CoreMemoryManager

kat-perez: The platform binary will never touch `dispatch()` directly

kat-perez:
@os-d @Javagedes this is ready for another review. thanks!

sdk/patina/src/component/service/dxe_dispatch.rs
```diff
@@ -0,0 +1,28 @@
...
+/// Service interface for DXE driver dispatch.
+///
+/// Provides access to the PI dispatcher for components that need to trigger
+/// driver dispatch passes (e.g., to interleave controller connection with
+/// driver dispatch during boot).
+#[cfg_attr(any(test, feature = "mockall"), automock)]
+pub trait DxeDispatch {
+    /// Performs a single DXE driver dispatch pass.
+    ///
+    /// Returns `true` if any drivers were dispatched, `false` if no drivers remain.
```
Javagedes: I believe `false` means no drivers were dispatched, not necessarily that there are no more remaining drivers.

patina_dxe_core/src/dxe_dispatch_service.rs
```diff
@@ -0,0 +1,74 @@
...
+/// Holds a function pointer to the concrete dispatch implementation,
+/// set at construction during Core initialization.
+#[derive(IntoService)]
+#[service(dyn DxeDispatch)]
+pub(crate) struct CoreDxeDispatch(fn() -> Result<bool>);
+
+impl CoreDxeDispatch {
+    /// Create a new dispatch service with the given dispatch function.
+    pub(crate) fn new(dispatch_fn: fn() -> Result<bool>) -> Self {
+        Self(dispatch_fn)
```
os-d: Why is this configurable? Can we not just always use the pi_dispatcher?

Javagedes: Probably for mocking.

os-d: Can't the whole trait just be mocked?

Javagedes: Fair :) I concur with @os-d we should just hardcode in the usage of the core instead of having the wrapper like this.

Also if you update `init_memory` to be `&'static self` (which I think is fair), then I think you could even do something like this:

```rust

pub(crate) struct CoreDxeDispatch<P: PlatformInfo>(&'static Core<P>);

// ...

component_dispatcher.add_service(dxe_dispatch_service::CoreDxeDispatch::new(self);

```

I'm not 100% confident in that though.

kat-perez:
@os-d @Javagedes does this new implementation work better?

kat-perez:
@makubacki added this in the inline docs:
```
/// Note: The DXE core already runs a PI dispatch loop automatically. This
/// service is only needed when a component must explicitly trigger a dispatch
/// pass at a specific point in its execution.
pub trait DxeDispatch {
...
```

PR #1422, opened 2026-03-19
[feature/patina-boot] patina_boot: Add connect-dispatch interleaving with DxeServices

kat-perez:
## Description

Interleave controller connection with DXE driver dispatch during device
enumeration. Connecting controllers can discover new firmware volumes
(e.g., PCI option ROMs) that contain drivers for devices behind that
controller. Those drivers must be dispatched before the next round of
enumeration, otherwise the devices they serve will not be found.

`SimpleBootManager` uses `interleave_connect_and_dispatch()` to alternate
between connecting controllers and dispatching newly-discovered drivers
until both stabilize. The `DxeDispatch` service trait (from #1421) is
consumed via dependency injection.

Note: `interleave_connect_and_dispatch()` currently uses `connect_all()`
which connects every controller on every round. This is functional but
inefficient for platforms with large device topologies — a future
optimization could connect only newly-discovered controllers.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- Built SBSA DXE core binary with `BootDispatcher` + `SimpleBootManager` replacing TianoCore BdsDxe
- Booted Windows ARM64 under QEMU SBSA-ref emulation with Patina BDS handling the full boot flow
- Verified interleaving: controller connection discovered AHCI device, partial device path expanded to full path, Windows bootloader loaded, ExitBootServices completed

## Integration Instructions

Depends on #1421 (`DxeDispatch` service trait) for platform binary integration.

Remove TianoCore `BdsDxe.inf` from platform DSC/FDF since the `BootDispatcher` provides the BDS architectural protocol.

### DIFF
--- a/components/patina_boot/README.md
+++ b/components/patina_boot/README.md
@@ -4,22 +4,24 @@ Boot orchestration component for Patina-based firmware implementing UEFI boot ma
 
 ## Components
 
-- **BootOrchestrator**: Orchestrates device enumeration, BDS phase events, and boot option execution.
-- **ConsoleDiscovery**: Discovers console devices and populates UEFI console variables.
+- **BootDispatcher**: Installs the BDS architectural protocol and delegates to a `BootOrchestrator` implementation.
+- **BootOrchestrator**: Trait for custom boot flows. Platforms implement this to define boot behavior.
+- **SimpleBootManager**: Default `BootOrchestrator` for platforms with straightforward boot topologies.
 
 ## Usage
 
 ```rust
-use patina_boot::{component::BootOrchestrator, config::BootOptions};
+use patina_boot::{BootDispatcher, SimpleBootManager, config::BootConfig};
 
-// Configure boot options (devices are tried in order)
-let config = BootOptions::new()
-    .with_device(primary_boot_path)
-    .with_device(fallback_boot_path)
-    .with_failure_handler(|| { /* handle boot failure */ });
+// Minimal boot:
+let orchestrator = SimpleBootManager::new(
+    BootConfig::new(nvme_esp_path())
+        .with_device(nvme_recovery_path()),
+);
+add.component(BootDispatcher::new(orchestrator));
 
-// Add BootOrchestrator as a platform component
-add.component(BootOrchestrator);
+// Custom orchestrator:
+add.component(BootDispatcher::new(MyCustomOrchestrator::new()));
 ```
 
 ## Helper Functions
--- a/components/patina_boot/src/orchestrators/simple_boot_manager.rs
+++ b/components/patina_boot/src/orchestrators/simple_boot_manager.rs
@@ -23,7 +23,39 @@ use patina::{
 };
 use r_efi::efi;
 
+use patina::component::service::dxe_dispatch::DxeDispatch;
+
+use patina::boot_services::BootServices;
+
 use crate::{boot_orchestrator::BootOrchestrator, config::BootConfig, helpers};
+
+/// Interleave controller connection with DXE driver dispatch.
+///
+/// Alternates between connecting all controllers and dispatching newly loaded
+/// drivers (e.g., PCI option ROMs) until both the device topology stabilizes
+/// and no new drivers are dispatched.
+///
+/// This ensures that drivers loaded from firmware volumes during device
+/// enumeration (such as PCI option ROM drivers) are dispatched before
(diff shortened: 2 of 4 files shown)


components/patina_boot/README.md
```diff
@@ -1,33 +1,36 @@
...
-## Helper Functions
-
-For custom boot flows, use the helper functions in the `helpers` module:
-
-- `connect_all()` - Connect all controllers for device enumeration
-- `signal_bds_phase_entry()` - Signal EndOfDxe event
-- `signal_ready_to_boot()` - Signal ReadyToBoot event
-- `discover_console_devices()` - Populate console variables
-- `boot_from_device_path()` - Load and start a boot image
+# Patina Boot
```
os-d, line 1: line endings change? Is this the correct or inocrrect?

components/patina_boot/README.md
```diff
@@ -1,33 +1,36 @@
...
+
+// Custom orchestrator:
+add.component(BootDispatcher::new(MyCustomOrchestrator::new(), my_dxe_services));
+```
+
+## Helper Functions
+
+For custom boot flows, use the helper functions in the `helpers` module:
+
+- `interleave_connect_and_dispatch()` - Interleave controller connection with DXE driver dispatch
```
os-d: Worth noting this is connecting all controllers (which is not the recommend path for most platforms)

os-d: As such, why put it in helpers? Maybe this should just go in the simple boot manager?

components/patina_boot/README.md
kat-perez: Yeah, the original README had CRLF and I switched it to LF to match everything else

components/patina_boot/README.md
kat-perez: I can document it. I thought it would be something that could be shared but since simple boot manager is the only thing implementing it right now, I can just move it there

os-d: Connect all is what really defines the simple boot manager from a not simple one, though, so I don't anticipate this would be used as much. Or, to state that differently, for standard platforms, we don't want to advertise the connect all route.

kat-perez: Yeah, makes sense. I've moved it out of helpers

components/patina_boot/src/orchestrators/simple_boot_manager.rs
```diff
@@ -23,16 +23,49 @@ use patina::{
...
+/// continuing enumeration, allowing those drivers to bind to newly
+/// discovered controllers.
+fn interleave_connect_and_dispatch<B: BootServices, D: DxeDispatch + ?Sized>(
+    boot_services: &B,
+    dxe_services: &D,
+) -> patina::error::Result<()> {
+    const MAX_ROUNDS: usize = 10;
+
+    for round in 0..MAX_ROUNDS {
+        helpers::connect_all(boot_services)?;
```
makubacki, line 49: Every iteration except the last (10th) will leave with `connect_all()` having been called with all drivers dispatched before exiting.

I think a debug assert (given the number of rounds) or something to distinguish the case in the return value when the loop fails to exhaust dispatchable drivers would at least let the caller know they might need to take some action (call again, log error, etc.).

components/patina_boot/README.md
```diff
@@ -1,33 +1,35 @@
...
+
+// Minimal boot:
+let orchestrator = SimpleBootManager::new(
+    BootConfig::new(nvme_esp_path())
+        .with_device(nvme_recovery_path()),
+);
+add.component(BootDispatcher::new(orchestrator, my_dxe_services));
+
+// Custom orchestrator:
+add.component(BootDispatcher::new(MyCustomOrchestrator::new(), my_dxe_services));
```
makubacki: It looks like `BootDispather::new()` only takes one argument:

```rust
    pub fn new(orchestrator: impl BootOrchestrator) -> Self {
        Self { orchestrator: Box::new(orchestrator) }
    }
```

kat-perez: Updated, thanks

components/patina_boot/src/orchestrators/simple_boot_manager.rs
kat-perez: Added in the debug assert

PR #1423, opened 2026-03-20
[patina_internal_cpu] remove alloc usage

kuqin12:
## Description

Patina internal CPU hosts the interrupt manager and other fundamental functionalities, which should not work on top of the allocator.

Admittedly, the current crate can support more advanced components that works on top of exception forwarding, but this should not be treated as a required dependency.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This was tested on QEMU Q35 and booted to UEFI shell.

## Integration Instructions

N/A

### DIFF
--- a/patina_dxe_core/src/cpu/cpu_arch_protocol.rs
+++ b/patina_dxe_core/src/cpu/cpu_arch_protocol.rs
@@ -414,10 +414,8 @@ mod tests {
     #[test]
     fn test_dxe_interrupt_manager_register_delegates() {
         let dxe_interrupt_manager = DxeInterruptManager(Interrupts::default());
-        let result = dxe_interrupt_manager.register_exception_handler(
-            ExceptionType::from(0_usize),
-            HandlerType::UefiRoutine(mock_interrupt_handler),
-        );
+        let result = dxe_interrupt_manager
+            .register_exception_handler(ExceptionType::from(0_usize), HandlerType::UefiRoutine(mock_interrupt_handler));
         assert!(result.is_ok());
     }
 


core/patina_internal_cpu/src/paging/null.rs
```diff
@@ -56,6 +55,6 @@ where
 /// Used to specify that this architecture paging implementation is not supported.
 pub fn create_cpu_null_paging<A: PageAllocator + 'static>(
     _page_allocator: A,
-) -> Result<Box<dyn PatinaPageTable>, efi::Status> {
+) -> Result<EfiCpuPagingNull<A>, efi::Status> {
```
kuqin12: This does not look right.

core/patina_internal_cpu/src/paging/x64.rs
```diff
@@ -9,6 +9,7 @@
 //! SPDX-License-Identifier: Apache-2.0
 //!
 use crate::paging::{CacheAttributeValue, PatinaPageTable};
+#[cfg(feature = "alloc")]
```
kuqin12: Not needed

core/patina_internal_cpu/src/paging.rs
```diff
@@ -12,10 +12,10 @@
 use patina_paging::{MemoryAttributes, PtError};
 
 cfg_if::cfg_if! {
-    if #[cfg(all(target_arch = "x86_64"))] {
```
kuqin12, line 15: revert

kuqin12: Reverted

core/patina_internal_cpu/src/paging/x64.rs
kuqin12: Removed.

core/patina_internal_cpu/src/paging/null.rs
kuqin12: Reverted

core/patina_internal_cpu/src/interrupts/aarch64/interrupt_manager.rs
```diff
@@ -7,12 +7,14 @@
...
+#[cfg(feature = "alloc")]
+use patina::component::service::IntoService;
 use patina::{
     base::{UEFI_PAGE_MASK, UEFI_PAGE_SIZE},
     bit,
-    component::service::IntoService,
     error::EfiError,
 };
+#[cfg(feature = "alloc")]
 use patina_paging::PageTable;
```
makubacki, line 15: Design-wise it seems weird to have:

```rust
#[cfg(feature = "alloc")]
use patina_paging::PageTable;
```

But no cfg gate in core/patina_internal_cpu/src/interrupts/x64/interrupt_manager.rs:

```rust
use patina_paging::PageTable;
```

makubacki: My understanding is that this is imported so the `PageTable` trait is in scope for `.dump_page_tables()` to resolve in `let _ = pt.dump_page_tables(far & !(UEFI_PAGE_MASK as u64), UEFI_PAGE_SIZE as u64);` later in the file.

Is that the case? If so, that would no longer work with this behind `alloc` and that code not.

core/patina_internal_cpu/src/cpu/stub.rs
```diff
@@ -17,8 +18,9 @@ use r_efi::efi;
 /// Struct to implement Null Cpu Init.
 ///
 /// This struct cannot be used directly. It replaces the `EfiCpu` struct when not compiling for x86_64 or AArch64 UEFI architectures.
-#[derive(Default, Copy, Clone, IntoService)]
-#[service(dyn Cpu)]
+#[derive(Default, Copy, Clone)]
+#[cfg_attr(feature = "alloc", derive(IntoService))]
```
makubacki: I think this is okay, but it would be a little more direct/clear to have something like a `patina_component` feature that is dependent on `alloc`. However, that adds more features/complexity. Feel free to ignore.

core/patina_internal_cpu/src/interrupts/aarch64/interrupt_manager.rs
kuqin12: This is a good observation. Indeed you are right that original inclusion was needed to resolve the `dump_page_tables` call. Will remove the change as it is not enforcing the alloc usage in the first place.

As per why this was not caught anywhere, this is because patina has default set to ['alloc']. So as long as any subcrate pulling in patina, it will pick up the default feature and enable alloc globally. I will create a sub-issue and handle that in a separate PR.

core/patina_internal_cpu/src/cpu/stub.rs
kuqin12: The entire `patina_component` is guarded behind "alloc" feature as of today. But I agree that we should mention it somewhere.

core/patina_internal_cpu/src/cpu/aarch64/cpu.rs
```diff
@@ -9,8 +9,9 @@
 use crate::cpu::Cpu;
 #[cfg(all(not(test), target_arch = "aarch64"))]
 use core::arch::asm;
+#[cfg(feature = "alloc")]
+use patina::component::service::IntoService;
```
cfernald: Is `IntoService` the only use case here? Have you thought about just removing this from `patina_internal_cpu` and instead have dxe core carry the service wrapper? It would create some boiler plate code, but would mean that we could remove the component code from this crate which seems tempting.

If you disagree, then i'm good with the change as it is.

core/patina_internal_cpu/src/lib.rs
```diff
@@ -11,8 +11,11 @@
 #![cfg_attr(all(not(feature = "std"), not(test)), no_std)]
 #![feature(abi_x86_interrupt)]
 #![feature(coverage_attribute)]
+
+#[cfg(feature = "alloc")]
 extern crate alloc;
 
 pub mod cpu;
 pub mod interrupts;
+#[cfg(feature = "alloc")]
```
cfernald: I wonder if this paging module actually belongs here. If we envision the CPU crate being a common crate abstracting just CPU mechanisms, then perhaps this module belongs in the dxe_core instead since it only exists to abstract the DXE specific MTRR management. 

Perhaps we just need to better define the purpose of this crate.

core/patina_internal_cpu/src/cpu/stub.rs
makubacki: Yeah, I remember that happening recently here:
https://github.com/OpenDevicePartnership/patina/blob/fa9f9ed3f62e6573b06d08d30614787b2b6b633c/sdk/patina/src/lib.rs#L41-L42

It's mainly that a blanket `alloc` switch kind of silently hides component functionality with components being so important in Patina that gives some pause. Again, I think this is still fine though. You could mention it in the [crate readme](https://github.com/OpenDevicePartnership/patina/tree/fa9f9ed3f62e6573b06d08d30614787b2b6b633c/core/patina_internal_cpu) if you like.

core/patina_internal_cpu/src/lib.rs
```diff
@@ -11,8 +11,11 @@
...
 pub mod paging;
```
cfernald, line 17: An additional thought. I'm not sure I _like_ this, but you can push the box up to the core by just returning the direct paging type and having the core do the boxing before taking dependence on the type.

i.e.
```rust
let page_table = alloc::boxed::Box::new(create_cpu_paging(page_allocator).expect("Failed to create CPU page table"))
        as alloc::boxed::Box<dyn patina_internal_cpu::paging::PatinaPageTable>;
```

core/patina_internal_cpu/src/cpu/aarch64/cpu.rs
makubacki: I think that would be cleaner conceptually.

core/patina_internal_cpu/src/lib.rs
os-d: Yeah, the design point of this module (which would be great to eventually get rid of with PAT) is really just so that DXE core can make architecture agnostic calls. That is relevant for cores in general, if we think about StMM core or PEI core. Supervisor core is a different story because it is inherently architecturally-aware. So any MTRR management it could do directly.

But, I think the layer still makes sense here until we switch to PAT, then we drop this layer and say Patina just uses permissive MTRRs, drivers can do what they want.

core/patina_internal_cpu/src/lib.rs
kuqin12: I started this change that way and decided it was changing dxe core unnecessarily. But I will move to change the core instead and see if we can decouple alloc completely.

core/patina_internal_cpu/src/lib.rs
cfernald: You know where I stand on leaving the MTRRs static, but idealism often dies in reality, so I won't push that too hard haha.

core/patina_internal_cpu/src/lib.rs
kuqin12: Should be fixed now. GCD will box it with the current function.

core/patina_internal_cpu/src/interrupts/aarch64/interrupt_manager.rs
kuqin12: This is reverted.

core/patina_internal_cpu/src/lib.rs
kuqin12: The original change is reverted as we updated the interface. But the fact that we wrapped a MTRR instance into a page table instance is understandable but a bit odd.

core/patina_internal_cpu/src/cpu/aarch64/cpu.rs
kuqin12: Actually, it seems that I do not need this cpu sub-module completely. I can just guard the entire sub-module and do not touch the content in here. In that case, it will make less sense to make such move because there will no usage to benefit from it.

But we will still need the split for interrupt manager... Any preference here?

Javagedes: Easier to maintain in the long run if we can remove all allocations, if its reasonable.

kuqin12: This is updated.

core/patina_internal_cpu/src/cpu/stub.rs
kuqin12: After the re-refactor, I updated the md file to mention that this crate should be alloc free

core/patina_internal_cpu/src/interrupts/aarch64/interrupt_manager.rs
```diff
@@ -36,10 +35,10 @@ cfg_if::cfg_if! {
     }
 }
 /// AARCH64 Implementation of the InterruptManager.
-#[derive(Default, Copy, Clone, IntoService)]
-#[service(dyn InterruptManager)]
+#[derive(Default, Copy, Clone)]
 pub struct InterruptsAarch64 {}
 
+#[allow(dead_code)]
```
cfernald, line 41: Did you figure out why this is needed? It's a public struct, so the compiler shouldn't care that it is dead since it potentially has external consumers, right?

cfernald (approved):

core/patina_internal_cpu/src/interrupts/aarch64/interrupt_manager.rs
kuqin12: It seems that interrups.rs does not re-export the x64 as public: https://github.com/OpenDevicePartnership/patina/blob/1ed7a98f6939b2ef3c6ce539d63a2190d8f5b114/core/patina_internal_cpu/src/interrupts.rs#L31. And `InterruptsX64` is aliased, and new/initialize is not in a public trait, so the dead code checker will flag this section as not used.

makubacki: I assume the issue is that it is not (clearly) reachable through a public path.

`interrupt_manager` is private in core\patina_internal_cpu\src\interrupts\aarch64.rs:

```rust
mod interrupt_manager;
```

It's publicly re-exported in core\patina_internal_cpu\src\interrupts\aarch64.rs:

```rust
#[allow(unused)]
pub use interrupt_manager::InterruptsAarch64;
```

But, the `aarch64` module itself is not public in core\patina_internal_cpu\src\interrupts.rs:

```rust
#[coverage(off)]
#[cfg(any(target_arch = "aarch64", test))]
mod aarch64;
```

It's made public through the type alias in core\patina_internal_cpu\src\interrupts.rs:

```rust
    } else if #[cfg(target_arch = "aarch64")] {
        pub type Interrupts = aarch64::InterruptsAarch64;
```

I'm not sure if that overcomes the private modules mentioned above. A lot of code is also gated for host tests.

Given there are at least 7 instances of `#[cfg(all(not(test), target_arch = "aarch64"))]` in core\patina_internal_cpu\src\interrupts\aarch64\interrupt_manager.rs, it might be worth splitting it into a `hw.rs` file and a `hw_stub.rs` file and try to have a single `cfg` gate in interrupt_manager.rs select between them (or something similar).

cfernald: Kun and I discussed offline. I think the architecture abstraction here need to be unified. I don't think we do it as part of this PR though.

makubacki: @kuqin12, can you please file an issue to track refactoring it?

kuqin12: Aye, created https://github.com/OpenDevicePartnership/patina/issues/1435 to track the refactoring.

core/patina_internal_cpu/src/interrupts/aarch64/interrupt_manager.rs
kuqin12: > As per why this was not caught anywhere, this is because patina has default set to ['alloc']. So as long as any subcrate pulling in patina, it will pick up the default feature and enable alloc globally. I will create a sub-issue and handle that in a separate PR.

Created this issue to try to untangle it: https://github.com/OpenDevicePartnership/patina/issues/1436

PR #1430, opened 2026-03-24
Add option to override hw_print_level for specific log targets

### LINKED ISSUE #1429: [Feature]: Enable changing/ignoring Advanced Logger `hw_print_level` for specific log targets
### Feature Overview

Closely related to #1427 - platforms may have specific wants and needs for verbosity of logs on the hardware serial port (in addition to the existing configurability of log verbosity in memory / to file). 

For example, as I understand it, there is no way to configure the serial logger to print only error messages, plus warnings from a specific module, while logging everything at INFO and above to memory/file.

Though I don't think there's are exactly equivalent options in an EDK2 build today, EDK2 does have the ability to fine tune verbosity more via additional bits in the `PcdDebugPrintErrorLevel` bitmask, like `DEBUG_LOAD` for example.

### Solution Overview

To enable this use case, the Advanced Logger structure could consume a list of log targets from the platform at the time of construction, similar to the current `target_filters`. Then, records matching a listed target could bypass the platform-wide `hw_print_level`. Or, an overriding `hw_print_level` could be provided along with each target, in bitmask form or otherwise. This would only apply to records generated from within the Patina binary (with an associated log target string).

For example:
```rs
    // as a bitmask
        let logger = AdvancedLogger::<UartNull>::new(
            Format::Standard,
            &[("test_target", log::LevelFilter::Info)],
            log::LevelFilter::Debug,
            serial,
            &[("test_target", 0x80000002)] // enable hw serial logging of error and warn records
        );

   // or in another way:
        let logger = AdvancedLogger::<UartNull>::new(
            Format::Standard,
            &[("test_target", log::LevelFilter::Info)],
            log::LevelFilter::Debug,
            serial,
            &[("test_target", HwPrintLevel::Error | HwPrintLevel::Warn)] // enable hw serial logging of error and warn records
        );
```

I have a draft PR as an example of how this could look: #1430 

### Alternatives Considered

- Handle platform requests to tune hw serial log verbosity centrally on a case-by-case basis (such as #1427 ).
  * Potential for issues when platforms/implementers/maintainers disagree on what is appropriate. For example, if I want stack traces treated like an error log record in release builds but others do not.
- Crate features for specific verbosity configuration (e.g. `promote_image_load_to_error`, `promote_stack_trace_to_error`), and just use the regular `hw_print_level` platform wide bitmask
- Use more of the `hw_print_level` bitmask than the current way of mapping log crate levels to bitmask bits. (i.e. introduce equivalents of DEBUG_LOAD etc). I haven't explored this, but it seems like this would be challenging to integrate cleanly with the `log` crate.

### Urgency

Medium

I will implement the feature

Maintainer feedback requested

### Anything else?

_No response_

thomashinds:
## Description

Allows platforms to provide hardware log level filters for specific log targets, allowing fine-tuning of the serial port logs. For example, allows configuring certain info / warn level records to print to serial when the platform-wide `hw_print_level` is set to error-only.

Implements this by adding a new `hw_filter_override: Option<log::LevelFilter>` field to the `target_filters` passed to `AdvancedLogger::new` (which are now a new struct instead of a tuple). If present, that filter is used for determining whether to print messages matching the given target prefix to the serial port, instead of the platform-wide mask from the log header.

Resolves #1429 

- [x] Impacts functionality?
- [ ] Impacts security?
- [x] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Verified HW filtering behavior works as expected on a physical platform (ARM).

## Integration Instructions

Change `target_filters` parameter of `AdvancedLogger::new()` to use a list of the new `TargetFilter` struct instead of the previous tuple of `(&'a str, log::LevelFilter)`.

For example:
```rust
static LOGGER: AdvancedLogger<UartNull> = AdvancedLogger::new(
   Format::Standard, // How logs are formatted
   &[("allocations", LevelFilter::Off)], // set custom log levels per module
   log::LevelFilter::Info, // Default log level
   UartNull { }, // Serial writer instance
);
```
becomes
```rust
static LOGGER: AdvancedLogger<UartNull> = AdvancedLogger::new(
   Format::Standard, // How logs are formatted
   &[TargetFilter { target: "allocations", log_level: LevelFilter::Off, hw_filter_override: None }], // set custom log levels per module
   log::LevelFilter::Info, // Default log level
   UartNull { }, // Serial writer instance
);
```

### DIFF
--- a/components/patina_adv_logger/src/logger.rs
+++ b/components/patina_adv_logger/src/logger.rs
@@ -38,8 +38,8 @@ pub struct TargetFilter<'a> {
     /// Optional override for the hardware print level for this target. Messages above this level will not be printed
     /// to the hardware port, but may still be logged to the memory log based on log_level and the overall max_level.
     /// - `None` = use global `hw_print_level` from memory log header.
-    /// - `Some(level_filter)` Use the provided level filter to control hardware printing for this target, instead 
-    /// of the global `hw_print_level`.
+    /// - `Some(level_filter)` Use the provided level filter to control hardware printing for this target, instead
+    ///   of the global `hw_print_level`.
     pub hw_filter_override: Option<log::LevelFilter>,
 }
 


components/patina_adv_logger/src/logger.rs
```diff
@@ -54,12 +55,16 @@ where
...
+    /// * `hw_target_filters` - Per-target overrides for the hardware print level bitmask.
+    ///   If a log record's target starts with the name, the associated bitmask is used
+    ///   instead of the hw_print_level from the memory log header. Use an empty slice for no overrides.
     ///
     pub const fn new(
         format: Format,
         target_filters: &'a [(&'a str, log::LevelFilter)],
         max_level: log::LevelFilter,
         hardware_port: S,
+        hw_target_filters: &'a [(&'a str, u32)],
```
cfernald: I think if we continue with this, consider the `pub fn with_hw_target_filters(Self) -> Self` function instead of expanding `new`. This avoids breaking changes and allows `new` to just handle the default for users that don't care.

components/patina_adv_logger/src/logger.rs
```diff
@@ -54,12 +55,16 @@ where
     /// * `target_filters` - A list of target filters to apply to the logger.
     /// * `max_level` - The maximum log level to log.
     /// * `hardware_port` - The hardware port to write logs to.
+    /// * `hw_target_filters` - Per-target overrides for the hardware print level bitmask.
+    ///   If a log record's target starts with the name, the associated bitmask is used
+    ///   instead of the hw_print_level from the memory log header. Use an empty slice for no overrides.
     ///
     pub const fn new(
         format: Format,
         target_filters: &'a [(&'a str, log::LevelFilter)],
```
cfernald: Did you consider just adding a new field to the existing filter i.e. (target, memory_log_level, hw_log_level)? Might make the intersection between the filters more understandable.

thomashinds: I like this idea - I do think it would be the clearest to understand

makubacki: Overall, having to coordinate the two lists is my main concern. For example, having to check if a target is in `target_filters` to determine if `hw_target_filters` is relevant.

I think a `TargetFilter` type that organizes the target and associated level/masks would clean the code up in several places, like:

```rust
pub struct TargetFilter<'a> {
    /// Target name prefix to match
    pub target: &'a str,
    /// Maximum log level for this target. Messages above this are dropped entirely.
    pub log_level: log::LevelFilter,
    /// Optional override for the hardware print level bitmask.
    /// - `None` = use global `hw_print_level` from memory log header.
    /// - `Some(mask)` = use this bitmask instead.
    pub hw_print_mask: Option<u32>,
}
```

This more clearly binds a single target with associated levels/masks and how they work together.

Then, the `AdvancedLogger` struct above can have a single entry for this (`target_filters`):

```rust
pub struct AdvancedLogger<'a, S: SerialIO + Send> {
    hardware_port: S,
    target_filters: &'a [TargetFilter<'a>],
    max_level: log::LevelFilter,
    format: Format,
    memory_log: RwLock<Option<AdvancedLogWriter>>,
    pub(crate) timer: Service<dyn ArchTimerFunctionality>,
}
```

It should also simplify the `enabled()` and `log()` logic since locating a single `TargetFilter` gives both pieces of information for the target.

components/patina_adv_logger/src/logger.rs
cfernald: Sounds like the lists will be merged, so this is moot.

components/patina_adv_logger/src/logger.rs
```diff
@@ -51,13 +67,13 @@ where
...
     /// * `format` - The format to use for logging.
-    /// * `target_filters` - A list of target filters to apply to the logger.
+    /// * `target_filters` - Per-target filters that control log level and optionally the hardware print mask.
     /// * `max_level` - The maximum log level to log.
     /// * `hardware_port` - The hardware port to write logs to.
     ///
     pub const fn new(
         format: Format,
-        target_filters: &'a [(&'a str, log::LevelFilter)],
+        target_filters: &'a [TargetFilter<'a>],
```
cfernald, line 74: Unless this change is urgent, I would target the major branch to avoid churn.

components/patina_adv_logger/src/logger.rs
```diff
@@ -401,4 +433,86 @@ mod tests {
...
     }
+
+    // Helper to build Metadata for a given target and level.
+    fn metadata(target: &str, level: log::Level) -> log::Metadata<'_> {
+        log::Metadata::builder().target(target).level(level).build()
+    }
+
+    // === Global level filtering (no target filters) ===
+
+    #[test]
```
cfernald, line 460: Please put the tests in a test mod at the end of the file with the appropriate cfg(test) and coverage(off) tags.

components/patina_adv_logger/src/memory_log.rs
```diff
@@ -32,9 +32,13 @@ pub const ADV_LOGGER_INFO_VERSION_V5: u16 = 5;
 pub const ADV_LOGGER_INFO_VERSION_V6: u16 = 6;
 
 // UEFI Debug Levels
+/// Error
 pub const DEBUG_LEVEL_ERROR: u32 = 0x80000000;
```
cfernald, line 36: I don't think these definitions are currently public, are they? Users will need access to it to get to the value needed for the hw filter option. Though this may become moot, see my comment on that attribute.

components/patina_adv_logger/src/logger.rs
```diff
@@ -24,18 +24,34 @@ use patina::{
...
+    /// Target name prefix to match.
+    pub target: &'a str,
+    /// Maximum log level for this target. Messages above this are dropped entirely.
+    pub log_level: log::LevelFilter,
+    /// Optional override for the hardware print level bitmask.
+    /// - `None` = use global `hw_print_level` from memory log header.
+    /// - `Some(mask)` = use this bitmask instead. Compose from
+    ///   [`DEBUG_LEVEL_ERROR`], [`DEBUG_LEVEL_WARNING`],
+    ///   [`DEBUG_LEVEL_INFO`], and [`DEBUG_LEVEL_VERBOSE`].
+    pub hw_mask_override: Option<u32>,
```
cfernald: Why not use `log::LevelFilter`? The target filters only apply for rust `log` events, so we might as well be consistent with the level definitions instead of using the EDK2 style levels that don't match perfectly. The only downside I see if that you can't skip a level, but not sure if that is needed.

components/patina_adv_logger/src/logger.rs
makubacki: I agree. @thomashinds, I've reset the `major` branch to the current `main` branch per the process described in https://github.com/OpenDevicePartnership/patina/blob/major/docs/src/rfc/text/0015-patina-release-process.md#branching.

Can you please rebase this PR onto that branch and change the PR branch target?

components/patina_adv_logger/src/memory_log.rs
thomashinds: These were made publically accessible as `patina_adv_logger::logger::DEBUG_LEVEL_ERROR` by the `pub use` in `logger.rs`, but I do agree that it would be cleaner to use the native Rust log level filter instead.

components/patina_adv_logger/src/logger.rs
thomashinds: Thanks, good suggestion, I agree. This could also help reduce confusion around all the bits in the bitmap that would be ignored by this implementation. I'll make the change.

components/patina_adv_logger/src/logger.rs
makubacki: Also, we're planning a Patina release with changes from the "major" branch tomorrow. If you can get your PR into the major branch, to follow the process, we can include it in that release if it's ready.

thomashinds: Thanks - rebased and retargeted.

components/patina_adv_logger/src/logger.rs
thomashinds: These are within the existing test mod starting at line 347
https://github.com/OpenDevicePartnership/patina/pull/1430/changes#diff-af18a6b5eba45f76a0df854ae0700de34d08bfaff823eef433c9dc70f8d62c70L305

cfernald: Oops, my mistake

components/patina_adv_logger/src/logger.rs
```diff
@@ -29,13 +29,27 @@ use spin::RwLock;
 #[used]
 static mut DBG_ADV_LOG_BUFFER: u64 = 0;
 
+/// Per-target filter that binds a target name prefix with its log level and optional hardware print level override.
+pub struct TargetFilter<'a> {
```
Javagedes, line 33: it may be worth having an "off" or "disabled" function, so that platforms can do this:

```rust
static LOGGER: AdvancedLogger<UartNull> = AdvancedLogger::new(
   Format::Standard, // How logs are formatted
   &[TargetFilter { target: "allocations", ..TargetFilter::off() }], // set custom log levels per module
   log::LevelFilter::Info, // Default log level
   UartNull { }, // Serial writer instance
);

// Example impl I'm referring to
impl TargetFilter {
  const fn off() -> Self {
    Self { target: "", log_level: LevelFilter::Off, hw_filter_override: None }
  }
}

```

thomashinds: Personally, I don't think this is needed - the existing arguments aren't too verbose, and I would expect platforms to declare any filters in one place and modify them rarely. 

We could always add helpers like this later in a non-breaking change.

I'm going to mark as resolved but feel free to reopen if you feel strongly that this must be added.

PR #1439, opened 2026-03-26
Update synchronization documentation.

joschock:
## Description

Cleans up and clarifies some elements of existing docs, adds a new section for rules of thumb for synchronization within Patina.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [x] Includes documentation?

## How This Was Tested

Passes md lint.

## Integration Instructions

N/A - documentation.

### DIFF
--- a/docs/src/dxe_core/synchronization.md
+++ b/docs/src/dxe_core/synchronization.md
@@ -40,7 +40,7 @@ scenarios.
 ```admonish warning
 Care must be taken when selecting the `tpl_lock_level` for a `TplMutex`. Code
 executing at a TPL higher than the `TplMutex` will panic if it attempts to
-accquire the lock (because it will attempt to raise the TPL to a lower level,
+acquire the lock (because it will attempt to raise the TPL to a lower level,
 which is an error). But setting a `tpl_lock_level` to a high TPL level will
 prevent other (unrelated) usage of that TPL, potentially reducing system
 responsiveness. It is recommended to set the `tpl_lock_level` as low as possible
@@ -78,7 +78,8 @@ mutex.
 
 The `try_lock()` routine in `TplMutex` allows a lock to be attempted and fail
 without blocking; this can be used for scenarios where a lock might be held by
-another agent in a lower TPL but the caller can handle not acquiring the lock.
+another agent in a lower TPL but the caller can handle not acquiring the lock,
+or in scenarios where a call is re-entrant at the same TPL.
 
 ## TplGuard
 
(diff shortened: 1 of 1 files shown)


docs/src/dxe_core/synchronization.md
```diff
@@ -109,3 +110,122 @@ assert!(tpl_mutex1.try_lock().is_err()); //mutex1 still locked.
...
+the data synchronization model of the code.
+
+1. Care should be taken to avoid UEFI spec caller restrictions on TPL as
+described in
+[Table 7.3](https://uefi.org/specs/UEFI/2.11/07_Services_Boot_Services.html#tpl-restrictions)
+of the UEFI spec. For example, the following usage of
+`TplMutex` would be an error:
+
+```rust,ignore
+let my_tpl_mutex = TplMutex<Data>::new(efi::TPL_NOTIFY, Data::new(), "my lock");
```
makubacki:
```suggestion
let my_tpl_mutex = TplMutex::<Data>::new(efi::TPL_NOTIFY, Data::new(), "my lock");
```

docs/src/dxe_core/synchronization.md
```diff
@@ -109,3 +110,122 @@ assert!(tpl_mutex1.try_lock().is_err()); //mutex1 still locked.
...
+`TplMutex` is not the same thing as the TPL level associated with an event
+callback routine. The TPL level associated with an event callback determines
+the TPL level at which the event callback is permitted to run and can be thought
+of as the "ground state" TPL that the event callback executes at. The callback
+is permitted to acquire a TplMutex at a higher level than the event callback is
+running at, and the TPL will be raised for the duration that the TPL guard is
+owned to prevent data races with event callbacks running at the higher context.
+```
+
+1. Care should be taken to ensure that `TplMutex` usages are scoped so that the
```
makubacki: It looks the numbering changed from increasing to all 1's after this point. Is that intentional?

docs/src/dxe_core/synchronization.md
makubacki: If you don't mind fixing this while editing the document.

```suggestion
acquire the lock (because it will attempt to raise the TPL to a lower level,
```

docs/src/dxe_core/synchronization.md
```diff
@@ -77,7 +78,7 @@ mutex.
 
 The `try_lock()` routine in `TplMutex` allows a lock to be attempted and fail
 without blocking; this can be used for scenarios where a lock might be held by
-another agent but the caller can handle not acquiring the lock.
+another agent in a lower TPL but the caller can handle not acquiring the lock.
```
makubacki: Isn't a reentrancy attempt another possibility in which the TPL might be the same?

docs/src/dxe_core/synchronization.md
```diff
@@ -109,3 +110,122 @@ assert!(tpl_mutex1.try_lock().is_err()); //mutex1 still locked.
...
+```admonish warning
+Note: it is important to understand that the TPL level associated with a
+`TplMutex` is not the same thing as the TPL level associated with an event
+callback routine. The TPL level associated with an event callback determines
+the TPL level at which the event callback is permitted to run and can be thought
+of as the "ground state" TPL that the event callback executes at. The callback
+is permitted to acquire a TplMutex at a higher level than the event callback is
+running at, and the TPL will be raised for the duration that the TPL guard is
+owned to prevent data races with event callbacks running at the higher context.
+```
```
makubacki: Sections like this (and the code examples below) would format more naturally alongside the ordered list paragraphs if indented a level.


```suggestion
  ```admonish warning
  Note: it is important to understand that the TPL level associated with a
  `TplMutex` is not the same thing as the TPL level associated with an event
  callback routine. The TPL level associated with an event callback determines
  the TPL level at which the event callback is permitted to run and can be thought
  of as the "ground state" TPL that the event callback executes at. The callback
  is permitted to acquire a TplMutex at a higher level than the event callback is
  running at, and the TPL will be raised for the duration that the TPL guard is
  owned to prevent data races with event callbacks running at the higher context.
  ```
```

docs/src/dxe_core/synchronization.md
```diff
@@ -109,3 +110,122 @@ assert!(tpl_mutex1.try_lock().is_err()); //mutex1 still locked.
...
+pointer. Sometimes the `context` pointer is known to be unique to the event
+callback and never accessed from other contexts, in which case a `TplMutex` is
+not required.
+
+1. Direct `raise_tpl` and `restore_tpl` calls should be avoided. Directly
+manipulating the TPL decouples the mutual exclusion primitives from the data
+that is being protected and makes it hard to associate the TPL requirements with
+the data synchronization model of the code.
+
+1. Care should be taken to avoid UEFI spec caller restrictions on TPL as
```
makubacki: Was this intended?

```suggestion
1. Care should be taken to avoid violating UEFI spec caller restrictions on TPL as
```

docs/src/dxe_core/synchronization.md
makubacki: It would be easier to follow as a cohesive numbered list.

docs/src/dxe_core/synchronization.md
```diff
@@ -109,3 +110,122 @@ assert!(tpl_mutex1.try_lock().is_err()); //mutex1 still locked.
...
+data between contexts. When implementing FFI interfaces to support these API
+contracts, `TplMutex` should be used to guard shared data accessed via these
+context raw pointers even though the raw pointers are not required to be `Sync`/
+`Send` by the compiler. Whether data races can occur and how they are prevented
+should be documented as part of the safety comments for usage of the raw context
+pointer. Sometimes the `context` pointer is known to be unique to the event
+callback and never accessed from other contexts, in which case a `TplMutex` is
+not required.
+
+1. Direct `raise_tpl` and `restore_tpl` calls should be avoided. Directly
```
cfernald: I wonder if these functions should just be made `unsafe`

joschock: What would the safety contract be that the caller has to uphold?

docs/src/dxe_core/synchronization.md
joschock: fixed.

docs/src/dxe_core/synchronization.md
joschock: fixed

docs/src/dxe_core/synchronization.md
joschock: updated the text to include that possibility.

docs/src/dxe_core/synchronization.md
joschock: fixed.

docs/src/dxe_core/synchronization.md
joschock: fixed

docs/src/dxe_core/synchronization.md
joschock: >What would the safety contract be that the caller has to uphold?

Proper TPL semantics? Those are enforced with asserts; do they merit `unsafe`?

cfernald: It's true the assert catches the most egregious of misuses, it doesn't (directly) catch scenarios where the caller doesn't call restore or restores with the incorrect value. Whether that warrants unsafe is a good question, probably not, since the atomic of the mutex should guard any actual invariants.

PR #1447, opened 2026-04-01
[feature/patina-boot] patina_boot: Add discover_boot_options helper

kat-perez:
## Description

Add `discover_boot_options()` helper to `patina_boot::helpers` that reads UEFI `BootOrder` and `Boot####` variables to build a `BootConfig` from standard UEFI boot options.

This enables any `BootOrchestrator` implementation that consumes `BootConfig` to use UEFI-compliant boot variables instead of requiring platforms to hardcode device paths. The function:

- Reads `BootOrder` to determine boot attempt order
- Parses each `Boot####` `EFI_LOAD_OPTION` structure to extract device paths
- Filters out inactive boot options (`LOAD_OPTION_ACTIVE`)
- Gracefully skips unreadable or malformed entries
- Returns a populated `BootConfig` with discovered devices in priority order
---

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- Unit tests covering: single/multiple boot options, inactive option filtering, unreadable variable handling, truncated load option data, empty BootOrder, and hex variable name generation
- Integration tested with patina-dxe-core-qemu `feature/patina-boot` on QEMU Q35 — full boot to UEFI Shell 2.0

## Integration Instructions

Platforms can call `discover_boot_options()` with runtime services to automatically populate a `BootConfig` from UEFI boot variables instead of constructing device paths manually. This works with any `BootOrchestrator` implementation that accepts a `BootConfig`:

```rust
let config = discover_boot_options(&runtime_services)?;
add.component(BootDispatcher::new(SimpleBootManager::new(config)));
```

### DIFF
--- a/components/patina_boot/src/helpers.rs
+++ b/components/patina_boot/src/helpers.rs
@@ -523,6 +523,163 @@ pub fn expand_device_path<B: BootServices>(boot_services: &B, partial_path: &Dev
     Err(EfiError::NotFound)
 }
 
+const LOAD_OPTION_ACTIVE: u32 = 0x00000001;
+
+use zerocopy::FromBytes;
+use zerocopy_derive::*;
+
+#[derive(FromBytes, KnownLayout, Immutable)]
+#[repr(C, packed)]
+struct LoadOptionHeader {
+    attributes: zerocopy::little_endian::U32,
+    file_path_list_length: zerocopy::little_endian::U16,
+}
+
+/// Discover boot options from UEFI `BootOrder` and `Boot####` variables.
+///
+/// Reads the `BootOrder` variable to determine boot attempt order, then reads
+/// each corresponding `Boot####` variable and parses the `EFI_LOAD_OPTION`
+/// structure to extract device paths. Only active boot options are returned.
+///
+/// Returns a [`BootConfig`](crate::config::BootConfig) populated with the discovered device paths, or
+/// an error if `BootOrder` cannot be read.
+pub fn discover_boot_options<R: RuntimeServices>(runtime_services: &R) -> Result<super::config::BootConfig> {
+    let namespace = EFI_GLOBAL_VARIABLE.into_inner();
+
+    let boot_order_name: Vec<u16> = "BootOrder\0".encode_utf16().collect();
+    let boot_order_name = boot_order_name.as_slice();
+
+    let (boot_order_bytes, _attributes): (Vec<u8>, u32) =
+        runtime_services.get_variable(boot_order_name, &namespace, None)?;
+
+    if boot_order_bytes.len() % 2 != 0 || boot_order_bytes.is_empty() {
+        log::error!("discover_boot_options: invalid BootOrder variable length");
+        return Err(EfiError::NotFound);
+    }
+
+    let boot_order: Vec<u16> =
+        boot_order_bytes.chunks_exact(2).map(|chunk| u16::from_le_bytes([chunk[0], chunk[1]])).collect();
+
+    let mut device_paths: Vec<DevicePathBuf> = Vec::new();
+
+    for option_number in &boot_order {
+        let var_name = boot_option_variable_name(*option_number);
+
+        let load_option_bytes = match runtime_services.get_variable::<Vec<u8>>(&var_name, &namespace, None) {
+            Ok((bytes, _)) => bytes,
+            Err(e) => {
+                log::warn!("discover_boot_options: failed to read Boot{:04X}: {:?}", option_number, e);
+                continue;
+            }
+        };
+
+        if let Some(device_path_buf) = parse_load_option(&load_option_bytes) {
+            device_paths.push(device_path_buf);
+        }
+    }
+
+    let mut iter = device_paths.into_iter();
+    let first = iter.next().ok_or(EfiError::NotFound)?;
+    let config = iter.fold(super::config::BootConfig::new(first), |config, dp| config.with_device(dp));
+
+    Ok(config)
+}
+
+/// Build a null-terminated UTF-16 variable name for `Boot####`.
+fn boot_option_variable_name(option_number: u16) -> Vec<u16> {
+    let mut name = alloc::format!("Boot{:04X}", option_number).encode_utf16().collect::<Vec<u16>>();
+    name.push(0);
+    name
+}
+
+/// Validate that all device path node lengths stay within the buffer.
+///
+/// Walks the node list checking that each node's Length field doesn't exceed the
+/// remaining buffer and that the path terminates with an EndEntire node.
+fn validate_device_path_nodes(buffer: &[u8]) -> bool {
+    const NODE_HEADER_SIZE: usize = 4;
+    const END_ENTIRE_TYPE: u8 = 0x7F;
+    const END_ENTIRE_SUBTYPE: u8 = 0xFF;
+
+    let mut pos = 0;
+    loop {
+        if pos + NODE_HEADER_SIZE > buffer.len() {
+            return false;
+        }
+
+        let node_type = buffer[pos];
+        let node_subtype = buffer[pos + 1];
+        let node_length = u16::from_le_bytes([buffer[pos + 2], buffer[pos + 3]]) as usize;
+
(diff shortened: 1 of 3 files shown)


components/patina_boot/src/helpers.rs
```diff
@@ -34,6 +36,13 @@ use r_efi::{
 /// Watchdog timeout in seconds per UEFI Specification Section 3.1.2.
 const WATCHDOG_TIMEOUT_SECONDS: usize = 300; // 5 minutes
 
+/// EFI_LOAD_OPTION attribute indicating an active boot option (UEFI Spec Section 3.1.3).
+const LOAD_OPTION_ACTIVE: u32 = 0x00000001;
+
+/// Null-terminated UTF-16 `BootOrder` variable name.
+const BOOT_ORDER_VARIABLE_NAME: &[u16] =
```
os-d: I missed this on earlier ones, but for these various UEFI spec defined things, we should be taking them to r-efi as well. Hosting them here until they are merged there is fine.

components/patina_boot/src/helpers.rs
```diff
@@ -523,6 +532,128 @@ pub fn expand_device_path<B: BootServices>(boot_services: &B, partial_path: &Dev
...
+        char::from_digit(((option_number >> 4) & 0xF) as u32, 16).unwrap_or('0'),
+        char::from_digit((option_number & 0xF) as u32, 16).unwrap_or('0'),
+    ];
+
+    let mut name: Vec<u16> = Vec::with_capacity(9);
+    for c in ['B', 'o', 'o', 't'] {
+        name.push(c as u16);
+    }
+    for c in hex_chars {
+        name.push(c.to_ascii_uppercase() as u16);
```
os-d: Aren't these all digits? What does to_ascii_uppercase do on a digit?

components/patina_boot/src/helpers.rs
```diff
@@ -523,6 +532,128 @@ pub fn expand_device_path<B: BootServices>(boot_services: &B, partial_path: &Dev
...
+///   u16    FilePathListLength
+///   [u16]  Description (null-terminated UTF-16)
+///   [u8]   FilePathList (device path, FilePathListLength bytes)
+///   [u8]   OptionalData (remaining bytes, ignored)
+fn parse_load_option(data: &[u8]) -> Option<DevicePathBuf> {
+    if data.len() < 6 {
+        return None;
+    }
+
+    let attributes = u32::from_le_bytes([data[0], data[1], data[2], data[3]]);
```
os-d: Lots of from_le_bytes throughout all these changes. A place to use zerocopy?

components/patina_boot/src/helpers.rs
kat-perez: `char::from_digit(n, 16)` returns lowercase hex and UEFI variable names use uppercase. The `to_ascii_uppercase` was there to fix the casing. I changed this to use `format!("Boot{:04X}", option_number)` to specify more clearly that we want uppercase hex.

components/patina_boot/src/helpers.rs
kat-perez: Updated to use zerocopy

os-d (approved):

components/patina_boot/src/helpers.rs
makubacki, line 189: It looks like this allocates a new vector for each handle which could be a lot of allocations, right? Do you know how many allocations (calls) are being made on QEMU?

components/patina_boot/src/helpers.rs
```diff
@@ -0,0 +1,1528 @@
...
+/// structure to extract device paths. Only active boot options are returned.
+///
+/// Returns a [`BootConfig`](crate::config::BootConfig) populated with the discovered device paths, or
+/// an error if `BootOrder` cannot be read.
+pub fn discover_boot_options<R: RuntimeServices>(runtime_services: &R) -> Result<super::config::BootConfig> {
+    let namespace = EFI_GLOBAL_VARIABLE.into_inner();
+
+    let boot_order_name: &[u16] = &[
+        'B' as u16, 'o' as u16, 'o' as u16, 't' as u16, 'O' as u16, 'r' as u16, 'd' as u16, 'e' as u16, 'r' as u16, 0,
+    ];
```
makubacki: You used `encode_utf16()` elsewhere in the file. Is there a reason not to use it here? Like:

```rust
let boot_order_name: Vec<u16> = "BootOrder\0".encode_utf16().collect();
```

components/patina_boot/src/helpers.rs
```diff
@@ -0,0 +1,1528 @@
...
+    }
+
+    let file_path_end = offset + file_path_list_length;
+    if file_path_end > rest.len() {
+        return None;
+    }
+
+    let file_path_bytes = &rest[offset..file_path_end];
+
+    // SAFETY: file_path_bytes points to a valid device path from the EFI_LOAD_OPTION.
```
makubacki: My understanding is that this depends on the boot variable (e.g. `Boot####`) to have correctly sized device path nodes. For example, if a boot UEFI variable were written with an very large node length (possible since these are just non-authenticated UEFI variables), this call to `try_from_ptr()` will call `from_raw_parts()` which will use that length to read past the actual device path buffer.

Let me know if this is already checked somewhere, but, if not, I suggest that each device path node length is validated before handing the pointer to `try_from_ptr()`.

components/patina_boot/src/orchestrators/simple_boot_manager.rs
```diff
@@ -0,0 +1,327 @@
...
+            log::error!("discover_console_devices failed: {:?}", e);
+        }
+
+        // Check for hotkey press after devices are connected and consoles discovered
+        let use_hotkey_devices =
+            if let Some(hotkey) = self.config.hotkey() { helpers::detect_hotkey(boot_services, hotkey) } else { false };
+
+        if let Err(e) = helpers::signal_ready_to_boot(boot_services) {
+            log::error!("signal_ready_to_boot failed: {:?}", e);
+        }
```
makubacki: If I'm reading this right, it is just calling ready to boot once and then looping through boot options? If so, it is expected to be called per boot option attempt.

> After all SysPrep#### variables have been launched and exited, the platform shall notify EFI_EVENT_GROUP_READY_TO_BOOT and EFI_EVENT_GROUP_AFTER_READY_TO_BOOT event groups. This should happen when the Boot Manager is about to load and execute Boot#### variables with Attributes set to LOAD_OPTION_CATEGORY_BOOT according to the order defined by BootOrder.

https://uefi.org/specs/UEFI/2.11/03_Boot_Manager.html#required-system-preparation-applications

components/patina_boot/src/helpers.rs
kat-perez: No specific reason that it needs to stay that way. Updated to use `encode_utf16()` instead

components/patina_boot/src/helpers.rs
kat-perez: According to Rust docs (https://doc.rust-lang.org/std/vec/struct.Vec.html#method.new), the vector won't allocate until elements are pushed onto it, so it seems fine to leave it as-is

components/patina_boot/src/orchestrators/simple_boot_manager.rs
kat-perez: Moved `signal_ready_to_boot` inside the boot device loop

components/patina_boot/src/helpers.rs
kat-perez: I didn't have any validation for the boot variable to have the correctly sized device nodes. But I added this: `validate_device_path_nodes()`. This checks that every node's Length is ≥ 4 (minimum header), doesn't exceed the remaining buffer, and that the path terminates with EndEntire.

components/patina_boot/src/helpers.rs
makubacki: That's a fair point. I'll resolve the thread.

makubacki (approved):

PR #1459, opened 2026-04-03
[feature/patina-boot] patina_boot: Add ConnectController service trait

kat-perez:
## Description

Add a pluggable `ConnectController` trait for controller connection strategies during device enumeration, replacing the hardcoded `connect_all()` call in `SimpleBootManager`.

- `ConnectController` trait with single `connect()` method
- `ConnectAllStrategy` default implementation (preserves existing behavior)
- `SimpleBootManager::with_connect_strategy()` constructor for custom strategies
- `interleave_connect_and_dispatch()` made generic over the connection function

This enables platforms to implement selective device enumeration (e.g., PCI-only, skip USB for headless) without writing a full custom `BootOrchestrator`.

`SimpleBootManager::new()` behavior is unchanged — zero breakage for existing platforms.

Ref: OpenDevicePartnership/modern-payload#205

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- 55 unit tests passing (7 new: custom connect function success/failure, connect strategy constructors, interleave with custom closures)
- `cargo fmt`, `cargo clippy` clean
- Integration tested with patina-dxe-core-qemu on QEMU Q35 — full boot to UEFI Shell 2.0

## Integration Instructions

No changes required for existing platforms. `SimpleBootManager::new()` defaults to `ConnectAllStrategy`.

Platforms wanting selective enumeration use the new constructor:

```rust
add.component(BootDispatcher::new(SimpleBootManager::with_connect_strategy(
    config,
    MyCustomStrategy,
)));
```

### DIFF
--- a/components/patina_boot/src/orchestrators/simple_boot_manager.rs
+++ b/components/patina_boot/src/orchestrators/simple_boot_manager.rs
@@ -15,46 +15,62 @@
 //!
 extern crate alloc;
 
-use alloc::vec::Vec;
+use alloc::{boxed::Box, vec::Vec};
 
 use patina::{
-    boot_services::StandardBootServices, device_path::paths::DevicePathBuf, error::EfiError,
+    boot_services::{BootServices, StandardBootServices, protocol_handler::HandleSearchType},
+    device_path::paths::DevicePathBuf,
+    error::EfiError,
     runtime_services::StandardRuntimeServices,
 };
 use r_efi::efi;
 
 use patina::component::service::dxe_dispatch::DxeDispatch;
 
-use patina::boot_services::BootServices;
-
-use crate::{boot_orchestrator::BootOrchestrator, config::BootConfig, helpers};
+use crate::{
+    boot_orchestrator::BootOrchestrator, config::BootConfig, connect_controller::ConnectController, helpers,
+    strategies::ConnectAllStrategy,
+};
 
 /// Interleave controller connection with DXE driver dispatch.
 ///
-/// Alternates between connecting all controllers and dispatching newly loaded
-/// drivers (e.g., PCI option ROMs) until both the device topology stabilizes
-/// and no new drivers are dispatched.
+/// Alternates between the given connect function and dispatching newly loaded
+/// drivers (e.g., PCI option ROM drivers) until the total system handle count
+/// stops growing across a connect+dispatch round.
 ///
 /// This ensures that drivers loaded from firmware volumes during device
 /// enumeration (such as PCI option ROM drivers) are dispatched before
 /// continuing enumeration, allowing those drivers to bind to newly
 /// discovered controllers.
 fn interleave_connect_and_dispatch<B: BootServices, D: DxeDispatch + ?Sized>(
+    connect_fn: impl Fn(&B) -> patina::error::Result<()>,
     boot_services: &B,
     dxe_services: &D,
 ) -> patina::error::Result<()> {
-    const MAX_ROUNDS: usize = 10;
+    let mut prev_handle_count = total_handle_count(boot_services)?;
+    #[cfg(debug_assertions)]
+    let mut rounds = 0usize;
 
-    for _round in 0..MAX_ROUNDS {
-        helpers::connect_all(boot_services)?;
-        if !dxe_services.dispatch()? {
+    loop {
+        connect_fn(boot_services)?;
+        dxe_services.dispatch()?;
+
+        let curr_handle_count = total_handle_count(boot_services)?;
+        if curr_handle_count == prev_handle_count {
             return Ok(());
         }
+        prev_handle_count = curr_handle_count;
+
+        #[cfg(debug_assertions)]
+        {
+            rounds += 1;
+            debug_assert!(rounds < 10_000, "connect-dispatch interleaving did not converge");
+        }
     }
+}
 
-    debug_assert!(false, "connect-dispatch interleaving did not converge after {MAX_ROUNDS} rounds");
-
-    Ok(())
+fn total_handle_count<B: BootServices>(boot_services: &B) -> patina::error::Result<usize> {
+    boot_services.locate_handle_buffer(HandleSearchType::AllHandle).map(|h| h.len()).map_err(EfiError::from)
 }
 
 /// Simple boot manager implementing [`BootOrchestrator`].
@@ -73,10 +89,15 @@ fn interleave_connect_and_dispatch<B: BootServices, D: DxeDispatch + ?Sized>(
 /// 7. Call failure handler if all options exhausted
 pub struct SimpleBootManager {
     config: BootConfig,
+    connect_strategy: Box<dyn ConnectController>,
 }
 
 impl SimpleBootManager {
     /// Create a `SimpleBootManager` from a boot configuration.
+    ///
+    /// Uses [`ConnectAllStrategy`] by default. To customize which controllers
+    /// are connected during device enumeration, use
+    /// [`with_connect_strategy()`](Self::with_connect_strategy).
     ///
     /// ## Example
     ///
@@ -92,7 +113,29 @@ impl SimpleBootManager {
     /// );
     /// ```
     pub fn new(config: BootConfig) -> Self {
-        Self { config }
+        Self { config, connect_strategy: Box::new(ConnectAllStrategy) }
+    }
+
+    /// Create a `SimpleBootManager` with a custom connection strategy.
+    ///
+    /// ## Example
+    ///
+    /// ```rust,ignore
+    /// use patina_boot::{ConnectController, SimpleBootManager, config::BootConfig};
+    /// use patina::{boot_services::BootServices, error::Result};
+    ///
+    /// struct MyPlatformConnect;
+    /// impl<B: BootServices> ConnectController<B> for MyPlatformConnect {
+    ///     fn connect(&self, bs: &B) -> Result<()> { /* sequence-specific connect */ Ok(()) }
+    /// }
+    ///
+    /// let manager = SimpleBootManager::with_connect_strategy(
+    ///     BootConfig::new(nvme_esp_path()),
+    ///     MyPlatformConnect,
+    /// );
+    /// ```
+    pub fn with_connect_strategy(config: BootConfig, strategy: impl ConnectController) -> Self {
+        Self { config, connect_strategy: Box::new(strategy) }
     }
 }
 
@@ -113,7 +156,9 @@ impl BootOrchestrator for SimpleBootManager {
         dxe_dispatch: &dyn DxeDispatch,
         image_handle: efi::Handle,
     ) -> Result<!, EfiError> {
-        if let Err(e) = interleave_connect_and_dispatch(boot_services, dxe_dispatch) {
+        if let Err(e) =
+            interleave_connect_and_dispatch(|bs| self.connect_strategy.connect(bs), boot_services, dxe_dispatch)
+        {
             log::error!("interleave_connect_and_dispatch failed: {:?}", e);
         }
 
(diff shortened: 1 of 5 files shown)


components/patina_boot/src/orchestrators/simple_boot_manager.rs
```diff
@@ -15,46 +15,50 @@
...
     for _round in 0..MAX_ROUNDS {
-        helpers::connect_all(boot_services)?;
+        connect_fn(boot_services)?;
         if !dxe_services.dispatch()? {
             return Ok(());
         }
     }
 
-    debug_assert!(false, "connect-dispatch interleaving did not converge after {MAX_ROUNDS} rounds");
+    log::error!("connect-dispatch interleaving did not converge after {MAX_ROUNDS} rounds");
```
os-d: Should you still leave the debug_assert to catch this easier?

components/patina_boot/src/strategies/connect_all.rs
```diff
@@ -0,0 +1,22 @@
+//! Default connection strategy: connect all controllers.
```
os-d: I wouldn't call this the default strategy generically. It is for the simple boot manager, but for a more advanced boot manager, it should not be

components/patina_boot/src/strategies/connect_by_protocol.rs
```diff
@@ -0,0 +1,191 @@
...
+}
+
+impl ConnectByProtocolStrategy {
+    /// Creates a new strategy that connects controllers matching the given protocol GUIDs.
+    pub fn new(protocol_guids: &[&'static efi::Guid]) -> Self {
+        Self { protocol_guids: protocol_guids.into() }
+    }
+}
+
+fn connect_by_protocol<B: BootServices>(boot_services: &B, protocol_guids: &[&'static efi::Guid]) -> Result<()> {
```
os-d: This seems a reasonable approach, for some platform it knows it only wants to connect NVMe devices, for example, but would appreciate others more familiar with BDS than me to weigh. @makubacki @apop5 @kuqin12 ?

os-d (commented):
I'm not as clear on what defines the Simple Boot Manager any more if the connect strategy is configurable. Besides connectall, what is the Simple Boot Manager doing that an "Advanced" Boot Manager wouldn't do or conversely, what isn't it doing?

i.e. it seems you are making Simple Boot Manager a more generic component than originally designed, which is fine, but perhaps the design point should be revisited on whether this is really just Simple or whether it is something that may extend to more platforms.

components/patina_boot/src/orchestrators/simple_boot_manager.rs
kat-perez: Yeah, I changed it back to `debug_assert`

components/patina_boot/src/strategies/connect_all.rs
kat-perez: Updated the doc so that it's clear that we're not implying that it's the universal default connection strategy

kat-perez:
@os-d 
> I'm not as clear on what defines the Simple Boot Manager any more if the connect strategy is configurable. Besides connectall, what is the Simple Boot Manager doing that an "Advanced" Boot Manager wouldn't do or conversely, what isn't it doing?

I can describe a bit better the expectations for `SimpleBootManager` and this is something I would add in the documentation: 

>  A declarative boot manager for platforms with static boot topologies — as opposed to a full UEFI BDS with variable-driven BootOrder and menu UI. Each phase (enumeration, hotkey dispatch, boot selection, failure) is a pluggable strategy with sensible defaults.

os-d:
> > A declarative boot manager for platforms with static boot topologies — as opposed to a full UEFI BDS with variable-driven BootOrder and menu UI. Each phase (enumeration, hotkey dispatch, boot selection, failure) is a pluggable strategy with sensible defaults.

Thanks. As it evolves, particularly with this notion of the pluggable strategies, I'm wondering whether the SimpleBootManager is becoming just a BootManager and one set of strategies is the "Simple" set and another set may include variable driven boot order, boot menu, etc.

kat-perez:
> > > A declarative boot manager for platforms with static boot topologies — as opposed to a full UEFI BDS with variable-driven BootOrder and menu UI. Each phase (enumeration, hotkey dispatch, boot selection, failure) is a pluggable strategy with sensible defaults.
> 
> Thanks. As it evolves, particularly with this notion of the pluggable strategies, I'm wondering whether the SimpleBootManager is becoming just a BootManager and one set of strategies is the "Simple" set and another set may include variable driven boot order, boot menu, etc.

Yeah, that's true that it's not really "Simple" anymore. The "simple" term feels overloaded / unclear for what this implementation is achieving. How about something like `StaticBootManager` or do you have any other recommendations for naming?

os-d:
> > > > A declarative boot manager for platforms with static boot topologies — as opposed to a full UEFI BDS with variable-driven BootOrder and menu UI. Each phase (enumeration, hotkey dispatch, boot selection, failure) is a pluggable strategy with sensible defaults.
> > 
> > 
> > Thanks. As it evolves, particularly with this notion of the pluggable strategies, I'm wondering whether the SimpleBootManager is becoming just a BootManager and one set of strategies is the "Simple" set and another set may include variable driven boot order, boot menu, etc.
> 
> Yeah, that's true that it's not really "Simple" anymore. The "simple" term feels overloaded / unclear for what this implementation is achieving. How about something like `StaticBootManager` or do you have any other recommendations for naming?

I think we can revisit the naming later if needed, I'm mostly thinking on the architectural side of this. Perhaps in the next Patina meeting we can discuss this with the group, to help with my understanding. Like I said, as more of the SimpleBootManager becomes pluggable, I'm not as sure the decisions being made in the SimpleBootManager that aren't relevant for other platforms.

apop5:
I found some old notes regarding what is happening the C Bds. Please don't take this as something we need to exactly reproduce.

- BdsEntry Protocol
- 	Signal gMsStartOfBdsNotifyGuid
- 	RebootReason	
- 	Set System Table Firmware Vendor
- 	Set System table Firmware revision
- 	Update CRC32 of system table
- 	Validate ConIn/ConOut/ErrOut
- 	Validate Os Indications 
- 	Lock Variables
- 	Initialize Hardware Error Records
- 	Initialize TimeOut variable
- 	Initialize BootOptionSupport variable
- 	Cache BootNext variable
- 	Initialize Language Support
- 	Initialize PlatformRecovery
- 	Status Code for Dxe Connet sequence
- 	Connect Root Bridges
- 	Register Default Boot Options
- 	Get Preferred Consoles
- 	Update ConIn/ConOut/ErrOut variables with preferred consoles
- 	Start HotKey Services
- 	Execute Driver#### Options
- 	Connect Consoles 
- 	Display Boot Graphics
- 	Display Console Messages
- 	Tpm Physical Prompts
- 	Get Platform Connect Sequence
- 	Handle Capsule
- 	Connect Connect Sequence specified  by Get Platform Connect Sequence
- 	Call Dispatch again to connect any lingering drivers 
- 	Process Capsules
- 	Act on OsIndications
- 	Debug Output  EFI Load Options
- 	If PlatformRecovery execute SysPrep#### options
- 	Bds wait for L"TimeOut"
- 	Check for any keystrokes (hotKeys)
- 	Handle any hotkeys that were pressed
- 	Check for Priority Boot (surface altboot, etc)
- 	Delete BootNext, if it exists to prevent boot loops
- 	Handle BootNext
- 	Try all boot options (loop indefinitely or as long as a single boot option succeeded )
- 	Print Boot Failed message

components/patina_boot/src/orchestrators/simple_boot_manager.rs
```diff
@@ -15,46 +15,50 @@
...
 /// continuing enumeration, allowing those drivers to bind to newly
 /// discovered controllers.
 fn interleave_connect_and_dispatch<B: BootServices, D: DxeDispatch + ?Sized>(
+    connect_fn: impl Fn(&B) -> patina::error::Result<()>,
     boot_services: &B,
     dxe_services: &D,
 ) -> patina::error::Result<()> {
     const MAX_ROUNDS: usize = 10;
 
     for _round in 0..MAX_ROUNDS {
```
apop5: I would recommend against the MAX_ROUNDS.

saying that the system will go through N numbers of connects makes it easy to do too many connects or too few.

I think the logic could be changed to use the LocateHandleBuffer to gets the total number of handles in the system, and when the total number of handles not change for a connet/dispatch loop, then to use that as an exit condition.

components/patina_boot/src/strategies/connect_by_protocol.rs
apop5: Some platform out there is going to want a specific connection sequence.

As long as we satisfy the RootBridges are connected prior to this, then it should be possible for a platform to provide their own implementation for whatever they want to connect, in whatever order they want.

apop5: I revise my earlier statement.

When Connect all is being called, it just has the buffer of handles that support DriverBinding.
Then it has a buffer of all handles. It was a double for loop that passes each handle to the driver binding supported function.

For the Nvme example, connecting by a protocol guid would not really do anything. At the point in the boot process, there would be no "nvme" protocols connected, it would just be handles for the root bridges (if they were already connected).
To get to Nvme

- PciBus would need to connect. (uses root bridge Io handles)
- NvmeController would need to connet. (uses PciIo handles produced by PciBus)
etc etc

This would need to be more complex to allow only connecting nvme.

kat-perez: Sure, that makes sense. I can drop ConnectByProtocolStrategy and ship just the ConnectController trait + ConnectAllStrategy so that platforms can just provide their own implementation

components/patina_boot/src/orchestrators/simple_boot_manager.rs
```diff
@@ -15,46 +15,62 @@
...
 /// continuing enumeration, allowing those drivers to bind to newly
 /// discovered controllers.
 fn interleave_connect_and_dispatch<B: BootServices, D: DxeDispatch + ?Sized>(
+    connect_fn: impl Fn(&B) -> patina::error::Result<()>,
     boot_services: &B,
     dxe_services: &D,
 ) -> patina::error::Result<()> {
-    const MAX_ROUNDS: usize = 10;
+    let mut prev_handle_count = total_handle_count(boot_services)?;
+    #[cfg(debug_assertions)]
```
makubacki: How do you envision this being used? We're trying to limit `cfg` logic in files to reduce conditional complexity.

components/patina_boot/src/orchestrators/simple_boot_manager.rs
```diff
@@ -15,46 +15,62 @@
...
             return Ok(());
         }
-    }
+        prev_handle_count = curr_handle_count;
 
-    debug_assert!(false, "connect-dispatch interleaving did not converge after {MAX_ROUNDS} rounds");
+        #[cfg(debug_assertions)]
+        {
+            rounds += 1;
+            debug_assert!(rounds < 10_000, "connect-dispatch interleaving did not converge");
```
makubacki: 10,000 seems like a lot of rounds. What are you typically seeing? How was that chosen?

components/patina_boot/src/orchestrators/simple_boot_manager.rs
```diff
@@ -15,46 +15,62 @@
...
 ) -> patina::error::Result<()> {
-    const MAX_ROUNDS: usize = 10;
+    let mut prev_handle_count = total_handle_count(boot_services)?;
+    #[cfg(debug_assertions)]
+    let mut rounds = 0usize;
 
-    for _round in 0..MAX_ROUNDS {
-        helpers::connect_all(boot_services)?;
-        if !dxe_services.dispatch()? {
+    loop {
```
makubacki, line 52: It looks like this loop no longer has a max on release builds. Is that intentional?

components/patina_boot/src/orchestrators/simple_boot_manager.rs
```diff
@@ -73,11 +89,16 @@ fn interleave_connect_and_dispatch<B: BootServices, D: DxeDispatch + ?Sized>(
 /// 7. Call failure handler if all options exhausted
 pub struct SimpleBootManager {
     config: BootConfig,
+    connect_strategy: Box<dyn ConnectController>,
 }
```
makubacki: Does `connect_strategy` need to be `Box<dyn ConnectController>`?

Versus (with a default of `ConnectAllStrategy`):

```rust
pub struct SimpleBootManager<C: ConnectController = ConnectAllStrategy> {
    config: BootConfig,
    connect_strategy: C,
}
```

components/patina_boot/src/strategies/connect_all.rs
```diff
@@ -0,0 +1,23 @@
...
+//! SPDX-License-Identifier: Apache-2.0
+//!
+
+use patina::{boot_services::BootServices, error::Result};
+
+use crate::{connect_controller::ConnectController, helpers};
+
+/// Connects all controllers recursively via
+/// [`helpers::connect_all()`](crate::helpers::connect_all).
+/// Used as the default strategy by [`SimpleBootManager::new()`](crate::SimpleBootManager::new).
```
makubacki: I suggest to avoid referencing the code that depends on this (`SimpleBootManager`) from here.

components/patina_boot/src/orchestrators/simple_boot_manager.rs
kat-perez: The cfg(debug_assertions) was scoping the rounds counter to debug builds, since the only consumer was the debug_assert!(rounds < 10_000). Though, in the latest revision the assert is removed in 79e6d4c.

components/patina_boot/src/orchestrators/simple_boot_manager.rs
kat-perez: Yeah, I was choosing a defensively high number, though I just checked in my local Q35 testing the loop converged in 1 round (215 handles). I'm rethinking this and I think it'd be safe to drop the MAX_ROUNDS at this point

components/patina_boot/src/orchestrators/simple_boot_manager.rs
kat-perez: Removed the MAX_ROUNDS in favor of simplicity

components/patina_boot/src/orchestrators/simple_boot_manager.rs
kat-perez: Doesn't need to be, I can do this instead. Updated it in my latest change

components/patina_boot/src/orchestrators/simple_boot_manager.rs
makubacki: I think it might be worth adding a break with some error message.

components/patina_boot/src/orchestrators/simple_boot_manager.rs
makubacki: Resolving the thread since it is removed now.

components/patina_boot/src/orchestrators/simple_boot_manager.rs
makubacki: I think it might still be useful to have an upper bound, but that's [covered in another thread](https://github.com/OpenDevicePartnership/patina/pull/1459/changes/BASE..fc5470b196b9fa4dc3104ef505e54999f30492d0#r3184623725), so I'll resolve this one.

makubacki (approved):
We'll probably talk more about connection strategies and SimpleBootManager in a larger design discussion. I'm approving the current state because I think this is reasonable for you to move forward with in the feature branch/to continue development.

apop5 (approved):

os-d (approved):

PR #1460, opened 2026-04-08
Add AI code review customization files [Rebase & FF]

makubacki:
## Description

Closes #1453 

Adds project-level AI customization files to support AI-assisted code review.

These files encode Patina's domain-specific conventions so that AI reviewers can flag violations locally, prior to human review.

The files are layered:

1. Always-on instructions (.github/copilot-instructions.md): Covers global conventions across the codebase.

2. File-based instructions (.github/instructions/): Currently, provides component-specific instructions in components.instructions.md(applyTo: components/**) to supplement the global instructions. These load only when editing matching files to minimize token cost.

3. Review prompt file (.github/prompts/review.prompt.md): A reusable /review slash command that defines a structured review checklist. This is to provide a convenient way to reliably perform the same set of review steps. Delegates convention details to the always-on instructions to avoid duplication.

4. Reviewer agent (.github/agents/reviewer.agent.md): A read-only review persona restricted to `search` and `read` tool sets (no edit or execute tools). Includes a handoff to the default agent for implementing fixes after review.

References:

- VS Code Customization Overview: https://code.visualstudio.com/docs/copilot/concepts/customization
- Custom Instructions: https://code.visualstudio.com/docs/copilot/customization/custom-instructions
- Prompt Files: https://code.visualstudio.com/docs/copilot/customization/prompt-files
- Custom Agents: https://code.visualstudio.com/docs/copilot/customization/custom-agents

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- Use the "Patina Code Reviewer" locally

## Integration Instructions

- N/A

---

This PR is in draft as it serves as a starting open to changes from others before publishing for broader review.

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -1,0 +1,194 @@
+# Patina Project Instructions
+
+Patina is a Rust-based UEFI firmware project. It currently replaces the traditional C-based EDK II
+DXE Core with a Rust implementation that introduces a component-based architecture with dependency
+injection in addition to a general-purpose UEFI Rust SDK.
+
+## Project Structure
+
+The workspace is organized into four top-level areas:
+
+- `components/` - Feature components (ACPI, Advanced Logger, MM, Performance, SMBIOS, etc.)
+- `core/` - Shared core internals (debugger, collections, CPU, dependency expressions, stack traces)
+- `patina_dxe_core/` - Main DXE Core library that ties everything together
+- `sdk/` - Public SDK (`patina` crate for Boot/Runtime Services, component infrastructure, `patina_ffs` for firmware
+   file system, `patina_macro` for proc-macros)
+
+Crate dependency rules are strict. Components depend on `sdk/` only - never on each other or on `core/`. The SDK
+depends only on generic external crates. Core crates may depend on each other and on the SDK. See
+[code_organization.md](docs/src/dev/code_organization.md) for the full dependency matrix.
+
+## Build and Test Commands
+
+All commands go through `cargo make`. Never run raw `cargo` commands.
+
+- `cargo make check` - Type-check all code
+- `cargo make fmt` - Format code (run after every edit)
+- `cargo make clippy` - Lint with clippy
+- `cargo make test` - Run all unit tests
+- `cargo make all` - Full PR readiness (fmt-check, deny, cspell, clippy, build, test, coverage, doc)
+- `cargo make build-x64` / `cargo make build-aarch64` - Build for UEFI targets
+- `cargo make coverage` - Generate test coverage reports
+- `cargo make doc` - Build documentation
+
+See [Makefile.toml](Makefile.toml) for the complete command list.
+
+## Module Organization
+
+- Never use `mod.rs` files. Use named module files (e.g., `my_module.rs`) instead.
+- No public definitions directly in `lib.rs` - only public module declarations.
+- Break implementations into logical files with clean namespaces.
+- Crate naming: `patina_` prefix for public crates, `patina_internal_` for internal
+  crates, `_macro` suffix for proc-macro crates.
+
+```rust
+// WRONG: using mod.rs
+// src/memory/mod.rs
+
+// CORRECT: using named module file
+// src/memory.rs
+```
+
+See [component requirements](docs/src/component/requirements.md) for crate layout standards.
+
+## Safety Conventions
+
+- Prefer `zerocopy` for binary data and memory layouts over manual unsafe pointer/slice
+  operations. `zerocopy` provides safe, zero-cost abstractions for interpreting byte
+  buffers as typed data.
+- Minimize `unsafe` code. Constrain it within safe abstraction wrappers.
+- Document preconditions, postconditions, and invariants for every `unsafe` block.
+- Prefer fallible constructors returning `Result` over constructors that silently
+  handle invalid input or panic.
+- Mark functions `unsafe` only when the caller must uphold a contract the function
+  itself cannot verify internally. If the function validates all inputs before the
+  unsafe operation, it can remain safe.
+
+See [unsafe.md](docs/src/dev/principles/unsafe.md) for the full safety philosophy
+(software safety vs. hardware safety distinctions).
+
+## Error Handling
+
+- Prefer `Result` return types. Avoid panics in production code.
+- At UEFI ABI boundaries (`extern "efiapi"` functions), use `efi::Status`.
+- Internally, use domain-specific Rust error types with `Debug`, `Display`, and `Error`
+  trait implementations.
+- Implement `From<>` conversions at error type boundaries to create clean error chains.
+- Use `expect("descriptive message")` over bare `unwrap()`. Reserve `unwrap()` for
+  test code only.
+
+See [error-handling.md](docs/src/dev/principles/error-handling.md) for error
+propagation patterns and examples.
+
+## Component Model
+
+Patina uses dependency injection through component entry point function signatures.
+A component only executes when all its declared dependencies are available.
+
+- Define components with the `#[component]` attribute macro on an `impl` block.
+  The impl must contain `fn entry_point(self, ...) -> Result<()>`.
+- Register service implementations with `#[derive(IntoService)]` and `#[service(dyn Trait)]`.
+- Param types: `Config<T>`, `ConfigMut<T>`, `Service<T>`, `Hob<T>`, `Commands`,
+  `Handle`, `StandardBootServices`, `StandardRuntimeServices`, `&Storage`/`&mut Storage`,
+  `Option<P>`, tuples.
+- `ConfigMut<T>` components run first (config is unlocked); calling `lock()` makes the
+  value immutable and enables `Config<T>` components to execute.
+- `Service<dyn Trait>` is preferred over `Service<ConcreteType>` for mockability. Use
+  a concrete wrapper struct only when generic method signatures are needed.
+- Use the **stored dependencies** pattern: store all dependency references as fields in
+  the component struct. The entry point stores references; methods use them.
+
+See [component interface](docs/src/component/interface.md) and
+[getting started](docs/src/component/getting_started.md) for details and examples.
+
+## Mocking and Testing
+
+- Target ≥80% code coverage. Patch coverage ≥80% is a required PR check.
+  - Write tests to cover critical logic and edge cases, not just to hit coverage numbers.
+- Test naming: prefix with `test_<component_name>_` or `test_<service_name>_`.
+- Prefer `mockall` for trait mocking in unit tests. Use `#[automock]` on traits to generate mocks.
+- Use `pretty_assertions` for readable test diffs.
+
+See [testing.md](docs/src/dev/testing.md) for the full testing strategy.
+
+## Trait Design
+
+Traits serve as **abstraction points** - not code reuse mechanisms. Use crates for
+code reuse (see [reuse.md](docs/src/dev/principles/reuse.md)).
+
+- Define traits to represent swappable behavior (e.g., `BootServices`, `SerialIO`).
+- Implementations can vary per platform without affecting consumers.
+- Keep traits focused on a single responsibility (Interface Segregation).
+
+See [abstractions.md](docs/src/dev/principles/abstractions.md) for the full trait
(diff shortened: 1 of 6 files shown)


.github/instructions/components.instructions.md
```diff
@@ -0,0 +1,91 @@
...
+Components use dependency injection through their entry point function signature.
+The entry point declares what the component needs and the dispatcher provides it.
+
+- Derive `IntoComponent` on structs that serve as components.
+- The entry point method must return `patina_sdk::error::Result<()>`.
+- Parameters in the entry point define dependencies - the component will not execute
+  until all dependencies are satisfied.
+
+```rust
+use patina_sdk::{
```
Javagedes: patina_sdk is outdated

.github/instructions/components.instructions.md
```diff
@@ -0,0 +1,91 @@
...
+    component::IntoComponent,
+    error::Result,
+};
+
+#[derive(IntoComponent)]
```
Javagedes: how we derive is outdated

.github/instructions/components.instructions.md
makubacki: Updated

.github/instructions/components.instructions.md
makubacki: Updated. Copied from old RFC.

.github/instructions/components.instructions.md
```diff
@@ -0,0 +1,111 @@
...
+- **Execution phase:** Use stored references via methods - no additional parameter
+  resolution needed.
+- Use `Service<T>` for service references (these are `Clone` with static lifetimes).
+- Use `Option<T>` for optional dependencies to handle unavailable services gracefully.
+
+## Parameter Semantics
+
+| Param | Available When | Notes |
+|---|---|---|
+| `Config<T>` | After config is locked | Immutable access to public configuration |
```
cfernald: For things like this, i wonder if it would be better to reference the component model documentation. I these instructions may get out of sync with the docs leading to a drop in quality.

.github/prompts/review.prompt.md
```diff
@@ -0,0 +1,63 @@
+---
+description: 'Review code changes against Patina project conventions and standards'
+tools:
+  - search
+  - read
```
vineelko, line 5: NIT: Should we also include the 'web' tool here so that the agent can go to the URLs(in the code comments) and may have more context?

.github/prompts/review.prompt.md
```diff
@@ -0,0 +1,63 @@
...
+
+### 2. Safety
+
+- Raw pointer/slice work where `zerocopy` would suffice.
+- `unsafe` blocks missing documented invariants or missing safe wrappers.
+- Functions marked `unsafe` that could validate inputs internally.
+
+### 3. Error Handling
+
+- `unwrap()` in non-test code.
```
os-d, line 30: I would suggest adding another line here checking for panics of any kind instead of graceful error handling. I think the reviewer agent should flag all cases of panicking without error handling, with a note that a very small set of cases, when execution truly cannot continue, is appropriate, but that each case should be flagged and reviewed by the developer to ensure this is one of those cases.

.github/prompts/review.prompt.md
```diff
@@ -0,0 +1,63 @@
...
+- Default methods on `#[automock]` traits instead of extension traits.
+
+### 6. Documentation
+
+- Public items missing doc comments.
+- `unsafe` functions missing `# Safety`; fallible functions missing `# Errors`.
+
+### 7. UEFI-Specific
+
+- Non-TPL-aware synchronization primitives.
```
os-d, line 51: This one should maybe point to the docs on synchronization to determine if the use case is appropriate

.github/copilot-instructions.md
```diff
@@ -0,0 +1,204 @@
+# Patina Project Instructions
+
+Patina is a Rust-based UEFI DXE Core firmware project. It replaces the traditional C-based EDK II
```
os-d: We're quickly growing beyond just DXE Core, do we maybe want to make this a bit broader or wait and update these instructions after the MM core is merged? I could easily see separate reviewer instructions for MM since the requirements are different.

.github/prompts/review.prompt.md
os-d: Scrolling down further, I see the avoid panics in production code in the copilot-instructions.md, so maybe that is good enough.

.github/copilot-instructions.md
```diff
@@ -0,0 +1,204 @@
...
+See [abstractions.md](docs/src/dev/principles/abstractions.md) for the full trait
+design philosophy.
+
+## Documentation Standards
+
+- All public items must be documented. Set `#[deny(missing_docs)]` in component crates.
+- Use `///` doc comments with these sections as appropriate: **Examples**, **Errors**,
+  **Safety**, **Panics**.
+- Document **traits**, not their implementations.
+- Do not add `# Arguments` or `# Returns` sections - the function signature is
```
os-d: Is this true? For some functions, sure, but for others, additional context is often needed for the arguments or returns. E.g. the name of an argument may be `MemoryType` and be of type `MemoryType` so it is clear what is accepted, but this function may only accept `MemoryType.X` and `MemoryType.Y` and that should be documented.

Overall thought being for some functions, I think #Arguments and #Returns are redundant, but I'm not sure about a blanket policy to never include them.

.github/prompts/review.prompt.md
os-d: Again, scrolling down I see this in copilot-instructions, so maybe not needed here.

os-d (approved):
Do you want to reference these in the mdbook? If so, I'd recommend stressing the optional nature of these, it is for local development and will not be an autoenforced reviewer, etc.

.github/copilot-instructions.md
```diff
@@ -0,0 +1,204 @@
...
+- `sdk/` - Public SDK (`patina` crate for Boot/Runtime Services, component infrastructure, `patina_ffs` for firmware
+   file system, `patina_macro` for proc-macros)
+
+Crate dependency rules are strict. Components depend on `sdk/` only - never on each other or on `core/`. The SDK
+depends only on generic external crates. Core crates may depend on each other and on the SDK. See
+[code_organization.md](docs/src/dev/code_organization.md) for the full dependency matrix.
+
+## Build and Test Commands
+
+All commands go through `cargo make`. Never run raw `cargo` commands.
```
cfernald, line 23: I actually find running `cargo test` while in specific crates very useful and time saving, should we consider loosening this or adding exceptions to that "never" rule?

.github/copilot-instructions.md
```diff
@@ -0,0 +1,204 @@
...
+- `cargo make build-x64` / `cargo make build-aarch64` - Build for UEFI targets
+- `cargo make coverage` - Generate test coverage reports
+- `cargo make doc` - Build documentation
+
+See [Makefile.toml](Makefile.toml) for the complete command list.
+
+## Module Organization
+
+- Never use `mod.rs` files. Use named module files (e.g., `my_module.rs`) instead.
+- No public definitions directly in `lib.rs` - only public module declarations.
```
cfernald, line 39: I don't think we generally follow this rule. Not saying we shouldnt strive to, but maybe we need to clarify this with the team as well

.github/copilot-instructions.md
```diff
@@ -0,0 +1,204 @@
...
+- Mark functions `unsafe` only when the caller must uphold a contract the function
+  itself cannot verify internally. If the function validates all inputs before the
+  unsafe operation, it can remain safe.
+
+```rust
+// WRONG: exposing raw unsafe to callers
+pub unsafe fn read_register(addr: *const u32) -> u32 { ... }
+
+// CORRECT: safe wrapper that validates internally
+pub fn read_register(addr: &MmioRegister) -> u32 {
```
cfernald: I think this needs to be updated as we have mentioned elsewhere MMIO shouldn't have references

.github/prompts/review.prompt.md
makubacki: I was thinking that it can find the documentation needed as general knowledge and local docs to keep the toolset scope limited. I'm also not sure how comfortable the general developer population is with AI accessing the Web, potential token impact etc.

If we think there is a substantial benefit and the concerns are not a problem, I'm happy to add it.

.github/prompts/review.prompt.md
makubacki: In my testing, I actually added a panic and it handled finding it well. Let me know if you want anything in addition to what is in copilot-instructions.md.

.github/copilot-instructions.md
makubacki: It makes sense to go ahead broaden this now. No need to restrict the overall project scope to the DXE core as opposed to saying it is a part of the project (+ the SDK, etc.). Then MM core and add its info in the future.

.github/copilot-instructions.md
makubacki: This is repeating the [function style guide](https://opendevicepartnership.github.io/patina/dev/documenting.html#function-style-guide) doc info.

I'm with you though and we've naturally added some extra context in a few places. If we loosen this up (which I'm fine with), I'd also want to update the linked doc as well.

Does anyone have concerns with that?

.github/prompts/review.prompt.md
vineelko: To be honest, I am not entirely sure about it either. One use case where it might be helpful is that we often reference UEFI spec HTML URLs in code comments. Apparently, Copilot can access those web links and get additional context if the web tool is enabled, otherwise, it relies on its general training knowledge about the topic. I am also not very clear about the guardrails for enabling the web tool whether it only looks at the URLs mentioned in the code or if it can access the general web. Feel free to resolve this comment.

vineelko (approved):

.github/instructions/components.instructions.md
makubacki: There is a balance between providing the most relevant and condensed data and other documentation. I can merge some of this into existing docs (like https://opendevicepartnership.github.io/patina/component/interface.html) if we'd like, but I'd want to keep references as direct as possible to the relevant info there.

I also expect that these files will become a regular part of our documentation maintenance, but I understand what you're saying.

.github/prompts/review.prompt.md
makubacki: I guess I'll go with the more conservative approach for now and keep it as is. We can always add it later if it's found to substantially help.

.github/prompts/review.prompt.md
makubacki: I'll keep it as-is and mark the thread resolved.

.github/copilot-instructions.md
makubacki: Updated it to describe Patina as a Rust-based UEFI project.

.github/copilot-instructions.md
makubacki: Updated it to allow args and returns sections in special cases.

.github/instructions/components.instructions.md
makubacki: Merged it and created a reference to the doc.

.github/copilot-instructions.md
makubacki: `cargo make test` [just calls `cargo test`](https://github.com/OpenDevicePartnership/patina/commit/acb5c1bf0dfb4ab55b40df5cdb9c9e1e50f3ea84) now and supports the same parameters. Does that cover your case?

.github/copilot-instructions.md
makubacki: This is already a rule in the [Crate Layout Standards](https://opendevicepartnership.github.io/patina/component/requirements.html#crate-layout-standards). I'll bring it up for visibility in the next Patina dev meeting.

.github/copilot-instructions.md
makubacki: Good catch. I removed that example.

.github/copilot-instructions.md
cfernald: yeah, i'm just lazy. This works

cfernald (approved):

PR #1463, opened 2026-04-13
patina: Use correct FPDT record types for MODULE_DB_END_ID

cfernald:
## Description

The patina implementation uses QWORD String event type for MODULE_DB_STOP_END_ID instead of MODULE_DB_END_ID, which is causing asserts when parsing the FDPT as this in unexpected. This commit fixes this to be consistent with the EDKII implementation.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on Q35 by parsing from shell with `dp`

## Integration Instructions

N/A

### DIFF
--- a/sdk/patina/src/performance/measurement.rs
+++ b/sdk/patina/src/performance/measurement.rs
@@ -508,6 +508,7 @@ mod tests {
         performance::{
             globals::set_perf_measurement_mask,
             logging::*,
+            record::PerformanceRecord,
             table::{FirmwarePerformanceVariable, MockFirmwareBasicBootPerfTable},
         },
         runtime_services::MockRuntimeServices,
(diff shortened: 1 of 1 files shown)


sdk/patina/src/performance/measurement.rs
makubacki, line 1: @cfernald, is it possible to verify this with a unit test? As we know, most platforms are not regularly building and testing `dp`.

Javagedes (approved):

sdk/patina/src/performance/measurement.rs
cfernald: I added a test that checks all the types compared to the parser. Is this what you had in mind or did you mean for full perf build and run of the C parser?

makubacki: That's all I had in mind here. We could consider that in patina-qemu but I think it's outside the scope of this PR.

makubacki (approved):

PR #1465, opened 2026-04-13
patina_performance: Simplify configuration

Javagedes:
## Description

patina_performance has a confusing configuration story. Prior to this change, configuration was done via patina `Config`, A platform could optionally register a secondary component that would read a HOB (if provided) and overwrite any `Config` with the HOB configuration.

This commit works to simplify the configuration story by making a couple of changes:

1. Removal of the `Config` object. As stated in documentation throughout patina, `Config` is meant for configuration that is expected to be shared between multiple components. This particular configuration does not need to be shared, so it is moved to be private configuration of the component.
2. Removal of the optional HOB reader driver. Now, a private configuration of the component will allow you to specify if the HOB is able to override local configuration or not.

This new story reduces patina_performance setup from three different definitions (Config, performance component, and hob component) to only one, the performance component.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [x] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Reading final configuration of the performance component on Q35 with various configuration settings both in the component and via the HOB

### Previous

```rust
use patina_dxe_core::*;

struct Q35;

impl ComponentInfo for Q35 {
  fn configs(mut add: Add<Config>) {
    add.config(patina_performance::config::PerfConfig {
      enable_component: true,
      enabled_measurements: patina::performance::Measurement::LoadImage | patina::performance::Measurement::StartImage
    }
  }
  
  fn components(mut add: Add<Component>) {
    add.component(patina_performance::component::performance_config_provider::PerformanceConfigurationProvider);
    add.component(patina_performance::component::performance::Performance);
  }
}
```

### After

```rust
use patina_dxe_core::*;
use patina_performance::component::*;

struct Q35;

impl ComponentInfo for Q35 {
  fn components(mut add: Add<Component>) {
    // Records LoadImage and StartImage measurements unless configuration is overwritten by a HOB
    add.component(Performance::new().with_measurements(Measurement::LoadImage | Measurement::StartImage));
    
    // Disabled. Enablement fully controlled by a HOB
    add.component(Performance::new());
}
```


components/patina_performance/src/component/performance.rs
```diff
@@ -38,30 +36,108 @@ use patina::{
...
 
+use patina::component::hob::FromHob;
+
+/// The configuration for the Patina Performance component.
+#[derive(Debug, Clone, Copy, FromHob, zerocopy_derive::FromBytes)]
+#[hob = "fd87f2d8-112d-4640-9c00-d37d2a1fb75d"]
+#[repr(C, packed)]
+struct PerformanceConfig {
+    /// Indicates whether the Patina Performance component is enabled.
+    pub enable_component: u8,
```
cfernald, line 54: Why not bool?

cfernald (approved):

components/patina_performance/src/component/performance.rs
Javagedes: bool is not a supported `zerocopy` type

components/patina_performance/src/component/performance.rs
```diff
@@ -38,30 +36,108 @@ use patina::{
...
+/// // Performance measurements are disabled by default, but can be overridden by a performance config HOB.
+/// let disabled_component = Performance::new();
+///
+/// // Performance measurements are enabled with no ability to override via a HOB.
+/// let enabled_no_hob_component = Performance::new().measure(Measurement::DriverBindingStart | Measurement::LoadImage).ignore_hob();
+/// ```
+#[derive(Default)]
+pub struct Performance {
+    config: PerformanceConfig,
+    ignore_hob: bool,
```
makubacki: HOBs are already dynamically produced by platform logic, for example, in this case whether or not the platform would want to configure performance. The `performance_config_provider` was more to keep this component free of HOB logic which it now has anyway.

makubacki: Can `ignore_hob` just be dropped entirely and tell the platform to produce the HOB to override config if they want?

components/patina_performance/src/component/performance.rs
cfernald: I think it is for `TryFromBytes` that will validate the only 0 or 1 semantic you assume below.

Image

Javagedes: Unfortunately, the current design of the `FromHob` trait is that parsing must be infallible. I actually think that it would make more sense if it was fallible. But at this point in time the `FromHob` macro expects `zerocopy::FromBytes` to be implemented. We could do a manual implementation, but we would have to unwrap the result.

components/patina_performance/src/component/performance.rs
Javagedes: yes, we could absolutely entirely drop `ignore_hob` if we wanted, and I'm happy to do that. It would technically result in a regression in configurability, however.

Right now, you it is optional to add the `PerformanceConfigurationProvider`, which means a platform has the possibility of ignoring the HOB even if it exists. Doing the above suggestion removes that possibility. I'm not against it. But it is a thing.

makubacki: Yeah, I think we have to take advantage of the limited set of consumers and just "simplify" it further by dropping HOB acknowledgement. Also, I'd prefer to avoid setting a precedent where there are multiple layers of configuration points for influencing the value which aligns with the spirit of the PR.

joschock (approved):

components/patina_performance/src/component/performance.rs
Javagedes: In addition to the offline talk where I confirmed we are going to switch to TryFromBytes, I'm going to make this a separate PR.

components/patina_performance/src/component/performance.rs
Javagedes: Removed / simplified

makubacki (approved):

PR #1468, opened 2026-04-15
Updated EFI_CPU_ARCH_PROTOCOL to provide DmaBufferAlignment

### LINKED ISSUE #1466: [Bug]: Arm EfiCpuArchProtocol::DmaBufferAlignment set to 0

### Current Behavior

The code file patina_dxe_core\src\cpu\cpu_arch_protocol.rs is publishing an EFI_CPU_ARCH_PROTOCOL that has a field called DmaBufferAlignment to provide the size, in bytes, of the alignment required for DMA buffer allocations.  It should be a non-zero number that is the largest cache line size.

### Expected Behavior

This value should be the largest cache line size of any cache in the system per the UEFI PI spec, version 1.9, section II-12.3.1.

### Steps To Reproduce

A platform using the CIX chipset and an NVME drive will not detect the NVME device at boot.

### Build Environment

```markdown
- OS(s): N/A

- Targets Impacted: AARCH64 platforms if they don't support cache-coherent DMA.  X86 CPUs have hardware cache-coherent DMA through PCIe root complex snooping.
```

### Version Information

```text
Tag: v21.0.1
```

### Urgency

High

I will fix it

Maintainer feedback requested

### Anything else?

_No response_

rogurr:
* Updated CPU trait to provide cache granule size
* Added granule size reporting to x86, arm, and stub components
* Updated publishing of CPU architecture protocol to include granule size.

## Description

Addressing [Issue #1466](https://github.com/OpenDevicePartnership/patina/issues/1466)

Current Behavior:  The code file patina_dxe_core\src\cpu\cpu_arch_protocol.rs is publishing an EFI_CPU_ARCH_PROTOCOL that has a field called DmaBufferAlignment to provide the size, in bytes, of the alignment required for DMA buffer allocations which is currently publishing a byte size of 0.  The value should be the largest cache line size of any cache in the system per the UEFI PI spec, version 1.9, section II-12.3.1.

- [X] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [X] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Found bug when porting Patina to a private build chipset reference and attempting to instantiate the NVME.

## Integration Instructions

N/A


core/patina_internal_cpu/src/cpu/aarch64/cpu.rs
```diff
@@ -139,6 +139,33 @@ impl Cpu for EfiCpuAarch64 {
...
+                };
+
+                // CWG (Cache Writeback Granule): CTR_EL0 bits [27:24]
+                let cwg = ((ctr_el0 >> 24) & 0xF) as u32;
+
+                // CWG is Log2 of the max size in words
+                if cwg > 0 {
+                    4 << cwg
+                } else {
+                    2048  // Default to 2K if register contains 0 per Armv8-A spec
```
os-d:
```suggestion
                    SIZE_2KB // Default to 2K if register contains 0 per Armv8-A spec
```

core/patina_internal_cpu/src/cpu/aarch64/cpu.rs
```diff
@@ -179,4 +206,11 @@ mod tests {
         assert_eq!(cpu_init.get_timer_value(1), Err(EfiError::Unsupported));
         assert_eq!(cpu_init.get_timer_value(0), Err(EfiError::Unsupported));
     }
+
+    #[test]
+    fn test_cache_writeback_granule() {
```
os-d, line 196: I would just mark the function as coverage(off), this doesn't test anything meaningful

core/patina_internal_cpu/src/cpu/x64/cpu.rs
```diff
@@ -173,6 +173,12 @@ impl Cpu for EfiCpuX64 {
 
         Ok((timer_value, self.timer_period))
     }
+
+    fn cache_writeback_granule(&self) -> u32 {
+        // x86_64 has cache-coherent DMA, so this value is only used as a
+        // safe alignment default. 64 bytes matches the typical cache line size.
+        64
```
os-d: edk2 sets this value to 4: https://github.com/tianocore/edk2/blob/30b1cb6cabb95f1c385564957a7246882f828519/UefiCpuPkg/CpuDxe/CpuDxe.c#L32.

Unless we have a good reason, we should follow that.

core/patina_internal_cpu/src/cpu/x64/cpu.rs
```diff
@@ -226,4 +232,10 @@ mod tests {
         assert_eq!(x64_cpu_init.get_timer_value(1), Err(EfiError::InvalidParameter));
         assert_eq!(x64_cpu_init.get_timer_value(0), Ok((0, 0)));
     }
+
+    #[test]
+    fn test_cache_writeback_granule() {
```
os-d: Same.

patina_dxe_core/src/cpu/cpu_arch_protocol.rs
```diff
@@ -431,6 +447,15 @@ mod tests {
         });
     }
 
+    #[test]
+    fn test_dxe_cpu_cache_writeback_granule_delegates() {
```
os-d, line 451: This test also doesn't test anything meaningful.

core/patina_internal_cpu/src/cpu/aarch64/cpu.rs
```diff
@@ -139,6 +139,33 @@ impl Cpu for EfiCpuAarch64 {
...
+
+                // CWG is Log2 of the max size in words
+                if cwg > 0 {
+                    4 << cwg
+                } else {
+                    2048  // Default to 2K if register contains 0 per Armv8-A spec
+                }
+            } else {
+                // For test mode or non-aarch64 platforms, return 64 bytes
+                64_u32
```
joschock: consider making this a named CONST for self-documentation purposes?

joschock (approved):

core/patina_internal_cpu/src/cpu/aarch64/cpu.rs
```diff
@@ -139,6 +139,33 @@ impl Cpu for EfiCpuAarch64 {
     fn get_timer_value(&self, _timer_index: u32) -> Result<(u64, u64), EfiError> {
         Err(EfiError::Unsupported)
     }
+
+    fn cache_writeback_granule(&self) -> u32 {
+        cfg_if::cfg_if! {
+            if #[cfg(all(not(test), target_arch = "aarch64"))] {
```
makubacki: This entire module should already be constrained to `aarch64` since it is the aarch64-specific instance.

core/patina_internal_cpu/src/cpu/aarch64/cpu.rs
makubacki: Yes. That should also explain why 64 was chosen and be used consistently (like in `test_cache_writeback_granule()` if that stays).

patina_dxe_core/src/cpu/cpu_arch_protocol.rs
```diff
@@ -298,6 +303,7 @@ mod tests {
     fn test_flush_data_cache() {
         with_locked_state(|| {
             let mut cpu_init = MockEfiCpuInit::new();
+            cpu_init.expect_cache_writeback_granule().return_const(64_u32);
```
makubacki, line 306: Is there a reason to add the same check to all of these unit tests?

core/patina_internal_cpu/src/cpu/aarch64/cpu.rs
makubacki: While I don't think the test adds much value, I'm not opposed to it staying either. It is possible that the code drifts from an expected/allowed value and the test catches it.

os-d: For aarch64 because the actual implementation is dynamic, I don't think it adds value to make sure a random test only value is used. For x64, since we have a defined value, I think fine to keep the test, but again, I also don't think adding much value.

core/patina_internal_cpu/src/cpu/aarch64/cpu.rs
rogurr: I was following the same precedence used by the other functions already present, but if the cpu.rs file is selectively choosing which file to use, do we really need any of those checks?

Is there a standard mechanism for checking to verify a file is not misplaced?  Maybe something like:
```
#[cfg(not(target_arch = "aarch64"))]
    compile_error!("This file only supports target_arch = aarch64");
```

core/patina_internal_cpu/src/cpu/aarch64/cpu.rs
rogurr: Fixed, will be in next push

patina_dxe_core/src/cpu/cpu_arch_protocol.rs
rogurr: I too didn't like it much, but creating a mock cpu_init that didn't have a call the test subsystem is looking for, or having a separate function called outside the construction got too complicated.  This seemed the easiest way to make the underlying code most readable instead of trying to make the tests more readable.

core/patina_internal_cpu/src/cpu/x64/cpu.rs
rogurr: I went back and forth on this one.  The original C code was written 17 years ago where 4 bytes might have made sense, but cache lines now are typically 64 bytes and ARM dictates 2KB minimum.  There may be a CPU we have to support that has a memory bus of only 32 bytes or something, so 4 would be safer but would it be out of the norm?

Thoughts?

core/patina_internal_cpu/src/cpu/aarch64/cpu.rs
rogurr: If we don't need the aarch cfg tags in this file, this line will be removed.

core/patina_internal_cpu/src/cpu/x64/cpu.rs
os-d: I think it's somewhat irrelevant because x64 is cache coherent for DMA and nowhere in edk2 uses this value for x64. I would suggest we leave it as what edk2 has, for what we expect will be ignored for this architecture. It clearly has worked for 17 years :), so no reason to change it unless we encounter an issue.

core/patina_internal_cpu/src/cpu/aarch64/cpu.rs
makubacki: Yes, I'm only speaking about the x64 case. The logic to have a test-specific value and verify that value is set in a test is unnecessary.

core/patina_internal_cpu/src/cpu/aarch64/cpu.rs
makubacki: In this case, if the value is only here for the test scenario, we're also advocating to just drop the test case and mark the function with `coverage(off)`.

core/patina_internal_cpu/src/cpu/x64/cpu.rs
joschock: >ARM dictates 2KB minimum.

Unrelated to what it should be for x64, but my understanding is that 2KB is the maximum possible size (not the minimum) before you need a new ARM architecture :)
I suspect actual cache lines are much smaller than that in most designs.

makubacki: I agree with setting the X64 instance to `4`. This is treated as a read-only field in a public facing interface (the CPU Arch protocol). Regardless of what is in the edk2 tree, platform code could be using the protocol that we're unaware of.

rogurr: John is correct, this is a reporting for max cache line size in the platform so X64 is most likely incorrect with 4, but sticking with it since it has precedence.  Change will be in next push.

makubacki: @rogurr, as the field description is written, I think we're all acknowledging that if it truly represents the max cache line size, it's `64`. In your update, you can include a comment that says while this conflicts with the field definition it is retained for compatibility with the edk2 implementation.

core/patina_internal_cpu/src/cpu/aarch64/cpu.rs
makubacki: I created a PR [here](https://github.com/OpenDevicePartnership/patina/pull/1471) to clean that up. After it is completed, you can rebase onto it and avoid going against the convention in the rest of the file.

core/patina_internal_cpu/src/cpu/aarch64/cpu.rs
rogurr: I changed what is compiled out when in test mode to verify the shift/mask portion of the function is correct.  So keeping this function, but removing the X64 test function since it is just testing for a hard coded value.

core/patina_internal_cpu/src/cpu/x64/cpu.rs
rogurr: Removed

patina_dxe_core/src/cpu/cpu_arch_protocol.rs
rogurr: Removed

core/patina_internal_cpu/src/cpu/aarch64/cpu.rs
rogurr: X86 architecture path removed, kept aarch64 path and only wrapped asm code with test cfg, so test was left to confirm bit mask/shift.

core/patina_internal_cpu/src/cpu/aarch64/cpu.rs
rogurr: No need to do another PR, I already added the change to this PR and included removing some of the curly braces to make it look the same as other usage models.

patina_dxe_core/src/cpu/cpu_arch_protocol.rs
rogurr: I put this back in because the other 2 functions had tests that really just tested that the stub returned and wanted them all to behave the same.  I'd be happy to open a task to scrub these tests and do them in 1 shot instead of tacking on changes to never get this PR closed.

core/patina_internal_cpu/src/cpu/aarch64/cpu.rs
makubacki: The PR was created before you made the changes. It is also an entirely unrelated changes to this PR and would need to be in a separate PR anyway. You just need to rebase and drop the change suggested in this thread.

core/patina_internal_cpu/src/cpu/aarch64/cpu.rs
```diff
@@ -15,6 +15,9 @@ use patina::{
 };
 use r_efi::efi;
 
+#[cfg(not(target_arch = "aarch64"))]
+compile_error!("This file only supports aarch64");
+
```
makubacki: The parent module (patina_internal_cpu/src/cpu.rs) already gates this module on `target_arch=aarch64`. Having logic like this in the .rs files pollutes the file with architectural conditions that we're trying to avoid in the first place with separating the module behind the architectural gate.

```suggestion
```

core/patina_internal_cpu/src/cpu/aarch64/cpu.rs
```diff
@@ -48,75 +52,63 @@ impl EfiCpuAarch64 {
...
-            // SAFETY: a data barrier has no impact on safety invariants.
-            unsafe {
-                asm!("dsb sy", options(nostack));
-            }
+        #[cfg(not(test))]
+        // we have a data barrier after all cache lines have had the operation performed on them as an optimization
+        // SAFETY: a data barrier has no impact on safety invariants.
+        unsafe {
+            asm!("dsb sy", options(nostack));
         }
```
makubacki, line 57: As stated in the comment thread, the change to remove redundant architectural cfg gates is entirely unrelated to this PR which is to specify the correct value for DmaBufferAlignment in the CPU arch protocol.

This PR should remain a simple, focused bug fix for that issue.

core/patina_internal_cpu/src/cpu/aarch64/cpu.rs
rogurr: That's why I was asking the question earlier if you had a standard way to check.  This file now has several sections running assembly code specific to aarch64, so this check is similar to using .ERR flags to make sure the file is only compiled as expected.  You can't rely on a calling file to guarantee this file is not used incorrectly. 

Please let me know if you have a preferred macro or compilation block besides the compile_error macro.

core/patina_internal_cpu/src/cpu/aarch64/cpu.rs
rogurr: As stated in the comment thread, the changes originally made were following precedent but I was asked to change them to a new method.  So as to not make the file messy, I was updating all of the calls to match what you wanted.

cfernald (approved):

core/patina_internal_cpu/src/cpu/aarch64/cpu.rs
makubacki: Sorry if there was a misunderstanding but it was not requested that you change all redundant architectural gates in the crate in this PR just that the line added in this PR was not needed in this PR.

core/patina_internal_cpu/src/cpu/aarch64/cpu.rs
makubacki: Can you describe the scenario you have in mind? This module is only resolved through the architectural condition:

https://github.com/OpenDevicePartnership/patina/blob/0e289d884f74983e629f8258cbc8b2587539ca41/core/patina_internal_cpu/src/cpu.rs#L20-L25

I'm not aware of a way this would get compiled when `target_arch` does not equal `aarch64` without deliberately bypassing this standard hierarchical path.

rogurr: Talked about this in the weekly meeting, sounds like my concern is not really a Rust-ism and more of a C-ism to enforce a specific build for files.  This file is currently being re-vamped so I'll pull this out so we can implement a standard mechanism.

PR #1470, opened 2026-04-16
Safety & UB Updates [Rebase & FF]

makubacki:
## Description

Fixes some safety and undefined behavior (UB) issues.

Reported by Claude with a UB/safety focused prompt. Manually filtered out less severe issues, false positives, and made fixes.

---

**boot_services: Fix UB in create_event_ex() Option<fn> transmute**

create_event_ex() transmuted Option<extern fn> to extern fn, dropping
the Option wrapper. Passing None would produce a null function pointer
typed as fn(...) which is instant UB since fn pointers are guaranteed
non-null.

Align create_event_ex() with create_event() by preserving the Option
wrapper through the transmute chain. Update create_event_ex_unchecked()
to accept Option<EventNotifyCallback> matching create_event_unchecked().

---

**patina_dxe_core: Fix TplGuard Deref lifetime unsoundness**

TplGuard's Deref and DerefMut returned references with the mutex's
lifetime ('a) instead of the guard's lifetime. This allowed extracting
references that outlive the guard, accessing mutex-protected data
without the lock held.

Tie the returned reference lifetime to &self (the guard) instead of
'a (the mutex), matching the correct pattern used in sdk/patina's
TplMutexGuard.

---

**patina_macro: Reject records without #[string_pool] at compile time**

The SmbiosRecordStructure trait requires string_pool_mut() to return
&mut Vec<String>. Records without a `#[string_pool]` field previously
used a static mut Vec which was mutable aliasing UB.

Replace with a compile-time error requiring all records to declare a
`#[string_pool]` field for trait conformance, even if never populated.
This makes the invalid state unrepresentable rather than deferring to
runtime.

----

**patina_sdk: Replace unreachable_unchecked() with unreachable in Service Deref**

The downcast failure path used unreachable_unchecked() which is UB if
reached. While all intended construction paths guarantee correct types,
the public From<&'static dyn Any> impl accepts any dyn Any, creating a
reachable path to UB. Replace with unreachable! which provides a safe
panic with a diagnostic message if the invariant is ever violated.

---

**boot_services: Use NonNull::dangling for ZST protocol references**

The ZST protocol path returned &'static mut T from a shared static (),
creating aliasing &mut references across concurrent callers.

Use NonNull::dangling() instead, which gives each caller a unique
aligned pointer without shared state.

---

**runtime: Use raw pointers for linked list manipulation to avoid aliasing**

When the runtime image/event lists are empty, the previous code created
two &mut references to the same image_head/event_head field on the same
line.

Follow the advice in https://rust-unofficial.github.io/too-many-lists/fifth-stacked-borrows.html#managing-stacked-borrows
and use pointers with addr_of_mut! throughout the list code.

---

**fv: Use checked arithmetic for FV read address calculation**

saturating_add() silently produces usize::MAX on overflow instead of
returning an error. A malformed FV with large LBA values could trigger
this, leading to a wild pointer in from_raw_parts().

Switch to checked_add and propagate InvalidParameter on overflow.

---

**fw_fs: Use read_unaligned() for FV header parsing from byte buffer**

The byte buffer has alignment 1 but fv::Header contains u32/u64 fields
requiring higher alignment. Creating a reference via &*(ptr as *const
fv::Header) is UB on unaligned data.

Use ptr::read_unaligned() to safely copy the header out of the buffer,
matching the approach used in patina_ffs.

---

**hob_list: Guard against zero-length or overflowing HOB entries**

A malformed HOB with length 0 causes an infinite loop, and a very
large length can wrap the pointer on 32-bit targets. Break out of
the parsing loop if the length is too small to contain a HOB header
or if the pointer addition overflows.

---

**patina_mm: Use read_unaligned() for communicate buffer header parsing**

The communicate buffer header fields were read with ptr::read() which
requires aligned pointers. The buffer comes from firmware memory that
may not be aligned to BinaryGuid or usize requirements depending on
the platform. Switch to ptr::read_unaligned() to be correct regardless
of alignment.

---

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- `cargo make all`

## Integration Instructions

- N/A

### DIFF
--- a/sdk/patina/src/component/service.rs
+++ b/sdk/patina/src/component/service.rs
@@ -354,27 +354,16 @@ impl<T: ?Sized + 'static> Deref for Service<T> {
             if let Some(service) = service.downcast_ref() {
                 service
             } else {
-                // Using core::hint::unreachable_unchecked() here results in the compiler optimizing away this entire
-                // code path, and will result in UB if the path is reached. This code path truly is unreachable as we
-                // (as patina developers) control all ways of instantiating a Service type.
+                // All construction paths (mock, from, initialize) guarantee that the downcast will succeed.
+                // unreachable! is used here instead of unreachable_unchecked to avoid UB if this invariant is
+                // ever violated (e.g., via the public From<&'static dyn Any> impl with a mismatched type).
                 //
-                // The performance impact of using this over unreachable! or panic! is about 25% improved performance
-                // in benchmarks.
-                //
-                // # SAFETY
-                // - The `Service` type tightly couples the underlying type to the `dyn Any` type for downcasting.
-                // - All ways of instantiating a `Service` type are tightly controlled to ensure that this downcast is
-                //   valid and will never fail including:
-                //   - The `mock` function, which requires a `Box<dyn T>` which is then manually leaked, ensuring the
-                //     underlying type is always available and the correct type.
-                //   - The `from` function, which is passed data from `Storage`, that is guaranteed to be the correct
-                //     type as it is generated via the `IntoService` macro and out of the hands of the user.
-                //   - The `initialize` function, which consumes another `Service` type, which has the same guarantees
-                //     as above.
-                // - If the Service is uninitialized, it will panic at the normal unreachable! macro call below.
-                // SAFETY: Service value was validated to be initialized in the Some(v) match arm.
-                // unreachable_unchecked provides optimizer hints for the impossible None case.
-                unsafe { core::hint::unreachable_unchecked() }
+                // However, this is a potential point for performance improvement in the future as the component
+                // infrastructure evolves and the number of components increases.
+                unreachable!(
+                    "Service downcast failed — this indicates a type mismatch in Service<{}> construction",
+                    core::any::type_name::<T>()
+                )
             }
         } else {
             // We use unreachable! here instead of panic! as this provides compiler hints to the optimizer. We cannot
--- a/components/patina_mm/src/config.rs
+++ b/components/patina_mm/src/config.rs
@@ -406,11 +406,13 @@ impl CommunicateBuffer {
 
         let header_slice = &self.as_slice()[..Self::MESSAGE_START_OFFSET];
 
-        // SAFETY: Buffer size validated, BinaryGuid is repr(transparent) over repr(C) efi::Guid at offset 0
-        let memory_guid = unsafe { core::ptr::read(header_slice.as_ptr() as *const patina::BinaryGuid) };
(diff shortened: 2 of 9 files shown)


os-d (approved):

vineelko (approved):

sdk/patina/src/component/service.rs
```diff
@@ -354,27 +354,13 @@ impl<T: ?Sized + 'static> Deref for Service<T> {
...
-                //   - The `initialize` function, which consumes another `Service` type, which has the same guarantees
-                //     as above.
-                // - If the Service is uninitialized, it will panic at the normal unreachable! macro call below.
-                // SAFETY: Service value was validated to be initialized in the Some(v) match arm.
-                // unreachable_unchecked provides optimizer hints for the impossible None case.
-                unsafe { core::hint::unreachable_unchecked() }
+                // All construction paths (mock, from, initialize) guarantee that the downcast will succeed.
+                // Using unreachable! instead of unreachable_unchecked to avoid UB if this invariant is
+                // ever violated (e.g., via the public From<&'static dyn Any> impl with a mismatched type).
+                unreachable!(
```
Javagedes, line 363: I will say, I purposefully used `unreachable_unchecked` here because I could reason and validate that this code path is completely unreachable.

This was purposefully done as a code optimization because `core::hint::unreachable_unchecked()` allows the compiler to eliminate all branches in the surrounding code. This is a pretty hot path. Every time you call `Service<dyn MyService>.<function>, this code is executed. So I suspect this will have a negative performance impact, though I have not investigated how many times we call this function in a run

makubacki: I assumed "25% improved performance" to be on the order of ns. If you expect a meaningful perf impact, I'm happy to revert this change.

Javagedes: For now it should be fine;  it's only called ~3200 times. I'm just slightly concerned about in the future.

makubacki: Do you happen to know what the time difference is for those 3200 calls? If it's too much trouble to get, don't worry about it.

I know this will vary per platform, etc. But a reference point would help if it is available.

Javagedes: Image

From our service benchmarks, I noted a 16% perf impact on dyn services `Service<dyn MyService>` from 3.5ns to 4.1ns per deref on average and a 29% perf impact on concrete services (`Service<MyStructService>`) from 2.6ns to 3.2ns per deref on average

makubacki: Per an offline discussion, we decided to proceed with `unreachable!()` for now, leaving a note about performance for reference in the future.

PR #1474, opened 2026-04-22
RFC: ACPI Service Interface Changes

Javagedes:
## Description

Current Status: **Draft**

Implementation Example: **In the works**

This RFC proposes interface changes to reduce the two services (`Service<dyn AcpiProvider>` and
`Service<AcpiTableManager`>) to a single `Service<dyn Acpi>` service and provide a public Helper struct, `Table` for
managing ACPI tables both internal to the service implementation and as a consumer of the service. This `Table`
structure would better support multiple types of ACPI tables, such as statically sized and dynamically sized ACPI
tables.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

N/A

## Integration Instructions

N/A

### DIFF
--- a/docs/src/rfc/text/0000-acpi-interface-improvements.md
+++ b/docs/src/rfc/text/0000-acpi-interface-improvements.md
@@ -1,7 +1,7 @@
 # RFC: `patina_acpi` service interface improvements
 
 This RFC proposes interface changes to reduce the two services (`Service<dyn AcpiProvider>` and
-`Service<AcpiTableManager`>) to a single `Service<dyn Acpi>` service and provide a public Helper struct,`Table` for
+`Service<AcpiTableManager`>) to a single `Service<dyn Acpi>` service and provide a public Helper struct, `Table` for
 managing ACPI tables both internal to the service implementation and as a consumer of the service. This `Table`
 structure would better support multiple types of ACPI tables, such as statically sized and dynamically sized ACPI
 tables.
@@ -14,12 +14,12 @@ tables.
 
 The main motivation behind this change is to add support for DSTs (Dynamically Sized Types) in the generics portion of
 the service interface. DST ACPI tables are currently supported via usage of the non-generic interfaces ACPI service
-methods, as they simply take a slice of bytes that start with an ACPI table header. This puts all safety burdons on the
+methods, as they simply take a slice of bytes that start with an ACPI table header. This puts all safety burdens on the
 developer. The proposed interface changes and additional support for DSTs would be highly beneficial as it would
 provide a standard interface for interacting with both static and dynamically sized ACPI tables.
 
 Supporting DSTs will require some fundamental changes to how ACPI table data is managed (described in-depth below). The
-high level description is that we will be removing the current generic table structure in favor of `Box<[u8]>`, which
+high-level description is that we will be removing the current generic table structure in favor of `Box<[u8]>`, which
 self manages the ACPI page allocation while relying on `zerocopy` to interpret the table byte prefixes as an
 `AcpiTableHeader`. Due to this, it will remove a large chunk of `unsafe` blocks throughout the crate.
 
@@ -71,7 +71,7 @@ it was not directly referencing the true ACPI table memory.
 
 This change is focused mostly on the service API to provide common logic for table management (creation, updating)
 on the consumer side of the service. Only a very minimal set of changes are made to the service implementation to
-accompidate the interface changes. These changes will remove unsafe code in the service implementation as table
+accommodate the interface changes. These changes will remove unsafe code in the service implementation as table
 access will move away from pointers and dereferencing to slices and safe byte interpretation via `zerocopy`.
 
 Starting with changes to the service implementation and data management: A new unsafe trait, `AcpiTable` is created to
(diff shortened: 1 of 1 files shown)


docs/src/rfc/text/0000-acpi-interface-improvements.md
```diff
@@ -0,0 +1,554 @@
+# RFC: `patina_acpi` service interface improvements
```
os-d, line 1: One process question, should this be an amendment to the original patina-acpi RFC instead of a new one?

docs/src/rfc/text/0000-acpi-interface-improvements.md
```diff
@@ -0,0 +1,554 @@
...
+provide a standard interface for interacting with both static and dynamically sized ACPI tables.
+
+Supporting DSTs will require some fundamental changes to how ACPI table data is managed (described in-depth below). The
+high-level description is that we will be removing the current generic table structure in favor of `Box<[u8]>`, which
+self manages the ACPI page allocation while relying on `zerocopy` to interpret the table byte prefixes as an
+`AcpiTableHeader`. Due to this, it will remove a large chunk of `unsafe` blocks throughout the crate.
+
+The final benefit is that of safety. This change removes many different areas of the code wrapped in `unsafe`, and most
+importantly, stops the `get_acpi_table` method from returning a pointer directly to the actual ACPI table in memory.
+Instead, it will now copy the bytes into new memory and provide that to the user. The EDKII EFIAPI side will need to
```
os-d, line 28:
```suggestion
Instead, it will now copy the bytes into new memory and provide that to the user. The C FFI interface will need to
```
nit: Not scoped to EDK II

docs/src/rfc/text/0000-acpi-interface-improvements.md
```diff
@@ -0,0 +1,554 @@
...
+The alternative is to continue the status quo. As mentioned above, DSTs are currently supported via the non-generic
+service interface, but it requires you to (1) Manually construct the `AcpiTable` struct (A wrapper around a raw
+pointer), (2) Ensure you've allocated to the correct memory type yourself, and (3) Manually attempt casting a generic
+table to a table of your choosing.
+
+## Prior Art
+
+The prior implementation generically stored each ACPI table as a pointer to a leaked union representing the ACPI
+table. It would then provide direct pointer access to the caller of `get_acpi_table` both for the patina component
+and caller of the EDK II protocol. Additionally, the generic methods (such as `get_acpi_table<T>`) would directly
```
os-d, line 66:
```suggestion
and caller of the UEFI spec protocol. Additionally, the generic methods (such as `get_acpi_table<T>`) would directly
```

docs/src/rfc/text/0000-acpi-interface-improvements.md
```diff
@@ -0,0 +1,554 @@
...
+/// # Safety
+/// 
+/// Implementors must ensure that:
+/// - The type begins with an `AcpiTableHeader` as its first field
+/// 
+/// # Examples
+/// 
+/// <Skipped for brevity here as there will be examples below>
+/// ```
+pub unsafe trait AcpiTable: FromBytes + IntoBytes + KnownLayout + Unaligned + Immutable {
```
os-d, line 121: Can this really be Immutable? Presumably you would return an AcpiTable from GetTable and it would be valid for someone to modify it and republish it? Same for the fields below. Perhaps I don't know what the Immutable trait bound is guaranteeing here.

os-d (approved):

docs/src/rfc/text/0000-acpi-interface-improvements.md
Javagedes: The `Immutable` trait is a bit of misnomer in my personal opinion. Implementing `Immutable` here means it is free from interior mutability.

https://docs.rs/zerocopy/0.8.48/zerocopy/trait.Immutable.html

os-d: Gotcha, thanks.

docs/src/rfc/text/0000-acpi-interface-improvements.md
```diff
@@ -0,0 +1,554 @@
...
+    /// The kind of ACPI table this is (Fixed, Dst). Used to gate slice style methods to only DST ACPI tables.
+    #[allow(private_bounds)]
+    type Kind: TableKind = Fixed;
+}
+
+/// An element that can be used for the variable-length array portion of DST ACPI tables.
+/// 
+/// This trait has two roles: (1) Ensure that each element can be converted to and from bytes using `zerocopy` and
+/// (2) to differentiate between Sized and Unsized elements for indexing optimizations.
+pub trait Element: IntoBytes + FromBytes + Immutable + KnownLayout + Unaligned {
```
cfernald, line 140: For the sake of avoiding internal panics, should these be `TryFromBytes` to allow handling malform/unexpected elements without panicking in the parser?

docs/src/rfc/text/0000-acpi-interface-improvements.md
```diff
@@ -0,0 +1,554 @@
...
+
+    /// The kind of ACPI table this is (Fixed, Dst). Used to gate slice style methods to only DST ACPI tables.
+    #[allow(private_bounds)]
+    type Kind: TableKind = Fixed;
+}
+
+/// An element that can be used for the variable-length array portion of DST ACPI tables.
+/// 
+/// This trait has two roles: (1) Ensure that each element can be converted to and from bytes using `zerocopy` and
+/// (2) to differentiate between Sized and Unsized elements for indexing optimizations.
```
cfernald, line 139: How does ZeroCopy work is element can take different forms? A good example is the [MADT](https://uefi.org/specs/ACPI/6.6/05_ACPI_Software_Programming_Model.html#multiple-apic-description-table-madt) where not only are the elements different sizes but also completely different but well-structured contents. This could be a union, but I assume zerocopy may be unhappy if the union size is greater than the bytes provided.

PR #1475, opened 2026-04-23
[REBASE & FF] Move Architectural Interfaces to SDK

os-d:
## Description

This contains three commits:

Drop x86_64 crate
--
Patina has light usage of the x86_64 crate and the main usage is for the IDT and GDT, where a different, simplified design is desired.

This updates all consumers to not use the crate and drops the crate.

Use Standard Exception Handler for #DF
--
Currently, a feature from x86_64 was being used to have special handling for a double fault. However, this is not required as we can use our standard exception handler and identify this is a double fault.

This commit drops use of the feature and instead handles the #DF as we normally do.

Move Architectural Interfaces to SDK's Arch Mod
--
Centralize architectural interfaces in the SDK instead of sprinkling them around the code.

As a result, the aarch64-cpu crate is no longer required.

This also fixes hyperlinks in the copilot-instructions.md doc.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [x] Includes documentation?

## How This Was Tested

Booting to Windows and Linux on Q35 and SBSA. Booting to Windows on a physical Intel device. On Q35 and the physical Intel device, connecting the debugger to confirm the IDT changes are working correctly.

Tested the #DF changes by generating a double fault on Q35 and a physical Intel platform by changing RSP to a bad value then generating a #PF.

## Integration Instructions

N/A.

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -16,7 +16,7 @@ The workspace is organized into four top-level areas:
 
 Crate dependency rules are strict. Components depend on `sdk/` only - never on each other or on `core/`. The SDK
 depends only on generic external crates. Core crates may depend on each other and on the SDK. See
-[code_organization.md](docs/src/dev/code_organization.md) for the full dependency matrix.
+`docs/src/dev/code_organization.md` for the full dependency matrix.
 
 ## Build and Test Commands
 
@@ -31,7 +31,7 @@ All commands go through `cargo make`. Never run raw `cargo` commands.
 - `cargo make coverage` - Generate test coverage reports
 - `cargo make doc` - Build documentation
 
-See [Makefile.toml](Makefile.toml) for the complete command list.
+See `Makefile.toml` for the complete command list.
 
 ## Module Organization
 
@@ -49,7 +49,7 @@ See [Makefile.toml](Makefile.toml) for the complete command list.
 // src/memory.rs
 ```
 
-See [component requirements](docs/src/component/requirements.md) for crate layout standards.
+See `docs/src/component/requirements.md` for crate layout standards.
 
 ## Safety Conventions
 
@@ -64,7 +64,7 @@ See [component requirements](docs/src/component/requirements.md) for crate layou
   itself cannot verify internally. If the function validates all inputs before the
   unsafe operation, it can remain safe.
 
-See [unsafe.md](docs/src/dev/principles/unsafe.md) for the full safety philosophy
+See `docs/src/dev/principles/unsafe.md` for the full safety philosophy
 (software safety vs. hardware safety distinctions).
 
 ## Error Handling
@@ -77,7 +77,7 @@ See [unsafe.md](docs/src/dev/principles/unsafe.md) for the full safety philosoph
 - Use `expect("descriptive message")` over bare `unwrap()`. Reserve `unwrap()` for
   test code only.
 
-See [error-handling.md](docs/src/dev/principles/error-handling.md) for error
+See `docs/src/dev/principles/error-handling.md` for error
 propagation patterns and examples.
 
 ## Component Model
@@ -98,8 +98,8 @@ A component only executes when all its declared dependencies are available.
 - Use the **stored dependencies** pattern: store all dependency references as fields in
   the component struct. The entry point stores references; methods use them.
 
-See [component interface](docs/src/component/interface.md) and
-[getting started](docs/src/component/getting_started.md) for details and examples.
+See `docs/src/component/interface.md` and
+`docs/src/component/getting_started.md` for details and examples.
 
 ## Mocking and Testing
 
@@ -109,18 +109,18 @@ See [component interface](docs/src/component/interface.md) and
 - Prefer `mockall` for trait mocking in unit tests. Use `#[automock]` on traits to generate mocks.
 - Use `pretty_assertions` for readable test diffs.
 
-See [testing.md](docs/src/dev/testing.md) for the full testing strategy.
+See `docs/src/dev/testing.md` for the full testing strategy.
 
 ## Trait Design
 
 Traits serve as **abstraction points** - not code reuse mechanisms. Use crates for
-code reuse (see [reuse.md](docs/src/dev/principles/reuse.md)).
+code reuse (see `docs/src/dev/principles/reuse.md`).
 
 - Define traits to represent swappable behavior (e.g., `BootServices`, `SerialIO`).
 - Implementations can vary per platform without affecting consumers.
 - Keep traits focused on a single responsibility (Interface Segregation).
 
-See [abstractions.md](docs/src/dev/principles/abstractions.md) for the full trait
+See `docs/src/dev/principles/abstractions.md` for the full trait
 design philosophy.
 
 ## Documentation Standards
@@ -133,7 +133,7 @@ design philosophy.
   should be self-evident. In cases where there is unavoidable ambiguity, add these sections
   to provide clarity.
 
-See [documenting.md](docs/src/dev/documenting.md) for templates and style guides.
+See `docs/src/dev/documenting.md` for templates and style guides.
 
 ## UEFI-Specific Guidelines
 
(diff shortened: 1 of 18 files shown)


os-d:
Note, there were several places where I chose not to move into the SDK because I did not feel it is appropriate at this time, namely in the IDT/GDT code, the serial code (where on x86 it disables interrupts to write to the serial port, then re-enables if they were enabled), and in the CPU arch protocol (disabling interrupts, enabling). I didn't think these were currently valid to expose in the SDK. Happy to discuss here.

core/patina_internal_cpu/src/cpu/x64/gdt.rs
```diff
@@ -8,252 +8,261 @@
...
+    limit15_0: u16,
+    base15_0: u16,
+    base23_16: u8,
+    type_: u8,
+    limit19_16_and_flags: u8,
+    base31_24: u8,
+}
+
+impl From<GdtEntry> for u64 {
+    fn from(entry: GdtEntry) -> Self {
```
joschock, line 28: minor, feel free to close as "not worth it" -> safety on conversion of u64 with invalid contents to GDT entry?

joschock (approved):

core/patina_internal_cpu/src/cpu/x64/gdt.rs
os-d: I'll add a comment, its true when you impl From you get Into for free, though we only use From

.github/copilot-instructions.md
```diff
@@ -16,7 +16,7 @@ The workspace is organized into four top-level areas:
 
 Crate dependency rules are strict. Components depend on `sdk/` only - never on each other or on `core/`. The SDK
 depends only on generic external crates. Core crates may depend on each other and on the SDK. See
-[code_organization.md](docs/src/dev/code_organization.md) for the full dependency matrix.
+[code_organization.md](../docs/src/dev/code_organization.md) for the full dependency matrix.
```
makubacki: I assumed AI viewed these from the perspective of the workspace not the instructions file. This change looks good. Did you find a problem with the files being accessed or just addressing a markdown lint?

makubacki (approved):

core/patina_internal_cpu/src/interrupts/x64/idt.rs
```diff
@@ -10,89 +10,135 @@
...
+    // SAFETY: Index has been validated to be in [0, 255].
+    unsafe { AsmGetVectorAddress(index) }
+}
 
-    // SAFETY: We have validated we are using the architecturally guaranteed indices
-    unsafe { VirtAddr::from_ptr(AsmGetVectorAddress(index) as *const ()) }
+lazy_static! {
+    static ref IDT: Idt = {
+        let mut idt = Idt { entries: [IdtEntry::empty(); 256] };
+        let cs = read_cs();
```
cfernald: Do you think the IDT could be static? Since we are setting up the GPT and IDT ourselves, the CS should be known?

core/patina_internal_cpu/src/interrupts/x64/idt.rs
```diff
@@ -10,89 +10,135 @@
...
 /// Handler for double faults.
 ///
-/// Handler for double faults that is configured to run as a direct interrupt
-/// handler without using the normal handler assembly or stack. This is done to
-/// increase the diagnosability of faults in the interrupt handling code.
-///
+/// Configured to run as a direct interrupt handler without using the normal
+/// handler assembly or stack. This is done to increase the diagnosability of
+/// faults in the interrupt handling code.
 extern "x86-interrupt" fn double_fault_handler(stack_frame: InterruptStackFrame, _error_code: u64) {
```
cfernald: Not necessarily for this PR, but e can probably drop this feature as well and just use our regular exception handler assembly.

sdk/patina/src/serial/uart.rs
```diff
@@ -31,7 +31,25 @@ cfg_if::cfg_if! {
...
+
+        /// Runs `f` with CPU interrupts disabled, restoring the previous interrupt state afterward.
+        fn without_interrupts<F, R>(f: F) -> R
+        where
+            F: FnOnce() -> R,
+        {
+            let flags: u64;
+            // SAFETY: Reading RFLAGS and disabling interrupts around the closure.
+            unsafe {
+                core::arch::asm!("pushfq; pop {0}; cli", out(reg) flags, options(nostack, preserves_flags));
```
cfernald: I need to double check, but i assume the push/pop breaks `nostack`

cfernald (approved):

sdk/patina/src/serial/uart.rs
os-d: Agreed, good catch

core/patina_internal_cpu/src/interrupts/x64/idt.rs
os-d: Thanks, meant to remove it

vineelkovvuri (approved):

vineelko (approved):

.github/copilot-instructions.md
os-d: I was just addressing the markdown lint. I took a look through the documentation (and asked copilot itself :) and I'm not sure it can follow these links. The docs say you should write this in valid markdown, but then also points to this repo for examples: https://github.com/github/awesome-copilot/blob/main/.github/copilot-instructions.md. And in there, it just writes the path out, not in a markdown link or just gives a direct link.

So maybe we should convert these to just paths (relative to the repo root as they were) and not as markdown links? Hard to really test if the links are automatically followed.

sdk/patina/src/serial/uart.rs
os-d: It also doesn't preserve flags because it is clearing the interrupt flag :)

.github/copilot-instructions.md
makubacki: Thanks for the link. I saw some examples that were relative to the repo root as well (as before) and I made this markdown links by habit and trying to be helpful.

I guess we should just stick with the example and leave them relative to the root and drop the link syntax.

os-d: If we don't mind it being in this PR, I'll amend that here. Otherwise I can make a new PR for it.

makubacki: I have no concern with it being amended in this PR.

core/patina_internal_cpu/src/interrupts/x64/idt.rs
os-d: Changed this to a static

core/patina_internal_cpu/src/interrupts/x64/idt.rs
os-d: Adding a separate commit to handle this

cfernald (approved):

core/patina_internal_cpu/src/interrupts/x64/idt.rs
```diff
@@ -7,92 +7,118 @@
...
+    unsafe { AsmGetVectorAddress(index) }
 }
 
+static IDT: StaticIdt = StaticIdt(Cell::new(Idt { entries: [IdtEntry::empty(); 256] }));
+
 pub fn initialize_idt() {
-    if &IDT as *const _ as usize >= SIZE_4GB {
+    let cs = crate::cpu::x64::gdt::CODE_SELECTOR;
+    // SAFETY: IDT setup is serialized during CPU initialization before interrupts are enabled.
+    let idt = unsafe { &mut *IDT.0.as_ptr() };
```
cfernald: Is this UB? Should it be `UnsafeCell`?

os-d: Discussed offline. Its not clear that this is UB because Cell safely offers this routine. But it's also not clear it is safe, so moved to UnsafeCell to be clear.

cfernald (approved):

PR #498, opened 2025-06-06
Performance Doc

magravel:
## Description

Add documentation for Performance Component.

Fix #70 

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [x] Includes documentation?

## How This Was Tested

N/A

## Integration Instructions

N/A

PR #1476, opened 2026-04-24
patina_performance: Log cross-module phase markers

cfernald:
## Description

The current implement does not log the cross-module markers for the transitions
from PEI to DXE. In EDK2, these are logged by the DXE Core. Here we need
to do this from the component because it initializes the underlying
functionality. This is not ideal, but it is the best we can do without
merging the performance component into the core.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Dumped the performance data using `dp` in the shell

## Integration Instructions

N/A


cfernald:
@makubacki I've talked about this some offline. This PR makes the situation better, but is really not ideal. I think we should probably fold the performance logic into the core. Right now, we cannot capture measurements until some point in component initialization. This means that our phase measurements are relatively inaccurate and we cannot trace earlier phases. What do you think?

I'd like to put this PR in, but I think we should plan to move this code into the core, perhaps gated by a crate feature if we want.

makubacki:
> @makubacki I've talked about this some offline. This PR makes the situation better, but is really not ideal. I think we should probably fold the performance logic into the core. Right now, we cannot capture measurements until some point in component initialization. This means that our phase measurements are relatively inaccurate and we cannot trace earlier phases. What do you think?
> 
> I'd like to put this PR in, but I think we should plan to move this code into the core, perhaps gated by a crate feature if we want.

There was [feedback](https://github.com/OpenDevicePartnership/patina/pull/284#issuecomment-2755619465) to move it to a component that led to that path. There was always supposed to be [a plan](https://github.com/OpenDevicePartnership/patina/pull/498) to allow logging directly in the core, but I don't think everything landed (and we shouldn't be referencing/using mu_rust_helpers at all now).

If you'd like to take up the new design, that would be great, but I think we would need a new RFC that enumerates the different performance sources from PEI records, MM records, to components, Patina DXE Core, and C drivers and how it all is cohesively handled.

makubacki:
Conversations that I recall about component support at that time were more about providing a component interface path not necessarily being exclusively implemented in a component.

I haven't looked at the design in a while, but I assume you have, and if this is the earliest we can do this for now, I'll approve to at least have the phase markers included.

cfernald:
> Conversations that I recall about component support at that time were more about providing a component interface path not necessarily being exclusively implemented in a component.
> 
> I haven't looked at the design in a while, but I assume you have, and if this is the earliest we can do this for now, I'll approve to at least have the phase markers included.

I've filed #1483 to track the more in-depth changes needed, but i'd like to take this to at least get tooling consistency for now.

makubacki (approved):

makubacki:
> > Conversations that I recall about component support at that time were more about providing a component interface path not necessarily being exclusively implemented in a component.
> > I haven't looked at the design in a while, but I assume you have, and if this is the earliest we can do this for now, I'll approve to at least have the phase markers included.
> 
> I've filed #1483 to track the more in-depth changes needed, but i'd like to take this to at least get tooling consistency for now.

I created this issue to encompass that one and track cleaning everything up: https://github.com/OpenDevicePartnership/patina/issues/1484

Javagedes (approved):

PR #1477, opened 2026-04-24
[REBASE & FF] Add Ability To Hide Debugger Monitor Commands

os-d:
## Description

 patina_debugger: Allow Hidden Monitor Cmds
--
Patina adds some monitor cmds that are not intended to be run by a user, such as the reload cmd. UefiExt is expected to run this in response to the !loadcore cmd.

However, the monitor help will show this as a valid command for a user to run, which is confusing. This hides that command as well as generically providing a mechanism to do so.

Drop GCD Debugger Monitor Cmd
--
A more advanced GCD command is available as of UefiExt v0.5.0. Drop the monitor cmd.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested by connecting Q35 and a physical Intel platform to the debugger and confirming the reload and gcd monitor cmds do not show up any longer but that !loadcore still works.

## Integration Instructions

Upgrade to the latest [UefiExt release](https://github.com/microsoft/uefi_debug_tools/releases) to ensure the UefiExt GCD command is available.

### DIFF
--- a/core/patina_debugger/src/lib.rs
+++ b/core/patina_debugger/src/lib.rs
@@ -42,7 +42,7 @@
 //!     patina_debugger::set_debugger(&DEBUGGER);
 //!
 //!     // Setup a custom monitor command for this platform.
-//!     patina_debugger::add_monitor_command("my_command", "Description of my_command", |args, writer| {
+//!     patina_debugger::add_monitor_command("my_command", Some("Description of my_command"), |args, writer| {
 //!         // Parse the arguments from _args, which is a SplitWhitespace iterator.
 //!         let _ = write!(writer, "Executed my_command with args: {:?}", args);
 //!     });
@@ -177,7 +177,7 @@ trait Debugger: Sync {
     fn add_monitor_command(
         &'static self,
         command: &'static str,
-        description: &'static str,
+        description: Option<&'static str>,
         callback: alloc::boxed::Box<MonitorCommandFn>,
     );
 }
@@ -297,14 +297,14 @@ pub fn initialized() -> bool {
 /// ## Example
 ///
 /// ```rust
-/// patina_debugger::add_monitor_command("my_command", "Description of my_command", |args, writer| {
+/// patina_debugger::add_monitor_command("my_command", Some("Description of my_command"), |args, writer| {
 ///     // Parse the arguments from _args, which is a SplitWhitespace iterator.
 ///     let _ = write!(writer, "Executed my_command with args: {:?}", args);
 /// });
 /// ```
 ///
 #[cfg(feature = "alloc")]
-pub fn add_monitor_command<F>(cmd: &'static str, description: &'static str, function: F)
+pub fn add_monitor_command<F>(cmd: &'static str, description: Option<&'static str>, function: F)
 where
     F: Fn(&mut core::str::SplitWhitespace<'_>, &mut dyn core::fmt::Write) + Send + Sync + 'static,
 {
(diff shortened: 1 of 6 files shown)


cfernald (approved):

makubacki (approved):

core/patina_debugger/src/lib.rs
```diff
@@ -42,7 +42,7 @@
 //!     patina_debugger::set_debugger(&DEBUGGER);
 //!
 //!     // Setup a custom monitor command for this platform.
-//!     patina_debugger::add_monitor_command("my_command", "Description of my_command", |args, writer| {
+//!     patina_debugger::add_monitor_command("my_command", "Description of my_command", true, |args, writer| {
```
makubacki: Was this supposed to have hidden set to `true`?

```suggestion
//!     patina_debugger::add_monitor_command("my_command", "Description of my_command", false, |args, writer| {
```

os-d:
@makubacki can you force merge this over the codecov complaint? I can't add any more tests. There's one uncovered line in the debugger and mostly this:

image

Which can't be tested. That function has other things going on that are tested, so I don't mark to mark it coverage off. I could move adding the monitor cmd to a separate function, but that seems odd.

core/patina_debugger/src/system/alloc.rs
```diff
@@ -200,7 +206,7 @@ mod tests {
         let callback: Box<MonitorCommandFn> = Box::new(|args, out| {
             let _ = writeln!(out, "Executed with args: {:?}", args.collect::<Vec<_>>());
         });
-        system_state.add_monitor_command(command, description, callback);
+        system_state.add_monitor_command(command, description, false, callback);
```
makubacki: I think providing named constants like `HIDDEN` and `VISIBLE` would be easier to read.

core/patina_debugger/src/lib.rs
os-d: This should be false as that's the usual case, thanks.

core/patina_debugger/src/system/alloc.rs
os-d: Sure, I'll change it

apop5 (approved):

core/patina_debugger/src/system/alloc.rs
```diff
@@ -39,16 +39,20 @@ impl SystemState {
         &mut self,
         command: &'static str,
         description: &'static str,
+        hidden: bool,
         callback: Box<MonitorCommandFn>,
     ) {
-        let monitor = MonitorCallback { command, description, callback };
+        let monitor = MonitorCallback { command, description, hidden, callback };
```
Javagedes: description is pointless if hidden is true. I would lean towards either checking description for an empty string, or making it a `Option<&'static str>` instead of using `hidden`.

os-d: That's fair, I'll switch it to an option

core/patina_debugger/src/system/alloc.rs
os-d: Not needed now that I dropped the hidden param

os-d:
@Javagedes @makubacki I addressed your feedback. Please re-review and if it looks good can we force merge since the patch coverage can't be increased over 80%?

makubacki:
> @makubacki can you force merge this over the codecov complaint?

I will force merge it after status checks are complete. Thanks for giving the rationale.

Javagedes (approved):

PR #1480, opened 2026-04-27
Patina: R-EFI 6.0 migration

### LINKED ISSUE #1235: [Task]: Make `core_free_pool()` and `core_free_pages()` unsafe

### LINKED ISSUE #1278: [Task]: Integrate r-efi 6.0.0 breaking changes - `unsafe fn` in all function prototypes
### Feature Overview

[r-efi v6.0.0](https://github.com/r-efi/r-efi/releases/tag/v6.0.0) is released. This is a major breaking change that updates function prototypes to be marked unsafe.

This needs to be integrated into Patina.

### Solution Overview

This integration was made to better reflect that safety of the code we use so that we can meaningfully and accurately track unsafe code boundaries. Therefore, this should be more than simply passing compilation and linter checks. We should think carefully about the relationships between functions and practical safety concerns, mitigating issues where possible and calling out relevant safety invariants in safety comments.

### Alternatives Considered

_No response_

### Urgency

Low

No maintainer feedback needed

### Anything else?

_No response_

### LINKED ISSUE #1298: [Task]: Update `"efiapi"` event, timer, and tpl functionality to be unsafe
### Request Description

As of r_efi 6.0.0, most `"efiapi"` function types have been marked unsafe. This tracks a sub-issue to make the following functions unsafe and appropriately comment on safety requirements:

Image

---

I have chosen to group these due by the UEFI spec functionality groupings. See https://uefi.org/specs/UEFI/2.9_A/07_Services_Boot_Services.html

I will make the change

No maintainer feedback needed

### Anything else?

_No response_

### LINKED ISSUE #1308: [Task]: Update `"efiapi"` memory allocation functions to be unsafe
### Request Description

As of r_efi 6.0.0, most `"efiapi"` function types have been marked unsafe. This tracks a sub-issue to make the following functions unsafe and appropriately comment on safety requirements:

Image

---

I have chosen to group these due by the UEFI spec functionality groupings. See https://uefi.org/specs/UEFI/2.9_A/07_Services_Boot_Services.html

I will make the change

No maintainer feedback needed

### Anything else?

_No response_

### LINKED ISSUE #1309: [Task]: Update `"efiapi"` protocol functions to be unsafe
### Request Description

As of r_efi 6.0.0, most `"efiapi"` function types have been marked unsafe. This tracks a sub-issue to make the following functions unsafe and appropriately comment on safety requirements:

Image
---

I have chosen to group these due by the UEFI spec functionality groupings. See https://uefi.org/specs/UEFI/2.9_A/07_Services_Boot_Services.html

I will make the change

No maintainer feedback needed

### Anything else?

_No response_

vineelko:
## Description

R-EFI 6.0 Migration

Upstream r-efi 6.0 marks all `extern "efiapi"` function pointers in EFI Boot
Services, Runtime Services, and other protocols as `unsafe`, since their safety
cannot be enforced by the Rust type system at compile time.

This PR audits almost all usage of UEFI service function pointers across the Patina
codebase for correctness and safety.

- There are three primary areas where `unsafe` usage has been audited:
  1. The FFI boundary - all `extern "efiapi"` functions
  2. The Patina wrappers - all functions within the `BootServices` trait that
     eventually call into the FFI
  3. Core Patina routines that implement the FFI functions

- Not all functions in the above categories need to be marked `unsafe`. For
  example, `extern "efiapi" close_event` is not marked `unsafe` because it can
  safely be called with an arbitrary event parameter without causing undefined
  behavior in the Patina implementation. The `event` parameter is treated as an
  opaque pointer and is never dereferenced.

  That said, any indirect caller that dereferences this Boot Services function
  pointer must still use an `unsafe` block, since the function pointer itself is
  defined as `unsafe` in r-efi 6.0. In addition, inherently unsafe Rust
  functions (such as `core_free_pool()`) are now explicitly marked `unsafe`.

- Each inspected function or call site is documented with appropriate safety
  comments where necessary, and with explanations where `unsafe` is not
  required.

- There are no functional changes.

- Clippy flags public functions that accept raw pointer parameters and pass them
  across an FFI boundary without being marked `unsafe`, with the following
  warning:
  "this public function might dereference a raw pointer but is not marked `unsafe`"

  ```
  unsafe extern "C" {
      fn some_ffi_call(ptr: *mut i32);
  }

  // Triggers clippy::not_unsafe_ptr_arg_deref:
  // - public function
  // - not marked `unsafe`
  // - raw pointer parameter `ptr` is passed into an `unsafe` block
  pub fn example(ptr: *mut i32) {
      unsafe { some_ffi_call(ptr) };
  }
  ```

  In Patina, this pattern is common for functions that take `efi::Event` or
  `efi::Handle` parameters and call Boot Services function pointers. We
  explicitly suppress this lint for such functions because these types are
  treated as opaque pointers and are never dereferenced:
  `#[allow(clippy::not_unsafe_ptr_arg_deref)]`

Geiger Unsafe Stats
---------------------
```
|             | x86_64-unknown-uefi(before) | x86_64-unknown-uefi(after) | aarch64-unknown-uefi(before) | aarch64-unknown-uefi(after) |
|-------------|-----------------------------|----------------------------|------------------------------|-----------------------------|
| overall     | 11.90% (yellow)             | 16.30% (red)               | 12.20% (yellow)              | 16.30% (red)                |
| functions   | 2.60%  (green)              | 11.40% (yellow)            | 2.80%  (green)               | 11.40% (yellow)             |
| exprs       | 12.40% (yellow)             | 17.00% (red)               | 12.60% (yellow)              | 17.00% (red)                |
| item_impls  | 12.50% (yellow)             | 13.80% (yellow)            | 13.80% (yellow)              | 13.80% (yellow)             |
| item_traits | 12.10% (yellow)             | 13.00% (yellow)            | 13.00% (yellow)              | 13.00% (yellow)             |
| methods     | 5.60%  (green)              | 6.60%  (green)             | 5.60%  (green)               | 6.60%  (green)              |
```

## How This Was Tested

Booted to UEFI Shell Q35/SBSA

## Integration Instructions

All consumers of Patina will need to pick the newer Patina version when published and also should update their r-efi version to 6.0.0.

### DIFF
--- a/patina_dxe_core/src/allocator.rs
+++ b/patina_dxe_core/src/allocator.rs
@@ -784,7 +784,10 @@ pub fn core_free_pages(memory: efi::PhysicalAddress, pages: usize) -> Result<(),
     res
 }
 
-extern "efiapi" fn copy_mem(destination: *mut c_void, source: *mut c_void, length: usize) {
+/// # Safety
+///
+/// The caller must ensure that `destination` and `source` are valid pointers for `length` bytes
+unsafe extern "efiapi" fn copy_mem(destination: *mut c_void, source: *mut c_void, length: usize) {
     if length == 0 {
         return;
     }
@@ -798,7 +801,11 @@ extern "efiapi" fn copy_mem(destination: *mut c_void, source: *mut c_void, lengt
     unsafe { core::ptr::copy(source as *mut u8, destination as *mut u8, length) }
 }
 
-extern "efiapi" fn set_mem(buffer: *mut c_void, size: usize, value: u8) {
+/// # Safety
+///
+/// The caller must ensure that `buffer` is a valid pointer to a contiguous region of at least
+/// `size` bytes.
+unsafe extern "efiapi" fn set_mem(buffer: *mut c_void, size: usize, value: u8) {
     if size == 0 {
         return;
     }
@@ -1841,34 +1848,37 @@ mod tests {
     fn copy_mem_should_copy_mem() {
         let mut dest = vec![0xa5u8; 0x10];
         let mut src = vec![0x5au8; 0x10];
-        copy_mem(dest.as_mut_ptr() as *mut c_void, src.as_mut_ptr() as *mut c_void, 0x10);
+        // SAFETY: The passed in values are safe because they are constructed in this test case.
+        unsafe { copy_mem(dest.as_mut_ptr() as *mut c_void, src.as_mut_ptr() as *mut c_void, 0x10) };
         assert_eq!(dest, src);
     }
 
     #[test]
     fn copy_mem_zero_length_is_noop_even_with_null() {
-        // A zero-length copy is a valid no-op
-        copy_mem(core::ptr::null_mut(), core::ptr::null_mut(), 0);
+        // SAFETY: A zero-length copy is a valid no-op, even with null pointers.
+        unsafe { copy_mem(core::ptr::null_mut(), core::ptr::null_mut(), 0) };
     }
 
     #[test]
     fn set_mem_should_set_mem() {
         let mut dest = vec![0xa5u8; 0x10];
-        set_mem(dest.as_mut_ptr() as *mut c_void, 0x10, 0x00);
+        // SAFETY: The passed in values are safe because they are constructed in this test case.
+        unsafe { set_mem(dest.as_mut_ptr() as *mut c_void, 0x10, 0x00) };
         assert_eq!(dest, vec![0x00u8; 0x10]);
     }
 
     #[test]
     fn set_mem_fills_bytes() {
         let mut buf: [u8; 8] = [0; 8];
-        set_mem(buf.as_mut_ptr() as *mut c_void, buf.len(), 0xAB);
+        // SAFETY: The buffer is valid for writes of its length.
+        unsafe { set_mem(buf.as_mut_ptr() as *mut c_void, buf.len(), 0xAB) };
         assert_eq!(buf, [0xAB; 8]);
     }
 
     #[test]
     fn set_mem_zero_size_is_noop_even_with_null() {
-        // A zero-size fill is a valid no-op
-        set_mem(core::ptr::null_mut(), 0, 0xFF);
+        // SAFETY: A zero-size fill is a valid no-op, even with a null pointer.
+        unsafe { set_mem(core::ptr::null_mut(), 0, 0xFF) };
     }
 
     #[test]
@@ -1877,7 +1887,8 @@ mod tests {
         let src: [u8; 4] = [1, 2, 3, 4];
         // When given a null destination with a non-zero length, the function either
         // asserts (if debug_assertions are enabled) or returns without dereferencing either pointer.
-        copy_mem(core::ptr::null_mut(), src.as_ptr() as *mut c_void, src.len());
+        // SAFETY: The function returns early on null pointers when debug_assertions are disabled.
+        unsafe { copy_mem(core::ptr::null_mut(), src.as_ptr() as *mut c_void, src.len()) };
     }
 
     #[test]
@@ -1886,7 +1897,8 @@ mod tests {
         let mut dst: [u8; 4] = [0xAA; 4];
         // When given a null source with a non-zero length, the function either
         // asserts (if debug_assertions are enabled) or returns without dereferencing either pointer.
-        copy_mem(dst.as_mut_ptr() as *mut c_void, core::ptr::null_mut(), dst.len());
+        // SAFETY: The function returns early on null pointers when debug_assertions are disabled.
+        unsafe { copy_mem(dst.as_mut_ptr() as *mut c_void, core::ptr::null_mut(), dst.len()) };
         assert_eq!(dst, [0xAA; 4]);
     }
 
@@ -1894,7 +1906,8 @@ mod tests {
     #[cfg(not(debug_assertions))]
     fn set_mem_null_buffer_returns_without_writing() {
         // A null buffer with a non-zero size returns without dereferencing.
-        set_mem(core::ptr::null_mut(), 4, 0xFF);
+        // SAFETY: The function returns early on null pointers when debug_assertions are disabled.
+        unsafe { set_mem(core::ptr::null_mut(), 4, 0xFF) };
     }
 
     #[test]
--- a/patina_dxe_core/src/allocator.rs
+++ b/patina_dxe_core/src/allocator.rs
@@ -772,14 +772,27 @@ pub fn memory_type_for_handle(handle: efi::Handle) -> Option<efi::MemoryType> {
     ALLOCATORS.lock().memory_type_for_handle(handle)
 }
 
-extern "efiapi" fn free_pages(memory: efi::PhysicalAddress, pages: usize) -> efi::Status {
-    match core_free_pages(memory, pages) {
+/// Frees pages previously allocated with [`allocate_pages`].
+///
+/// # Safety
+///
+/// `memory` must be a valid physical address corresponding to a previously allocated page range.
+/// The caller must have exclusive ownership of the memory being freed.
+unsafe extern "efiapi" fn free_pages(memory: efi::PhysicalAddress, pages: usize) -> efi::Status {
+    // SAFETY: The caller is responsible for ensuring `memory` is a valid, previously allocated address.
+    match unsafe { core_free_pages(memory, pages) } {
         Ok(_) => efi::Status::SUCCESS,
         Err(status) => status.into(),
     }
 }
 
-pub fn core_free_pages(memory: efi::PhysicalAddress, pages: usize) -> Result<(), EfiError> {
+/// Frees pages previously allocated with [`core_allocate_pages`].
+///
+/// # Safety
+///
+/// `memory` must be a valid physical address corresponding to a previously allocated page range.
+/// The caller must have exclusive ownership of the memory being freed.
+pub unsafe fn core_free_pages(memory: efi::PhysicalAddress, pages: usize) -> Result<(), EfiError> {
     let size = match pages.checked_mul(UEFI_PAGE_SIZE) {
         Some(size) => size,
         None => return Err(EfiError::InvalidParameter),
@@ -1824,7 +1837,8 @@ mod tests {
                 },
                 efi::Status::SUCCESS
             );
-            free_pages(buffer_ptr as u64, 0x10);
+            // SAFETY: `buffer_ptr` was successfully allocated above and points to a valid page range of 0x10 pages.
+            unsafe { free_pages(buffer_ptr as u64, 0x10) };
 
             //test successful allocate_address at the address that was just freed
             assert_eq!(
@@ -1839,7 +1853,8 @@ mod tests {
                 },
                 efi::Status::SUCCESS
             );
-            free_pages(buffer_ptr as u64, 0x10);
+            // SAFETY: `buffer_ptr` was successfully allocated above and points to a valid page range of 0x10 pages.
+            unsafe { free_pages(buffer_ptr as u64, 0x10) };
 
             //test successful allocate_max where max is greater than the address that was just freed.
             buffer_ptr = buffer_ptr.wrapping_add(0x11 * 0x1000);
@@ -1855,7 +1870,8 @@ mod tests {
                 },
                 efi::Status::SUCCESS
             );
-            free_pages(buffer_ptr as u64, 0x10);
+            // SAFETY: `buffer_ptr` was successfully allocated above and points to a valid page range of 0x10 pages.
+            unsafe { free_pages(buffer_ptr as u64, 0x10) };
 
             //test invalid allocation type
             assert_eq!(
@@ -1884,7 +1900,8 @@ mod tests {
                 },
                 efi::Status::SUCCESS
             );
-            free_pages(buffer_ptr as u64, 0x10);
+            // SAFETY: `buffer_ptr` was successfully allocated above and points to a valid page range of 0x10 pages.
+            unsafe { free_pages(buffer_ptr as u64, 0x10) };
             let allocators = ALLOCATORS.lock();
             let allocator = allocators.get_allocator(0x71234567).unwrap();
             let handle = allocator.handle();
@@ -1924,10 +1941,14 @@ mod tests {
     #[test]
     fn free_pages_error_scenarios_should_be_handled_properly() {
         with_locked_state(GcdInit::WithSize(0x1000000), |_physical_hob_list| {
-            assert_eq!(free_pages(0x12345000, !0xFFF), efi::Status::INVALID_PARAMETER);
-            assert_eq!(free_pages(!0xFFF, 0x10), efi::Status::INVALID_PARAMETER);
-            assert_eq!(free_pages(0x12345678, 1), efi::Status::INVALID_PARAMETER);
-            assert_eq!(free_pages(0x12345000, 1), efi::Status::NOT_FOUND);
+            // SAFETY: invalid page count to test overflow detection.
+            assert_eq!(unsafe { free_pages(0x12345000, !0xFFF) }, efi::Status::INVALID_PARAMETER);
+            // SAFETY: address near end of address space to test overflow detection.
+            assert_eq!(unsafe { free_pages(!0xFFF, 0x10) }, efi::Status::INVALID_PARAMETER);
+            // SAFETY: misaligned address to test alignment validation.
+            assert_eq!(unsafe { free_pages(0x12345678, 1) }, efi::Status::INVALID_PARAMETER);
+            // SAFETY: valid-looking but unallocated address to test not-found path.
+            assert_eq!(unsafe { free_pages(0x12345000, 1) }, efi::Status::NOT_FOUND);
         });
     }
 
--- a/patina_dxe_core/src/allocator.rs
+++ b/patina_dxe_core/src/allocator.rs
@@ -879,7 +879,15 @@ unsafe extern "efiapi" fn set_mem(buffer: *mut c_void, size: usize, value: u8) {
     }
 }
 
-extern "efiapi" fn get_memory_map(
+/// Returns the current UEFI memory map.
+///
+/// # Safety
+///
+/// `memory_map_size` must be a valid pointer to a `usize` indicating the buffer size.
+/// `memory_map` must point to a writable buffer of at least `*memory_map_size` bytes, or be null
+/// (in which case only the required size is returned). `map_key`, `descriptor_size`, and
+/// `descriptor_version` must each be valid pointers or null.
+unsafe extern "efiapi" fn get_memory_map(
     memory_map_size: *mut usize,
     memory_map: *mut efi::MemoryDescriptor,
     map_key: *mut usize,
@@ -2058,13 +2066,17 @@ mod tests {
             let mut map_key = 0;
             let mut descriptor_size = 0;
             let mut version = 0;
-            let status = get_memory_map(
-                core::ptr::addr_of_mut!(memory_map_size),
-                core::ptr::null_mut(),
-                core::ptr::addr_of_mut!(map_key),
-                core::ptr::addr_of_mut!(descriptor_size),
-                core::ptr::addr_of_mut!(version),
-            );
+            // SAFETY: all pointers are derived from local variables declared above and are valid.
+            // memory_map is null to perform a size query.
+            let status = unsafe {
+                get_memory_map(
+                    core::ptr::addr_of_mut!(memory_map_size),
+                    core::ptr::null_mut(),
+                    core::ptr::addr_of_mut!(map_key),
+                    core::ptr::addr_of_mut!(descriptor_size),
+                    core::ptr::addr_of_mut!(version),
+                )
+            };
             assert_eq!(status, efi::Status::BUFFER_TOO_SMALL);
             assert_ne!(memory_map_size, 0);
             assert_eq!(descriptor_size, core::mem::size_of::<efi::MemoryDescriptor>());
@@ -2082,13 +2094,17 @@ mod tests {
                 memory_map_size / descriptor_size
             ];
 
-            let status = get_memory_map(
-                core::ptr::addr_of_mut!(memory_map_size),
-                memory_map_buffer.as_mut_ptr(),
-                core::ptr::addr_of_mut!(map_key),
-                core::ptr::addr_of_mut!(descriptor_size),
-                core::ptr::addr_of_mut!(version),
-            );
+            // SAFETY: all pointers are derived from local variables. memory_map_buffer is
+            // properly sized from the previous size query.
+            let status = unsafe {
+                get_memory_map(
+                    core::ptr::addr_of_mut!(memory_map_size),
+                    memory_map_buffer.as_mut_ptr(),
+                    core::ptr::addr_of_mut!(map_key),
+                    core::ptr::addr_of_mut!(descriptor_size),
+                    core::ptr::addr_of_mut!(version),
+                )
+            };
             assert_eq!(status, efi::Status::SUCCESS);
             assert_eq!(descriptor_size, core::mem::size_of::<efi::MemoryDescriptor>());
             assert_eq!(version, 1);
@@ -2126,13 +2142,17 @@ mod tests {
                 .expect("Failed to find runtime allocation.");
 
             //get_memory_map with null size should return invalid parameter
-            let status = get_memory_map(
-                core::ptr::null_mut(),
-                memory_map_buffer.as_mut_ptr(),
-                core::ptr::addr_of_mut!(map_key),
-                core::ptr::addr_of_mut!(descriptor_size),
-                core::ptr::addr_of_mut!(version),
-            );
+            // SAFETY: memory_map_size is intentionally null to test the error path.
+            // All other pointers are valid local variables.
+            let status = unsafe {
+                get_memory_map(
+                    core::ptr::null_mut(),
+                    memory_map_buffer.as_mut_ptr(),
+                    core::ptr::addr_of_mut!(map_key),
+                    core::ptr::addr_of_mut!(descriptor_size),
+                    core::ptr::addr_of_mut!(version),
+                )
+            };
             assert_eq!(status, efi::Status::INVALID_PARAMETER);
         })
     }
@@ -2175,13 +2195,17 @@ mod tests {
             let mut map_key = 0;
             let mut descriptor_size = 0;
             let mut version = 0;
-            let status = get_memory_map(
-                core::ptr::addr_of_mut!(memory_map_size),
-                core::ptr::null_mut(),
-                core::ptr::addr_of_mut!(map_key),
-                core::ptr::addr_of_mut!(descriptor_size),
-                core::ptr::addr_of_mut!(version),
-            );
+            // SAFETY: all pointers are derived from local variables declared above and are valid.
+            // memory_map is null to perform a size query.
+            let status = unsafe {
+                get_memory_map(
+                    core::ptr::addr_of_mut!(memory_map_size),
+                    core::ptr::null_mut(),
+                    core::ptr::addr_of_mut!(map_key),
+                    core::ptr::addr_of_mut!(descriptor_size),
+                    core::ptr::addr_of_mut!(version),
+                )
+            };
             assert_eq!(status, efi::Status::BUFFER_TOO_SMALL);
 
             let mut memory_map_buffer: Vec<efi::MemoryDescriptor> = vec![
@@ -2195,13 +2219,17 @@ mod tests {
                 memory_map_size / descriptor_size
             ];
 
-            let status = get_memory_map(
-                core::ptr::addr_of_mut!(memory_map_size),
-                memory_map_buffer.as_mut_ptr(),
-                core::ptr::addr_of_mut!(map_key),
-                core::ptr::addr_of_mut!(descriptor_size),
-                core::ptr::addr_of_mut!(version),
-            );
+            // SAFETY: all pointers are derived from local variables. memory_map_buffer is
+            // properly sized from the previous size query.
+            let status = unsafe {
+                get_memory_map(
+                    core::ptr::addr_of_mut!(memory_map_size),
+                    memory_map_buffer.as_mut_ptr(),
+                    core::ptr::addr_of_mut!(map_key),
+                    core::ptr::addr_of_mut!(descriptor_size),
+                    core::ptr::addr_of_mut!(version),
+                )
+            };
             assert_eq!(status, efi::Status::SUCCESS);
 
             assert!(terminate_memory_map(map_key).is_ok());
--- a/patina_dxe_core/src/allocator.rs
+++ b/patina_dxe_core/src/allocator.rs
@@ -589,7 +589,17 @@ impl AllocatorMap {
     }
 }
 
-extern "efiapi" fn allocate_pool(pool_type: efi::MemoryType, size: usize, buffer: *mut *mut c_void) -> efi::Status {
+/// Allocates pool memory of the specified type.
+///
+/// # Safety
+///
+/// `buffer` must be a pointer pointing to valid, writable memory for a `*mut c_void`.
+/// The pointer is null-checked, but validity of the referenced memory is the caller's responsibility.
+unsafe extern "efiapi" fn allocate_pool(
+    pool_type: efi::MemoryType,
+    size: usize,
+    buffer: *mut *mut c_void,
+) -> efi::Status {
     if buffer.is_null() {
         return efi::Status::INVALID_PARAMETER;
     }
@@ -604,6 +614,10 @@ extern "efiapi" fn allocate_pool(pool_type: efi::MemoryType, size: usize, buffer
     }
 }
 
+/// Allocates pool memory of the specified type.
+///
+/// This function is safe because all input parameters are value types and any internal unsafe
+/// operations are fully encapsulated with validated preconditions.
 pub fn core_allocate_pool(pool_type: efi::MemoryType, size: usize) -> Result<*mut c_void, EfiError> {
     // It is not valid to attempt to allocate these memory types
     if matches!(pool_type, efi::CONVENTIONAL_MEMORY | efi::PERSISTENT_MEMORY | efi::UNACCEPTED_MEMORY_TYPE) {
@@ -1594,32 +1608,38 @@ mod tests {
 
             // test that disallowed types cannot be allocated
             assert_eq!(
-                allocate_pool(efi::CONVENTIONAL_MEMORY, 0x1000, core::ptr::addr_of_mut!(buffer_ptr)),
+                // SAFETY: `buffer_ptr` is a valid local variable and its address points to writable memory.
+                unsafe { allocate_pool(efi::CONVENTIONAL_MEMORY, 0x1000, core::ptr::addr_of_mut!(buffer_ptr)) },
                 efi::Status::INVALID_PARAMETER
             );
 
             assert_eq!(
-                allocate_pool(efi::PERSISTENT_MEMORY, 0x1000, core::ptr::addr_of_mut!(buffer_ptr)),
+                // SAFETY: `buffer_ptr` is a valid local variable and its address points to writable memory.
+                unsafe { allocate_pool(efi::PERSISTENT_MEMORY, 0x1000, core::ptr::addr_of_mut!(buffer_ptr)) },
                 efi::Status::INVALID_PARAMETER
             );
 
             assert_eq!(
-                allocate_pool(efi::UNUSABLE_MEMORY, 0x1000, core::ptr::addr_of_mut!(buffer_ptr)),
+                // SAFETY: `buffer_ptr` is a valid local variable and its address points to writable memory.
+                unsafe { allocate_pool(efi::UNUSABLE_MEMORY, 0x1000, core::ptr::addr_of_mut!(buffer_ptr)) },
                 efi::Status::SUCCESS
             );
 
             assert_eq!(
-                allocate_pool(efi::UNACCEPTED_MEMORY_TYPE, 0x1000, core::ptr::addr_of_mut!(buffer_ptr)),
+                // SAFETY: `buffer_ptr` is a valid local variable and its address points to writable memory.
+                unsafe { allocate_pool(efi::UNACCEPTED_MEMORY_TYPE, 0x1000, core::ptr::addr_of_mut!(buffer_ptr)) },
                 efi::Status::INVALID_PARAMETER
             );
 
             assert_eq!(
-                allocate_pool(efi::BOOT_SERVICES_DATA, 0x1000, core::ptr::addr_of_mut!(buffer_ptr)),
+                // SAFETY: `buffer_ptr` is a valid local variable and its address points to writable memory.
+                unsafe { allocate_pool(efi::BOOT_SERVICES_DATA, 0x1000, core::ptr::addr_of_mut!(buffer_ptr)) },
                 efi::Status::SUCCESS
             );
 
             assert_eq!(
-                allocate_pool(efi::BOOT_SERVICES_DATA, 0x2000000, core::ptr::null_mut()),
+                // SAFETY: null pointer is passed intentionally to validate the error path.
+                unsafe { allocate_pool(efi::BOOT_SERVICES_DATA, 0x2000000, core::ptr::null_mut()) },
                 efi::Status::INVALID_PARAMETER
             );
         });
@@ -1630,7 +1650,8 @@ mod tests {
         with_locked_state(GcdInit::WithSize(0x1000000), |_physical_hob_list| {
             let mut buffer_ptr = core::ptr::null_mut();
             assert_eq!(
-                allocate_pool(efi::BOOT_SERVICES_DATA, 0x1000, core::ptr::addr_of_mut!(buffer_ptr)),
+                // SAFETY: `buffer_ptr` is a valid local variable and its address points to writable memory.
+                unsafe { allocate_pool(efi::BOOT_SERVICES_DATA, 0x1000, core::ptr::addr_of_mut!(buffer_ptr)) },
                 efi::Status::SUCCESS
             );
 
--- a/patina_dxe_core/src/allocator.rs
+++ b/patina_dxe_core/src/allocator.rs
@@ -635,14 +635,31 @@ pub fn core_allocate_pool(pool_type: efi::MemoryType, size: usize) -> Result<*mu
     }
 }
 
-extern "efiapi" fn free_pool(buffer: *mut c_void) -> efi::Status {
-    match core_free_pool(buffer) {
+/// Frees pool memory previously allocated with [`allocate_pool`].
+///
+/// # Safety
+///
+/// `buffer` must be a pointer pointing to valid, previously allocated memory
+/// from [`allocate_pool`]. This is needed to locate allocation signatures. The
+/// current implementation does not track allocations, so the caller is
+/// responsible for ensuring the pointer is valid.
+unsafe extern "efiapi" fn free_pool(buffer: *mut c_void) -> efi::Status {
+    // SAFETY: The caller is responsible for ensuring `buffer` points to a valid allocation.
+    match unsafe { core_free_pool(buffer) } {
         Ok(_) => efi::Status::SUCCESS,
         Err(status) => status.into(),
     }
 }
 
-pub fn core_free_pool(buffer: *mut c_void) -> Result<(), EfiError> {
+/// Frees pool memory previously allocated with [`core_allocate_pool`].
+///
+/// # Safety
+///
+/// `buffer` must be a pointer pointing to valid, previously allocated memory
+/// from [`core_allocate_pool`]. This is needed to locate allocation signatures.
+/// The current implementation does not track allocations, so the caller is
+/// responsible for ensuring the pointer is valid.
+pub unsafe fn core_free_pool(buffer: *mut c_void) -> Result<(), EfiError> {
     if buffer.is_null() {
         return Err(EfiError::InvalidParameter);
     }
@@ -1655,9 +1672,11 @@ mod tests {
                 efi::Status::SUCCESS
             );
 
-            assert_eq!(free_pool(buffer_ptr), efi::Status::SUCCESS);
+            // SAFETY: `buffer_ptr` was successfully allocated above and points to valid pool memory.
+            assert_eq!(unsafe { free_pool(buffer_ptr) }, efi::Status::SUCCESS);
 
-            assert_eq!(free_pool(core::ptr::null_mut()), efi::Status::INVALID_PARAMETER);
+            // SAFETY: null pointer is intentional to test the error path.
+            assert_eq!(unsafe { free_pool(core::ptr::null_mut()) }, efi::Status::INVALID_PARAMETER);
             //TODO: these cause non-unwinding panic which crashes the test even with "#[should_panic]".
             //assert_eq!(free_pool(buffer_ptr), efi::Status::INVALID_PARAMETER);
             //assert_eq!(free_pool(((buffer_ptr as usize) + 10) as *mut c_void), efi::Status::INVALID_PARAMETER);
--- a/patina_dxe_core/src/allocator.rs
+++ b/patina_dxe_core/src/allocator.rs
@@ -646,29 +646,53 @@ pub fn core_free_pool(buffer: *mut c_void) -> Result<(), EfiError> {
     }
 }
 
-extern "efiapi" fn allocate_pages(
+/// Allocates pages of the specified type and returns the address through the `memory` pointer.
+///
+/// # Safety
+///
+/// The caller is responsible for ensuring that `memory` points to writable memory for an
+/// `efi::PhysicalAddress`. When `allocation_type` is `ALLOCATE_MAX_ADDRESS` or `ALLOCATE_ADDRESS`,
+/// the value at `*memory` must also be initialized; for `ALLOCATE_ANY_PAGES` the input value is
+/// ignored per the UEFI specification.
+unsafe extern "efiapi" fn allocate_pages(
     allocation_type: efi::AllocateType,
     memory_type: efi::MemoryType,
     pages: usize,
     memory: *mut efi::PhysicalAddress,
 ) -> efi::Status {
-    match core_allocate_pages(allocation_type, memory_type, pages, memory, None) {
-        Ok(_) => efi::Status::SUCCESS,
+    if memory.is_null() {
+        return efi::Status::INVALID_PARAMETER;
+    }
+
+    // Per the UEFI spec, the input value at `*memory` is only consumed for `ALLOCATE_MAX_ADDRESS`
+    // and `ALLOCATE_ADDRESS`. For `ALLOCATE_ANY_PAGES`, the input is ignored, so we must not read
+    // from `*memory` (it may be uninitialized).
+    let address = match allocation_type {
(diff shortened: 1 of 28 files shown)


patina_dxe_core/src/pi_dispatcher/image.rs
```diff
@@ -651,32 +658,43 @@ impl ImageData {
...
             return Ok((buffer, false, device_handle, 0));
         }
 
-        if let Ok((buffer, device_handle)) =
+        // SAFETY: file_path is non-null (checked above) and valid per this function's safety contract.
+        let result = unsafe {
             get_file_buffer_from_load_protocol(efi::protocols::load_file::PROTOCOL_GUID, boot_policy, file_path)
-        {
+        };
+        if let Ok((buffer, device_handle)) = result {
```
os-d: nit: this could stay as one statement, i.e. saving the Result in result is not needed?

patina_dxe_core/src/pi_dispatcher/image.rs
```diff
@@ -796,23 +824,23 @@ impl<P: super::PlatformInfo> super::PiDispatcher<P> {
...
-    // * device_path - the file path from which the image is loaded.
-    // * source_buffer - if not null, pointer to the memory location containing the
-    //                   image to be loaded.
-    //  * source_size - size in bytes of source_buffer. ignored if source_buffer is
-    //                  null.
-    //  * image_handle - pointer to the returned image handle that is created on
-    //                   successful image load.
+    /// Loads the image specified by the device_path or source_buffer argument.
+    ///
+    /// See EFI_BOOT_SERVICES::LoadImage() API definition in UEFI spec for usage
```
os-d:
```suggestion
    /// See the EFI_BOOT_SERVICES::LoadImage() API definition in the UEFI spec for usage
```
nit

patina_dxe_core/src/pi_dispatcher/image.rs
```diff
@@ -796,23 +824,23 @@ impl<P: super::PlatformInfo> super::PiDispatcher<P> {
...
+    /// details.
+    ///
+    /// # Safety
```
os-d, line 808: This safety section reads as an arguments section. I think this can be kept as an arguments section (though for the efiapi pointing to the UEFI spec is also fine), but that we should write a more specific safety section targeted at what the caller must uphold.

patina_dxe_core/src/pi_dispatcher/image.rs
```diff
@@ -834,20 +862,37 @@ impl<P: super::PlatformInfo> super::PiDispatcher<P> {
...
-                Ok(handle) => (handle, efi::Status::SUCCESS),
-                Err(ImageStatus::AccessDenied) => (null_mut(), efi::Status::ACCESS_DENIED),
-                Err(ImageStatus::SecurityViolation(handle)) => (handle, efi::Status::SECURITY_VIOLATION),
-                Err(ImageStatus::LoadError(err)) => return err.into(),
-            };
+        // SAFETY: Caller must ensure that device_path (if non-null) points to a
+        // valid device path structure.
+        let result =
+            unsafe { Self::instance().load_image(boot_policy.into(), parent_image_handle, device_path, image) };
+        let (handle, status) = match result {
```
os-d, line 843: nit: this could stay as one statement, don't need result

patina_dxe_core/src/pi_dispatcher/image.rs
```diff
@@ -882,17 +927,18 @@ impl<P: super::PlatformInfo> super::PiDispatcher<P> {
...
+                // SAFETY: Invokes the entry point. The code behind this pointer
+                // is FFI, which is inherently unsafe. The caller is responsible
+                // for ensuring that `self` refers to a valid loaded image
+                // (mapped), and that `entry_point` points to valid executable
+                // code.
+                status = unsafe { entry_point(image_handle, system_table) };
+
+                // SAFETY: any variables with "Drop" routines that need to run
+                // need to be explicitly dropped before calling exit(). Since exit()
+                // effectively "longjmp"s back to StartImage(), rust automatic
```
os-d:
```suggestion
                // effectively "longjmp"s back to StartImage(), Rust automatic
```

This safety comment (and I know it existed before this PR) is for what this function must do before calling exit(), but it does not address the parameters of exit. I think this needs to be changed in two ways:
- Change the current statement from explaining the safety rules (those should be moved to exit()'s function safety comment) and instead explain how this function is guaranteeing that contract is met (i.e. something like, this function has already dropped all memory that needs to be freed).
- Add explanation here how the passed parameters are meeting the safety contract of exit() (image_handle is valid because...)

patina_dxe_core/src/pi_dispatcher/image.rs
```diff
@@ -937,20 +983,29 @@ impl<P: super::PlatformInfo> super::PiDispatcher<P> {
...
 
-    // Transfers control to the entry point of an image that was loaded by
-    // load_image. See EFI_BOOT_SERVICES::StartImage() API definition in UEFI spec
-    // for usage details.
-    // * image_handle - handle of the image to be started.
-    // * exit_data_size - pointer to receive the size, in bytes, of exit_data.
-    //                    if exit_data is null, this is parameter is ignored.
-    // * exit_data - pointer to receive a data buffer with exit data, if any.
+    /// Transfers control to the entry point of an image that was loaded by
+    /// load_image. See EFI_BOOT_SERVICES::StartImage() API definition in UEFI
```
os-d:
```suggestion
    /// load_image. See the EFI_BOOT_SERVICES::StartImage() API definition in the UEFI
```

patina_dxe_core/src/pi_dispatcher/image.rs
```diff
@@ -937,20 +983,29 @@ impl<P: super::PlatformInfo> super::PiDispatcher<P> {
...
+    /// spec for usage details.
+    ///
+    /// # Safety
```
os-d, line 961: Same comment here, lines 992-993 are safety info, but below that is just arg info, not the safety contract

patina_dxe_core/src/pi_dispatcher/image.rs
```diff
@@ -1128,21 +1214,34 @@ impl<P: super::PlatformInfo> super::PiDispatcher<P> {
...
 
-    // Terminates a loaded EFI image and returns control to boot services.
-    // See EFI_BOOT_SERVICES::Exit() API definition in UEFI spec for usage details.
-    // * image_handle - the handle of the currently running image.
-    // * exit_status - the exit status for the image.
-    // * exit_data_size - the size of the exit_data buffer, if exit_data is not
-    //                    null.
-    // * exit_data - optional buffer of data provided by the caller.
+    /// Terminates a loaded EFI image and returns control to boot services. See
+    /// EFI_BOOT_SERVICES::Exit() API definition in UEFI spec for usage details.
```
os-d:
```suggestion
    /// the EFI_BOOT_SERVICES::Exit() API definition in the UEFI spec for usage details.
```

patina_dxe_core/src/pi_dispatcher/image.rs
```diff
@@ -1128,21 +1214,34 @@ impl<P: super::PlatformInfo> super::PiDispatcher<P> {
...
+    ///
+    /// `image_handle` is validated against the private image data map; if not
+    /// found, `INVALID_PARAMETER` is returned.
+    ///
+    /// # Safety
```
os-d, line 1193: Same here. Consider having an argument section to split it from what the safety contract the caller must uphold is. This one does explain the safety contract but then also mixes in the arguments.

patina_dxe_core/src/allocator.rs
```diff
@@ -591,7 +591,17 @@ fn alloc_error_handler(layout: alloc::alloc::Layout) -> ! {
     panic!("allocation error: {:?}", layout)
 }
 
-extern "efiapi" fn allocate_pool(pool_type: efi::MemoryType, size: usize, buffer: *mut *mut c_void) -> efi::Status {
+/// Allocates pool memory of the specified type.
+///
+/// # Safety
+///
+/// `buffer` must be a valid pointer pointing to valid, writable memory for a `*mut c_void`.
+/// The pointer is null-checked, but validity of the referenced memory is the caller's responsibility.
```
os-d:
```suggestion
/// `buffer` must be a pointer pointing to valid, writable memory.
```

patina_dxe_core/src/allocator.rs
```diff
@@ -623,14 +637,29 @@ pub fn core_allocate_pool(pool_type: efi::MemoryType, size: usize) -> Result<*mu
...
 
-extern "efiapi" fn free_pool(buffer: *mut c_void) -> efi::Status {
-    match core_free_pool(buffer) {
+/// Frees pool memory previously allocated with [`allocate_pool`].
+///
+/// # Safety
+///
+/// `buffer` must be a valid pointer pointing to valid, previously allocated memory.
+/// This is needed to locate allocation signatures. The current implementation does not
+/// track allocations, so the caller is responsible for ensuring the pointer is valid.
```
os-d:
```suggestion
/// `buffer` must be point to valid readable memory that was previously allocated with [`allocate_pool`].
/// This is needed to locate allocation signatures.
```

patina_dxe_core/src/allocator.rs
```diff
@@ -623,14 +637,29 @@ pub fn core_allocate_pool(pool_type: efi::MemoryType, size: usize) -> Result<*mu
...
 }
 
-pub fn core_free_pool(buffer: *mut c_void) -> Result<(), EfiError> {
+/// Frees pool memory previously allocated with [`core_allocate_pool`].
+///
+/// # Safety
+///
+/// `buffer` must be a valid pointer pointing to valid, previously allocated memory.
+/// This is needed to locate allocation signatures. The current implementation does not
+/// track allocations, so the caller is responsible for ensuring the pointer is valid.
```
os-d:
```suggestion
/// `buffer` must point to valid readable memory previously allocated with [`core_allocate_pool`].
/// This is needed to locate allocation signatures.
```

patina_dxe_core/src/allocator.rs
```diff
@@ -648,19 +677,34 @@ pub fn core_free_pool(buffer: *mut c_void) -> Result<(), EfiError> {
     }
 }
 
-extern "efiapi" fn allocate_pages(
+/// Allocates pages of the specified type and returns the address through the `memory` pointer.
+///
+/// # Safety
+///
+/// The caller is responsible for ensuring that `memory` points to valid, writable memory for a
+/// `efi::PhysicalAddress`. The pointer is null-checked, but validity of the referenced memory is
```
os-d:
```suggestion
/// The caller is responsible for ensuring that `memory` points to valid writable memory.
/// The pointer is null-checked, but validity of the referenced memory is
```

patina_dxe_core/src/allocator.rs
```diff
@@ -648,19 +677,34 @@ pub fn core_free_pool(buffer: *mut c_void) -> Result<(), EfiError> {
...
     }
 }
 
-pub fn core_allocate_pages(
+/// Allocates pages of the specified type and returns the address through the `memory` pointer.
+///
+/// # Safety
+///
+/// The caller is responsible for ensuring that `memory` points to valid, writable memory for a
+/// `efi::PhysicalAddress`. The pointer is null-checked, but validity of the referenced memory is
```
os-d:
```suggestion
/// The caller is responsible for ensuring that `memory` points to valid writable memory.
/// The pointer is null-checked, but validity of the referenced memory is
```

patina_dxe_core/src/allocator.rs
```diff
@@ -958,13 +1031,20 @@ fn process_hob_allocations(hob_list: &HobList) {
...
                             // so that we can track the allocation in the allocator
-                            GcdMemoryType::SystemMemory => core_allocate_pages(
-                                efi::ALLOCATE_ADDRESS,
-                                desc.memory_type,
-                                uefi_size_to_pages!(desc.memory_length as usize),
-                                &mut address as *mut efi::PhysicalAddress,
-                                None,
-                            ),
+
+                            // SAFETY: The address is described by the HOB and
```
os-d: The safety contract is only that the `memory` param is valid and writeable, so I think this safety comment should say that address is a local so we are guaranteed it is valid to use.

patina_dxe_core/src/allocator.rs
```diff
@@ -1078,15 +1158,20 @@ fn process_hob_allocations(hob_list: &HobList) {
...
-                efi::BOOT_SERVICES_DATA,
-                1,
-                &mut address as *mut efi::PhysicalAddress,
-                None,
-            )
-            .is_err()
-            {
+
+            // SAFETY: we are attempting to allocate page 0 for null pointer
+            // detection, This is valid expected usage.
```
os-d: Same here, the safety contract is about the `memory` param, not what we are attempting to allocate.

patina_dxe_core/src/memory_manager.rs
```diff
@@ -60,8 +60,11 @@ impl MemoryManager for CoreMemoryManager {
             }
         };
 
-        let result =
-            core_allocate_pages(alloc_type, options.memory_type().into(), page_count, &mut address, Some(alignment));
+        // SAFETY: The caller must ensure that the provided page_count is valid
```
os-d: This safety contract is just about the `memory` param pointing to valid writeable memory, as such this comment should just say that address is a local variable, so it is safe to pass the address.

patina_dxe_core/src/misc_boot_services.rs
```diff
@@ -62,7 +62,15 @@ static WATCHDOG_ARCH_PTR: ArchProtocolPtr<protocols::watchdog::Protocol> = ArchP
...
 pub const PRE_EBS_GUID: patina::BinaryGuid = patina::BinaryGuid::from_string("5F1D7E16-784A-4DA2-B084-F812F23A8DCE");
 // TODO [END]: LOCAL (TEMP) GUID DEFINITIONS (MOVE LATER)
-extern "efiapi" fn calculate_crc32(data: *mut c_void, data_size: usize, crc_32: *mut u32) -> efi::Status {
+
+/// Calculates the CRC32 of the given data buffer.
+///
+/// # Safety
+///
+/// The caller is responsible for ensuring that `data` points to valid, readable memory of at least
+/// `data_size` bytes and that `crc_32` points to valid, writable memory for a `u32`. Both pointers
```
os-d:
```suggestion
/// `data_size` bytes and that `crc_32` points to valid, writable memory. Both pointers
```

sdk/patina/src/runtime_services.rs
```diff
@@ -416,8 +424,12 @@ impl RuntimeServices for StandardRuntimeServices {
         // the appropriate size that the buffer should be resized to for the second call.
         let mut first_try: bool = true;
         loop {
-            let status =
-                get_next_variable_name(ptr::addr_of_mut!(next_name_size), next_name.as_mut_ptr(), next_namespace);
+            // SAFETY: It is the caller's responsibility to ensure that the prev_name is always
```
os-d, line 439: For the caller's responsibility we should add a function level safety comment

sdk/patina/src/boot_services.rs
```diff
@@ -308,7 +308,12 @@ pub trait BootServices {
     /// Frees memory pages.
     ///
     /// [UEFI Spec Documentation: 7.2.2. EFI_BOOT_SERVICES.FreePages()](https://uefi.org/specs/UEFI/2.10/07_Services_Boot_Services.html#efi-boot-services-freepages)
-    fn free_pages(&self, address: usize, nb_pages: usize) -> Result<(), efi::Status>;
+    ///
+    /// # Safety
+    ///
+    /// `address` must be a valid physical address corresponding to a previously allocated page range.
```
os-d:
```suggestion
    /// `address` must be a valid physical address corresponding to a previously allocated page range from [`allocate_pages`].
```

sdk/patina/src/boot_services.rs
```diff
@@ -329,7 +334,13 @@ pub trait BootServices {
     /// Returns pool memory to the system.
     ///
     /// [UEFI Spec Documentation: 7.2.5. EFI_BOOT_SERVICES.FreePool()](https://uefi.org/specs/UEFI/2.10/07_Services_Boot_Services.html#efi-boot-services-freepool)
-    fn free_pool(&self, buffer: *mut u8) -> Result<(), efi::Status>;
+    ///
+    /// # Safety
+    ///
+    /// `buffer` must be a valid pointer pointing to valid, previously allocated memory.
+    /// This is needed to locate allocation signatures. The current implementation does not
+    /// track allocations, so the caller is responsible for ensuring the pointer is valid.
```
os-d:
```suggestion
    /// `buffer` must be a valid pointer pointing to valid readable memory previously allocated from [`allocate_pool`].
    /// This is needed to locate allocation signatures.
```

sdk/patina/src/boot_services.rs
```diff
@@ -1006,21 +1056,25 @@ impl BootServices for StandardBootServices {
...
         // of boot services don't merit such complexity at this time.
         let create_event = unsafe { efi_boot_services_fn!(*self.as_mut_ptr(), create_event) };
-        let status = create_event(
-            event_type.into(),
-            notify_tpl.into(),
-            // Safety: Transmuting function pointer types with matching ABIs and compatible signatures.
-            // Both are extern "efiapi" callbacks taking a context pointer - the generic parameter T
-            // is erased to c_void for the UEFI FFI interface.
-            unsafe {
+        // SAFETY: In Rust 2024 Edition, the unsafe_op_in_unsafe_fn lint is enabled by default, so calling unsafe
```
os-d: This comment isn't needed, the actual safety comment below should be here

sdk/patina/src/boot_services.rs
```diff
@@ -1036,65 +1090,109 @@ impl BootServices for StandardBootServices {
...
         // SAFETY: See safety comment in create_event_unchecked for details on corner cases around external modifications.
         let create_event_ex = unsafe { efi_boot_services_fn!(*self.as_mut_ptr(), create_event_ex) };
-        let status = create_event_ex(
-            event_type.into(),
-            notify_tpl.into(),
-            // Safety: Transmuting function pointer types with matching ABIs and compatible signatures.
-            // Both are extern "efiapi" callbacks - the generic parameter T is erased to c_void for FFI.
-            unsafe {
+
+        // SAFETY: In Rust 2024 Edition, the unsafe_op_in_unsafe_fn lint is enabled by default, so calling unsafe
```
os-d: Same here

sdk/patina/src/boot_services.rs
```diff
@@ -1125,26 +1227,40 @@ impl BootServices for StandardBootServices {
...
         let allocate_pages = unsafe { efi_boot_services_fn!(*self.as_mut_ptr(), allocate_pages) };
-        match allocate_pages(
-            alloc_type.into(),
-            memory_type.into(),
-            nb_pages,
-            ptr::addr_of_mut!(memory_address) as *mut u64,
-        ) {
-            s if s.is_error() => Err(s),
+
+        // SAFETY: Although the current function `allocate_pages` is a safe Rust wrapper, the
```
os-d, line 1228: This function should be marked unsafe then, since it gets the address from the caller and we can't verify it

sdk/patina/src/boot_services.rs
```diff
@@ -1125,26 +1227,40 @@ impl BootServices for StandardBootServices {
...
         }
     }
 
-    fn free_pages(&self, address: usize, nb_pages: usize) -> Result<(), efi::Status> {
+    unsafe fn free_pages(&self, address: usize, nb_pages: usize) -> Result<(), efi::Status> {
         // SAFETY: See safety comment in create_event_unchecked for details on corner cases around external modifications.
         let free_pages = unsafe { efi_boot_services_fn!(*self.as_mut_ptr(), free_pages) };
-        match free_pages(address as u64, nb_pages) {
-            s if s.is_error() => Err(s),
+        // SAFETY: `address` must be a valid physical address corresponding to a previously allocated
```
os-d, line 1255: Need fn safety comment

sdk/patina/src/boot_services.rs
```diff
@@ -1191,21 +1316,29 @@ impl BootServices for StandardBootServices {
         let mut buffer = ptr::null_mut();
         // SAFETY: See safety comment in create_event_unchecked for details on corner cases around external modifications.
         let allocate_pool = unsafe { efi_boot_services_fn!(*self.as_mut_ptr(), allocate_pool) };
-        match allocate_pool(memory_type.into(), size, ptr::addr_of_mut!(buffer)) {
-            s if s.is_error() => Err(s),
+        // SAFETY: buffer is declared above, we pass the address which guarantees it is a valid pointer. Unsafe block is
+        // still needed because the API itself is declared unsafe in r-efi.
```
os-d:
```suggestion
        // SAFETY: buffer is declared above, we pass the address which guarantees it is a valid pointer.
```

sdk/patina/src/boot_services.rs
```diff
@@ -1191,21 +1316,29 @@ impl BootServices for StandardBootServices {
...
         }
     }
 
-    fn free_pool(&self, buffer: *mut u8) -> Result<(), efi::Status> {
+    unsafe fn free_pool(&self, buffer: *mut u8) -> Result<(), efi::Status> {
         // SAFETY: See safety comment in create_event_unchecked for details on corner cases around external modifications.
         let free_pool = unsafe { efi_boot_services_fn!(*self.as_mut_ptr(), free_pool) };
-        match free_pool(buffer as *mut c_void) {
-            s if s.is_error() => Err(s),
+        // SAFETY: The caller is responsible for ensuring `buffer` points to a valid allocation.
```
os-d, line 1339: fn safety comment needed

sdk/patina/src/boot_services.rs
```diff
@@ -1423,30 +1667,56 @@ impl BootServices for StandardBootServices {
         };
         // SAFETY: See safety comment in create_event_unchecked for details on corner cases around external modifications.
         let connect_controller = unsafe { efi_boot_services_fn!(*self.as_mut_ptr(), connect_controller) };
-        match connect_controller(controller_handle, driver_image_handles, remaining_device_path, recursive.into()) {
-            s if s.is_error() => Err(s),
+        // SAFETY: The caller must ensure the function's unsafe contract.
```
os-d:
```suggestion
        // SAFETY: The caller must ensure the function's safety contract.
```

os-d (approved):

makubacki:
I haven't looked at the PR changes yet, but it should target the `major` branch due to the severity of breaking changes.

vineelko:
> I haven't looked at the PR changes yet, but it should target the `major` branch due to the severity of breaking changes.

Michael, The branch is now updated to target major instead of main.

patina_dxe_core/src/pi_dispatcher/image.rs
vineelko: True it is not needed but the if/match statements are becoming unmanageable in some places. So, I prefer taking it out.

patina_dxe_core/src/pi_dispatcher/image.rs
vineelko: Same here I would prefer it this way to make it slightly more manageable especially since we have to write the safety comment over it.

patina_dxe_core/src/allocator.rs
vineelko: Fixed it.

patina_dxe_core/src/allocator.rs
vineelko: Fixed it.

patina_dxe_core/src/memory_manager.rs
vineelko: I will have to fix the comment regarding the page_count and make it more about the memory param contract. 

This one specifically, Eventhough the address is a local variable it is derived from the `options` argument which is not in our control. So, we should mark the `allocate_pages()` API on MemoryManager as unsafe.

sdk/patina/src/component/service/memory.rs
```diff
@@ -130,12 +134,23 @@ pub trait MemoryManager {
...
     /// }
     /// ```
     ///
-    fn allocate_pages(&self, page_count: usize, options: AllocationOptions) -> Result<PageAllocation, MemoryError>;
+    /// ## Safety
+    /// Caller must ensure that the allocation options are valid. If using
+    /// [`PageAllocationStrategy::Address`], the provided address must be a valid physical address
+    /// that is safe to use for the requested number of pages.
+    ///
+    unsafe fn allocate_pages(
```
joschock: I don't know if there has been discussion on this, but I think making this unsafe is a pretty big blast radius, esp. since the targeted address case is really a minority of usage. 

what if we left allocate_pages and friends as safe abstractions and instead moved the safety check down to with_strategy() where the caller has to guarantee the invariant
```rust
let allocation_options = AllocationOptions::new(); // default strategy is safe. 

// # Safety: I promise <foo> is good. 
let allocation_options_risky = unsafe { AllocationOptions::new().with_strategy(PageAllocationStrategy::Address(<foo>)) };

File: patina_dxe_core/src/pi_dispatcher/image.rs
Comment (joschock) [inline patina_dxe_core/src/pi_dispatcher/image.rs]: there is a lot of "unsafe poisoning" downstream of these raw pointers to device_path::Protocol. Could we instead make a newtype of these with an unsafe constructor so that the safety invariant is not spread everywhere?
```rust
struct DevicePathPtr(*mut efi::protocols::device_path::Protocol);
impl DevicePathPtr {
   /// Safety: Caller must guarantee that device path pointer is non-null and points to a valid device path.
   pub unsafe fn new(ptr: *mut efi::Protocols::device_path::Protocol) -> Self {
      Self(ptr)
   }
   pub fn as_raw(&mut self) -> *mut T {
     self.0
   }
}

pub load_image(
  &self,
  boot_policy: bool,
  parent_image_handle: efi::Handle, 
  file_path: DevicePathPtr
) -> Result<etc> {
...
}

// Safety: my_raw_path is a properly constructed valid device path with static lifetime (or whatever)
let file_path = unsafe { DevicePathPtr::new(my_raw_path); }
PI_DISPATCHER.load_image(boot_policy, parent_image_handle, file_path); // safe

File: patina_dxe_core/src/pi_dispatcher/image.rs [hunk 30]
```diff
@@ -834,20 +866,37 @@ impl<P: super::PlatformInfo> super::PiDispatcher<P> {
...
+    /// refer to a properly loaded image whose entry point is executable code.
+    /// It is the caller's responsibility to first load such an image via
+    /// [`Self::load_image`] before calling this function.
+    ///
+    /// # Safety
+    ///
+    /// The caller must ensure that `image_handle` was obtained from a prior
+    /// successful call to [`Self::load_image`] and that the loaded image's
+    /// memory and entry point remain valid.
+    pub unsafe fn start_image(&'static self, image_handle: efi::Handle) -> Result<(), efi::Status> {
```
joschock: I'm not sure this needs to be unsafe. If the image_handle is not valid, you can just return an error.  We can add additional validation (e.g. make sure that the handle is not only valid, but that it is actually associated with an image).  The caller can't really uphold the invariant of whether the loaded_images memory and entry point are valid; those are upheld by load_image implementation.

patina_dxe_core/src/pi_dispatcher/image.rs
```diff
@@ -937,20 +988,29 @@ impl<P: super::PlatformInfo> super::PiDispatcher<P> {
...
-    // for usage details.
-    // * image_handle - handle of the image to be started.
-    // * exit_data_size - pointer to receive the size, in bytes, of exit_data.
-    //                    if exit_data is null, this is parameter is ignored.
-    // * exit_data - pointer to receive a data buffer with exit data, if any.
+    /// Transfers control to the entry point of an image that was loaded by
+    /// load_image. See the EFI_BOOT_SERVICES::StartImage() API definition in
+    /// the UEFI spec for usage details.
+    ///
+    /// # Safety
```
joschock, line 961: The safety here is pretty much something like: "caller must uphold the UEFI spec contract with respect to the semantics of this API"

patina_dxe_core/src/allocator.rs
```diff
@@ -623,14 +637,29 @@ pub fn core_allocate_pool(pool_type: efi::MemoryType, size: usize) -> Result<*mu
...
         Err(status) => status.into(),
     }
 }
 
-pub fn core_free_pool(buffer: *mut c_void) -> Result<(), EfiError> {
+/// Frees pool memory previously allocated with [`core_allocate_pool`].
+///
+/// # Safety
+///
+/// `buffer` must be a pointer pointing to valid, previously allocated memory.
```
joschock: I would explicitly state that the memory has to be allocated with `core_allocate_pool` to be completely clear.

patina_dxe_core/src/allocator.rs
```diff
@@ -648,19 +677,32 @@ pub fn core_free_pool(buffer: *mut c_void) -> Result<(), EfiError> {
...
 }
 
-pub fn core_allocate_pages(
+/// Allocates pages of the specified type and returns the address through the `memory` pointer.
+///
+/// # Safety
+///
+/// The caller is responsible for ensuring that `memory` points to valid writable memory. The
+/// pointer is null-checked, but validity of the referenced memory is the caller's responsibility.
+pub unsafe fn core_allocate_pages(
```
joschock: Could we change the return type to `Result<efi::PhysicalAddress, EfiError>` and contain the raw pointer in the `extern "efiapi" allocate_pages` code?

patina_dxe_core/src/allocator.rs
```diff
@@ -790,20 +845,36 @@ pub fn core_free_pages(memory: efi::PhysicalAddress, pages: usize) -> Result<(),
     res
 }
 
-extern "efiapi" fn copy_mem(destination: *mut c_void, source: *mut c_void, length: usize) {
+/// # Safety
+///
+/// The caller must ensure that `destination` and `source` are valid pointers for `length` bytes
+/// and that the regions they point to do not violate Rust's aliasing rules for `core::ptr::copy`.
```
joschock: Since this is a UEFI spec function, I don't think the caller needs to be aware of or uphold rust's aliasing rules, unless it is a rust caller (in which case it should just use core::ptr::copy() directly).

patina_dxe_core/src/driver_services.rs
```diff
@@ -347,7 +364,34 @@ pub unsafe fn core_connect_controller(
...
+///   services. That means this routine can't guarantee that driver bindings
+///   remain valid for the duration of its execution. For example, if a driver
+///   were be unloaded in a timer callback after returning true from Supported()
+///   before Start() is called, then the driver binding instance would be
+///   uninstalled or invalid and Start() would be an invalid function pointer
+///   when invoked. In general, the spec implicitly assumes that driver binding
+///   instances that are valid at the start of the call to ConnectController()
+///   must remain valid for the duration of the ConnectController() call. If
+///   this is not true, then behavior is undefined. This function is marked
+///   unsafe for this reason.
```
joschock: I'm not sure the caller can directly uphold the "no weird edge cases exist in the platform where the driver is uninstalled between `Supported()` and `Start()`". 

Other safety conditions hold tho.

patina_dxe_core/src/memory_manager.rs
```diff
@@ -100,6 +112,12 @@ impl MemoryManager for CoreMemoryManager {
         Ok(allocator as &dyn core::alloc::Allocator)
     }
 
+    /// # Safety
+    ///
+    /// Changing tha attributes of a page of memory can result in undefined behavior
```
joschock:
```suggestion
    /// Changing the attributes of a page of memory can result in undefined behavior
```

sdk/patina/src/component/service/memory.rs
vineelko: Actually that sounds better. I think the same can be done for `allocate_pages()` in other places as well. Will fix it.

patina_dxe_core/src/allocator.rs
vineelko: Fixed it.

patina_dxe_core/src/driver_services.rs
vineelko: Agreed, I don't mean to make it as a caller's responsibility 😄. But I really want to document that edge case in the function comment so that the caller is aware of it. I will move it above the SAFETY comments.

patina_dxe_core/src/pi_dispatcher/image.rs
vineelko: I am prototyping safe `DevicePathPtr` around `*mut efi::protocols::device_path::Protocol`. This newtype looks very promising and resembles much like `NonNull<T>` type. And it feels like it is not limited to DevicePath and can be applicable to other raw pointer types which are used in Patina. 


```rust 
pub struct RawPtr<T>(*mut T);

impl<T> RawPtr<T> {
    pub unsafe fn new(ptr: *mut T) -> Self {
        Self(ptr)
    }

    pub unsafe fn from_raw(ptr: *mut T) -> Option<Self> {
        if ptr.is_null() {
            None
        } else {
            Some(unsafe { Self::new(ptr) })
        }
    }

    pub fn as_ptr(self) -> *mut T {
        self.0
    }
}
```

My only concern is that, the change will have far more impact beyond the scope of this PR. Especially, when `patina\sdk\patina\src\device_path\walker.rs` is also rewritten based on `DevicePathPtr`. So, I think it is better to pull in that change immediately after this PR. I can create a work item for tracking and can work on it.

patina_dxe_core/src/pi_dispatcher/image.rs
vineelko:
```
    /// Starts execution of a previously loaded image.
    ///
    /// The `image_handle` is validated against the protocol database and the
    /// private image data map; an error is returned if the handle is unknown or
    /// the image has already been started. However, a valid handle must still
    /// refer to a properly loaded image whose entry point is executable code.
    /// It is the caller's responsibility to first load such an image via
    /// [`Self::load_image`] before calling this function.
    ///
    /// # Safety
    ///
    /// The caller must ensure that `image_handle` was obtained from a prior
    /// successful call to [`Self::load_image`] and that the loaded image's
    /// memory and entry point remain valid.
```

My thought process was about the case where a malformed PE image has a valid entry point but contains invalid code (or something similar), which we cannot guarantee against. LoadImage will load the image because it satisfies the PE format requirements, but StartImage can still fail during execution.

patina_dxe_core/src/allocator.rs
vineelko: Fixed it

patina_dxe_core/src/pi_dispatcher/image.rs
joschock: I'm not sure how I feel about a universal `RawPtr` safe wrapper. That seems a bit generic? But I guess it would be unsafe at both point of creation and point of deref; so I guess it doesn't do too much to mask unsafety.

patina_dxe_core/src/pi_dispatcher/image.rs
joschock: I think concerns about a malformed image buffer should be on load_image. I think the promise chain goes: "This is a good image" -> load_image -> image_handle -> start_image. So start_image is basically relying on the fact that it is an image_handle to signify a promise that it was properly constructed. Practically speaking, if the image buffer is junk, you won't change the arguments to start_image, you'll change the arguments to load_image.

sdk/patina/src/component/service/memory.rs
joschock: I am granting the assumption that `PageAllocationStrategy::Address(<foo>)` is actually unsafe - 

I'm not sure it is, since we have the GCD/Allocator framework to deal with the memory map and qualify the address. The safety of all allocations in the system is predicated on the correctness of the map. If the map is correct, then PageAllocationStrategy::Address(<bogus value>) should always error, instead of doing something unsafe.

patina_dxe_core/src/allocator.rs
vineelko: Fixed it.

patina_dxe_core/src/pi_dispatcher/image.rs
vineelko: Agreed, but do you think this need not be marked unsafe? I thought the caller should be made aware of the exit_data and exit_data_size parameters explicitly.

joschock: Yeah, agree it should be unsafe around exit_data* parameters. This was more of an overall general statement about safety of UEFI spec APIs that happened to occur to me at this point of the review. In this specific case, I'd simply swap the order of the parameter info with the safety comment:

```rust
    /// * image_handle - handle of the image to be started.
    /// * exit_data_size - pointer to receive the size, in bytes, of exit_data.
    ///   if exit_data is null, this is parameter is ignored.
    /// * exit_data - pointer to receive a data buffer with exit data, if any.
    ///
    /// # Safety
    /// If `exit_data_size` and `exit_data` are non-null, they must point to
    /// valid writable memory. The caller owns the returned exit data buffer.

File: patina_dxe_core/src/pi_dispatcher/image.rs
Comment (vineelko) [reply patina_dxe_core/src/pi_dispatcher/image.rs]: John, pushed an update with DevicePathPtr. Please take a look at the last few commits.

File: sdk/patina/src/device_path/ptr.rs [hunk 37]
```diff
@@ -0,0 +1,55 @@
...
+// caller's responsibility; the pointer itself has no thread-affinity.
+unsafe impl Sync for DevicePathPtr {}
+
+impl DevicePathPtr {
+    /// Wraps a raw pointer as a `DevicePathPtr`.
+    ///
+    /// # Safety
+    ///
+    /// `ptr` must be non-null and must point to a valid, well-formed UEFI device path
+    /// node that remains valid for as long as the returned `DevicePathPtr` is used.
```
joschock: nit: I think in "Rust-ese" this would be
```suggestion
    /// node that remains valid for the lifetime of the returned `DevicePathPtr`.
```

patina_dxe_core/src/pi_dispatcher/image.rs
joschock: I think this looks good.

joschock (approved):

patina_dxe_core/src/allocator.rs
```diff
@@ -648,29 +679,43 @@ pub fn core_free_pool(buffer: *mut c_void) -> Result<(), EfiError> {
...
 ) -> efi::Status {
-    match core_allocate_pages(allocation_type, memory_type, pages, memory, None) {
-        Ok(_) => efi::Status::SUCCESS,
+    // SAFETY: The caller is responsible for ensuring `memory` is valid.
+    if memory.is_null() {
+        return efi::Status::INVALID_PARAMETER;
+    }
+
+    // SAFETY: caller must ensure that "memory" is a valid pointer. It is null-checked above.
+    let address = unsafe { memory.read_unaligned() };
```
makubacki: `memory` is treated differently depending on the allocation type.

https://uefi.org/specs/UEFI/2.11/07_Services_Boot_Services.html#efi-boot-services-allocatepages

For `AllocateAnyPages`, the address pointed to by `memory` is supposed to be ignored on input.

Can you please read through the API constraints in the spec and update this implementation to be compliant?

patina_dxe_core/src/driver_services.rs
```diff
@@ -389,23 +437,21 @@ extern "efiapi" fn connect_controller(
...
+/// call. If this is not true, then behavior is undefined.
 ///
 /// ## Example
 ///
 /// ```ignore
 /// let result = core_disconnect_controller(controller_handle, None, None);
 /// ```
 ///
-pub unsafe fn core_disconnect_controller(
+pub fn core_disconnect_controller(
```
makubacki: I notice that you changed `core_disconnect_controller()` from `unsafe` to safe, and I understand the reasoning, given the platform/environment uncertainties during disconnect versus conditions controlled by the caller for connect. While the nuance is fine, at the function level, the asymmetry is more apparent by the fact that `core_connect_controller()` remains `unsafe`.

Can you please add a brief clarifying comment about this in the `core_disconnect_controller()` doc comment?

sdk/patina/src/device_path/ptr.rs
```diff
@@ -0,0 +1,55 @@
...
+    pub unsafe fn new(ptr: *mut Protocol) -> Self {
+        // SAFETY: Caller guarantees ptr is non-null and valid.
+        debug_assert!(!ptr.is_null());
+        Self(ptr)
+    }
```
makubacki: What's the difference betwen this and just directly using `NonNull<Protocol>`?

---

From the function safety comment:

> `ptr` must be non-null and must point to a valid, well-formed UEFI device path
> node that remains valid for as long as the returned `DevicePathPtr` is used.

`DevicePathPtr::new()` let's this be violated in cases where debug asserts are not enabled:

```rust
    pub unsafe fn new(ptr: *mut Protocol) -> Self {
        // SAFETY: Caller guarantees ptr is non-null and valid.
        debug_assert!(!ptr.is_null());
        Self(ptr)
    }
```

Versus an implementation like that in [`NonNull::new()`](https://doc.rust-lang.org/src/core/ptr/non_null.rs.html#269):

```suggestion
    pub unsafe fn new(ptr: *mut Protocol) -> Option<Self> {
       if !ptr.is_null() {
            // SAFETY: The pointer is already checked and is not null
            Some(unsafe { Self::new_unchecked(ptr) })
        } else {
            None
        }
    }
```

_(the example above references a non-existing `DevicePathPtr::new_unchecked(ptr)` function)_

---

My other concerns is the constructor just stores the pointer. A pointer storage operation can trivially verify that the pointer is non-null but the "validity" part of this really applies to dereference time.

All of this gives the impression that `DevicePathPtr` promises more safety than is actually delivered which might end up complicating safe handling of device paths overall. For example, I saw other code in image.rs leaning on safety assumptions of the type.

---

If this stays mostly as it is, could it just be a transparent wrapper around `NonNull`. Something like:

```rust
#[repr(transparent)]
#[derive(Copy, Clone, Debug)]
pub struct DevicePathPtr(NonNull<Protocol>);
//...
impl DevicePathPtr {
    pub unsafe fn new_unchecked(ptr: *mut Protocol) -> Self {
        // SAFETY: caller guarantees non-null.
        Self(unsafe { NonNull::new_unchecked(ptr) })
    }

    /// Safe constructor. Returns `None` on null. No unsafe needed.
    pub fn new(ptr: *mut Protocol) -> Option<Self> {
        NonNull::new(ptr).map(Self)
    }

    pub fn as_ptr(self) -> *mut Protocol { self.0.as_ptr() }

    /// Reads the device path node header.
    ///
    /// # Safety
    /// The pointee must be a valid, well-formed UEFI device path node and remain
    /// live for the duration of the read.
    pub unsafe fn read_header(self) -> Protocol { /* ... */ }

// ...
}
```

This simplifies implementation reusing `NonNull` logic, focuses `new()` on storage and the nullness check, and moves header reads to be in their own `unsafe` blocks in callers where they have to uphold the safety invariant at the time it is relevant.

But, really, it seems to me like just using `NonNull<Protocol>` (this type removed) wouldn't lose much, be a little more clear, and reduce maintenance.

patina_dxe_core/src/allocator.rs
```diff
@@ -962,7 +1025,7 @@ fn process_hob_allocations(hob_list: &HobList) {
                                 efi::ALLOCATE_ADDRESS,
                                 desc.memory_type,
                                 uefi_size_to_pages!(desc.memory_length as usize),
-                                &mut address as *mut efi::PhysicalAddress,
+                                address as efi::PhysicalAddress,
```
makubacki: You shouldn't need to cast `address` here.

```suggestion
                                address,
```

patina_dxe_core/src/protocols.rs
```diff
@@ -77,6 +92,10 @@ extern "efiapi" fn install_protocol_interface(
     efi::Status::SUCCESS
 }
 
+/// Uninstalls a protocol interface on a handle.
+///
+/// This function is safe because `interface` is an opaque pointer used for
+/// comparision but never dereferenced. All other parameters are value types.
```
makubacki:
```suggestion
/// comparison but never dereferenced. All other parameters are value types.
```

patina_dxe_core/src/protocols.rs
```diff
@@ -50,7 +57,15 @@ pub fn core_install_protocol_interface(
...
 
-extern "efiapi" fn install_protocol_interface(
+/// Installs a protocol interface on a handle.
+///
+/// # Safety
+///
+/// `handle` must be a valid pointer to an `efi::Handle` (may point to a null handle for new handle
+/// creation). `protocol` must be a valid pointer to an `efi::Guid`. Both are null checked, but
+/// validity of the referenced memory is the caller's responsibility. Throughout the lifetime of the
+/// interface reference, the caller must ensure it remains valid.
```
makubacki: Can you clarify this to specifically call out the lifetime of the protocol database (e.g. until uninstalled)? That will take the explanation from generic to more precise.

sdk/patina/src/boot_services.rs
```diff
@@ -1561,22 +1885,41 @@ impl BootServices for StandardBootServices {
                         self,
                     )
                 });
-                Err((s, data))
+                Err((status, data))
             }
             _ => Ok(()),
         }
```
makubacki: I'm not sure that this is doing what is intended.

- I believe `exit_data.as_ptr()` is returning the address of `exit_data` on the stack (not the contained value).
- So, `exit_data.as_ptr().is_null()` is always false because the stack address is not null.
  - Then, in here:

  ```rust
                      BootServicesBox::from_raw_parts_mut(
                        exit_data.as_mut_ptr() as *mut u8,
                        exit_data_size.assume_init(),
                        self,
                    )
  ```

The same thing happens. `BootServicesBox` is built from the address of the local stack variable with the correct `exit_data_size` value.

Also, are you sure `exit_data.assume_init()` is valid? The write to it in `start_image_efiapi()` is conditionalized.

My understanding is you can clearly get the exit data pointer and size and then do the null check and box construction. Like (not tested):

```rust
let exit_data_ptr = unsafe { exit_data.assume_init() };
let exit_data_len = unsafe { exit_data_size.assume_init() };
let data = (!exit_data_ptr.is_null() && exit_data_len > 0).then(|| unsafe {
    BootServicesBox::from_raw_parts_mut(exit_data_ptr as *mut u8, exit_data_len, self)
});
```

sdk/patina/src/boot_services.rs
```diff
@@ -1509,39 +1804,64 @@ impl BootServices for StandardBootServices {
...
     }
 
+    /// # Safety
+    ///
+    /// If `device_path` is `Some`, it must wrap a valid UEFI device path in readable memory.
+    /// If `source_buffer` is provided, its contents must be a valid PE/COFF image.
+    // This is triggered by the fact that efi::Handle aliases to *mut c_void, but
+    // it is an opaque handle used as a database key.
+    #[allow(clippy::not_unsafe_ptr_arg_deref)]
     fn load_image(
```
makubacki, line 1825: `load_image()` is marked safe but still has a function safety comment. This is confusing/not enforceable.

sdk/patina/src/boot_services.rs
```diff
@@ -1586,57 +1929,98 @@ impl BootServices for StandardBootServices {
...
+    ///
+    /// This function is not marked `unsafe` because both parameters are safe to accept from
+    /// unverified sources:
+    /// - `image_handle` is passed through to the firmware which validates it internally.
+    /// - `map_key` is validated against the current memory map key; an invalid key causes the call
+    ///   to fail gracefully with an error status rather than undefined behavior.
+    // This is triggered by the fact that efi::Event aliases to *mut c_void, but
+    // it is an opaque handle used as a database key.
+    #[allow(clippy::not_unsafe_ptr_arg_deref)]
     fn exit_boot_services(&self, image_handle: efi::Handle, map_key: usize) -> Result<(), efi::Status> {
```
makubacki, line 1958: Since there already is a "Safety Considerations" section, can you please include that the entire Boot Services table is no longer callable after this function successfully completes?

sdk/patina/src/boot_services.rs
vineelko: Wow that's clearly a bug! Let me fix it properly. Original code incorrectly assumed `as_ptr()` instead of `assume_init()`.

sdk/patina/src/device_path/ptr.rs
vineelko: I originally had similar impression [here](https://github.com/OpenDevicePartnership/patina/pull/1480#discussion_r3263231651) when working on `DevicePathPtr` as it somewhat resembled `NonNull` with slightly different semantics. I am fine with removing this indirection and directly relying on `NonNull` for more clarity.

patina_dxe_core/src/allocator.rs
vineelko: Updated the implementation based on the spec.

patina_dxe_core/src/driver_services.rs
vineelko: This one feel bit odd to me, because marking the core_connect_controller() as `unsafe` for its undefined behavior not caused by the current inputs(but purely based on spec assumptions). This cannot be guaranteed by the caller. So, even though we have a safety comment explaining the scenario it feels like that function is marked unsafe incorrectly. In any case, we should have marked both `unsafe` or both `safe`. For now, I am marking the core_disconnect_controller() also as unsafe as similar undefined behavior can happen in disconnect call. I am open to change both of them otherwise.

sdk/patina/src/boot_services.rs
vineelko: Fixed it.

patina_dxe_core/src/driver_services.rs
makubacki: Yeah, my comment here is specifically about clearly explaining the decision made regardless of the decision itself.

Regarding the decision, my take is:

Handles are disjoint from the type system but should be able to be safely verified by the protocol database implementation. Since `core_disconnect_controller()` exclusively takes three handles, I thought it was reasonable to be safe. Implementation-aware safety invariants can be considered separately from UEFI-specification environmental dangers.

Because `core_connect_controller()` takes handles + a raw pointer (optional, to a protocol), I thought it was reasonable to be unsafe. The pointer can be checked for being non-null simple enough, but there is a lifetime aspect to its protocol interface buffer being valid throughout the execution of `core_connect_controller()` that is, at least partially, the caller's responsibility and cannot be verified otherwise than the safety contract.

sdk/patina/src/device_path/ptr.rs
makubacki: I think sticking with `NonNull<T>` works best right now. If we had more capabilities for `DevicePathPtr`, a new type might be justified. In the meantime, `NonNull<T>` accomplishes this in a simple, idiomatic way.

patina_dxe_core/src/driver_services.rs
vineelko: Updated the function comments to explain why they are being marked unsafe. The lifetime aspect of the protocol interface also seems to apply for the core_disconnect_controller() because it internally calls `DriverBinding->Stop()`. Hence keeping both of them as unsafe(Even though the inputs are exclusively handles which can be validated against the ProtocolDB).

makubacki (approved):

PR #1488, opened 2026-04-30
[feature/patina-boot] patina_boot: Add NVMe boot-partition write-lock helper

kat-perez:
## Description

Adds `lock_partition_write(boot_services, device_path)` to a new `patina_boot::partition` module. Looks up the NVMe Pass-Thru protocol on the device path and sets the Boot Partition Write Protection feature to "Write Protect Until Power Cycle" on both boot partitions. Lock clears on controller reset or power cycle.

Partition-I/O helpers live in their own module (`partition.rs`) rather than the existing `helpers.rs` so that file doesn't accumulate disparate concerns.

Closes OpenDevicePartnership/odp-platform-common#61.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- `cargo make all`

## Integration Instructions

- N/A


components/patina_boot/src/partition.rs
```diff
@@ -0,0 +1,363 @@
...
+/// Separated from `lock_partition_write` because it dereferences a raw protocol pointer and
+/// invokes its function pointer directly. Tests use mock protocol function pointers to exercise
+/// the dispatch path.
+///
+/// # Safety
+///
+/// `protocol` must be a valid, non-null pointer to an `EFI_NVM_EXPRESS_PASS_THRU_PROTOCOL`
+/// instance owned by an NVMe controller for the duration of this call.
+unsafe fn lock_partition_write_inner(protocol: *mut nvme_pass_thru::Protocol) -> Result<()> {
+    use nvme_pass_thru::{
```
os-d, line 68: I would like to wait to check this in until we’ve had a discussion on using protocols from Patina boot. It may be a necessary intermediate step, but we should consider what the right abstraction is, whether to accept using the protocol directly or have a service Patina boot consumes (which for now gets the protocol). Can we discuss in the next Patina meeting?

makubacki: We can discuss in the Patina meeting. However, I don't think we want `patina_boot` to directly implement storage abstraction details regardless of the underlying mechanism to do so (protocol or service).

components/patina_boot/src/partition.rs
makubacki, line 1: Setting the implementation in this PR, what is the underlying assumption when adding this to `patina_boot` on when the boot partition will be locked in the overall boot flow?

kat-perez:
Thanks both for the architectural feedback. Holding off on changes here pending the Patina meeting discussion — happy to bring this up.

Quick context for @makubacki's question on `partition.rs:1` (when the lock fires in the overall boot flow):

The intended consumer is `SreBootManager` (#1492), which calls the lock helper between the `EndOfDxe` signal and `ReadyToBoot` — i.e. after security lockdown and console discovery, but before any boot option is attempted. The goal is that nothing executing post-ReadyToBoot (including the OS loader and ultimately the OS itself) can modify the NVMe boot partition for the rest of the power cycle. State 001b (Write Protect Until Power Cycle) is chosen so a controller reset clears it on the next boot, allowing legitimate firmware updates to the BP via capsule apply on the next cold boot.

On the larger abstraction question: agree that `patina_boot` shouldn't directly own the storage protocol details long-term. Possible shapes I see for the meeting:

1. A `PartitionLocker` / `StorageService` trait in `patina_boot` (or `patina`), with an NVMe Pass-Thru concrete impl that lives elsewhere — `patina_boot` consumes the trait
2. Pull the NVMe Pass-Thru protocol bindings out of `patina_boot` into a dedicated storage component crate; `patina_boot` only consumes a generic "lock the boot partition" service
3. Move the binding upstream to `r-efi` (already filed as patina #1494) and let consumers compose the lock op without `patina_boot` knowing about NVMe

Happy to bring all three to the meeting and pick whichever fits the broader Patina architecture direction.

kat-perez:
Closing in favor of OpenDevicePartnership/patina-components#11, which moves this helper into the new `patina_nvme` crate per the reviewer thread on this PR (`patina_boot` shouldn't implement storage abstraction details).

PR #1490, opened 2026-04-30
Add EFI_RAM_DISK_PROTOCOL bindings

### LINKED ISSUE #65: Add EFI_RAM_DISK_PROTOCOL binding to patina SDK
Check whether `EFI_RAM_DISK_PROTOCOL` bindings exist in `sdk/patina` today. If not, add them. Required for installing the SRE WIM as a RAM disk.

Part of #51.

kat-perez:
## Description

Adds `sdk/patina/src/uefi_protocol/ram_disk.rs` providing the FFI binding for `EFI_RAM_DISK_PROTOCOL`. Exposes the `Protocol` struct, `RegisterFn` / `UnregisterFn` type aliases, and the four well-known `RamDiskType` GUIDs. `r-efi` doesn't provide this binding today, so it lands here.

Closes OpenDevicePartnership/odp-platform-common#65.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- `cargo make all`

## Integration Instructions

- N/A

### DIFF
--- a/sdk/patina/src/uefi_protocol/ram_disk.rs
+++ b/sdk/patina/src/uefi_protocol/ram_disk.rs
@@ -1,0 +1,100 @@
+//! UEFI RAM Disk Protocol bindings.
+//!
+//! Used to install a contiguous range of system memory as a virtual block device the system
+//! firmware can boot from.
+//!
+//! ## License
+//!
+//! Copyright (c) Microsoft Corporation.
+//!
+//! SPDX-License-Identifier: Apache-2.0
+//!
+use r_efi::efi;
+
+use crate::uefi_protocol::ProtocolInterface;
+
+/// Well-known `RamDiskType` GUID for a generic virtual disk.
+pub const RAM_DISK_VIRTUAL_DISK_GUID: efi::Guid =
+    efi::Guid::from_fields(0x77ab535a, 0x45fc, 0x624b, 0x55, 0x60, &[0xf7, 0xb2, 0x81, 0xd1, 0xf9, 0x6e]);
+
+/// Well-known `RamDiskType` GUID for a virtual CD (ISO9660) image.
+pub const RAM_DISK_VIRTUAL_CD_GUID: efi::Guid =
+    efi::Guid::from_fields(0x3d5abd30, 0x4175, 0x87ce, 0x6d, 0x64, &[0xd2, 0xad, 0xe5, 0x23, 0xc4, 0xbb]);
+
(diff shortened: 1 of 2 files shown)


sdk/patina/src/uefi_protocol/ram_disk.rs
```diff
@@ -0,0 +1,101 @@
+//! UEFI RAM Disk Protocol bindings.
+//!
+//! Used to install a contiguous range of system memory as a virtual block device the system
+//! firmware can boot from. The SRE flow uses this to boot the recovery WIM after reading it
```
os-d: This reads as a platform behavior, we should keep the protocol to the broad definition (i.e. nothing in here should talk about SRE)

os-d (commented):
Same comment here as the nvme one if Patina boot is planning on consuming this.

sdk/patina/src/uefi_protocol/ram_disk.rs
kat-perez: Thanks for the catch — pushed an update that drops the SRE / recovery-flow language from the module-level doc. The binding is now described purely in terms of the UEFI protocol behavior (install a host memory range as a virtual block device the firmware can boot from), with no references to specific consumers.

makubacki:
@kat-perez, given that this is a protocol defined in the UEFI Specification, can you please list the plan you have in mind here to coordinate your work using it with r-efi?

kat-perez:
> @kat-perez, given that this is a protocol defined in the UEFI Specification, can you please list the plan you have in mind here to coordinate your work using it with r-efi?

I'm treating this as the interim — I'll prep the upstream r-efi patch in parallel and follow up here to swap sdk/patina/uefi_protocol/ram_disk to a re-export once it ships.

makubacki:
> I'm treating this as the interim — I'll prep the upstream r-efi patch in parallel and follow up here to swap sdk/patina/uefi_protocol/ram_disk to a re-export once it ships.

If this is temporary and you already have the `patina_boot` crate in patina-components [using the `patina` code from `feature/patina-boot` branch](https://github.com/OpenDevicePartnership/patina-components/blob/1bcdd7a34b7558a38dacb7e2db069ff73e610188/patina_boot/Cargo.toml#L17), can you just add that there until it is in r-efi?

It would be best to avoid adding public content to the normal crate that are intentionally short-lived/misplaced.

PR #1496, opened 2026-05-04
patina_boot: Add install_ram_disk helper

### LINKED ISSUE #67: Add install_ram_disk helper to patina-boot
Add `install_ram_disk(bytes) -> DevicePathBuf` helper built on the RAM disk protocol binding. Produces a device path that can be passed to `boot_from_device_path`.

Part of #51.

kat-perez:
## Description

Adds `install_ram_disk(boot_services, bytes) -> Result<DevicePathBuf>` to `patina_boot::helpers`. Leaks `bytes` into a stable heap allocation, locates the UEFI RAM Disk Protocol, and registers the buffer as a virtual disk of type `RAM_DISK_VIRTUAL_DISK_GUID`. Returns the firmware-assigned device path as an owned `DevicePathBuf`, suitable for `boot_from_device_path` or similar consumers.

The backing memory lives for the rest of the firmware's lifetime — there is no `Drop` impl that frees it. Callers may invoke the protocol's `unregister` to remove the device-path entry, but the memory itself stays leaked.

Closes OpenDevicePartnership/odp-platform-common#67.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- `cargo make all`

## Integration Instructions

Stacked on **#1490** (RAM Disk protocol bindings). Merge #1490 first, then this rebases trivially onto `feature/patina-boot`.

### DIFF
--- a/sdk/patina/src/uefi_protocol.rs
+++ b/sdk/patina/src/uefi_protocol.rs
@@ -9,6 +9,7 @@
 
 pub mod decompress;
 pub mod performance_measurement;
+pub mod ram_disk;
 pub mod status_code;
 
 use crate::BinaryGuid;
--- a/sdk/patina/src/uefi_protocol/ram_disk.rs
+++ b/sdk/patina/src/uefi_protocol/ram_disk.rs
@@ -1,0 +1,100 @@
+//! UEFI RAM Disk Protocol bindings.
+//!
+//! Used to install a contiguous range of system memory as a virtual block device the system
+//! firmware can boot from.
+//!
+//! ## License
+//!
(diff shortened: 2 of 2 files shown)


kat-perez:
Closing in favor of OpenDevicePartnership/patina-components#13, which moves the `install` helper into a new `patina_ram_disk` crate. The protocol binding stays in sdk/patina per patina#1490 (next to decompress / performance_measurement / status_code); only the higher-level helper moves.

PR #1498, opened 2026-05-05
[REBASE & FF]Add USB and HID protocol definitions to SDK.

joschock:
## Description

* Add the EFI_USB_IO_PROTOCOL interface and associated USB descriptor types
as defined in the UEFI Specification, Chapter 17.2.4 (USB I/O Protocol).

* Add the HID_IO protocol interface as a vendor-specific protocol definition.
This protocol is defined by Project Mu and provides the interface between
HID transport drivers (producers) and HID class drivers (consumers).

This PR consists primarily of structural definitions of the protocols themselves, and does not include any consuming logic.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested via work in progress component development that requires these definitions. 

## Integration Instructions

N/A.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -75,6 +75,9 @@ serde_yaml = { version = "0.9" }
 serde_json = { version = "1.0", default-features = false, features = ["alloc"] }
 lzma-rs = { version = "0.3" }
 
+[patch.crates-io]
+r-efi = { git = "https://github.com/joschock/r-efi.git", branch = "add_usb_io" }
+
 [workspace.lints.clippy]
 indexing_slicing = "deny"
 undocumented_unsafe_blocks = "warn"
--- a/sdk/patina/src/lib.rs
+++ b/sdk/patina/src/lib.rs
@@ -61,6 +61,8 @@ pub mod tpl_mutex;
 pub mod uefi_decompress;
 #[cfg(any(test, feature = "alloc"))]
 pub mod uefi_protocol;
+#[cfg(any(test, feature = "alloc"))]
+pub mod vendor_protocols;
 
 /// Re-export of the [`safe-mmio`](https://crates.io/crates/safe-mmio) crate.
 ///
--- a/sdk/patina/src/uefi_protocol.rs
+++ b/sdk/patina/src/uefi_protocol.rs
@@ -87,3 +87,4 @@ impl_r_efi_protocol!(tcp6);
 impl_r_efi_protocol!(timestamp);
 impl_r_efi_protocol!(udp4);
 impl_r_efi_protocol!(udp6);
+impl_r_efi_protocol!(usb_io);
--- a/sdk/patina/src/vendor_protocols.rs
+++ b/sdk/patina/src/vendor_protocols.rs
@@ -1,0 +1,13 @@
+//! Vendor-specific UEFI protocol definitions.
+//!
+//! Protocols in this module are not part of the UEFI or PI specifications but
+//! follow the UEFI protocol model. They are defined by platform vendors.
+//!
+//! ## License
+//!
+//! Copyright (c) Microsoft Corporation.
+//!
+//! SPDX-License-Identifier: Apache-2.0
+//!
+
+pub mod hid_io;
--- a/sdk/patina/src/vendor_protocols/hid_io.rs
+++ b/sdk/patina/src/vendor_protocols/hid_io.rs
@@ -1,0 +1,76 @@
+//! HidIo protocol FFI definitions.
+//!
+//! Defines the HidIo protocol interface struct, types, and GUID shared between
+//! producers (e.g. `usb_hid`) and consumers (e.g. `uefi_hid`) of the protocol.
+//!
+//! ## License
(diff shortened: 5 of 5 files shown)


makubacki:
@joschock is there a reason not to take the UEFI Specification defined content to r-efi?

sdk/patina/src/uefi_protocol/usb_io.rs
```diff
@@ -0,0 +1,145 @@
...
+
+/// The EFI_USB_IO_PROTOCOL FFI interface.
+///
+/// Provides services to manage and communicate with USB devices.
+#[repr(C)]
+pub struct EfiUsbIoProtocol {
+    /// Manages a USB device with a control transfer pipe. A control transfer is
+    /// typically used to perform device initialization and configuration.
+    pub usb_control_transfer: unsafe extern "efiapi" fn(
+        this: *const EfiUsbIoProtocol,
```
makubacki: With the way this is defined in the UEFI Specification, I don't think we can consider it `*const` here.

```c
typedef
EFI_STATUS
(EFIAPI *EFI_USB_IO_CONTROL_TRANSFER) (
  IN EFI_USB_IO_PROTOCOL           *This,
  IN EFI_USB_DEVICE_REQUEST        *Request,
  IN EFI_USB_DATA_DIRECTION        Direction,
  IN UINT32                        Timeout,
  IN OUT VOID                      *Data OPTIONAL,
  IN UINTN                         DataLength OPTIONAL,
  OUT UINT32                       *Status
  );
```

sdk/patina/src/uefi_protocol/usb_io/types.rs
```diff
@@ -0,0 +1,177 @@
...
+    pub configuration: u8,
+    /// Configuration attributes (self-powered, remote wakeup, etc.).
+    pub attributes: u8,
+    /// Maximum power consumption in 2mA units.
+    pub max_power: u8,
+}
+
+/// Common header for USB descriptors.
+#[derive(Debug, Clone, Copy)]
+#[repr(C, packed)]
```
makubacki: Why is this structure (`UsbDescHead`) packed while the others are not?

sdk/patina/src/uefi_protocol/usb_io/types.rs
```diff
@@ -0,0 +1,177 @@
...
+/// USB system error.
+pub const EFI_USB_ERR_SYSTEM: u32 = 0x100;
+
+/// USB device descriptor.
+#[derive(Debug, Clone, Copy, Default)]
+#[repr(C)]
+pub struct EfiUsbDeviceDescriptor {
+    /// Size of this descriptor in bytes.
+    pub length: u8,
+    /// Descriptor type (USB_DESC_TYPE_DEVICE).
```
makubacki: `USB_DESC_TYPE_DEVICE` is not defined with the other USB descriptor types above (CONFIG and INTERFACE are currrently there). Since it's referenced here, it might be worth including for clarity. Were the other types (1, 3, 5) excluded intentionally?

joschock:
> @joschock is there a reason not to take the UEFI Specification defined content to r-efi?

Primarily for ease of prototyping this approach and getting feedback. I agree that it should be upstreamed.

sdk/patina/src/uefi_protocol/usb_io/types.rs
joschock: initially excluded because they weren't needed for the implementation; but agree they should be added to fully define the protocol

sdk/patina/src/uefi_protocol/usb_io/types.rs
joschock: copilot added it and it was an oversight on my part not to make it consistent with the others. In this case, it should not be needed as packed will result in the same as natural alignment for this struct.

joschock:
> > @joschock is there a reason not to take the UEFI Specification defined content to r-efi?
> 
> Primarily for ease of prototyping this approach and getting feedback. I agree that it should be upstreamed.

(definitely also needs to be more comprehensive as part of that)

os-d (approved):
Approving; agree with Michael's review. Can we file an issue to track upstream to r-efi?

makubacki (approved):
Approving with the plan to move USB IO out of this PR and take directly to r-efi (filling out all definitions).

joschock:
> Approving with the plan to move USB IO out of this PR and take directly to r-efi (filling out all definitions).

Branch for r_efi here: https://github.com/joschock/r-efi/commits/add_usb_io/, but waiting on https://github.com/OpenDevicePartnership/patina/pull/1480 to facilitate proper testing.

joschock:
Converted to draft -> moving the USB definition to a PR on r-EFI.

joschock:
r-efi PR to add USB IO is here: https://github.com/r-efi/r-efi/pull/101. Will promote this out of draft once that completes and release.

PR #1501, opened 2026-05-06
patina_smbios: Add SMBIOS types

kat-perez:
## Description

Define types for SMBIOS records according the SMBIOS specs 3.0+ and update records to use types. These structs/enums/bitfields force typing when creating SMBIOS records.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [x] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

`cargo make all`, plus patina-dxe-core-qemu integration: ported the Q35 and
SBSA SMBIOS platform components onto the new typed records, booted Q35 in
QEMU, and confirmed `q35_smbios_ffi_test` passes against the published SMBIOS
table.

## Integration Instructions

Downstream consumers of `patina_smbios` Type 0 / 1 / 2 / 3 / 4 / 7 / 16 / 17 /
19 records must construct fields with the new types instead of raw integers
(e.g. `BiosCharacteristics::from_bits(0x08)`, `WakeUpType::PowerSwitch`,
`BootUpState::Safe`). For Type 4, set `processor_family: u8 = 0xFE` and put
the typed value in `processor_family2`.

Co-Authored-By: Ansley Thompson <ansley.thompson@dell.com>

### DIFF
--- a/components/patina_smbios/src/service.rs
+++ b/components/patina_smbios/src/service.rs
@@ -696,14 +696,18 @@ mod tests {
             bios_starting_address_segment: 0xE800,
             firmware_release_date: 3,
             firmware_rom_size: 0xFF,
-            characteristics: smbios_types::BiosCharacteristics::from_bits(0x08),
-            characteristics_ext1: smbios_types::BiosCharacteristicsExt1::from_bits(0x03),
-            characteristics_ext2: smbios_types::BiosCharacteristicsExt2::from_bits(0x03),
+            characteristics: smbios_types::BiosCharacteristics::new().with_pci_supported(true),
+            characteristics_ext1: smbios_types::BiosCharacteristicsExt1::new()
+                .with_acpi_supported(true)
+                .with_usb_legacy_supported(true),
+            characteristics_ext2: smbios_types::BiosCharacteristicsExt2::new()
+                .with_bios_boot_specification_supported(true)
+                .with_fn_network_service_boot_supported(true),
             system_bios_major_release: 1,
             system_bios_minor_release: 0,
             embedded_controller_major_release: 0xFF,
             embedded_controller_minor_release: 0xFF,
-            extended_bios_rom_size: smbios_types::ExtendedBiosRomSize::from_bits(0),
+            extended_bios_rom_size: smbios_types::ExtendedBiosRomSize::new(),
             string_pool: vec![String::from("Vendor"), String::from("1.0"), String::from("2025")],
         };
 
@@ -809,14 +813,18 @@ mod tests {
             bios_starting_address_segment: 0xE800,
             firmware_release_date: 3,
             firmware_rom_size: 0xFF,
-            characteristics: smbios_types::BiosCharacteristics::from_bits(0x08),
-            characteristics_ext1: smbios_types::BiosCharacteristicsExt1::from_bits(0x03),
-            characteristics_ext2: smbios_types::BiosCharacteristicsExt2::from_bits(0x03),
+            characteristics: smbios_types::BiosCharacteristics::new().with_pci_supported(true),
+            characteristics_ext1: smbios_types::BiosCharacteristicsExt1::new()
+                .with_acpi_supported(true)
+                .with_usb_legacy_supported(true),
+            characteristics_ext2: smbios_types::BiosCharacteristicsExt2::new()
+                .with_bios_boot_specification_supported(true)
+                .with_fn_network_service_boot_supported(true),
             system_bios_major_release: 1,
             system_bios_minor_release: 0,
             embedded_controller_major_release: 0xFF,
             embedded_controller_minor_release: 0xFF,
-            extended_bios_rom_size: smbios_types::ExtendedBiosRomSize::from_bits(0),
+            extended_bios_rom_size: smbios_types::ExtendedBiosRomSize::new(),
             string_pool: vec![String::from("Vendor"), String::from("1.0"), String::from("01/01/2025")],
         };
 
@@ -842,14 +850,18 @@ mod tests {
             bios_starting_address_segment: 0xE800,
             firmware_release_date: 3,
             firmware_rom_size: 0xFF,
-            characteristics: smbios_types::BiosCharacteristics::from_bits(0x08),
-            characteristics_ext1: smbios_types::BiosCharacteristicsExt1::from_bits(0x03),
-            characteristics_ext2: smbios_types::BiosCharacteristicsExt2::from_bits(0x03),
+            characteristics: smbios_types::BiosCharacteristics::new().with_pci_supported(true),
+            characteristics_ext1: smbios_types::BiosCharacteristicsExt1::new()
+                .with_acpi_supported(true)
+                .with_usb_legacy_supported(true),
+            characteristics_ext2: smbios_types::BiosCharacteristicsExt2::new()
+                .with_bios_boot_specification_supported(true)
+                .with_fn_network_service_boot_supported(true),
             system_bios_major_release: 1,
             system_bios_minor_release: 0,
             embedded_controller_major_release: 0xFF,
             embedded_controller_minor_release: 0xFF,
-            extended_bios_rom_size: smbios_types::ExtendedBiosRomSize::from_bits(0),
+            extended_bios_rom_size: smbios_types::ExtendedBiosRomSize::new(),
             string_pool: vec![String::from("Vendor"), String::from("1.0"), String::from("01/01/2025")],
         };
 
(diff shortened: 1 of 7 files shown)


kat-perez:
@makubacki Can I please get help with overriding the QEMU PR validation? This is expected to fail and a follow-up PR will be opened in patina-dxe-core-qemu to address the breaking changes

components/patina_smbios/src/service.rs
```diff
@@ -693,14 +696,14 @@ mod tests {
             bios_starting_address_segment: 0xE800,
             firmware_release_date: 3,
             firmware_rom_size: 0xFF,
-            characteristics: 0x08,
-            characteristics_ext1: 0x03,
-            characteristics_ext2: 0x03,
+            characteristics: smbios_types::BiosCharacteristics::from_bits(0x08),
```
os-d: Seems odd to do this as from_bits when these are real values. May make the test easier to read if these follow the same convention as the real code.

components/patina_smbios/src/smbios_record.rs
```diff
@@ -950,10 +948,10 @@ mod tests {
             version: 2,
             serial_number: 3,
             asset_tag_number: 4,
-            bootup_state: 0x03,       // Safe
-            power_supply_state: 0x03, // Safe
-            thermal_state: 0x03,      // Safe
-            security_status: 0x03,    // None
+            bootup_state: smbios_types::BootUpState::Safe, // 0x03
```
os-d: don't need all the comments in the tests, it is self describing now

os-d (approved):

makubacki:
> @makubacki Can I please get help with overriding the QEMU PR validation? This is expected to fail and a follow-up PR will be opened in patina-dxe-core-qemu to address the breaking changes

We can do that. This would probably be a better target for the `major` branch instead of `main` though so we can pull it into a single release with other pending breaking changes.

components/patina_smbios/src/service.rs
kat-perez: Changed all `from_bits` fields to the corresponding real values

kat-perez:
> We can do that. This would probably be a better target for the `major` branch instead of `main` though so we can pull it into a single release with other pending breaking changes.

Thanks, retargeted for `major`

components/patina_smbios/src/smbios_types.rs
```diff
@@ -0,0 +1,1075 @@
...
+#[derive(Copy, Clone, Debug, IntoBytes, Immutable, KnownLayout)]
+pub enum ErrorCorrectionTypes {
+    Other = 0x01,
+    Unknown = 0x02,
+    NoEcc = 0x03,
+    Parity = 0x04,
+    SingleBitEcc = 0x05,
+    MultiBitEcc = 0x06,
+    Crc = 0x07,
+}
```
makubacki, line 1611: Is this enum (`ErrorCorrectionTypes`) and the one earlier in the file (`ErrorCorrectionType`) both needed? They have the exact same variants except this one includes `Crc = 0x07`.

The names a variant duplication are probably going to confuse people.

components/patina_smbios/src/smbios_types.rs
```diff
@@ -0,0 +1,1075 @@
...
+/// Cache Error Correction Type (Type 7, offset 0x10)
+#[repr(u8)]
+#[derive(Copy, Clone, Debug, IntoBytes, Immutable, KnownLayout)]
+pub enum ErrorCorrectionType {
+    Other = 0x01,
+    Unknown = 0x02,
+    NoEcc = 0x03,
+    Parity = 0x04,
+    SingleBitEcc = 0x05,
+    MutliBitEcc = 0x06,
```
makubacki:
```suggestion
    MultiBitEcc = 0x06,
```

cspell.yml
```diff
@@ -33,6 +33,36 @@ allowCompoundWords: true
...
+  - eeprom
+  - efficeon
+  - eprom
+  - expressx
+  - feprom
+  - loong
+  - loongson
+  - lpddr
+  - mrdimm
+  - mutli
```
makubacki: Please try to avoid adding "mutli" to the dictionary. I left a code suggestion for the typo fix in the only place I assume this was needed for.

cspell.yml
```diff
@@ -33,6 +33,36 @@ allowCompoundWords: true
 words:
```
makubacki, line 33: Sort the new words into the existing list to have an overall sorted list. This helps with maintenance.

components/patina_smbios/src/smbios_types.rs
```diff
@@ -0,0 +1,1075 @@
...
+//! `#[repr(u16)]` per the SMBIOS specification field width and likewise
+//! derive `IntoBytes`.
+//!
+//! ## License
+//!
+//! Copyright (c) Microsoft Corporation.
+//!
+//! SPDX-License-Identifier: Apache-2.0
+
+#![allow(missing_docs)]
```
makubacki: What's the main reason for `allow(missing_docs)` versus just adding the docs now?

makubacki (approved):

components/patina_smbios/src/smbios_types.rs
```diff
@@ -0,0 +1,1075 @@
...
+    Other = 0x01,
+    Unknown = 0x02,
+    Dram = 0x03,
+    NvdimmN = 0x04,
+    NvdimmF = 0x05,
+    NvdimmP = 0x06,
+    IntelOptanePersistentMemory = 0x07,
+}
+
+/// Memory Device Attributes (Type 17, offset 0x21)
```
makubacki:
```suggestion
/// Memory Device Attributes (Type 17, offset 0x1B)
```

components/patina_smbios/src/smbios_types.rs
```diff
@@ -0,0 +1,1075 @@
...
+    Unknown = 0x02,
+    Isa = 0x03,
+    Mca = 0x04,
+    Eisa = 0x05,
+    Pci = 0x06,
+    PcCard = 0x07,
+    VlVesa = 0x08,
+    Proprietary = 0x09,
+    ProcessorCardSlot = 0x0A,
+    ProprietaryMemroyCardSlot = 0x0B,
```
makubacki:
```suggestion
    ProprietaryMemoryCardSlot = 0x0B,
```

components/patina_smbios/src/smbios_types.rs
```diff
@@ -0,0 +1,1075 @@
...
+    pub escd_supported: bool,
+    pub cd_boot_supported: bool,
+    pub selectable_boot_supported: bool,
+    pub bios_rom_socketed: bool,
+    pub pc_card_boot_supported: bool,
+    pub edd_spec_supported: bool,
+    pub japanese_nec_9800_supported: bool,
+    pub japanese_toshiba_supported: bool,
+    pub kb_525_360_supported: bool,
+    pub mb_535_12_supported: bool,
```
makubacki:
```suggestion
    pub mb_525_12_supported: bool,
```

components/patina_smbios/src/smbios_types.rs
```diff
@@ -0,0 +1,1075 @@
...
+}
+
+/// BIOS Characteristics Extension Byte 1 (Type 0, offset 0x12)
+#[bitfield(u8)]
+#[derive(IntoBytes, Immutable, KnownLayout)]
+pub struct BiosCharacteristicsExt1 {
+    pub acpi_supported: bool,
+    pub usb_legacy_supported: bool,
+    pub agp_supported: bool,
+    pub i20_supported: bool,
```
makubacki: Intelligent Input/Output `O` instead of `0`.

```suggestion
    pub i2o_supported: bool,
```

makubacki (approved):

components/patina_smbios/src/smbios_types.rs
kat-perez: Each enum and struct in this file already has a type-level doc comment pointing to its SMBIOS spec section (e.g. /// Cache Error Correction Type (Type 7, offset 0x10)), and the variant/field names are taken straight from the spec — NoEcc, MultiBitEcc, Parity, etc. Per-variant /// would mostly just restate the identifier. So I left `#![allow(missing_docs)]` to avoid that noise.

Happy to drop the attribute and write per-variant docs if you'd rather have them in-tree — let me know.

components/patina_smbios/src/smbios_types.rs
kat-perez: Renamed to make the table-type association explicit: `ErrorCorrectionType` → `CacheErrorCorrectionType` (Type 7) and `ErrorCorrectionTypes` → `MemoryArrayErrorCorrectionType` (Type 16)

Also fixed `Type16PhysicalMemoryArray::memory_error_correction` — it was typed as the cache enum, which would've made it unable to emit Crc. Now uses `MemoryArrayErrorCorrectionType`.

components/patina_smbios/src/smbios_types.rs
makubacki: I understand what you're saying for the variants. More globally though, I'd like to avoid increasing the frequency of lint suppression like `missing_docs` to avoid reduce precedent for other places. It also seems like AI could handle it without much manual effort. So, please try to include it.

makubacki (approved):

components/patina_smbios/src/smbios_types.rs
kat-perez: Gotcha. Will remove the `missing_docs` and update the doc comments

PR #1503, opened 2026-05-07
[REBASE & FF] patina_debugger: Don't Touch Uncached Memory By Default

os-d:
## Description

This PR contains two commits:

Debugger: Move to PatinaPageTable Trait
--
Currently, the debugger uses the PageTable trait from patina_paging directly. On X64, this does not allow getting caching attributes since they are managed by MTRRs.

In preparation for using caching attributes in the debugger, move to the PatinaPageTable trait from patina_internal_cpu to get caching attributes uniformly.

This requires a new API in patina_internal_cpu to get the existing page table instead of creating a new one.

patina_debugger: Don't Access Uncached Memory By Default
--
It has been observed on physical platforms that WinDbg can request addresses in uncached memory. This results in the debugger attempting to read them. It uses the poke test to tell if this is safe. Sometimes the poke test works, but in one case, a machine check is generated to all cpus and the poke test cannot recover from this.

This changes this to check if memory is uncached. If so, the debugger will fail to read it. A new monitor cmd will toggle the ability of the debugger to read uncached memory or not.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

On a physical Intel platform that was getting a #MC when an uncached range was requested to be read by WinDbg and so the debugger read it because it was mapped. After this, the #MC doesn't occur.

## Integration Instructions

If uncached memory must be touched from the debugger, call `!uefiext.monitor toggle_uncached_access` (or otherwise call the monitor command) to toggle the ability.

### DIFF
--- a/core/patina_debugger/src/dbg_target/monitor.rs
+++ b/core/patina_debugger/src/dbg_target/monitor.rs
@@ -15,6 +15,7 @@ use gdbstub::target::ext::{self, monitor_cmd::ConsoleOutput};
 
 use crate::{
     arch::{DebuggerArch, SystemArch},
+    memory,
     system::SystemStateTrait,
 };
 
@@ -22,19 +23,21 @@ use super::PatinaTarget;
 
 const MONITOR_HELP: &str = "
 Patina Rust Debugger monitor commands:
-    help - Display this help.
     ? - Display information about the state of the machine.
-    reboot - Prepares to reboot the machine on the next continue.
-    mod ... - Commands for breaking on or quering modules.
     arch ... - Architecture specific commands.
+    mod ... - Commands for breaking on or quering modules.
+    help - Display this help.
+    reboot - Prepares to reboot the machine on the next continue.
+    toggle_uncached_access - Toggle accessing uncached memory regions.
+                             By default uncached memory access is disabled.
 ";
 
 const MOD_HELP: &str = "
 Mod commands:
-    list [count] [index] - List loaded modules.
     break [module] - Set load breakpoint for a module.
     breakall - Break on all module loads.
     clear - clear all module breakpoints.
+    list [count] [index] - List loaded modules.
 ";
 
 impl ext::monitor_cmd::MonitorCmd for PatinaTarget {
@@ -91,6 +94,14 @@ impl ext::monitor_cmd::MonitorCmd for PatinaTarget {
             }
             Some("arch") => {
                 SystemArch::monitor_cmd(&mut tokens, &mut buf);
+            }
+            Some("toggle_uncached_access") => {
+                let enabled = memory::toggle_uncached_access();
+                if enabled {
+                    let _ = buf.write_str("Uncached memory access enabled.");
+                } else {
+                    let _ = buf.write_str("Uncached memory access disabled.");
+                }
             }
             Some(cmd) => match self.system_state.try_lock() {
                 Some(state) => {
--- a/core/patina_debugger/src/memory.rs
+++ b/core/patina_debugger/src/memory.rs
@@ -7,7 +7,10 @@
 //! SPDX-License-Identifier: Apache-2.0
 //!
 
-use core::ptr;
+use core::{
+    ptr,
+    sync::atomic::{AtomicBool, Ordering},
+};
 
 use patina_internal_cpu::paging::PatinaPageTable;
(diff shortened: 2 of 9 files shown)


cfernald (approved):

kuqin12 (approved):

core/patina_debugger/src/memory.rs
```diff
@@ -7,15 +7,32 @@
...
+/// on memory-mapped I/O. A monitor command can toggle this to `true`.
+static ALLOW_UNCACHED_ACCESS: AtomicBool = AtomicBool::new(false);
+
+/// Toggles uncached memory access and returns the new state.
+pub fn toggle_uncached_access() -> bool {
+    let prev = ALLOW_UNCACHED_ACCESS.load(Ordering::Relaxed);
+    let new = !prev;
+    ALLOW_UNCACHED_ACCESS.store(new, Ordering::Relaxed);
+    new
+}
```
makubacki, line 31: I think this could be done in a single call with:

```suggestion
pub fn toggle_uncached_access() -> bool {
    !ALLOW_UNCACHED_ACCESS.fetch_xor(true, Ordering::Relaxed)
}
```

core/patina_debugger/src/dbg_target/monitor.rs
```diff
@@ -15,26 +15,29 @@ use gdbstub::target::ext::{self, monitor_cmd::ConsoleOutput};
...
-    help - Display this help.
     ? - Display information about the state of the machine.
-    reboot - Prepares to reboot the machine on the next continue.
-    mod ... - Commands for breaking on or quering modules.
     arch ... - Architecture specific commands.
+    mod ... - Commands for breaking on or quering modules.
+    help - Display this help.
+    reboot - Prepares to reboot the machine on the next continue.
+    toggle_uncached_access - Toggle reading uncached memory regions.
+                             By default uncached memory access is disabled.
```
makubacki: It looks like this gets used through `write_memory()` as well.

```suggestion
    toggle_uncached_access - Toggle accessing uncached memory regions.
                             By default uncached memory access is disabled.
```

core/patina_debugger/src/memory.rs
```diff
@@ -71,6 +88,7 @@ pub fn write_memory<Arch: DebuggerArch>(address: u64, buffer: &[u8]) -> Result<(
         // modify the page table to allow writing.
         let attributes = page_table
             .query_memory_region(page, PAGE_SIZE)
+            .map_err(|_| ())
             .expect("Unexpected failure on address that was already checked.");
```
makubacki, line 89: Are you sure this is necessary? `CacheAttributeValue` (returned by `query_memory_region()`) derives `Debug`.

core/patina_debugger/src/memory.rs
```diff
@@ -350,4 +376,45 @@ mod tests {
         // SAFETY: dest was allocated with this layout.
         unsafe { std::alloc::dealloc(dest, layout) };
     }
+
+    #[test]
```
makubacki, line 381: It seems like `ALLOW_UNCACHED_ACCESS` being used across these tests is a little dangerous. Like a test running in parallel to `test_access_check_uncached_page_allowed_when_toggled()` when `ALLOW_UNCACHED_ACCESS` is true.

Is there a reason not to use `PAGE_LOCK` in these?

core/patina_debugger/src/memory.rs
```diff
@@ -118,18 +136,24 @@ fn check_range_access<Arch: DebuggerArch>(
...
-        let res = page_table.query_memory_region(page, PAGE_SIZE).map_err(|_| ());
+        let res = page_table.query_memory_region(page, PAGE_SIZE);
         let valid = match res {
-            Ok(attributes) => !attributes.contains(MemoryAttributes::ReadProtect),
-            Err(_) => false,
+            Ok(attributes) => {
+                !attributes.contains(MemoryAttributes::ReadProtect)
+                    && (allow_uncached || !attributes.contains(MemoryAttributes::Uncached))
+            }
+            _ => false,
```
makubacki: This is okay, but I'm curious if it was necessary given `query_memory_region()` returns `Result`.

makubacki (approved):

PR #1509, opened 2026-05-13
gcd: Move mapping memory range message prior to adding memory space

makubacki:
## Description

Allows the memory range details to be visible if the call to `GCD.add_memory_space()` fails.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- `cargo make all`
- Boot and examine when the message is printed.

## Integration Instructions

- N/A

### DIFF
--- a/patina_dxe_core/src/gcd.rs
+++ b/patina_dxe_core/src/gcd.rs
@@ -635,6 +635,10 @@ pub fn add_hob_resource_descriptors_to_gcd(hob_list: &HobList) {
                     .take_while(|r| r.is_some())
                     .flatten()
             {
+                log::info!(
+                    "Mapping memory range {split_range:#x?} as {gcd_mem_type:?} with attributes {memory_attributes:#x?}",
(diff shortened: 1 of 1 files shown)


patina_dxe_core/src/gcd.rs
```diff
@@ -635,6 +635,10 @@ pub fn add_hob_resource_descriptors_to_gcd(hob_list: &HobList) {
                     .take_while(|r| r.is_some())
                     .flatten()
             {
+                log::info!(
```
makubacki, line 638: I thought this was more helpful in the case when `add_memory_space()` fails. If we'd like a separate message to stay before `set_memory_space_attributes()` about setting attributes, I can add it, but I thought it was unnecessary.

os-d (approved):

vineelko (approved):

PR #1511, opened 2026-05-15
docs: Add FFI docs

os-d:
## Description

This adds a doc for guidance on writing C FFIs.

Note: I wrote this document to have a place to document no VA_LIST across an FFI because there wasn't a good other place I could find. So, I added a small amount of other FFI guidance; if there is another spot folks want this, happy to move it, or if folks have suggestions for what else to add to this section, happy to add.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [x] Includes documentation?

## How This Was Tested

N/A.

## Integration Instructions

N/A.

### DIFF
--- a/docs/src/dev/principles/ffi.md
+++ b/docs/src/dev/principles/ffi.md
@@ -1,0 +1,38 @@
+# FFI Authoring
+
+Patina exposes C FFIs for interoperability with C based drivers. This documentation gives guidance on how to write these
+FFIs with patina specific guidance. For general guidance, view
+[the Rust FFI docs](https://doc.rust-lang.org/nomicon/ffi.html).
+
+## Patina Specific Considerations
+
+### EFIAPI
+
+All FFI functions must be marked with `extern "efiapi"` in order to use the EFIAPI ABI. Failure to include this
+will cause undefined behavior.
+
+>**Note:**: The one exception to this, temporarily, is for variadic functions. Rust only contains support for these
+through the
+[unstable c_variadics feature](https://doc.rust-lang.org/beta/unstable-book/language-features/c-variadic.html). This
+feature only supports variadics for `extern "C"`. The UEFI spec has some variadic functions that are required. As such,
+these functions use `extern "C"` for now. [This issue](https://github.com/r-efi/r-efi/issues/95) tracks adding support
+for `extern "efiapi"`.
+
+### va_list
+
+AARCH64 defines the [AAPCS64 ABI](https://github.com/ARM-software/abi-aa/blob/main/aapcs64/aapcs64.rst) as the single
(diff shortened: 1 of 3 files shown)


docs/src/dev/principles/ffi.md
```diff
@@ -0,0 +1,72 @@
...
+            descriptor.write_unaligned(target_descriptor);
+        },
+    }
+    efi::Status::SUCCESS
+}
+```
+
+The required safety checks are performed and then the Rust safe abstraction is called. This minimizes the amount of
+code duplication required to support both interfaces as well as restricting the amount of time we spend in the FFI
+layer.
```
makubacki: Do you envision this enumerating what the "required safety checks" are? For example, checking pointers for null is implied in the example but that's not the only one. As an example, structs transferred across the C FFI should use `#[repr(C)]`. I'm not sure how far we want to take it. A middle ground might be to keep the special notes like variadic usage already here and link to other good, official resources on FFI best practices to avoid repetition.

docs/src/dev/principles/ffi.md
```diff
@@ -0,0 +1,72 @@
...
+
+### va_list
+
+AARCH64 defines the [AAPCS64 ABI](https://github.com/ARM-software/abi-aa/blob/main/aapcs64/aapcs64.rst) as the single
+ABI for AARCH64. However, MSVC (and clang's aarch64-unknown-windows-msvc target to align to it) has broken this ABI in
+some places, resulting in an MS ABI for AARCH64 as well.
+
+The [aarch64-unknown-uefi target](https://doc.rust-lang.org/rustc/platform-support/unknown-uefi.html) used by Patina
+uses the MS ABI because that is what LLVM supports for building PE/COFF images, which are required for UEFI. Within
+Patina and Rust build code, this immaterial, as long as the target is the same. However, it affects interoperability
```
vineelko: NIT: "this is immaterial"

vineelko (approved):

docs/src/dev/principles/ffi.md
os-d: I think we can link to this: https://doc.rust-lang.org/nomicon/ffi.html. Which includes this kind of information. Certainly if we have any Patina specific guidance, then I would add it here.

makubacki: That works for me and then clarify the type of information that augments it in this file.

makubacki (approved):
Approving assuming a link to the FFI section in the Rustonomicon is added.

docs/src/dev/principles/ffi.md
os-d: Added.

os-d: I dropped the duplicate section on fast transition to safe Rust APIs, as well, so this document only has Patina specifics now (the Rust link covers that).

PR #1514, opened 2026-05-19
pi_dispatcher: Add UI Names to Prints

os-d:
## Description

Currently when various states occur, such as drivers not dispatched, only the GUID of the driver is printed. This is useful, but also it is nicer to have the UI name printed so the GUID does not have to be cross-referenced.

This parses the UI name from the FV, if present, and also prints it.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

N/A.

## Integration Instructions

N/A.

### DIFF
--- a/patina_dxe_core/src/pi_dispatcher.rs
+++ b/patina_dxe_core/src/pi_dispatcher.rs
@@ -17,6 +17,7 @@ mod section_decompress;
 use alloc::{
     boxed::Box,
     collections::{BTreeMap, BTreeSet},
+    string::{String, ToString},
     vec::Vec,
 };
 use core::{cmp::Ordering, ffi::c_void};
@@ -105,7 +106,11 @@ impl<P: PlatformInfo> PiDispatcher<P> {
     /// Displays drivers that were discovered but not dispatched.
     pub fn display_discovered_not_dispatched(&self) {
         for driver in &self.dispatcher_context.lock().pending_drivers {
-            log::warn!("Driver {:?} found but not dispatched.", guid_fmt!(driver.file_name));
+            log::warn!(
+                "Driver {} ({:?}) found but not dispatched.",
+                driver.name.as_deref().unwrap_or("Unnamed"),
+                guid_fmt!(driver.file_name)
+            );
         }
     }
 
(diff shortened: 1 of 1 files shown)


patina_dxe_core/src/pi_dispatcher.rs
```diff
@@ -105,7 +106,11 @@ impl<P: PlatformInfo> PiDispatcher<P> {
     /// Displays drivers that were discovered but not dispatched.
     pub fn display_discovered_not_dispatched(&self) {
         for driver in &self.dispatcher_context.lock().pending_drivers {
-            log::warn!("Driver {:?} found but not dispatched.", guid_fmt!(driver.file_name));
+            if let Some(name) = &driver.name {
+                log::warn!("Driver {name} ({:?}) found but not dispatched.", guid_fmt!(driver.file_name));
```
Javagedes: Personal nitpick: I like the `.unwrap_or("Unnamed")` method you do elsewhere, and I think it would make sense here also.

Feel free to close this one with no action.

Javagedes (approved):

makubacki (approved):

vineelko (approved):

patina_dxe_core/src/pi_dispatcher.rs
```diff
@@ -662,6 +670,16 @@ impl DispatcherContext {
                             .transpose()?
                             .map(Depex::from);
 
+                        let name = sections
+                            .iter()
+                            .find(|x| x.section_type() == Some(ffs::section::Type::UserInterface))
+                            .and_then(|x| x.try_content_as_slice().ok())
+                            .map(|data| {
+                                let chars: Vec<u16> =
+                                    data.chunks(2).map(|c| u16::from_le_bytes([c[0], c[1]])).collect();
```
makubacki: Does `c[1]` work as expacted if `data.len()` is odd? It seems like only `c[0]` would be available and it might panic.

os-d: Yes, good point, let me fix this

cfernald (approved):

os-d:
@makubacki @Javagedes can we force merge this over codecov complaints? I added a new unit test which got the patch coverage up to 75%, but the remaining lines are for log messages uncovered, either in cases where the lines are actually covered but we have the null logger so they show up as uncovered or cases where I updated log messages that existed in code that was uncovered

makubacki:
> @makubacki @Javagedes can we force merge this over codecov complaints?

No problem in this case if it's needed based on the final coverage numbers.

os-d:
After getting the fix for log messages not being covered, this is over 80% now, so doesn't need the force merge

PR #1515, opened 2026-05-20
Small Set of FFI Safety Improvements [Rebase & FF]

makubacki:
## Description

Makes a few changes to improve FFI safety noticed while reviewing the r-efi v6.0.0 PR (pre-existing problems that can be fixed in this PR before v6.0.0 integration).

I tried to find places where null checks and unaligned pointer accesses were not used and should be. Some places may have been missed if they didn't match a search pattern used.

Also, I added tests within reason. In particular, there is a large code coverage gap in `components/patina_acpi/src/acpi_protocol.rs` and `patina_dxe_core/src/protocols.rs` which I did not address since the lines modified are insignificant relative to the surrounding code already lacking coverage.

---

**Treat C FFI pointers as unaligned consistently**

Several UEFI protocol entry points dereference caller-supplied
pointers directly (`*ptr` or `ptr.as_mut()`). This can be undefined
behavior when the C caller passes an unaligned address.

The UEFI ABI does not guarantee that out-parameters or input
structures meet Rust's natural alignment for the pointee type, so
most loads and stores through a raw FFI pointer must go through
`read_unaligned()` / `write_unaligned()`.

While we've followed this pattern in the past, some cases were missed
that are updated in this change.

Also adds a Pointer Alignment section to the FFI Authoring doc (ffi.md)
that describes the rationale along with a few examples.

---

**Improve FFI pointer validation in a couple of places**

Two FFI entry points dereferenced caller-supplied pointers without
validating them:

  - patina_dxe_core: `copy_mem()` and `set_mem()` called
    `core::ptr::copy()` / `slice::from_raw_parts_mut()` on a null
    pointer, if the caller passed one, with a non-zero length.
  - patina_adv_logger: `adv_log_write()` constructed a slice from
    `buffer` and a reference from `this` with no validation.

The boot services functions now `debug_assert!()` on null with a
non-zero length to catch caller bugs clearly in debug builds, then
fall through to a safe early return in release. Since the function
does not return a value, this prevents a panic from propagating across
the FFI boundary.

`adv_log_write()` now returns `EFI_INVALID_PARAMETER` when `this` is
null or when `buffer` is null. The SAFETY comments are updated to reflect
the new invariants.

---

**cpu_arch_protocol: Prevent panic on null `this` pointers**

CPU Arch Protocol functions call a common helper to convert the raw
`this` pointer to a reference. Before, the helper panicked on null
pointers. This commit changes the helper to return `None` on null
pointers, so that the FFI entry points can return
`EFI_INVALID_PARAMETER` instead of panicking across the FFI
boundary.

---

**patina_adv_logger: Add unit tests to components.rs**

There are currently no unit tests for this module. Though the test
support needed to write tests without much overhead are in place,
such as UartNull.

This commit adds some basic tests for `adv_log_write` that verify the
pointer validation logic and calls to write the message when
expectations are valid.

---

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [x] Includes documentation?

## How This Was Tested

- `cargo make all`

## Integration Instructions

- N/A

### DIFF
--- a/docs/src/dev/principles/ffi.md
+++ b/docs/src/dev/principles/ffi.md
@@ -51,21 +51,142 @@ unaligned access intrinsics [`read_unaligned()`] and [`write_unaligned()`] or a 
 
 #### Rules
 
-1. **Read FFI pointers with [`read_unaligned()`].** Do not use methods that assume alignment, such as `*ptr`, `ptr::read`,
-   `&*ptr`, `(*ptr).field`, or `(*(ptr as *const T)).clone()` on a caller-supplied pointer.
-2. **Write FFI pointers with [`write_unaligned()`].** Do not use `*ptr = value` or `ptr::write`.
-3. **Be consistent within a function.** Mixing aligned and unaligned access on the same pointer is a code smell and
-   risks UB if the pointer is in fact misaligned.
-4. **Null-check before every unaligned access.** `read_unaligned()` and `write_unaligned()` only relax the *alignment*
-   requirement. The pointer must still be non-null and valid for the access. A null check should be the first operation
-   performed on an untrusted caller-supplied pointer before any other dereference or access, including unaligned ones.
-5. **Prefer [`zerocopy`] for structured reads.** When pulling a typed value out of a caller-supplied byte buffer,
+1. **Prefer [`zerocopy`] for structured reads.** When pulling a typed value out of a caller-supplied byte buffer,
    `zerocopy::FromBytes` provides a safe, alignment-agnostic alternative to manual `read_unaligned()` of a casted
    pointer.
+2. **Read FFI pointers with [`read_unaligned()`].** Do not use methods that assume alignment, such as `*ptr`, `ptr::read`,
+   `&*ptr`, `(*ptr).field`, or `(*(ptr as *const T)).clone()` on a caller-supplied pointer.
+3. **Write FFI pointers with [`write_unaligned()`].** Do not use `*ptr = value` or `ptr::write`.
+4. **Be consistent within a function.** Mixing aligned and unaligned access on the same pointer is a code smell and
+   risks UB if the pointer is in fact misaligned.
+5. **Null-check before every unaligned access.** `read_unaligned()` and `write_unaligned()` only relax the *alignment*
+   requirement. The pointer must still be non-null and valid for the access. A null check should be the first operation
+   performed on an untrusted caller-supplied pointer before any other dereference or access, including unaligned ones.
 
 [`read_unaligned()`]: https://doc.rust-lang.org/std/primitive.pointer.html#method.read_unaligned()
 [`write_unaligned()`]: https://doc.rust-lang.org/std/primitive.pointer.html#method.write_unaligned()
 [`zerocopy`]: https://crates.io/crates/zerocopy
+
+#### Zerocopy example
+
+The following example shows an `extern "efiapi"` function that receives an untrusted, potentially unaligned byte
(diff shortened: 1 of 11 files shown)


joschock (approved):

apop5 (approved):

docs/src/dev/principles/ffi.md
```diff
@@ -36,3 +36,67 @@ the va_list type in FFIs. In practice, this has been seen very little.
...
+// WRONG: aligned read of a caller-supplied integer.
+let n = unsafe { *num_bytes };
+
+// WRONG: aligned struct read out of an arbitrary byte buffer.
+let header = unsafe { (*(buffer as *const Header)).clone() };
+
+// WRONG: inconsistent as the same pointer is read aligned and is written unaligned.
+let n = unsafe { *num_bytes };
+// ...
+unsafe { num_bytes.write_unaligned()(actual) };
```
vineelko: NIT: minor typo `num_bytes.write_unaligned()(actual)`

vineelko (approved):

docs/src/dev/principles/ffi.md
```diff
@@ -36,3 +36,67 @@ the va_list type in FFIs. In practice, this has been seen very little.
...
+
+1. **Read FFI pointers with [`read_unaligned()`].** Do not use methods that assume alignment, such as `*ptr`, `ptr::read`,
+   `&*ptr`, `(*ptr).field`, or `(*(ptr as *const T)).clone()` on a caller-supplied pointer.
+2. **Write FFI pointers with [`write_unaligned()`].** Do not use `*ptr = value` or `ptr::write`.
+3. **Be consistent within a function.** Mixing aligned and unaligned access on the same pointer is a code smell and
+   risks UB if the pointer is in fact misaligned.
+4. **Null-check before every unaligned access.** `read_unaligned()` and `write_unaligned()` only relax the *alignment*
+   requirement. The pointer must still be non-null and valid for the access. A null check should be the first operation
+   performed on an untrusted caller-supplied pointer before any other dereference or access, including unaligned ones.
+5. **Prefer [`zerocopy`] for structured reads.** When pulling a typed value out of a caller-supplied byte buffer,
```
os-d: Nit: I would consider moving this to the first item, since this is what we want people to do if they can. I would also consider adding a zerocopy example for the same reason.

os-d (approved):

Javagedes (approved):

docs/src/dev/principles/ffi.md
makubacki: I added the example, but I thought it'd be nice if it could run in the playground. I got it to work, mainly by not directly using r-efi (since it is not supported by Rust Playground). During that exercise, I realized mdbook tests are kind of awkward locally in general, so I made some changes in https://github.com/OpenDevicePartnership/patina/pull/1518. I'd like to get that in first and then I'll push the example here.

makubacki: 1518 was merged. The zerocopy example has been pushed here. @os-d, can you take a look and close this thread?

cfernald (approved):

os-d (approved):

makubacki:
For the record, I meant for [ffi.md fixup - treat c ffi pointers consistently](https://github.com/OpenDevicePartnership/patina/commit/b0e7746b874b6c3258149e44f589a9ceab18e822) to be "fixed up" into the earlier commit that added the md file. Forgot to run the interactive rebase. Not a big deal though.

PR #1517, opened 2026-05-21
[REBASE & FF] Enable Clippy's Deny indexing_slicing Lint

os-d:
## Description

This enables an optional clippy lint to disallow direct slice indexing. First, all instances in the codebase are fixed or explicitly allowed.

The only places they are explicitly allowed are:
- In const functions because .get() is [currently not const](https://github.com/rust-lang/rust/issues/143775). An unstable feature could be enabled to allow this, but because const functions are evaluated at build time, it did not seem worthwhile to enable
- In test code

Each crate was updated separately to make any future bisects easier to identify where an issue occurred and to allow reviewers to focus on crates they care about.

- [x] Impacts functionality?
- [x] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Unit tests, booting to Win/Linux on Q35/SBSA and Windows on a physical Intel platform.

## Integration Instructions

Clippy will now fail when direct slice indexing is used. Developers must either use .get() (preferred) or disable the lint at the smallest level possible if it is not possible to use .get().

### DIFF
--- a/components/patina_acpi/src/acpi.rs
+++ b/components/patina_acpi/src/acpi.rs
@@ -510,7 +510,10 @@ where
             // Next entry goes after header + existing address entries.
             let entry_offset = ACPI_HEADER_LEN + xsdt_data.n_entries * ACPI_XSDT_ENTRY_SIZE;
             // Fill in the bytes of the new address entry.
-            xsdt_data.slice[entry_offset..entry_offset + ACPI_XSDT_ENTRY_SIZE]
+            xsdt_data
+                .slice
+                .get_mut(entry_offset..entry_offset + ACPI_XSDT_ENTRY_SIZE)
+                .ok_or(AcpiError::XsdtOverflow)?
                 .copy_from_slice(&new_table_addr.to_le_bytes());
 
             // Increase XSDT length by one entry.
@@ -661,6 +664,9 @@ where
                 let start_idx = ACPI_HEADER_LEN + idx * ACPI_XSDT_ENTRY_SIZE; // Find where the target entry starts.
                 let end_idx = ACPI_HEADER_LEN + xsdt_data.n_entries * ACPI_XSDT_ENTRY_SIZE; // Find where the XSDT ends.
 
+                // Validate the full range before mutating.
+                let _ = xsdt_data.slice.get(start_idx..end_idx).ok_or(AcpiError::XsdtOverflow)?;
+
                 // Shift all entries after the one being removed to the left.
                 // [.. before .. | target | <- .. after .. ]
                 // becomes [.. before .. | .. after.. ]
@@ -669,10 +675,13 @@ where
                 // Decrement entries.
                 xsdt_data.n_entries -= 1;
 
-                // Zero out the end of the XSDT.
-                // (After removing and shifting all entries, there is one extra slot at the end.)
-                // This is not technically necessary for correctness but is good practice for consistency.
-                xsdt_data.slice[end_idx - ACPI_XSDT_ENTRY_SIZE..end_idx].iter_mut().for_each(|b| *b = 0);
+                // Zero out the trailing slot freed by the shift.
+                xsdt_data
+                    .slice
+                    .get_mut(end_idx - ACPI_XSDT_ENTRY_SIZE..end_idx)
+                    .ok_or(AcpiError::XsdtOverflow)?
+                    .iter_mut()
+                    .for_each(|b| *b = 0);
 
                 // Decrease XSDT length.
                 xsdt_data.set_length(xsdt_data.get_length()? - ACPI_XSDT_ENTRY_SIZE as u32);
@@ -693,11 +702,12 @@ where
     pub(crate) fn checksum_common_tables(&self) {
         if let Some(ref mut xsdt_data) = *self.xsdt_metadata.lock() {
             // Zero the old checksum byte.
-            xsdt_data.slice[ACPI_CHECKSUM_OFFSET] = 0;
+            *xsdt_data.slice.get_mut(ACPI_CHECKSUM_OFFSET).expect("XSDT has a standard header") = 0;
             // Sum all bytes (wrapping since the checksum is a u8 between 0-255).
             let sum_of_bytes: u8 = xsdt_data.slice.iter().fold(0u8, |acc, &b| acc.wrapping_add(b));
             // Write new checksum: equivalent to -1 * `sum_of_bytes` (so the sum is zero modulo 256).
-            xsdt_data.slice[ACPI_CHECKSUM_OFFSET] = sum_of_bytes.wrapping_neg();
+            *xsdt_data.slice.get_mut(ACPI_CHECKSUM_OFFSET).expect("XSDT has a standard header") =
+                sum_of_bytes.wrapping_neg();
         }
 
         // The RSDP doesn't have a standard header, so it is easier to calculate the checksum manually.
@@ -713,16 +723,22 @@ where
         };
 
         // Zero out old checksums before recalculating.
-        rsdp_bytes[RSDP_CHECKSUM_OFFSET] = 0;
-        rsdp_bytes[RSDP_EXTENDED_CHECKSUM_OFFSET] = 0;
+        *rsdp_bytes.get_mut(RSDP_CHECKSUM_OFFSET).expect("RSDP is large enough for checksum offset") = 0;
+        *rsdp_bytes
+            .get_mut(RSDP_EXTENDED_CHECKSUM_OFFSET)
+            .expect("RSDP is large enough for extended checksum offset") = 0;
 
         // Calculate the `checksum` field (first 20 bytes).
-        let sum20: u8 = rsdp_bytes[..20].iter().fold(0u8, |acc, &b| acc.wrapping_add(b));
-        rsdp_bytes[RSDP_CHECKSUM_OFFSET] = sum20.wrapping_neg();
+        let sum20: u8 =
+            rsdp_bytes.get(..20).expect("RSDP is at least 20 bytes").iter().fold(0u8, |acc, &b| acc.wrapping_add(b));
+        *rsdp_bytes.get_mut(RSDP_CHECKSUM_OFFSET).expect("RSDP is large enough for checksum offset") =
+            sum20.wrapping_neg();
 
         // Calculate the `extended_checksum` (checksums over all bytes).
         let sum_of_bytes: u8 = rsdp_bytes.iter().fold(0u8, |acc, &b| acc.wrapping_add(b));
-        rsdp_bytes[RSDP_EXTENDED_CHECKSUM_OFFSET] = sum_of_bytes.wrapping_neg();
+        *rsdp_bytes
+            .get_mut(RSDP_EXTENDED_CHECKSUM_OFFSET)
+            .expect("RSDP is large enough for extended checksum offset") = sum_of_bytes.wrapping_neg();
     }
 
     /// Publishes ACPI tables after installation.
--- a/components/patina_adv_logger/src/logger.rs
+++ b/components/patina_adv_logger/src/logger.rs
@@ -311,8 +311,9 @@ where
             return;
         }
 
-        let data = &self.buffer[0..self.buffer_size];
-        self.writer.log_write(self.level, self.hw_print_mask_override, data);
+        if let Some(data) = self.buffer.get(0..self.buffer_size) {
+            self.writer.log_write(self.level, self.hw_print_mask_override, data);
+        }
         self.buffer_size = 0;
     }
 }
@@ -331,8 +332,10 @@ where
             if len > WRITER_BUFFER_SIZE - self.buffer_size {
                 self.flush();
             }
-            self.buffer[self.buffer_size..self.buffer_size + len].copy_from_slice(data);
-            self.buffer_size += len;
+            if let Some(dest) = self.buffer.get_mut(self.buffer_size..self.buffer_size + len) {
+                dest.copy_from_slice(data);
+                self.buffer_size += len;
+            }
         } else {
             // this message is too big to buffer, flush then write the message.
             self.flush();
--- a/components/patina_mm/src/component/communicator.rs
+++ b/components/patina_mm/src/component/communicator.rs
@@ -319,7 +319,7 @@ impl<E: MmExecutor + 'static> MmCommunication for MmCommunicator<E> {
         })?;
 
         log::debug!(target: "mm_comm", "Outgoing MM communication request: buffer_id={}, data_size={}, recipient={:?}", id, data_buffer.len(), recipient);
-        log::debug!(target: "mm_comm", "Request Data (hex): {:02X?}", &data_buffer[..core::cmp::min(data_buffer.len(), 64)]);
+        log::debug!(target: "mm_comm", "Request Data (hex): {:02X?}", data_buffer.get(..core::cmp::min(data_buffer.len(), 64)).unwrap_or(data_buffer));
         log::trace!(target: "mm_comm", "Comm buffer before request: {:?}", comm_buffer);
 
         // Set the mailbox status to indicate buffer is valid before triggering MMI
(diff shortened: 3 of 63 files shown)


components/patina_acpi/src/acpi.rs
```diff
@@ -661,18 +664,22 @@ where
                 let start_idx = ACPI_HEADER_LEN + idx * ACPI_XSDT_ENTRY_SIZE; // Find where the target entry starts.
                 let end_idx = ACPI_HEADER_LEN + xsdt_data.n_entries * ACPI_XSDT_ENTRY_SIZE; // Find where the XSDT ends.
 
+                // Validate the full range before mutating.
+                let _ = xsdt_data.slice.get(start_idx..end_idx).ok_or(AcpiError::XsdtOverflow)?;
+
                 // Shift all entries after the one being removed to the left.
-                // [.. before .. | target | <- .. after .. ]
-                // becomes [.. before .. | .. after.. ]
```
os-d, line 666: I missed a couple of small changes like this one where I fixed some comments that got dropped. Will fix.

patina_dxe_core/src/allocator/fixed_size_block_allocator.rs
```diff
@@ -252,15 +252,17 @@ impl FixedSizeBlockAllocator {
 
         match list_index(&layout) {
             Some(index) => {
-                match self.list_heads[index].take() {
+                let head = self.list_heads.get_mut(index).ok_or(FixedSizeBlockAllocatorError::InvalidLayout)?;
```
joschock: nit: consider a different error code? `InvalidLayout` implies that the input is bad, but this failure would reflect a logic problem in the implementation itself.

joschock (approved):

os-d:
Hmm, the clippy logic with all targets is not working correctly because it didn't check all the arm64 code. Will run on an arm64 machine and fix those up; will also take a look at why clippy with all targets isn't working here

patina_dxe_core/src/allocator/fixed_size_block_allocator.rs
os-d: Added an InternalError error code and return that here

components/patina_mm/src/component/communicator.rs
```diff
@@ -318,7 +318,7 @@ impl<E: MmExecutor + 'static> MmCommunication for MmCommunicator<E> {
         })?;
 
         log::debug!(target: "mm_comm", "Outgoing MM communication request: buffer_id={}, data_size={}, recipient={:?}", id, data_buffer.len(), recipient);
-        log::debug!(target: "mm_comm", "Request Data (hex): {:02X?}", &data_buffer[..core::cmp::min(data_buffer.len(), 64)]);
+        log::debug!(target: "mm_comm", "Request Data (hex): {:02X?}", data_buffer.get(..core::cmp::min(data_buffer.len(), 64)).unwrap_or(data_buffer));
```
Javagedes, line 322: This one seems like slicing is acceptable. I have only started scrolling through this code so I'm not sure if you are making some exceptions with allow(clippy::indexing_slicing)

components/patina_performance/src/component/performance.rs
```diff
@@ -311,14 +311,14 @@ impl<'a> Iterator for PerformanceRecordIterator<'a> {
...
-                self.bytes = &self.bytes[1..];
+                self.bytes = self.bytes.get(1..).unwrap_or(&[]);
                 return Some(Err(MmPerformanceError::RecordError(err.into())));
             }
         };
 
         let rec_len = header.length as usize;
         if rec_len < PerformanceRecordHeader::SIZE {
-            self.bytes = &self.bytes[PerformanceRecordHeader::SIZE..];
+            self.bytes = self.bytes.get(PerformanceRecordHeader::SIZE..).unwrap_or(&[]);
```
Javagedes, line 321: I'm wondering if its better to leave self.bytes untouched if we cannot get the bytes. something like

```rust
if rec_len < PerformanceRecordHeader::SIZE
  and let Some(bytes) = self.bytes.get(PerformanceRecordHeader::SIZE..)
{
    self.bytes = bytes;
}
```

core/patina_internal_collections/src/node.rs
```diff
@@ -74,18 +74,19 @@ where
...
         storage
     }
 
     fn build_linked_list(buffer: &[Node<D>]) {
-        let mut node = &buffer[0];
-        for next in buffer.iter().skip(1) {
-            node.set_right(Some(next));
-            next.set_left(Some(node));
-            node = next;
+        for pair in buffer.windows(2) {
```
Javagedes: I think this is cleaner:

```rust
for [node, next] in buffer.array_windows::<2>() {
    node.set_right(Some(next));
    next.set_left(Some(node));
}
```

core/patina_internal_collections/src/node.rs
```diff
@@ -235,49 +236,56 @@ where
         if self.capacity() == 0 {
             self.data = buffer;
             Self::build_linked_list(self.data);
-            self.available.set(self.data[0].as_mut_ptr());
+            // first() is safe: we just assigned a non-empty buffer (assert above ensures len >= capacity).
+            self.available.set(self.data.first().expect("non-empty buffer").as_mut_ptr());
```
Javagedes: This one seems unnecessary and results in a safety check followed by an unwrap uncheck. The performance difference is minimal but still not sure if this helps anything at all.

core/patina_internal_collections/src/sorted_slice.rs
```diff
@@ -72,12 +72,13 @@ where
             }
         }
 
-        let Err(idx) = self.search(elements[0]) else {
+        let first = *elements.first().ok_or(Error::NotSorted)?;
```
Javagedes, line 75: These new deref's make me nervous as to if we are copying or dereferencing.

core/patina_internal_collections/src/sorted_slice.rs
```diff
@@ -123,18 +124,28 @@ where
     ///
     /// Returns the exact datum if it exists, or the closest datum that is greater than the key if it does not.
     pub fn search_with_key(&self, key: &T::Key) -> Result<&T, &T> {
```
Javagedes, line 126: to keep the immutable and mutable functions almost the same, maybe we do something like below. the `self.is_empty()` assert should make sure you can index without using `get()` you could still use get and "expect" if you wanted, to provide a better error message. Otherwise, we need to change the interface for this function IMO.

```rust
pub fn search_with_key(&self, key: &T::Key) -> Result<&T, &T> {
assert!(!self.is_empty());    
match self.binary_search_by_key(key, |e| e.key()) {
        Ok(idx) => Ok(&self[idx]),
        Err(idx) => {
            // Clamp to last valid index when binary_search returns past-the-end.
            let idx = idx.min(self.len().saturating_sub(1));
            Err(&self[idx])
        }
    }
}

pub fn search_with_key_mut(&mut self, key: &T::Key) -> Result<&mut T, &mut T> {
    match self.binary_search_by_key(key, |e| e.key()) {
        Ok(idx) => Ok(&mut self[idx]),
        Err(idx) => {
            // Clamp to last valid index when binary_search returns past-the-end.
            let idx = idx.min(self.len().saturating_sub(1));
            Err(&mut self[idx])
        }
    }
}
```

core/patina_debugger/src/dbg_target/monitor.rs
```diff
@@ -215,7 +215,9 @@ impl<const N: usize> Write for MonitorBuffer<'_, N> {
             if len > N - self.pos {
                 self.flush();
             }
-            self.buffer[self.pos..self.pos + len].copy_from_slice(data);
+            if let Some(dest) = self.buffer.get_mut(self.pos..self.pos + len) {
+                dest.copy_from_slice(data);
+            }
             self.pos += len;
```
makubacki: It seems like something more should be done if the copy cannot be performed than just bump `self.pos`.

components/patina_adv_logger/src/logger.rs
```diff
@@ -331,7 +332,9 @@ where
             if len > WRITER_BUFFER_SIZE - self.buffer_size {
                 self.flush();
             }
-            self.buffer[self.buffer_size..self.buffer_size + len].copy_from_slice(data);
+            if let Some(dest) = self.buffer.get_mut(self.buffer_size..self.buffer_size + len) {
+                dest.copy_from_slice(data);
+            }
             self.buffer_size += len;
```
makubacki: Similar to the case in monitor.rs, it seems strange to keep incrementing `self.buffer_size` here if the copy was not possible.

sdk/patina/src/component/storage.rs
```diff
@@ -64,7 +64,9 @@ impl<V> SparseVec<V> {
         if index >= self.values.len() {
             self.values.resize_with(index + 1, || None);
         }
-        self.values[index] = Some(value);
+        if let Some(slot) = self.values.get_mut(index) {
+            *slot = Some(value);
+        }
```
makubacki: This code is a little more verbose/less readable. It seems like the `resize_with()` call should catch any practical problems. Is that your understanding?

core/patina_debugger/src/arch/x64.rs
```diff
@@ -293,7 +293,7 @@ impl Registers for X64CoreRegs {
                     return Err(());
                 }
                 let mut array = [0u8; core::mem::size_of::<$t>()];
```
makubacki, line 294: Can this follow what was done for the `read!()` macro in core\patina_debugger\src\arch\aarch64.rs?

components/patina_mm/src/component/communicator.rs
os-d: My general pattern was to not allow direct slice indexing except for cases where we had to, i.e. const functions. Otherwise, only tests were allowed to have it.

There are a lot of places where it is acceptable because of checks being done, but I tried to have migrate to using .get() and dropping the redundant checks in the code. This reduces the number of allows required and then also the .get() stays with the slice if you move logic around, but you may accidentally move the slice indexing before the check and blow up, etc.

So, I'm not opposed to allow direct slice indexing some places, but I do think we should be judicious in where we choose to. I would prefer to have a reason to allow it, i.e. perf, instead of having a reason to not use it.

For this case, it's a debug print that isn't compiled in most of the time, I'd just leave .get and say, yeah, it's a little silly it does .get() on a min of the buffer len, but who cares. But I don't have a strong feeling on it.

components/patina_acpi/src/acpi.rs
```diff
@@ -795,6 +811,7 @@ where
 }
 
 #[cfg(test)]
+#[allow(clippy::indexing_slicing)]
```
Javagedes: I believe you can put this in a clippy.toml so that you don't need to do this on all the tests:

```toml
# Workspace clippy.toml

allow-indexing-slicing-in-tests = true
```

os-d: Thanks, that's a lot nicer

components/patina_mm/src/component/communicator.rs
os-d: As noted, because this is debug code, I'm going to leave it rather than explicitly allow it.

components/patina_performance/src/component/performance.rs
os-d: This one seems to be the original intention, it follows another error case: https://github.com/OpenDevicePartnership/patina/blob/c059284b203e3c34f4bc95097e87d88a142c679a/components/patina_performance/src/component/performance.rs#L332, where if we have a bad record, just consume the rest of the bytes so the iterator is empty. I'm not really sure if that is required though, since the iterator returns an error in this case

core/patina_internal_collections/src/node.rs
os-d: Actually, the comment looks wrong in this case and maybe a slightly different flow is needed because the assert actually only checks that the buffer is >= 0, which is clearly not useful

core/patina_internal_collections/src/sorted_slice.rs
os-d: Does it matter in this case? Other than a potential perf difference of copying? We aren't changing the values, just using them in the search, where they are getting passed by value anyway

core/patina_internal_collections/src/sorted_slice.rs
os-d: I adopted this, keeping .get for the better error message and just in the event something really bad happened and the search fn is wrong or the self maintained len is wrong.

core/patina_internal_collections/src/node.rs
Javagedes: I think the right answer is when the old and new buffer are both of length zero - this expect will hit and we will panic. The right route to go is an unwrap_or_default() to set it back to a null ptr for the head node.

core/patina_internal_collections/src/sorted_slice.rs
os-d: Chatted offline, the concern was just on readability, added an explicit .clone() instead of the implicit copy.

core/patina_debugger/src/dbg_target/monitor.rs
os-d: This condition should not be hit due to the flush before this, but I agree it is weird to increment the position if we didn't copy, so I moved that inside the if condition.

components/patina_adv_logger/src/logger.rs
os-d: Also moved this to inside the if statement.

sdk/patina/src/component/storage.rs
os-d: I agree. I also think it is weird that insert could fail to insert. I changed this to use an expect instead and then always assign. If you'd prefer, I can explicitly allow the slice indexing here, but I like the property that the .get travels with the assignment, so even if it got moved above the resize, we wouldn't have a bad indexing. But in this case it is unlikely.

core/patina_debugger/src/arch/x64.rs
os-d: Agreed that is more readable. Updated.

core/patina_internal_collections/src/sorted_slice.rs
os-d: Turns out clippy doesn't like .clone() on something that impls Copy and so I put it back :)

os-d:
@makubacki @Javagedes I have addressed all comments, please re-review. I also rebased on main, there was one additional fix in pi_dispatcher.rs

makubacki:
> @makubacki @Javagedes I have addressed all comments, please re-review. I also rebased on main, there was one additional fix in pi_dispatcher.rs

Thanks. I'm reviewing now.

Javagedes (approved):

sdk/patina/src/performance/record/hob.rs
```diff
@@ -49,7 +49,7 @@ impl FromHob for HobPerformanceData {
             log::error!("Performance: error while parsing HobPerformanceRecordBuffer, return default value.");
             return Self::default();
         };
-        let records_data_buffer = bytes[offset..offset + size_of_all_entries as usize].to_vec();
+        let records_data_buffer = bytes.get(offset..offset + size_of_all_entries as usize).unwrap_or_default().to_vec();
```
makubacki: It seems like an error might be worth raising here even if with `debug_assert!()`.

patina_dxe_core/src/memory_bin.rs
makubacki, line 1: For each `expect()` did you confirm that is the best path versus returning an error? Not in reference to any place in particular, just want to make sure it was considered.

makubacki (approved):

sdk/patina/src/performance/record/hob.rs
os-d: I'll add that

patina_dxe_core/src/memory_bin.rs
os-d: Yeah, I did review each one and in places where I believed we should not fail, I used expect

PR #1518, opened 2026-05-21
Improve local mdbook testing [Rebase & FF]

makubacki:
## Description

Changes to simplify testing mdbooks.

---

**Simplify local mdbook testing**

Today, trying to run mdbook tests locally is a bit of a pain because:

1. The `dev` profile is used to build dependencies used by code in the mdbook, and it has LTO enabled. This means that the resulting rlibs don't contain machine code and can't be linked by `rustdoc --test` when running the doctests embedded in the mdbook. This results in link errors.
2. mdbook defaults to using the stable toolchain ignoring the `rust-toolchain.toml` file, which means that you have to manually switch to the nightly toolchain before running the tests.
3. The dependencies have to be built separately (`cargo make build-lib`). which are built using the toolchain in `rust-toolchain.toml`.

This is accounted for in CI with manual steps:

https://github.com/OpenDevicePartnership/patina-devops/blob/main/.github/workflows/MdbookWorkflow.yml

This commit simplifies all of this by introducing a new profile specifically for mdbook that inherits from `dev` but disables LTO and adding a `cargo-make` task to build the dependencies using this profile and run `mdbook test` using the toolchain specified in `rust-toolchain.toml`.

Another task was added called `serve-mdbook` to make it easier to open the mdbook locally similar to `doc-open` for the docs.

- Test mdbook code: `cargo make test-mdbook`
- Serve mdbook locally: `cargo make serve-mdbook`

This also has the advantage that mdbook tests are checked during `cargo make all` now.

---

**book.toml: Change playpen to playground**

I could not find "playpen" in the mdbook documentation:

https://rust-lang.github.io/mdBook/

As far as I can tell, "playpen" is a deprecated name for "playground".

This changes the name of the section in book.toml to align with current mdbook documentation.

---

Note: I kept the dedicated job in [ci-workflow.yml](https://github.com/OpenDevicePartnership/patina/blob/main/.github/workflows/ci-workflow.yml) that calls to [patina-devops/MdbookWorkflow.yml](https://github.com/OpenDevicePartnership/patina-devops/blob/main/.github/workflows/MdbookWorkflow.yml) since `all` is not run in CI and having it in a separate job still seems reasonable even though the job itself has some redundancies now. I'll probably make some changes to MdbookWorkflow.yml in a patina-devops PR to remove complexity now handled by the make task.

---

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- `cargo make test-mdbook`
- `cargo make serve-mdbook`

## Integration Instructions

- N/A

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -301,6 +301,48 @@ clear = true
 command = "cargo"
 args = ["test", "--doc"]
 
+[tasks.build-mdbook-deps]
+description = """Builds workspace libraries with the `mdbook` profile.
+This is a prerequisite for `cargo make test-mdbook`."""
+clear = true
+env = { RUSTC_PROFILE = "mdbook" }
+command = "cargo"
+args = ["build", "--all-features", "@@split(INDIVIDUAL_PACKAGE_TARGETS, )", "@@split(STD_FLAGS, )"]
+dependencies = ["individual-package-targets"]
+
+[tasks.test-mdbook]
+description = """Builds the mdbook and runs all of its embedded doctests with
+`mdbook test`. Uses the workspace toolchain (from `rust-toolchain.toml`) so the
+doctests link against rlibs built by the same compiler, and points `mdbook test`
+to the mdbook profile `target/mdbook/deps/` directory produced by `build-mdbook-deps`.
+"""
+clear = true
+script_runner = "@duckscript"
+script = '''
+toolchain_file = readfile rust-toolchain.toml
+channel = set "stable"
+lines = split ${toolchain_file} "\n"
+for line in ${lines}
+    trimmed = trim ${line}
+    starts = starts_with ${trimmed} "channel"
+    if ${starts}
+        parts = split ${trimmed} "\""
+        channel = array_get ${parts} 1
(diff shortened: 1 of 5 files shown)


os-d (approved):

Javagedes (approved):

Makefile.toml
```diff
@@ -301,6 +301,48 @@ clear = true
 command = "cargo"
 args = ["test", "--doc"]
 
+[tasks.build-mdbook-deps]
```
Javagedes, line 304: I'm half tempted to say it would make more sense if these were named "mdbook-build-deps", "mdbook-test" and "mdbook-serve". However, it is so minimal of a nitpick that I don't think it has much impact. I leave it up to you and your thoughts. Feel free to close it.

makubacki: I was torn between those as well, but I thought leading with the verb might be more natural. If there is not a strong opinion, I'll keep what's present just to avoid having to change references.

PR #1520, opened 2026-05-21
[REBASE & FF] Use CRLF for Log Messages

os-d:
## Description

Currently Patina only uses LF for log message line endings. In some terminal emulators, the implicit CR is not added and so the logs look stilted, e.g.

```text
INFO - Some Log Message
                                            INFO - Oops, no CR
```

This contains two commits to fix that. One is in the SDK to convert strings printed with log::<level>!() to have CRLF as the line ending. The other is use write!() with CRLF explicitly stated instead of writeln!() as that only does LF.

Note: This matches edk2 behavior and C drivers using DEBUG(()) already get CRLF as their line ending: https://github.com/tianocore/edk2/blob/b03a21a63e3bd001f52c527e5a57feddb53a690b/MdePkg/Library/BasePrintLib/PrintLibInternal.c#L1107-L1136

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

On a physical ARM64 platform where the terminal emulator does not implicitly add CR.

## Integration Instructions

N/A.

### DIFF
--- a/components/patina_mm/src/component/communicator.rs
+++ b/components/patina_mm/src/component/communicator.rs
@@ -27,6 +27,7 @@ use patina::{
         service::{IntoService, Service},
     },
     pi::protocols::communication::EfiMmCommunicateHeader,
+    writelncrlf,
 };
 
 use alloc::vec::Vec;
@@ -255,11 +256,11 @@ impl MmCommunicator {
 
 impl<E: MmExecutor + 'static> Debug for MmCommunicator<E> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        writeln!(f, "MM Communicator:")?;
+        writelncrlf!(f, "MM Communicator:")?;
         for buffer in self.comm_buffers.borrow().iter() {
-            writeln!(f, "Comm Buffer: {buffer:?}")?;
+            writelncrlf!(f, "Comm Buffer: {buffer:?}")?;
         }
-        writeln!(f, "MM Executor Set: {}", self.mm_executor.is_some())?;
+        writelncrlf!(f, "MM Executor Set: {}", self.mm_executor.is_some())?;
         Ok(())
     }
 }
--- a/patina_dxe_core/src/debugger_reload.rs
+++ b/patina_dxe_core/src/debugger_reload.rs
@@ -19,7 +19,7 @@ use patina::{
     component::service::memory::{AllocationOptions, MemoryManager, PageAllocationStrategy},
     guids,
     pi::hob::{self},
-    uefi_size_to_pages,
+    uefi_size_to_pages, writelncrlf,
 };
 
 use crate::{memory_manager::CoreMemoryManager, pecoff};
@@ -51,7 +51,7 @@ pub fn tear_down_debugger_reload() {
 /// extension and so is not intended to be human friendly.
 fn reload_monitor(args: &mut core::str::SplitWhitespace<'_>, out: &mut dyn core::fmt::Write) {
     if TEAR_DOWN.load(core::sync::atomic::Ordering::Relaxed) {
-        let _ = writeln!(out, "Only supported at initial breakpoint");
+        let _ = writelncrlf!(out, "Only supported at initial breakpoint");
         return;
     }
 
@@ -63,7 +63,7 @@ fn reload_monitor(args: &mut core::str::SplitWhitespace<'_>, out: &mut dyn core:
             load_command(args, out);
         }
         _ => {
-            let _ = writeln!(out, "Unknown command");
+            let _ = writelncrlf!(out, "Unknown command");
         }
     }
 }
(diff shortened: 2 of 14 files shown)


patina_dxe_core/src/debugger_reload.rs
```diff
@@ -51,7 +51,7 @@ pub fn tear_down_debugger_reload() {
 /// extension and so is not intended to be human friendly.
 fn reload_monitor(args: &mut core::str::SplitWhitespace<'_>, out: &mut dyn core::fmt::Write) {
     if TEAR_DOWN.load(core::sync::atomic::Ordering::Relaxed) {
-        let _ = writeln!(out, "Only supported at initial breakpoint");
+        let _ = write!(out, "Only supported at initial breakpoint\r\n");
```
cfernald: nit: i'm fine with the change, but also the monitor commands are being handled by a terminal emulator so doesn't strictly matter.

core/patina_debugger/src/system/alloc.rs
```diff
@@ -50,7 +50,7 @@ impl SystemStateTrait for SystemState {
     fn dump_monitor_commands(&self, out: &mut dyn core::fmt::Write) {
         for cmd in &self.monitor_commands {
             if let Some(description) = cmd.description {
-                let _ = writeln!(out, "    {} - {}", cmd.command, description);
+                let _ = write!(out, "    {} - {}\r\n", cmd.command, description);
```
cfernald: Same in the patina_debugger changes, monitor commands are handled by the debugger application and so the carriage return may not be needed.

cfernald (approved):

Javagedes (approved):
We are going to have a hard time enforcing this.

core/patina_debugger/src/system/alloc.rs
os-d: Yeah, I just replaced it globally, fine to remove it here if you want

os-d:
> We are going to have a hard time enforcing this.

I don’t think it matters too much to enforce, a thing to pay attention to in review, but also if it slips through that’s fine, the log will just look bad in some places, this fixes major logs, like gcd, if it looks bad enough for something else, we can fix it later.

components/patina_mm/src/component/communicator.rs
makubacki, line 1: Can you update the [SDK: log: Use CRLF Instead of LF on Output Strings](https://github.com/OpenDevicePartnership/patina/pull/1520/changes/d240f7bb45fdd4394abc636f1a1df72f23aa7b8d) commit to include an update somewhere in the mdbook to state this convention? It's also something simple that could be added to the AI file.

makubacki, line 1: Are we sure that this is better than just writing a custom macro? It was nice to not have to worry about line endings when using `writeln!()`. And we could probably enforce it a little easier.

core/patina_debugger/src/system/alloc.rs
cfernald: I don't really care. You're welcome to leave it for consistency or remove it for churn reduction. I'll let you decide if you care enough either way.

os-d: I'll leave it, if nothing else for people to copy when doing a new one

components/patina_mm/src/component/communicator.rs
os-d: I don't have a preference, I can add a macro to the SDK. Main question would be whether to confusingly but also maybe helpfully also call it writeln!()?

makubacki: As suggested offline, to make the semantics clear, maybe `writelncrlf!()`.

os-d:
@makubacki I converted this to the writelncrlf macro and addressed other feedback. Please re-review.

@cfernald I dropped the debugger changes since they were not logging uses of writeln and I had to go make changes anyway.

sdk/patina/src/base.rs
```diff
@@ -383,6 +383,29 @@ macro_rules! signature {
...
+/// # Note
+/// This macro is typically used to format logs messages to maintain
+/// compatibility with a wide range of terminal emulators and log viewers.
+#[macro_export]
+macro_rules! writelncrlf {
+    ($target:expr, $($arg:tt)*) => {
+        write!($target, $($arg)*).and_then(|()| write!($target, "\r\n"))
+    };
+}
+
```
vineelko, line 408: NIT: Should this be part of log.rs?

vineelko (approved):

makubacki (approved):

sdk/patina/src/base.rs
makubacki: I personally think it is okay here as that contains more implementation-oriented code and this is a simple, generic macro.

os-d: Yeah, this macro is intended for all logging cases but isn't restricted to logging cases, it can be used for general formatting as well, so that's why I put it here. I'll leave it.

PR #1524, opened 2026-05-23
[x64] Do not issue `cli` if interrupt is not enabled

kuqin12:
## Description

The module could be used in the user environment, and `cli` will cause #GP.

This change will check on the interrupt state and opt out the `cli` if it is already disabled.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This was tested for building the uart module for ring 3 module and booted to OS desktop properly.

## Integration Instructions

N/A

### DIFF
--- a/sdk/patina/src/serial/uart.rs
+++ b/sdk/patina/src/serial/uart.rs
@@ -32,26 +32,28 @@ cfg_if::cfg_if! {
         use uart_16550::MmioSerialPort;
         use uart_16550::SerialPort as IoSerialPort;
 
+        /// Returns the Current Privilege Level (CPL) from the CS selector.
+        fn current_privilege_level() -> u16 {
+            let cs: u16;
+            // SAFETY: Reading the CS register has no side effects.
+            unsafe { core::arch::asm!("mov {0:x}, cs", out(reg) cs, options(nostack, nomem)); }
+            cs & 0b11
+        }
+
         /// Runs `f` with CPU interrupts disabled, restoring the previous interrupt state afterward.
-        ///
-        /// If interrupts were already disabled on entry, this is a no-op around `f`: interrupts
-        /// are neither disabled nor re-enabled.
         fn without_interrupts<F, R>(f: F) -> R
         where
             F: FnOnce() -> R,
         {
             let flags: u64;
-            // SAFETY: Reading RFLAGS has no side effects.
+            // SAFETY: Reading RFLAGS and disabling interrupts around the closure is safe because we restore
+            // interrupts afterwards, but only if they were previously enabled.
             unsafe {
-                core::arch::asm!("pushfq; pop {0}", out(reg) flags, options(nomem));
-            }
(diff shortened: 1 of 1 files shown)


os-d (approved):
Is the ring3 code always run with interrupts disabled? Or is it that reading rflags from ring 3 will always show interrupts disabled?

cfernald (approved):

sdk/patina/src/serial/uart.rs
```diff
@@ -33,19 +33,25 @@ cfg_if::cfg_if! {
         use uart_16550::SerialPort as IoSerialPort;
 
         /// Runs `f` with CPU interrupts disabled, restoring the previous interrupt state afterward.
+        ///
+        /// If interrupts were already disabled on entry, this is a no-op around `f`: interrupts
+        /// are neither disabled nor re-enabled.
         fn without_interrupts<F, R>(f: F) -> R
```
makubacki, line 44: Whether we want to take a dependency in this code on the `x86_64` is one thing, but it would clean this up quite a bit.

- [`pub fn without_interrupts<F, R>(f: F) -> R`](https://docs.rs/x86_64/latest/src/x86_64/instructions/interrupts.rs.html#58)

The IF bit can be read more clearly with:

```rust
x86_64::registers::rflags::read().contains(RFlags::INTERRUPT_FLAG)
```

If that's not used, I still think there is value in defining `1 << 9` as a named const (e.g. `IF_BIT`) for clarity.

sdk/patina/src/serial/uart.rs
makubacki, line 1: If this is about not being able to execute `cli` due to privilege, would it be cleaner to wrap the `without_interrupts()` call with a check for the current privilege level? To me, that more directly addresses the issue and keeps this implementation simple.

makubacki (approved):

sdk/patina/src/serial/uart.rs
os-d: I just stripped the x86_64 crate from Patina, including removing it from here :). I would prefer not to add it back.

At the time, we discussed making this a public interface in the SDK, but decided against it because it felt too risky to give consumers a general option to do this without TPLs, that it should generally be done within the TPL framework, which is then why this lives here.

makubacki: I recall the rationale to not make it generally available in Patina (through the SDK) now and that makes sense. We can use this thread to track making the code read a little more naturally than like the `x86_64` call. At a minimum, just naming the bit position would help.

sdk/patina/src/serial/uart.rs
kuqin12: Thanks, will update.

cfernald (approved):

sdk/patina/src/serial/uart.rs
kuqin12: This is updated.

PR #1531, opened 2026-05-26
component: Normalize core::any::type_name output (1.89.0 MSRV compat)

### LINKED ISSUE #1530: [Bug]: A few `component` unit tests fail to compile with 1.89.0 (MSRV)

### Current Behavior

Testing with the `nightly-2026-06-23` toolchain (`cargo +nightly-2026-06-23 test`) results in the following:

```
---- component::params::tests::test_config_cannot_be_accessed_while_unlocked stdout ----

thread 'component::params::tests::test_config_cannot_be_accessed_while_unlocked' panicked at sdk\patina\src\component\params.rs:922:9:
assertion failed: Config::<i32>::try_validate(&id,
        (&storage).into()).is_err_and(|err|
        err == "patina::component::params::Config<'_, i32> not available.")    

---- component::params::tests::test_config_mut_cannot_be_accessed_while_locked stdout ----

thread 'component::params::tests::test_config_mut_cannot_be_accessed_while_locked' panicked at sdk\patina\src\component\params.rs:936:9:
assertion failed: ConfigMut::<i32>::try_validate(&id,
        (&storage).into()).is_err_and(|err|
        err == "patina::component::params::ConfigMut<'_, i32> not available.") 

---- component::struct_component::tests::test_component_run_handling_works_as_expected stdout ----

thread 'component::struct_component::tests::test_component_run_handling_works_as_expected' panicked at sdk\patina\src\component\struct_component.rs:204:9:    
assertion `left == right` failed
  left: Some("patina::component::params::ConfigMut<u32>")
 right: Some("patina::component::params::ConfigMut<'_, u32>")
```

The tests are written to assert against the string rendered in newer toolchains that includes the generic lifetime parameter.

### Expected Behavior

Test pass with the MSRV to the currently supported toolchain.

### Steps To Reproduce

- `rustup toolchain install nightly-2026-06-23`
- `cargo +nightly-2026-06-23 test`

### Build Environment

```markdown
N/A
```

### Version Information

```text
Tag (as of): 21.0.4
```

### Urgency

Medium

I will fix it

No maintainer feedback needed

### Anything else?

_No response_

makubacki:
## Description

Resolves #1530 

The strings emitted by `core::any::type_name::<T>()` are explicitly best-effort and documented to vary between rustc releases. Per the official documentation:

  > This is intended for diagnostic use. The exact contents and format
  > of the string returned are not specified, other than being
  > best-effort description of the type. For example, amongst the
  > strings that type_name::<Option<String>>() might return are
  > "Option<String>" and "std::option::Option<std::string::String>".

  > The returned string must not be considered to be a unique
  > identifier of a type as multiple types may map to the same type
  > name. Similarly, there is no guarantee that all parts of a type
  > will appear in the returned string. In addition, the output may
  > change between versions of the compiler. For example, lifetime
  > specifiers were omitted in some earlier versions.

  > The current implementation uses the same infrastructure as
  > compiler diagnostics and debuginfo, but this is not guaranteed.

  https://doc.rust-lang.org/beta/core/any/fn.type_name.html

The component infrastructure embeds those strings directly in user-visible diagnostics (error messages, panic payloads, `MetaData::name`) and a handful of unit tests assert against their exact form. These unit tests currently fail on the MSRV 1.89.0 (which means consumers on 1.89.0) also see different strings.

For example, recent/current toolchains render lifetime-parameterized generics as `Config<'_, u32>` while older ones render the same type as `Config<u32>`, and three tests in `component::params` and `component::struct_component` hardcode the more recent rendered form.

This change introduces a small internal helper, `component::type_name`, that wraps `core::any::type_name` and strips anonymous lifetime tokens before returning the string.

This stabilizes user-visible diagnostics, which always read `Config<T>` regardless of which toolchain compiled the build.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- 1.89.0 equivalent:
  - `cargo +nightly-2025-06-23 test -p patina`
- Current (1.93.1):
  - `cargo test -p patina`

## Integration Instructions

- N/A

### DIFF
--- a/sdk/patina/src/component.rs
+++ b/sdk/patina/src/component.rs
@@ -142,6 +142,7 @@ pub mod params;
 pub mod service;
 mod storage;
 mod struct_component;
+mod type_name;
 
 use crate::error::Result;
 
--- a/sdk/patina/src/component/hob.rs
+++ b/sdk/patina/src/component/hob.rs
@@ -271,7 +271,7 @@ impl<'h, T: FromHob + 'static> Iterator for HobIter<'h, T> {
             return Some(
                 any_box
                     .downcast_ref::<T>()
-                    .unwrap_or_else(|| panic!("Hob should be of type {}", core::any::type_name::<T>())),
+                    .unwrap_or_else(|| panic!("Hob should be of type {}", super::type_name::normalized::<T>())),
             );
         }
         None
--- a/sdk/patina/src/component/metadata.rs
+++ b/sdk/patina/src/component/metadata.rs
@@ -28,7 +28,7 @@ pub struct MetaData {
 impl MetaData {
     /// Creates a new metadata object for a component.
     pub fn new<S>() -> Self {
-        Self { access: Access::new(), name: Cow::from(core::any::type_name::<S>()), error_message: None }
+        Self { access: Access::new(), name: Cow::from(super::type_name::normalized::<S>()), error_message: None }
     }
 
     /// Returns the name of the component, including the module path.
--- a/sdk/patina/src/component/params.rs
+++ b/sdk/patina/src/component/params.rs
@@ -154,7 +154,7 @@ pub unsafe trait Param {
         if Self::validate(state, storage) {
             Ok(())
         } else {
-            Err(Cow::from(alloc::format!("{} not available.", core::any::type_name::<Self>())))
+            Err(Cow::from(alloc::format!("{} not available.", super::type_name::normalized::<Self>())))
         }
     }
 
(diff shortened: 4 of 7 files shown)


os-d (approved):
This fixes the current issue, but still leaves open future changes to this. Do we think it is still reasonable to rely on this and fix it up as things change or do we need to rethink using it?

makubacki:
> This fixes the current issue, but still leaves open future changes to this. Do we think it is still reasonable to rely on this and fix it up as things change or do we need to rethink using it?

I considered this versus:

- Since `check`/`build` work, modify the test - These strings are used outside tests though, so I avoided that.
- Change the MSRV - Since this is the only problem I ran into that seemed extreme.

This does need us to update the abstraction if needed, but I thought having the means to do so and keeping it consistent for these use cases was reasonable. I'm open to alternatives if others prefer.

Javagedes (approved):

makubacki:
> > This fixes the current issue, but still leaves open future changes to this. Do we think it is still reasonable to rely on this and fix it up as things change or do we need to rethink using it?
> 
> I considered this versus...

I'll go ahead and merge this change to get it in today's release if there are no further requests for the approach.

PR #1532, opened 2026-05-26
[REBASE & FF] Run Clippy/Check on UEFI Target Triples

os-d:
## Description

Currently the clippy/check configuration is expecting that --all-targets is building for x86_64-unknown-uefi and aarch64-unknown-uefi. However, --all-targets means build lib, docs, test, bin, etc, not target triples. As such, various clippy and check errors are missed in architecture specific code. This adds new steps to cargo make clippy/check to check the architectures. It leaves linting the std code as well because --all-features cannot be used in no-std; it attempts to include std as a feature. Short of manually listing all features, this seemed the best way to get coverage, though it does increase clippy/check times.

Open to other suggestions on how to implement this or to drop the std checking (nice to keep for tests for use in rust-analyzer). With this PR, times increase (all times in seconds, cargo clean run before each invocation):

|        | New | Original |
|--------|-------|--------|
| Clippy | 46    | 17     |
| Check      | 49    | 39     |

>**Note:** The Makefile.toml change will be dropped from this PR and brought to all applicable makefiles in patina-devops. It is here for convenience for anyone who wants to test this change before recommending a different path.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Running `cargo make clippy` and `cargo make check`.

## Integration Instructions

N/A.

### DIFF
--- a/docs/src/background/rust_tools.md
+++ b/docs/src/background/rust_tools.md
@@ -25,6 +25,14 @@ The table below shows the comprehensive tool suite used in Patina compared to tr
 | **Test Coverage** | `cargo-llvm-cov` | gcov, OpenCppCoverage | Integrated coverage collection |
 | **Undefined Behavior Analysis** | `cargo miri` | Valgrind, UBSan | Catches memory safety issues in unsafe code |
 
+## Definitions
+
+CI Features: The set of all features in a crate that should be tested in CI against a `no_std` build. This only omits
+the features that require `std`.
+
+Default Features: The features for a given crate that are in the `default` feature, i.e. the features that are enabled
+by default when building a crate.
+
 ## The cargo Ecosystem: Central Command Hub
 
 ### Tooling Strategy
@@ -460,6 +468,38 @@ words:
 - [CSpell Configuration](https://cspell.org/docs/getting-started)
 - [Patina Configuration](https://github.com/OpenDevicePartnership/patina/blob/main/cspell.yml)
 
+### cargo-check (Fast Compilation Without Code Generation)
+
+**Purpose**: Fast build checking, use by Rust Analyzer to aid in development
+
+**Value**:
+
+- **Compiler Integration**: Leverages Rust's compiler infrastructure to do a quick build check
+- **Rust Analyzer Integration**: Allows a quick dev inner loop with Rust Analyzer reporting results in near real time
+
+**Comparison to C Development**:
+
+Code is written and later built fully to test, harder to have near real time results as code
+is being written.
+
+**Coverage in Patina**:
+
+Patina runs check against the UEFI targets and test code. `std` is not done separately from
+tests because this is a small set of Patina code and is expected to be built with the normal
+build commands. The `--all-features` parameter can only be used for `std/tests` because it pulls in
+std which conflicts with the `no_std` build. As such, the ci_features are tested for the
+`no_std` builds to test all features that do not require `std`.
+
+|                    | aarch64-unknown-uefi | x86_64-unknown-uefi | std | test |
+|:------------------:|:--------------------:|:-------------------:|:---:|:----:|
+| All Features       |                      |                     |     |  X   |
+| CI Features        |          X           |          X          |     |  X   |
+| No Default Features|                      |                     |  X  |  X   |
+
+**Further Documentation**:
+
+- [cargo-check docs](https://doc.rust-lang.org/cargo/commands/cargo-check.html)
+
 ### clippy (Static Analysis and Linting)
 
 **Purpose**: Advanced static analysis tool that catches bugs, performance issues, and style problems beyond what the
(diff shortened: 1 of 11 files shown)


Makefile.toml
```diff
@@ -113,12 +113,24 @@ clear = true
...
 
-[tasks.check_code]
-description = "Checks rust code for no_std build errors with results."
+[tasks.check_code_std]
+description = "Checks rust code for std build errors with results."
 private = true
 command = "cargo"
 args = ["check", "--all-targets", "--all-features", "@@split(CARGO_MAKE_TASK_ARGS, )"]
 
+[tasks.check_code_x64]
```
makubacki: I think the no-std checks should include `NO_STD_FLAGS`,

makubacki (approved):
Reminder to take the Makefile.toml changes to patina-devops.

Makefile.toml
antklein, line 11: Should this be updated?
`--timings` is no longer an unstable-option.
`NO_STD_FLAGS = "--profile ${RUSTC_PROFILE} -Zbuild-std=core,compiler_builtins,alloc -Zbuild-std-features=compiler-builtins-mem --timings"`

Please don't consider this as a blocking comment, as it's out of scope of the current PR.

Makefile.toml
```diff
@@ -278,8 +290,25 @@ command = "cargo"
...
-description = "Run cargo clippy."
+description = "Run cargo clippy for the host target and both UEFI cross-compilation targets."
 clear = true
+run_task = [{ name = ["clippy-x64", "clippy-aarch64", "clippy-std"], parallel = true }]
+
+[tasks.clippy-x64]
+description = "Run cargo clippy for x86_64-unknown-uefi."
+private = true
+command = "cargo"
+args = ["clippy", "--target", "x86_64-unknown-uefi", "@@split(NO_STD_FLAGS, )", "--", "-D", "warnings"]
```
antklein: Is it worth running arch specific clippy on STD as well?
Something like,
`args = ["clippy", "--target", "x86_64-unknown-uefi", "--all-features", "--", "-D", "warnings"]`

makubacki: With how much time is already increasing, I'd prefer to deprioritize/drop this if the impact is noticeable at all.

os-d: Yeah, I was debating dropping std for clippy at all, the reason I kept it was to get —all-features for broader coverage there. If we were okay with a static list of features (not great, but no other choice), then I’d drop the std clippy and just pass —features …

os-d: Same for check running on std. Running on tests is nice for rust-analyzer to work correctly there, but I think the std only buys —all-features there as well

vineelko (approved):

makubacki:
Please include a matrix in the mdbook with tools and targets so we can visually assess what's being run against which targets.

Makefile.toml
os-d: I'll take a look at this when I put up the patina-devops PR.

os-d:
@Javagedes @makubacki can you take a look at the latest? I added the docs, trying to integrate them into existing documentation on these tools. I'm not sure the tables really are that helpful in that format, let me know if something else would be better.

After more playing with this, removing `std` isn't enough, there are other features that pull in `std` like mockall, serde, and a few others. So, for now, I have gone with the ci_features approach, as I don't think we will get to `--all-features` for no_std.

Note the CI here will fail until patina-devops is updated (and then pushed out here) to drop no-default-features from the CI runs. With this change, I don't think that is needed any longer. We do need to change our features to be additive only, so there will be a small gap, but I don't think no-default-features completely covers the gap and I want to reduce the check time.

os-d:
I have dropped the Makefile change now that it is queued in patina-devops

makubacki:
> Please include a matrix in the mdbook with tools and targets so we can visually assess what's being run against which targets.

@os-d, I may have missed it, but did you add this yet? Nvm, I reread your comment. Let me look at the file holistically.

docs/src/background/rust_tools.md
```diff
@@ -460,6 +460,37 @@ words:
...
+Patina runs check against the UEFI targets and test code. std is not done separately from
+tests because this is a small set of Patina code and is expected to be built with the normal
+build commands. The `--all-features` parameter can only be used for std/tests because it pulls in
+std which conflicts with the no-std build. As such, the ci_features are tested for the
+no_std builds to test all features that do not require std.
+
+|              | aarch64-unknown-uefi | x86_64-unknown-uefi | std | test |
+|:------------:|:--------------------:|:-------------------:|:---:|:----:|
+| All Features |                      |                     |     |  X   |
+| CI Features  |          X           |          X          |     |  X   |
```
makubacki: Can you define "CI Features" somewhere toward the top of the file?

os-d: Yes, I am also adding default features to the check matrix, since that is kept

os-d:
@makubacki updated

cfernald (approved):

makubacki (approved):

PR #1540, opened 2026-05-28
Ensure cache consistency when changing to uncached memory [FF & REBASE] 

cfernald:
### patina_internal_cpu: Flush data cache which switching to uncached

When switching memory from a cached normal memory to a uncached or device
memory, the cache lines may stuill exist. Some platforms may snoop this
cache, but this is not guaranteed. To ensure data consistency accross all
platforms, the data cache should be explicitly cleaned.

This commit adds a arc architectural callout for when caching is changed.
For AARCH64 this will flush and invalidate the data cache when chaning
from a cached to a uncached memory type.

### patina_internal_cpu: Fix stride bug in cache operations

The current cache operations for aarch64 is using the cache line size mask
as the stride causing later cache lines to be skipped. This commit cleans
up the variables for clarity and adds input validation.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on physical and virtual platform 
- [x] Boot to OS
- [x] One off test for cache coherency

## Integration Instructions

N/A

### DIFF
--- a/core/patina_debugger/src/memory.rs
+++ b/core/patina_debugger/src/memory.rs
@@ -193,6 +193,13 @@ mod tests {
             fn install_page_table(&mut self) -> Result<(), PtError>;
             fn query_memory_region(&self, address: u64, size: u64) -> Result<MemoryAttributes, (PtError, CacheAttributeValue)>;
             fn dump_page_tables(&self, address: u64, size: u64) -> Result<(), PtError>;
+            fn handle_cacheability_change(
+                &self,
+                address: u64,
+                size: u64,
+                old_cache_attributes: MemoryAttributes,
+                new_cache_attributes: MemoryAttributes,
+            );
         }
     }
 
--- a/core/patina_internal_cpu/src/cpu/aarch64/cpu.rs
+++ b/core/patina_internal_cpu/src/cpu/aarch64/cpu.rs
@@ -29,13 +29,24 @@ impl EfiCpuAarch64 {
     }
 
     // AArch64 related cache functions
-    fn cache_range_operation(&self, _start: efi::PhysicalAddress, _length: u64, _op: CpuFlushType) {
-        let cacheline_alignment = self.data_cache_line_len() - 1;
-        let mut aligned_addr = _start - (_start & cacheline_alignment);
-        let end_addr = _start + _length;
+    fn cache_range_operation(&self, start: efi::PhysicalAddress, length: u64, op: CpuFlushType) {
+        if length == 0 {
+            return;
+        }
 
-        loop {
-            match _op {
+        let cacheline_size = self.data_cache_line_len();
+        let cacheline_mask = cacheline_size - 1;
+        let mut aligned_addr = start & !cacheline_mask;
+        let end_addr = match start.checked_add(length) {
+            Some(end_addr) => end_addr,
+            None => {
+                debug_assert!(false, "Cache range overflow");
+                return;
+            }
+        };
+
+        while aligned_addr < end_addr {
+            match op {
                 CpuFlushType::EfiCpuFlushTypeWriteBack => self.clean_data_entry_by_mva(aligned_addr),
                 CpuFlushType::EfiCpuFlushTypeInvalidate => self.invalidate_data_cache_entry_by_mva(aligned_addr),
                 CpuFlushType::EfiCpuFlushTypeWriteBackInvalidate => {
@@ -43,9 +54,9 @@ impl EfiCpuAarch64 {
                 }
             }
 
-            aligned_addr += cacheline_alignment;
-            if aligned_addr >= end_addr {
-                break;
+            match aligned_addr.checked_add(cacheline_size) {
+                Some(next) => aligned_addr = next,
+                None => break,
             }
         }
 
--- a/core/patina_internal_cpu/src/cpu/aarch64/cpu.rs
+++ b/core/patina_internal_cpu/src/cpu/aarch64/cpu.rs
@@ -6,7 +6,7 @@
 //!
 //! SPDX-License-Identifier: Apache-2.0
 //!
-use crate::cpu::Cpu;
+use crate::cpu::{Cpu, aarch64::cache};
 #[cfg(not(test))]
 use core::arch::asm;
 use patina::{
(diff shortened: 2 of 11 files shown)


kuqin12 (approved):

core/patina_internal_cpu/src/paging/x64.rs
```diff
@@ -94,6 +94,15 @@ where
...
     }
+
+    fn handle_cacheability_change(
+        &self,
+        _address: u64,
+        _size: u64,
+        _old_cache_attributes: MemoryAttributes,
+        _new_cache_attributes: MemoryAttributes,
+    ) {
+    }
```
os-d, line 106: perhaps a comment here on why it is empty

os-d (approved):

vineelko (approved):

core/patina_debugger/src/memory.rs
makubacki, line 1: I'm reviewing this now. Please hold for me to post any pending comments.

patina_dxe_core/src/gcd/spin_locked_gcd.rs
```diff
@@ -2163,6 +2163,14 @@ impl SpinLockedGcd {
                         if let Some(old_cache_attrs) = old_cache_attributes
                             && old_cache_attrs != new_cache_attributes
                         {
+                            // perform any architecture-specific cache management.
+                            page_table.handle_cacheability_change(
+                                base_address as u64,
+                                len as u64,
+                                old_cache_attrs,
+                                new_cache_attributes,
+                            );
```
makubacki: I thought this would happen before the region is remapped to the new attributes. For example, on x86 you would want to flush the cache before a MTRR/PAT change, not after. Is that a misunderstanding of what this callout is used for?

core/patina_internal_cpu/src/cpu/aarch64/cpu.rs
```diff
@@ -29,23 +29,28 @@ impl EfiCpuAarch64 {
...
-        let end_addr = _start + _length;
+    fn cache_range_operation(&self, start: efi::PhysicalAddress, length: u64, op: CpuFlushType) {
+        if length == 0 {
+            return;
+        }
+
+        let cacheline_size = self.data_cache_line_len();
+        let cacheline_mask = cacheline_size - 1;
+        let mut aligned_addr = start & !cacheline_mask;
+        let end_addr = start.saturating_add(length);
```
makubacki: I think `checked_add()` + raising an error or a `debug_assert!()`/return makes more sense here. `saturating_add()` is going to set `end_addr` to `u64::MAX` and continue with a likely wrong end address versus making it obvious the overflow happened and preventing propagation of a likely mistake.

core/patina_internal_cpu/src/paging/x64.rs
makubacki: Yeah, I'm interested to know why it is empty for x64.

core/patina_internal_cpu/src/paging.rs
```diff
@@ -103,4 +103,20 @@ pub trait PatinaPageTable {
...
+
+    /// Function to handle a change in the cacheability of a memory region.
+    /// This function is called when the cacheability of a memory region changes.
+    ///
+    /// ## Arguments
+    /// * `address` - The memory address of the region.
+    /// * `size` - The size of the region.
+    /// * `new_cache_attributes` - The new cache attributes for the region.
+    /// * `old_cache_attributes` - The old cache attributes for the region.
+    fn handle_cacheability_change(
```
makubacki, line 115:
```suggestion
    ///
    /// ## Arguments
    /// * `address` - The memory address of the region.
    /// * `size` - The size of the region.
    /// * `old_cache_attributes` - The old cache attributes for the region.
    /// * `new_cache_attributes` - The new cache attributes for the region.
    fn handle_cacheability_change(
```

core/patina_internal_cpu/src/paging/aarch64.rs
```diff
@@ -50,6 +60,21 @@ where
...
+        address: u64,
+        size: u64,
+        old_cache_attributes: MemoryAttributes,
+        new_cache_attributes: MemoryAttributes,
+    ) {
+        // If the region was previously cached (WB/WT) and is now uncached (UC/WC),
+        // clean and invalidate the data cache for the range so that any dirty lines
+        // are written back to memory before subsequent accesses bypass the cache.
+        if old_cache_attributes.intersects(CACHED_ATTRS) && new_cache_attributes.intersects(UNCACHED_ATTRS) {
+            let _ = EfiCpuAarch64.flush_data_cache(address, size, CpuFlushType::EfiCpuFlushTypeWriteBackInvalidate);
```
makubacki: This only works because `EfiCpuAarch64` is a unit struct (has not fields) so you don't need to instanatiate the struct to call the function. It feels like it is taking advantage of a current implementation situation rather than properly calling the function in a sustainable way.

I think you could also drop `pub(crate) mode aarch64;` if this was just afunction with the access needed in `core/patina_internal_cpu/src/cpu/aarch64/cpu.rs` like `pub(crate) fn flush_data_cache_range`.

core/patina_debugger/src/memory.rs
makubacki: Done

patina_dxe_core/src/gcd/spin_locked_gcd.rs
cfernald: Until the mapping is changed, the cache line can be recreated (speculatively or otherwise). The point of this is to ensure that before anyone actually uses the memory, all cache lines are cleaned. Doing this after the mapping change ensures that new lines aren't pulled in after the invalidate.

core/patina_internal_cpu/src/paging/x64.rs
cfernald: Well, frankly, it's empty because this hasn't been an issue. I'm not as familiar with the cache coherency guarantees on X64 as they haven't ever bitten me the same way that arm has. I'm happy to add the flush for this as well if we believe it is the correct thing to do.

cfernald: Actually, looking it seems we do a `wbinvd` in patina_mtrr, so I think it's already handled at that layer.

makubacki: That seems like a reasonable layer. Calling it out here would be a helpful future note.

patina_dxe_core/src/gcd/spin_locked_gcd.rs
makubacki: As long as that is documented, it is fine. The confusing part would be an x86 developer that thinks this is a way to handle typical maintenance prior to making a change.

core/patina_internal_cpu/src/paging/aarch64.rs
cfernald: Fair enough. I'll move some stuff around

makubacki (approved):

PR #1544, opened 2026-05-29
[mm] Do not introduce `talk_to_supervisor` bit

kuqin12:
## Description

Even the status buffer definition from basecore has this bit set:
- https://github.com/microsoft/mu_basecore/commit/597b1808db31d72bdd71fa6781474ec206ebc824
- https://github.com/microsoft/mu_basecore/commit/20b0561023e158bd245d37b8bddc137d2bd2fcfe

However, with the current usage of MM communicate, this bit is not needed for the proper communication invocation:
- https://github.com/microsoft/mu_feature_mm_supv/commit/b0743476d6384ca66d3e6505cf18f90532e149e1

In addition, the updated MM supervisor will no longer rely on this bit to differentiate the target channel, thus will not need this bit in the future, either.

This change removes the bit definition and the associated usage.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This was tested locally and booted to Windows desktop. Pipeline should verify the same as well.

## Integration Instructions

N/A


apop5 (approved):

makubacki (approved):

components/patina_mm/src/config.rs
makubacki, line 1: @kuqin12, can you please put change specifics (Mu versions/commit hashes) into the PR description to help connect the corresponding changes?

kuqin12: Added 2 commits from basecore and the commit that removes the need of this bit from supv.

PR #1550, opened 2026-06-02
patina_debugger: Do not initialize the transport by default

cfernald:
## Description

This change removes the default behavior to initialize the debugger transport serial port. Most systems will use a shared transport, and so the re-initialization is at best unnecessary. In some scenarios the Patina serial transport initialization may stomp on the existing configuration. This has caused confusion on multiple platforms during enablement. It is best to just skip unless requested. This commit inverts the default, deprecates `without_transport_init`, and introduces `with_transport_init` for the rare platforms that require this.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested that Q35 & SBSA debuggers still function

## Integration Instructions

Platforms using `without_transport_init()` may safely drop that function call. There are no known platforms that require transport initialization, but if so those should now call `with_transport_init()` as needed.

### DIFF
--- a/core/patina_debugger/src/debugger.rs
+++ b/core/patina_debugger/src/debugger.rs
@@ -77,8 +77,8 @@ where
     exception_types: &'static [usize],
     /// Controls what the debugger does with logging.
     log_policy: DebuggerLoggingPolicy,
-    /// Whether initializing the transport should be skipped.
-    no_transport_init: bool,
+    /// Whether the transport will be initialized.
+    transport_init: bool,
     /// Debugger enabled state.
     enabled: AtomicBool,
     /// Debugger Initialized state
@@ -120,7 +120,7 @@ impl<T: SerialIO> PatinaDebugger<T> {
         PatinaDebugger {
             transport,
             log_policy: DebuggerLoggingPolicy::SuspendLogging,
-            no_transport_init: false,
+            transport_init: false,
             exception_types: SystemArch::DEFAULT_EXCEPTION_TYPES,
             enabled: AtomicBool::new(false),
             initialized: AtomicBool::new(false),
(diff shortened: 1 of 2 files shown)


os-d (approved):

core/patina_debugger/src/debugger.rs
```diff
@@ -156,8 +156,21 @@ impl<T: SerialIO> PatinaDebugger<T> {
 
     /// Prevents the debugger from initializing the transport. This is suggested in
     /// cases where the transport is shared with the logging device.
-    pub const fn without_transport_init(mut self) -> Self {
-        self.no_transport_init = true;
+    ///
+    /// ## DEPRECATED
```
makubacki, line 160: Can you please update the platform integration guidance here to no longer use `without_transport_init()` as an example?

https://github.com/OpenDevicePartnership/patina/blob/b0a53df25c14c22703cbd7c7cba0158c77e6330d/core/patina_debugger/README.md?plain=1#L44-L54

Also, this section needs to be updated (and maybe others):

https://github.com/OpenDevicePartnership/patina/blob/b0a53df25c14c22703cbd7c7cba0158c77e6330d/core/patina_debugger/README.md?plain=1#L78-L97

core/patina_debugger/src/debugger.rs
```diff
@@ -156,8 +156,21 @@ impl<T: SerialIO> PatinaDebugger<T> {
...
+    ///
+    /// The transport is no longer initialized by default. This function has no effect and is deprecated. It will be
+    /// removed in a future version.
+    ///
+    #[deprecated(note = "This function has no effect and is deprecated. It will be removed in a future version.")]
```
makubacki: You might also want to consider adding some actionable info to the deprecation message as well, like:

> Default changed. Call `with_transport_init(true)` if you previously relied on default initialization.

core/patina_debugger/src/debugger.rs
```diff
@@ -156,8 +156,21 @@ impl<T: SerialIO> PatinaDebugger<T> {
...
+    pub const fn without_transport_init(self) -> Self {
+        self
+    }
+
+    /// Enables the debugger to initialize the transport. This is typically only required if the transport is not shared
+    /// with the logging device. Initializing the transport when it is shared may lead to unexpected behavior.
+    pub const fn with_transport_init(mut self, init: bool) -> Self {
```
makubacki: I can see why the new API takes a `bool` argument, but the old one was cleaner and `with_transport_init(self) -> Self` would be here too if it could override a default.

core/patina_debugger/src/debugger.rs
```diff
@@ -156,8 +156,21 @@ impl<T: SerialIO> PatinaDebugger<T> {
 
     /// Prevents the debugger from initializing the transport. This is suggested in
     /// cases where the transport is shared with the logging device.
```
makubacki: Right now, the doc comment is confusing to the current functionality. Example of a new message:

```suggestion
    /// Deprecated API that previously prevented the debugger from initializing...
```

makubacki (commented):
I added the design change label to at least rev the minor version for visibility to the deprecation change in the release notes. There should be integration guidance in the PR description as well though.

core/patina_debugger/src/debugger.rs
cfernald: Adding to the note that the call may be safely removed. The people seeing this warning need no additional action since this is now the default behavior. Callers who _need_ initialization do need to use with init, but doesn't apply to any platforms i'm aware of.

makubacki (approved):

PR #1552, opened 2026-06-03
pi_dispatcher: Report protocols present for Drivers Not Dispatched

yangrongwei:
## Description

`pi_dispatcher.rs` reports protocols present for each "found but not dispatched" driver.

```c
08:12:04.220 : INFO - Finished Dispatching Drivers
08:12:04.221 : WARN - Driver MacAddressEmulationDxe (72AA9388-70DD-4669-8EC1-87444DD62CF3) found but not dispatched.
08:12:04.221 : WARN - Driver DisplayEngine (C09E4CE4-AE29-4340-B941-078871BD89D1) found but not dispatched.
08:12:04.221 : WARN - ..
08:12:04.222 : DEBUG - Begin Report of Drivers Not Dispatched:
08:12:04.224 : DEBUG - Driver MacAddressEmulationDxe (72AA9388-70DD-4669-8EC1-87444DD62CF3) found but not dispatched. Protocols present:
08:12:04.224 : DEBUG -   13a3f0f6-264a-3ef0-f2e0-dec512342f34 : true
08:12:04.224 : DEBUG -   81d1675c-86f6-48df-bd95-9a6e4f0925c3 : true
08:12:04.224 : DEBUG -   e355fc05-37c2-468a-9f52-52e5bf1871d3 : true
08:12:04.224 : DEBUG -   7b9cc1a0-218d-4c1c-9fcf-99f3a34d35f8 : false
08:12:04.224 : DEBUG -   End
08:12:04.224 : DEBUG - Driver DisplayEngine (C09E4CE4-AE29-4340-B941-078871BD89D1) found but not dispatched. Protocols present:
08:12:04.224 : DEBUG -   ef9fc172-a1b2-4693-b327-6d32fc416042 : true
08:12:04.224 : DEBUG -   587e72d7-cc50-4f79-8209-ca291fc1a10f : true
08:12:04.224 : DEBUG -   a770c357-b693-4e6d-a6cf-d21c728e550b : true
08:12:04.232 : DEBUG -   9d400d20-6f35-4268-904f-dc04b1877b62 : false
08:12:04.232 : DEBUG -   13a3f0f6-264a-3ef0-f2e0-dec512342f34 : true
08:12:04.232 : DEBUG -   fc111ae5-f073-4348-aa5c-49124bbfed7b : true
08:12:04.232 : DEBUG -   0fd96974-23aa-4cdc-b9cb-98d17750322a : true
08:12:04.232 : DEBUG -   End
08:12:04.256 : DEBUG - ..
08:12:04.257 : DEBUG - End Report of Drivers Not Dispatched.
08:12:04.257 : [Bds] Entry...
```

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on real hardware.

## Integration Instructions

No action needed.

### DIFF
--- a/core/patina_internal_depex/src/lib.rs
+++ b/core/patina_internal_depex/src/lib.rs
@@ -282,6 +282,7 @@ impl Depex {
         false
     }
 
+    /// Returns an iterator over internal opcodes.
     pub fn iter(&self) -> impl Iterator<Item = &Opcode> {
         self.expression.iter()
     }
--- a/core/patina_internal_depex/src/lib.rs
+++ b/core/patina_internal_depex/src/lib.rs
@@ -282,25 +282,8 @@ impl Depex {
         false
     }
 
-    /// Report protocol present status in the DEPEX expression.
-    pub fn report_protocol_present(&self) {
-        for opcode in self.expression.iter() {
-            match opcode {
-                Opcode::Push(guid, present) => {
-                    log::warn!(
-                        target: "patina_dxe_core",
-                        "  {guid:?} : {present}"
-                    );
-                }
-                Opcode::End => {
-                    log::warn!(
-                        target: "patina_dxe_core",
-                        "  {opcode:?}"
-                    );
-                }
-                _ => {}
-            }
-        }
+    pub fn iter(&self) -> impl Iterator<Item = &Opcode> {
+        self.expression.iter()
     }
 
     /// If the depex expression is an associated dependency, it returns the associated dependency.
--- a/core/patina_internal_depex/src/lib.rs
+++ b/core/patina_internal_depex/src/lib.rs
@@ -282,6 +282,27 @@ impl Depex {
         false
     }
 
+    /// Report protocol present status in the DEPEX expression.
+    pub fn report_protocol_present(&self) {
+        for opcode in self.expression.iter() {
+            match opcode {
+                Opcode::Push(guid, present) => {
+                    log::warn!(
+                        target: "patina_dxe_core",
+                        "  {guid:?} : {present}"
+                    );
+                }
+                Opcode::End => {
+                    log::warn!(
+                        target: "patina_dxe_core",
+                        "  {opcode:?}"
+                    );
+                }
+                _ => {}
+            }
+        }
+    }
+
     /// If the depex expression is an associated dependency, it returns the associated dependency.
     pub fn is_associated(&self) -> Option<AssociatedDependency> {
         match self.expression.first() {
--- a/patina_dxe_core/src/pi_dispatcher.rs
+++ b/patina_dxe_core/src/pi_dispatcher.rs
@@ -110,13 +110,34 @@ impl<P: PlatformInfo> PiDispatcher<P> {
 
     /// Displays drivers that were discovered but not dispatched.
     pub fn display_discovered_not_dispatched(&self) {
-        for driver in &self.dispatcher_context.lock().pending_drivers {
+        // Summary report
+        for driver in self.dispatcher_context.lock().pending_drivers.iter() {
             log::warn!(
                 "Driver {} ({:?}) found but not dispatched.",
                 driver.name.as_deref().unwrap_or("Unnamed"),
                 guid_fmt!(driver.file_name)
             );
         }
+
+        // Detail report
+        log::warn!("Begin Detail Report:");
+        for driver in self.dispatcher_context.lock().pending_drivers.iter() {
+            log::warn!(
+                "Driver {} ({:?}) found but not dispatched. Protocols present:",
+                driver.name.as_deref().unwrap_or("Unnamed"),
+                guid_fmt!(driver.file_name)
+            );
+
+            match &driver.depex {
+                Some(depex) => {
+                    depex.report_protocol_present();
+                }
+                _ => {
+                    log::warn!("  No Depex");
+                }
+            }
+        }
+        log::warn!("End Detail Report.");
     }
 
     /// Initializes the dispatcher by registering for FV protocol installation events.
--- a/patina_dxe_core/src/pi_dispatcher.rs
+++ b/patina_dxe_core/src/pi_dispatcher.rs
@@ -120,24 +120,35 @@ impl<P: PlatformInfo> PiDispatcher<P> {
         }
 
         // Detail report
-        log::warn!("Begin Detail Report:");
+        self.detail_report();
+    }
+
+    fn detail_report(&self) {
+        log::debug!("Begin Report of Drivers Not Dispatched:");
         for driver in self.dispatcher_context.lock().pending_drivers.iter() {
-            log::warn!(
+            log::debug!(
                 "Driver {} ({:?}) found but not dispatched. Protocols present:",
                 driver.name.as_deref().unwrap_or("Unnamed"),
                 guid_fmt!(driver.file_name)
             );
 
-            match &driver.depex {
-                Some(depex) => {
-                    depex.report_protocol_present();
+            if let Some(depex) = &driver.depex {
+                for opcode in depex.iter() {
+                    match opcode {
+                        Opcode::Push(guid, present) => {
+                            log::debug!("  {guid:?} : {present}");
+                        }
+                        Opcode::End => {
+                            log::debug!("  {opcode:?}");
+                        }
+                        _ => {}
+                    }
                 }
-                _ => {
-                    log::warn!("  No Depex");
-                }
+            } else {
+                log::debug!("  No Depex");
             }
         }
-        log::warn!("End Detail Report.");
+        log::debug!("End Report of Drivers Not Dispatched.");
     }
 
     /// Initializes the dispatcher by registering for FV protocol installation events.


patina_dxe_core/src/pi_dispatcher.rs
```diff
@@ -112,6 +112,27 @@ impl<P: PlatformInfo> PiDispatcher<P> {
...
+            for driver in &mut self.dispatcher_context.lock().pending_drivers {
+                log::warn!(
+                    "Driver {} ({:?}) found but not dispatched.",
+                    driver.name.as_deref().unwrap_or("Unnamed"),
+                    guid_fmt!(driver.file_name)
+                );
+
+                match &mut driver.depex {
+                    Some(depex) => {
+                        depex.eval_with_report(&PROTOCOL_DB.registered_protocols());
```
os-d: Rather than this approach, which will print out the hard to parse depex report, e.g.:

Image

I would suggest we print out something nicer here. I know this can get a little hairy because depex statements are allowed to get complex, but in practice we typically don't see very complex ones.

i.e. instead of the stack of true and false, I would hope to see a report like:

Driver X failed to dispatch because GUID Y was not found.
Driver Z failed to dispatch because GUID alpha was found and GUID beta was not found.

etc. I think that is feasible and a lot easier to parse from the logs.

@Javagedes @makubacki thoughts?

makubacki: I agree that should be explored here while the feature is being added. We should be able to map back image names from PDB paths and/or UI sections to the module GUID.

yangrongwei: @os-d 
I’ll figure out a way to make this nicer, based on my understanding.

@makubacki 
I’ll keep this in mind. I also have a more aggressive idea.

yangrongwei: If we prefer "nicer", to balance verbosity and correctness, my understanding is this:
Trying to construct an expression tree from the DEPEX op list, then report the false node in the evaluation.
I feel this might be a bit too aggressive for debug code.

This problem shifts from the computer engineering domain to the computer science domain.
We are trying to touch the SAT (Boolean Satisfiability) problem.

After reviewing the cost-benefit tradeoffs against our purpose, we decided to use a "simpler" format. This format reports the present status of dependent protocols. Engineers can then open the corresponding INF file (or build report) and perform a one-to-one mapping check.

One ref for SAT: https://people.eecs.berkeley.edu/~sseshia/219c/spr16/lectures/SATSolving.pdf

yangrongwei: This is the updated format.  

a) The detail report section is marked as Debug.
b) Intentionally use the same leading description for DXE in the summary table.

image

patina_dxe_core/src/pi_dispatcher.rs
```diff
@@ -112,6 +112,30 @@ impl<P: PlatformInfo> PiDispatcher<P> {
...
         }
+
+        const EVAL_REPORT: &str = "patina_internal_depex.report";
+
+        if log::log_enabled!(target: EVAL_REPORT, log::Level::Debug) {
+            log::debug!(target: EVAL_REPORT, "Begin Detail Report:");
+            for driver in self.dispatcher_context.lock().pending_drivers.iter() {
+                log::debug!(
+                    target: EVAL_REPORT,
+                    "Driver {} ({:?}) found but not dispatched. Protocols present:",
```
joschock: Should we keep the original "found but not dispatched" (without depex summary) if patina_internal_depx.report is not set?

joschock (approved):

patina_dxe_core/src/pi_dispatcher.rs
yangrongwei: It’s already using this logic now.

Summary report is controlled by "patina_dxe_core" (default target)
Detail report is controlled by **"patina_depex_report"**
Eval steps are controlled by "patina_internal_depx" (default target)

**Updates** of Commits f02fd80  on Jun 8, 2026: 
Changed the target name to "patina_internal_depx"

<del> Detail report is controlled by "patina_internal_depx.report" </del>

<del> The tricky point is that `starts_with()` of logger filter enforces us to put the "patina_internal_depx.report" before "patina_internal_depx" in the config wrapper unless we use a different prefixed target name OR change the filter matching algorithm. </del>

image

==

image

yangrongwei: @antklein @os-d @makubacki @joschock 

**Updates** of Commits f02fd80  on Jun 8, 2026: 
Changed the target name to "patina_internal_depx"

**For two reasons:**
The filter is enabled by default for Debug builds without modifying existing wrapper configurations.
It is straightforward for users, avoiding unnecessary nesting and tricky ordering in the wrapper configuration.

**Con:**
Introduces a top-level filter.

joschock: ah, I missed the initial print; it was hidden in the default view in git diff.

core/patina_internal_depex/src/lib.rs
```diff
@@ -282,6 +282,27 @@ impl Depex {
         false
     }
 
+    /// Report protocol present status in the DEPEX expression.
+    pub fn report_protocol_present(&self) {
+        for opcode in self.expression.iter() {
+            match opcode {
+                Opcode::Push(guid, present) => {
+                    log::debug!(
+                        target: "patina_depex_report",
```
os-d: I think this information is generally useful, I don’t think it needs to be hidden behind a target

os-d (approved):

os-d:
@Javagedes @makubacki can one of you please review?

core/patina_internal_depex/src/lib.rs
yangrongwei: "patina_depex_report" target removed.

Using the same target and level of the summary section, which is `log::warn!(target: "patina_dxe_core"`

image

makubacki:
> @Javagedes @makubacki can one of you please review?

I was out of office Mon & Tue. I'm reviewing now.

core/patina_internal_depex/src/lib.rs
```diff
@@ -282,6 +282,27 @@ impl Depex {
...
 
+    /// Report protocol present status in the DEPEX expression.
+    pub fn report_protocol_present(&self) {
+        for opcode in self.expression.iter() {
+            match opcode {
+                Opcode::Push(guid, present) => {
+                    log::warn!(
+                        target: "patina_dxe_core",
+                        "  {guid:?} : {present}"
+                    );
```
makubacki: Having this at `warn` level coupled with the `patina_dxe_core` target is a code smell in terms of coupling. Basically, this assumes code exists a certain way in the code (`patina_dxe_core`) that uses this crate (`patina_internal_depex`) as a dependency.

This type of print should be at least `info` or a more verbose type if left at the `patina_dxe_core` target level.

But I really think this code should really just return something like an iterator of protocols present for a given `Depex` so all logging stays in the caller.

patina_dxe_core/src/pi_dispatcher.rs
```diff
@@ -105,13 +105,34 @@ impl<P: PlatformInfo> PiDispatcher<P> {
...
+                Some(depex) => {
+                    depex.report_protocol_present();
+                }
+                _ => {
+                    log::warn!("  No Depex");
+                }
+            }
+        }
+        log::warn!("End Detail Report.");
     }
```
makubacki, line 152: The messages in this "detailed report" should be at `info` level or higher. I think `debug` level.

The normal case of a driver not dispatched is already covered by `warn` above. This is debug information.

patina_dxe_core/src/pi_dispatcher.rs
```diff
@@ -105,13 +105,34 @@ impl<P: PlatformInfo> PiDispatcher<P> {
...
+        for driver in self.dispatcher_context.lock().pending_drivers.iter() {
             log::warn!(
                 "Driver {} ({:?}) found but not dispatched.",
                 driver.name.as_deref().unwrap_or("Unnamed"),
                 guid_fmt!(driver.file_name)
             );
         }
+
+        // Detail report
+        log::warn!("Begin Detail Report:");
```
makubacki: "Begin Detail Report" is too ambiguous. Please change to something like "Begin Report of Drivers Not Dispatched".

makubacki (changes requested):

makubacki:
@yangrongwei, can you please put an example of output in the PR description to help reviewers and those reading the release notes (that come from the PR description) to know what to expect?

core/patina_internal_depex/src/lib.rs
os-d: Agreed. I just meant don't use a target at all for the logging. It will inherit the patina_internal_depex target if a consumer wants to filter it.

But, to Michael's point, I think that returning an iterator here and letting the caller print it works well.

makubacki:
@yangrongwei, please rebase your branch locally for your next push. It is 59 commits behind now and that should allow it to run against the latest CI checks.

core/patina_internal_depex/src/lib.rs
yangrongwei: Done.  Removed the reporting code and return an interator now.

patina_dxe_core/src/pi_dispatcher.rs
yangrongwei: Using `Debug` level now.

patina_dxe_core/src/pi_dispatcher.rs
yangrongwei: Changed to your suggestion.

yangrongwei:
> @yangrongwei, can you please put an example of output in the PR description to help reviewers and those reading the release notes (that come from the PR description) to know what to expect?


> @yangrongwei, please rebase your branch locally for your next push. It is 59 commits behind now and that should allow it to run against the latest CI checks.

@makubacki 
Done. Please review.

patina_dxe_core/src/pi_dispatcher.rs
makubacki: Thanks!

makubacki:
@yangrongwei, thanks for the updates.

I have a proposal I'd like to make to the output format, but I don't think that needs to be made in this PR as this is useful as-is and it might need further discussion that doesn't need to hold up initial support. But I would be interested in your feedback on my proposed changes though.

makubacki (approved):

PR #1554, opened 2026-06-03
[REBASE & FF] Add Stack Overflow Hint in Exception Handlers

os-d:
## Description

Stack overflows show up as page faults in the exception handler, which can be hard to distinguish from other kinds of page faults. This PR contains three commits to provide a hint that a stack overflow occurred.

x64: IDT: Use a Separate Stack for Page Fault Exceptions
--
Currently, only double faults use a separate stack. However, a stack overflow manifests as a page fault, so taking a stack overflow causes a double fault. This makes it harder to reason about what the fault was.

This commit instead using a separate stack for x64 page faults, aligning with the aarch64 side. This is in preparation for giving better information about stack overflows in the exception handlers.

exception handlers: Print Stack Overflow Hint
--
Currently, stack overflows show up as page faults in the exception handlers. This can make it hard to identify when a different page fault has occurred.

This commit introduces a heuristic to hint that a stack overflow likely occurred in the exception handler. It checks if the faulting address is on the same page as or one page lower than the stack pointer. In some cases, the stack pointer is incremented first, then data is written (in which case the faulting address is on the same page as the stack pointer) and in other cases the data is written before the stack pointer is decremented (so the faulting address is on the page below the stack pointer).

A pretty message is then printed to indicate a stack overflow occurred.

internal_cpu: Rename EfiExceptionStackTrace Trait
--
The EfiExceptionStackTrace trait is an internal trait that originally was just used so the common exception handler can use arch specific stack trace dumping.

However, this was expanded to also include dumping a page table walk, so the name is not accurate.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

In Q35 and SBSA testing stack overflows which now look like:

image

image

## Integration Instructions

N/A.

### DIFF
--- a/core/patina_internal_cpu/src/interrupts/x64/idt.rs
+++ b/core/patina_internal_cpu/src/interrupts/x64/idt.rs
@@ -83,8 +83,8 @@ pub fn initialize_idt() {
 
     // Point every vector at its corresponding assembly handler.
     for vector in 0..=255usize {
-        // Use IST 1 for double fault (vector 8) to ensure it has a valid stack
-        let ist_index = if vector == 8 { 1 } else { 0 };
+        // Use IST 1 for double fault (vector 8) and page fault (vector 14) to ensure they have a valid stack.
+        let ist_index = if vector == 8 || vector == 14 { 1 } else { 0 };
         idt.entries.get_mut(vector).expect("vector out of bounds").set_handler(
             get_vector_address(vector),
             cs,
--- a/core/patina_internal_cpu/src/interrupts/x64/interrupt_manager.rs
+++ b/core/patina_internal_cpu/src/interrupts/x64/interrupt_manager.rs
@@ -18,7 +18,7 @@ use patina_mtrr::Mtrr;
 use patina_paging::PageTable;
 use patina_stacktrace::{StackFrame, StackTrace};
 
-use crate::interrupts::{EfiExceptionStackTrace, HandlerType, InterruptManager, x64::ExceptionContextX64};
+use crate::interrupts::{EfiExceptionInfoDump, HandlerType, InterruptManager, x64::ExceptionContextX64};
 
 /// X64 Implementation of the InterruptManager.
 ///
--- a/core/patina_internal_cpu/src/interrupts/x64/interrupt_manager.rs
+++ b/core/patina_internal_cpu/src/interrupts/x64/interrupt_manager.rs
@@ -125,7 +125,21 @@ extern "efiapi" fn page_fault_handler(_exception_type: isize, context: EfiSystem
     // to report. The system is dead anyway.
     let x64_context = unsafe { context.system_context_x64.as_ref().unwrap() };
 
-    log::error!("EXCEPTION: PAGE FAULT");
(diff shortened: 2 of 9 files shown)


core/patina_internal_cpu/src/interrupts/x64/idt.rs
```diff
@@ -83,8 +83,8 @@ pub fn initialize_idt() {
 
     // Point every vector at its corresponding assembly handler.
     for vector in 0..=255usize {
-        // Use IST 1 for double fault (vector 8) to ensure it has a valid stack
-        let ist_index = if vector == 8 { 1 } else { 0 };
+        // Use IST 1 for double fault (vector 8) and page fault (vector 14) to ensure they have a valid stack.
+        let ist_index = if vector == 8 || vector == 14 { 1 } else { 0 };
```
cfernald, line 87: Did you confirm that a debugger breaking in on an page fault operates within our stack size? It should, but would be good to confirm.

cfernald (approved):

core/patina_internal_cpu/src/interrupts/x64/idt.rs
os-d: Yes, all looked good

core/patina_internal_cpu/src/interrupts/x64/interrupt_manager.rs
```diff
@@ -125,7 +125,20 @@ extern "efiapi" fn page_fault_handler(_exception_type: isize, context: EfiSystem
...
     let x64_context = unsafe { context.system_context_x64.as_ref().unwrap() };
 
-    log::error!("EXCEPTION: PAGE FAULT");
+    // Stack overflows cause page faults, let's give a hint if it looks like one.
+    // The heuristic we are using is that stack overflow will either have CR2 on the same page
+    // as RSP or directly below it. Sometimes RSP is decremented first, sometimes second. Stack overflows
+    // are always data access, so ensure that is the case as well.
+    if (((x64_context.cr2 & !UEFI_PAGE_MASK as u64) == (x64_context.rsp & !UEFI_PAGE_MASK as u64))
+        || ((x64_context.cr2 & !UEFI_PAGE_MASK as u64)
+            == ((x64_context.rsp & !UEFI_PAGE_MASK as u64) - UEFI_PAGE_SIZE as u64)))
```
makubacki: You could also clean this up a bit with some variables:

- `let rsp_page = x64_context.rsp & !(UEFI_PAGE_MASK as u64);`
- `let cr2_page = x64_context.cr2 & !(UEFI_PAGE_MASK as u64);`
- `let same_or_below = cr2_page == rsp_page || cr2_page == rsp_page.wrapping_sub(UEFI_PAGE_SIZE as u64);`

makubacki (approved):

core/patina_internal_cpu/src/interrupts/x64/interrupt_manager.rs
os-d: Cleaned this up for both x64 and aarch64.

PR #1560, opened 2026-06-15
uart: bump uart_16550 dependency

phip1611:
## Description

Since v0.5.0, the crate got a major overhaul with a focus to work on physical hardware next to VMs. I've bumped the dependency to 0.6.0.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

I simply bumped the crate without major testing (I am relying on the CI). I am the author of `uart_16550` and tested this in various VMs (QEMU, Cloud Hypervisor) and also on real x86 hardware. I just bumped the dependency here in a straight forward way.

## Integration Instructions

- N/A

### DIFF
--- a/core/patina_debugger/README.md
+++ b/core/patina_debugger/README.md
@@ -43,7 +43,8 @@ The self-hosted debugger is lightweight and tightly integrated with Patina, offe
 
 ## Platform Integration
 
-1. Instantiate a `PatinaDebugger` with the platform UART configuration (for example, `Uart16550::Io { base: 0x3F8 }`).
+1. Instantiate a `PatinaDebugger` with the platform UART configuration (for example,
+   `unsafe { Uart16550::new_port(0x3F8) }`).
 2. Apply any policy overrides such as `.with_force_enable`, `.with_log_policy`, or `.with_transport_init` when
    if the debugger must initialize the transport.
 3. Register the debugger using `patina_debugger::set_debugger(&DEBUGGER)` before the Patina DXE Core starts dispatching
--- a/docs/src/dev/principles/abstractions.md
+++ b/docs/src/dev/principles/abstractions.md
@@ -60,9 +60,11 @@ a simple stdio writer.
 ``` rust
 # extern crate log;
 # extern crate core;
+# extern crate spin;
 # extern crate uart_16550;
 // The starting abstraction point, the `Log` trait
 use log::Log;
+use spin::Mutex;
 use std::io::{Read, Write};
 
 /// Our Abstraction point for implementing different ways to perform a serial write
@@ -75,23 +77,23 @@ pub trait SerialIO {
 
 pub struct SerialLogger<S>
 where
-    S: SerialIO + Send + Sync,
+    S: SerialIO + Send,
 {
     /// An implementation of the abstraction point
-    serial: S,
+    serial: Mutex<S>,
     /// Will not log messages above this level
     max_level: log::LevelFilter,
 }
 
 impl<S> SerialLogger<S>
 where
-    S: SerialIO + Send + Sync,
+    S: SerialIO + Send,
 {
     pub const fn new(
         serial: S,
         max_level: log::LevelFilter,
     ) -> Self {
-        Self { serial, max_level }
+        Self { serial: Mutex::new(serial), max_level }
     }
 }
 
@@ -99,7 +101,7 @@ where
 // be implemented to complete the interface implementation
 impl<S> Log for SerialLogger<S>
 where
-    S: SerialIO + Send + Sync,
+    S: SerialIO + Send,
 {
     fn enabled(&self, metadata: &log::Metadata) -> bool {
         return metadata.level().to_level_filter() <= self.max_level
@@ -109,7 +111,7 @@ where
         let formatted = format!("{} - {}\n", record.level(), record.args());
         // We know our "serial" object must have the "write" function, we just don't know the
         // implementation details, which is fine.
-        self.serial.write(&formatted.into_bytes());
+        self.serial.lock().write(&formatted.into_bytes());
     }
 
     fn flush(&self) {}
@@ -142,40 +144,32 @@ impl SerialIO for Terminal {
     }
 }
 
-use uart_16550::MmioSerialPort;
-struct Uart16550(usize);
+use core::ptr::{NonNull, with_exposed_provenance_mut};
+use uart_16550::backend::MmioBackend;
+use uart_16550::{Config, Uart16550};
 
-impl Uart16550 {
-    fn new(addr: usize) -> Self {
-        Self(addr)
-    }
-}
+struct MmioUart(Mutex<Uart16550<MmioBackend>>);
 
-impl SerialIO for Uart16550 {
+impl SerialIO for MmioUart {
     fn init(&self) {
-        unsafe { MmioSerialPort::new(self.0).init() };
+        self.0
+            .lock()
+            .init(Config::default())
+            .expect("UART 16550 MMIO device initialization must succeed");
     }
 
     fn write(&self, buffer: &[u8]) {
-        let mut port = unsafe { MmioSerialPort::new(self.0) };
-
-        for b in buffer {
-            port.send(*b);
-        }
+        self.0.lock().send_bytes_exact(buffer);
     }
 
     fn read(&self) -> u8 {
-        let mut port = unsafe { MmioSerialPort::new(self.0) };
-        port.receive()
+        let mut byte = [0];
+        self.0.lock().receive_bytes_exact(&mut byte);
+        byte[0]
     }
 
     fn try_read(&self) -> Option<u8> {
-        let mut port = unsafe { MmioSerialPort::new(self.0) };
-        if let Ok(value) = port.try_receive() {
-            Some(value)
-        } else {
-            None
-        }
+        self.0.lock().try_receive_byte().ok()
     }
 }
 
@@ -183,6 +177,14 @@ impl SerialIO for Uart16550 {
 fn main() {
     let terminal_logger = SerialLogger::new(Terminal, log::LevelFilter::Trace);
 
-    let uart16550_logger = SerialLogger::new(Uart16550::new(0x4000), log::LevelFilter::Trace);
+    let uart16550_logger = SerialLogger::new(
+        MmioUart(Mutex::new(unsafe {
+            let base = NonNull::new(with_exposed_provenance_mut::<u8>(0x4000))
+                .expect("UART 16550 MMIO base address must be non-null");
+            Uart16550::new_mmio(base, 1)
+                .expect("UART 16550 MMIO base address and register stride must be valid")
+        })),
+        log::LevelFilter::Trace,
+    );
 }
 ```
--- a/sdk/patina/src/serial/uart.rs
+++ b/sdk/patina/src/serial/uart.rs
@@ -29,8 +29,10 @@ impl super::SerialIO for UartNull {
 cfg_if::cfg_if! {
     if #[cfg(all(target_arch = "x86_64", any(target_os = "uefi", feature = "doc")))] {
 
-        use uart_16550::MmioSerialPort;
-        use uart_16550::SerialPort as IoSerialPort;
+        use core::ptr::{NonNull as PtrNonNull, with_exposed_provenance_mut};
+        use spin::Mutex;
+        use uart_16550::backend::{MmioBackend, PioBackend};
+        use uart_16550::{Config, Uart16550 as Uart16550Device};
 
         /// Returns the Current Privilege Level (CPL) from the CS selector.
         fn current_privilege_level() -> u16 {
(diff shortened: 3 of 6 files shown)


sdk/patina/src/serial/uart.rs
```diff
@@ -62,46 +64,62 @@ cfg_if::cfg_if! {
...
         /// An interface for writing to a Uart16550 device.
         #[derive(Debug)]
-        pub enum Uart16550 {
+        pub enum Uart16550Wrapper {
             /// The I/O interface for the Uart16550 serial port.
-            Io {
-                /// The base address of the UART control registers.
-                base: u16
-            },
+            Io(Mutex<Uart16550Device<PioBackend>>),
```
phip1611, line 69: hint for reviewers. this is now a mutex as we need `&mut self` in several methods

os-d: Thanks for the PR, appreciate the contribution!

I think using the mutex is okay here because we are calling this without interrupts. That's also true in MM, right @kuqin12?

@makubacki @cfernald any concerns on using mutex here?

makubacki: Regarding the interrupt in MM question, an SMI is non-maskable, all interrupts are pended while in SMM, and code executes on a single core.

I am a little concerned with the number of `expect()` messages (potentially bricking boot due to UART init) and the fallibility of the code though and potential re-entrant paths though errors being logged in panic handlers. 

---

Do we see any major downsides with `try_lock()`?

For example:

- Lazy init `Uart16550`:

  ```rust
        #[derive(Debug)]
        pub enum Uart16550 {
            /// The x86 port-I/O backed serial port.
            Io {
                /// Base I/O port of the UART register block.
                base: u16,
                /// The device, constructed on the first successful `init`.
                device: Mutex<Option<Uart16550Device<PioBackend>>>,
            },
            /// The memory-mapped serial port.
            Mmio {
                /// Base address of the UART register block.
                base: usize,
                /// Byte distance between consecutive registers.
                reg_stride: u8,
                /// The device, constructed on the first successful `init`.
                device: Mutex<Option<Uart16550Device<MmioBackend>>>,
            },
        }
  ```
- Make `new_port()` and `new_mmio()` `const`:

  ```rust
            pub const unsafe fn new_port(base: u16) -> Self {
                Self::Io { base, device: Mutex::new(None) }
            }
  ```

  ```rust
            pub const unsafe fn new_mmio(base: usize, reg_stride: u8) -> Self {
                Self::Mmio { base, reg_stride, device: Mutex::new(None) }
            }
  ```
- Use `try_lock()` in `impl super::SerialIO) for Uart16550`:

  ```rust
        impl super::SerialIO for Uart16550 {
            fn init(&self) {
                match self {
                    Uart16550::Io { base, device } => {
                        let Some(mut guard) = device.try_lock() else { return };
                        // SAFETY: `new_port`'s safety contract requires `base` is valid
                        // 16550 I/O port that is safe to access.
                        match unsafe { Uart16550Device::new_port(*base) } {
                            Ok(mut uart) => match uart.init(line_config()) {
                                Ok(()) => *guard = Some(uart),
                                Err(err) => log::error!("UART 16550 I/O initialization failed: {err:?}"),
                            },
                            Err(err) => log::error!("UART 16550 I/O construction failed: {err:?}"),
                        }
                    }
                    // ...
  ```
- `write()` also can handle not acquiring the lock:

  ```rust
            fn write(&self, buffer: &[u8]) {
                let send = || match self {
                    Uart16550::Io { device, .. } => {
                        if let Some(mut guard) = device.try_lock()
                            && let Some(uart) = guard.as_mut()
                        {
                            uart.send_bytes_exact(buffer);
                        }
                    }
  ```

- Similarly use `try_lock()` in `read()` and `try_read()`

phip1611: > I am a little concerned with the number of expect() messages (potentially bricking boot due to UART init) and the fallibility of the code though and potential re-entrant paths though errors being logged in panic handlers.

I understand that concern.

I am currently maintaining `uart_16550`, and the newer versions, starting with `0.5.0`, are a complete rewrite of the crate. The goal of that rewrite was to make the crate compatible with real hardware, as the previous implementation was not flexible enough for that use case.

As a result, the new API necessarily exposes more of the underlying complexity, including possible failure cases that were previously silent.

For this dependency bump, I chose the least invasive approach.

> Do we see any major downsides with try_lock()?

nope

cfernald: Mostly, though there is an edge case with the read routines. If you were to take an interrupt while reading from the core (say for a serial console) then you would hit deadlocks if trying to write to the log in the interrupt. Those should probably be converts to without interrupts too.

makubacki: @phip1611, my understanding is that you're planning to refactor to the `try_lock()` approach, is that right?

phip1611: Yup 👍

sdk/patina/src/serial/uart.rs
```diff
@@ -64,44 +66,64 @@ cfg_if::cfg_if! {
         #[derive(Debug)]
         pub enum Uart16550 {
             /// The I/O interface for the Uart16550 serial port.
-            Io {
-                /// The base address of the UART control registers.
-                base: u16
-            },
+            Io(Mutex<Uart16550Device<PioBackend>>),
```
phip1611, line 69: I wonder if we should use a Mutex on a higher level instead but I think this is out of scope for this PR

cfernald: Agreed, actually filed this yesterday as well coming to the same conclusion in a similar PR.

https://github.com/OpenDevicePartnership/patina/issues/1567

makubacki:
@phip1611, the "How This Was Tested" and "Integration Instructions" sections in the PR description were not filled in. I added some text based on what I know from PR validation tests, but please adjust them if needed.

sdk/patina/src/serial/uart.rs
makubacki: Thanks for the update. Resolving based on latest.

sdk/patina/src/serial/uart.rs
```diff
@@ -64,44 +66,64 @@ cfg_if::cfg_if! {
...
-                        // SAFETY: The base address is provided during Uart16550 construction and is assumed to be valid for I/O port access.
-                        let mut serial_port = unsafe { IoSerialPort::new(*base) };
-                        let mut send = || {
-                            for b in buffer {
-                                serial_port.send(*b);
-                            }
+                    Uart16550::Io(uart) => {
+                        let send = || {
+                            uart.try_lock()
+                                .expect("UART 16550 I/O device lock must not be re-entered")
```
makubacki: We have to be very careful with panics in production firmware because it will potentially brick a system. I was thinking that if the lock cannot be acquired either return or maybe use something like `debug_assert!()` to identify the problem in debug scenarios.

phip1611: I see - thanks for the reminder!

sdk/patina/src/serial/uart.rs
phip1611: Just to clarify: In this PR I stick to the Mutex and in a follow-up (not done by me) this is changed as described in #1567?

sdk/patina/src/serial/uart.rs
makubacki: @phip1611, are you still planning to make an update?

sdk/patina/src/serial/uart.rs
cfernald: Correct, please continue with your current approach here. We will make this consistent across implementations after this is merged.

sdk/patina/src/serial/uart.rs
phip1611: Hi, sorry for the delay. I have that on my desk for the next work days!

makubacki: Just checking in, no problem. Thanks!

docs/src/dev/principles/abstractions.md
```diff
@@ -142,47 +144,47 @@ impl SerialIO for Terminal {
...
     }
 }
 
 // Now we can initialize them with our implementations
 fn main() {
     let terminal_logger = SerialLogger::new(Terminal, log::LevelFilter::Trace);
 
-    let uart16550_logger = SerialLogger::new(Uart16550::new(0x4000), log::LevelFilter::Trace);
+    let uart16550_logger = SerialLogger::new(
+        MmioUart(Mutex::new(unsafe {
```
phip1611, line 181: Full transparency for reviewers: For `abstractions.md`, I've used an LLM to generate and streamline the code.

sdk/patina/src/serial/uart.rs
phip1611: I removed the panicking for read() and write() and kept it for the fundamental new() and init() - is this what you've intended?

phip1611:
> @phip1611, the "How This Was Tested" and "Integration Instructions" sections in the PR description were not filled in. I added some text based on what I know from PR validation tests, but please adjust them if needed.

Updated PR description

cfernald:
> > @phip1611, the "How This Was Tested" and "Integration Instructions" sections in the PR description were not filled in. I added some text based on what I know from PR validation tests, but please adjust them if needed.
> 
> Updated PR description

@phip1611, I can do a sanity check validation on our system after the builds are fixed.

phip1611:
> > > @phip1611, the "How This Was Tested" and "Integration Instructions" sections in the PR description were not filled in. I added some text based on what I know from PR validation tests, but please adjust them if needed.
> > 
> > 
> > Updated PR description
> 
> @phip1611, I can do a sanity check validation on our system after the builds are fixed.

Yikes, sorry. Developer Experience was a little unusual as my IDE didn't work that nicely in patinas tree (cfg_if etc). I think I have it now!

core/patina_debugger/README.md
```diff
@@ -43,7 +43,7 @@ The self-hosted debugger is lightweight and tightly integrated with Patina, offe
 
 ## Platform Integration
 
-1. Instantiate a `PatinaDebugger` with the platform UART configuration (for example, `Uart16550::Io { base: 0x3F8 }`).
+1. Instantiate a `PatinaDebugger` with the platform UART configuration (for example, `unsafe { Uart16550::new_port(0x3F8) }`).
```
makubacki: To fix this markdownlint issue:

```
core/patina_debugger/README.md:46:121 MD013/line-length Line length [Expected: 120; Actual: 126]
```

```suggestion
1. Instantiate a `PatinaDebugger` with the platform UART configuration
   (for example, `unsafe { Uart16550::new_port(0x3F8) }`).
```

sdk/patina/src/serial/uart.rs
```diff
@@ -64,44 +66,79 @@ cfg_if::cfg_if! {
...
+        impl Uart16550 {
+            /// Constructs a UART backed by x86 port I/O.
+            ///
+            /// # Safety
+            ///
+            /// The base port must be valid and safe to access for the lifetime
+            /// of the returned wrapper.
+            pub unsafe fn new_port(base: u16) -> Self {
+                let uart = unsafe { Uart16550Device::new_port(base) }
+                    .expect("UART 16550 I/O base address must allow access to all registers");
```
makubacki, line 84: This unsafe block will need a SAFETY message.

sdk/patina/src/serial/uart.rs
```diff
@@ -64,44 +66,79 @@ cfg_if::cfg_if! {
...
+            ///
+            /// # Safety
+            ///
+            /// The base address must be valid and safe to access for the
+            /// lifetime of the returned wrapper.
+            pub unsafe fn new_mmio(base: usize, reg_stride: u8) -> Self {
+                let base = PtrNonNull::new(with_exposed_provenance_mut::<u8>(base))
+                    .expect("UART 16550 MMIO base address must be non-null");
+                let uart = unsafe { Uart16550Device::new_mmio(base, reg_stride) }
+                    .expect("UART 16550 MMIO base address and register stride must be valid");
```
makubacki, line 99: This will also need a SAFETY message.

phip1611: adressed

sdk/patina/src/serial/uart.rs
phip1611: addressed

core/patina_debugger/README.md
phip1611: thanks!

makubacki:
@phip1611, in case you didn't see it, looks like there is one more markdown change needed:

```
core/patina_debugger/README.md:46:85 MD009/no-trailing-spaces Trailing spaces [Expected: 0 or 2; Actual: 1]
```

phip1611:
Yikes, CI is still failing :/ My long summer vacation starts tomorrow.. I'll come back to this after my vacation.

I'll reopen this in August. Thanks for your patience and see you soon

PR #1562, opened 2026-06-15
dxe_core: dispatcher: Don't Re-Extract FV Files with FV2/3 HOBs

os-d:
## Description

FV2 and extracted FV3 HOBs are produced by the HOB producer phase to indicate to the HOB consumer phase that an FV is already extracted. Patina has been ignoring this and re-extracting FV files if an FV HOB exists for an FV that contains compressed child FVs. On some platforms this has been observed to add
\>1 second of boot time which is resolved by this change.

This follows what edk2 does and does not re-process child FVs that already have FV2 or extracted FV3 HOBs for them.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on an AARCH64 platform that was seeing a large perf delta between Patina and the C DXE Core. This change reduced overall boot time by 1.2 seconds. Tested on an Intel physical platform that did not have an FV produced for the parent FV of compressed child FVs, so there was no perf delta there.

## Integration Instructions

N/A.

### DIFF
--- a/patina_dxe_core/src/pi_dispatcher.rs
+++ b/patina_dxe_core/src/pi_dispatcher.rs
@@ -23,9 +23,14 @@ use alloc::{
 use core::{cmp::Ordering, ffi::c_void};
 use mu_rust_helpers::guid::guid_fmt;
 use patina::{
+    BinaryGuid,
     device_path::walker::concat_device_path_to_boxed_slice,
     error::EfiError,
-    pi::{fw_fs::ffs, hob::HobList, protocols::firmware_volume_block},
+    pi::{
+        fw_fs::ffs,
+        hob::{Hob, HobList},
+        protocols::firmware_volume_block,
+    },
 };
 use patina_ffs::{
     section::{Section, SectionExtractor},
@@ -138,6 +143,9 @@ impl<P: PlatformInfo> PiDispatcher<P> {
                 Some(efi::EVENT_GROUP_EXIT_BOOT_SERVICES),
             )
             .expect("Failed to create callback for runtime image memory protection fixups.");
+
+        // Get the FV2/3 HOBs before we register for FV protocol installation callbacks
+        self.dispatcher_context.lock().cache_pre_extracted_fv_hobs(hob_list);
 
         //set up call back for FV protocol installation.
         let event = EVENT_DB
@@ -581,6 +589,7 @@ struct DispatcherContext {
     pending_drivers: Vec<PendingDriver>,
     fv_section_data: Vec<Box<[u8]>>,
     pending_firmware_volume_images: Vec<PendingFirmwareVolumeImage>,
+    pre_extracted_fv_hobs: BTreeSet<(BinaryGuid, BinaryGuid)>,
     associated_before: BTreeMap<OrdGuid, Vec<PendingDriver>>,
     associated_after: BTreeMap<OrdGuid, Vec<PendingDriver>>,
     processed_fvs: BTreeSet<efi::Handle>,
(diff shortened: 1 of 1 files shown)


patina_dxe_core/src/pi_dispatcher.rs
```diff
@@ -760,6 +770,32 @@ impl DispatcherContext {
                         let file = file.clone();
                         let file_name = file.name().into_inner();
 
+                        // if we have an FV2 or extracted FV3 HOB for this FV, it has already been decompressed and we
```
joschock: It seems a bit weird to thread the whole hoblist as context through the fv handles. What about processing the hoblist at Dispatcher init time and storing the "pre-extracted" FV data in the dispatcher and then matching against that dispatcher state instead? This would avoid retaining a hoblist pointer for the lifetime of the dispatcher and threading it through the event dispatch...

os-d: I also didn’t like this. I do think it is odd that dxe core doesn’t have a good way of getting the hob list other than just passing it around. And clearly instead of introducing a general way to retrieve it, I was lazy :).

But I think it is reasonable for the dispatcher to just grab the hobs it cares about and store them, I’ll make the change.

os-d: @joschock made the change.

joschock (approved):

cfernald (approved):

patina_dxe_core/src/pi_dispatcher.rs
makubacki, line 1: @os-d, please let me finish an in-progress review before mering.

patina_dxe_core/src/pi_dispatcher.rs
```diff
@@ -604,6 +614,19 @@ impl DispatcherContext {
...
+        self.pre_extracted_fv_hobs.extend(hob_list.iter().filter_map(|hob| match hob {
+            Hob::FirmwareVolume2(fv2) => Some((fv2.file_name, fv2.fv_name)),
+            Hob::FirmwareVolume3(fv3) if fv3.extracted_fv.into() => Some((fv3.file_name, fv3.fv_name)),
+            _ => None,
+        }));
+    }
+
+    fn has_pre_extracted_fv_hob(&self, file_name: efi::Guid, fv_name: Option<BinaryGuid>) -> bool {
+        let default_fv_name = BinaryGuid::from_string("00000000-0000-0000-0000-000000000000");
+        self.pre_extracted_fv_hobs.contains(&(file_name.into(), fv_name.unwrap_or(default_fv_name)))
```
makubacki: Can you use `patina::guids::ZERO`?


```suggestion
        self.pre_extracted_fv_hobs.contains(&(file_name.into(), fv_name.unwrap_or(patina::guids::ZERO)))
```

patina_dxe_core/src/pi_dispatcher.rs
```diff
@@ -760,6 +784,12 @@ impl DispatcherContext {
                         let file = file.clone();
                         let file_name = file.name().into_inner();
 
+                        // if we have an FV2 or extracted FV3 HOB for this FV, it has already been decompressed and we
+                        // don't need to do it again
```
makubacki: Technically, decompression is a potential form of extraction. At this level, I think the comment should stick to the actual operation of extraction.

```suggestion
                        // If we have an FV2 or extracted FV3 HOB for this FV, the producer phase already
                        // extracted (processed) this FV image, so we don't need to extract it again.
```

makubacki (approved):

patina_dxe_core/src/pi_dispatcher.rs
os-d: Updated, also added function comments as I realized I hadn't.

PR #1565, opened 2026-06-16
Patina: Implement VirtIO serial port and supporting change [FF & REBASE]

cfernald:
### Patina: Implement interrupt support for arch module

This commit implements basic interrupt management support in the Patina
SDK. This is needed for some low level implementations such as SerialIO
implementations where the TPL may not be available or accessible.

This commit also creates an abstraction model for the architecture-specific
implementations in the arch module. This uses an internal trait-based
approach to provide a consistent interface across different architectures.
These implementations are then exposed to static wrapper functions.

### Patina: Implement VirtIO console based SerialIO

The commit adds a new SerialIO implementation for a VirtIO console based
on a VirtIO MMIO transport. VirtIO devices are common for virtualized
environments such as QEMU, and a SerialIO implementation allows for greater
flexibility in those environments.

The implementation adds generic support for the basic VirtIO queue and
MMIO operations, but that is left as internal to this implementation for
the time being.
- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- [x] ArmVirt debug transport
- [x] Unit Tests

## Integration Instructions

N/A

### DIFF
--- a/sdk/patina/src/arch.rs
+++ b/sdk/patina/src/arch.rs
@@ -7,8 +7,73 @@
 //! SPDX-License-Identifier: Apache-2.0
 //!
 
-#[cfg(all(not(test), target_arch = "aarch64"))]
-pub mod aarch64;
+cfg_if::cfg_if! {
+    if #[cfg(test)] {
+        #[coverage(off)]
+        mod null;
+        type Arch = null::NullArch;
+    } else if #[cfg(target_arch = "aarch64")] {
+        #[coverage(off)] // Architecture code cannot be unit tested.
+        pub mod aarch64;
+        type Arch = aarch64::AArch64;
+    } else if #[cfg(target_arch = "x86_64")] {
+        #[coverage(off)] // Architecture code cannot be unit tested.
+        pub mod x64;
+        type Arch = x64::X64;
+    }
+}
 
-#[cfg(all(not(test), target_arch = "x86_64"))]
-pub mod x64;
+#[allow(unused)]
+/// A trait for architecture-specific functionality. It inherits from subtraits
+/// for code organization.
+trait ArchSupport: Interrupts {}
+
+//
+// Interrupts support.
+//
+// Only `with_interrupts_disabled` is being exposed publically for now to
+// avoid misuse of direct interrupt manipulation.
+//
+
+trait Interrupts {
+    fn enable_interrupts();
+    fn disable_interrupts();
+    fn interrupts_enabled() -> bool;
+}
+
+/// Enables CPU interrupts on the current architecture.
+fn enable_interrupts() {
+    <Arch as Interrupts>::enable_interrupts();
+}
+
+/// Disables CPU interrupts on the current architecture.
+fn disable_interrupts() {
+    <Arch as Interrupts>::disable_interrupts();
+}
+
+/// Returns whether CPU interrupts are currently enabled on the current architecture.
+fn interrupts_enabled() -> bool {
+    <Arch as Interrupts>::interrupts_enabled()
+}
+
+/// Executes a closure with CPU interrupts disabled, interrupts will be restored to their
+/// previous state after the closure is executed.
+///
+/// ## Important
+///
+/// This function is only intended for low-level operations, and should not be used in
+/// components. Components should use the TPL (Task Priority Level) to manage interrupts.
+pub(crate) fn with_interrupts_disabled<F, R>(f: F) -> R
+where
+    F: FnOnce() -> R,
+{
+    let enabled = interrupts_enabled();
+    if enabled {
+        disable_interrupts();
+    }
+    let result = f();
+    if enabled {
+        enable_interrupts();
+    }
+    result
+}
--- a/sdk/patina/src/arch/aarch64.rs
+++ b/sdk/patina/src/arch/aarch64.rs
@@ -9,6 +9,10 @@
 //! Portions Copyright 2023 The arm-gic Authors.
 //! arm-gic is dual-licensed under Apache 2.0 and MIT terms.
 
+pub(super) struct AArch64;
+
+impl super::ArchSupport for AArch64 {}
+
 /// Reads and returns the value of the given aarch64 system register.
 #[macro_export]
 macro_rules! read_sysreg {
@@ -126,4 +130,19 @@ macro_rules! write_sysreg {
             }
         }
     };
+}
+
+impl super::Interrupts for AArch64 {
+    fn enable_interrupts() {
+        write_sysreg!(reg daifclr, imm 0x02, "isb sy");
+    }
+
+    fn disable_interrupts() {
+        write_sysreg!(reg daifset, imm 0x02, "isb sy");
+    }
+
+    fn interrupts_enabled() -> bool {
+        let daif = read_sysreg!(daif);
+        daif & 0x80 == 0
+    }
 }
--- a/sdk/patina/src/serial/virtio.rs
+++ b/sdk/patina/src/serial/virtio.rs
@@ -1,0 +1,282 @@
+//! [`SerialIO`](crate::serial::SerialIO) implementation for a virtio-console device on the virtio-MMIO transport.
+//!
+//! ## License
+//!
+//! Copyright (C) Microsoft Corporation.
+//!
+//! SPDX-License-Identifier: Apache-2.0
+//!
+
+// The internal virtio infrastructure is being left private for now. It's written to be re-usable,
+// but until there is a known use-case, don't expose a public API for it.
+mod mmio;
+mod queue;
+
+use spin::Mutex;
+
+use mmio::VirtioMmio;
+use queue::VirtQueue;
+
+/// Virtio device id for a console device.
+const VIRTIO_ID_CONSOLE: u32 = 3;
+
+/// Receive virt queue index.
+const RX_QUEUE: u32 = 0;
+/// Transmit virt queue index.
+const TX_QUEUE: u32 = 1;
+
+struct State<const N: usize, const B: usize> {
+    initialized: bool,
+    transport: VirtioMmio,
+    rx: VirtQueue<N, B>,
+    tx: VirtQueue<N, B>,
+    rx_block_offset: usize,
+}
+
+impl<const N: usize, const B: usize> State<N, B> {
+    const unsafe fn new(base_address: usize) -> Self {
+        Self {
+            initialized: false,
+            // SAFETY: forwarded from the caller of this function.
+            transport: unsafe { VirtioMmio::new(base_address) },
+            rx: VirtQueue::new(),
+            tx: VirtQueue::new(),
+            rx_block_offset: 0,
+        }
+    }
+}
+
+/// A [`SerialIO`](crate::serial::SerialIO) implementation for a virtio-console
+/// device attached via the virtio-MMIO transport.
+pub struct VirtioSerial<const N: usize = 8, const B: usize = 128> {
+    state: Mutex<State<N, B>>,
+}
+
+impl<const N: usize, const B: usize> VirtioSerial<N, B> {
+    /// Constructs a new instance for a virtio-MMIO device at the given base
+    /// address. [`SerialIO::init`](crate::serial::SerialIO::init) must be
+    /// called before any I/O.
+    ///
+    /// # Safety
+    ///
+    /// `base_address` must point to an appropriately mapped virtio-MMIO register
+    /// window for a virtio-console. This address must be exclusively used by
+    /// this instance.
+    pub const unsafe fn new(base_address: usize) -> Self {
+        // SAFETY: forwarded from caller.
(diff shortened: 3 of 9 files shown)


sdk/patina/src/arch.rs
```diff
@@ -7,8 +7,70 @@
...
+        type Arch = x64::X64;
+    }
+}
 
-#[cfg(all(not(test), target_arch = "x86_64"))]
-pub mod x64;
+#[allow(unused)]
+/// A trait for architecture-specific functionality. It inherits from subtraits
+/// for code organization.
+trait Architecture: Interrupts {}
```
joschock: super-nit: `Architecture` is a pretty generic name.

sdk/patina/src/arch/aarch64.rs
```diff
@@ -127,3 +131,18 @@ macro_rules! write_sysreg {
         }
     };
 }
+
+impl super::Interrupts for AArch64 {
```
os-d, line 135: We had a conversation recently about not exposing this functionality generically in the SDK, because almost all consumers should go through TPL manipulations instead and we left this in the UART impl itself. So, we either should align to that and not put this here or put it here (should updated patina_internal_cpu to use this) and then put some major comments about not using this in most scenarios.

sdk/patina/src/serial/virtio/mmio.rs
```diff
@@ -0,0 +1,260 @@
...
+
+/// A virtio-MMIO transport at a fixed base address.
+#[derive(Debug)]
+pub(super) struct VirtioMmio {
+    regs: UniqueMmioPointer<'static, VirtioMmioRegs>,
+}
+
+// SAFETY: `VirtioMmio` owns a unique pointer to a device-memory MMIO window
+// All device-touching methods take `&mut self`, so there is no concurrent
+// access through the same instance.
```
os-d, line 122: This isn't really true, right? The safety is disabling interrupts while touching the device? The &mut self doesn't protect us, which is why we have to do the  unsafe impl Send.

sdk/patina/src/serial/virtio/mmio.rs
```diff
@@ -0,0 +1,260 @@
...
+impl VirtioMmio {
+    /// Constructs a new transport for the virtio-MMIO device at the given
+    /// base address.
+    ///
+    /// # Safety
+    ///
+    /// `base_address` must be a non-null address of an appropriately mapped
+    /// virtio-MMIO register window exclusively owned for the lifetime of the
+    /// returned transport.
+    pub(super) const unsafe fn new(base_address: usize) -> Option<Self> {
```
os-d: nit: This function isn't really unsafe, in that it doesn't touch this address. I would lean towards this is a function expectation, pass a good address, but the function won't explode if it is a bad address (others may), so this wouldn't be unsafe. But fine if others disagree.

Since you can't make the struct unsafe to indicate use a real address, I think using new() as unsafe could be fine.

sdk/patina/src/arch/aarch64.rs
joschock: As a data point, I've hardcoded direct aarch64 asm interrupt support into at least one platform hardware SerialIo instance I've written. So - the alternative might not be "no one will use it if we don't provide it" - the alternative is likely "they will roll their own if they believe they need it". I can see arguments both pro/con; Pro - if we provide it, then maybe more eyes will make it less likely to be wrong. Con - if we provide it, then it authorizes a broader usage. If we don't provide it and a roll-your-own breaks, we can be like "not recommended."

os-d: Yep, exactly. I also can go either way on this, but we had the conversation on my PR a couple months ago and decided to keep it out of the SDK, so just wanted to make sure we were all aligned on it going in if that is the case. Again, if it does go in, it should point to your doc on why TPL locks should generally be used.

os-d (approved):

sdk/patina/src/arch/aarch64.rs
cfernald: This was the reason that most of this is not `pub`, I did choose to expose the `with_interrupts_disabled`, with that exact comment, but I could see the argument that should be `pub(crate)`, but the problem we have scales externally. Folks have their own `SerialIO` implementations after all

sdk/patina/src/serial/virtio/mmio.rs
cfernald: This unsafe is required for us to be able to justify our own unsafes. That is, we will use this address internal assuming that it is a correct, valid address. This unsafe is the contract the caller must satisfy for those assumptions to be true.

sdk/patina/src/arch/aarch64.rs
cfernald: My opinion is that `with_interrupts_disabled` (the only `pub` interrupt interface) provides a scoped and contained functionality for this to the point that the comment that this is not for components seems reasonable. Let me know what y'all think.

sdk/patina/src/serial/virtio/mmio.rs
```diff
@@ -0,0 +1,260 @@
...
+
+    /// Sets `STATUS_FAILED`.
+    pub(super) fn set_failed(&mut self) {
+        let mut regs = self.regs.reborrow();
+        let current = field!(regs, status).read();
+        field!(regs, status).write(current | STATUS_FAILED);
+    }
+
+    /// Notifies the device that new entries are available on `queue_index`.
+    /// Issues a SeqCst fenc e before the write so all prior queue updates are
```
makubacki:
```suggestion
    /// Issues a SeqCst fence before the write so all prior queue updates are
```

sdk/patina/src/serial/virtio.rs
```diff
@@ -0,0 +1,281 @@
...
+
+impl<const N: usize, const B: usize> VirtioSerial<N, B> {
+    /// Constructs a new instance for a virtio-MMIO device at the given base
+    /// address. [`SerialIO::init`](crate::serial::SerialIO::init) must be
+    /// called before any I/O.
+    ///
+    /// # Safety
+    ///
+    /// `base_address` must point to an appropriately mapped virtio-MMIO register
+    /// window  for a virtio-console. This address must be exclusively used by
```
makubacki:
```suggestion
    /// window for a virtio-console. This address must be exclusively used by
```

sdk/patina/src/serial/virtio.rs
```diff
@@ -0,0 +1,281 @@
...
+const RX_QUEUE: u32 = 0;
+/// Transmit virt queue index.
+const TX_QUEUE: u32 = 1;
+
+struct State<const N: usize, const B: usize> {
+    initialized: bool,
+    transport: VirtioMmio,
+    rx: VirtQueue<N, B>,
+    tx: VirtQueue<N, B>,
+    rx_block_offset: u32,
```
makubacki: Can `rx_block_offset` be a `usize`? It's already cast to that in several places.

sdk/patina/src/serial/virtio.rs
```diff
@@ -0,0 +1,281 @@
...
+}
+
+impl<const N: usize, const B: usize> State<N, B> {
+    const unsafe fn new(base_address: usize) -> Self {
+        Self {
+            initialized: false,
+            // SAFETY: forwarded from the caller of this function.
+            transport: unsafe { VirtioMmio::new(base_address).expect("Caller provided NULL address.") },
```
makubacki: If this is the only place calling `VirtioMmio::new()` and it panics on `None`, is there a point in returning `Option<Self>` in `VirtioMmio::new()`?

That function is already marked `unsafe` and the safety comment already says.

```
    /// `base_address` must be a non-null address of an appropriately mapped
    /// virtio-MMIO register window exclusively owned for the lifetime of the
    /// returned transport.
```

I think it'd be cleaner to just add a `# Panics` section to the function doc comment there and have the expect directly on the `NonNull::new()` call:

```rust
let ptr = NonNull::new(base_address as *mut VirtioMmioRegs).expect("base_address must be non-null");
```

sdk/patina/src/serial/virtio.rs
```diff
@@ -0,0 +1,281 @@
...
+pub struct VirtioSerial<const N: usize = 8, const B: usize = 128> {
+    state: Mutex<State<N, B>>,
+}
+
+impl<const N: usize, const B: usize> VirtioSerial<N, B> {
+    /// Constructs a new instance for a virtio-MMIO device at the given base
+    /// address. [`SerialIO::init`](crate::serial::SerialIO::init) must be
+    /// called before any I/O.
+    ///
+    /// # Safety
```
makubacki, line 60: Should this also call out that `base_address` should be a stable address (not move after serial I/O is initialized)? That's expected, right?

sdk/patina/src/arch/aarch64.rs
makubacki: I don't have a strong opinion either, but `with_interrupts_disabled()` goes against the spirit of what we've talked about not officially exposing in prior conversations. It is also not technically limited to components but really most places the SDK might be used (which is optimistically, largely components).

makubacki: That said, I'm indifferent to leaving it with guidance in the function comment as well.

cfernald: It does seem we need a better division between component-level interfaces in the SDK vs things that may be needed to build a core, like the SerialIO interfaces.

I think that this is something we address in the SDK refactor.

sdk/patina/src/serial/virtio/mmio.rs
cfernald: The disabling interrupts prevents deadlock, but the Mutex that guards the &mut is what protects this layer. This layer only knows that its being given exclusive access via the mut, it doesn't know how that occurred. So I think the statement stands.

sdk/patina/src/serial/virtio.rs
cfernald: Do you mean that the MMIO must not move? I suppose since it's an emulated device, perhaps that is possible? I could add it, but not sure it provides much to the contract since it's outside of the scope of the callers control, i assume.

sdk/patina/src/arch.rs
cfernald: Switched to `ArchSupport`, but open to better ideas.

sdk/patina/src/arch/aarch64.rs
cfernald: I moved this to `pub(crate)`. I think I still want it to be public, but I will defer that for the larger SDK cleanup effort and not expose it yet.

sdk/patina/src/serial/virtio/mmio.rs
cfernald: I'm going to close this, IMO this is a pretty common pattern when creating a wrapper around a raw pointer. E.g. we do this in Patina-paging, and a bunch of other places. Please re-open if you disagree though.

os-d: I’m fine with it. Imo it falls in the category of not really meaningful (yeah, you have to pass the right parameter when calling a function), but it’s not harmful to have by any means

sdk/patina/src/serial/virtio/mmio.rs
os-d: I missed the mutex, hence the comment. I think its fine.

sdk/patina/src/serial/virtio.rs
makubacki: > Do you mean that the MMIO must not move? 

Yes. I know it is unusual from a platform perspective but regardless, the code makes this assumption, so I think it'd be better to explicitly state that assumption.

makubacki (approved):

sdk/patina/src/serial/virtio.rs
cfernald: Fair enough, will add.

cfernald: Added as must be "present" as well as exclusive for the lifetime.

sdk/patina/src/arch/aarch64.rs
cfernald: @joschock @os-d @makubacki , I thought about this some and decided that I am wrong. I think the real issue here is that our frameworks require interior mutability for the SerialIO implementations. So, I want to make SerialIO use `&mut self` and have static consumers such as SerialLogger, AdvLogger, PatinaDebugger use a new SDK struct `StaticSerialIO<T: SerialIO>` that will provide the appropriate interrupt protected interior mutability.

I filed this bug to track this as part of the breaking SDK changes: https://github.com/OpenDevicePartnership/patina/issues/1567  

For this PR, the solution is to keep this `pub(crate)`, which I think is appropriate and will be used by the solution above.

os-d (approved):

PR #1593, opened 2026-06-26
Prevent ldmxcsr from using absolute address

### LINKED ISSUE #1601: [Bug]: initialize_fpu() should not be marked `preserves_flags`

### Current Behavior

[`initialize_fpu()`](https://github.com/OpenDevicePartnership/patina/blob/035849666a7be0f663e11d1b2f3e830ae3309854/core/patina_internal_cpu/src/cpu/x64/cpu.rs#L108-L135) currently has `preserves_flags`:

https://github.com/OpenDevicePartnership/patina/blob/035849666a7be0f663e11d1b2f3e830ae3309854/core/patina_internal_cpu/src/cpu/x64/cpu.rs#L108-L135

---

For x86, [`preserves_flags` requires](https://doc.rust-lang.org/reference/inline-assembly.html#r-asm.rules.preserved-registers):

> These flags registers must be restored upon exiting the assembly code if the preserves_flags option is set:
> - x86
>   - Status flags in EFLAGS (CF, PF, AF, ZF, SF, OF).
>   - Floating-point status word (all).
>   - Floating-point exception flags in MXCSR (PE, UE, OE, ZE, DE, IE).

1. The [`or` operation](https://www.felixcloutier.com/x86/or#flags-affected) to set OSFXSR results in:
   > The OF and CF flags are cleared; the SF, ZF, and PF flags are set according to the result. The state of the AF flag is undefined.
2. `finit` [clears the floating-point status word](https://www.felixcloutier.com/x86/finit:fninit):
   > The status word is cleared (no exception flags set, TOP is set to 0).
3. I think `ldmxcsr` is also going to clear exception flag bits (PR, UE, etc.)

### Expected Behavior

Do not set `preserves_flags` in an assembly block that doesn't uphold expectations.

### Steps To Reproduce

Review the code in core/patina_internal_cpu/src/cpu/x64/cpu.rs.

### Build Environment

```markdown
N/A
```

### Version Information

```text
Tag: patina-v22.1.0
```

### Urgency

Low

Someone else needs to fix it

No maintainer feedback needed

### Anything else?

_No response_

apop5:
## Description

Closes https://github.com/OpenDevicePartnership/patina/issues/1601

ldmxcsr was attempting to use an absolute address. On a X86 system, this resulted in a 0xF extended address

i.e. ldmxcsr dword ptr [0xffffffffa240bac8] when the value was at 0xa240bac8

Switching the syntax resulted in using a relative address (rip + n).

The absolute address was resulting in a GP# fault when executing.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Virtual x86 system. GP fault before change, boot past after the change.

## Integration Instructions

No integration necessary.

### DIFF
--- a/core/patina_internal_cpu/src/cpu/x64/cpu.rs
+++ b/core/patina_internal_cpu/src/cpu/x64/cpu.rs
@@ -105,6 +105,7 @@ impl EfiCpuX64 {
         // unimplemented!();
     }
 
+    #[coverage(off)]
     fn initialize_fpu(&self) {
         #[cfg(not(test))]
         // SAFETY: This assembly writes only hard coded values to CR4 register, and MMX and FPU control words. No
@@ -115,21 +116,23 @@ impl EfiCpuX64 {
 
             // sdm vol. 1, MMX Control Status Register configuration
             static MMX_CONTROL_WORD: u32 = 0x1F80;
+            let fpu_cw = &raw const FPU_CONTROL_WORD;
+            let mmx_cw = &raw const MMX_CONTROL_WORD;
             asm!(
                 "finit",
-                "fldcw [{FPU_CONTROL_WORD}]",
+                "fldcw [{fpu_cw}]",
 
                 // Set OSFXSR (bit 9) in CR4 to enable SSE instructions
                 "mov {temp}, cr4",
                 "or {temp}, {BIT9}",
                 "mov cr4, {temp}",
 
-                "ldmxcsr [{MMX_CONTROL_WORD}]",
+                "ldmxcsr [{mmx_cw}]",
                 temp = out(reg) _,
-                FPU_CONTROL_WORD = sym FPU_CONTROL_WORD,
-                MMX_CONTROL_WORD = sym MMX_CONTROL_WORD,
+                fpu_cw = in(reg) fpu_cw,
+                mmx_cw = in(reg) mmx_cw,
                 BIT9 = const patina::bit!(9),
-                options(nostack, preserves_flags)
+                options(nostack)
             );
         }
     }


core/patina_internal_cpu/src/cpu/x64/cpu.rs
```diff
@@ -116,7 +116,7 @@ impl EfiCpuX64 {
             // sdm vol. 1, MMX Control Status Register configuration
             static MMX_CONTROL_WORD: u32 = 0x1F80;
             asm!(
-                "finit",
+                "fninit",
```
os-d: Why this change? I know the disassembly shows this but the edk2 asm does have finit.

core/patina_internal_cpu/src/cpu/x64/cpu.rs
```diff
@@ -127,7 +127,7 @@ impl EfiCpuX64 {
                 "ldmxcsr [{MMX_CONTROL_WORD}]",
                 temp = out(reg) _,
                 FPU_CONTROL_WORD = sym FPU_CONTROL_WORD,
-                MMX_CONTROL_WORD = sym MMX_CONTROL_WORD,
+                MMX_CONTROL_WORD = in(reg) MMX_CONTROL_WORD,
```
os-d: Also change FPU_CONTROL_WORD? Can you look at the other x64 inline asm in the codebase and make sure this isn't also done other places? I suspect not.

core/patina_internal_cpu/src/cpu/x64/cpu.rs
apop5: Reverted in latest push

core/patina_internal_cpu/src/cpu/x64/cpu.rs
apop5: Also changed, but the changes became more involved to try to satisfy clippy

core/patina_internal_cpu/src/cpu/x64/cpu.rs
```diff
@@ -115,19 +115,21 @@ impl EfiCpuX64 {
...
 
-                "ldmxcsr [{MMX_CONTROL_WORD}]",
+                "ldmxcsr [{mmx_cw}]",
                 temp = out(reg) _,
-                FPU_CONTROL_WORD = sym FPU_CONTROL_WORD,
-                MMX_CONTROL_WORD = sym MMX_CONTROL_WORD,
+                fpu_cw = in(reg) fpu_cw,
+                mmx_cw = in(reg) mmx_cw,
                 BIT9 = const patina::bit!(9),
                 options(nostack, preserves_flags)
```
makubacki: Unrelated to this change, but this block should not be marked as `preserves_flags` (as is).

For x86, [`preserves_flags` requires](https://doc.rust-lang.org/reference/inline-assembly.html#r-asm.rules.preserved-registers):

> These flags registers must be restored upon exiting the assembly code if the preserves_flags option is set:
> - x86
>   - Status flags in EFLAGS (CF, PF, AF, ZF, SF, OF).
>   - Floating-point status word (all).
>   - Floating-point exception flags in MXCSR (PE, UE, OE, ZE, DE, IE).

1. The [`or` operation](https://www.felixcloutier.com/x86/or#flags-affected) to set OSFXSR results in:
   > The OF and CF flags are cleared; the SF, ZF, and PF flags are set according to the result. The state of the AF flag is undefined.
2. `finit` [clears the floating-point status word](https://www.felixcloutier.com/x86/finit:fninit):
   > The status word is cleared (no exception flags set, TOP is set to 0).
3. I think `ldmxcsr` is also going to clear exception flag bits (PR, UE, etc.)

It's a completely separate change and I'm happy to follow up with a standalone PR to describe it in detail. However, I'm also not opposed to including it here if you'd like.

makubacki (approved):

makubacki:
For code coverage, open to other's feedback, but I think adding `#[coverage(off)]` to `initialize_fpu()` is fine.

core/patina_internal_cpu/src/cpu/x64/cpu.rs
apop5: Updated in latest

Javagedes (approved):

core/patina_internal_cpu/src/cpu/x64/cpu.rs
makubacki: Filed a bug for this so it is visible outside the PR for the original issue - https://github.com/OpenDevicePartnership/patina/issues/1601.

PR #1594, opened 2026-06-27
[FCP] RFC: Consolidate common core code into a single internal library crate

cfernald:
## Description

This RFC proposes creating an internal crate (`core/patina_internal_lib`) for code shared by core implementations. This single crate would remove overhead and simplify changing internal code and interfaces. This will become increasingly important with the introduction of the Standalone MM core.

This RFC builds on the crate organization and naming conventions established by [RFC: Categorizing and Renaming Patina Crates for Consistency](./0007-patina-crates-naming-and-categorization.md).

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [x] Includes documentation?

## How This Was Tested

N/A

## Integration Instructions

N/A


joschock:
Agree with the general motivation of the RFC, but one of the original motivations for several smaller crates was to help clarify/enforce separation of concerns. With everything in a large monocrate, it becomes cheaper to cross-couple things. Are there any organizational or technical mechanisms that we could retain to help with that motivation?

docs/src/rfc/text/0000-shared-core-library.md
```diff
@@ -0,0 +1,210 @@
...
+   in the future, if a purpose arises.
+4. The common crate must allow gating on the `alloc` feature to support all patina
+   core environments.
+
+## Unresolved Questions
+
+- **MM Supervisor coupling.** The exact subset of `patina_internal_lib` consumed
+  by the MM Supervisor is out of scope here and will be driven by the MM
+  Supervisor work.
+- **CPU Crate Independence.** The CPU crate is used in more limited environments
```
os-d, line 135: I don’t like the current structure of the cpu crate, but I do think it probably makes sense to have standalone. It’s a generic enough name that I think it is easy to add new functionality to without being restricted by the name. It would feel odd to me to say, oh, yeah, all the interrupt handling logic? That’s buried in some miscellaneous crate over there.

I don’t have too big a preference, I think ergonomics for the internal crates is more important, but if you are already seeing reasons to keep it separate, then I think it makes sense. 17 files is a decent size, imo

os-d (approved):
I don’t have strong feelings on this one. A small nit, naming is hard, but patina_internal_lib feels a bit too generic, maybe patina_internal_core to indicate its use?

docs/src/rfc/text/0000-shared-core-library.md
cfernald: I'm on the fence on CPU. It is compiled in a unique way and used in things like supervisor that is more restrictive. In the end, it already exists, so leaving it alone is reasonable and we can always merge it in later. The main goal here was to support some refactoring without having to create a bunch of new crates following our current patterns.

cfernald:
> Agree with the general motivation of the RFC, but one of the original motivations for several smaller crates was to help clarify/enforce separation of concerns. With everything in a large monocrate, it becomes cheaper to cross-couple things. Are there any organizational or technical mechanisms that we could retain to help with that motivation?

@joschock, this is true. I think this will be similar to the SDK where we perhaps need to define some rules and enforce them as practice instead of technologically. That is, on us as the maintainers and reviews to prevent bad code design from making it in.

I was thinking about adding a "Here are the rules for this crate", but I think until the crate matures it's probably difficult to say what they should be exactly. Do you have any ideas what types of rules you'd like to see?

makubacki:
> I don’t have strong feelings on this one. A small nit, naming is hard, but patina_internal_lib feels a bit too generic, maybe patina_internal_core to indicate its use?

Yeah, `patina_internal_lib` also struck me as too generic when I first saw it. `patina_internal_core` seems reasonable.

makubacki (approved):
I think it makes sense to step back and reduce our crate footprint for content we do not intend to be consumed outside a small set of core code.

It's less Patina-related crate noise in the registry and scales more sustainably as noted in the RFC. To my understanding, our sharing boundary at the crate level is largely "core consumption only" and the content separation, documentation, and access details can be handled within the module system. Also, as noted we can always split off crates meaningfully when we find justification to do so.

docs/src/rfc/text/0000-shared-core-library.md
cfernald: I will update this RFC to leave CPU as a future decision.

joschock (approved):

cfernald:
Added draft PR: https://github.com/OpenDevicePartnership/patina/pull/1597

kuqin12 (approved):

cfernald:
It seems we have reached agreement. Moving to FCP.

Javagedes (approved):

PR #1596, opened 2026-06-29
Unify Component Parameter Validation [Rebase & FF]

### LINKED ISSUE #1145: [Task]: Make logic for static and runtime component parameter validation more consistent
### Request Description

Pull request https://github.com/OpenDevicePartnership/patina/pull/1124 added compile-time component parameter validation. For future maintainability, it would be ideal to have the method used to validate component parameters at compile-time and runtime as similar as possible.

This task tracks consolidating and unifying that logic.

I will make the change

No maintainer feedback needed

### Anything else?

_No response_

makubacki:
## Description

Addresses a backlog item to consolidate and standardize component parameter rules between compile-time and runtime checks. This was raised for follow up back when compile-time parameter validation was first introduced.

---

Closes #1145 

---

**Consistently apply LF line endings to .stderr files**

Currently LF is enforced for .rs files with rustfmt. .stderr files
are used to hold reference text for compile-time test messages.

Some .stderr files are checked in with CRLF line endings. Others with
LF line endings. This change converts them to LF line endings and
updates .gitattributes to convert .stderr files to LF when committed
going forward.

Note: This is just done here to make actual content edits to these files
clean going forward.

---

**patina_macro: Use common component param rules**

Replace the bespoke `ParamType` conflict matrix in the component macro
with a single `ParamKind` plus `conflict()` rule set in a new
`param_conflict` module.

In turn, this eliminates all of the custom rule logic that was
implemented in `ParamType::conflicts_with()` with rule logic that
matches what is used for runtime validation.

Note:

Unforunately, the `param_conflict` module cannot directly be shared
between compile-time and runtime validation because the `patina`
crate depends on `patina_macro` which is a `proc_macro` crate. The
code could be placed into a new internal crate that is depended on by
both `patina` and `patina_macro`, but that would require publishing a
new internal crate, because a published crate cannot depend on an
unpublished crate, which is not desirable just for this small module.
So, instead, the `param_conflict` module is duplicated in both crates,
with a note in the module doc.

---

**patina: Use common component param rules for runtime validation**

Replace the per-parameter `bitset` conflict checks with a single
`Access::register()` call backed by the shared `conflict()` rules, so
runtime validation and compile-time validation are consistent.

This also makes runtime reject duplicate `StandardBootServices` and
`StandardRuntimeServices` parameters, matching the macro.

The `fixedbitset` dependency is dropped since it is no longer used.

---

**patina_macro: Detect param conflicts in Option and tuple at compile time**

Runtime validation already flattens `Option<P>` and tuples through
their inner `init_state()` calls.

This change makes the macro do the same so compile-time validation
catches the same conflicts instead of letting them slip through to
runtime.

---

**docs: Update component parameter validation rule docs**

Note that compile-time and runtime validation share one conflict rule
set and that the rules also cover `Option` and tuple members.

---

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [x] Includes documentation?

## How This Was Tested

- `cargo make all`

## Integration Instructions

- N/A


makubacki:
This PR makes some opinionated changes to the ways params are handled as part of the consolidation between compile-time and runtime logic. Part of the decisions include "what we do today" but also "what we want to do in the future". And given "what we want in the future", how we implement support for that today.

This change purposefully simplifies built-in parameter validation and eliminates the bitset that was previously used to primarily support parallel execution of components in the future.

To allow these design-level decisions to be discussed in a forum more focused on the design (e.g. RFC), I'll close this PR for now and carry forward some non-design impacting changes separately. @Javagedes and I can follow up on design tradeoffs and how to formalize the discussion.

PR #1603, opened 2026-06-30
[REBASE & FF] Check EL Before Setting HCR_EL2

os-d:
## Description

This PR contains two commits:

 SDK: Move get_current_el() to SDK and Introduce EL Enum
--
get_current_el() is a common AARCH64 function. It is currently being shared in patina_internal_cpu, but is broadly applicable.

This moves it to the SDK and introduces an EL enum for it to return instead of each caller needing to know the register values.

aarch64 intr: Only Set TGE if Running in EL2
--
When initializing exceptions, TGE is always set in HCR_EL2. However, this will hang the system when running at EL1.

Trapping general exceptions is not required at EL1, we are only getting the EL1 exceptions. As such, this commit just changes the code to check the current EL and only set the bit if we are running at EL2.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on a platform that runs at EL1. Before this change, it hangs, after this, it boots past this point.

## Integration Instructions

N/A.

### DIFF
--- a/sdk/patina/src/arch/aarch64.rs
+++ b/sdk/patina/src/arch/aarch64.rs
@@ -146,3 +146,21 @@ impl super::Interrupts for AArch64 {
         daif & 0x80 == 0
     }
 }
+
+/// Supported AARCH64 Exception Levels
+#[derive(Debug, PartialEq, Eq)]
+pub enum AArch64El {
+    EL1,
+    EL2,
+}
+
+// Determine the current exception level
+pub fn get_current_el() -> AArch64El {
+    match read_sysreg!(CurrentEL) & 0xC {
+        0xC => panic!("EL3 is not supported"),
+        0x8 => AArch64El::EL2,
+        0x4 => AArch64El::EL1,
+        0x0 => panic!("EL0 is not supported"),
+        _ => unreachable!(),
+    }
+}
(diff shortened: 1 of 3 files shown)


sdk/patina/src/arch/aarch64.rs
```diff
@@ -146,3 +146,23 @@ impl super::Interrupts for AArch64 {
...
+    EL3,
+}
+
+// Determine the current exception level
+pub fn get_current_el() -> AArch64El {
+    match read_sysreg!(CurrentEL) & 0xC {
+        0xC => AArch64El::EL3,
+        0x8 => AArch64El::EL2,
+        0x4 => AArch64El::EL1,
+        0x0 => AArch64El::EL0,
```
cfernald: I think it's actually better to just Panic here and only allow returning EL1 and EL2. As written all consumers now have to handle this themselves, and if we were to ever add another EL all those panics would have to be individually found. I think it's better to just unify that assumption and then if we add a new option each one of the consumers now becomes a compile error to fix.

cfernald (approved):

sdk/patina/src/arch/aarch64.rs
os-d: I'm good with that

os-d: Updated, also made the enum only have EL1 and EL2.

kuqin12 (approved):

PR #1608, opened 2026-07-01
[REBASE & FF] Always Allocate FV Regions

os-d:
## Description

This PR contains five commits, two of which are set up/clean up for the third commit which is a bug fix. The final two commits are bugs noticed doing this work.

dxe: GCD: Get Descriptors With Closure, Not Filter
--
Currently, get_memory_descriptor_for_address will return any descriptor for an address. get_memory_descriptors accepts a filter to determine which descs to return.

This commit updates both fns to accept a closure so that custom logic can be applied to determine when a desc should
be returned. This removes increasingly complicated and use case specific descriptor filters in favor of a more flexible mechanism.

dxe: GCD: Consolidate get_existent_memory_descriptor_for_address
--
Now that get_memory_descriptor_for_address accepts a closure, convert users of get_existent_memory_descriptor_for_address and drop the function.

dxe: GCD: Always Allocate FV Regions
--
Currently, the DXE Core initialization only attempts to allocate FV regions if they are marked as MMIO in the GCD. However, many platforms mark these regions as system memory (particularly for FVs that were decompressed into memory).

These regions are not allocated and so are free to allocate by regular allocations. This causes corruption in the FV and failure to parse it.

This commit changes the behavior from the EDK II behavior to always allocate FV regions if they are not already allocated (some platforms produce memory allocation HOBs for the FV region; also three FVs are produced for each extracted FV, so we only allocate it once).

This ensures that the FVs do not get corrupted.

dxe: GCD: Account For Direct Allocated GCD Memory in EFI_MEMORY_MAP
--
is_efi_memory_map_descriptor() currently assumes all system memory is efi::ConventionalMemory because it assumes that the allocators have been checked for this descriptor before it runs.

However, this misses direct allocated GCD memory (calling through DXE services).

This commit does three things:
- Move is_efi_memory_map_descriptor() from the SDK to the core so that it can directly call memory_type_for_handle() and be a complete function. Given any memory descriptor it will correctly report what the EFI_MEMORY_MAP type will be, instead of relying on memory_type_handle() being called first. It is not a useful function in the SDK, only intended for core use. It cannot be used by SDK consumers because it relies on a core internal function.
- Update the system memory logic in that function to check whether it is allocated and if the runtime flag is set.
- Add tests.

This was noticed because the FVs are directly allocated in the GCD.

dxe: GCD: Don't Assert on Non-Page Aligned GCD Allocations
--
This updates the EFI_MEMORY_MAP building logic to not assert on non-page aligned GCD allocations. FVs are expected to not be page aligned, but must be allocated to avoid being corrupted by other allocations.

This is a temporary workaround to log and drop the allocations. A longer term solution is tracked in https://github.com/OpenDevicePartnership/patina/issues/1614.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on a platform that had an FV starting at 0x1000. When the page table was allocated bottom up, the root landed at 0x1000. This overwrote the FV signature, causing Patina to fail to parse the FV and therefore could not dispatch any drivers. Obviously, it is incredibly dangerous to have the page table conflict with any other memory as well, though the FV should be read only.

After this change, the FV is marked as allocated and the page table lands on a different page and the system is able to read the FV and dispatch drivers.

Also booted to Windows on a physical Intel platform to confirm no regression.

## Integration Instructions

N/A.

### DIFF
--- a/patina_dxe_core/src/allocator.rs
+++ b/patina_dxe_core/src/allocator.rs
@@ -38,7 +38,7 @@ use crate::{
 };
 pub use fixed_size_block_allocator::SpinLockedFixedSizeBlockAllocator;
 use patina::pi::{
-    dxe_services::{self, GcdMemoryType},
+    dxe_services::GcdMemoryType,
     hob::{self, EFiMemoryTypeInformation, Hob, HobList, MEMORY_TYPE_INFO_HOB_GUID},
 };
 use r_efi::{efi, system::TPL_HIGH_LEVEL};
@@ -1094,7 +1094,8 @@ fn process_hob_allocations(hob_list: &HobList) {
                 }
 
                 let address = desc.memory_base_address;
-                match GCD.get_existent_memory_descriptor_for_address(address) {
+                match GCD.get_memory_descriptor_for_address(address, |d, _| d.memory_type != GcdMemoryType::NonExistent)
+                {
                     // we found the region in the GCD, so we can allocate it
                     Ok(gcd_desc) => {
                         if gcd_desc.base_address == desc.memory_base_address
@@ -1229,7 +1230,7 @@ fn process_hob_allocations(hob_list: &HobList) {
     // EFI_MEMORY_MAP reports as EfiConventionalMemory), which will cause a failure that is unnecessary. We do this
     // after HOB processing because we want to ensure that the GCD is fully populated with the memory map
     // before we allocate page 0, as it may not live in system memory, in which case we cannot allocate it.
-    match GCD.get_existent_memory_descriptor_for_address(0) {
+    match GCD.get_memory_descriptor_for_address(0, |d, _| d.memory_type != GcdMemoryType::NonExistent) {
         Ok(desc) if desc.memory_type == GcdMemoryType::SystemMemory => {
             let address: efi::PhysicalAddress = 0;
 
@@ -1639,7 +1640,10 @@ mod tests {
     };
 
     use super::*;
-    use patina::pi::hob::{GUID_EXTENSION, GuidHob, Hob, header};
+    use patina::pi::{
+        dxe_services,
+        hob::{GUID_EXTENSION, GuidHob, Hob, header},
+    };
     use r_efi::efi;
 
     enum GcdInit {
@@ -2108,7 +2112,9 @@ mod tests {
             ]
             .iter()
             {
-                let allocator = allocators.get_allocator(*memory_type).unwrap();
+                let allocator = allocators
+                    .get_allocator(*memory_type)
+                    .unwrap_or_else(|| panic!("no allocator for type {:#x}", memory_type));
 
                 let granularity = match *memory_type {
                     efi::RESERVED_MEMORY_TYPE
@@ -2132,14 +2138,18 @@ mod tests {
             }
 
             // confirm the MMIO memory allocation occurred in the GCD
-            let mmio_desc = GCD.get_existent_memory_descriptor_for_address(0x10000000).unwrap();
+            let mmio_desc = GCD
+                .get_memory_descriptor_for_address(0x10000000, |d, _| d.memory_type != GcdMemoryType::NonExistent)
+                .unwrap();
             assert_eq!(mmio_desc.memory_type, dxe_services::GcdMemoryType::MemoryMappedIo);
             assert_eq!(mmio_desc.base_address, 0x10000000);
             assert_eq!(mmio_desc.length, 0x2000);
             assert_eq!(mmio_desc.image_handle, protocol_db::DXE_CORE_HANDLE);
 
             // confirm the rest of the MMIO region is not allocated
-            let mmio_desc = GCD.get_existent_memory_descriptor_for_address(0x10002000).unwrap();
+            let mmio_desc = GCD
+                .get_memory_descriptor_for_address(0x10002000, |d, _| d.memory_type != GcdMemoryType::NonExistent)
+                .unwrap();
             assert_eq!(mmio_desc.memory_type, dxe_services::GcdMemoryType::MemoryMappedIo);
             assert_eq!(mmio_desc.base_address, 0x10002000);
             assert_eq!(mmio_desc.length, 0x1000000 - 0x2000);
--- a/patina_dxe_core/src/allocator.rs
+++ b/patina_dxe_core/src/allocator.rs
@@ -1192,25 +1192,20 @@ fn process_hob_allocations(hob_list: &HobList) {
             }) => {
                 log::trace!("[{}] Processing Firmware Volume HOB:\n{:#x?}\n\n", function!(), hob);
 
-                //The EDK2 C reference core maps FVs to MMIO space, but many implementations don't declare the
-                //corresponding resource descriptor. Check the current region in the GCD to see whether a resource
-                //descriptor of the appropriate type has been reported. If not, print a warning and skip attempting
-                //to reserve it in the GCD.
-                if let Ok(existing_desc) = GCD.get_existent_memory_descriptor_for_address(*base_address)
-                    && (existing_desc.memory_type != dxe_services::GcdMemoryType::MemoryMappedIo
-                        || existing_desc.image_handle != INVALID_HANDLE)
-                {
-                    log::info!(
-                        "Skipping FV HOB at {base_address:#x?} of length {length:#x?}. Containing region is not MMIO."
-                    );
-                    continue;
-                }
+                // EDK II assumes all FVs are MMIO, however many platforms do not report the
+                // backing address space as an MMIO resource, causing the allocation to silently fail.
+                // Instead of assuming MMIO here, retrieve the backing memory type from the GCD if it
+                // exists and allocate the FV region using that type. Note: there are often multiple
+                // FV HOBs for the same FV. This will only allocate for the first HOB since
+                // get_memory_descriptor_for_address() is filtering for free descriptors.
 
                 //The 4K granularity rule does not apply to FV hobs, so allocate_pages cannot be used.
                 //This means they must be direct-allocated in the GCD, and no stats will be tracked for them.
-                let _ = GCD.allocate_memory_space(
+                match GCD.get_memory_descriptor_for_address(*base_address, |_, allocated| !allocated) {
+                    Ok(desc) => {
+                        let _ = GCD.allocate_memory_space(
                     AllocationStrategy::Address(*base_address as usize),
-                    dxe_services::GcdMemoryType::MemoryMappedIo,
+                    desc.memory_type,
                     0,
                     *length as usize,
                     protocol_db::DXE_CORE_HANDLE,
@@ -1220,6 +1215,9 @@ fn process_hob_allocations(hob_list: &HobList) {
                             "Failed to allocate memory space for firmware volume HOB at {base_address:#x?} of length {length:#x?}. Error: {err}",
                         );
                     });
+                    }
+                    Err(_) => continue,
+                }
             }
             _ => continue,
         };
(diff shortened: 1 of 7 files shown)


patina_dxe_core/src/gcd/spin_locked_gcd.rs
```diff
@@ -1241,6 +1278,70 @@ impl GCD {
...
+    /// types are included in the EFI memory map.
+    ///
+    /// # Returns
+    /// * `Some(memory_type)` if the descriptor should be included in the EFI memory map
+    /// * `None` if the descriptor should be excluded
+    fn is_efi_memory_map_descriptor(descriptor: &MemorySpaceDescriptor) -> Option<efi::MemoryType> {
+        // Validate page alignment and size
+        let number_of_pages = ((descriptor.length as usize + UEFI_PAGE_MASK) / UEFI_PAGE_SIZE) as u64;
+        if number_of_pages == 0 {
+            debug_assert!(false, "GCD returned a memory descriptor smaller than a page.");
```
joschock: I think the present GCD supports arbitrary sized blocks. If this is a requirement (GCD only have page-granularity descriptors), then maybe we should enforce this requirement on block add/create so we can find the producer of the problem. If this debug_assert! trips, I don't have an easy way of figuring out where the small descriptor came from. Maybe this check already exists on the input side? if so, then no-op. Otherwise maybe we should file an issue to catch this on the input side.

patina_dxe_core/src/allocator.rs
```diff
@@ -1191,25 +1191,18 @@ fn process_hob_allocations(hob_list: &HobList) {
...
-                {
-                    log::info!(
-                        "Skipping FV HOB at {base_address:#x?} of length {length:#x?}. Containing region is not MMIO."
-                    );
-                    continue;
-                }
+                // EDK II assumes all FVs are MMIO, however many platforms do not report it as MMIO and the allocation
+                // silently fails. Check the GCD and get the type if it exists so that we do not allocate it if it is
+                // marked as system memory. We usually will have three FV HOBs for each FV, an FV, FV2, and FV3 HOB,
+                // so only allocate it once.
```
joschock: bit confusing.  would submit this as a suggestion, but line wrappings are hard.

> EDK II assumes all FVs are MMIO, however many platforms do not report the backing address space as an MMIO resource, causing the allocation to silently fail. Instead of assuming MMIO here, retrieve the backing memory type from the GCD if it exists and allocate the FV region using that type. Note: there are often multiple FV HOBs for the same FV. This will only allocate for the first HOB since get_memory_descriptor_for_address() is filtering for free descriptors.

patina_dxe_core/src/allocator.rs
```diff
@@ -1191,25 +1191,18 @@ fn process_hob_allocations(hob_list: &HobList) {
...
 
                 //The 4K granularity rule does not apply to FV hobs, so allocate_pages cannot be used.
```
joschock, line 1202: Reference to previous comment; if FV hobs are not 4K granularity, then this might trip that debug assert.

joschock (approved):

patina_dxe_core/src/gcd/spin_locked_gcd.rs
os-d: Yeah, its a good point (and the FV one below). The problem is that the GCD is not required to be page aligned and the EFI_MEMORY_MAP is:

image

edk2 has an assert here as well that the merged GCD descriptors must be page aligned:

https://github.com/tianocore/edk2/blob/0809aa4dceb1fafc1aa11de07cb6c8735d77bf9e/MdeModulePkg/Core/Dxe/Mem/Page.c#L2168.

However, all system memory is going to be page aligned because it is owned by gMemoryMap there which is going to gate things as page aligned. Technically a system memory allocation could occur behind the scenes and break this.

This feels like something that should be resolved in the PI spec, only allowing GCD descriptors to be page aligned. I'll consider whether that would be worthwhile to take up.

We also can't force the FV allocation to be page aligned because it may well sit next to something else. We also can't say that the merged GCD entry will even be page aligned for the FV because if it is sitting next to free memory, then it will have a different image handle and not be merged.

There are implications for memory protections as well, of course.

We could do a few things:
- When building the EFI_MEMORY_MAP, if we hit a GCD descriptor that isn't page aligned and can't be merged (optionally promoting free memory adjacent to the same type), just don't report it. There are memory protections implications here, we may also miss giving the OS memory it should have.
- Allocate FV region. If not page aligned, also allocate rest of page. If that fails, fine, we hope it gets merged. If future allocations fail, we hope it is packed FVs and okay, fine we've already allocated.
- Require memory allocation hobs are produced for FV regions. These must be page aligned, i.e. FVs themselves don't have to be, but the memory they take up must be. However, this would require copying FVs out of flash. In DXE, this is likely already done because we typically (but not always) have compressed DXE FVs that get decompressed in pre-DXE. This could be a PI spec update.
- Allow non-page aligned GCD descriptors with some requirements:
  - Adjacent descriptors must be able to merge to page granularity and must be the same GCD type. They must also have no image handle (free) or the same image handle.
  - Free memory will get promoted to being mapped RW since the whole page must be mapped.
  - When we build the EFI_MEMORY_MAP we will merge adjacent descriptors that are not page aligned as long as the GCD type and image handle is the same.

I don't love any of this :). It's complicated, but maybe the last option is the most workable. edk2 does some stuff like this.

patina_dxe_core/src/allocator.rs
os-d: Agreed. Closing this to discuss in above thread.

patina_dxe_core/src/gcd/spin_locked_gcd.rs
joschock: I'm tentatively supportive of an opinionated GCD that forces page-alignment/granularity at least for RAM-backed types. I've always felt it a bit weird that GCD (which I see as a relatively slow coarse-grained allocator) supports arbitrary size/align. A 3-byte GCD allocation at 0x01234567 is anarchy.  I am a bit worried on the MMIO side, though, because I'm not sure that hardware always shares my view on this, and things like XIP FVs that exist for hardware reasons might not be strictly page-aligned are real. There is less reliance on "Allocate/Free" semantics on the MMIO side though; i.e. it is unlikely that something is going to request an "arbitrary" MMIO space allocation and then scribble an FV over it. 

This feels like a bigger issue than this PR though, I would be comfortable if we switched the debug_assert to a log::warn() and filed an issue to address this more wholistically. Maybe I'm even okay with a debug_assert() if we don't have anyone hitting it.

os-d: I agree. I am not concerned about MMIO allocations and the regions really may not be page aligned.

I was going to remove the debug_assert(), but the function does rely on the descs being aligned. I'm tempted to leave it in and see if we hit it after desc merging is taken into account. For an Intel platform, arm64 platform, and the two patina-qemu virtual platforms, it didn't cause an issue, but obviously it is dependent on memory layout.

Probably the safer option is to warn and drop the desc from the EFI_MEMORY_MAP. But then we get into the, well, should we try to merge better, etc.

This is already an issue, so good to defer it, but as you pointed out the FV allocation we believe will not be page aligned and then we are relying on the merging to be okay, which who knows.

Instead, as a temporary solution, we could not allocate the FV region at all and instead copy FVs to a new location that we allocate as page aligned. This does require updating the HOBs in DXE which is obviously bad :). So let's not do that.

I think I would opt for filing an issue and just warning and dropping the non-page aligned descriptor for now. The OS will get a little bit less memory in that case, but I think it is a quick safe option that won't break platforms while I'm figuring out the complete story.

makubacki: I agree with a warning and tracking issue for now. It would also be nice to clean up approaches and discuss them in the Patina Dev & Design Sync meeitng.

patina_dxe_core/src/gcd/spin_locked_gcd.rs
```diff
@@ -1048,6 +1063,7 @@ impl GCD {
     pub fn get_memory_descriptor_for_address(
```
makubacki, line 1028: Is it intentional for `get_memory_descriptors()` to match allocated and unallocated for `MmioAndReserved` but `get_memory_descriptor_for_address()` to just match `MemoryBlock::Unallocated`?

patina_dxe_core/src/gcd/spin_locked_gcd.rs
```diff
@@ -1031,12 +1035,23 @@ impl GCD {
...
-                    if descriptor.memory_type == dxe_services::GcdMemoryType::MemoryMappedIo
-                        || descriptor.memory_type == dxe_services::GcdMemoryType::Reserved =>
+                (
+                    DescriptorFilter::MmioAndReserved,
+                    MemoryBlock::Unallocated(descriptor) | MemoryBlock::Allocated(descriptor),
+                ) if descriptor.memory_type == dxe_services::GcdMemoryType::MemoryMappedIo
+                    || descriptor.memory_type == dxe_services::GcdMemoryType::Reserved =>
                 {
                     buffer.push(*descriptor);
                 }
```
makubacki: `init_paging_with()` gets `MmioAndReserved` descriptors:

```rust
        let mut mmio_res_descs: Vec<dxe_services::MemorySpaceDescriptor> =
            Vec::with_capacity(self.memory_descriptor_count() + 10);
        self.memory
            .lock()
            .get_memory_descriptors(mmio_res_descs.as_mut(), DescriptorFilter::MmioAndReserved)
            .expect("Failed to get MMIO descriptors!");
```

And `Allocated` descriptors:

```rust
        let mut descriptors: Vec<dxe_services::MemorySpaceDescriptor> =
            Vec::with_capacity(self.memory_descriptor_count() + 10);
        self.memory
            .lock()
            .get_memory_descriptors(&mut descriptors, DescriptorFilter::Allocated)
            .expect("Failed to get allocated memory descriptors!");
```

My understanding is that with `get_memory_descriptors()` now returning allocated descriptors for `MmioAndReserved` there will potentially be duplicates in the two vecs.

Can you confirm this is okay? It looks like allocated MMIO/reserved might just have `GCD.memory_protection_policy.apply_allocated_memory_protection_policy(desc.attributes)` applied twice?

makubacki (approved):

patina_dxe_core/src/gcd/spin_locked_gcd.rs
os-d: Ah, thanks, that was my mistake. I had forgotten that the allocated one would pull more than just system memory. So I was viewing it as missing the allocate MMIO and reserved. I'll change this for the filters to be a bit more clear, then not have the duplication.

patina_dxe_core/src/gcd/spin_locked_gcd.rs
os-d: No, I'll unify

patina_dxe_core/src/gcd/spin_locked_gcd.rs
os-d: Filed https://github.com/OpenDevicePartnership/patina/issues/1614 and switched to a warning.

patina_dxe_core/src/gcd/spin_locked_gcd.rs
os-d: Rewrote this.

os-d:
@joschock @makubacki I significantly changed the desc filtering code to take a closure instead of trying to continue to update the descriptor filter for different situations. I restested on QEMU and a physical Intel platform.

patina_dxe_core/src/gcd/spin_locked_gcd.rs
```diff
@@ -1006,48 +994,41 @@ impl GCD {
     pub fn get_memory_descriptors(
         &self,
         buffer: &mut Vec<dxe_services::MemorySpaceDescriptor>,
-        filter: DescriptorFilter,
+        mut filter: impl FnMut(&dxe_services::MemorySpaceDescriptor, bool) -> bool,
```
Javagedes: I like this implementation. However, this is making it a generic function (due to the `impl`) which means it's implementation gets copied as a new function for each usage. Making this generic over a `FnMut` is only necessary if you are wanting to capture state outside of your filter function for usage inside the filter. I don't see that happening in any usages, so it's probably better to just accept a function pointer, which means the function will no longer be generic over the filter. Something like this:

```rust
filter: fn(&MemorySpaceDescriptor, bool) -> bool
```

Javagedes (approved):

patina_dxe_core/src/gcd/spin_locked_gcd.rs
os-d: Thanks, I'm not sure why I did it that way. Updated.

PR #1611, opened 2026-07-02
SDK: Migrate to lzma-rust2 instead of patina-lzma-rs

os-d:
## Description

patina-lzma-rs is a fork of the unmaintained lzma-rs crate that adds no_std support for it.

lzma-rust2 is a maintained fork of lzma-rs that already contains no_std support and large performance improvements over lzma-rs.

On an example AARCH64 platform, decompressing an LZMA compressed FV went from 2500 ms with patina-lzma-rs to 1314 ms with lzma-rust2.

Host based benchmarking shows similar performance improvements.

This commit moves patina over to the maintained crate with better performance that doesn't require a custom override.

The optimization feature is not enabled because it is primarily for encoding (which patina does not do) and does not support the FORMAT_ONLY compression type that our FV use, so the performance is the same but it adds unsafe code.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested with local benchmarks and running on an AARCH64 platform.

## Integration Instructions

N/A. The changed dependency will be picked up automatically. patina-lzma-rs will be deprecated.

### DIFF
--- a/sdk/patina_ffs/src/volume.rs
+++ b/sdk/patina_ffs/src/volume.rs
@@ -774,7 +774,7 @@ impl TryFrom<(&VolumeRef<'_>, &dyn SectionExtractor)> for Volume {
 mod test {
     use core::{mem, sync::atomic::AtomicBool};
     use log::{self, Level, LevelFilter, Metadata, Record};
-    use lzma_rs::lzma_decompress;
+    use lzma_rust2::{LzmaOptions, LzmaReader, LzmaWriter, Read, Write};
     use patina::pi::fw_fs::{self, ffs, fv};
     use r_efi::efi;
     use serde::Deserialize;
@@ -783,7 +783,6 @@ mod test {
         env,
         error::Error,
         fs::{self, File},
-        io::Cursor,
         path::Path,
     };
     use uuid::Uuid;
@@ -1289,9 +1288,17 @@ mod test {
                     && guid_header.section_definition_guid == fw_fs::guid::LZMA_SECTION
                 {
                     let data = section.try_content_as_slice()?;
-                    let mut decompressed: Vec<u8> = Vec::new();
-                    lzma_decompress(&mut Cursor::new(data), &mut decompressed)
+                    let mut reader = LzmaReader::new_mem_limit(data, patina::base::SIZE_512MB as u32, None)
                         .map_err(|_| FirmwareFileSystemError::DataCorrupt)?;
+                    let mut decompressed: Vec<u8> = Vec::new();
+                    let mut chunk = [0u8; 4096];
+                    loop {
+                        let n = reader.read(&mut chunk).map_err(|_| FirmwareFileSystemError::DataCorrupt)?;
+                        if n == 0 {
+                            break;
+                        }
+                        decompressed.extend_from_slice(chunk.get(..n).ok_or(FirmwareFileSystemError::DataCorrupt)?);
+                    }
                     return Ok(decompressed);
                 }
                 Err(FirmwareFileSystemError::Unsupported)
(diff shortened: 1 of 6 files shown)


apop5 (approved):

kouchekiniad (approved):

makubacki:
If this can wait until Monday, I'd be happy to look at it then. As mentioned offline, it's great to move off our custom fork of lzma-rs and pick up the perf improvement.

makubacki (approved):

sdk/patina_ffs/src/volume.rs
```diff
@@ -1348,9 +1355,17 @@ mod test {
                     && guid_header.section_definition_guid == fw_fs::guid::LZMA_SECTION
                 {
                     let data = section.try_content_as_slice()?;
-                    let mut decompressed: Vec<u8> = Vec::new();
-                    lzma_decompress(&mut Cursor::new(data), &mut decompressed)
+                    let mut reader = LzmaReader::new_mem_limit(data, u32::MAX, None)
                         .map_err(|_| FirmwareFileSystemError::DataCorrupt)?;
+                    let mut decompressed: Vec<u8> = Vec::new();
+                    let mut chunk = [0u8; 4096];
+                    loop {
```
Javagedes, line 1362: Unless there is a specific reason you are reading chunks, I think you can do something like this:

```rust
let mut decompressed: Vec<u8> = Vec::new();
reader
    .read_to_end(&mut decompressed)
    .map_err(|_| FirmwareFileSystemError::DataCorrupt)?;
```

Javagedes: For here and in patina_ffs_extractors

Javagedes (approved):

sdk/patina_ffs/src/volume.rs
os-d: I believe `read_to_end()` is only usable in `std`. `LzmaReader` itself just impls `read()`: https://github.com/hasenbanck/lzma-rust2/blob/1e05e7bcb3e2a5991e9c204caeb6475e384577ad/src/lzma2_reader.rs#L176. When `std` is used, it also has the `std::io::Read` trait. We could update the tests, but then we aren't testing what we are actually doing.

sdk/patina_ffs/src/volume.rs
```diff
@@ -1289,9 +1288,17 @@ mod test {
                     && guid_header.section_definition_guid == fw_fs::guid::LZMA_SECTION
                 {
                     let data = section.try_content_as_slice()?;
-                    let mut decompressed: Vec<u8> = Vec::new();
-                    lzma_decompress(&mut Cursor::new(data), &mut decompressed)
+                    let mut reader = LzmaReader::new_mem_limit(data, u32::MAX, None)
```
makubacki: @os-d, can you please consider setting an upper memory limit instead of `u32::Max`?

We should also be careful when parsing the section itself, but I think a reasonable upper bound makes sense here too.

sdk/patina_ffs/src/volume.rs
Javagedes: ah I missed that the `Read` Trait is `std` and that it is a custom `Read` trait for `no_std`. Closing with no action needed.

sdk/patina_ffs/src/volume.rs
os-d: Yeah, technically the uncompressed FV can be up to u32::Max, but in practice it should be much, much lower. We can try to avoid a corrupted or malicious FV (though that is generally considered inside the trust boundary) and use 512MB. The largest sizes I've seen for platforms have all been 10MB or less, so that seems a lot of padding while protecting us from running out of resources. If we encounter a platform with a larger FV, we can have the discussion then, though I would assume this is a bug.

PR #1612, opened 2026-07-02
Switch to the stable toolchain (and RUSTC_BOOTSTRAP=1)

makubacki:
## Description

Per RFC 0030, switch to the stable toolchain and set RUSTC_BOOTSTRAP=1 to allow the stable toolchain to accept the unstable feature gates Patina still depends on.

Sets `allow-features` to the list of unstable features Patina uses.

Update associated documentation to reflect the change.

---

Plan is to get a single change into this repo (this PR) that includes the exact changes from patina-devops + doc updates then the patina-devops changes will not make any subsequent updates to the files.

- patina-devops PR: https://github.com/OpenDevicePartnership/patina-devops/pull/169

---

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [x] Includes documentation?

## How This Was Tested

- `cargo make all`

## Integration Instructions

- See [RFC 0030 - Build on Stable Rust with RUSTC_BOOTSTRAP](https://github.com/OpenDevicePartnership/patina/blob/main/docs/src/rfc/text/0030-rustc-bootstrap.md) and the documentation updates in this change.

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -353,6 +353,7 @@ doctests link against rlibs built by the same compiler, and points `mdbook test`
 to the mdbook profile `target/mdbook/deps/` directory produced by `build-mdbook-deps`.
 """
 clear = true
+env = { RUSTC_BOOTSTRAP = "1" }
 script_runner = "@duckscript"
 script = '''
 toolchain_file = readfile rust-toolchain.toml
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -6,6 +6,15 @@
 
 [env]
 PATINA_CONFIG_VERSION = "1"
+RUSTC_BOOTSTRAP = "1"
+
+# Restrict the unstable features RUSTC_BOOTSTRAP makes available to only those Patina has accepted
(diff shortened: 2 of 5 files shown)


vineelko (approved):

Makefile.toml
```diff
@@ -353,6 +353,7 @@ doctests link against rlibs built by the same compiler, and points `mdbook test`
 to the mdbook profile `target/mdbook/deps/` directory produced by `build-mdbook-deps`.
 """
 clear = true
+env = { RUSTC_BOOTSTRAP = "1" }
```
os-d, line 356: I would think this isn’t needed with the config.toml change?

os-d (approved):

Makefile.toml
makubacki: config.toml only applies to `cargo` commands.

os-d: Oh, I missed this was scoped to the mdbook commands, I was reviewing on my phone :).

PR #1620, opened 2026-07-08
patina_internal_cpu: Prevent sign extension of symbol addresses in x64 interrupt handlers

apop5:
## Description

In GAS assembly syntax, lea <global-variable> will create a 32bit absolute address and place it into the register. If the register is a 64 bit register, the value will be signed extended. 

This is a problem if the patina image is loaded above address 0x8000_0000. The 32bit address will be sign extended.

This was found on a physical platform where patina was loaded into system memory at 0xA000_0000. 

Switch to using RIP relative form syntax. 

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested
On a platform where PEI allocated dxe_core at a higher address, exceptions would occur, but the exception handlers were never reached. After changing assembly syntax, exception handlers were reached. 

## Integration Instructions
No integration necessary.

### DIFF
--- a/core/patina_internal_cpu/src/interrupts/x64/interrupt_handler.asm
+++ b/core/patina_internal_cpu/src/interrupts/x64/interrupt_handler.asm
@@ -22,7 +22,7 @@ AsmIdtVectorBegin:
         sub     rsp, 8
         mov     qword ptr [rsp], vector
         push    rax
-        mov     rax, offset common_interrupt_entry
+        lea     rax, [rip + common_interrupt_entry]
         jmp     rax
         .set vector, vector+1
     .endr
@@ -293,8 +293,8 @@ stack_normalized:
 # This routine only uses volatile registers for the EFI calling convention.
 #
 AsmGetVectorAddress:
-    lea     r10, AsmIdtVectorBegin
-    lea     rax, AsmIdtVectorEnd
+    lea     r10, [rip + AsmIdtVectorBegin]
+    lea     rax, [rip + AsmIdtVectorEnd]
     sub     rax, r10
     shr     rax, 8 # >> 8 == / 256
 


core/patina_internal_cpu/src/interrupts/x64/interrupt_handler.asm
```diff
@@ -22,7 +22,7 @@ AsmIdtVectorBegin:
         sub     rsp, 8
         mov     qword ptr [rsp], vector
         push    rax
-        mov     rax, offset common_interrupt_entry
+        lea     rax, [rip + common_interrupt_entry]
```
os-d, line 25: For my knowledge, was offset not already doing a rip relative offset?

os-d (approved):

core/patina_internal_cpu/src/interrupts/x64/interrupt_handler.asm
apop5: when Stepping through the code, it was doing the same sign extending behavior. 


Searching for documentation, this seems to be specific to x86-64:

Rust’s inline asm produces a memory source operand, which encodes a displacement field. On x86‑64, this displacement is a signed 32‑bit immediate, which is sign‑extended to 64 bits at runtime.

cfernald: I think the fix is good, but I'm confused why the assembly would choose that. Best to avoid is all together though with LEA.

cfernald (approved):

makubacki (approved):

core/patina_internal_cpu/src/interrupts/x64/interrupt_handler.asm
makubacki: My understanding is that it is falling into the [`REX.W + C7 /0 id` form](https://www.felixcloutier.com/x86/mov):

image

Where `common_interrupt_entry` is a 32-bit immediate. It looks like [`movabs`](https://sourceware.org/binutils/docs/as/i386_002dVariations.html) would treat the immediate (`common_interrupt_entry`) as a 64-bit operand:

> In 64-bit code, ‘movabs’ can be used to encode the ‘mov’ instruction with the 64-bit displacement or immediate operand.

In any case, I think the current PR solution is better.

PR #1628, opened 2026-07-08
[major] Refactor performance into core service

### LINKED ISSUE #1483: [Bug]: Performance tracking has a blackout as DXE entry

### Current Behavior

Currently the performance protocol backing the performance calls initializes in the patina_performance component, and so is not available until component dispatching. Additionally, depending on the components installed, the operations that may occur before this would not be consistent. This leads to a significant gap in potential performance metrics as well as misleading time information in the PEI to DXE transition.

### Expected Behavior

The core performance infrastructure should be implemented in the core and available early. This could then be exposed to patina performance as a service that it publishes as a protocol or all of the functionality could be merged into the core. 

### Steps To Reproduce

N/.A

### Build Environment

```markdown
N/A
```

### Version Information

```text
N/A
```

### Urgency

Medium

Someone else needs to fix it

No maintainer feedback needed

### Anything else?

_No response_

cfernald:
## Description

This PR addresses multiple issues with the current performance structure

1. Performance data is not available until the component is dispatched. This can cause significant misrepresentation of the DXE execution time and cause events from images to be lost if not dispatched in the first loop.
2. Performance implementations in the SDK require SDK statics, which is generally an SDK antipattern. 

This PR does the following

1. Moves the core performance implementation into the core, into a new struct names `CorePerformance`, following the pattern established by the timer and memory management.
2. Creates a new service trait `PerfortmanceMeasurement` that exposes this core support.
3. Updates the component to now use this core service to produce the protocol & publish tables.
4. Change initialization and configuration of performance measurements to be done by the core.
5. Updates all documentation according to these changes
6. Some smaller rework and restructuring 

Resolves: #1483 

## Reviewing this PR

This PR is intended to be merged as a single commit. The individual commits do attempt to follow a basic progression of the code rework but may be incomplete at the partial steps. Recommended review flow

1. Start with the new theory of operation md (performance_measurement.md) to understand the intended structure. 
2. Review the new service interface (`PerformanceMeasurement`)
3. Review the new core implementation
4. Review the changes to the component to consume the new service

- [x] Impacts functionality?
- [ ] Impacts security?
- [x] Breaking change?
- [x] Includes tests?
- [x] Includes documentation?

## How This Was Tested

Tested for consistent performance data using `dp` before and after refactor.

## Integration Instructions

Platforms must change use of `with_measurements` from the `patina_performance` to instead override `PlatformInfo::default_performance_config()`


patina_dxe_core/src/lib.rs
```diff
@@ -237,6 +239,15 @@ pub trait PlatformInfo: 'static {
 
     /// The platform's section extractor type, used when extracting sections from firmware volumes.
     type Extractor: SectionExtractor;
+
+    /// The performance measurement configuration used when no performance configuration HOB is present.
+    ///
+    /// Defaults to disabled. Platforms may override this option to control the default behavior of the performance
+    /// measurement service when no configuration HOB is present.
+    #[inline(always)]
+    fn default_performance_config() -> PerformanceConfig {
```
cfernald: @Javagedes , curious about your thoughts on this. I'd rather not create a new sub-type here, but also i'm hesitant for `PlatformInfo` to become a dumping ground.

os-d (approved):

components/patina_performance/README.md
```diff
@@ -1,53 +1,65 @@
 # Patina Performance Component
 
-The Patina performance component maintains the infrastructure to report firmware performance information.
+The Patina performance component acts as the translation layer between the core's performance implementation, and the
+UEFI annd ACPI implementations.
```
magravel: Is this a spelling error? "annd" -> "and"

magravel (approved):

makubacki:
I haven't taken a look through this one yet, but similar to #1630, please try to keep each file about 80% (ideally 85%) coverage.

cfernald:
@makubacki @Javagedes , i will add tests, but please review the changes for the new structure. The PR comment has a review guide since the commits are a bit of a mess due to the complexity of the refactor.

patina_dxe_core/src/performance.rs
```diff
@@ -0,0 +1,773 @@
...
+
+        // Record the PEI-end / DXE-begin cross-module markers. This runs during core memory initialization, as early
+        // as the performance engine can record into its table, so the DXE span is captured close to the phase boundary.
+        let dxe_core_guid = patina::guids::DXE_CORE.into_inner();
+        CorePerformance.perf_cross_module_end("PEI", &dxe_core_guid);
+        CorePerformance.perf_cross_module_begin("DXE", &dxe_core_guid);
+    }
+
+    pub(crate) fn enabled() -> bool {
+        PERF_MEASUREMENT_MASK.load(Ordering::Relaxed) != 0
```
makubacki: I may be misunderstanding, but should this use `enabled` instead of the mask? 

```rust
pub struct PerformanceConfig {
    pub enabled: u8,               // ENABLED / DISABLED
    pub enabled_measurements: u32, // bitmask of `Measurement`
}
```

edk2 used BIT0 for "performance enabled":

```
  ## The mask is used to control PerformanceLib behavior.<BR><BR>
  #  BIT0 - Enable Performance Measurement.<BR>
  #  BIT1 - Disable Start Image Logging.<BR>
  #  BIT2 - Disable Load Image logging.<BR>
  #  BIT3 - Disable Binding Support logging.<BR>
  #  BIT4 - Disable Binding Start logging.<BR>
  #  BIT5 - Disable Binding Stop logging.<BR>
  #  BIT6 - Disable all other general Perfs.<BR>
  #  BIT1-BIT6 are evaluated when BIT0 is set.<BR>
  # @Prompt Performance Measurement Property.
  # @Expression  0x80000002 | (gEfiMdePkgTokenSpaceGuid.PcdPerformanceLibraryPropertyMask & 0x80) == 0
  gEfiMdePkgTokenSpaceGuid.PcdPerformanceLibraryPropertyMask|0|UINT8|0x00000009
```

But my understanding is that we've broken that out to a separate field.

patina_dxe_core/src/performance.rs
```diff
@@ -0,0 +1,773 @@
...
+            }
+
+            // SAFETY: To be honest there is no way to guarantee the memory read here is valid and owned by us,
+            // but we have at least validated that the type gives a known layout and the layout of the device path
+            // matches its claimed length.
+            unsafe {
+                let guid_ptr = (loaded_image.file_path as *const u8)
+                    .add(mem::size_of::<efi::protocols::device_path::Protocol>())
+                    as *const BinaryGuid;
+                guid = ptr::read(guid_ptr);
```
makubacki: I don't think device path nodes carry an alignment guarantee.

```suggestion
                guid = ptr::read_unaligned(guid_ptr);
```

patina_dxe_core/src/lib.rs
```diff
@@ -469,13 +480,28 @@ impl<P: PlatformInfo> Core<P> {
...
         log::info!("GCD - After memory init:\n{GCD}");
 
+        // Initialize the core performance service from the HOB configuration (or the platform default) before
+        // registering it below, so the service is published only when performance measurement is enabled. The engine
+        // relies only on the arch timer and its own `TplMutex`, both of which are available at this point.
+        let perf_config =
+            performance::read_performance_config(self.hob_list()).unwrap_or(P::default_performance_config());
+        let perf_hob_records = performance::read_hob_performance_records(self.hob_list());
+        performance::CorePerformance::init(
+            P::CpuInfo::perf_timer_frequency().unwrap_or(0),
```
makubacki: It is not a big deal, but I think that reading `P::CpuInfo::perf_timer_frequency().unwrap_or(0)` into a local would be a little cleaner.

patina_dxe_core/src/lib.rs
```diff
@@ -469,13 +480,28 @@ impl<P: PlatformInfo> Core<P> {
 
         log::info!("GCD - After memory init:\n{GCD}");
 
+        // Initialize the core performance service from the HOB configuration (or the platform default) before
+        // registering it below, so the service is published only when performance measurement is enabled. The engine
+        // relies only on the arch timer and its own `TplMutex`, both of which are available at this point.
+        let perf_config =
+            performance::read_performance_config(self.hob_list()).unwrap_or(P::default_performance_config());
```
makubacki: `unwrap_or_else()` would prevent evaluating the argument (`P::default_performance_config()`) unless needed.

makubacki (dismissed):

Javagedes:
I will be reviewing this today, please do not merge :)

docs/src/dxe_core/performance_measurement.md
```diff
@@ -0,0 +1,108 @@
...
+## Layers
+
+- **SDK** - Shared types and the service contract. It defines the interfaces and common structures used by all layers.
+- **DXE Core** - The measurement engine. `CorePerformance` implements `PerformanceMeasurement` and owns all
+  global state and will process new performance records. This is implemented in the core to ensure early
+  availability of performance data.
+- **Component** - The DXE integration and interoperability layer. It consumes the service and, at boot,
+  publishes the FBPT for the ACPI FPDT, installs the `EdkiiPerformanceMeasurement` protocol for C drivers,
+  installs the `PerformanceProperty` configuration table, and optionally merges Management Mode (MM) performance
+  records.
```
Javagedes, line 20: I like this abstraction, however does it need to be a a standalone crate? e.g. is there a scenario where a platform would not want to publish this? There are other internal service / component combinations that produce the UEFI equivalent protocols, all from inside the core. e.g. the DxeCpu component & cpu_arch_protocol.

docs/src/integrate/dxe_core.md
```diff
@@ -439,30 +439,44 @@ impl CpuInfo for ExamplePlatform {
...
-# extern crate patina_performance;
+```rust,ignore
+use patina::performance::config::PerformanceConfig;
 use patina_dxe_core::*;
 
 struct ExamplePlatform;
 
+impl PlatformInfo for ExamplePlatform {
+    // Optional override if the platform does not publish the performance configuration HOB.
+    fn default_performance_config() -> PerformanceConfig {
```
Javagedes: There has been a reluctance in the past to have actual configuration in the `PlatformInfo` config, instead pushing it off to a sub trait. Are we OK with this?

patina_dxe_core/src/lib.rs
makubacki: This is needed because we don't want to always require the HOB, right?

sdk/patina/src/component/service/performance.rs
```diff
@@ -0,0 +1,187 @@
...
+        guid: Option<&efi::Guid>,
+        string: Option<&str>,
+        ticker: u64,
+        address: usize,
+        perf_id: u16,
+        attribute: PerfAttribute,
+    ) -> Result<(), Error>;
+
+    /// Adds an already-formed generic performance record to the FBPT.
+    fn add_generic_record(&self, record: GenericPerformanceRecord<&[u8]>) -> Result<(), Error>;
```
Javagedes: I think it would be better if you just defined a DST Record type here and used zerocopy.

```rust
#[derive(zerocopy::things)]
struct Record {
    record_type: u16,
    length: u8,
    revision: u8,
    data: [u8]
}

...
fn add_generic_record(&self, record: &Record) -> Result<(), Error>;
```

What this allows is other types to also utlize zerocopy to transmute to bytes and then to Record. And if you wanted to get really fancy you could make an extension trait to use generics:

```rust
pub trait PerformanceMeasurementExt {
    fn add_record<R: TryInto<&Record>>(&self, record: R) -> Result<(), Error>;
}

impl PerformanceMeasurementExt for Service<dyn PerformanceMeasurement> {
    fn add_record<R: TryInto<&Record>>(&self, record: R) -> Result<(), Error> {
        self.add_generic_record(record.try_into()?)
    }
}
```

sdk/patina/src/component/service/performance.rs
```diff
@@ -0,0 +1,187 @@
...
+        error::Error,
+        measurement::CallerIdentifier,
+        record::{GenericPerformanceRecord, known::KnownPerfId},
+    },
+    uefi_protocol::performance_measurement::PerfAttribute,
+};
+
+/// Service that records firmware performance measurements into the Firmware Basic Boot Performance Table (FBPT).
+///
+pub trait PerformanceMeasurement: Send + Sync {
```
Javagedes: I don't really have a better name, but `PerformanceMeasurement` sounds like a struct more then something providing functionality.

sdk/patina/src/performance/config.rs
```diff
@@ -0,0 +1,58 @@
...
+//! Copyright (c) Microsoft Corporation.
+//!
+//! SPDX-License-Identifier: Apache-2.0
+//!
+use crate::{BinaryGuid, component::hob::FromHob};
+
+/// The configuration for performance measurement.
+#[derive(Debug, Clone, Copy, zerocopy_derive::FromBytes)]
+#[repr(C, packed)]
+pub struct PerformanceConfig {
```
Javagedes, line 18: Does this make sense here? To me, this should be in patina_dxe_core similar to `GicBases`

sdk/patina/src/performance.rs
```diff
@@ -8,9 +8,8 @@
 //!
 //! SPDX-License-Identifier: Apache-2.0
 //!
+pub mod config;
```
Javagedes, line 11: I question how much of this `performance` module needs to exist anymore. The way I see it, The user interface for working with the service should be in `patina/component/service/performance.rs` and any internal funcitonality should just be in `patina_dxe_core`.

patina_dxe_core/src/performance.rs
```diff
@@ -0,0 +1,773 @@
...
+};
+
+use crate::{cpu::PerfTimer, protocols::PROTOCOL_DB, tpl_mutex::TplMutex};
+
+use r_efi::{
+    efi,
+    protocols::device_path::{Media, TYPE_MEDIA},
+};
+
+static LOAD_IMAGE_COUNT: AtomicU32 = AtomicU32::new(0);
```
Javagedes: We should not have these global statics now that the separation between the sdk and core has been fixed. `CorePerformance` should be a instanced struct with data in it, and become field of the patina dxe core. Measurement calls should properly called on the instance. 

What we are doing here adds more state that needs to be reset during tests among the other issues we've discussed regarding global statics.

docs/src/integrate/dxe_core.md
makubacki: You may have missed a conversation on this topic in https://github.com/OpenDevicePartnership/patina/pull/1628#discussion_r3547597007. Perhaps we can keep it going there for continuity. I'd prefer to avoid this as-is in `PlatformInfo`, but I also think the configuration here might be tied to larger points around how performance will be configured.

sdk/patina/src/performance.rs
makubacki: Not to take away from the previous comment, but at this point, we should also factor MM (potential code sharing) into a performance refactor.

docs/src/integrate/dxe_core.md
cfernald: Resolving this to comment there.

patina_dxe_core/src/lib.rs
cfernald: @makubacki, correct.

@Javagedes, responding to your other comment. I agree I'd rather not making the top level structure a dumping ground, but i'm also reluctant to make platform define a bunch of empty impls (e.g. `PerformanceInfo`). If we have default associated types, then this wouldn't be such a big deal.

An alternative to traits would be to instead have configuration structures, like PerformanceConfig. This does allow for easier less-taxing defaults.

I do also like `gdbstub`'s approach to this where you have functions that can optionally return a `Self` implementation of a trait if desired: https://docs.rs/gdbstub/latest/gdbstub/target/ext/index.html#how-protocol-extensions-work---inlineable-dyn-extension-traits-idets

cfernald: Discussed offline, going to switch to this: Will leave open for feedback on this though:

```rust
    /// The performance measurement configuration used when no performance configuration HOB is present.
    ///
    /// Defaults to disabled. Platforms may override this option to control the default behavior of the performance
    /// measurement service when no configuration HOB is present.
    const DEFAULT_PERFORMANCE_CONFIG: PerformanceConfig = PerformanceConfig::new();
```

patina_dxe_core/src/lib.rs
cfernald: Resolving this as I'm changing the default to a const so `unwrap_or_else` is more appripriate now.

docs/src/dxe_core/performance_measurement.md
cfernald: I think that the place where it makes the most since is the ACPI table, this way the ACPI service does not need to be consumed by the core. The protocol is a bit more debatable, but I figured since we already have the component it mades sense to offline everything that doesn't _have to_ be in the core to the component. 

I'm open to suggestions on other organization though.

sdk/patina/src/performance.rs
cfernald: My intention was that a second refactor would likely happen between the dxe_core and internal_core crates, but I decided to not try to predict that line, but that would be a non-breaking refactor since it would be internal to the patina cores at that point. Moving the config would break that though.

sdk/patina/src/performance/config.rs
cfernald: I think because this is a reflection of a HOB from Patina-edk2, and will be used in StMM core, it makes sense to keep it here. The other code in this section though perhaps not.

patina_dxe_core/src/performance.rs
cfernald: yeah, the testing is a bit awkward though, because when the dispatcher, for example, calls into the performance it will have to do `crate::instance()::performance::perf_image_start_begin(...)` which means it's still global state either way. 

I wonder if all sub-structures need a referenced back to the core. e.g. 

```rust
/// The internal state of the PI Dispatcher.
pub(crate) struct PiDispatcher<P: PlatformInfo> {
    /// Reference to the parent core
    core: &'static Core<P>,

    /// State for the dispatcher itself.
    dispatcher_context: TplMutex<DispatcherContext>,
    /// Image management data for executing images.
    image_data: TplMutex<image::ImageData>,
    /// Debug image data managing the debug image info table published as a configuration table.
    debug_image_data: RwLock<debug_image_info_table::DebugImageInfoData>,
    /// State tracking firmware volumes installed by the Patina DXE Core.
    fv_data: TplMutex<fv::FvProtocolData<P>>,
    /// Section extractor used when working with firmware volumes.
    section_extractor: CoreExtractor<P::Extractor>,
}
```

This seems like a pain though. Let me know what you think.

Javagedes: Well my idea would be more like this:

```rust
pub(crate) struct PiDispatcher<P: PlatformInfo> {
    /// State for the dispatcher itself.
    dispatcher_context: TplMutex<DispatcherContext>,
    /// Image management data for executing images.
    image_data: TplMutex<image::ImageData>,
    /// Debug image data managing the debug image info table published as a configuration table.
    debug_image_data: RwLock<debug_image_info_table::DebugImageInfoData>,
    /// Performance measuring
    performance: Service<dyn PerformanceMeasurement>,
    /// State tracking firmware volumes installed by the Patina DXE Core.
    fv_data: TplMutex<fv::FvProtocolData<P>>,
    /// Section extractor used when working with firmware volumes.
    section_extractor: CoreExtractor<P::Extractor>,
}

impl<P: super::PlatformInfo> super::PiDispatcher<P> {
    pub fn load_image(
        &self,
        boot_policy: bool,
        parent_image_handle: efi::Handle,
        file_path: Option<NonNull<Protocol>>,
        image: Option<&[u8]>,
    ) -> Result<efi::Handle, ImageStatus> {
        // Calls the lambda if performance is initialized.
        self.performance.map_or_default(|s| s.perf_load_image_begin(core::ptr::null_mut());
        
        /* Rest */
     }
}
```

cfernald: Makes sense. I don't think the `dyn` is needed, but I will move to an instantiated reference model.

makubacki:
Note: Added `[major]` prefix to PR title per [RFC: Patina Release Process](https://github.com/OpenDevicePartnership/patina/blob/main/docs/src/rfc/text/0015-patina-release-process.md).

sdk/patina/src/performance.rs
cfernald: Thinking about this, I want to do another round of refactoring haha. I think there is a lot of cruft here still that can be removed/moved..

patina_dxe_core/src/performance.rs
cfernald: Unfortunately, the measurements here are not consistent with EDK2's performance. The measurements are this:

```rust
pub enum Measurement {
    /// Dispatch modules entry point execution
    StartImage = bit!(0),
    /// Load a dispatched module.
    LoadImage = bit!(1),
    /// Diver binding support function call.
    DriverBindingSupport = bit!(2),
    /// Diver binding start function call.
    DriverBindingStart = bit!(3),
    /// Diver binding stop function call.
    DriverBindingStop = bit!(4),
}
```

This code basically treads disabled and enabled with no measurements as equivalent. This is true before the refactor, but it is a bit confusing so I will just separate it for clarity.

makubacki: That was intentional at one point because `PcdPerformanceLibraryPropertyMask` was confusing (opt out vs opt in) and we didn't see a need to carry the problem forward to an independent Patina config interface.

  EDK II Format (`PcdPerformanceLibraryPropertyMask`):
    - BIT0: Enable Performance Measurement (master switch, handled separately by EnableComponent)
    - BIT1: DISABLE Start Image Logging
    - BIT2: DISABLE Load Image logging
    - BIT3: DISABLE Binding Support logging
    - BIT4: DISABLE Binding Start logging
    - BIT5: DISABLE Binding Stop logging
    - BIT6: DISABLE all other general Performance Measurements

  Patina Format (`enabled_measurements`):
    - BIT0 (1):  ENABLE StartImage
    - BIT1 (2):  ENABLE LoadImage
    - BIT2 (4):  ENABLE DriverBindingSupport
    - BIT3 (8):  ENABLE DriverBindingStart
    - BIT4 (16): ENABLE DriverBindingStop

cfernald:
@makubacki @Javagedes , I've moved more code into the implementations and out of the SDK. I also moved the CorePErformance to be instanced. This has some downstream affect's i don't love with having to pass around references and even one static. I think we may need to evaluate late-retrieval of core services, or perhaps there is already a better way.

@os-d @magravel , i cleared your review since there has been significant changes.

components/patina_performance/src/component/performance.rs
```diff
@@ -413,12 +311,13 @@ impl<'a> Iterator for PerformanceRecordIterator<'a> {
             ))));
         }
```
makubacki, line 313: It looks like this assignment will always cause `self.bytes.len()` to be `0`.


```suggestion
        if rec_len > self.bytes.len() {
            let available = self.bytes.len();
            self.bytes = &[];
            return Some(Err(MmPerformanceError::RecordError(alloc::format!(
                "Truncated record (needed {}, had {})",
                rec_len,
                available
            ))));
        }
```

components/patina_performance/src/component/performance.rs
makubacki: This was pre-existing but it looks like this should `continue` instead of `break`? It looks like `PerformanceRecordIterator` attempts to advance iteration in case of `Err` to avoid missing records.

components/patina_performance/src/component/protocol.rs
makubacki: It looks like the FBPT full message was already logged in `report_add_record_error()` and doesn't need to be logged again here.

patina_dxe_core/src/performance.rs
```diff
@@ -0,0 +1,1050 @@
...
+        match self.performance_table.try_lock() {
+            Some(table) => Ok(table.published_table_size()),
+            None => Err(EfiError::InvalidParameter.into()),
+        }
+    }
+
+    fn publish_table(&self, buffer: &'static mut [u8]) -> Result<(), Error> {
+        match self.performance_table.try_lock() {
+            Some(mut table) => table.publish_table(buffer).map(|_| ()),
+            None => Err(EfiError::InvalidParameter.into()),
```
makubacki: Is there a reason to return `EfiError:InvalidParameter` instead of something like `OutOfResources` or `NotReady` if the lock cannot be acquired?

components/patina_performance/src/component/performance.rs
```diff
@@ -490,45 +390,142 @@ where
...
+        log::error!("Performance: Fail to allocate FBPT buffer.");
+        return;
+    };
+    let fbpt_address = buffer.as_ptr() as usize;
+
+    // Provide the allocated memory for the service to serialize the table into.
+    if let Err(e) = performance.publish_table(buffer) {
+        log::error!("Performance: Fail to serialize FBPT: {e:?}");
+        return;
+    }
```
makubacki, line 440: Could the reserved memory buffer from `allocate_fbpt_buffer()` could be freed if `performance.publish_table(buffer)` fails?

makubacki (approved):
@cfernald, I'm going to go ahead and approve now to avoid blocking the PR but if you could leave it open at least through Monday, I'd like to do another more detailed pass through.

components/patina_performance/src/component/performance.rs
cfernald: So it does, I'll fix this.

components/patina_performance/src/component/performance.rs
cfernald: In the event of an error it just empties the remaining bytes, so I don't think continue would do anything other then just return `None` on the next call, right?

components/patina_performance/src/component/protocol.rs
cfernald: Yeah, I'll drop this. Not needed the redundant lock it a plus.

patina_dxe_core/src/performance.rs
cfernald: Switching to `AccessDenied` which feels the most appropriate.

components/patina_performance/src/component/performance.rs
cfernald: Adding

components/patina_performance/src/component/performance.rs
makubacki: The issue I have is that the `next()` implementation does both:

- "Attempts to advance iteration":

  ```rust
          let header = match PerformanceRecordHeader::try_from(self.bytes) {
              Ok(h) => h,
              Err(err) => {
                  self.bytes = self.bytes.get(1..).unwrap_or(&[]);
                  return Some(Err(MmPerformanceError::RecordError(err.into())));
              }
          };
  ```

- "Empties the remaining bytes":

  ```rust
          if rec_len > self.bytes.len() {
              let available = self.bytes.len();
              self.bytes = &[];
              return Some(Err(MmPerformanceError::RecordError(alloc::format!(
                  "Truncated record (needed {}, had {})",
                  rec_len,
                  available
              ))));
          }
  ```

https://github.com/OpenDevicePartnership/patina/blob/edf2cd9b7b1dcae51b33d9480f249b581ad7a091/components/patina_performance/src/component/performance.rs#L384-L427

I take the "Attempts to advance iteration" cases to exist so `continue` would be possible while the "Empties the remaining bytes" case to exist to return `None` on the next call which would naturally break iteration (likely as intended).

makubacki:
> @cfernald, I'm going to go ahead and approve now to avoid blocking the PR but if you could leave it open at least through Monday, I'd like to do another more detailed pass through.

Considering that additional performance refactoring and feature changes will be made for https://github.com/OpenDevicePartnership/patina/issues/467 and potentially for MM infrastructure support, this looks good to me as an incremental improvement in the right direction and I'm done reviewing.

components/patina_performance/src/component/performance.rs
cfernald: Ah, i missed that before. I think that's fair. I will switch to continue and if it bails on the next iteration then so be it.

patina_dxe_core/src/pi_dispatcher.rs
```diff
@@ -88,6 +88,8 @@ pub(crate) struct PiDispatcher<P: PlatformInfo> {
     fv_data: TplMutex<fv::FvProtocolData<P>>,
     /// Section extractor used when working with firmware volumes.
     section_extractor: CoreExtractor<P::Extractor>,
+    /// Optional performance service reference.
+    performance: Once<&'static performance::CorePerformance>,
```
Javagedes: Is there a reason you want to use this instead of `Service<dyn PerformanceManager>`? The service wrapper pretty much does exactly what you are doing here. 

Using `Service<dyn PerformanceManager>` will allow you to make calls like this:

```rust
self.performance.replace(...)` // instead of your `call_once` logic

// and for measuring
self.performance.map_or_default(|perf| perf.perf_load_image_begin(...))
```

Javagedes (approved):

patina_dxe_core/src/pi_dispatcher.rs
cfernald: Mostly just avoiding the `dyn` because the type here is known. It doesnt really make a difference though, so can switch it if you think that is better.

Javagedes: I would prefer using the Service interface, but if it causes too many issues with fully integrating, then it's not that big of a deal to keep what you have.

Javagedes (approved):

PR #1630, opened 2026-07-09
[major] Remove `SerialIO` interior mutability and cleanup serial implementations [REBASE & FF]

### LINKED ISSUE #1567: [Bug]: SeralIO should use mutable references

### Current Behavior

The current Serial IO implementation assumes internal mutability. This is problematic because providing the appropriate locking for serial is unconventional because of its existence outside of the general extensibility model for Patina. 

### Expected Behavior

SerialIO should me made to use mutable references, and the SDK should provide a `StaticSerialIO<T>` struct that will provide the appropriate interior mutability for protecting T from concurrent or contested use

### Steps To Reproduce

N/A

### Build Environment

```markdown
N/A
```

### Version Information

```text
N/A
```

### Urgency

Medium

I will fix it

No maintainer feedback needed

### Anything else?

_No response_

cfernald:
### Change SerialIO trait to be mutable

The current `SerialIO` implementation takes `&self` instead of `&mut self`
for its method, forcing all serial implementations to implement
interior mutability. However, because of the nature of the serial port
use cases, the typical locking mechanisms that avoid deadlocks, such as
the `TplMutex` is not readily accessible. Because of this, internal
implementations have one-off, and at times questionable, interior mutability
implementations.

This commit removes the expectation on the serial port to ensure this,
and instead provides a re-usable `SharedSerial` wrapper that implements
interior mutability in a suitable way for serial port use cases.

### patina: Simplify serial implementations

This commit takes advantage of the new mutable references to cleanup
the serial port implementations. Namely by dropping re-initialization of
interior references used as a workaround for the lack of mutability
in the `SerialIO` trait.

### patina: split uart module submodules

This commit splits the null, 16550, and pl011 UART implementations into
their own submodules under the `uart` module. This better aligns to
general code organization practices. This additionally removes some
unnecessary conditional complication.

- [x] Impacts functionality?
- [ ] Impacts security?
- [x] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- [x] boot to shell on q35
- [x] boot to shell on armvirt

## Integration Instructions

1. All custom `SerialIO` implementations should drop any interior mutalability and adopt the `&mut self` references.
2. Users of `Uart16550` must swap to the new `new_io` and `new_mmio` instantiations.


Javagedes (approved):

os-d (approved):

cfernald:
@joschock, you previously expressed some thoughts on this, so wanted to make sure you saw this PR. Will leave it open for a bit.

sdk/patina/src/serial/shared.rs
```diff
@@ -0,0 +1,91 @@
...
+//!
+
+use spin::Mutex;
+
+use crate::{arch::with_interrupts_disabled, serial::SerialIO};
+
+/// A wrapper that provides guaranteed single threaded exclusive access to the serial port **in the patina environment**.
+/// This is a special implementation because of unique use-case of serial ports that sit outside of the general
+/// Patina TPL based locking model.
+pub struct SharedSerial<T: SerialIO> {
```
joschock, line 20: Is it worth a doc comment mentioning that these traits cannot be re-entrant?

sdk/patina/src/serial/shared.rs
```diff
@@ -0,0 +1,91 @@
...
+}
+
+impl<T: SerialIO> SharedSerial<T> {
+    /// Creates a new shared serial port wrapper.
+    pub const fn new(serial: T) -> Self {
+        SharedSerial { serial: Mutex::new(serial) }
+    }
+
+    fn get(&self) -> impl core::ops::DerefMut<Target = T> + '_ {
+        self.serial.try_lock().expect("Reentrant access to serial port.")
```
joschock: Perhaps this could be fallible (Result or Option) with a debug assert? Gives some resilience against re-entrancy edge cases in release.

joschock (approved):

sdk/patina/src/serial/shared.rs
```diff
@@ -0,0 +1,91 @@
...
+        with_interrupts_disabled(|| self.get().init());
+    }
+
+    /// Write a buffer to the serial port.
+    pub fn write(&self, buffer: &[u8]) {
+        with_interrupts_disabled(|| self.get().write(buffer));
+    }
+
+    /// Read a byte from the serial port, blocking until a byte is available.
+    pub fn read(&self) -> u8 {
```
makubacki: I think it's also worth calling out that interrupts are disabled while blocking.

sdk/patina/src/serial/shared.rs
makubacki: Yeah, I think we should avoid turning serial contention/errors into a fatal condition.

makubacki:
Note: This already does not need QEMU PR Validation to pass since it is going into the `major` branch. However, please try to get each file above 80% test coverage if possible.

sdk/patina/src/serial/uart/pl011.rs
```diff
@@ -0,0 +1,194 @@
...
+    /// # Safety
+    ///
+    /// The given base address must point to the MMIO control registers of a
+    /// PL011 device, which must be mapped into the address space of the process
+    /// as device memory and not have any other aliases for the lifetime of the
+    /// returned instance.
+    pub const unsafe fn new(base_address: usize) -> Self {
+        // SAFETY: The caller guarantees `base_address` points to an exclusively-owned PL011
+        // register block mapped as device memory, satisfying `UniqueMmioPointer::new`.
+        let regs = unsafe { UniqueMmioPointer::new(NonNull::new(base_address as *mut Pl011Registers).unwrap()) };
```
makubacki: If this will panic on null base, a message would add a little more clarity.


```suggestion
        let regs = unsafe { UniqueMmioPointer::new(NonNull::new(base_address as *mut Pl011Registers).expect("UART PL011 should have a non-null base address")) };
```

sdk/patina/src/serial/std.rs
```diff
@@ -33,3 +33,22 @@ impl super::SerialIO for Terminal {
         }
     }
 }
+
+#[cfg(test)]
+#[coverage(off)]
```
makubacki: Might just be a matter of rebasing your branch, but please gate on `coverage_nightly` now.

```suggestion
#[cfg_attr(coverage_nightly, coverage(off))]
```

makubacki (approved):

sdk/patina/src/serial/shared.rs
cfernald: Switched all returns to results and updated callers to take the appropriate action (or inactions)

PR #1640, opened 2026-07-10
patina_dxe_core: Consistently apply test guards [Rebase & FF]

### LINKED ISSUE #1637: Look into intermittent unit test failures
I've noticed an occasional test failure when rerunning unit tests.

- `allocator::tests::process_hob_allocations_should_allocate_fv_in_system_memory()` sometimes results in: "Failed to add initial region to GCD: OutOfResources"
- `allocator::tests::process_hob_allocations_should_handle_stack_attribute_set_failure()` has also failed.

This issue tracks looking to these failures further.

makubacki:
## Description

Resolves #1637

Currently, tests use `test_support::with_global_lock()` to often wrap the test and its setup/teardown logic inside the global test lock. This ensures everything happens exclusively while holding the lock.

Over time, test cleanup because more intricate and an issue so a "test guard" (`test_support::StateGuard`) was added that accepts a generic cleanup closure where the guard ensures that the cleanup code is run even if the test panics (as it is at the same scope as the test invocation).

This all works well enough but:

1. Some module tests are missing a guard.
2. Common globals (GCD, protocol database, and allocator state) should really always be reset before each test.
3. Guard creation often occurrred after test initialization logic which meant it was not in place to clean if the initialization logic itself panicked.

This change adds a generic `test_support::reset_global_state()` function that can conveniently be used in guards, adds a guard to missing patina_dxe_core test modules, and moves guard creation to the top of test wrappers.

---

Specifically, in this most recent case of flakiness, it appears that tests in runtime.rs were leaving dirty state in the globals getting picked up by init code in allocator.rs that was occasionally failing because of it. Those runtime.rs tests now exit in clean state due to the guard added.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [x] Includes documentation?

## How This Was Tested

- `cargo make all`
- `cargo make test` about 10 times

## Integration Instructions

- N/A

### DIFF
--- a/patina_dxe_core/src/pi_dispatcher.rs
+++ b/patina_dxe_core/src/pi_dispatcher.rs
@@ -926,6 +926,9 @@ mod tests {
         F: Fn() + std::panic::RefUnwindSafe,
     {
         test_support::with_global_lock(|| {
+            // Reset global state on exit (even if setup panics) so nothing leaks to the next test.
+            let _guard = test_support::StateGuard::new(test_support::reset_global_state);
+
             // SAFETY: Test-only initialization of the protocol database occurs under the global lock.
             unsafe { test_support::init_test_protocol_db() };
             f();
--- a/patina_dxe_core/src/pi_dispatcher.rs
+++ b/patina_dxe_core/src/pi_dispatcher.rs
@@ -925,10 +925,7 @@ mod tests {
     where
         F: Fn() + std::panic::RefUnwindSafe,
     {
-        test_support::with_global_lock(|| {
-            // Reset global state on exit (even if setup panics) so nothing leaks to the next test.
-            let _guard = test_support::StateGuard::new(test_support::reset_global_state);
-
+        test_support::with_clean_global_lock(|| {
             // SAFETY: Test-only initialization of the protocol database occurs under the global lock.
             unsafe { test_support::init_test_protocol_db() };
             f();
--- a/patina_dxe_core/src/test_support.rs
+++ b/patina_dxe_core/src/test_support.rs
@@ -268,6 +268,27 @@ pub(crate) fn with_global_lock<F: Fn() + std::panic::RefUnwindSafe>(f: F) -> Res
     })
 }
 
+/// Like [`with_global_lock`], but additionally resets the shared global state via
+/// [`reset_global_state`] both before running `f` (so the test starts from a clean slate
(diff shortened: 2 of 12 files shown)


patina_dxe_core/src/test_support.rs
```diff
@@ -333,6 +333,29 @@ pub(crate) unsafe fn reset_allocators() {
...
+/// so that cleanup runs even if setup panics.
+///
+/// This intentionally does not reset subsystem-specific state (for example the system table or
+/// per-module statics) so it is more broadly applicable.
+///
+/// ## Locking
+/// Must be called with the global test lock held (see [`with_global_lock`]). It does not acquire
+/// the lock itself.
+pub(crate) fn reset_global_state() {
+    // SAFETY: Callers hold the global test lock, so no other test can access the state concurrently.
```
cfernald, line 372: Consider adding an assert that `try_lock` on the global lock returns `None` to catch misuse.

patina_dxe_core/src/pi_dispatcher.rs
```diff
@@ -926,6 +926,9 @@ mod tests {
         F: Fn() + std::panic::RefUnwindSafe,
     {
         test_support::with_global_lock(|| {
```
cfernald: Seems like a common pattern, should we instead provide a `test_support::with_clean_global_lock(||` or similar to do both.

cfernald (approved):

os-d (approved):

patina_dxe_core/src/pi_dispatcher.rs
makubacki: I added this in a second commit (https://github.com/OpenDevicePartnership/patina/pull/1640/changes/8f90c5d4542a652fdb6fc2325b6bf563b99c6caa) as a refactor to improve ergonomics in the first commit.

patina_dxe_core/src/test_support.rs
makubacki: This one I'll probably leave as an open to consider for the future. I think it is part of a larger design decision of whether to allow any parallelization in `patina_dxe_core` tests. If we do, some contention would be expected between OS threads. But even then, we wouldn't expect re-entrancy within a given test thread. In that case, maybe a thread-local re-entrancy detection mechanism could help. In any case, I'm a hesitant about making such a change here now.

PR #1644, opened 2026-07-16
[REBASE & FF] Reorganize cpu mod and Enable Interrupts Before Sleeping

os-d:
## Description

EfiCpu: Move Relevant Functions to SDK
--
The EfiCpu trait existed before the arch mod in the SDK.

This commit moves functions that are simply wrappers over basic architectural functionality to the SDK.

EfiCpu: Move Remainder of Trait to DXE Core
--
The final remaining function in the EfiCpu trait is DXE specific logic for initializing the FPU and GDT on X64 systems.

This commit moves that logic to the DXE core and drops the cpu module in patina_internal_cpu.

SDK: Allow Exprs in write_sysreg Macro
--
This allows an expr instead of just a literal in the write_sysreg macro. This allows defining consts to use in the macro instead of just literal values.

This commit then uses that to make DAIF_WR_IRQ_BIT a const instead of a literal. It also updates x64 comments to be clear on why preserves_flags is used.

SDK/DXE: Update sleep to enable_interrupts_and_sleep
--
In no case should FW ever sleep with interrupts off. The only place we sleep is gated by TPL_APPLICATION, so this is not strictly required, but make the API safer by enabling interrupts and then sleeping. This also allows for more performant sleeping in the case of device interrupts, because interrupts are only enabled (for x64) after the next instruction, so there is not a window where the interrupt can be missed.

Move Timer Operations to SDK
--
The SDK timer operations were stubbed out and real functionality existed in the DXE Core.

This moves the DXE Core arch timer operations to the SDK and removes the stub impls there.

## How This Was Tested

Booting Q35 and ArmVirt.

## Integration Instructions

N/A. The changes here are for an internal only crate.


core/patina_internal_cpu/src/arch.rs
```diff
@@ -0,0 +1,39 @@
...
+//!
+//! SPDX-License-Identifier: Apache-2.0
+//!
+
+use crate::cpu::EfiCpu;
+#[cfg(all(target_os = "uefi", any(target_arch = "x86_64", target_arch = "aarch64")))]
+use core::arch::asm;
+
+/// A trait providing architecture-specific CPU management operations.
+pub trait CpuMgmt {
```
cfernald: Is there a reason for the trait?, why not just implement on EfiCpu directly since there is just the one implementation?

sdk/patina/src/arch.rs
cfernald: The intention of this was to have `ArchSupport` require all of the following, just a way to help organize.

so:
```rust
trait ArchSupport: Interrupts, CpuMgmt  {}
```

sdk/patina/src/arch.rs
```diff
@@ -77,3 +80,85 @@ where
     }
     result
 }
+
+/// Architecture-specific CPU operations.
+///
+/// This trait is implemented by the per-architecture zero-sized `Arch` type. Callers should use
+/// the free functions in this module rather than referencing the trait directly.
+pub trait CpuMgmt {
```
cfernald: My original thinking on this was to try to keep the ArchSupport sub traits organized by functionality. To that end, unless you disagree with that intention, I think it would be better to separate into `CacheManagment` and `Timer`, and maybe put `sleep` under the existing Interrupt group

sdk/patina/src/arch/x64.rs
```diff
@@ -66,3 +71,73 @@ impl super::Interrupts for X64 {
...
+            return Err(EfiError::InvalidParameter);
+        }
+        Ok(timer_period())
+    }
+
+    fn cache_writeback_granule() -> u32 {
+        CACHE_WRITEBACK_GRANULE
+    }
+
+    #[cfg_attr(coverage_nightly, coverage(off))]
```
cfernald: nit, the whole file is already bypassed on coverage

sdk/patina/src/arch/x64.rs
```diff
@@ -66,3 +71,73 @@ impl super::Interrupts for X64 {
...
+        #[cfg(target_os = "uefi")]
+        // SAFETY: This halts the CPU until the next interrupt, which has no memory safety implications.
+        unsafe {
+            asm!("hlt");
+        }
+    }
+}
+
+fn asm_wbinvd() {
+    #[cfg(not(test))]
```
cfernald: Same below, but this file shouldnt be invoked on test ad the null instance will be used on the matrix in super

sdk/patina/src/arch/aarch64.rs
```diff
@@ -164,3 +167,108 @@ pub fn get_current_el() -> AArch64El {
...
+
+        // CWG is Log2 of the max size in words
+        if cwg > 0 {
+            4 << cwg
+        } else {
+            crate::base::SIZE_2KB as u32 // Default to 2K if register contains 0 per Armv8-A spec
+        }
+    }
+
+    #[cfg_attr(coverage_nightly, coverage(off))]
```
cfernald: nit: file already uncovered.

sdk/patina/src/arch/aarch64.rs
```diff
@@ -164,3 +167,108 @@ pub fn get_current_el() -> AArch64El {
...
+    fn sleep() {
+        #[cfg(target_os = "uefi")]
```
cfernald: Use of this file is gated already

sdk/patina/src/arch/null.rs
```diff
@@ -21,3 +24,24 @@ impl super::Interrupts for NullArch {
...
+
+    fn get_timer_period(_timer_index: u32) -> Result<u64, EfiError> {
+        Ok(0)
+    }
+
+    fn cache_writeback_granule() -> u32 {
+        64
+    }
+
+    #[cfg_attr(coverage_nightly, coverage(off))]
```
cfernald: file is already uncovered

sdk/patina/src/arch.rs
os-d: I agree with the intention, I also just wasn't sure if we wanted a misc one as well because having a few traits with a single function seemed odd. But I think `CacheManagement`, `Timer` and putting `enable_interrupts_and_sleep` under `Interrupts` seems reasonable.

core/patina_internal_cpu/src/arch.rs
os-d: Can you check the final state and see if you still have a question? This is marked outdated, I think you reviewed one of the early commits that only moved things partway.

sdk/patina/src/arch/x64.rs
os-d: Will drop all of these and cfg not test

sdk/patina/src/arch.rs
cfernald: oh, and the trait shouldn't be pub

sdk/patina/src/arch/x64.rs
```diff
@@ -66,3 +71,73 @@ impl super::Interrupts for X64 {
...
+    // SAFETY: Invalidating the cache has no memory safety implications.
+    unsafe {
+        asm!("invd");
+    }
+}
+
+/// Reads the timestamp counter used for the CPU timer value. Currently returns 0 until a real
+/// implementation is provided.
+fn read_tsc() -> u64 {
```
os-d: I just noticed these are dead functions and the core is already using SDK functions. I will drop these.

sdk/patina/src/arch.rs
```diff
@@ -77,3 +80,85 @@ where
...
+pub fn flush_data_cache(start: efi::PhysicalAddress, length: u64, flush_type: CpuFlushType) -> Result<(), EfiError> {
+    <Arch as CpuMgmt>::flush_data_cache(start, length, flush_type)
+}
+
+/// Returns a timer value from the current architecture's CPU.
+pub fn get_timer_value(timer_index: u32) -> Result<u64, EfiError> {
+    <Arch as CpuMgmt>::get_timer_value(timer_index)
+}
+
+/// Returns a timer period from the current architecture's CPU.
```
cfernald: Probably worth calling out in the comment that on some platforms this may not be supported, and callers should use the timer Service instead.

sdk/patina/src/arch/x64.rs
cfernald: Could you instead converge under the new definition. That is for consistency just converge under the arch::functions and remove the arch specific abstraction in the core for CorePerfTimer?

os-d: Yes, I agree with that. When you sent the code snippet offline, I thought it was in the SDK already

sdk/patina/src/arch.rs
os-d: I thought I had done so, but apparently not.

patina_dxe_core/src/events.rs
```diff
@@ -177,10 +177,11 @@ unsafe extern "efiapi" fn wait_for_event(
...
         // event are the CPU architectural drivers which use it to enter a low power state until the next interrupt.
-        // Patina implements CPU architectural support as part of the core, so directly call the sleep() method to avoid
-        // exposing the idle event to outside consumers (this event group is not specified in UEFI or PI specs). In the
-        // event that a need arises to expose the idle event to consumers outside of Patina, it can be signaled here.
-        EfiCpu::sleep();
+        // Patina implements CPU architectural support in the SDK, so directly call the enable_interrupts_and_sleep()
+        // function to avoid exposing the idle event to outside consumers (this event group is not specified in UEFI or
+        // PI specs). In the event that a need arises to expose the idle event to consumers outside of Patina, it can be
+        // signaled here.
+        patina::arch::enable_interrupts_and_sleep();
```
joschock: nit: This feels side-effect-y to me. I agree that it's a programming error if you attempt to sleep with interrupts disabled for sure; but to me it doesn't follow that we should auto-enable them here and mask bad behavior. Put another way: calling wait_for_event with interrupts disabled is a programming error and should assert in debug rather than silently fixing. Could we do something like:
```rust
if (patina::arch::interupts_enabled()) {
  patina::arch::sleep();
} else {
  debug_assert!("wait_for_events called with interrupts disabled");
  //fall through without sleeping for release.
}
```

joschock (approved):

patina_dxe_core/src/events.rs
os-d: I think that's fine here. The real reason for this change was because a platform needs enable_interrupts_and_sleep and I think it is reasonable to have in the SDK. Which then led me to, well is it reasonable to have sleep only in the SDK? Maybe not and then a footgun is removed. Though its an unlikely one.

Since this was the only consumer of sleep in the tree, I decided to just remove it and now a potentially unsafe API is not possible (w/o a platform just doing it on their own). So it wasn't about this case ever.

It also could be 

```rust
if (patina::arch::interupts_enabled()) {
  patina::arch::enable_interrupts_and_sleep();
} else {
  debug_assert!("wait_for_events called with interrupts disabled");
  //fall through without sleeping for release.
}
```
or

```rust
if (!patina::arch::interupts_enabled()) {
  debug_assert!("wait_for_events called with interrupts disabled");
}

patina::arch::enable_interrupts_and_sleep();
```

Which is a bit silly, but then sleep in general goes away

joschock: I mean, it's a nit; so I don't really have a super strong opinion if you want to leave it this way. I'm just noting that enabling interrupts is not a side effect I would expect out of `wait_for_event` and I would much prefer for it to tell me I did a bad thing than fix it for me on the sly.

os-d: Yeah, I think that's fair. I'll update this to do the interrupt check when also checking TPL_APPLICATION (that's part of the implication of that check anyway) and add a debug_assert for the two conditions there (to catch which) and return invalid param. I'd be a little concerned about falling through. Probably better to just return up front.

sdk/patina/src/arch/x64.rs
```diff
@@ -65,4 +74,85 @@ impl super::Interrupts for X64 {
...
+        CACHE_WRITEBACK_GRANULE
+    }
+}
+
+impl super::Timer for X64 {
+    fn get_timer_value() -> u64 {
+        rdtsc()
+    }
+
+    fn get_timer_frequency() -> u64 {
```
cfernald: Considering this is fallible, and returns a 0 for something that is intended to be divided by, maybe returning a `Option<NonZeroU64>` is "safer"

cfernald (approved):

sdk/patina/src/arch/x64.rs
os-d: Fair. I'll update. I'm also making a fix because I accidentally dropped the cpu arch protocol period calculation. Which then leads to needing microsecond_delay, because that was also stubbed out

os-d:
@cfernald @joschock I have addressed all comments and then some :). I encountered duplication in the interrupts crate, so dealt with that here as well. Fixed the aarch64 unit test failure which was due to cfg[test] only working when running tests for that crate, when other crates (say patina_dxe_core) build for tests, they build all dependencies as not(test) and so the aarch64 asm was attempting to be run. This led to the stub.rs changes.

Note, QEMU PR validation is broken right now due to a policy change that doesn't let it run.

sdk/patina/src/arch.rs
```diff
@@ -14,8 +14,22 @@ use r_efi::efi;
...
-        type Arch = null::NullArch;
+        mod stub;
+        type Arch = stub::StubArch;
+
+        // Re-export the stub as the matching architecture module so that `patina::arch::aarch64` /
+        // `patina::arch::x64` paths (and the associated macros/types) resolve on host builds. Currently,
+        // only the pieces required for patina to run tests are re-exported.
+        #[doc(hidden)]
+        #[cfg(target_arch = "aarch64")]
+        pub mod aarch64 {
```
cfernald: Mention above, but seems reasonable to make those a mod in stub.rs that you just republish here: `pub use stub::aarch64`

sdk/patina/src/arch/stub.rs
```diff
@@ -0,0 +1,119 @@
...
+        None
+    }
+}
+
+//
+// aarch64-specific stubs. These mirror the public surface of the real `aarch64` module that
+// dependent crates reference from `patina::arch::aarch64` on host builds.
+//
+
+#[cfg(target_arch = "aarch64")]
```
cfernald, line 69: I would put this all in a submodule aarch64 (either in file file or a seperate), that was you can just republish that mod from arch.rs. That was you dont have to republish them one-by-one and you only have to write the arch check on the mod.

cfernald (approved):

joschock (approved):

sdk/patina/src/arch.rs
os-d: That was how I started, but the `stub` mod has to become `pub` then, which maybe we are okay with, because you can't make a mod in a private mod public.

os-d: Updated.

PR #1645, opened 2026-07-17
Serialize access to serial log

kuqin12:
## Description

When printing to the serial port from multiple processors, the log could be garbled due to interleaved access to the UART device.

This change adds a mutex to the serial logging device to serialize the prints from multiple processors.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This is tested on QEMU Q35 and physical hardware.

## Integration Instructions

N/A

Co-authored-by: Vineel Kovvuri[MSFT] <vineelko@microsoft.com>
Signed-off-by: Kun Qin <kun.qin@microsoft.com>

### DIFF
--- a/sdk/patina/src/log/serial_logger.rs
+++ b/sdk/patina/src/log/serial_logger.rs
@@ -8,8 +8,13 @@
 //!
 use crate::serial::SerialIO;
 use core::marker::Send;
+use spin::Mutex;
 
 use super::Format;
+
+/// Global lock serializing whole log records so concurrent callers do not
+/// interleave formatted output on the serial port.
+static LOG_LOCK: Mutex<()> = Mutex::new(());
 
 /// A Base implementation for a logger.
 ///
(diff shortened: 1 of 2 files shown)


cfernald:
I think this is no longer needed with this: https://github.com/OpenDevicePartnership/patina/pull/1630

Though the locking semantics are a bit different, so we may need to amend how serial logger handles failures from the `SharedSerial`

makubacki:
> I think this is no longer needed with this: #1630

Can we consolidate everything related to serial synchronization in `major`? Downside is it won't be released for a few weeks.

kuqin12:
> Though the locking semantics are a bit different, so we may need to amend how serial logger handles failures from the `SharedSerial`

Will give this a shot.

kuqin12:
@vineelko tested the change in major branch and verified it worked on physical platform. Closing this PR in favor of existing change in major.

PR #1650, opened 2026-07-22
Devpath macros

joschock:
## Description

Adds macros for build-time construction of device path byte arrays from string literals. This is an ergonomic change, but it helps change this:
```rust

   let mut path = [0u8; DEVICE_PATH_LENGTH];

    // ACPI(PNP0A03,0): PCI root bridge for segment 0.
    path[0] = 0x02;
    path[1] = 0x01;
    path[2] = ACPI_PCI_ROOT_NODE_LENGTH as u8;
    let pci_root_hid = 0x0A03_41D0u32.to_le_bytes();
    let mut index = 0;
    while index < pci_root_hid.len() {
        path[4 + index] = pci_root_hid[index];
        index += 1;
    }

    // Pci(0x11,0) on bus 0.
    let pci_offset = ACPI_PCI_ROOT_NODE_LENGTH;
    path[pci_offset] = 0x01;
    path[pci_offset + 1] = 0x01;
    path[pci_offset + 2] = PCI_NODE_LENGTH as u8;
    path[pci_offset + 4] = 0;
    path[pci_offset + 5] = 0x11;
```
into 
```rust
const path: &[u8] =  &devpath!("PciRoot(0)/Pci(0x11,0)");
```

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [x] Includes documentation?

## How This Was Tested

New macro includes unit tests and compilation tests. 

## Integration Instructions

N/A - usage is documented in the readme and rustdocs.

### DIFF
--- a/cspell.yml
+++ b/cspell.yml
@@ -223,6 +223,7 @@ words:
   - unmaps
   - unrecovered
   - unreg
+  - unrepresentable
   - unregisters
   - unspec
   - unsynchronized
--- a/cspell.yml
+++ b/cspell.yml
@@ -223,8 +223,8 @@ words:
   - unmaps
   - unrecovered
   - unreg
-  - unrepresentable
   - unregisters
+  - unrepresentable
   - unspec
   - unsynchronized
   - variadics
--- a/sdk/patina/README.md
+++ b/sdk/patina/README.md
@@ -38,9 +38,6 @@ The SDK is organized into the following root-level modules.
 | **standard** | Strict definitions of industry standards. |
 | **uefi** | UEFI specification definitions and wrappers. |
 
-An optional invocation-local `vendors { ... };` registry can define named vendor hardware nodes with fixed GUIDs and
-typed payload fields. See the `devpath!` API documentation for the schema syntax and supported field encodings.
-
 ## Feature Overview
 
 | Feature | Purpose |
--- a/sdk/patina/README.md
+++ b/sdk/patina/README.md
@@ -38,6 +38,9 @@ The SDK is organized into the following root-level modules.
 | **standard** | Strict definitions of industry standards. |
 | **uefi** | UEFI specification definitions and wrappers. |
 
+An optional invocation-local `vendors { ... };` registry can define named vendor hardware nodes with fixed GUIDs and
+typed payload fields. See the `devpath!` API documentation for the schema syntax and supported field encodings.
+
 ## Feature Overview
 
 | Feature | Purpose |
--- a/sdk/patina_macro/README.md
+++ b/sdk/patina_macro/README.md
@@ -96,3 +96,23 @@ fn watchdog_negative_path() -> Result {
     Ok(())
 }
 ```
+
+### `devpath!`
+
+- Converts a UEFI 2.11 Device Path From Text string literal into an owned `[u8; N]` at compile time.
+- Supports standard device path nodes, aliases, decimal and hexadecimal integers, named parameters, and multiple
+  device path instances.
+- Inserts End Instance nodes between top-level comma-separated instances and an End Entire node after the final
+  instance.
+- Requires no runtime parser, allocation, or Patina device-path feature in the consuming crate.
+- Reports malformed syntax, invalid fields, unknown nodes, and unrepresentable values as compile-time errors.
+
+```rust
+use patina::devpath;
+
+const PCI_DEVICE_PATH: [u8; 22] = devpath!("PciRoot(0)/Pci(0x11,0)");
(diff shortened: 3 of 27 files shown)


sdk/patina_macro/README.md
```diff
@@ -96,3 +96,41 @@ fn watchdog_negative_path() -> Result {
...
+- Reports malformed syntax, invalid fields, unknown nodes, and unrepresentable values as compile-time errors.
+
+```rust
+use patina::devpath;
+
+const PCI_DEVICE_PATH: [u8; 22] = devpath!("PciRoot(0)/Pci(0x11,0)");
+const MULTI_INSTANCE_PATH: [u8; 20] = devpath!("Pci(1,0),USB(2,1)");
+```
+
+Without a vendor registry, the input must be exactly one string literal. Separate nodes with `/` or `\`; use a
```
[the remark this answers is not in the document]
joschock: adopted and fixed in latest.

sdk/patina_macro/src/lib.rs
```diff
@@ -9,12 +9,88 @@
...
 mod hob_macro;
 mod service_macro;
 mod smbios_record_macro;
 mod test_macro;
 mod validate_params_macro;
 
+/// Encodes a UEFI text device path as an owned byte array at compile time.
+///
+/// The input is normally one string literal using the UEFI 2.11 Device Path
+/// From Text syntax. Nodes are separated by `/` or `\`. A top-level comma
```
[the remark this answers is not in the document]
joschock: adopted and fixed in latest.

sdk/patina_macro/src/device_path_parser.rs
```diff
@@ -0,0 +1,340 @@
...
+/// Parse a complete UEFI text device path.
+pub(crate) fn parse_device_path(input: &str) -> Result<ParsedDevicePath, DevicePathError> {
+    if input.is_empty() {
+        return Err(DevicePathError::new(0, "device path cannot be empty"));
+    }
+
+    let mut offset = 0;
+    let mut instances = vec![Vec::new()];
+
+    if matches!(byte_at(input, offset), Some(b'/' | b'\\')) {
```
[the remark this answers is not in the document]
joschock: adopted and fixed in latest.

cspell.yml
```diff
@@ -223,6 +223,7 @@ words:
   - unmaps
   - unrecovered
   - unreg
+  - unrepresentable
   - unregisters
```
makubacki: nit: Can you drop this down one line to retain alphabetical order?

makubacki (approved):

makubacki:
As mentioned offline, I think we should keep in mind if consolidation with the runtime (SDK) parsing logic could be consolidated in the future, but I don't think it's required now.

sdk/patina/README.md
```diff
@@ -38,6 +38,9 @@ The SDK is organized into the following root-level modules.
 | **standard** | Strict definitions of industry standards. |
 | **uefi** | UEFI specification definitions and wrappers. |
 
+An optional invocation-local `vendors { ... };` registry can define named vendor hardware nodes with fixed GUIDs and
+typed payload fields. See the `devpath!` API documentation for the schema syntax and supported field encodings.
```
os-d: This seems an odd place for this note

sdk/patina_macro/README.md
```diff
@@ -96,3 +96,41 @@ fn watchdog_negative_path() -> Result {
     Ok(())
 }
 ```
+
+### `devpath!`
+
+- Converts a UEFI 2.11 Device Path From Text string literal into an owned `[u8; N]` at compile time.
```
os-d: Maybe just say UEFI spec defined device path instead of tying it to a version so it doesn't go out of date? If the device path string gets updated in future UEFI spec versions, we'll be updating this.

os-d (approved):

sdk/patina_macro/README.md
makubacki: I was on the fence about this as well. I think you could also say "as of UEFI 2.11" so it's clear when it was last evaluated and that it may still be valid.

sdk/patina/README.md
joschock: Agree. Will just remove it. I think it made more sense before this file was refactored as part of the SDK layout refactoring.

sdk/patina_macro/README.md
joschock: Removed the spec reference and reworded the section a bit to be less clunky.

joschock:
@makubacki @os-d - pushed an update to add VenMsg and VenMedia (previously the `vendor` syntax only supported shortcuts for VenHw).

PR #1654, opened 2026-07-23
[Rebase & FF] Memory Protection Policy: Default System Memory to WB

### LINKED ISSUE #1418: [Feature]: Consider setting cache attributes on freed pages
### Feature Overview

When memory is freed via `free_pages`, Patina sets certain attributes before freeing - preserving cache attributes and setting RP & XP. 

ref: https://github.com/OpenDevicePartnership/patina/blob/4e014bd3d49bc194c4dd94c55797aa0c75988134/patina_dxe_core/src/gcd/spin_locked_gcd.rs#L2652-L2656
https://github.com/OpenDevicePartnership/patina/blob/4e014bd3d49bc194c4dd94c55797aa0c75988134/patina_dxe_core/src/gcd.rs#L228-L238

### Solution Overview

Consider enforcing opinionated cache attributes on freed pages, in order to have a stronger guarantee on the state of the pages' attributes if allocated in the future, and to allow freed regions to coalesce together if cache attributes were changed while in use. This should be reasonable to do because platforms should not impose requirements on the cache attributes of pooled free memory.

### Alternatives Considered

- Make no changes - as I understand it this is closer to how EDK2's core behaves today but has the drawbacks identified above.
- Enforce that the caller should restore attributes before freeing.

### Urgency

Low

Maintainer feedback requested

### Anything else?

_No response_

os-d:
## Description
Currently, the add/allocate/free memory protection policies preserve the cache attributes of the memory. However, the entity consuming this memory may have changed the cache attributes. And added memory may have 0, which is invalid.

In order to maintain a consistent state for system memory, Patina is taking the stance that system memory should be writeback by default. It will be allocated and returned as free as such. This allows a new allocation to have WB and allows free system memory blocks to coalesce correctly.

Other memory types Patina preserves the cache attribute for because Patina doesn't know what the right attribute should be.

This also contains a commit to clean up the includes in the GCD to be less verbose.

Closes #1418 

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Booting Q35/ArmVirt to Windows, booting a physical Intel system to Windows.

## Integration Instructions

N/A.

### DIFF
--- a/docs/src/dxe_core/memory_management.md
+++ b/docs/src/dxe_core/memory_management.md
@@ -298,7 +298,8 @@ auditability and ensuring consistency.
 > **Note:** This section primarily deals with access attributes. Caching attributes are platform and driver driven and
 > outside the scope of this document. The core gets the initial platform specified caching attributes via the Resource
 > Descriptor HOB v2 and persists whatever the GCD entry has on every other call. After this point, drivers (such as
-> the PCI Host Bridge driver) may update memory regions with different caching attributes.
+> the PCI Host Bridge driver) may update memory regions with different caching attributes. The one place Patina
+> is opinionated in caching is that the default caching attribute for system memory is write back.
 
 ### General Flow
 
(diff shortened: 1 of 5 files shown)


cfernald (approved):

makubacki (approved):
I think this is reasonable for System Memory. It feels like this might benefit from a documentation update in the mdbook to call it out.

os-d:
> I think this is reasonable for System Memory. It feels like this might benefit from a documentation update in the mdbook to call it out.

Agreed. Added.

PR #1657, opened 2026-07-27
Add Patina SDK UEFI String Types [Rebase & FF]

### LINKED ISSUE #1656: [Feature]: Add native UEFI string types
### Feature Overview

This feature tracks adding two string families to the `sdk` modeled, to an extent,on `CStr` and `CString`. `Char8Str`/`Char8String` will cover 8-bit Latin-1/ASCII strings which follows the exact format defined as `CHAR8` in the UEFI Specification. `Char16Str`/`Char16String` will cover the fixed 2-byte UCS-2 encoding format (following Unicode 2.1) defined as `CHAR16` in the UEFI Specification.

These will provide an idiomatic and safe type to use for these strings in Patina.

Note: This will be used to support native Rust strings with UEFI Services APIs, but will also be available outside UEFI Services.

### Solution Overview

A `strings` module will be added to the `sdk` to contain the types. There will be an error type (e.g. `StringError`) to represent errors working with the string types.

At a high-level, `Char16` is `#[repr(transparent)]` over `[u16]` and supports conversion between raw pointers. `Char16String` is gated by `alloc` and backed by a `Vec<u16>`. Some implementations like `TryFrom<&str>` are supported and will reject invalid characters.

`Char8Str` & `Char8String` work similarly with `u8`.

Const support will be available through types `#[repr(transparent)]` over arrays.

### Alternatives Considered

_No response_

### Urgency

Low

I will implement the feature

No maintainer feedback needed

### Anything else?

_No response_

makubacki:
## Description

Resolves #1656

The UEFI Specification defines two string types: [`CHAR8` and `CHAR16`](https://uefi.org/specs/UEFI/2.11/02_Overview.html#common-uefi-data-types).

- `CHAR8`: 1-byte character. Unless otherwise specified, all 1-byte or ASCII characters and strings are stored in 8-bit ASCII encoding format, using the ISO-Latin-1 character set.
- `CHAR16`: 2-byte Character. Unless otherwise specified all characters and strings are stored in the UCS-2 encoding format as defined by Unicode 2.1 and ISO/IEC 10646 standards.

`strings.md` (added to the mdbook) and doc comments explain the types in more detail, but at a high-level:

| Role                                       | CHAR8           | CHAR16           |
|--------------------------------------------|-----------------|------------------|
| Borrowed, unsized                          | `Char8Str`      | `Char16Str`      |
| Owned, heap-backed (requires `alloc`)      | `Char8String`   | `Char16String`   |
| Owned, fixed capacity (usable in `const`)  | `Char8Array<N>` | `Char16Array<N>` |

Some code usage is updated to use the new types that might not be exhaustive (future changes can adopt the types as needed). **However, more substantial changes to use the new types throughout the codebase would require changing public APIs** and that was avoided in this change so the types can be introduced to the `main` branch without the PR being a breaking change. This makes existing code updates not as clean as they'd be otherwise. For example, if a public API takes a string today as `Vec<u16>`, this will update the caller to convert the string to a `&Char16Str` to check correctness, but it will ultimately need to be converted to `Vec<u16>` when passed as an argument (instead of directly as `&Char16Str`).

After this PR is complete, it will be added to the `major` branch and the "breaking changes" to more completely use these types will be made there.

---

### Commits

**docs: set mdbook doctest edition to 2021**

mdbook doctests default to Rust edition 2015, since patina is on the
2021 edition, this updates the mdbook to it as well.

In particular, this is done now to allow path-based macro imports such
as (`use patina::{char16, char8};`) and invocations like
`patina::char16!(...)` to resolve.

---

**sdk: Add UEFI string types**

Adds two string families to the `sdk` modeled, to an extent, on
`CStr` and `CString`.

`Char8Str`/`Char8String` covers 8-bit Latin-1/ASCII strings which
follows the exact format defined as `CHAR8` in the UEFI
Specification. `Char16Str`/`Char16String` covers the fixed 2-byte
UCS-2 encoding format (following Unicode 2.1) defined as `CHAR16`
in the UEFI Specification.

These provide idiomatic and safe types to use for these strings
in Patina.

---

**docs: Add UEFI string docs**

Adds a "UEFI Strings" page to explain how to use the string types
recently added to the SDK.

---

**components: Use Patina SDK UEFI string types**

Updates a couple of components that have interfaces with raw string
pointers to use the `Char8Str` types for processing.

---

**sdk: Use Patina SDK UEFI string types**

Updates string handling to use the new Patina SDK string types. This
commit intentionally avoids larger refactors that impact public APIs.

Note: Device path buffers with surrogate code units now return a parse
error rather than being mapped to `U+FFFD`.

---

**patina_dxe_core: Use Patina SDK UEFI string types**

Updates string handling to use the new Patina SDK string types. This
commit intentionally avoids larger refactors that impact public APIs.

---

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [x] Includes documentation?

## How This Was Tested

- `cargo make all`
- Boot to EFI shell on ArmVirt and Q35

## Integration Instructions

- Use these new string types when appropriate per the instructions in `strings.md`.

### DIFF
--- a/sdk/patina/src/base/string/char16.rs
+++ b/sdk/patina/src/base/string/char16.rs
@@ -11,8 +11,13 @@
 use super::{StringError, char_count, decode_utf8};
 use zerocopy_derive::{FromBytes, Immutable, IntoBytes, KnownLayout};
 
+use super::char8::Char8Array;
+
 #[cfg(any(test, feature = "alloc"))]
-use alloc::vec::Vec;
+use alloc::{string::String, vec::Vec};
+
+#[cfg(any(test, feature = "alloc"))]
+use super::char8::{Char8Str, Char8String};
 
 /// The inclusive range of UTF-16 surrogate code units, which are not valid UCS-2.
 const SURROGATE_RANGE: core::ops::RangeInclusive<u16> = 0xD800..=0xDFFF;
@@ -463,6 +468,36 @@ impl<'a> TryFrom<&'a str> for Char16String {
 }
 
 #[cfg(any(test, feature = "alloc"))]
+impl From<&Char8Str> for Char16String {
+    /// Converts Latin-1 to UCS-2.
+    ///
+    /// Always succeeds since every Latin-1 byte maps directly onto an identically-valued, non-surrogate
+    /// UCS-2 code unit.
+    fn from(value: &Char8Str) -> Self {
+        let mut units = Vec::with_capacity(value.len() + 1);
+        for &byte in value.iter() {
+            units.push(byte as u16);
+        }
+        units.push(0);
+        Self(units)
+    }
+}
+
+#[cfg(any(test, feature = "alloc"))]
+impl From<&Char8String> for Char16String {
+    fn from(value: &Char8String) -> Self {
+        Self::from(value.as_char8_str())
+    }
+}
+
+#[cfg(any(test, feature = "alloc"))]
+impl From<Char8String> for Char16String {
+    fn from(value: Char8String) -> Self {
+        Self::from(&value)
+    }
+}
+
+#[cfg(any(test, feature = "alloc"))]
 impl core::fmt::Display for Char16String {
     fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
         core::fmt::Display::fmt(self.as_char16_str(), f)
@@ -666,6 +701,57 @@ impl<const N: usize> Default for Char16Array<N> {
     /// [`Char16Array::try_from_str`], a capacity of `0` cannot represent a valid string.
     fn default() -> Self {
         Self([0; N])
+    }
+}
+
+impl<'a, const N: usize> TryFrom<&'a str> for Char16Array<N> {
+    type Error = StringError;
+
+    /// Equivalent to [`Char16Array::try_from_str`], provided for generic code written against the
+    /// standard [`TryFrom`] trait.
+    fn try_from(s: &'a str) -> Result<Self, StringError> {
+        Self::try_from_str(s)
+    }
+}
+
+impl<const N: usize> From<Char8Array<N>> for Char16Array<N> {
+    /// Converts Latin-1 to UCS-2.
+    ///
+    /// Always succeeds since every Latin-1 byte maps directly onto an identically-valued, non-surrogate
+    /// UCS-2 code unit, and a `Char8Array<N>`'s logical content (at most `N - 1` characters) always
+    /// fits within `Char16Array<N>`'s equal capacity.
+    // Note: `out[position]` is safe because `position < src.len() < N`, except when `N == 0`, where `src`
+    // is empty and the loop never runs.
+    #[allow(clippy::indexing_slicing)]
+    fn from(value: Char8Array<N>) -> Self {
+        let src = value.as_char8_str();
+        // Note: The terminator and any trailing padding are already present due to zero-initialization.
+        let mut out = [0u16; N];
+        for (position, &byte) in src.iter().enumerate() {
+            out[position] = byte as u16;
+        }
+        Self(out)
+    }
+}
+
+#[cfg(any(test, feature = "alloc"))]
+impl<const N: usize> From<Char16Array<N>> for Char16String {
+    /// Converts a fixed-capacity array to a heap-owned string, dropping the const capacity `N`.
+    fn from(value: Char16Array<N>) -> Self {
+        use alloc::borrow::ToOwned;
+        value.as_char16_str().to_owned()
+    }
+}
+
+#[cfg(any(test, feature = "alloc"))]
+impl<const N: usize> From<Char16Array<N>> for String {
+    /// Converts to a native Rust string.
+    ///
+    /// Always succeeds. Lossless for arrays built through a safe constructor; a surrogate code unit
+    /// reachable only via an unchecked constructor is replaced with [`char::REPLACEMENT_CHARACTER`]
+    /// rather than causing a panic.
+    fn from(value: Char16Array<N>) -> Self {
+        value.as_char16_str().chars().collect()
     }
 }
 
@@ -1086,5 +1172,62 @@ mod tests {
     fn test_char16_capacity_helper_runtime() {
         assert_eq!(ucs2_capacity("EFI"), 4);
         assert_eq!(ucs2_capacity(""), 1);
+    }
+
+    #[test]
+    fn test_char16_array_try_from_str_trait() {
+        let array = Char16Array::<9>::try_from("Firmware").unwrap();
+        assert!(array.as_char16_str() == "Firmware");
+    }
+
+    #[test]
+    fn test_char16_array_from_char8_array() {
+        let narrow = Char8Array::<9>::from_str("Firmware");
+        let wide: Char16Array<9> = narrow.into();
+        assert!(wide.as_char16_str() == "Firmware");
+    }
+
+    #[test]
+    fn test_char16_array_from_char8_array_high_latin1() {
+        let narrow = Char8Array::<2>::from_str("\u{00E9}"); // 'é'
+        let wide: Char16Array<2> = narrow.into();
+        assert!(wide.as_char16_str() == "\u{00E9}");
+    }
+
+    #[test]
+    fn test_char16_array_from_char8_array_empty_capacity() {
+        let narrow = Char8Array::<0>::default();
+        let wide: Char16Array<0> = narrow.into();
+        assert!(wide.as_char16_str().is_empty());
+    }
+
+    #[test]
+    fn test_char16_array_to_char16string() {
+        let array = Char16Array::<9>::from_str("Firmware");
+        let owned: Char16String = array.into();
+        assert!(owned == "Firmware");
+    }
+
+    #[test]
+    fn test_char16_array_to_string() {
+        let array = Char16Array::<9>::from_str("Firmware");
+        let owned: alloc::string::String = array.into();
+        assert_eq!(owned, "Firmware");
+    }
+
+    #[test]
+    fn test_char16string_from_char8str() {
+        let narrow = Char8Array::<4>::from_str("EFI");
+        let wide = Char16String::from(narrow.as_char8_str());
+        assert!(wide == "EFI");
+    }
+
+    #[test]
+    fn test_char16string_from_char8string() {
+        let narrow = Char8String::try_from_str("EFI").unwrap();
+        let wide = Char16String::from(&narrow);
+        assert!(wide == "EFI");
+        let wide_owned = Char16String::from(narrow);
+        assert!(wide_owned == "EFI");
     }
 }
--- a/sdk/patina/src/base/string/char16.rs
+++ b/sdk/patina/src/base/string/char16.rs
@@ -1,0 +1,1090 @@
+//! CHAR16 (UCS-2) string types
+//!
+//! See the [module documentation](crate::base::string) for an overview of the UEFI string types.
+//!
+//! ## License
+//!
+//! Copyright (C) Microsoft Corporation.
+//!
+//! SPDX-License-Identifier: Apache-2.0
+
+use super::{StringError, char_count, decode_utf8};
+use zerocopy_derive::{FromBytes, Immutable, IntoBytes, KnownLayout};
+
+#[cfg(any(test, feature = "alloc"))]
+use alloc::vec::Vec;
+
+/// The inclusive range of UTF-16 surrogate code units, which are not valid UCS-2.
+const SURROGATE_RANGE: core::ops::RangeInclusive<u16> = 0xD800..=0xDFFF;
+
+/// A borrowed, NUL-terminated UCS-2 (UEFI CHAR16) string.
+///
+/// This is the CHAR16 analogue of [`core::ffi::CStr`]. It is an unsized type that is always accessed
+/// behind a reference (`&Char16Str`). The referenced data is guaranteed to:
+///
+/// - End with a single NUL (`0x0000`) terminator,
+/// - Contain no interior NUL, and
+/// - Contain only valid UCS-2 code units (no surrogates in `0xD800..=0xDFFF`).
+///
+/// Because the terminator is guaranteed, [`Char16Str::as_ptr`] can be passed directly to
+/// `extern "efiapi"` interfaces expecting a `*const CHAR16`.
+///
+/// # Examples
+///
+/// ```rust
+/// use patina::base::string::Char16Str;
+///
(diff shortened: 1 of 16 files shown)


sdk/patina/src/base/strings/array.rs
```diff
@@ -0,0 +1,577 @@
...
+        let mut index = 0;
+        let mut out_index = 0;
+        let mut position = 0;
+
+        while index < bytes.len() {
+            let (value, len) = decode_utf8(bytes, index);
+            if value == 0 {
+                return Err(StringError::InteriorNul { position });
+            }
+            if value > 0xFFFF {
```
joschock: pedantry: technically, I think UCS-2 rejects 0xD800-0xDFFF (https://en.wikipedia.org/wiki/UTF-16#U+D800_to_U+DFFF_(surrogates)). I am not sure we care.

joschock: I guess maybe `decode_utf8` would never generate these? I haven't done the math.

sdk/patina/src/base/string/char16.rs
```diff
@@ -0,0 +1,798 @@
...
+        let body = match units.split_last() {
+            Some((0, body)) => body,
+            _ => return Err(StringError::MissingNulTerminator),
+        };
+
+        for (position, &unit) in body.iter().enumerate() {
+            if unit == 0 {
+                return Err(StringError::InteriorNul { position });
+            }
+            if SURROGATE_RANGE.contains(&unit) {
```
joschock, line 72: Ah - I see we are checking surrogates here during runtime conversion; good. I sure hope we don't have a lot of sloppy code out there that's working by accident.

I wonder if we have an issue with e.g. "Someone created a UEFI var in the varstore with surrogate chars in the name prior to updating to this UEFI and now it crashes". Does the existing EDK2 var stack enforce surrogates?

joschock (approved):

sdk/patina/src/base/string/char16.rs
```diff
@@ -0,0 +1,798 @@
...
+            if value == 0 {
+                return Err(StringError::InteriorNul { position });
+            }
+            if value > 0xFFFF {
+                return Err(StringError::NotUcs2 { position, value });
+            }
+            units.push(value as u16);
+        }
+        units.push(0);
+        Ok(Self(units))
```
vineelko, line 346: Should we shrink the `units` vector after populating here? Because `s.len()` is always in bytes(u8) but `with_capacity()` is always in elements(u16).

vineelko (approved):

sdk/patina/src/base/strings/char16.rs
```diff
@@ -0,0 +1,798 @@
...
+    pub fn from_units_until_nul(units: &[u16]) -> Result<&Char16Str, StringError> {
+        let end = units.iter().position(|&unit| unit == 0).ok_or(StringError::MissingNulTerminator)?;
+        Self::from_units_with_nul(&units[..=end])
+    }
+
+    /// Creates a `&Char16Str` by scanning a raw pointer for a NUL terminator.
+    ///
+    /// # Safety
+    ///
+    /// The caller must ensure that `ptr` is non-null and points to a sequence of `u16` values that is
```
Javagedes: While it is vital that the pointer points to data that eventually ends with a `\0`, I don't think we are correctly characterizing the safety concerns of this function.

1. `ptr.add` has safety requirements that `[u16] * len` does not overflow `isize::MAX` (which I think in this usage we can and should check for and don't need to bubble up)

2. That the pointer owns all of the data between pointer and length. Maybe we can hand wave that correctly ending in a `\0` resolves this.

This would also be true for the char8 - regarding needing to check for `[u8] * len` does not overflow `isize::MAX`.

sdk/patina/src/base/string/char16.rs
```diff
@@ -0,0 +1,798 @@
...
+        }
+        // SAFETY: `len + 1 <= max_units` code units up to and including the terminator are valid.
+        let units = unsafe { core::slice::from_raw_parts(ptr, len + 1) };
+        Self::from_units_with_nul(units)
+    }
+
+    /// Returns a pointer to the first code unit, suitable for passing to `extern "efiapi"` interfaces.
+    ///
+    /// The pointed-to sequence is always NUL-terminated.
+    pub const fn as_ptr(&self) -> *const u16 {
```
Javagedes, line 176: would it be better to return these as `NonNull<u16>` and `NonNull<u8>`?

sdk/patina/src/base/string/char16.rs
```diff
@@ -0,0 +1,798 @@
...
+        self.0.as_ptr()
+    }
+
+    /// Returns the code units including the trailing NUL terminator.
+    pub const fn as_units_with_nul(&self) -> &[u16] {
+        &self.0
+    }
+
+    /// Returns the code units excluding the trailing NUL terminator.
+    pub fn as_units(&self) -> &[u16] {
```
Javagedes, line 186: why not name this `as_slice`?

sdk/patina/src/base/strings/char16.rs
```diff
@@ -0,0 +1,798 @@
...
+        self.as_units().len()
+    }
+
+    /// Returns `true` if the string contains no characters (only the NUL terminator).
+    pub fn is_empty(&self) -> bool {
+        self.len() == 0
+    }
+
+    /// Returns an iterator over the raw `u16` code units, excluding the terminator.
+    pub fn code_units(&self) -> impl Iterator<Item = u16> + '_ {
```
Javagedes: `as_iter()` is typically the normal way to do this.

I also don't like forcing a copy here. We should let the caller decide if they want to copy the bytes or not in my opinion.

sdk/patina/src/base/string/char16.rs
```diff
@@ -0,0 +1,798 @@
...
+}
+
+#[cfg(any(test, feature = "alloc"))]
+impl alloc::borrow::ToOwned for Char16Str {
+    type Owned = Char16String;
+
+    fn to_owned(&self) -> Char16String {
+        Char16String(self.0.to_vec())
+    }
+}
```
Javagedes, line 451: I think a From<Char16Str> would be beneficial.

sdk/patina/src/base/string/char16.rs
```diff
@@ -0,0 +1,798 @@
...
+
+#[cfg(any(test, feature = "alloc"))]
+impl core::fmt::Debug for Char16String {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        core::fmt::Debug::fmt(self.as_char16_str(), f)
+    }
+}
+
+#[cfg(any(test, feature = "alloc"))]
+impl PartialEq<str> for Char16String {
```
Javagedes, line 515: and comparisons between Char16String and Char16Str.

sdk/patina/src/base/strings/array.rs
```diff
@@ -0,0 +1,577 @@
...
+/// ```rust
+/// use patina::base::strings::Char16Array;
+///
+/// const NAME: Char16Array<4> = Char16Array::from_str("EFI");
+/// assert_eq!(NAME.as_char16_str().len(), 3);
+/// assert!(NAME.as_char16_str() == "EFI");
+/// ```
+#[repr(transparent)]
+#[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash, FromBytes, IntoBytes, Immutable, KnownLayout)]
+pub struct Char16Array<const N: usize>([u16; N]);
```
Javagedes: I don't really think I agree with putting this in an `array.rs` file. In myopinion it makes more sense going in the same file as the `char16.rs` code (and `Char8Array` in `char8.rs.`

sdk/patina/src/base/strings/array.rs
```diff
@@ -0,0 +1,577 @@
...
+    }
+}
+
+impl<const N: usize> core::fmt::Display for Char16Array<N> {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        core::fmt::Display::fmt(self.as_char16_str(), f)
+    }
+}
+
+impl<const N: usize> Default for Char16Array<N> {
```
Javagedes: I think we need more conversions. Specifically converting this array into a string. My a tryInto for converting a string type into an array if the lengths match up. things like that for convenience.

Javagedes (approved):
I made a lot of comments specific to Char16 that apply to Char8 as well if you decide to implement any of them. otherwise looks fine though. I think interactions with `Cow<TheseTypes>` would be interesting too.

sdk/patina/src/base/strings/char16.rs
joschock: The `*const u16` also needs to be properly aligned. Consider adding a safety note to that effect.

sdk/patina/src/base/strings/array.rs
makubacki: This particular function (`patina::base::strings::array::Char16Array::try_from_str`) takes a [str](https://doc.rust-lang.org/std/primitive.str.html#invariant) which the docs say "Rust libraries may assume that string slices are always valid UTF-8". This is repeated in places like [`str::from_utf8`](https://doc.rust-lang.org/std/primitive.str.html#method.from_utf8).

The [`char`](https://doc.rust-lang.org/std/primitive.char.html#validity-and-layout) documentation also requires are Unicode scalar values (the "exact same set of values that may be encoded in UTF-8") and calls out `0xD800` specifically as rejected.

So, given the `str` type, and these defined constraints, I didn't think it was necessary here.

sdk/patina/src/base/string/char16.rs
makubacki: I'm not aware of code enforcement in places the like the edk2 variable stack on `CHAR16*` names. It's just treated as an opaque array of code units there. However, it's very unlikely in practice and already a violation of the `CHAR16` definition in the UEFI Specification, so I think it's justified to strictly enforce string expectations here.

sdk/patina/src/base/string/char16.rs
makubacki: This is an optimization decision I'm hesitant to make without data. `shrink_to_fit()` will use [`Allocator::shrink`](https://doc.rust-lang.org/std/alloc/trait.Allocator.html#method.shrink) and that may introduce copies and new allocations, which have their own impact. Since vectors are expected to have a larger capacity than may be needed and the common use case of ASCII characters will have the same number of elements, I think there's a few more cons than pros to do this without a substantial motivator to do so.

sdk/patina/src/base/strings/char16.rs
makubacki: [`CStr::from_ptr`](https://doc.rust-lang.org/std/ffi/struct.CStr.html#method.from_ptr) runs into the same situation and does bubble the `ptr.add` requirements up to its requirements.

Looking into it briefly, it seems to have to do with the fact that scanning into the unknown (finding the terminator) would lead to the UB anyway. In any case, I included the safety detail they have there in the context of this code. I also called out the pointer alignment requirement.

sdk/patina/src/base/string/char16.rs
makubacki: It's an analogue of [`CStr::as_ptr`](https://doc.rust-lang.org/std/ffi/struct.CStr.html#method.as_ptr). I see where you're coming from and I considered that too, but I don't want to stray too far aware from established conventions.

I suspected it has something to do with ergonomics like C FFIs using the `*const` form and not needing to apply `nonnull.as_ptr()` or rely on coercion.

sdk/patina/src/base/string/char16.rs
makubacki: It is meant to be a more descriptive name for the types of elements in the CHAR16 slice (code units) and consistent with terminology used elsewhere. In char8.rs, it's `as_bytes`, as those are fixed-width bytes, here the equivalent is `as_units`.

sdk/patina/src/base/strings/char16.rs
makubacki: I agree on both points. I renamed the function to `iter()`. I also dropped the `.copied` call and updated the return type with guidance that callers who want owned values can call `.copied` themselves. I also added an equivalent function to char8.rs.

sdk/patina/src/base/string/char16.rs
makubacki: Agree and added.

sdk/patina/src/base/string/char16.rs
makubacki: Agree and added.

sdk/patina/src/base/strings/array.rs
makubacki: I think that's reasonable and will start making the change and see if anything unexpected comes up. A few common helpers will likely need to be moved to strings.rs, but not a problem.

makubacki:
I'm going to rename `strings.rs` to `string.rs`. That should align with `sdk` naming conventions.

sdk/patina/src/base/string/char16.rs
joschock: Agree with enforcing it here; just trying to think through how it might work out in the face of malicious creation of such a variable name, and whether we need to adjust elsewhere to defend. For example, if UefiServices uses this string handling to implement "GetNextVariableName" wrappers - would the existence of a bogus name in the varstore terminate the search early? I think it is okay to deny access to a maliciously named variable, but I'm worried about that chaining into denying access to legit variables somehow or otherwise breaking the var stack. I'm not sure I want something different at this line of code though, so I'm good with resolving this comment here.

sdk/patina/src/base/strings/array.rs
makubacki: Done

sdk/patina/src/base/strings/array.rs
makubacki: Since `Char8Array` and `Char16Array` already implement `Display`, `.to_string()` works. I suppose adding a `From<Char8Array>`/`<Char16Array>` for `alloc::String` helps a little with ergonomics.

I went ahead and added what I think are the most useful additional conversions in commit https://github.com/OpenDevicePartnership/patina/pull/1657/changes/08b84d051884d9d93392d676c4e5d74f63ed0778.

@Javagedes, can you check and resolve this conversation if it meets expectations?

sdk/patina/src/base/string/char16.rs
makubacki: @joschock, I'd classify this as a broader concern that may need code change or guidance. I'll file a GitHub issue to track it.

makubacki: Issue: https://github.com/OpenDevicePartnership/patina/issues/1665

PR #1658, opened 2026-07-28
Remove `doc` feature in favor of `doc` cfg

### LINKED ISSUE #1535: [Task]: Remove doc Feature From Crates

### Current Behavior

The doc feature existed in a previous version of running doctests and is believed not to be needed any longer.

### Expected Behavior

Evaluate if the feature is needed and remove from all crates if not.

### Steps To Reproduce

n/a

### Build Environment

```markdown
- OS(s):
- Targets Impacted:
```

### Version Information

```text
top of tree
```

### Urgency

Low

Someone else needs to fix it

No maintainer feedback needed

### Anything else?

_No response_

Javagedes:
## Description

Sync'd files are also updated via this PR: https://github.com/OpenDevicePartnership/patina-devops/pull/183

ref: [doc cfg functionality](https://doc.rust-lang.org/rustdoc/advanced-features.html#cfgdoc-documenting-platform-specific-or-feature-specific-information)

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

N/A

## Integration Instructions

N/A

### DIFF
--- a/sdk/patina/Cargo.toml
+++ b/sdk/patina/Cargo.toml
@@ -67,7 +67,6 @@ trybuild = { version = "1.0", features = ["diff"] }
 [features]
 core = ['alloc']
 std = ['alloc']
-doc = ['alloc']
 alloc = ["dep:goblin", "dep:fixedbitset"]
 mockall = ["dep:mockall", "std"]
 global_allocator = []
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -336,13 +336,13 @@ dependencies = ["individual-package-targets"]
 [tasks.doc]
 description = "Builds all rust documentation in the workspace. Example `cargo make doc`"
 command = "cargo"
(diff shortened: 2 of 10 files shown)


vineelko (approved):

makubacki (approved):

sdk/patina/Cargo.toml
makubacki, line 1: This is technically a breaking change if consumers are referencing the feature, right?

Since `doc` cfg is only set by rustdoc, it also doesn't exactly replace a feature that can be controlled in other invocations (e.g. build, check, test, etc.). I think that's fine and the doc link added helps explain when it's set.

Javagedes: I agree that it is technically a breaking change, however that feature flag's only purpose was to allow both `aarch64` and `x64` specific functions to show up in docs when built. It should not have been a feature flag consumed by any other crate.

I've verified this to be true in all patina repos.

Issue #1610, opened 2026-07-02
[Bug]: Handle SMRR and code access programming


kuqin12:
### Is there an existing issue for this?

- [x] I have searched existing issues

### Current Behavior

The current code does not handle SMRR like c implementation, where it will probe the SMRR on the first entry and potentially lock SMRR on the exit.

In addition, the code access MSR is not programmed and locked down.

### Expected Behavior

Do the needful per SMRR security standard.

Also lock down the SMM code access MSR.

### Steps To Reproduce

It is just not implemented.

### Build Environment

```markdown
- OS(s): Windows 11
- Targets Impacted: DEBUG/RELEASE
```

### Version Information

```text
Top of feature branch.
```

### Urgency

High

### Are you going to fix this?

Someone else needs to fix it

### Do you need maintainer feedback?

No maintainer feedback needed

PR #1659, opened 2026-07-28
[feature/supv] SMM: Configure SMRR and SMM Code Access

vineelko:
## Description

- This commit programs the SMRRs in two phases
    1. Initialize the SMRR base and mask registers on the first SMI entry
    2. Enable the Valid bit (if not already) on every subsequent SMI entry
- Also programs the SMM Code Access for every CPU on the first SMI entry
   
    Address https://github.com/OpenDevicePartnership/patina/issues/1610

---
- [x] Impacts functionality?
- [x] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Validated boot to desktop on physical platform.

## Integration Instructions

NA

### DIFF
--- a/patina_mm_supervisor/src/lib.rs
+++ b/patina_mm_supervisor/src/lib.rs
@@ -94,7 +94,7 @@ pub use cpu::CpuInfo;
 pub use init::PolicyInitError;
 pub use supervisor_handlers::SupervisorMmiHandler;
 
-use crate::smrr::{configure_smm_code_access, smrr_enable, smrr_initialize};
+use crate::smrr::smrr_enable;
 
 // The entry-point shim references `rust_main`, which is provided by the platform binary, and is
 // only meaningful on the firmware (UEFI) target. Exclude it from host builds (tests, doctests)
@@ -541,14 +541,6 @@ impl<P: PlatformInfo, const MAX_CPUS: usize> MmSupervisorCore<P, MAX_CPUS> {
         // Track that this core has completed per-core init
         let init_count = init_state().inc_per_core_init_count();
         log::trace!("CPU {} (index {}) completed per-core init ({} cores initialized)", cpu_id, cpu_index, init_count);
-
-        // If smrr region in the initstate is set, use that otherwise don't
-        // program smrr
-        if let Some((base, size)) = init_state().smrr_base_size() {
-            smrr_initialize(base, size);
-        }
-
-        configure_smm_code_access();
 
         // BSP waits for all registered CPUs to complete per-core init before returning
         if is_bsp {
--- a/patina_mm_supervisor/src/lib.rs
+++ b/patina_mm_supervisor/src/lib.rs
@@ -57,6 +57,7 @@ mod perf_timer;
 mod privilege_mgmt;
 mod runtime;
 mod save_state;
+mod smrr;
 mod state;
 mod supervisor_handlers;
 
@@ -92,6 +93,8 @@ pub use cpu::CpuInfo;
 // their function signatures and return types.
 pub use init::PolicyInitError;
 pub use supervisor_handlers::SupervisorMmiHandler;
+
+use crate::smrr::{configure_smm_code_access, smrr_enable, smrr_initialize};
 
 // The entry-point shim references `rust_main`, which is provided by the platform binary, and is
 // only meaningful on the firmware (UEFI) target. Exclude it from host builds (tests, doctests)
@@ -455,6 +458,7 @@ impl<P: PlatformInfo, const MAX_CPUS: usize> MmSupervisorCore<P, MAX_CPUS> {
                 cpu_id,
                 cpu_index
             );
+            smrr_enable();
             self.enter_runtime(cpu_id);
 
             return;
@@ -537,6 +541,14 @@ impl<P: PlatformInfo, const MAX_CPUS: usize> MmSupervisorCore<P, MAX_CPUS> {
         // Track that this core has completed per-core init
         let init_count = init_state().inc_per_core_init_count();
         log::trace!("CPU {} (index {}) completed per-core init ({} cores initialized)", cpu_id, cpu_index, init_count);
+
+        // If smrr region in the initstate is set, use that otherwise don't
+        // program smrr
+        if let Some((base, size)) = init_state().smrr_base_size() {
+            smrr_initialize(base, size);
+        }
+
+        configure_smm_code_access();
 
         // BSP waits for all registered CPUs to complete per-core init before returning
         if is_bsp {
--- a/patina_mm_supervisor/src/mem/page_allocator.rs
+++ b/patina_mm_supervisor/src/mem/page_allocator.rs
@@ -532,9 +532,7 @@ pub(crate) fn coalesced_smrr_range(regions: &[SmramRegion]) -> Option<SmramRegio
         }
     }
 
-    let (base, size) = current?;
-    let mut smrr_base = base as u32;
-    let mut smrr_size = size as u32;
+    let (mut smrr_base, mut smrr_size) = current?;
 
     // Coalesce any physically adjacent ranges into the selected range. This
     // scans the (unsorted) region array repeatedly until no further
@@ -546,14 +544,14 @@ pub(crate) fn coalesced_smrr_range(regions: &[SmramRegion]) -> Option<SmramRegio
         for region in regions {
             let region_base = region.base;
             let region_size = region.size;
-            if region_base < smrr_base as u64 && smrr_base as u64 == region_base + region_size {
+            if region_base < smrr_base && smrr_base == region_base + region_size {
                 // Region sits immediately before the current range: extend downward.
-                smrr_base = region_base as u32;
-                smrr_size = smrr_size.wrapping_add(region_size as u32);
+                smrr_base = region_base;
+                smrr_size = smrr_size.checked_add(region_size)?;
                 found = true;
-            } else if smrr_base as u64 + smrr_size as u64 == region_base && region_size > 0 {
+            } else if smrr_base + smrr_size == region_base && region_size > 0 {
                 // Region sits immediately after the current range: extend upward.
-                smrr_size = smrr_size.wrapping_add(region_size as u32);
+                smrr_size = smrr_size.checked_add(region_size)?;
                 found = true;
             }
         }
@@ -561,6 +559,9 @@ pub(crate) fn coalesced_smrr_range(regions: &[SmramRegion]) -> Option<SmramRegio
             break;
         }
     }
+
+    let smrr_base = u32::try_from(smrr_base).ok()?;
+    let smrr_size = u32::try_from(smrr_size).ok()?;
 
     if !verify_smrr_base_size(smrr_base, smrr_size) {
         log::warn!(
--- a/patina_mm_supervisor/src/mem/page_allocator.rs
+++ b/patina_mm_supervisor/src/mem/page_allocator.rs
@@ -49,6 +49,8 @@ use patina_paging::{MemoryAttributes, PageTable};
 use r_efi::efi;
 use spin::{Mutex, MutexGuard, relax::Spin};
 
+use crate::smrr::{SmrrRange, verify_smrr_base_size};
+
 /// Bits per byte.
 const BITS_PER_BYTE: usize = 8;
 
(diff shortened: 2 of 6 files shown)


kuqin12 (approved):

makubacki:
@vineelko, I plan to review this tomorrow. Are the unit test coverage numbers intentionally low?

vineelko:
Right now for the smrr module, much of the functionality has been writing/reading to msr. So could not increase more coverage, Will take a stab at adding `#[cfg_attr(coverage, coverage(off))]` for other functions.

makubacki:
> Right now for the smrr module, much of the functionality has been writing/reading to msr. So could not increase more coverage, Will take a stab at adding `#[cfg_attr(coverage, coverage(off))]` for other functions.

Yeah, disabling coverage is fine for those cases.

patina_mm_supervisor/src/lib.rs
```diff
@@ -538,6 +542,14 @@ impl<P: PlatformInfo, const MAX_CPUS: usize> MmSupervisorCore<P, MAX_CPUS> {
...
 
+        // If smrr region in the initstate is set, use that otherwise don't
+        // program smrr
+        if let Some((base, size)) = init_state().smrr_base_size() {
+            smrr_initialize(base, size);
+        }
+
+        configure_smm_code_access();
+
         // BSP waits for all registered CPUs to complete per-core init before returning
```
makubacki, line 545: My understanding is that the BSP is going to spin on `per_core_init_count()` to determine when all cores haves finished "init" (BSPs + APs).

If this indicates when a core has completed its init:

```rust
        // Track that this core has completed per-core init
        let init_count = init_state().inc_per_core_init_count();
```

It seems out of order for `smrr_initialize()` & `configure_smm_code_access()` to be called after.

Is this more what was intended?

```suggestion
// All cores perform per-core initialization
self.per_core_init(cpu_id, is_bsp);

// Mark this core as initialized in the per-core buffer
mark_core_initialized(cpu_index);

        // If smrr region in the initstate is set, use that otherwise don't
        // program smrr
        if let Some((base, size)) = init_state().smrr_base_size() {
            smrr_initialize(base, size);
        }

        configure_smm_code_access();

        // Track that this core has completed per-core init
        let init_count = init_state().inc_per_core_init_count();
        log::trace!("CPU {} (index {}) completed per-core init ({} cores initialized)", cpu_id, cpu_index, init_count);

// BSP waits for all registered CPUs to complete per-core init before returning
```

---

The GitHub preview doesn't look right, but it's saying to move the call to `inc_per_core_init_count()` to after `configure_smm_code_access()`.

patina_mm_supervisor/src/mem/page_allocator.rs
```diff
@@ -545,6 +545,71 @@ impl ScannedRegions {
...
+                    found = true;
+                }
+            }
+            if !found {
+                break;
+            }
+        }
+
+        log::info!("SMRR Base: 0x{:x}, SMRR Size: 0x{:x}", smrr_base, smrr_size);
+        Some((smrr_base, smrr_size))
```
makubacki: I think you should call `verify_smrr_base_size()` before returning `Some((smrr_base, smrr_size))` here and return `None` if it fails to verify the coalesced sum is still a power of two. Afaict, that's not guaranteed here and will result in a panic later.

patina_mm_supervisor/src/lib.rs
```diff
@@ -538,6 +542,14 @@ impl<P: PlatformInfo, const MAX_CPUS: usize> MmSupervisorCore<P, MAX_CPUS> {
         let init_count = init_state().inc_per_core_init_count();
         log::trace!("CPU {} (index {}) completed per-core init ({} cores initialized)", cpu_id, cpu_index, init_count);
 
+        // If smrr region in the initstate is set, use that otherwise don't
+        // program smrr
+        if let Some((base, size)) = init_state().smrr_base_size() {
+            smrr_initialize(base, size);
+        }
```
makubacki: Would code would have panicked already if `smrr_base_size()` would return `None` here?

...in this code in `init_page_allocators()`:

```rust
        match smram_scanned_regions.coalesced_smrr_range() {
            Some((smrr_base, smrr_size)) => {
                log::info!("Discovered SMRR range: base=0x{:08x}, size=0x{:08x}", smrr_base, smrr_size);
                init_state().set_smrr_base_size(smrr_base, smrr_size);
            }
            None => panic!("Failed to determine SMRR range from scanned SMRAM regions"),
        }
```

In which case, it seems this code should expect SMRR regions are available at this point (not optional) or that would indicate a fatal pre-condition is broken.

patina_mm_supervisor/src/smrr.rs
```diff
@@ -0,0 +1,339 @@
...
+        panic!("Unsupported CPU: MTRR not supported");
+    }
+
+    if !is_smrr_supported() {
+        panic!("Unsupported CPU: SMRR not supported");
+    }
+
+    if !is_smrr_ext_supported() {
+        panic!("Unsupported CPU: SMRR extended capability not supported");
+    }
```
makubacki: Given this was already called in `smrr_initialize()` which must be called on the first SMI, is it really necessary to call them again on every SMI?

patina_mm_supervisor/src/smrr.rs
```diff
@@ -0,0 +1,339 @@
...
+
+/// Sets the memory type field ([7:0]) of a raw SMRR base register value.
+fn base_reg_set_memtype(raw: u64, memtype: u8) -> u64 {
+    (raw & !MEMTYPE_MASK) | (memtype as u64 & MEMTYPE_MASK)
+}
+
+/// Sets the physical base address field ([31:12]) of a raw SMRR base register
+/// value.
+fn base_reg_set_base(raw: u64, base: u32) -> u64 {
+    // Stores `base >> 12` into [31:12], i.e. the address bits [31:12].
```
makubacki: The comment about shifting is confusing since the code is masking bits 12-31 in place.

patina_mm_supervisor/src/smrr.rs
```diff
@@ -0,0 +1,339 @@
...
+    (raw & !PHYS_ADDR_MASK) | (base as u64 & PHYS_ADDR_MASK)
+}
+
+/// Sets the mask field ([31:12]) of a raw SMRR mask register value for the
+/// given region `size`.
+fn mask_reg_set_mask(raw: u64, size: u32) -> u64 {
+    // Mask field = ~(size - 1) >> 12, placed back into [31:12].
```
makubacki: Same here.

patina_mm_supervisor/src/smrr.rs
```diff
@@ -0,0 +1,339 @@
...
+const MTRR_CAP_SMRR_BIT: u64 = 1 << 11;
+const MTRR_CAP_SMRR_EXT_BIT: u64 = 1 << 14;
+
+const MSR_SMRR_BASE: u32 = 0x1F2;
+const MSR_SMRR_MASK: u32 = 0x1F3;
+
+const _MTRR_CACHE_WRITE_PROTECTED: u8 = 5;
+const MTRR_CACHE_WRITE_BACK: u8 = 6;
+
+const CPUID_VERSION_INFO: u32 = 0x1;
```
makubacki: This is also defined privately in cpu.rs. Could just have a single `pub(crate)` definition.

patina_mm_supervisor/src/state.rs
```diff
@@ -59,6 +59,8 @@ pub(crate) struct InitState {
     ap_startup_fn: Once<fn(u64, u64, u64) -> u64>,
     /// Set once ExitBootServices has been signaled.
     at_runtime: Once<()>,
+    /// SMRR base and size
+    smrr_base_size: Once<(u32, u32)>,
```
makubacki: nit: I think a small named struct would read better in code than the tuple. Like: `SmrrRange { base, size }`

patina_mm_supervisor/src/smrr.rs
```diff
@@ -0,0 +1,339 @@
...
+/// When enabled, the CPU raises a machine check if code is fetched from outside
+/// the SMRR protected region while in SMM. This also sets the lock bit in the
+/// SMM feature control MSR, after which the feature configuration cannot be
+/// modified until the next processor reset.
+///
+/// # Panics
+///
+/// Panics if the CPU does not report support for SMM Code Access Check.
+#[cfg_attr(coverage_nightly, coverage(off))]
+pub fn configure_smm_code_access() {
```
makubacki: Do all of the `pub` functions in here need to be `pub` or could they be `pub(crate)`?

patina_mm_supervisor/src/mem/page_allocator.rs
```diff
@@ -545,6 +545,71 @@ impl ScannedRegions {
...
+        // adjacent range is found, so ordering does not matter. Adjacency is
+        // considered regardless of allocation state, because the SMRR must
+        // cover a single contiguous physical range.
+        loop {
+            let mut found = false;
+            for &(region_base, region_size, _) in regions {
+                if region_base < smrr_base as u64 && smrr_base as u64 == region_base + region_size {
+                    // Region sits immediately before the current range: extend downward.
+                    smrr_base = region_base as u32;
+                    smrr_size = smrr_size.wrapping_add(region_size as u32);
```
makubacki: Instead of truncating to `u32` before adding and wrapping on overflow, it would be more robust to keep working in `u64` and convert once at the end with checked functions.

- Get rid of most of the casts
- Use `checked_add()` here: `smrr_size = smrr_size.checked_add(region_size)?;`
- Do a final conversion to `u32` at the end:

  ```rust
        let smrr_base = u32::try_from(smrr_base).ok()?;
        let smrr_size = u32::try_from(smrr_size).ok()?;
  ```

patina_mm_supervisor/src/smrr.rs
```diff
@@ -0,0 +1,339 @@
...
+        // SAFETY: We only set the code check enable and lock bits, which is the
+        // architecturally defined way to enable SMM Code Access Check. The write
+        // is idempotent and performed only when the value actually changes.
+        unsafe { write_msr(MSR_SMM_FEATURE_CONTROL, new_smm_feature_control_msr) };
+    }
+}
+
+/// Returns the largest power of two less than or equal to `value`, or `0` if
+/// `value` is `0`.
+const fn get_power_of_two32(value: u32) -> u32 {
```
makubacki, line 92: Functions like this should be moved somewhere else. But that can wait until later.

patina_mm_supervisor/src/smrr.rs
```diff
@@ -0,0 +1,339 @@
...
+/// Sets the mask field ([31:12]) of a raw SMRR mask register value for the
+/// given region `size`.
+fn mask_reg_set_mask(raw: u64, size: u32) -> u64 {
+    // Mask field = ~(size - 1) >> 12, placed back into [31:12].
+    let mask_bits = (!(size.wrapping_sub(1))) as u64 & PHYS_ADDR_MASK;
+    (raw & !PHYS_ADDR_MASK) | mask_bits
+}
+
+/// Returns `true` if bit 10 is set in a raw SMRR mask register value.
+fn mask_reg_bit10_set(raw: u64) -> bool {
```
makubacki: Simple functions operating on scalars like this could be `const`.

patina_mm_supervisor/src/lib.rs
vineelko: `per_core_init()` is more appropriate here. I have moved the code inside that.

patina_mm_supervisor/src/mem/page_allocator.rs
vineelko: Updated it.

patina_mm_supervisor/src/smrr.rs
vineelko: Updated the code to not have them here.

patina_mm_supervisor/src/state.rs
vineelko: Fixed it.

patina_mm_supervisor/src/smrr.rs
vineelko: switched to pub(crate) as appropriate.

patina_mm_supervisor/src/smrr.rs
vineelko: I agree I felt the same, will do it in a later PR.

patina_mm_supervisor/src/mem/page_allocator.rs
vineelko: Updated.

patina_mm_supervisor/src/lib.rs
vineelko: Fixed it and added `expect()` to that call as part of the moved code to `per_core_init()`

makubacki (approved):

makubacki:
Overriding CI since this is to a feature branch and current failures are expected.

PR #1666, opened 2026-07-29
[major] SDK Refactor:  Republish base, update consumers, and cleanup [REBASE & FF]

cfernald:
## Description

Second phase of large Patina SDK refactor, including:

### Add `CorePerformance` to services

The performance refactor left out declaring the `CorePerformance` service
in the derive service macro for `CorePerformance`. This commit simply
adds the missing service declaration.

### SDK Refactor: Consume r_efi republish from patina crate

This commit changes all uses of r_efi in the patina repro to instead use
the r_efi republished from the patina crate, and removing r_efi from
individual cargo dependencies.
 
### SDK Refactor: Publish all of base from Patina root

This commit republishes all of the `Patina::base::*` from the Patina
root. This flattens callers' imports and makes accessing common base types
easier.

### patina: Scope allow(static_mut_refs) only to necessary tests

Moves the global `#![allow(static_mut_refs)]` attributes to the test
modules that require it, rather than having it at the crate level.
This change improves code safety and clarity by limiting the scope of
this allowance to only where it's needed.

###

- [ ] Impacts functionality?
- [ ] Impacts security?
- [x] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- [x] Q35 Boot
- [x] ArmVirt Boot
- [x] Cargo make all

## Integration Instructions

Consumers that import anything under `patina::base::` should instead import directly from `patina::`

### DIFF
--- a/components/patina_acpi/Cargo.toml
+++ b/components/patina_acpi/Cargo.toml
@@ -13,7 +13,6 @@ workspace = true
 log = { workspace = true }
 memoffset = { workspace = true }
 mockall = { workspace = true, optional = true }
-r-efi = { workspace = true }
 zerocopy = { workspace = true }
 zerocopy-derive = { workspace = true }
 
--- a/components/patina_acpi/src/acpi.rs
+++ b/components/patina_acpi/src/acpi.rs
@@ -20,7 +20,7 @@ use core::{
 };
 
 use patina::{
-    base::SIZE_4GB,
+    SIZE_4GB,
     component::{
         hob::Hob,
         service::{IntoService, Service, memory::MemoryManager},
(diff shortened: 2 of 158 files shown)


vineelko (approved):
Minor: Looks like there are few more references to `r_efi` in `patina\core\patina_internal_cpu` and `patina\sdk\patina\src\arch`.  

Also should we just remove `r-efi = { version = "7", default-features = false }` from workspace Cargo.toml and make it local to patina sdk's Cargo.toml so others won't use it by accident in the future? No a big fan of this but just calling it out.

cfernald:
> Minor: Looks like there are few more references to `r_efi` in `patina\core\patina_internal_cpu` and `patina\sdk\patina\src\arch`.
> 
> Also should we just remove `r-efi = { version = "7", default-features = false }` from workspace Cargo.toml and make it local to patina sdk's Cargo.toml so others won't use it by accident in the future? No a big fan of this but just calling it out.

Thanks, think I missed in the rebase. Will fix.

makubacki (approved):

PR #1671, opened 2026-07-30
Integrate major branch changes into main [REBASE & FF]

cfernald:
## Description

Merged the following changes from the major branch

### Description

This PR addresses multiple issues with the current performance structure:

1. Performance data is not available until the component is dispatched. This can cause significant
   misrepresentation of the DXE execution time and cause events from images to be lost if not
   dispatched in the first loop.
2. Performance implementations in the SDK require SDK statics, which is generally an SDK antipattern.

This PR does the following:

1. Moves the core performance implementation into the core, into a new struct named `CorePerformance`,
   following the pattern established by the timer and memory management.
2. Creates a new service trait `PerformanceMeasurement` that exposes this core support.
3. Updates the component to now use this core service to produce the protocol & publish tables.
4. Change initialization and configuration of performance measurements to be done by the core.
5. Updates all documentation according to these changes.
6. Some smaller rework and restructuring.

Resolves: [#1483](https://github.com/OpenDevicePartnership/patina/issues/1483)

### Reviewing this PR

This PR is intended to be merged as a single commit. The individual commits do attempt to follow a
basic progression of the code rework but may be incomplete at the partial steps. Recommended review flow:

1. Start with the new theory of operation md (`performance_measurement.md`) to understand the intended structure.
2. Review the new service interface (`PerformanceMeasurement`).
3. Review the new core implementation.
4. Review the changes to the component to consume the new service.

---

### Description

**Change SerialIO trait to be mutable**

The current `SerialIO` implementation takes `&self` instead of `&mut self` for its methods, forcing
all serial implementations to implement interior mutability. However, because of the nature of the
serial port use cases, the typical locking mechanisms that avoid deadlocks, such as the `TplMutex`,
are not readily accessible. Because of this, internal implementations have one-off, and at times
questionable, interior mutability implementations.

This commit removes the expectation on the serial port to ensure this, and instead provides a
re-usable `SharedSerial` wrapper that implements interior mutability in a suitable way for serial
port use cases.

**patina: Simplify serial implementations**

This commit takes advantage of the new mutable references to cleanup the serial port implementations.
Namely by dropping re-initialization of interior references used as a workaround for the lack of
mutability in the `SerialIO` trait.

**patina: split uart module submodules**

This commit splits the null, 16550, and pl011 UART implementations into their own submodules under
the `uart` module. This better aligns to general code organization practices. This additionally
removes some unnecessary conditional complication.

---

### Description

This PR removes the deprecated `without_transport_init` function for the next major release. This
[...]

PR #1677, opened 2026-07-31
`--no-default-features` updates for tests [Rebase & FF]

makubacki:
## Description

Includes fixes for issues found while running `test --no-default-features` against individual crates.

---

**sdk/patina: Uncouple test and alloc**

Fixes failures running: `cargo test -p patina --no-run --no-default-features`

component/metadata.rs uses `fixedbitset` which is an optional crate
dependency:

  `fixedbitset = { workspace = true, optional = true }`

It will only be pulled in when the `alloc` feature is active:

  `alloc = ["dep:goblin", "dep:fixedbitset"]`

The decision on whether which features are enabled and dependencies
included is made before `rustc` runs for tests.

This made `test` in `#[cfg(any(test, feature = "alloc"))]` not work
when `--no-default-features` is used as `alloc` is not enabled and
`fixedbitset` is therefore not linked.

Since `alloc` is enabled by default, this does not change the normal
`cargo test` build. This will cause `component` to not be included on
`--no-default-features` test invocations which was already the case
for commands like `cargo make check --no-default-features`.

`cargo test -p patina --no-run --no-default-features` was run to
filter the modules down what could be enabled.

---

**core: Gate `dfs()` method calls in tests on alloc**

The `dfs()` method on the `Bst` and `Rbt` types is gated on the alloc
feature, so the calls to the methods in tests need to be as well.

Allows running tests with `--no-default-features` to succeed.

---

**sdk: Gate c_ptr and status_code**

sdk/patina/src/lib.rs gates the alloc crate:

```rust
#![cfg_attr(any(test, feature = "alloc"), feature(allocator_api))]
```

Code in sdk/patina/src/base/c_ptr.rs was using `alloc::boxed::Box`
without applying a similar gate. This change applies the gate on
usage there so the crate can be built without the alloc feature.

A similar fix in the sdk is made in status_code.rs. When built with
`cargo hack check --workspace --no-default-features` the following
error was seen:

```
error[E0599]: no method named `concat` found for array `[&[u8]; 2]` in the current scope
    let mut data_buffer = [any_as_u8_slice(&header), any_as_u8_slice(&data)].concat();
```

`[T]::concat()` on an array of slices returns `Vec<u8>` with alloc's
`Concat` trait.

Three changes were made for this:

1. Splitting the import block, so `use core::{mem, ptr, slice}` so
   `mem` and `slice` are gated on `alloc`.
2. Updating `pub fn report_status_code_with_data<T>` to be gated on
   the `alloc` feature.
3. Updating `fn any_as_u8_slice<T: Sized>(p: &T) -> &[u8]` to be
   gated on the `alloc` feature as well to avoid a dead code warning
   since it was only called by `report_status_code_with_data<T>`.

---

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- `cargo test --no-default-features`
- `cargo hack check --workspace --no-default-features`
- `cargo hack test --no-run --workspace --no-default-features`
- `cargo make test`

Note: https://github.com/OpenDevicePartnership/patina/issues/1680 tracks potentially using `cargo-hack` to perform no default testing on each crate in the future.

## Integration Instructions

- N/A

### DIFF
--- a/components/patina_adv_logger/README.md
+++ b/components/patina_adv_logger/README.md
@@ -41,6 +41,8 @@ below.
 #### Example
 
 ```rust
+# #[cfg(feature = "component")]
+# {
 use patina_dxe_core::*;
 use patina::{debug::log::Format, peripheral::serial::uart::UartNull};
 use patina_adv_logger::{component::AdvancedLoggerComponent, logger::{AdvancedLogger, TargetFilter}};
@@ -75,6 +77,7 @@ pub extern "efiapi" fn _start(physical_hob_list: *const c_void) -> ! {
 
    # loop { }
 }
+# }
 ```
 
 ### Platform Integration
--- a/core/patina_internal_core/src/collections/bst.rs
+++ b/core/patina_internal_core/src/collections/bst.rs
@@ -706,8 +706,11 @@ mod tests {
         assert_eq!(bst.storage.len(), 8);
         assert!(bst.add(10).is_err()); // Can't add the same value twice
 
-        let values = bst.dfs();
-        assert_eq!(values, [2, 3, 5, 6, 7, 8, 9, 10]);
+        #[cfg(feature = "alloc")]
+        {
+            let values = bst.dfs();
+            assert_eq!(values, [2, 3, 5, 6, 7, 8, 9, 10]);
+        }
     }
 
     #[test]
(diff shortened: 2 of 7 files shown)


makubacki:
This can wait to be merged until after the `major` branch is merged into `main` (https://github.com/OpenDevicePartnership/patina/pull/1671).

cfernald (approved):

vineelko (approved):

Issue #1664, opened 2026-07-29
[Feature]: make `compatability_mode_allowed` feature additive


Javagedes:
### Feature Overview

The `compatability_mode_allowed` feature has an implementation of `activate_compatiability_mode` that is swapped depending on the feature flag. Multiple tests are conditionally compiled based off of this flag, which means some unit tests are never actually run.

### Solution Overview

Paths to resolve this feature:

1. Only have one implementation of `MemoryProtectionPolicy::activate_compatibility_mode` and instead have a `cfg-if` statement based off the feature flag in `PrivateImageData::activate_compatability_mode` for actually calling `MemoryProtectionPolicy::activate_compatability_mode` or not. This will require only a single conditional compilation and will also ensure the function always exists and can be tested while also being conditionally compiled out when not used. 
2. Leave as-is. The replaced functionality is minor (simple error log and return error).

### Alternatives Considered

N/A

### Urgency

Medium

### Are you going to implement the feature request?

I will implement the feature

### Do you need maintainer feedback?

Maintainer feedback requested

### Anything else?

@os-d @makubacki for comments on solutions. I've provided three and am looking for feedback.

Issue #1663, opened 2026-07-29
[Feature]: make `v1_resource_descriptor_support` feature additive


Javagedes:
### Feature Overview

The `v1_resource_dscriptor_support` feature currently swaps `ResourceDescriptorV2` support for `ResourceDescriptorV1` support. As the parent issue suggests, this goes against rust's rules that Features should be _additive_ (as specified in [Feature Unification](https://doc.rust-lang.org/cargo/reference/features.html#feature-unification).

### Solution Overview

Paths to resolve this feature:

1. If both HOBs can exist at the same time, then we should elect to always support `ResourceDescritporV2` HOBs and the feature should only add support for parsing `ResourceDescriptorV1` HOBs.

2. Create a `v2_resource_descriptor_support` feature that is enabled by default and allow the platform to elect to enable one or the other (or both)

2. If both HOBs cannot exist at the same time, then the feature should be completely removed, and a platform configuration should be added to decide which HOB to acknowledge.

### Alternatives Considered

N/A

### Urgency

Medium

### Are you going to implement the feature request?

I will implement the feature

### Do you need maintainer feedback?

Maintainer feedback requested

### Anything else?

@os-d @makubacki for comments on solutions. I've provided three and amlooking for feedback.

Javagedes:
The intent is to remove this feature in the next few weeks, so this will be put on the backlog with the intent to remove the feature rather than fix it in a way that makes it additive.

PR #1685, opened 2026-08-01
[major] Remove "alloc" feature from default feature list of all Patina crates

vineelko:
## Description

**Background:** Behavior of `--all-features` with explicit package vs without explicit package:
```
cargo clippy --all-features --tests
workspace
  crate a(feature fa)           <--- all features defined in crate_a are enabled and built as a test binary
    crate b(feature fb)         <--- all features defined in crate_b are enabled and built as a lib

cargo clippy -p crate_a --all-features --tests
workspace
  crate a(feature fa)           <--- all features defined in crate_a are enabled and built as a test binary
    crate b(feature fb)         <--- only features selected when specifying crate_b as dependency are enabled(if non selected, default features are applied) and built as a lib
```
**NOTE:** `--tests` is nothing special here any target can be picked

**Problem:** Rust MM Supervisor avoids hidden heap allocations from dependencies
by ensuring the "alloc" feature is not enabled. Since Patina currently includes
"alloc" in its default feature set, the MM Supervisor branch had to explicitly
remove it from its default features(`default = []`) and also by not defining an
explicit `#[global_allocator]`. Because `default-features = false` only affects
a direct dependency, not transitive dependencies

Now with `default=[]` in place on patina sdk in feature/supv branch,

Before the SDK refactor, c_ptr.rs is gated as shown below. Which means trying to
run `cargo clippy --tests` did not break the build with or without
--all-features because in both cases c_ptr.rs is fully stubbed out as patina
sdk is neither built with test feature(remember this is a dependency to whatever
crate is being built) nor alloc(since we have set default=[]).

```
sdk\patina\src\lib.rs
#[cfg(any(test, feature = "alloc"))]        <----- stubbed out because patina sdk is a dependency for whatever test crate is being built right now
extern crate alloc;

#[cfg(any(test, feature = "alloc"))]        <----- stubbed out because patina sdk is a dependency for whatever test crate is being built right now
pub mod boot_services;

sdk\patina\src\boot_services\c_ptr.rs       <----- stubbed out because patina sdk is a dependency for whatever test crate is being built right now
use alloc::boxed::Box;
```

After the SDK refactor, c_ptr.rs is NOT gated. Which means trying to run 
`cargo clippy --tests` will break the build with or without `--all-features` because in
both the cases c_ptr.rs is always present and `extern crate alloc` is always
absent as neither test feature(again remember this is a dependency to whatever
crate is being built) nor alloc(since we have set default=[]).

```
sdk\patina\src\lib.rs
#[cfg(any(test, feature = "alloc"))]        <----- stubbed out because patina sdk is a dependency for whatever test crate is being built right now
extern crate alloc;

pub mod base;                               <----- NOT STUBBED OUT

sdk\patina\src\base\c_ptr.rs                <----- NOT STUBBED OUT
use alloc::boxed::Box;
```

Resulting in below error:

  ```
  error[E0433]: cannot find module or crate `alloc` in this scope
 --> sdk\patina\src\base\c_ptr.rs:9:5
  |
9 | use alloc::boxed::Box;
  |     ^^^^^ use of unresolved module or unlinked crate `alloc`
  ```

While restoring the feature gate on `c_ptr` fixes this immediate issue(also done
in https://github.com/OpenDevicePartnership/patina/pull/1677), debugging "feature"
related failures like this is not a very pleasant experience. In particular,
`--all-features` only applies to the package being built and does not propagate
to dependencies. To some extent `cargo hack` makes it easier to run failing
command on individual packages but that brings in its own set of new challenges.
After spending a considerable amount of time untangling the dependencies and
finding the root cause, it seems that removing "alloc" from the default feature set
across all Patina crates would alleviate some of the debugging pain.

**Benefits:**
1. Patina crates no longer assume the presence of the `alloc` feature.
2. Crates within Patina repo are nothing special and they have to explicitly
   declare the required feature set, making feature usage easier to audit and
   maintain.

    ```
    E:\repos\patina>cargo tree -e features -p patina_ffs_extractors -e=no-dev
    patina_ffs_extractors v23.0.1 (E:\repos\patina\sdk\patina_ffs_extractors)
    ...
    ├── patina feature "default"  <----

    vs

    E:\repos\patina>cargo tree -e features -p patina_ffs_extractors -e=no-dev
    patina_ffs_extractors v23.0.1 (E:\repos\patina\sdk\patina_ffs_extractors)
    ...
    ├── patina feature "alloc"    <----
    ```
3. Rust MM Supervisor can now opt out of `alloc` without requiring any hacks.
   Please note that #[global_allocator] should still be avoided so that the
   compiler can flag any dependent code paths that require heap allocations.

This is a breaking change because existing consumers relying on Patina's default
feature set will need to explicitly enable the required features. Most existing
crates have already been updated as part of this commit.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [x] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

`cargo make all` and validated on hardware platform.

## Integration Instructions

All external crates relying on patina crates will need to add "alloc" to the features list like below
`patina = { workspace = true, features = ["alloc"] }`

### DIFF
--- a/components/patina_acpi/Cargo.toml
+++ b/components/patina_acpi/Cargo.toml
@@ -21,7 +21,7 @@ patina_test = { workspace = true }
 
 [dev-dependencies]
 mockall = { workspace = true }
-patina = { workspace = true, features = ["mockall"] }
+patina = { workspace = true, features = ["alloc", "mockall"] }
 
 [features]
 mockall = ["dep:mockall", "std"]
--- a/components/patina_adv_logger/Cargo.toml
+++ b/components/patina_adv_logger/Cargo.toml
@@ -17,7 +17,7 @@ required-features = ['std']
 
 [dependencies]
 log = { workspace = true }
-patina = { workspace = true }
+patina = { workspace = true, features = ["alloc"] }
 patina_test = { workspace = true }
 spin = { workspace = true }
 zerocopy = { workspace = true }
--- a/components/patina_mm/Cargo.toml
+++ b/components/patina_mm/Cargo.toml
@@ -14,7 +14,7 @@ workspace = true
 cfg-if = { workspace = true }
 log = { workspace = true }
 mockall = { workspace = true, optional = true }
-patina = { workspace = true }
+patina = { workspace = true, features = ["alloc"] }
 zerocopy = { workspace = true }
 zerocopy-derive = { workspace = true }
 
--- a/components/patina_performance/Cargo.toml
+++ b/components/patina_performance/Cargo.toml
@@ -13,7 +13,7 @@ workspace = true
 [dependencies]
 log = { workspace = true }
 
-patina = { workspace = true }
+patina = { workspace = true, features = ["alloc"] }
 patina_mm = { workspace = true }
 zerocopy = { workspace = true }
 zerocopy-derive = { workspace = true }
--- a/components/patina_samples/Cargo.toml
+++ b/components/patina_samples/Cargo.toml
@@ -12,7 +12,7 @@ workspace = true
 
 [dependencies]
 log = { workspace = true }
-patina = { workspace = true }
+patina = { workspace = true, features = ["alloc"] }
 patina_macro = { workspace = true }
 patina_smbios = { workspace = true }
 
--- a/components/patina_smbios/Cargo.toml
+++ b/components/patina_smbios/Cargo.toml
@@ -10,7 +10,7 @@ description = "System Management BIOS (SMBIOS) support for Patina UEFI component
 [dependencies]
 log = { workspace = true }
 mockall = { workspace = true, optional = true }
-patina = { workspace = true }
+patina = { workspace = true, features = ["alloc"] }
 patina_macro = { workspace = true }
 zerocopy = { workspace = true }
 zerocopy-derive = { workspace = true }
--- a/components/patina_test/Cargo.toml
+++ b/components/patina_test/Cargo.toml
@@ -13,7 +13,7 @@ workspace = true
 [dependencies]
 linkme = { workspace = true }
 log = { workspace = true }
-patina = { workspace = true }
+patina = { workspace = true, features = ["alloc"] }
 patina_macro = { workspace = true }
 spin = { workspace = true }
 
--- a/core/patina_debugger/Cargo.toml
+++ b/core/patina_debugger/Cargo.toml
@@ -13,7 +13,7 @@ workspace = true
 [dependencies]
 cfg-if = { workspace = true }
 gdbstub = { workspace = true }
-patina = { workspace = true }
+patina = { workspace = true, features = ["alloc"] }
 patina_internal_cpu = { workspace = true }
 log = { workspace = true }
 spin = { workspace = true }
@@ -28,7 +28,6 @@ serial_test = { workspace = true }
 patina_mtrr = { workspace = true }
 
 [features]
-default = []
 alloc = []
 # All non-std features to test in CI
 ci_features = ["alloc"]
--- a/core/patina_internal_core/Cargo.toml
+++ b/core/patina_internal_core/Cargo.toml
@@ -13,14 +13,13 @@ workspace = true
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [features]
-default = ["alloc"]
 alloc = []
 # All non-std features to test in CI
 ci_features = ["alloc"]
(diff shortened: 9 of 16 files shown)


makubacki (approved):
I agree with the change. Especially after sorting through similar issues last week. However, I think this should target the major branch given it is a breaking change and given the immediate code changes made in #1677, it shouldn't' be urgent.

vineelko:
Sure Michael, will get this into major branch.

makubacki:
Added @Javagedes since he's looking at additive feature changes.

Javagedes:
Overall I don't agree with disabling default features across patina. That is very uncommon across the ecosystem.

In my opinion the answer is that if you don't think the feature should be enabled by default, then we should not have it in the default features list. But there are some features that should realistically be always on (alloc features for example).

Regardless of my opinion and whether you follow through with this, there are three features that are not additive, so if any of them are swapped from enabled to disabled by default, will cause breaking changes:

1. `alloc` feature in `patina_debugger` will disable almost all of the functionality we know and love in the debugger. However enabling / disabling this feature at the moment also requires a manual change in `patina` crate to change the location at which `patina_debugger::initialize` is called. See: https://github.com/OpenDevicePartnership/patina/issues/1686
2. `compatability_mode_allowed`: This is also not additive; it swaps implementation of `activate_compatiability_mode`. I think this is off by default already, so I don't think this will have an issue. See https://github.com/OpenDevicePartnership/patina/issues/1664
3. `v1_resource_descriptor_support`. This is also not additive. It swaps how hobs are parsed. I think this is off by default so it should not be an issue. See https://github.com/OpenDevicePartnership/patina/issues/1663

kuqin12:
> In my opinion the answer is that if you don't think the feature should be enabled by default, then we should not have it in the default features list. But there are some features that should realistically be always on (alloc features for example).
> 
The `alloc` is actually the main issue here. We will not support the `alloc` nor put global allocator in the supervisor binary for simplicity.

Javagedes:
> > In my opinion the answer is that if you don't think the feature should be enabled by default, then we should not have it in the default features list. But there are some features that should realistically be always on (alloc features for example).
> 
> The `alloc` is actually the main issue here. We will not support the `alloc` nor put global allocator in the supervisor binary for simplicity.

I should have looked at the PR more closely. The PR title is poorely worded. This is not disabling all default features across the patina crates. This is simply removing `alloc` from the default list of features specifically for `patina_internal_core`. That is not as big of a deal in my eyes.

However I still don't understand why you cannot just do `default-features = false` for your crate, rather removing it from the default list of features somewhere else. It at least prevents a breaking change.

vineelko:
Thanks Joey, let me update the title. I was more furious after scratching my head debugging it, hence the title 😄  

> However I still don't understand why you cannot just do default-features = false for your crate, rather removing it from the default list of features somewhere else. It at least prevents a breaking change.

The problem is not about the direct usage of a dependency. It is the indirect inclusion of "alloc" crate which is a pain to debug and isolate. default-features=false only prevent an immediate dependency to not include its default feature set but it does not prevent a transitive dependency to include its defaults. On top that, it works fine for normal `cargo build` with minor fixes, but fails for various clippy/test configurations we run as part of `cargo make all`.

kuqin12:
> > However I still don't understand why you cannot just do default-features = false for your crate, rather removing it from the default list of features somewhere else. It at least prevents a breaking change.
> 
> The problem is not about the direct usage of a dependency. It is the indirect inclusion of "alloc" crate which is a pain to debug and isolate. default-features=false only prevent an immediate dependency to not include its default feature set but it does not prevent a transitive dependency to include its defaults. On top that, it works fine for normal `cargo build` with minor fixes, but fails for various clippy/test configurations we run as part of `cargo make all`.

Yes, the `alloc` should not be propagated to the platform level. But we should not let it creep into the basic crates unconditionally, either.

vineelko:
> Yes, the alloc should not be propagated to the platform level. But we should not let it creep into the basic crates unconditionally, either.

Agreed, this PR only makes the dependency on "alloc" crate explicit but does not prevent any alloc feature usage from a transitive dependency(though it makes auditing the dependency tree via "cargo tree" slightly better to pin point). 

Why a crate needs a dependency on "alloc" is a different question which requires more scrutiny and if needed should be isolated with a feature guard as already done in https://github.com/OpenDevicePartnership/patina/pull/1677

Javagedes (approved):

PR #1687, opened 2026-08-03
[major] Remove "alloc" from "core" feature

vineelko:
## Description

[major] Remove "alloc" from "core" feature

- The core feature is redundant and ambiguous with its dependent
  feature "alloc" now explicitly called out. Hence removing it.
- The actual usage of the "core" feature need to reevaluated eventually,
  As its current usage is hacky just to expose
  arch::enable_interrupts() and arch::disable_interrupts() only to
  Patina crates (also referred to as "core" crates). However, it does
  not prevent external crates from enabling the core feature in their
  dependencies and using these restricted APIs. Overall, feature flags
  are intended for conditional compilation, not for enforcing
  visibility or access control over internal APIs.
---
- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

`cargo make all` in `Patina` and `patina-dxe-core-qemu`

## Integration Instructions

NA: core feature is expected to be enabled only by internal crates of Patina. So removal of "alloc" should not affect consumers of patina crates. Hence the change is not marked as breaking.

### DIFF
--- a/core/patina_internal_cpu/Cargo.toml
+++ b/core/patina_internal_cpu/Cargo.toml
@@ -15,7 +15,7 @@ features = ["std"]
 
 [dependencies]
 log = { workspace = true }
-patina = { workspace = true, features = ["alloc", "core"] }
+patina = { workspace = true, features = ["alloc"] }
 spin = { workspace = true, features = ["rwlock"] }
 patina_paging = { workspace = true }
 cfg-if = { workspace = true }
--- a/core/patina_internal_cpu/Cargo.toml
+++ b/core/patina_internal_cpu/Cargo.toml
@@ -15,7 +15,7 @@ features = ["std"]
 
 [dependencies]
 log = { workspace = true }
-patina = { workspace = true, features = ["alloc"] }
+patina = { workspace = true, features = ["alloc", "core"] }
 spin = { workspace = true, features = ["rwlock"] }
 patina_paging = { workspace = true }
 cfg-if = { workspace = true }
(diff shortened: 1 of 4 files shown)


kuqin12 (approved):

os-d (commented):
I think you can just remove `alloc` from the `core` feature rather than strip it out. It isn't needed and solves your problem. I think we need to think a bit more about exposing everything in the SDK. Right now it is these two functions, not the biggest deal (though we really don't want folks to use these directly), but we should decide if there is a separate location where this lives if we think it is inappropriate for the SDK to contain these at all

makubacki:
> I think you can just remove `alloc` from the `core` feature rather than strip it out. It isn't needed and solves your problem. I think we need to think a bit more about exposing everything in the SDK. Right now it is these two functions, not the biggest deal (though we really don't want folks to use these directly), but we should decide if there is a separate location where this lives if we think it is inappropriate for the SDK to contain these at all

I agree with this approach for now. Not that removing `core` is an issue at the moment necessarily, but the immediate problem can be solved without impacting another larger topic.

vineelko:
> I think you can just remove `alloc` from the `core` feature rather than strip it out. It isn't needed and solves your problem. I think we need to think a bit more about exposing everything in the SDK. Right now it is these two functions, not the biggest deal (though we really don't want folks to use these directly), but we should decide if there is a separate location where this lives if we think it is inappropriate for the SDK to contain these at all

I am fine with that too and will update the PR.

I don't mind where these APIs live today (SDK or not). But what we are trying to do here is,
"Let's allow these two APIs for access with-in Patina crates but not beyond Patina crates". Unfortunately, there is nothing in the language that lets us do this cleanly other than redefining the crate boundaries. I think not even a common crate/location will address that.

makubacki (approved):

PR #1688, opened 2026-08-04
[major][Rebase & FF] Consolidate alloc crate usage and opt out of alloc feature for patina_internal_cpu's patina sdk dependency

### LINKED ISSUE #1689: [Feature]: Consolidate/Segregate usage of alloc crate and alloc feature across Patina crates
### Feature Overview

The goal of this work is to appropriately segregate Patina code based on the `alloc` feature. This segregation can be achieved either by guarding entire modules behind the `alloc` feature or by selectively gating the individual types within a module that depend on `alloc`.

Today, the usage of the `alloc` crate and the `alloc` feature is not consistent across the Patina crates. Historically, this was not an issue because most downstream components implicitly assumed the presence of `alloc` in some form. However, this assumption no longer holds for Rust MM Supervisor, whose design intentionally avoids any form of dynamic allocation in the Patina crates and modules that it consumes.

As a result, if a Patina crate inadvertently introduces a dependency on `alloc`, Rust MM Supervisor fails to build because the dependency requires a global allocator:

```text
Compiling mm_supervisor v14.3.0 (E:\repos\patina_bins\mm_supervisor)
error: no global memory allocator found but one is required; link to std or add `#[global_allocator]` to a static item that implements the GlobalAlloc trait

error: could not compile `mm_supervisor` (bin "mm_supervisor") due to 1 previous error
```

Unfortunately, this error is not very actionable because it does not identify which Patina crate or module introduced the dependency on `alloc`. To make matters worse, `alloc` is a built-in crate rather than a normal Cargo dependency, so tools such as `cargo tree` cannot be used to trace where it enters the dependency graph. In practice, this leaves manual code inspection as the primary way to identify and eliminate unintended `alloc` usage.

### Solution Overview

- Identify current usage of alloc crate cross Patina crate
- Constrain the usage to non-test code for easy auditing
- Consolidate the usage of alloc feature around alloc crate
- Ultimately if a consume is not opting into "alloc" feature, all alloc related function should be stubbed out from Patina.

### Alternatives Considered

_No response_

### Urgency

Low

I will implement the feature

No maintainer feedback needed

### Anything else?

_No response_

vineelko:
## Description

This PR originally set out to consolidate and gate all `alloc` crate usage across the Patina repo behind an appropriate `alloc` feature flag. That turned out to be trickier and more fragile than expected, due to implicit assumptions about `alloc` crate usage baked in throughout Patina. This effort actually started while building the Rust MM supervisor, where the Patina crates it depends on should not implicitly opt into the `alloc` crate. After a few iterations, we deliberately narrowed the scope of this change to only the crates the supervisor relies on(`patina_internal_cpu`), gating those explicitly. As a result, this change is now trimmed down to the 3 commits below. The first two are good general cleanup/consolidation on their own, the third commit stops `patina_internal_cpu` from enabling the alloc feature on its patina SDK dependency, removing the unnecessary alloc requirement for the Rust MM supervisor's current dependency set.

[major] Consolidate extern crate alloc declarations to crate root(lib.rs/main.rs)

- This clean up will enable guarding alloc crate usage in a later PR

---
[major] Use `std` instead of `alloc` in tests

- Isolate alloc crate usage to non-test code
- Test code is always compiled for the host target with std, so alloc
  imports should not be conditionally guarded with test.
- Remove `extern crate std;` declaration from all tests.

---

[major] patina_internal_cpu: Do not enable "alloc" feature on patina sdk crate

- `alloc` feature on patina is not required to build patina_internal_cpu.
- This eliminates the dependency on alloc for rust mm_supervisor(with
  its current dependencies)
- explicitly validated by running below commands on `patina_internal_cpu`

```
cargo make all
cargo clippy -p patina_internal_cpu --all-targets --all-features -- -D warnings
cargo test -p patina_internal_cpu --no-run --no-default-features
```
---

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

`cargo make all` in patina and patina-dxe-core-qemu

## Integration Instructions

This does not change the interface between external consumers and patina crates. Hence not marking as breaking change.

### DIFF
--- a/components/patina_acpi/src/acpi.rs
+++ b/components/patina_acpi/src/acpi.rs
@@ -813,8 +813,6 @@ where
 
 #[cfg(test)]
 mod tests {
-    extern crate std;
-
     use crate::signature::MAX_INITIAL_ENTRIES;
 
     use super::*;
--- a/components/patina_acpi/src/service.rs
+++ b/components/patina_acpi/src/service.rs
@@ -168,9 +168,9 @@ pub(crate) trait AcpiProvider {
 #[cfg(test)]
 #[cfg_attr(coverage, coverage(off))]
 mod tests {
-    use alloc::boxed::Box;
     use core::mem;
     use patina::component::service::memory::StdMemoryManager;
+    use std::boxed::Box;
 
     use crate::acpi_table::AcpiFadt;
 
--- a/components/patina_adv_logger/src/lib.rs
+++ b/components/patina_adv_logger/src/lib.rs
@@ -8,7 +8,7 @@
 #![deny(missing_docs)]
 #![cfg_attr(coverage, feature(coverage_attribute))]
 
-#[cfg(any(feature = "alloc", test, doc))]
+#[cfg(any(feature = "alloc", doc))]
 extern crate alloc;
 
 mod memory_log;
--- a/components/patina_adv_logger/src/logger.rs
+++ b/components/patina_adv_logger/src/logger.rs
@@ -351,7 +351,6 @@ where
 mod tests {
     use core::{ffi::c_void, ptr};
 
-    use alloc::boxed::Box;
     use log::Log;
     use patina::standard::efi;
     use patina::{
@@ -360,6 +359,7 @@ mod tests {
         peripheral::serial::uart::UartNull,
         pi::hob::{GUID_EXTENSION, GuidHob, HobHeader},
     };
+    use std::boxed::Box;
 
     use crate::{
         logger::{AdvancedLogger, TargetFilter},
--- a/components/patina_adv_logger/src/memory_log.rs
+++ b/components/patina_adv_logger/src/memory_log.rs
@@ -423,9 +423,9 @@ impl AdvLoggerMessageEntry {
 const TEST_DATA_SIZE: usize = 128;
 
 #[cfg(test)]
-pub(crate) fn create_buffer_v5(timer_frequency: u64, hw_port_disabled: bool) -> alloc::boxed::Box<[u8]> {
+pub(crate) fn create_buffer_v5(timer_frequency: u64, hw_port_disabled: bool) -> std::boxed::Box<[u8]> {
     let header_size = size_of::<AdvLoggerInfoV5>();
-    let mut buffer = alloc::vec![0_u8; header_size + TEST_DATA_SIZE].into_boxed_slice();
+    let mut buffer = std::vec![0_u8; header_size + TEST_DATA_SIZE].into_boxed_slice();
     let header = AdvLoggerInfoV5 {
         signature: AdvLoggerInfo::SIGNATURE,
         version: ADV_LOGGER_INFO_VERSION_V5,
@@ -457,9 +457,9 @@ pub(crate) fn create_buffer_v5(timer_frequency: u64, hw_port_disabled: bool) -> 
 }
 
 #[cfg(test)]
-pub(crate) fn create_buffer_v6(timer_frequency: u64, new_address: u64) -> alloc::boxed::Box<[u8]> {
+pub(crate) fn create_buffer_v6(timer_frequency: u64, new_address: u64) -> std::boxed::Box<[u8]> {
     let header_size = size_of::<AdvLoggerInfoV6>();
-    let mut buffer = alloc::vec![0_u8; header_size + TEST_DATA_SIZE].into_boxed_slice();
+    let mut buffer = std::vec![0_u8; header_size + TEST_DATA_SIZE].into_boxed_slice();
     let header = AdvLoggerInfoV6 {
         v5: AdvLoggerInfoV5 {
             signature: AdvLoggerInfo::SIGNATURE,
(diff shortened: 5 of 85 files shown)


makubacki:
@vineelko, since this is 1/n, can you write the complete plan you have in mind in a single GitHub issue and link that to the PR? Then we can see where things are heading, agree that makes sense, and review the PR in that context.

vineelko:
@makubacki Added https://github.com/OpenDevicePartnership/patina/issues/1689 issue with additional details to support the changes being made. Feel free to let me know if you see any edits.

vineelko:
> @makubacki Added #1689 issue with additional details to support the changes being made. Feel free to let me know if you see any edits.

@makubacki, As I dig more into it, it feels like reliance on alloc feature/crate is very fragile and more intertwined than I originally thought within patina crates. This is more evident due to the fact that we built different combinations of the crates via `cargo clippy` and `cargo test` runs. 

Fully resolving these dependency issues is likely to take more time (not worth the effort), so I'm narrowing the scope of this PR to the changes needed to make the Rust MM Supervisor buildable without depending on the `alloc` crate/feature. As a result, I'll abandon the planned follow up patch series(will adjust the PR title accordingly).

The first 3 commits are good to have in terms of code cleanup without any functional impact. The fourth commit removes opting into the `alloc` feature on the `patina` SDK crate when building `patina_internal_cpu`. This is the immediate change needed for rust mm supervisor to build without `#[global_allocator]`. All changes have been verified to be buildable on q35 and other rust bins.

```
cargo make all
cargo clippy -p patina_internal_cpu --all-targets --all-features -- -D warnings
cargo test -p patina_internal_cpu --no-run --no-default-features
```

kuqin12 (approved):

makubacki:
> > @makubacki Added #1689 issue with additional details to support the changes being made. Feel free to let me know if you see any edits.
> 
> @makubacki, As I dig more into it, it feels like reliance on alloc feature/crate is very fragile and more intertwined than I originally thought within patina crates. This is more evident due to the fact that we built different combinations of the crates via `cargo clippy` and `cargo test` runs.
> 
> Fully resolving these dependency issues is likely to take more time (not worth the effort), so I'm narrowing the scope of this PR to the changes needed to make the Rust MM Supervisor buildable without depending on the `alloc` crate/feature. As a result, I'll abandon the planned follow up patch series(will adjust the PR title accordingly).
> 

It's unclear what the final state of `alloc` is in the patina repo and what developer guidance is for going forward.

Can you add an "Alloc Guidance" page to the "Developer Guides" section of the mdbook that explains (simply) how the feature should be used and how developers should use it when writing their code? You also have things in this PR like "use `std` instead of `alloc` in tests" and "Refer to types by name instead of fully qualified paths" that should be documented. Then, what state is the overall patina codebase left in after this change? If partial, what's the plan to finish the work or change direction to it is not left in a partial state?

vineelko:
@makubacki, Added the documentation for the guidance of using alloc crate and feature in patina crates.

makubacki:
> @makubacki, Added the documentation for the guidance of using alloc crate and feature in patina crates.

Thanks. As noted, this is very tedious. I'll look through the rest of the file changes on Monday.

Are the first three commits complete for their intended purpose? For example, does [[major] Use std instead of alloc in tests](https://github.com/OpenDevicePartnership/patina/pull/1688/changes/5c38c4c33272656a067ed0c33985a9995969d465) completely update all cases where std should be used instead of alloc in tests? Does [[major] Refer to types by name instead of fully qualified paths](https://github.com/OpenDevicePartnership/patina/pull/1688/changes/5ed2eecaa81541131b18b4ff8584633f1141093c) update all occurrences?

vineelko:
> > @makubacki, Added the documentation for the guidance of using alloc crate and feature in patina crates.
> 
> Thanks. As noted, this is very tedious. I'll look through the rest of the file changes on Monday.
> 
> Are the first three commits complete for their intended purpose? For example, does [[major] Use std instead of alloc in tests](https://github.com/OpenDevicePartnership/patina/pull/1688/changes/5c38c4c33272656a067ed0c33985a9995969d465) completely update all cases where std should be used instead of alloc in tests? Does [[major] Refer to types by name instead of fully qualified paths](https://github.com/OpenDevicePartnership/patina/pull/1688/changes/5ed2eecaa81541131b18b4ff8584633f1141093c) update all occurrences?

No worries, Michael, take your time. I am pretty sure I did cover the entire code base multiple times and made sure any missing instances(discovered later) have been adjusted back into those 3 commits as appropriate.

vineelko:
Updated the PR with the commits we discussed offline. 

```
> f267b062 2026-08-05 Vineel Kovvuri[MSFT]  [major] patina_internal_cpu: Do not enable "alloc" feature on patina sdk crate 
> c785213f 2026-08-13 Vineel Kovvuri[MSFT]  [major] Use `std` instead of `alloc` in tests
> bb7b5e06 2026-08-05 Vineel Kovvuri[MSFT]  [major] Consolidate extern crate alloc declarations to crate root(lib.rs/main.rs)
```

Left below commits
```
> 4913ff10 2026-08-05 Vineel Kovvuri[MSFT]  Docs: Alloc crate and feature guidance
> 5ed2eeca 2026-08-04 Vineel Kovvuri[MSFT]  [major] Refer to types by name instead of fully qualified paths
```

makubacki:
Thanks for updating the scope to `patina_internal_cpu`. As discussed, we'll plan to merge this into `major` after UEFI Services if there is no urgency to do so earlier.

PR #1692, opened 2026-08-06
Ignore slow tests [Rebase & FF]

makubacki:
## Description

Since a small set of unit tests increase time significantly, this follows the guidance in the [Rust Programming Language book](https://doc.rust-lang.org/book/ch11-02-running-tests.html#ignoring-tests-unless-specifically-requested) to ignore them during normal test runs.

- `cargo make test` before avg (after clean): 192 seconds
- `cargo make test` after avg (after clean): 105 seconds

This does not include ignored tests in test coverage, though that's possible if preferred.

---

**Ignore long-running tests by default**

Excluding `compile_fail_tests` reduces "cargo make test" reduces the
test run time by about 45% (after clean). Since it is common to run
`cargo make test` throughout development, this has a significant
impact on iterative development.

This change marks these tests as ignored by default, so that they will
only run when explicitly requested with `cargo test --ignored`.

A few tests in `components\patina_test\src\component.rs` were
already using the `#[ignore]` attribute but don't successfully run.
The `storage::new()` calls initialize `StandardBootServices`. This
code ends up registering a Ready to Boot notification, but
`create_event_ex()` is not initialized in the Boot Services table and
panics with a message.

This change removes these tests for now with a tracking issue filed
to add them back when the UEFI services are available to mock.

---

**Makefile.toml: Add test-ignored task**

Adds a new task to run ignored tests excluding those in doctests.

- `cargo make test` - Runs all tests excluding ignored tests.
- `cargo make test-ignored` - Runs ignored tests excluding those in
  doctests.

`cargo make all` runs both sets.

This allows `cargo make test` to run much faster by ignoring a small
set of slow tests.

---

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- `cargo make test`
- `cargo make test-ignored`
- `cargo make all`

## Integration Instructions

- N/A

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -235,6 +235,12 @@ clear = true
 command = "cargo"
 args = ["test", "@@split(CARGO_MAKE_TASK_ARGS,;)"]
 
+[tasks.test-ignored]
+description = "Runs ignored tests with native cargo test behavior. Example `cargo make test-ignored` or `cargo make test-ignored -p package_name"
+clear = true
+command = "cargo"
+args = ["test", "@@split(CARGO_MAKE_TASK_ARGS,;)", "--lib", "--tests", "--examples", "--", "--ignored"]
+
 [tasks.test-cov]
 description = "Run tests and collect coverage data without generating reports."
 install_crate = false
@@ -487,5 +493,6 @@ dependencies = [
     "coverage",
     "doc-test",
     "doc",
+    "test-ignored",
     "test-mdbook",
 ]
--- a/components/patina_test/src/component.rs
+++ b/components/patina_test/src/component.rs
@@ -144,10 +144,8 @@ pub(crate) mod tests {
     extern crate std;
 
     use super::*;
-    use crate::{
-        __private_api::TestCase,
-        alloc::{boxed::Box, format},
-    };
+
+    use crate::alloc::{boxed::Box, format};
     use core::mem::MaybeUninit;
     use patina::{
         BinaryGuid,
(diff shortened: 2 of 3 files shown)


vineelko (approved):
Minor: Should we just keep the `ignore` instead of changing them to `cfg(any())`? Just trying to avoid another pattern to skip running/building tests.

makubacki:
> Minor: Should we just keep the `ignore` instead of changing them to `cfg(any())`? Just trying to avoid another pattern to skip running/building tests.

I'll start by saying, the goal here is to have a way to move slow tests out of the main test path. After some research, [Ignoring Tests Unless Specifically Requested - Rust Programming Language](https://doc.rust-lang.org/book/ch11-02-running-tests.html#ignoring-tests-unless-specifically-requested) is an authoritative source that describes this exact problem, so I went with that.

The tests that picked up `cfg(any())` did not run successfully before (`cargo make test --ignored` will fail today). Intuitively, I haven't considered that a problem because they're ignored by normal test commands. However, that contradicts the guidance for the use case above.

I lean toward liking the policy that ignored tests still pass/fail as expected but are ignored in the main test path. Then, the question comes what to do with the tests previously `ignore`d but that don't run? I chose `cfg(any())` to keep them dormant. I'm now favoring just removing them entirely (they would still be available in source history). In any case, I'll file a tracking issue to come back to them.

makubacki:
> > Minor: Should we just keep the `ignore` instead of changing them to `cfg(any())`? Just trying to avoid another pattern to skip running/building tests.
> 
> I'm now favoring just removing them entirely (they would still be available in source history). In any case, I'll file a tracking issue to come back to them.

Removed them from this PR. Tracking issue: https://github.com/OpenDevicePartnership/patina/issues/1697

Javagedes (approved):

PR #1693, opened 2026-08-06
[FCP] RFC: Use cargo-nextest

### LINKED ISSUE #1691: [Feature]: Consider cargo-nextest
### Feature Overview

This issue tracks using [cargo-nextest](https://nexte.st/) instead of `cargo test` as the default test runner in patina.

The primary motivating factor is [performance improvements](https://nexte.st/docs/benchmarks/), in addition to its other features such as a more intuitive UI and the ability to identify misbehaving tests.

In general, patina is cautious to establish new third-party tool dependencies, so further evaluation will take place to make a proposal in an RFC.

### Solution Overview

Use `cargo-nextest` instead of `cargo-test`.

### Alternatives Considered

_No response_

### Urgency

Low

I will implement the feature

Maintainer feedback requested

### Anything else?

_No response_

makubacki:
## Description

**Status: FCP**

Proposes using [cargo-nextest](https://nexte.st/) for running tests in Patina.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [x] Includes documentation?

## How This Was Tested

- N/A

## Integration Instructions

- N/A

### DIFF
--- a/docs/src/rfc/text/0032-cargo-nextest.md
+++ b/docs/src/rfc/text/0032-cargo-nextest.md
@@ -1,0 +1,160 @@
+# RFC: Run Tests with `cargo-nextest`
+
+This RFC proposes replacing `cargo test` with [cargo-nextest](https://nexte.st/) as the test runner behind Patina's
+`cargo make` test tasks, while keeping `cargo test` for running doctests since nextest can't run them.
+
+## Change Log
+
+- 2026-08-06: Initial draft of RFC.
+- 2026-08-11: Update "Unresolved Questions" to indicate that `#[serial]` should be removed in cases where nextest's
+  isolation is sufficient. Add details about how nextest works with Miri in Technology Background.
+
+## Motivation
+
+Patina's has a few test-related tasks in `Makefile.toml`:
+
+- `test`
+- `check_tests`
+- `check-no-default-features-tests`
+- `test-asan`
+- Coverage tasks
+
+Those run tests through plain `cargo test`. `cargo test` works, but its output is hard to read at a glance for a
+workspace this size. It does not give a summary of which tests were slow, and offers no way to retry a test that fails
+intermittently.
+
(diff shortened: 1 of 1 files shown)


os-d (approved):
imo, the only feature we were trying to get with `#[serial]` is resolved by cargo-nextest's isolation and I would converge on cargo-nextest to keep it simpler.

The other thing to note is that miri supports cargo-nextest, so there are the same benefits to running miri with nextest instead of cargo-test and timing on miri matters even more because it is so slow. However, it is almost certain that miri's own runtime is going to far overshadow the test harness time; I don't have data on if nextest helps significantly in overall miri time.

makubacki:
> imo, the only feature we were trying to get with `#[serial]` is resolved by cargo-nextest's isolation and I would converge on cargo-nextest to keep it simpler.

I haven't looked thoroughly, but I agree that I think we can drop them all.

> The other thing to note is that miri supports cargo-nextest, so there are the same benefits to running miri with nextest instead of cargo-test and timing on miri matters even more because it is so slow. However, it is almost certain that miri's own runtime is going to far overshadow the test harness time; I don't have data on if nextest helps significantly in overall miri time.

I'll look into this a bit more and add some miri info to the RFC.

vineelko (approved):

makubacki:
> I'll look into this a bit more and add some miri info to the RFC.

Done and pushed.

makubacki:
Note: The RFC has entered Final Comment Period (FCP).

cfernald (approved):
I agree. It seems if we aren't regularly exercising the correctness of serializiation or locking in tests in CI, then it will just start to fail and will be a burden to fix them up all the time. Better to drop it and only support nextext. Will be a bit annoying to no longer expect `cargo test` within a crate to not hit concurrency issues, though I guess that is solved with `cargo test --test-threads=1`

makubacki:
Note: Moved to FCP on Aug 11. Planning to merge this week.

PR #1694, opened 2026-08-06
Update and integrate uart_16550 to 0.8 [Rebase & FF]

makubacki:
## Description

⚠️ **Waiting on https://github.com/rust-osdev/uart_16550/issues/65 before mering**

This PR originally targeted updating to uart_16550 0.6 but was later updated to target 0.8 as discussed in the PR comments. The commit messages copied/pasted into this PR description below reflect the final changes to 0.8.

---

**Update and integrate uart_16550 to 0.8**

The 0.3.2 version of uart_16550 depends on the unmaintained x86 crate
(not updated in ~4 years), which pulls in an old 1.x bitflags release
that duplicates the version used everywhere else in Patina. This
uart_16550 update was raised before in PR 1560, but the duplicate
dependency has become enough of a problem to make the update now.

Bumping to 0.8 lets us drop the bitflags skip entry in deny.toml and
reduce the copies of bitflags in the dependency tree.

In this crate update, the old `SerialPort` and `MmioSerialPort` types
are replaced with a generic `Uart16550<Backend>` driver. Construction
is now fallible because it validates the address range and performs
real hardware presence checks before applying the baud rate and format.

Because the constructors are fallible, there is a tendency to `expect()`
on the return value which panics if the address is bad or the probe
fails. We generally don't want to panic because a UART probe check
failed as that could boot a device. In reality, it will be obvious
in most platforms if UART fails because there will be no output.

`new_io()` and `new_mmio()` are `const fn` and do not fail. They
just record basic information and defer everything else. The underlying
driver is built lazily the first time it is actually needed.

Presence probing and configuration happen through the explicit
`SerialIO::init` call, and a failed probe there does not block
reads or writes afterward.

The goal is to make UART safe while reducing potential panics and
allowing graceful degradation in cases where a UART might not behave
as expected.

---

**Cargo.toml: No longer allow multiple_crate_versions**

This was enabled until the uart_16550 crate could be updated so the
crate dependency on bitflags v1 could be removed. Since that is now
done, the lint no longer needs to be allowed.

---

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [x] Includes documentation?

## How This Was Tested

- `cargo make all`
- Boot ArmVirt and Q35 to EFI shell and check UART output
- Verify UART output on a physical Intel platform

**Today**

```
bitflags v1.3.2
├── raw-cpuid v10.7.0
│   └── x86 v0.52.0
│       └── uart_16550 v0.3.2
│           ├── patina v23.0.1 (patina/sdk/patina)

...

bitflags v2.13.1
├── gdbstub v0.7.10
│   └── patina_debugger v23.0.1 (patina/core/patina_debugger) (*)
├── patina_paging v11.0.6

...
```

**After**: bitflags 2.x used everywhere

## Integration Instructions

- N/A

---

This is a little opinionated in how fatal I think UART operations should be to boot. I'm happy to make changes if others have reasons to handle errors differently.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -54,7 +54,7 @@ quote = { version = "1" }
 scroll = { version = "0.13", default-features = false, features = ["derive"]}
 spin = { version = "0.12" }
 syn = { version = "3" }
-uart_16550 = { version = "0.3.2" }
+uart_16550 = { version = "0.8" }
 corosensei = { version = "0.3.4", default-features = false }
 uuid = { version = "1.23", default-features = false }
 zerocopy = { version = "0.8" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -90,10 +90,6 @@ suspicious = { level = "warn", priority = -1 }
 #     `the "_support" suffix in the feature name "v1_resource_descriptor_support" is redundant`
 #   This is allowed until the v1_resource_descriptor_support feature is dropped.
 redundant_feature_names = "allow"
-# cargo::multiple_crate_versions violation:
-#  `multiple versions for dependency `bitflags`: 1.3.2, 2.13.1 `
-# This is allowed until the uart_16550 crate is updated to v0.6+ which removes the bitflags v1 dependency.
-multiple_crate_versions = "allow"
 # pedantic::cast_possible_truncation - Temporarily allowed until all cases are updated (each fix needs a lot of
 #                                      attention) to determine if the truncation is intentional and safe or not.
 cast_possible_truncation = "allow"
--- a/docs/src/dev/principles/abstractions.md
+++ b/docs/src/dev/principles/abstractions.md
@@ -142,40 +142,47 @@ impl SerialIO for Terminal {
     }
 }
 
-use uart_16550::MmioSerialPort;
+use core::num::NonZero;
+use uart_16550::backend::MmioBackend;
+use uart_16550::{Config, Uart16550 as InnerUart16550};
 struct Uart16550(usize);
 
 impl Uart16550 {
     fn new(addr: usize) -> Self {
         Self(addr)
     }
+
+    /// Returns `None` if the address is rejected or the device can't be detected.
+    fn port(&self) -> Option<InnerUart16550<MmioBackend>> {
+        let addr = core::ptr::NonNull::<u8>::with_exposed_provenance(NonZero::new(self.0)?);
+        // SAFETY: `self.0` is assumed to be a valid, exclusively-owned MMIO address for this example.
+        unsafe { InnerUart16550::new_mmio(addr, 1) }.ok()
+    }
 }
 
 impl SerialIO for Uart16550 {
     fn init(&self) {
-        unsafe { MmioSerialPort::new(self.0).init() };
+        if let Some(mut port) = self.port() {
+            let _ = port.init(Config::DEFAULT);
+        }
     }
 
     fn write(&self, buffer: &[u8]) {
-        let mut port = unsafe { MmioSerialPort::new(self.0) };
-
-        for b in buffer {
-            port.send(*b);
+        if let Some(mut port) = self.port() {
+            port.send_bytes_exact(buffer);
         }
     }
 
     fn read(&self) -> u8 {
-        let mut port = unsafe { MmioSerialPort::new(self.0) };
-        port.receive()
+        let mut byte = 0u8;
+        if let Some(mut port) = self.port() {
+            port.receive_bytes_exact(core::slice::from_mut(&mut byte));
+        }
+        byte
     }
 
     fn try_read(&self) -> Option<u8> {
-        let mut port = unsafe { MmioSerialPort::new(self.0) };
-        if let Ok(value) = port.try_receive() {
-            Some(value)
-        } else {
-            None
-        }
+        self.port()?.try_receive_byte().ok()
     }
 }
 
--- a/sdk/patina/src/peripheral/serial/uart/uart_16550.rs
+++ b/sdk/patina/src/peripheral/serial/uart/uart_16550.rs
@@ -7,20 +7,116 @@
 //! SPDX-License-Identifier: Apache-2.0
 //!
 
-use uart_16550::MmioSerialPort;
-use uart_16550::SerialPort as IoSerialPort;
+use core::ptr::NonNull;
+
+use crate::log_debug_assert;
+use uart_16550::backend::{MmioBackend, PioBackend};
+use uart_16550::{BaudRate, Config, Uart16550 as Inner};
+
+/// The configuration applied the first time a `Uart16550` port is used.
+const CONFIG: Config = Config { baud_rate: BaudRate::Baud38400, ..Config::DEFAULT };
+
+/// A lazily-initialized I/O port-mapped backend.
+///
+/// Construction only records the base port. The underlying driver is validated but not
+/// hardware-probed on first use (see `driver`).
+#[derive(Debug)]
+pub struct IoPort {
+    base: u16,
+    inner: Option<Inner<PioBackend>>,
+}
+
+impl IoPort {
+    const fn new(base: u16) -> Self {
+        Self { base, inner: None }
+    }
+
+    /// Returns the constructed driver, building it from the base port on the first call.
+    /// Returns `None` if the port is rejected (e.g. its register range overflows `u16`). The
+    /// attempt is retried on every call since there is not much overhead to retry.
+    ///
+    /// This only validates the address, it does not probe for hardware presence, so ports that
+    /// don't implement a full 16550 register file still work for raw reads/writes even though
+    /// `init` below can't confirm that they're present.
+    fn driver(&mut self) -> Option<&mut Inner<PioBackend>> {
+        if self.inner.is_none() {
+            // SAFETY: Forwarded from `Uart16550::new_io`'s caller contract that `self.base` is a
+            // valid I/O port range, exclusively owned for the lifetime of this `Uart16550`.
+            self.inner = unsafe { Inner::new_port(self.base) }.ok();
+        }
+        self.inner.as_mut()
+    }
+
+    /// Applies [`CONFIG`] to the device. Some ports may not have a probeable identity, so a failed
+    /// presence probe here does not block later reads/writes, it just means the baud rate/format may
+    /// not match [`CONFIG`].
+    fn init(&mut self) {
+        if let Some(driver) = self.driver() {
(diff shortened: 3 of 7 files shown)


docs/src/dev/principles/abstractions.md
```diff
@@ -142,40 +142,46 @@ impl SerialIO for Terminal {
...
 struct Uart16550(usize);
 
 impl Uart16550 {
     fn new(addr: usize) -> Self {
         Self(addr)
     }
+
+    /// Returns `None` if the address is rejected or the device can't be detected.
+    fn port(&self) -> Option<InnerUart16550<MmioBackend>> {
+        let addr = core::ptr::NonNull::new(core::ptr::with_exposed_provenance_mut::<u8>(self.0))?;
```
Javagedes: `NonNull` has a `with_exposed_provenance` function. Is this usable here?

Javagedes (approved):

docs/src/dev/principles/abstractions.md
makubacki: Yeah, I had to bring in `core::num::NonZero`, but updated it to use `NonNull::<T>::with_exposed_provenance()`.

os-d (approved):

phip1611 (approved):
thanks for bumping the dependency! Just a heads up. `uart_16550 @ 0.6.0` might be a little stricted than needed which can influence your testing on real hardware. I'll fix that in the next couple of weeks and cut a new release!

Context: https://github.com/rust-osdev/uart_16550/issues/65

makubacki:
> thanks for bumping the dependency! Just a heads up. `uart_16550 @ 0.6.0` might be a little stricted than needed which can influence your testing on real hardware. I'll fix that in the next couple of weeks and cut a new release!
> 
> Context: [rust-osdev/uart_16550#65](https://github.com/rust-osdev/uart_16550/issues/65)

Hey @phip1611! Thanks for the info and review.

makubacki:
> > thanks for bumping the dependency! Just a heads up. `uart_16550 @ 0.6.0` might be a little stricted than needed which can influence your testing on real hardware. I'll fix that in the next couple of weeks and cut a new release!
> > Context: [rust-osdev/uart_16550#65](https://github.com/rust-osdev/uart_16550/issues/65)
> 

I've confirmed that this does impact at least one relevant physical platform. I was able to work around it by updating [`ready_to_send()`](https://github.com/rust-osdev/uart_16550/blob/25b95ae2f3774c6e6624becd13ef6410d9e8efc8/src/lib.rs#L669-L689) locally with:

```diff
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -668,8 +668,6 @@ impl<B: Backend> Uart16550<B> {
     /// an established connection.
     pub fn ready_to_send(&mut self) -> Result<(), ByteSendError> {
         let lsr = self.lsr();
-        let msr = self.msr();
-        let mcr = self.mcr();
 
         // In FIFO mode, this bit is set when the transmitter’s FIFO is
         // completely empty, being 0 if there is at least one byte in the
@@ -679,12 +677,6 @@ impl<B: Backend> Uart16550<B> {
             return Err(ByteSendError::NoCapacity);
         }
 
-        // Software flow control. TODO, what to do with hardware flow control?
-        // Is this something we can and should support?
-        if !mcr.contains(MCR::LOOP_BACK) && !msr.contains(MSR::CTS) {
-            return Err(ByteSendError::RemoteNotClearToSend);
-        }
-
         Ok(())
     }
```

I'll leave this PR open until that release is made.

vineelko (approved):

sdk/patina/src/peripheral/serial/uart/uart_16550.rs
```diff
@@ -57,31 +156,47 @@ impl crate::peripheral::serial::SerialIO for Uart16550 {
...
+                if let Some(driver) = port.driver() {
+                    driver.send_bytes_exact(buffer);
                 }
             }
         }
     }
 
     fn read(&mut self) -> u8 {
+        // Blocks until a byte is available once the device is constructed. If the address was
+        // rejected, there is no hardware to wait on, so a `0` sentinel is returned rather than
```
cfernald: Should such a failure debug_assert?

cfernald (approved):

phip1611:
> > > thanks for bumping the dependency! Just a heads up. `uart_16550 @ 0.6.0` might be a little stricted than needed which can influence your testing on real hardware. I'll fix that in the next couple of weeks and cut a new release!
> > > Context: [rust-osdev/uart_16550#65](https://github.com/rust-osdev/uart_16550/issues/65)
> 
> I've confirmed that this does impact at least one relevant physical platform. I was able to work around it by updating [`ready_to_send()`](https://github.com/rust-osdev/uart_16550/blob/25b95ae2f3774c6e6624becd13ef6410d9e8efc8/src/lib.rs#L669-L689) locally with:


Could you please check if the comments in the latest PR version (**including the commit message**) align with your knowledge about Uart hardware? https://github.com/rust-osdev/uart_16550/pull/66

I'd appreciate if you could approve the PR then, so we can move forward!

makubacki:
> > > > thanks for bumping the dependency! Just a heads up. `uart_16550 @ 0.6.0` might be a little stricted than needed which can influence your testing on real hardware. I'll fix that in the next couple of weeks and cut a new release!
> > > > Context: [rust-osdev/uart_16550#65](https://github.com/rust-osdev/uart_16550/issues/65)
> > 
> > 
> > I've confirmed that this does impact at least one relevant physical platform. I was able to work around it by updating [`ready_to_send()`](https://github.com/rust-osdev/uart_16550/blob/25b95ae2f3774c6e6624becd13ef6410d9e8efc8/src/lib.rs#L669-L689) locally with:
> 
> Could you please check if the comments in the latest PR version (**including the commit message**) align with your knowledge about Uart hardware? [rust-osdev/uart_16550#66](https://github.com/rust-osdev/uart_16550/pull/66)
> 
> I'd appreciate if you could approve the PR then, so we can move forward!

@phip1611, the PR content and commit message look good to me. I also tried the PR (at commit [fe65e0b](https://github.com/rust-osdev/uart_16550/commit/fe65e0b003f2a320993d3cef15119f504a0d0877)) and it worked. I left a minor comment, but I'll approve the PR to show support for it moving ahead.

Cargo.toml
```diff
@@ -54,7 +54,7 @@ r-efi = { version = "7", default-features = false }
 scroll = { version = "0.13", default-features = false, features = ["derive"]}
 spin = { version = "0.12" }
 syn = { version = "3" }
-uart_16550 = { version = "0.3.2" }
+uart_16550 = { version = "0.6" }
```
phip1611: thank you for helping me to test things (especially 0.6.0 -> 0.7.0)! I also did some tests on real hardware and made sure the crate becomes even more flexible and safely usable on real hardware.

`0.8.0` is the latest version: https://github.com/rust-osdev/uart_16550/blob/main/CHANGELOG.md

I hope to unblock you now. Please let me know what your tests on real hardware are reporting, once you had time to test things

sdk/patina/src/peripheral/serial/uart/uart_16550.rs
```diff
@@ -7,20 +7,115 @@
...
 
-use uart_16550::MmioSerialPort;
-use uart_16550::SerialPort as IoSerialPort;
+use core::ptr::NonNull;
+
+use uart_16550::backend::{MmioBackend, PioBackend};
+use uart_16550::{BaudRate, Config, Uart16550 as Inner};
+
+/// The configuration applied the first time a `Uart16550` port is used.
+const CONFIG: Config = Config { baud_rate: BaudRate::Baud38400, ..Config::DEFAULT };
```
phip1611, line 17: Please note there was a small breaking change to the default config, see https://github.com/rust-osdev/uart_16550/blob/main/CHANGELOG.md#080---2026-08-16 / https://github.com/rust-osdev/uart_16550/pull/69

sdk/patina/src/peripheral/serial/uart/uart_16550.rs
makubacki: @cfernald, sorry for the delay in getting back. I got distracted by the uart_16550 0.6 -> 0.8 updates.

I'd expect that, at least for a `SerialLogger`, the write would also fail, however this would make it more obvious in a debug environment when a debugger is connected in the panic handler.

In that case, I think `log_debug_assert!()` *might* be incrementally a little more helpful since that would give a logger an opportunity to log somewhere else (e.g. `AdvancedLogger` would write to the memory log) and that might assist debug, depending on logger implementations.

All that to say, would this address your open?

```rust
    fn read(&mut self) -> u8 {
        // Blocks until a byte is available once the device is constructed. A rejected address
        // logs an error and panics in debug builds. In release builds, a `0` sentinel is returned
        // instead of spinning forever.
        let mut byte = 0u8;
        match self {
            Uart16550::Io(port) => match port.driver() {
                Some(driver) => driver.receive_bytes_exact(core::slice::from_mut(&mut byte)),
                None => log_debug_assert!("Uart16550::read on a rejected I/O port. Returning a 0 sentinel byte"),
            },
            Uart16550::Mmio(port) => match port.driver() {
                Some(driver) => driver.receive_bytes_exact(core::slice::from_mut(&mut byte)),
                None => log_debug_assert!("Uart16550::read on a rejected MMIO port. Returning a 0 sentinel byte"),
            },
        }
        byte
    }
```

Cargo.toml
makubacki: @phip1611, I don't see an issue with 0.8.0 in the hardware I've tested.

sdk/patina/src/peripheral/serial/uart/uart_16550.rs
cfernald: Yeah, I think that would address it. This seems pretty fundamental to mess us so I don't think we should dwell on it much and this would ensure there is at least an attempt to indicate what has happened.

makubacki: Updated.

phip1611:
Cool! Out of curiosity: is Patina already deployed somewhere (on scale) or is it still in an experimental stage?

makubacki:
> Cool! Out of curiosity: is Patina already deployed somewhere (on scale) or is it still in an experimental stage?

It's used in newer Surface devices (search for "Patina" on [this page](https://www.microsoft.com/d/surface-pro-for-business-13-inch-12th-edition-intel/8mzbmmcjzpn4#tab-tech-specs)) with other vendors at varying levels of evaluation and adoption.

PR #1701, opened 2026-08-08
[feature/supv] Add the support of programming MSEG during each core initialization

kuqin12:
## Description

This change adds a parser for the `MSEG_SMRAM_HOB` so that the SMM can program the MSEG MSR during the first MMI entry.

This change also exposed the `efi_main` to make it discoverable in the PDB file.

- [x] Impacts functionality?
- [x] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This was tested on a customized QEMU Q35 platform and booted to UEFI shell.

## Integration Instructions

For platforms elect to support SEA, integrate the MSEG prepare PEI into the platform and link the STM based CPU feature library into the MM initializer.

### DIFF
--- a/patina_mm_supervisor/src/init.rs
+++ b/patina_mm_supervisor/src/init.rs
@@ -31,6 +31,7 @@ use crate::{
     AllocationType, CommBufferConfig, MmSupervisorCore, PageOwnership, PlatformInfo, SharedPagingAllocator,
     intrinsics::read_cr3,
     is_buffer_inside_mmram,
+    mem::page_allocator::SmramDescriptor,
     mem::{self, page_allocator::coalesced_smrr_range},
     mm_policy::{self, MemDescriptorV1_0, dump_policy, gate::PolicyGate, walk_page_table},
     query_address_ownership,
@@ -69,6 +70,16 @@ pub enum PolicyInitError {
 
 /// Offset from SMBASE where the SMI handler code is located.
 const SMM_HANDLER_OFFSET: u64 = 0x8000;
+
+/// MSR index for `IA32_SMM_MONITOR_CTL`, which holds the MSEG base used to
+/// activate the dual-monitor treatment (Intel SDM Vol. 4).
+const IA32_SMM_MONITOR_CTL: u32 = 0x9b;
+
+/// `IA32_SMM_MONITOR_CTL.Valid` (bit 0). An STM may only be invoked when set.
+const SMM_MONITOR_CTL_VALID: u64 = 1;
+
+/// `IA32_SMM_MONITOR_CTL.MsegBase` (bits 31:12).
+const SMM_MONITOR_CTL_MSEG_BASE_MASK: u64 = 0xffff_f000;
 
 /// Index into the Fixup64 array for the SMI handler IDTR pointer.
 const FIXUP64_SMI_HANDLER_IDTR: usize = 5;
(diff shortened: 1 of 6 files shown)


makubacki:
I don't have time to review now, but am interested to do so next week if it can wait.

patina_mm_supervisor/src/init.rs
```diff
@@ -598,6 +633,18 @@ impl<P: PlatformInfo, const MAX_CPUS: usize> MmSupervisorCore<P, MAX_CPUS> {
...
+        // 1b-ii. Process the MSEG SMRAM HOB (`gMsegSmramGuid`), if published. It carries the
+        //        MSEG region reserved for an STM. Each core programs the base into
+        //        IA32_SMM_MONITOR_CTL during per-core init. Platforms without STM/SEA
+        //        integration do not publish this HOB, so its absence is not an error.
+        match find_guid_hob(hob_list_info, crate::MSEG_SMRAM_HOB_GUID).and_then(parse_mseg_smram_hob) {
+            Some(mseg_base) => {
+                init_state().set_mseg_base(mseg_base);
+                log::info!("MSEG base 0x{:x} discovered from MSEG SMRAM HOB", mseg_base);
+            }
+            _ => panic!("No usable MSEG SMRAM HOB; IA32_SMM_MONITOR_CTL will not be programmed"),
```
makubacki: This panic contradicts the text above:

```rust
        // 1b-ii. Process the MSEG SMRAM HOB (`gMsegSmramGuid`), if published. It carries the
        //        MSEG region reserved for an STM. Each core programs the base into
        //        IA32_SMM_MONITOR_CTL during per-core init. Platforms without STM/SEA
        //        integration do not publish this HOB, so its absence is not an error.
```

patina_mm_supervisor/src/init.rs
```diff
@@ -509,6 +523,27 @@ impl<P: PlatformInfo, const MAX_CPUS: usize> MmSupervisorCore<P, MAX_CPUS> {
...
+            return;
+        };
+
+        let value = (mseg_base & SMM_MONITOR_CTL_MSEG_BASE_MASK) | SMM_MONITOR_CTL_VALID;
+
+        // SAFETY: IA32_SMM_MONITOR_CTL is an architectural MSR available whenever VMX is
+        // reported by CPUID.01H:ECX.VMX. Only the architecturally defined Valid and MsegBase
+        // fields are set; reserved bits are masked off above, so the write cannot #GP on a
+        // reserved-bit violation. The write affects only this logical processor's MSR.
+        unsafe { write_msr(IA32_SMM_MONITOR_CTL, value) };
```
makubacki: I know that by this point the MSEG HOB was found and it's likely the CPU supports VMX. However, since `get_current_cpu_id()` is already available in intrinsics.rs:

https://github.com/OpenDevicePartnership/patina/blob/d2f470a2727f5aedf4403f385698fb924981a8c9/patina_mm_supervisor/src/intrinsics.rs#L82-L96

Conditionalizing the MSR write on `CpuidResult.ecx & (1 << 5) != 0` shouldn't be too much extra work, right?

patina_mm_supervisor/src/init.rs
```diff
@@ -70,6 +71,16 @@ pub enum PolicyInitError {
 /// Offset from SMBASE where the SMI handler code is located.
 const SMM_HANDLER_OFFSET: u64 = 0x8000;
 
+/// MSR index for `IA32_SMM_MONITOR_CTL`, which holds the MSEG base used to
+/// activate the dual-monitor treatment (Intel SDM Vol. 4).
+const IA32_SMM_MONITOR_CTL: u32 = 0x9b;
```
makubacki: More clear?

```suggestion
const IA32_SMM_MONITOR_CTL_MSR: u32 = 0x9b;
```

patina_mm_supervisor/src/init.rs
```diff
@@ -820,6 +867,33 @@ fn find_guid_hob(hob_list_info: &PhaseHandoffInformationTable, target_guid: pati
     None
 }
 
+/// Parses the MSEG SMRAM HOB payload (`gMsegSmramGuid`), a single
+/// [`SmramDescriptor`] describing the MSEG region carved out of SMRAM.
+///
+/// Returns the MSEG base address, or `None` if the region is empty.
+fn parse_mseg_smram_hob(data: &[u8]) -> Option<u64> {
```
makubacki, line 883: This seems like an easy function to add a unit test for now.

patina_mm_supervisor/src/init.rs
kuqin12: Good call. This panic was later added. Will update the comment.

patina_mm_supervisor/src/init.rs
kuqin12: Thanks. Agreed. Will update.

kuqin12: Updated the intrinsic interface to expose the other registers to cover this usage. Otherwise the existing interface does not have the needed information returned.

kuqin12: Updated the intrinsic interface to expose the other registers to cover this usage. Otherwise the existing interface does not have the needed information returned.

patina_mm_supervisor/src/init.rs
kuqin12: I ended up changing the panic into a warning.

makubacki (approved):

patina_mm_supervisor/src/init.rs
kuqin12: Agreed. This is added.

vineelko (approved):

PR #1702, opened 2026-08-08
SerialLogger: Add opt-in blocking, uncorrupted serial logging

vineelko:
## Description

The existing `SharedSerial` implementation uses non-blocking serial port acquisition and returns `DeviceError` when the port is unavailable. When used by the Rust MM Supervisor, this can result in log messages being dropped, especially when multiple cores attempt to log concurrently.

Add an opt-in blocking mode that lets callers block while acquiring the serial port instead of returning `DeviceError`:

- `SharedSerial::into_blocking()` converts a wrapper to spin (block) on acquisition rather than failing fast on contention.
- `Logger::with_blocking()` enables blocking mode on the serial logger.

Blocking the port alone only makes each individual `write_str` fragment atomic, not a whole record. Because `Format::write()` emits a single log record via multiple `write_str` calls (prefix, level, message, terminator), concurrent cores could still interleave those fragments and produce corrupted lines such as:

    INFO - AP (CPU INFO - AP (CPU 64) exiting holding pen

To fix this, add a logger-level `write_lock` that is held for the entire record when blocking mode is enabled, so a message is emitted atomically with respect to other cores.

```
UEFI_UART: 8/8/2026 12:42:14 AM | INFO - EBS completed successfully.
UEFI_UART: 8/8/2026 12:42:20 AM | INFO - BSP (CPU 0) waiting for APs to arrive...
UEFI_UART: 8/8/2026 12:42:20 AM | INFO - AP (CPU 68) checked in, entering holding pen...
UEFI_UART: 8/8/2026 12:42:20 AM | INFO - AP (CPU 16) checked in, entering holding pen...
UEFI_UART: 8/8/2026 12:42:20 AM | INFO - AP (CPU 24) checked in, entering holding pen...
UEFI_UART: 8/8/2026 12:42:20 AM | INFO - AP (CPU 64) checked in, entering holding pen...
UEFI_UART: 8/8/2026 12:42:20 AM | INFO - AP (CPU 8) checked in, entering holding pen...
UEFI_UART: 8/8/2026 12:42:20 AM | INFO - AP (CPU 66) checked in, entering holding pen...
UEFI_UART: 8/8/2026 12:42:20 AM | INFO - AP (CPU 70) checked in, entering holding pen...
UEFI_UART: 8/8/2026 12:42:20 AM | INFO - All 7 APs arrived
...
UEFI_UART: 8/8/2026 12:42:20 AM | INFO - AP (CPU 8) exiting holding pen
UEFI_UART: 8/8/2026 12:42:20 AM | INFO - AP (CPU 24) exiting holding pen
UEFI_UART: 8/8/2026 12:42:20 AM | INFO - AP (CPU 16) exiting holding pen
UEFI_UART: 8/8/2026 12:42:20 AM | INFO - AP (CPU 64) exiting holding pen
UEFI_UART: 8/8/2026 12:42:20 AM | INFO - AP (CPU 66) exiting holding pen
UEFI_UART: 8/8/2026 12:42:20 AM | INFO - AP (CPU 68) exiting holding pen
UEFI_UART: 8/8/2026 12:42:20 AM | INFO - AP (CPU 70) exiting holding pen
```

Existing consumers continue to use the default `new()` constructor and retain the current non-blocking, best-effort (lossy) behavior, making this change non-breaking. Consumers with use cases that require guaranteed, uncorrupted logging, such as the Rust MM Supervisor, can opt in through `with_blocking()`.

Note: blocking mode reintroduces a self-deadlock hazard if the same core re-enters the logger/port while already writing a record (e.g. logging from a panic handler mid-write); it is intended for opt-in use where guaranteed logging outweighs that risk.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Validated on physical platforms.

## Integration Instructions

NA

### DIFF
--- a/sdk/patina/src/debug/log/serial_logger.rs
+++ b/sdk/patina/src/debug/log/serial_logger.rs
@@ -133,6 +133,7 @@ mod tests {
     use alloc::{string::String, sync::Arc, vec::Vec};
     use log::{Level, LevelFilter, Log, Metadata};
     use spin::Mutex;
+    use std::thread;
 
     fn metadata(level: Level, target: &str) -> Metadata<'_> {
         Metadata::builder().level(level).target(target).build()
@@ -196,5 +197,48 @@ mod tests {
         let args = format_args!("hello");
         logger.log(&log::Record::builder().args(args).level(Level::Info).target("test").build());
         assert_eq!(String::from_utf8(buffer.lock().clone()).expect("valid utf8"), "INFO - hello\r\n");
+    }
+
+    #[test]
+    #[ignore = "Stress test: spawns threads and logs many records; run explicitly with --ignored."]
+    fn test_serial_logger_blocking_concurrent_records_not_interleaved() {
+        // Each record is emitted as several `write_str` fragments (level, separator, message,
+        // terminator). In blocking mode the logger-level `write_lock` must hold for the whole
+        // record so two threads writing concurrently cannot interleave those fragments. We log
+        // two distinct messages from two threads and assert every rendered record is intact -
+        // i.e. the stream is a sequence of whole "A" or "B" records, never a garbled mix.
+        const ITERATIONS: usize = 500;
+
+        let (mock, buffer) = recording_serial();
+        let logger = Logger::new(Format::Standard, &[], LevelFilter::Trace, mock).with_blocking();
+
+        thread::scope(|scope| {
+            for message in ["aaaa", "bbbb"] {
+                let logger = &logger;
+                scope.spawn(move || {
+                    for _ in 0..ITERATIONS {
+                        logger.log(
+                            &log::Record::builder()
+                                .args(format_args!("{message}"))
+                                .level(Level::Info)
+                                .target("test")
+                                .build(),
+                        );
+                    }
+                });
+            }
+        });
+
+        let output = String::from_utf8(buffer.lock().clone()).expect("valid utf8");
+        let (mut a_count, mut b_count) = (0usize, 0usize);
+        for record in output.split_terminator("\r\n") {
+            match record {
+                "INFO - aaaa" => a_count += 1,
+                "INFO - bbbb" => b_count += 1,
+                garbled => panic!("interleaved/garbled record observed: {garbled:?}"),
+            }
+        }
+        assert_eq!(a_count, ITERATIONS, "missing or corrupted 'aaaa' records");
+        assert_eq!(b_count, ITERATIONS, "missing or corrupted 'bbbb' records");
     }
 }
--- a/sdk/patina/src/debug/log/serial_logger.rs
+++ b/sdk/patina/src/debug/log/serial_logger.rs
@@ -8,6 +8,7 @@
 //!
 use crate::peripheral::serial::{SerialIO, shared::SharedSerial};
 use core::marker::Send;
+use spin::Mutex;
 
 use super::Format;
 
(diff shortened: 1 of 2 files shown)


sdk/patina/src/debug/log/serial_logger.rs
```diff
@@ -25,20 +26,53 @@ where
...
+    }
+
+    /// Switches the logger to blocking (spinning) serial acquisition, consuming and returning `self`.
+    ///
+    /// In blocking mode, each record is written under a logger-level lock so concurrent cores
+    /// cannot interleave the individual `write_str` fragments of a single message, and the serial
+    /// port itself spins until free instead of failing fast on contention. This guarantees lossless,
+    /// uncorrupted output at the cost of a self deadlock hazard if the same core re-enters the
+    /// logger while already writing a record (e.g. logging from a panic handler mid record).
+    pub fn with_blocking(self) -> Self {
```
os-d: nit: In other places we've had this pattern: https://github.com/OpenDevicePartnership/patina/blob/801fa8fa6b7be059976d90a170aecdfb88b929b1/core/patina_debugger/src/debugger.rs#L145-L171. I don't have a preference, other than we stay consistent unless there is a good reason not to.

sdk/patina/src/debug/log/serial_logger.rs
```diff
@@ -149,4 +186,15 @@ mod tests {
         logger.log(&log::Record::builder().args(args).level(Level::Info).target("test").build());
         logger.flush();
     }
+
+    #[test]
```
os-d, line 186: If you wanted to have an interesting test you could have two threads attempt writing to the serial port and ensure that the order received is either A,B or B,A but not garbled. That adds new test dependencies for multithreading, but may be useful elsewhere. Don't consider it required from me, just an idea if we actually want to make sure this doesn't regress.

os-d (approved):
Does adv logger need a change as well? Or it is already handled there?

sdk/patina/src/debug/log/serial_logger.rs
vineelko: I considered that, but went with `with_blocking()` without a boolean parameter because the default behavior is always non-blocking, which is also the recommended behavior according to the current module docs due to the reentrancy issue.

`with_blocking(false)` didn't seem particularly useful for our use case. If we ever need the blocking behavior, it would be explicitly enabled with `with_blocking()`. So I intentionally chose not to expose an additional choice that doesn't provide much value.

os-d: Oh, I didn't mean passing in a parameter, I agree on not doing that, I just meant mutating the existing self instead of consuming self and producing a new one.

vineelko: Fixed it.

sdk/patina/src/debug/log/serial_logger.rs
vineelko: Added it.

vineelko:
> Does adv logger need a change as well? Or it is already handled there?

I only looked at it briefly, and it looks like `AdvancedLogger::log` does need it for creating the format record, though the actual memory logging and hardware write are protected by an `RwLock`. I haven't tested it with the Supervisor scenario yet, so I'll try to address it in a later PR.

kuqin12 (approved):

makubacki (approved):

PR #1703, opened 2026-08-10
smbios: use r-efi constants

kat-perez:
## Description

Require `r-efi` 7.1 and replace the remaining Patina-owned SMBIOS constants with the definitions released upstream in r-efi/r-efi#102.

The existing `SMBIOS_HANDLE_PI_RESERVED` and `SMBIOS_STRING_MAX_LENGTH` public names remain available as re-exports, so downstream Patina users do not need to change imports. The SMBIOS protocol GUID now comes directly from `r_efi::efi::protocols::smbios::PROTOCOL_GUID`.

Tracking: #1457

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- `cargo make fmt`
- `cargo make test -p patina_smbios`
- `cargo make check`

## Integration Instructions

N/A

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,7 +50,7 @@ patina_stacktrace = { version = "23.0.1", path = "core/patina_stacktrace" }
 patina_test = { version = "23.0.1", path = "components/patina_test" }
 proc-macro2 = { version = "1" }
 quote = { version = "1" }
-r-efi = { version = "7", default-features = false }
+r-efi = { version = "7.1", default-features = false }
 scroll = { version = "0.13", default-features = false, features = ["derive"]}
 spin = { version = "0.12" }
 syn = { version = "3" }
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -50,7 +50,7 @@ patina_stacktrace = { version = "23.0.1", path = "core/patina_stacktrace" }
 patina_test = { version = "23.0.1", path = "components/patina_test" }
 proc-macro2 = { version = "1" }
 quote = { version = "1" }
-r-efi = { version = "7", default-features = false }
+r-efi = { version = "7.1", default-features = false }
```
os-d, line 53: My understanding now is that repos should not depend on r-efi directly and should only consume through the SDK. Right @makubacki @cfernald ?

makubacki: That's correct and what is happening. The sdk is the only Cargo.toml that inherits r-efi from the workspace Cargo.toml (updated here).

r-efi could be specified directly in the sdk/patina/Cargo.toml, but I think there was a preference to keep the version maintained alongside others at the workspace level.

makubacki: If there is no concern with doing so, I'll create a separate PR to move r-efi directly there to reduce likelihood of a crate inadvertently picking up it directly.

os-d: Ah, my misreading. I thought that smbios was in patina-components

os-d (approved):

makubacki (approved):

PR #1704, opened 2026-08-10
Drop indoc Crate Dependency

os-d:
## Description

The indoc crate provides the indoc! macro to pretty print a set of strings with expected spacing instead of calling multiple writeln!() macros.

However, it does not support CRLF, leading to poor readability on some terminal emulators.

It is also only used in a single location. As a result, this commit drops the crate dependency and switches to using multiple writelncrlf!() macros in the single caller.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Seeing prints with poor formatting when printing the hob list. After this fix the prints align correctly.

## Integration Instructions

N/A.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,6 @@ fallible-streaming-iterator = { version = "0.1.9" }
 fixedbitset = { version = "0.5", default-features = false }
 gdbstub = { version = "0.7.10", default-features = false }
 goblin = { version = "0.10.7", default-features = false }
-indoc = { version = "2.0" }
 linked_list_allocator = { version = "0.10" }
 linkme = { version = "0.3.36" }
 log = { version = "0.4", default-features = false }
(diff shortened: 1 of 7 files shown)


makubacki (approved):
I suggest adding an entry to the [deny section](https://github.com/OpenDevicePartnership/patina/blob/801fa8fa6b7be059976d90a170aecdfb88b929b1/deny.toml#L138) in deny.toml to explain why this crate should not be used in the future.

os-d:
> I suggest adding an entry to the [deny section](https://github.com/OpenDevicePartnership/patina/blob/801fa8fa6b7be059976d90a170aecdfb88b929b1/deny.toml#L138) in deny.toml to explain why this crate should not be used in the future.

Added.

vineelko (approved):

PR #1705, opened 2026-08-11
Enable the cargo and pedantic clippy lint groups [Rebase & FF]

### LINKED ISSUE #1513: [Feature]: Enable Additional Clippy Lints
### Feature Overview

Clippy provides a large set of optional lints that provide additional safety.

### Solution Overview

Patina should consider which set to enable.

### Alternatives Considered

_No response_

### Urgency

Low

No maintainer feedback needed

### Anything else?

_No response_

makubacki:
## Description

Enables the [cargo](https://doc.rust-lang.org/clippy/lints.html#cargo) and [pedantic](https://doc.rust-lang.org/clippy/lints.html#pedantic) clippy lint groups. 

As noted in Cargo.toml, several pedantic lints are allowed either temporarily or permanently as noted.

There was a mix of manual modifications and those autogenerated by `clippy --fix` where available. Although there is a lot of what was repetitive work (running clippy frequently), AI was not used in the creation of these changes. It just seemed inefficient given the amount of text and manual work needed anyway. Lint descriptions used in individual commit messages were copied from the [clippy documentation](https://rust-lang.github.io/rust-clippy/master/).

Each lint that needs follow up has a tracking issue and is tracked collectively in: https://github.com/OpenDevicePartnership/patina/issues/1707

Commit messages where an individual lint was fixed are not included below due to prevent the PR description from being unreasonably long.

Some modifications are made to public APIs. If they are not substantial and semantic breaking, the plan is to leave them in this PR per [Patina breaking change criteria](https://github.com/OpenDevicePartnership/patina/blob/main/docs/src/rfc/text/0015-patina-release-process.md#criteria-for-breaking-changes).

---

**Inherit keywords and categories in each crate from the workspace**

- Define `keywords` and `categories` for the workspace in the root
  Cargo.toml.
- Update workspace crates to inherit `keywords` and `categories` from
  the workspace.

---

**Enable the cargo clippy lint group**

Ref: https://doc.rust-lang.org/clippy/lints.html#cargo

From the clippy documentation:

  The clippy::cargo group gives you suggestions on how to improve your
  Cargo.toml file. This might be especially interesting if you want to
  publish your crate and are not sure if you have all useful
  information in your Cargo.toml.

Temporary allowances were made for the `redundant_feature_names` and
`multiple_crate_versions` lints for the reasons given in the
Cargo.toml comments.

---

**Cargo.toml: Reset `undocumented_unsafe_blocks` back to deny from warn**

The `undocumented_unsafe_blocks` lint was sent to `warn` level when
Patina had a large number of SAFETY blocks missing comments. This is
no longer the case, so the lint is removed so it falls back to the
default value of `deny` to prevent future regressions.

---

**Explicitly enable default clippy lint groups**

As more clippy lints are being managed in the workspace Cargo.toml
file, this adds the default lint groups at their default levels so
it's clear what is enabled looking at `workspace.lints.clippy` in
Cargo.toml. These do not need any code changes since there is no
change from the implicit default settings applied previously.

Deny Lints

- correctness

  Ref: https://doc.rust-lang.org/clippy/lints.html#correctness

  From the clippy documentation:

    The clippy::correctness group is the only lint group in Clippy
    whose lints are deny-by-default and abort the compilation when
    triggered. This is for good reason: if you see a correctness lint,
    it means that your code is outright wrong or useless, and you
    should try to fix it.

    Lints in this category are carefully picked and should be free of
    false positives. So just #[allow]ing those lints is not
    recommended.

Warning Lints

- complexity

  Ref: https://doc.rust-lang.org/clippy/lints.html#complexity

  From the clippy documentation:

    The clippy::complexity group offers lints that give you suggestions
    on how to simplify your code. It mostly focuses on code that can be
    written in a shorter and more readable way, while preserving the
    semantics.

    If you should see a complexity lint, it usually means that you can
    remove or replace some code, and it is recommended to do so.
    However, if you need the more complex code for some expressiveness
    reason, it is recommended to allow complexity lints on a
    case-by-case basis.

- perf

  Ref: https://doc.rust-lang.org/clippy/lints.html#perf

  From the clippy documentation:

    The clippy::perf group gives you suggestions on how you can
    increase the performance of your code. Those lints are mostly
    about code that the compiler can’t trivially optimize, but has to
    be written in a slightly different way to make the optimizer job
    easier.

    Perf lints are usually easy to apply, and it is recommended to do
    so.

- style

  Ref: https://doc.rust-lang.org/clippy/lints.html#style

  From the clippy documentation:

    The clippy::style group is mostly about writing idiomatic code.
    Because style is subjective, this lint group is the most
    opinionated warn-by-default group in Clippy.

    If you see a style lint, applying the suggestion usually makes
    your code more readable and idiomatic. But because we know that
    this is opinionated, feel free to sprinkle #[allow]s for style
    lints in your code or #![allow] a style lint on your whole crate
    if you disagree with the suggested style completely.

- suspicious

  Ref: https://doc.rust-lang.org/clippy/lints.html#suspicious

  From the clippy documentation:

    The clippy::suspicious group is similar to the correctness lints
    in that it contains lints that trigger on code that is really sus
    and should be fixed. As opposed to correctness lints, it might be
    possible that the linted code is intentionally written like it is.

    It is still recommended to fix code that is linted by lints out of
    this group instead of #[allow]ing the lint. In case you
    intentionally have written code that offends the lint you should
    specifically and locally #[allow] the lint and add give a reason
    why the code is correct as written.

----

~39 commits addressing individual lints

---

**Fix remaining trivial lints in the pedantic clippy group**

Previous commits applied fixes for various clippy pedantic lints on
a per-lint basis. This commit applies fixes for the remaining trivial
lints in the pedantic group.

---

**sdk/hob_list: Defer clippy::same_length_and_capacity**

Leave a TODO to revisit this lint in the future and determine the
best way to address it. For now, this marks the lint as allowed to
unblock the overall clippy pedantic lint group.

---

**patina_dxe_core/pecoff: Defer clippy::case_sensitive_file_extension_comparisons**

Leave a TODO to revisit this lint in the future and determine the
best way to address it. For now, this marks the lint as allowed to
unblock the overall clippy pedantic lint group.

---

**patina_dxe_core/runtime: Allow clippy::pedantic::linkedlist**

Mark linkedlist as allowed for runtime_events and runtime_images
since the data structure is intentionally used there.

---

**Cargo.toml: Enable the pedantic clippy lint group**

REF: https://doc.rust-lang.org/clippy/lints.html#pedantic

At this point, a large number of clippy lints in the pedantic group
have been fixed. Remaining lints are marked as allowed with comments
about why each lint is allowed. Some are intentional and others are
temporary until the code is refactored to fix the lint. Cases where
a lint is temporarily allowed will have a GitHub tracking issue filed.

---

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- `cargo make all`

## Integration Instructions

- If an issue is encountered using an API modified by this PR, refer to the updated docs and the commit message to determine how to handle the change. For example, a return value may need to be used that wasn't previously.


components/patina_acpi/src/acpi_protocol.rs
```diff
@@ -284,17 +282,17 @@ mod tests {
         let mut table_key: usize = 0;
         let status = AcpiTableProtocol::install_acpi_table_ext(
             &AcpiTableProtocol::new(),
-            dummy_table.as_ptr() as *const c_void,
+            dummy_table.as_ptr().cast::<c_void>(),
             dummy_table.len(),
-            &mut table_key as *mut usize,
+            &raw mut table_key,
```
Javagedes, line 287: This is pretty cool. good to know for the future and a good clippy lint.

core/patina_debugger/src/debugger.rs
```diff
@@ -201,36 +206,34 @@ impl<T: SerialIO> PatinaDebugger<T> {
...
-                let conn = SerialConnection::new(&self.transport);
-
-                let builder = GdbStubBuilder::new(conn)
-                    .with_packet_buffer(gdb_buffer)
-                    .build()
-                    .map_err(|_| DebugError::GdbStubInit)?;
-
-                builder.run_state_machine(&mut target).map_err(|_| DebugError::GdbStubInit)?
-            }
+        let mut gdb = if let Some(_) = debug.gdb {
```
Javagedes: I think this could be even simpler:

```rust
let mut gdb = debug.gdb.take().ok_or_else(|| {
    /* everything in the else branch */
})?;
```

core/patina_internal_core/benches/bench_delete.rs
```diff
@@ -102,15 +102,13 @@ fn benchmark_delete_function(c: &mut Criterion) {
...
-                    match bst.delete(i.key()) {
-                        Ok(_) => {}
-                        Err(_) => {
-                            std::println!("{}", nums.len());
-                            std::println!("{nums:?}");
-                            std::println!("{}", nums_shuffled.len());
-                            std::println!("{nums_shuffled:?}");
-                            panic!("Failed to delete {i}");
-                        }
+                    if let Ok(()) = bst.delete(i.key()) {
```
Javagedes: can this just be:

```rust
if bst.delete(i.key()).is_err() {
  std::println!("{}", nums.len());
  std::println!("{nums:?}");
  std::println!("{}", nums_shuffled.len());
  std::println!("{nums_shuffled:?}");
  panic!("Failed to delete {i}");
}
```

core/patina_stacktrace/src/error.rs
```diff
@@ -148,7 +148,7 @@ impl Error {
             Error::UnwindCodeOutOfBounds { module, requested, available } => {
                 Error::UnwindCodeOutOfBounds { module: module.or(fallback), requested, available }
             }
-            other => other,
+            other @ Error::ImageNotFound { .. } => other,
```
Javagedes, line 151: learn something new every day.

Javagedes (approved):

core/patina_internal_core/benches/bench_delete.rs
makubacki: Yes, updated.

core/patina_debugger/src/debugger.rs
makubacki: I tried that but got:

```
error[E0277]: the `?` operator can only be used in a closure that returns `Result` or `Option` (or another type that implements `FromResidual`)
   --> core\patina_debugger\src\debugger.rs:219:86
    |
217 | ....take().ok_or_else(|| {
    |                       -- this function should return `Result` or `Option` to accept `?`
218 | ...data from the transport.
219 | ...t.try_read().map_err(|_| DebugError::TransportFailure)?.is_some() {}
    |                                                          ^ cannot use the `?` operator in a closure that returns `GdbStubStateMachine<'_, PatinaTarget, SerialConnection<'_, T>>`
```

In the original, the `builder.run_state_machine(&mut target).map_err(|_| DebugError::GdbStubInit)?` call in the `else` blocked returned `GdbStubStateMachine` on success, but `ok_or_else()` only calls the closure to produce an error value (wrapping it in `Err()`). The closure return type is also `DebugError` instead of `GdbStubStateMachine`.

We could refactor it to make it work, maybe like:

```rust
let mut gdb = match debug.gdb.take() {
    Some(gdb) => gdb,
    None => {
        /* else code */
        builder.run_state_machine(&mut target).map_err(|_| DebugError::GdbStubInit)?
    }
};
```

But it's not that much cleaner, imo.

Javagedes: Current solution in this PR works for me.

core/patina_internal_core/src/collections/sorted_slice.rs
```diff
@@ -198,7 +198,7 @@ where
     T: Debug,
 {
     fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
-        f.debug_struct("MemoryBlockSlice").field("block_count", &self.item_count).field("slice", &self.deref()).finish()
+        f.debug_struct("MemoryBlockSlice").field("block_count", &self.item_count).field("slice", &&**self).finish()
```
os-d: Some of these I would argue the readability is worse. &&** is very to understand. True also of &** below. Perhaps adding parens here would help readability.

vineelko (approved):

core/patina_stacktrace/src/aarch64/unwind.rs
```diff
@@ -622,7 +622,7 @@ impl UnwindCode {
         // The main unwind decode logic
         while i < unwind_codes.len() {
             let byte = at(i)?;
-            if (byte >> 5) & 0b111 == 0b000 {
+            if (byte >> 5).trailing_zeros() >= 3 {
```
os-d: This lint I personally don’t agree is more readable. I now have to go find out what this function is vs simple bit math. Won’t complain if it is enabled, but wouldn’t choose it myself.

os-d (approved):

core/patina_stacktrace/src/aarch64/unwind.rs
makubacki: I don't have a strong preference but could see how some might find it more natural to recognize the intention. There are only a couple of instances and the numbers being in binary help with readability. I'm fine to drop the code changes and mark it as permanently enabled.

core/patina_internal_core/src/collections/sorted_slice.rs
makubacki: Going through that lint's changes, I agree that this case is the one that sticks out a bit, partly because `field()` requires a trait object and it's in a longer chain of method calls. We could revert this and allow the explicit `.deref()` call here.

I think other cases are concise and follow normal and expected operations. In most cases, they're just doing something like `&**self` where `*self` is a normal deref to the type (`T`), `**self` is the actual `deref()` on the type, and `&` is a reference.

Does that sound okay?

os-d: That's fine with me. I was just calling out the lints that were for readability that I didn't necessarily think improved readability, but I don't have strong feelings on them.

core/patina_stacktrace/src/aarch64/unwind.rs
makubacki: Dropped the commit and set `verbose_bit_mask` as "allowed" in Cargo.toml.

core/patina_internal_core/src/collections/sorted_slice.rs
makubacki: Done

PR #1742, opened 2026-08-12
[feature/supv] Guard save state result write inside SMAP disable routine

kuqin12:
## Description

This change wraps the current raw memory writes into SMAP disable -> write memory -> reenable SMAP routine.

This will allow us to enable the SMAP from MMI entry block again.

- [x] Impacts functionality?
- [x] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This was tested on QEMU Q35 platform with artificial save state transactions.

## Integration Instructions

The CR4 value from the MMI entry block can enable SMAP bit now.

### DIFF
--- a/core/patina_internal_cpu/Cargo.toml
+++ b/core/patina_internal_cpu/Cargo.toml
@@ -20,6 +20,8 @@ spin = { workspace = true, features = ["rwlock"] }
 patina_paging = { workspace = true }
 cfg-if = { workspace = true }
 patina_stacktrace = { workspace = true }
+zerocopy = { workspace = true }
+zerocopy-derive = { workspace = true }
 
 [target.'cfg(all(target_arch="x86_64"))'.dependencies]
 patina_mtrr = { workspace = true }
--- a/core/patina_internal_cpu/src/save_state.rs
+++ b/core/patina_internal_cpu/src/save_state.rs
@@ -42,6 +42,8 @@ pub mod intel;
 
 pub mod amd;
 
+use zerocopy_derive::{Immutable, IntoBytes};
+
 /// `EFI_MM_SAVE_STATE_REGISTER` values from the PI Specification.
 ///
 /// These correspond to the registers that can be read from the SMRAM save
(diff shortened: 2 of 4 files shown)


makubacki:
Do you have other SMAP changes planned? Also, did you intentionally choose not to implement `with_user_access()` as a type that implements `Drop` and calls `enable_smap()` in the `drop()` implementation?

kuqin12:
> Do you have other SMAP changes planned? Also, did you intentionally choose not to implement `with_user_access()` as a type that implements `Drop` and calls `enable_smap()` in the `drop()` implementation?

I went through the supported syscall interface once again, this is the only instance that the supervisor will write to the user pointer directly. So no other SMAP related changes are planned. The serial port access is worked around by going through the IO port.

As per `drop` type, I looked at this earlier, the thoughts are that I would prefer the confine the usage of user buffer access as minimal as possible. In that case, either we need to manually disable SMAP before gaining the access and call `drop` after, on each usage. Alternatively, we create a function that wraps the access operation in it to trigger the out-of-scope drop. But that is not much different from the current implementation other than jumping through the hoops to program the same register. So I just kept it as is.

But please let me know if you have different thoughts.

makubacki:
> > Do you have other SMAP changes planned? Also, did you intentionally choose not to implement `with_user_access()` as a type that implements `Drop` and calls `enable_smap()` in the `drop()` implementation?
> 
> Alternatively, we create a function that wraps the access operation in it to trigger the out-of-scope drop. But that is not much different from the current implementation other than jumping through the hoops to program the same register. So I just kept it as is.
> 
> But please let me know if you have different thoughts.

The main difference to the current implementation is that in `drop()`, `enable_smap()` would be called if the given function (`access`) panics and right now it would not.

kuqin12:
> The main difference to the current implementation is that in `drop()`, `enable_smap()` would be called if the given function (`access`) panics and right now it would not.

Makes sense. This is added.

vineelko (approved):

makubacki (approved):
We probably want to look at buffer safety in more detail later, but there's been some good improvements made so far in the PR.

kuqin12:
> We probably want to look at buffer safety in more detail later, but there's been some good improvements made so far in the PR.

Agreed. The buffer access needs to be reviewed throughout the branch. We will do that when all functional blocks are in place.

PR #1743, opened 2026-08-13
[major] Add Initial Patina UEFI Services [Rebase & FF]

### LINKED ISSUE #432: [Feature]: Add Install Configuration Table service
### Feature Overview

Add a service to allow components to install configuration tables.

### Solution Overview

See above.

### Alternatives Considered

_No response_

### Urgency

Low

No maintainer feedback needed

### Anything else?

_No response_

### LINKED ISSUE #515: [Feature]: Provide a simple component that demonstrates common C module functionality
### Feature Overview

Ideally new Rust code should be written as components. Most code can be but may still need to install global structures like protocols or manipulate global stores like the PCD database. A simple component should demonstrate how to do these actions, so it is available as a reference to new component writers that also need their component to interact with C code.

### Solution Overview

Demonstate:

- Install, locate, uninstall protocol
- Hook on and produce a status code
- Get and set a UEFI variable with variable policy
- Get and set a PCD
- Register an event notification callback
- Publish an event notification
- Consume a HOB

### Alternatives Considered

_No response_

### Urgency

Low

I will implement the feature

No maintainer feedback needed

### Anything else?

_No response_

### LINKED ISSUE #1537: [Feature]: Add UEFI Services
### Feature Overview

This is a tracking issue for "UEFI Services" which are a set of Patina services for "services" defined in the UEFI Specification such as the UEFI Boot Services and UEFI Runtime Services.

Other services from the PI Specification may be included such as "DXE Services".

### Solution Overview

UEFI Services follow the Patina component model and provide Rust idiomatic interfaces to other Patina components and code. The scope of UEFI Services begins after the FFI layer and provides safe interfaces down to component usage.

The services are currently broken into the following subset:

- Event Services
- Image Services
- Protocol Services
- System Table Services

### Alternatives Considered

_No response_

### Urgency

Low

I will implement the feature

No maintainer feedback needed

### Anything else?

_No response_

makubacki:
## Description

Includes an initial merge commit from the current main branch into major.

Adds a new set of Patina component services called "UEFI Services". These services are defined in the Patina SDK, produced by the Patina DXE Core, and used by Patina components.

Main changes:

1. Add Patina UEFI Services trait and related definitions - see `sdk\patina\src\component\service\uefi_services`.
2. Produce Patina UEFI Services in the DXE Core - see `patina_dxe_core\src\uefi_services`.
3. Add Patina component samples using the services - see 
`components\patina_samples\src\component\uefi_services`.

Note: Throughout the development of this PR, other changes were added that built on the code that currently remains. Those have been removed to reduce the size of this PR, and the above set of changes are the primary set that remain.

- [x] Impacts functionality?
- [ ] Impacts security?
- [x] Breaking change?
- [x] Includes tests?
- [x] Includes documentation?

## How This Was Tested

- `cargo make all`
- QEMU ArmVirt & Q35 boot to EFI shell with Patina samples included
  - Initial component dispatch (DXE core and sample components):

    ```
    INFO - Dispatching Drivers
    INFO - Dispatching patina_dxe_core::cpu::cpu_arch_protocol::CpuArchProtocolInstaller
    INFO - InstallProtocolInterface: 26BACCB1-6F42-11D4-BCE7-0080C73C8881 @ 0x000000007a1ff800
    INFO - installed EFI_CPU_ARCH_PROTOCOL_GUID
    INFO - Dispatching patina_dxe_core::systemtables::SystemTableChecksumInstaller
    INFO - Dispatching patina_dxe_core::decompress::DecompressProtocolInstaller
    INFO - InstallProtocolInterface: D8117CFE-94A6-11D4-9A3A-0090273FC14D @ 0x000000007a977700
    INFO - Dispatching patina_adv_logger::component::AdvancedLoggerComponent<patina::peripheral::serial::uart::uart_16550::Uart16550>
    INFO - InstallProtocolInterface: 434F695C-EF26-4A12-9EBA-DDEF0097497C @ 0x000000007a200220
    INFO - Advanced Logger protocol installed.
    INFO - Dispatching patina_samples::component::uefi_services::configuration_table::ConfigurationTableSample
    INFO - Installed SampleVendorTable v1
    INFO - Read back signature "PTNA", 0 entries
    INFO - Dispatching patina_samples::component::uefi_services::driver_connect::DriverConnectSample
    INFO - Dispatching patina_samples::component::uefi_services::protocol_publisher::ProtocolPublisherSample
    INFO - InstallProtocolInterface: A1B2C3D4-E5F6-4789-ABCD-EF0123456789 @ 0x000000007ad727c8
    INFO - published SampleVendorProtocol on handle Handle(0x688cdc10f3d4e4b5)
    INFO - Dispatching patina_samples::component::uefi_services::protocol_publisher::ProtocolConsumerSample
    INFO - SampleVendorProtocol rev 0x10000 returned status 0x12345678
    INFO - Dispatching patina_samples::component::uefi_services::protocol_consumer::ProtocolConsumerSample
    INFO - with_protocol() read status is 0x12345678
    INFO - open_protocol() revision is 0x10000
    INFO - open_protocol() status is 0x12345678
    INFO - Protocol token resolved, status is 0x12345678
    INFO - SampleVendorProtocol notification: Protocol is installed on Handle(0x688cdc10f3d4e4b5)
    INFO - Dispatching patina_samples::component::uefi_services::tpl_critical_section::TplCriticalSectionSample
    INFO - Counters were kept consistent under raised TPL
    INFO - Dispatching patina_samples::component::uefi_services::end_of_dxe_protocol_consumer::EndOfDxeProtocolConsumerSample
    INFO - Registered End-of-DXE callback for SampleVendorProtocol
    INFO - Dispatching qemu_resources::q35::component::service::mm_config_provider::MmConfigurationProvider
    INFO - ACPI I/O Port Address: 0xB00F8040
    INFO - ACPI (PMBASE) I/O Port: 0x600
    INFO - Found 1 MM Communicate Region HOBs
    INFO - Dispatching qemu_resources::q35::component::service::mm_control::QemuQ35PlatformMmControl
    INFO - Dispatching patina_mm::component::sw_mmi_manager::SwMmiManager
    INFO - Dispatching patina_mm::component::communicator::MmCommunicator
    INFO - Dispatching qemu_resources::q35::component::service::mm_test::QemuQ35MmTest
    INFO - MM Supervisor Version: 0x0, Patch Level: 0x0, Max Request Level: 0x3533715F756D6571
    INFO - Dispatching patina_performance::component::protocol::MeasurementProtocolPublisher
    INFO - InstallProtocolInterface: C85D06BE-5F75-48CE-A80F-1236BA3B87B1 @ 0x000000007ad715e8
    INFO - Dispatching patina_performance::component::property::PropertyPublisher
    INFO - Dispatching patina_performance::component::fbpt::FbptPublisher
    INFO - Dispatching patina_performance::component::mm_records::MmRecordCollector
    INFO - Dispatching patina_smbios::component::provider::SmbiosProvider
    INFO - Dispatching patina_smbios::component::protocol_publisher::SmbiosProtocolPublisher
    INFO - InstallProtocolInterface: 03583FF6-CB36-4940-947E-B9B39F4AFAF7 @ 0x000000007a973800
    INFO - Dispatching qemu_resources::q35::component::service::smbios_platform::Q35SmbiosPlatform
    INFO - Dispatching patina_acpi::component::AcpiComponent
    INFO - InstallProtocolInterface: FFE06BDD-6107-46A6-7BB2-5A9C7EC5275C @ 0x000000007ad35028
    INFO - InstallProtocolInterface: 7F3C1A92-8B4E-4D2F-A6C9-3E12F4B8D7C1 @ 0x000000007ad35010
    INFO - Depex evaluation complete, scheduled 2 drivers
    ```

  - Later after the Timer Architectural protocol is installed:

    ```
    INFO - Dispatching patina_samples::component::uefi_services::overview::UefiServicesSample
    INFO - Sample crc32 = 0x7c22221c
    INFO - Dispatching patina_samples::component::uefi_services::timers::TimerSample
    INFO - Armed one-shot (50 ms) and periodic (10 ms) timers
    INFO - Logged from the periodic timer event
    INFO - Logged from the one-shot timer event
    INFO - Logged from the periodic timer event
    INFO - Logged from the periodic timer event
    INFO - Logged from the periodic timer event
    INFO - Logged from the periodic timer event
    INFO - Logged from the periodic timer event
    ```

  - End of DXE messages:

    ```
    INFO - End of DXE: sample_get_status returned 0x12345678
    INFO - Logged from the periodic timer event
    All EndOfDxe callbacks have returned successfully
    ```

  - All Patina on-system tests pass:

    ```
    INFO - adv_logger_test: Direct log message!!!
    adv_logger_test: Logged through the protocol!!!
    INFO - Patina on-system unit-test results:
      patina_acpi::integration_test::acpi_protocol_test ... ok (1 passes)
      patina_acpi::integration_test::acpi_test ... ok (1 passes)
      patina_adv_logger::integration_test::adv_logger_test ... ok (1 passes)
      patina_dxe_core::core_patina_tests::audit_tests::gcd_free_memory_merged_test ... ok (1 passes)
      patina_dxe_core::core_patina_tests::stability_tests::page_table_tests_1gb_split ... ok (4 passes)
      patina_dxe_core::core_patina_tests::stability_tests::page_table_tests_1gb_unmap_2mb_remap ... ok (4 passes)
      patina_dxe_core::core_patina_tests::stability_tests::page_table_tests_1gb_unmap_4kb_remap ... ok (4 passes)
      patina_dxe_core::core_patina_tests::stability_tests::page_table_tests_2mb_split ... ok (4 passes)
      patina_dxe_core::core_patina_tests::stability_tests::page_table_tests_2mb_unmap ... ok (4 passes)
      patina_dxe_core::memory_manager::memory_manager_allocations_test ... ok (1 passes)
      patina_dxe_core::memory_manager::memory_manager_attributes_test ... ok (1 passes)
      qemu_resources::q35::component::service::smbios_test::q35_smbios_ffi_test ... ok (1 passes)
    Patina on-system unit-test result totals: 27 passes, 0 fails
    ```
- Verified EFI shell console rendering after removing `GraphicsConsoleDxe` from QEMU platforms and using the `patina_graphics_console` component instead.

## Integration Instructions

- See individual crate documentation samples. In addition, a patina-dxe-core-qemu PR will demonstrate the changes needed to integrate and use the new components.


makubacki:
I'm working on some changes to improve unit test coverage.

makubacki:
> I'm working on some changes to improve unit test coverage.

Pushed. Also updated the `major` branch to include the clippy pedantic changes in `main` (merged at [a1f6643](https://github.com/OpenDevicePartnership/patina/commit/a1f66431b933f129b0ec75b24f198f9b106a465b)) and rebased changes against those new lints.

sdk/patina/src/component/service/uefi_services/tpl.rs
```diff
@@ -0,0 +1,205 @@
...
+///         // ... critical section runs at TPL_NOTIFY ...
+///     } // previous TPL restored here
+///
+///     // Or run a closure at a raised TPL.
+///     let value = tpl.with_raised_tpl(Tpl::Notify, || 42);
+///     assert_eq!(value, 42);
+///     Ok(())
+/// }
+/// ```
+pub trait TplServicesExt: TplServices {
```
cfernald: Did you consider just having these as default implemented function in the `TplServices` definition? The seperation is nice, but I feel like it add overhead to the callers who have to now know to use this additional trait.

sdk/patina/src/component/service/uefi_services/timing.rs
```diff
@@ -0,0 +1,138 @@
...
+/// use patina::error::Result;
+///
+/// fn entry_point(timing: Service<dyn TimingServices>) -> Result<()> {
+///     timing.stall(Duration::from_millis(10))?;
+///     Ok(())
+/// }
+/// ```
+#[cfg_attr(any(test, feature = "mockall"), automock)]
+pub trait TimingServices {
+    /// Stalls execution for at least `duration`.
```
cfernald, line 81: nit: please include the duration units int he description (and ideally the argument name)

sdk/patina/src/component/service/uefi_services/timing.rs
```diff
@@ -0,0 +1,138 @@
...
+/// use patina::component::service::{Service, uefi_services::timing::TimingServices};
+/// use patina::error::Result;
+///
+/// fn entry_point(timing: Service<dyn TimingServices>) -> Result<()> {
+///     timing.stall(Duration::from_millis(10))?;
+///     Ok(())
+/// }
+/// ```
+#[cfg_attr(any(test, feature = "mockall"), automock)]
+pub trait TimingServices {
```
cfernald, line 80: I'm wondering if this should include RT get/set time? Is there any plan to pull RT services into this model as well? Variables in particular seem necessary.

sdk/patina/src/component/service/uefi_services/timer_event.rs
```diff
@@ -0,0 +1,138 @@
...
+}
+
+/// Type-safe extension methods for [`TimerEventServices`].
+///
+/// This method accepts a plain closure and boxes it internally, so callers don't need to write
+/// `Box::new` themselves. The trait is implemented for every [`TimerEventServices`] implementor
+/// (including [`Service<dyn TimerEventServices>`]).
+///
+/// [`Service<dyn TimerEventServices>`]: crate::component::service::Service
+pub trait TimerEventServicesExt: TimerEventServices {
```
cfernald: Similar question on if this should just be default function in TimerEventServices

sdk/patina/src/component/service/uefi_services/timer_event.rs
```diff
@@ -0,0 +1,138 @@
...
+
+#[cfg(any(test, feature = "mockall"))]
+use mockall::automock;
+
+/// Describes how a timer configured with [`TimerEventServices::set_timer`] should fire.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum TimerType {
+    /// Cancels a previously configured timer.
+    Cancel,
+    /// Fires the timer once, `Duration` from now.
```
cfernald, line 31: nit: include duration units.

sdk/patina/src/component/service/uefi_services/protocol.rs
```diff
@@ -0,0 +1,589 @@
...
+/// Dropping the token does not cancel the notification. The registration is intentionally long-lived
+/// so it can outlive the component that created it.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct NotifyRegistration {
+    event: usize,
+    registration: usize,
+    context: usize,
+}
+
+impl NotifyRegistration {
```
cfernald, line 134: Here and elsewhere where we have the hidden internal function, should the impl be gated on the `core` feature, to avoiding folks calling it externally by accidednt?

sdk/patina/src/component/service/uefi_services/protocol.rs
```diff
@@ -0,0 +1,589 @@
...
+/// Unlike a `&P`, a token never dangles. It only stores a handle, so it is safe to keep for the
+/// whole boot. Each use is re-validated through [`ProtocolServicesExt::resolve`], which returns
+/// `None` if the interface is no longer installed on the handle.
+///
+/// The token identifies a specific handle. It does not detect handle reuse. For example, if the
+/// DXE Core were to recycle a handle value for a different object that also installs `P`, a stale
+/// token could resolve to that object. Since the Patina DXE Core does not recycle handle values,
+/// this is not expected to occur in practice.
+#[derive(Debug)]
+pub struct ProtocolToken<P: ProtocolInterface> {
```
cfernald, line 268: It seems `ProtocolHandle` might be a more appropriate name, since this is just a handle that is guaranteed for the generic protocol, right?

sdk/patina/src/component/service/uefi_services/protocol.rs
```diff
@@ -0,0 +1,589 @@
...
+///
+/// Returned by [`ProtocolServicesExt::on_protocol_installed`]. Keep it for as long as the
+/// notification should stay active, then pass it to [`ProtocolServicesExt::cancel`] to stop it.
+/// Dropping the token does not cancel the notification. The registration is intentionally long-lived
+/// so it can outlive the component that created it.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct NotifyRegistration {
+    event: usize,
+    registration: usize,
+    context: usize,
```
cfernald, line 131: Does context serve any purpose? It seems this is always going to come from the core as the handle, right? Or do I misunderstand what context is here?

sdk/patina/src/component/service/uefi_services/tpl.rs
cfernald: After reading more. It seems this is to allow the underlying implementation to swill bd dyn referenced, is that true?

sdk/patina/src/component/service/uefi_services/misc.rs
```diff
@@ -0,0 +1,50 @@
...
+/// fn entry_point(misc: Service<dyn MiscServices>) -> Result<()> {
+///     let checksum = misc.calculate_crc32(b"hello");
+///     log::info!("crc32 = {checksum:#x}");
+///     Ok(())
+/// }
+/// ```
+#[cfg_attr(any(test, feature = "mockall"), automock)]
+pub trait MiscServices {
+    /// Computes the CRC-32 of the given data, matching the UEFI `CalculateCrc32` boot service.
+    fn calculate_crc32(&self, data: &[u8]) -> u32;
```
cfernald: Does this really need to be a service? It seems we could just expose this through a direct function in the SDK since its not variable nor does it depend on any infrastructure.

sdk/patina/src/component/service/uefi_services/image.rs
```diff
@@ -0,0 +1,204 @@
...
+///     Ok(())
+/// }
+/// ```
+#[cfg_attr(any(test, feature = "mockall"), automock)]
+pub trait ImageServices {
+    /// Loads a UEFI image from an in-memory buffer.
+    ///
+    /// `parent` must be a valid image handle (for example, an image handle the caller already
+    /// holds). The loaded image's handle is returned. It can then be started with
+    /// [`Self::start_image`].
```
cfernald, line 97: It's probably worth calling out here motivations behind divergence from the specification arguments. e.g. `DevicePath` for `load_image`

sdk/patina/src/component/service/uefi_services/config_table.rs
```diff
@@ -0,0 +1,216 @@
...
+/// ```
+#[cfg_attr(any(test, feature = "mockall"), automock)]
+pub trait ConfigurationTableServices {
+    /// Installs or replaces the configuration table associated with `guid`.
+    ///
+    /// # Errors
+    ///
+    /// Returns [`ConfigTableError::NotFound`] if the system table is not available, or
+    /// [`ConfigTableError::OutOfResources`] if the table could not be stored.
+    fn install_table(&self, guid: BinaryGuid, table: ConfigTablePtr) -> Result<(), ConfigTableError>;
```
cfernald: I think this should be unsafe (or perhaps the creation on ConfigTablePtr) because the provider need to ensure that the memory is valid and is `'static`.

sdk/patina/src/component/service/uefi_services/config_table.rs
```diff
@@ -0,0 +1,216 @@
...
+/// that the table installed under `guid` actually has type `T`.
+pub trait ConfigurationTableServicesExt: ConfigurationTableServices {
+    /// Installs `table` under `guid`.
+    ///
+    /// The table must live for the lifetime of the configuration table entry (`'static`).
+    ///
+    /// # Errors
+    ///
+    /// Returns [`ConfigTableError::InvalidParameter`] if the table could not be installed.
+    fn install_configuration_table<T>(&self, guid: BinaryGuid, table: &'static T) -> Result<(), ConfigTableError> {
```
cfernald: Should we create a ConfigTable trait that associates a type and GUID like we do with protocols?

cfernald (commented):
just reviewed the SDK interface for now. Will review the full changes later.

sdk/patina/src/component/service.rs
```diff
@@ -348,7 +240,7 @@ impl<T: ?Sized + 'static> Service<T> {
 
 impl<T: ?Sized + 'static> From<&'static dyn Any> for Service<T> {
     fn from(value: &'static dyn Any) -> Self {
-        let s = Self::new_uninit();
+        let s = Self { value: OnceCell::new(), _marker: PhantomData };
```
Javagedes: If you pulling the `OnceCell` type logic out of `Service` and into `ServiceCell`, then you no longer need the struct internals to be wrapped in a `OnceCell` itself. You can basically revert this commit:

https://github.com/OpenDevicePartnership/patina/commit/a66694748c7a986744e9d592dc1383c2b24c4afe

sdk/patina/src/component/service/cell.rs
```diff
@@ -0,0 +1,169 @@
...
+//!
+use core::{
+    cell::OnceCell,
+    sync::atomic::{AtomicBool, Ordering},
+};
+
+/// A wait-free, write-once-then-read-many cell.
+///
+/// See the [module documentation](self) for the motivation and the reasoning behind its design.
+pub struct ServiceCell<T> {
```
Javagedes: I think the struct definition needs to be something like this:

```rust
pub struct ServiceCell<T> {
  value: OnceCell<Service<T>>,
  writing: AtomicBool
}

impl<T> ServiceCell<T> {
    /// Publishes `value` into the cell.
    ///
    /// Never blocks. If another call to `publish` is in progress, or the cell is already published, `value` is
    /// returned back to the caller immediately rather than waiting.
    ///
    /// # Errors
    ///
    /// Returns `Err(value)` if the cell was already published, or another call to `publish` is currently in
    /// progress.
    pub fn publish(&self, value: Service<T>) -> Result<(), Service<T>> {
        if self.writing.compare_exchange(false, true, Ordering::Acquire, Ordering::Relaxed).is_ok() {
            let result = self.value.set(*value);
            self.writing.store(false, Ordering::Release);
            return result;
        }
        Err(value)
    }
}
```

sdk/patina/src/component/service/cell.rs
```diff
@@ -0,0 +1,169 @@
...
+    /// `publish` is in progress.
+    pub fn get(&self) -> Option<&T> {
+        if self.writing.load(Ordering::Acquire) { None } else { self.value.get() }
+    }
+
+    /// Returns `true` if a value has been published.
+    pub fn is_published(&self) -> bool {
+        self.get().is_some()
+    }
+}
```
Javagedes, line 153: You did not add back the map functions that existed on the `Service` side. Not saying you need to, but it is missing.

sdk/patina/src/component/service/uefi_services/misc.rs
makubacki: It is mostly a service for consistency with other boot service functionality. I don't have a problem with moving this to be an SDK function though.

sdk/patina/src/component/service/uefi_services/protocol.rs
makubacki: It's used internally to hold the leaked box to the component's call back plus the registration key. That's also why it's not public, since only methods operate on it.

sdk/patina/src/component/service/uefi_services/protocol.rs
makubacki: Since it wraps a `Handle` and `Handle` continues to be used as it historically has been, I wanted to avoid potential type confusion.

sdk/patina/src/component/service/uefi_services/timer_event.rs
makubacki: Similarly updated the doc comment to mention:

> The duration can be expressed in any unit supported by [`core::time::Duration`].

sdk/patina/src/component/service/uefi_services/timing.rs
makubacki: Updated to include:

> `duration` can be expressed in any unit supported by [`core::time::Duration`].

sdk/patina/src/component/service/uefi_services/timing.rs
makubacki: Expansion is possible, but I didn't want to make any commitments until tackling runtime in detail.

sdk/patina/src/component/service/uefi_services/config_table.rs
makubacki: I'm looking into some config table handling improvements and including that.

sdk/patina/src/component/service/uefi_services/image.rs
makubacki: I'll mention that it's a simpler interface. It just sets `boot_policy` to false and does not include a device path. The device path version (`load_image_from_device_path()`) uses the `DevicePath` type which is gated behind the `unstable-device-path` feature in:

https://github.com/OpenDevicePartnership/patina/blob/c62469965bdb7c532b3427f47cca7a3afc2e11c0/sdk/patina/src/uefi/device_path.rs#L25-L27

So, it also gated.

sdk/patina/src/component/service/uefi_services/config_table.rs
makubacki: As mentioned in the other comment in the file, I might slightly rework config table APIs and will consider this in the overall flow.

sdk/patina/src/component/service/uefi_services/protocol.rs
makubacki: Going to close for now. Feel free to reopen if you'd like to discuss further.

sdk/patina/src/component/service/uefi_services/protocol.rs
makubacki: Going to close for now. Feel free to reopen if you'd like to discuss further.

patina_dxe_core/src/uefi_services/event.rs
```diff
@@ -0,0 +1,256 @@
...
+        let context = EVENT_DB.get_notification_data(efi_event).ok().and_then(|data| data.notify_context);
+
+        EVENT_DB.close_event(efi_event).map_err(EventError::from)?;
+
+        if let Some(context) = context
+            && !context.is_null()
+        {
+            // SAFETY: `context` is the `ClosureHolder` pointer created in `create_event_internal`.
+            // The event has just been closed, so no further notifications can reference it.
+            drop(unsafe { Box::from_raw(context as *mut ClosureHolder) });
```
kat-perez: `close_event()` can run from this event's own callback (the new `on_event_group_once()` helper does exactly that). In that case `notify_trampoline()` is still invoking `holder.callback`, so dropping `ClosureHolder` here destroys the closure and its captures during their own call, causing use-after-free/UB. Please defer reclamation until notification dispatch has unwound, or track in-flight callbacks, and add a self-closing-event test.

patina_dxe_core/src/uefi_services/protocol.rs
```diff
@@ -0,0 +1,449 @@
...
+        let context = registration.context();
+
+        PROTOCOL_DB.unregister_protocol_notify_events(vec![event]);
+
+        EVENT_DB.close_event(event).map_err(ProtocolError::from)?;
+
+        if !context.is_null() {
+            // SAFETY: `context` is the `NotifyHolder` created in `register_install_notify`. The
+            // event is now closed and unregistered, so no further notification can reference it.
+            drop(unsafe { Box::from_raw(context as *mut NotifyHolder) });
```
kat-perez: A future-install callback can retain its `NotifyRegistration` and cancel itself. That drops `NotifyHolder` while `notify_install_trampoline()` is still invoking the callback; after it returns, the trampoline reads `holder.registration` again from freed memory. Please make cancellation callback-safe by deferring destruction until dispatch exits and add a self-cancellation test.

sdk/patina/src/component/service/uefi_services/protocol.rs
```diff
@@ -0,0 +1,589 @@
...
+///     Ok(())
+/// }
+/// ```
+pub trait ProtocolServicesExt: ProtocolServices {
+    /// Locates the first interface for protocol `P` and returns a typed reference to it.
+    ///
+    /// # Errors
+    ///
+    /// Returns [`ProtocolError::NotFound`] if no interface for `P` is installed.
+    fn locate_protocol<P: ProtocolInterface>(&self) -> Result<&'static P, ProtocolError> {
```
kat-perez: This safe API returns `&'static P` even though this module correctly notes that protocols may later be uninstalled. `open_protocol()` and `resolve()` also only perform database lookups; they do not establish an OpenProtocol record or otherwise prevent uninstall while a reference exists. Safe component code can therefore retain/use a dangling reference. Please avoid `'static` here and either implement real open/close ownership for guards or make pointer-to-reference conversion unsafe with an explicit lifetime contract.

sdk/patina/src/component/service/uefi_services/protocol.rs
```diff
@@ -0,0 +1,589 @@
...
+#[cfg(any(test, feature = "mockall"))]
+use mockall::automock;
+
+/// An opaque pointer to a protocol interface.
+///
+/// This token is returned by the erased [`ProtocolServices::locate_interface`] method and consumed
+/// by [`ProtocolServices::install_interface`]. Components should generally use the typed methods on
+/// [`ProtocolServicesExt`] rather than handling this token directly.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct ProtocolPtr(NonNull<c_void>);
```
kat-perez: UEFI permits marker protocols with a null interface, but `NonNull` makes those impossible to install through this replacement service and causes an existing marker installation to be reported as `NotFound` (the core test explicitly documents this). Please represent null interfaces or provide explicit marker-protocol install/locate operations so the service preserves the UEFI contract.

patina_dxe_core/src/uefi_services/timing.rs
```diff
@@ -0,0 +1,96 @@
...
+/// Registered with the component dispatcher only once both the Metronome and Watchdog Timer
+/// Architectural Protocols are installed, so components depending on this service are not dispatched
+/// until stall and watchdog timer control is available.
+#[derive(IntoService)]
+#[service(dyn TimingServices)]
+pub(crate) struct CoreTimingServices;
+
+impl TimingServices for CoreTimingServices {
+    fn stall(&self, duration: core::time::Duration) -> Result<(), TimingError> {
+        core_stall(duration.as_micros() as usize).map_err(TimingError::from)
```
kat-perez: `Duration::as_micros()` floors a nonzero sub-microsecond duration to zero, contradicting the trait's “at least duration” contract, and the unchecked `u128 -> usize` cast can wrap. The timer conversion has the same issue with 100 ns units and `u64`. Please use checked ceiling conversion, return an error on overflow, and add sub-unit/overflow tests.

kat-perez (dismissed):
I found several safety and contract issues in the new service layer that should be addressed before merge. The inline comments focus on callback lifetime, protocol reference lifetime, marker protocol compatibility, and duration conversion.

components/patina_smbios/src/manager.rs
```diff
@@ -26,62 +24,36 @@ pub(crate) use record::SmbiosRecord;
...
-    boot_services: &'static StandardBootServices,
-) -> Result<efi::Handle, SmbiosError> {
-    // Create the protocol instance with internal struct
-    let internal = SmbiosProtocolInternal::new(major_version, minor_version, manager_mutex, boot_services);
-    let interface = Box::into_raw(Box::new(internal));
+    service: Service<dyn Smbios>,
+    protocols: &Service<dyn ProtocolServices>,
+) -> Result<(), SmbiosError> {
+    let internal = SmbiosProtocolInternal::new(major_version, minor_version, service);
+    let internal: &'static SmbiosProtocolInternal = Box::leak(Box::new(internal));
```
kat-perez: `SmbiosProtocolInternal` is `Box::leak`ed before `install_protocol`, so an installation failure permanently leaks the allocation and its embedded `Service<dyn Smbios>`. The previous implementation used `Box::into_raw` and reconstructed the box on the error path, and logged the failing status. Could this retain ownership until installation succeeds, and log the underlying error before mapping to `AllocationFailed`?

sdk/patina/src/component/service/uefi_services/protocol.rs
makubacki: I think it's not a problem and gated them with:
`#[cfg(any(test, feature = "core"))]`

From a practical standpoint, because all crates are built in a virtual workspace, I think feature unification from `patina_dxe_core` and `patina_internal_cpu` (that enable `core`) is going to make workspace commands, like `cargo make check`, work regardless. But it still provides incrementally better separation and will help if/when components are split outside a separate repo/build process from core components.

sdk/patina/src/component/service/uefi_services/config_table.rs
makubacki: Yeah, I think it should be `unsafe` and made the update.

sdk/patina/src/component/service/uefi_services/config_table.rs
makubacki: Added a `ConfigTable` trait and implemented it where needed (in a separate commit: "Implement the ConfigTable trait on config table types").

`DynamicConfigurationTableSample` in components/patina_samples/src/component/uefi_services/configuration_table.rs has an example.

Will be in the next PR push.

makubacki: Pushed

sdk/patina/src/component/service.rs
makubacki: Amended the changes into https://github.com/OpenDevicePartnership/patina/pull/1743/changes/30280a6c3174f1898d93a9e7f6ec2fc273b42798.

sdk/patina/src/component/service/cell.rs
makubacki: There are quite a few sites with `.get().map_or()` or `.get().map().unwrap_or_default()`, so they make sense to add for ergonomics. Will be in the next push.

sdk/patina/src/component/service.rs
makubacki: Let me know if you have any additional feedback on it. Resolving the conversation for now.

sdk/patina/src/component/service/uefi_services/misc.rs
makubacki: Added to the SDK in https://github.com/OpenDevicePartnership/patina/pull/1743/changes/6e776c948a9c3e66283b70014691d80941eeb045. Replaced `patina_dxe_core` using `crc32fast` directly in https://github.com/OpenDevicePartnership/patina/pull/1743/changes/c5cdb0a201d337b20ed9ab58b175affdaff86c91. Service removal and documentation updates amended in relevant commits.

sdk/patina/src/component/service/cell.rs
makubacki: Added and updated callers.

sdk/patina/src/component/service/cell.rs
makubacki: Just curious, is there a strong practical advantage in doing so, or more for expected type clarity? Most soundness cases I can think of are handled either way.

Javagedes: You are calling this structure a `ServiceCell` but your actual implementation has nothing to do with the `Service` struct. It is just a `OnceCell` that is `Sync + Send` by adding a safety mechanism that prevents setting internal value simultaneously and also removing the ability to get the internal value mutably.

So I think there are two options:

1. Rename this to something more generic because the implementation has nothing to do with `Service`
2. Change the implementation to be `Service` specific by forcing the inner value to be `Service<T>` rather then `T`.

makubacki: That's what I meant by type clarity. I'll update it to `OnceCell<Service<T>>` in the next push.

components/patina_smbios/src/manager.rs
makubacki: For this, I think I'm going to make it more ergonomic to pass a boxed protocol that leaks statically if installation succeeds, since this will be a common pattern. I originally made the `interface` parameter in `install_protocol` be `&'static P` to most simply ensure `'static` lifetime.

However, this can be fairly clean by giving the implementor (the DXE Core) an abstraction to commit a given protocol interface. For a `&'static P` (the currently supported case), it doesn't need to do anything. For a `Box<P>` (which is owned), it can leak it.

Then a caller can simply pass the `Box`:

```rust
    let internal = SmbiosProtocolInternal::new(major_version, minor_version, service);
    protocols.install_protocol(None, Box::new(internal)).map_err(|e| {
        log::error!("Failed to install SMBIOS protocol: {e:?}");
        SmbiosError::AllocationFailed
    })?;
    Ok(())
```

sdk/patina/src/component/service/cell.rs
makubacki: Amended the commit. In https://github.com/OpenDevicePartnership/patina/pull/1743/changes/f892554ecc5272ce041bca19bdff003091059bfa.

makubacki: Can you double check it meets expectations? Following commits were amended as needed.

components/patina_smbios/src/manager.rs
makubacki: Amended in the previous commit. Updated commit is https://github.com/OpenDevicePartnership/patina/pull/1743/changes/b8bd05a2690df552b29d8615e6f404a1805b970b.

Now just does:

```rust
pub fn install_smbios_protocol(
    major_version: u8,
    minor_version: u8,
    service: Service<dyn Smbios>,
    protocols: &Service<dyn ProtocolServices>,
) -> Result<(), SmbiosError> {
    let internal = SmbiosProtocolInternal::new(major_version, minor_version, service);

    protocols.install_protocol(None, Box::new(internal)).map_err(|e| {
        log::error!("Failed to install SMBIOS protocol: {e:?}");
        SmbiosError::AllocationFailed
    })?;

    Ok(())
}
```

makubacki:
@Javagedes, @cfernald, @kat-perez, this is ready for you review again when you get time.

Javagedes (approved):

kat-perez (approved):

sdk/patina/src/component/service/cell.rs
makubacki: @Javagedes, since I went with the approach we discussed, I'll mark this resolved for now.

makubacki:
I've decided to spend a bit more time improving open protocol and driver binding since that might impact some other interfaces and there's no point having those in an intermediate state. Will push that update within a few days.

makubacki:
> I've decided to spend a bit more time improving open protocol and driver binding since that might impact some other interfaces and there's no point having those in an intermediate state. Will push that update within a few days.

Merged main into the major branch (at commit 35f9055e) and rebased the PR onto it. The main set of changes added are in commit https://github.com/OpenDevicePartnership/patina/pull/1743/changes/59cc862c597ef0d7bc0cbfb2c86fb04515242b96.

makubacki:
> > I've decided to spend a bit more time improving open protocol and driver binding since that might impact some other interfaces and there's no point having those in an intermediate state. Will push that update within a few days.
> 
> Merged main into the major branch (at commit [35f9055](https://github.com/OpenDevicePartnership/patina/commit/35f9055eb9ac231b5404a4f08e4ae0b437b1fe39)) and rebased the PR onto it. The main set of changes added are in commit [59cc862](https://github.com/OpenDevicePartnership/patina/commit/59cc862c597ef0d7bc0cbfb2c86fb04515242b96).

I made some cumulative changes to better support UEFI Driver Model components, the PR description is updated.

The summary of new changes is a `locate_first_handle` function in Protocol Services, support for producing the EFI Component Name protocol, support for producing the EFI Driver Supported EFI Version protocol, support for a new component parameter `Protocol<P>` that allows a protocol to be a component dependency, a minimal PCD service that only abstracts access to Dynamic PCDs using an underlying PCD protocol, and a Patina component port of `GraphicsConsoleDxe` to show many of the UEFI Services changes working together.

sdk/patina/src/component/service/uefi_services/config_table.rs
```diff
@@ -0,0 +1,521 @@
...
+///
+/// static TABLE: VendorTable = VendorTable { version: 1 };
+///
+/// fn entry_point(config: Service<dyn ConfigurationTableServices>) -> Result<()> {
+///     config.install(&TABLE)?;
+///     let installed: Option<&VendorTable> = config.get::<VendorTable>();
+///     Ok(())
+/// }
+/// ```
+pub trait ConfigTable: Sized + 'static {
```
cfernald, line 271: Why not `?Sized`? A lot of config tables are DSTs.

cfernald, line 271: It seems odd for the trait to require `'static`. I think if this is a requirement to install it should be in the install function instead of the trait.

makubacki: Is there a particular reason why? From a boot timeline perspective, a config table is expected to persist through the OS transition, which we treat as `'static`. So, it accurately reflects that a UEFI configuration table must stay valid throughout that time once published.

Centralizing it on the trait avoids the lifetime being forgotten elsewhere. The functions rely on [`TypeId::of::<T>()`](https://doc.rust-lang.org/core/any/struct.TypeId.html#method.of) which requires `T` be `'static`. Functions like `get()` only return `Option<&'static T>` so the bound there can just be derived from parameters like in `install()` where a `T` parameter is present.

I think it also simplifies tracking what's expected versus manually coordinating/managing it in individual functions.

makubacki: I went with a different approach than an actual DST as I'm not sure it would work here if following the Config Table ABI in the UEFI Spec (GUID and pointer pairs). An example is [`SampleDynamicTable`](https://github.com/makubacki/patina/blob/89f426b4ab90662e6f22ee12021f5a71f1229012/components/patina_samples/src/component/uefi_services/configuration_table.rs#L122).

I thought that gets most of the practical value for a header + trailing data.

makubacki:
I've made some additional changes to handle pool allocations returned from C code, split GOP-specific logic out to a dedicated component/service and move the default font package into its own component (out of `patina_graphics_console`), so it, along with other font packages, can be independently managed from the graphics console component. Will push these updates soon. I've also started breaking out standalone changes into PRs into `main` to start reducing this PR to incremental changes that need to go to `major`. When that's done, I'll merge `main` into `major` and rebase this PR, dropping those commits.

makubacki:
This branch has accumulated several months worth of changes now. I'm going to remove everything except the initial set of commits to add the UEFI Services themselves and break the remaining changes up into other PRs for more focused review. The remaining set of changes (and some commits after) were already reviewed and approved, so I'll complete this PR after CI passes. There might be some redundant review in the next few PRs created, but a second look doesn't hurt.

sdk/patina/src/component/service/uefi_services/config_table.rs
cfernald: I had a misunderstanding about the implications of inheriting `'static` thinking it added requirements to the lifetime of any struct associated with it. I think that your reasoning seems fair. Will resolve.

cfernald: Makes sense. Thanks!

patina_dxe_core/src/uefi_services/event.rs
cfernald, line 119: Couldn't this be the closure that called `close_event`?

patina_dxe_core/src/uefi_services/config_table.rs
```diff
@@ -0,0 +1,326 @@
...
+impl ConfigurationTableServices for CoreConfigurationTableServices {
+    unsafe fn install_table(&self, guid: BinaryGuid, table: ConfigTablePtr) -> Result<(), ConfigTableError> {
+        let mut st_guard = SYSTEM_TABLE.lock();
+        let st = st_guard.as_mut().ok_or(ConfigTableError::NotFound)?;
+        core_install_configuration_table(guid.into_inner(), table.as_raw(), st)
+            .map(|_| ())
+            .map_err(ConfigTableError::from)
+    }
+
+    fn remove_table(&self, guid: BinaryGuid) -> Result<(), ConfigTableError> {
```
cfernald, line 46: I think this routine needs to also try `CONFIG_TABLE_TYPES.lock().remove(&guid);` to avoid potential type aliasing.

cfernald (commented):
looks like I missed the window a bit, but a few comments.

PR #1744, opened 2026-08-13
SDK: BinaryGuid: Use Guid's Debug Impl

os-d:
## Description

Currently, BinaryGuid's Display impl uses Guid's Display impl, which is also used by Guid's Debug impl, because splitting out the fields of a GUID is often worse for the reader who wants to see a canonical form GUID for easy visual reading, copying, etc.

However, BinaryGuid derived its Debug impl, which leads to all the fields being printed which is both actively hard to read and doesn't use CRLF, making it very hard to read on some terminal emulators.

This simply makes both scenarios better by having BinaryGuid use Guid's Debug impl to get a better print.

Before this change, print a single GUID looked like this:

image

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

The unit test; on a terminal emulator that doesn't auto CRLF.

## Integration Instructions

N/A.

### DIFF
--- a/sdk/patina/src/base/guid.rs
+++ b/sdk/patina/src/base/guid.rs
@@ -202,7 +202,7 @@ pub type OwnedGuid = Guid<'static>;
 /// println!("Header GUID: {}", header.guid.as_guid());
 /// ```
 #[repr(transparent)]
-#[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub struct BinaryGuid(pub efi::Guid);
 
 impl BinaryGuid {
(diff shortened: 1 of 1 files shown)


vineelko (approved):

sdk/patina/src/base/guid.rs
```diff
@@ -1529,7 +1536,7 @@ mod tests {
         let debug_string = format!("{:?}", binary_guid);
 
         // Debug should show the full type and inner value
-        assert!(debug_string.contains("BinaryGuid"));
+        assert_eq!(debug_string, "550E8400-E29B-41D4-A716-446655440000");
```
makubacki: Does this work?

```suggestion
        assert_eq!(debug_string, TEST_GUID_STRING_UPPER);
```

sdk/patina/src/base/guid.rs
```diff
@@ -1529,7 +1536,7 @@ mod tests {
         let debug_string = format!("{:?}", binary_guid);
 
         // Debug should show the full type and inner value
```
makubacki: Can you update the comment? For example: "full type and inner value" -> "defers to Guid's Debug impl".

makubacki (approved):

sdk/patina/src/base/guid.rs
os-d: Thanks, I thought I had.

PR #1745, opened 2026-08-14
deny.toml: Ban direct dependecies on r-efi outside the SDK crate [Rebase & FF]

makubacki:
## Description

`r_efi::efi` is re-exported from `patina::standard::efi`, with the guidance that code should depend on the re-export.

While r-efi is only an explicit crate dependency of the `patina` crate today, it could inadvertently be added as a direct dependency of other crates in the repo in the future.

This change adds a deny rule to the `deny.toml` file to prevent direct dependencies on `r-efi` outside of the SDK crate.

This also allows tracking of transitive dependencies on `r-efi`.

Both `patina` and `getrandom` are allowed to depend on `r-efi` as `getrandom` does so today used as follows by `patina_internal_core`:

```
warning[unmatched-wrapper]: direct parent 'getrandom = 0.4.3' of
banned crate 'r-efi = 6.0.0' was not marked as a wrapper
    ┌─ deny.toml:146:16
    │
146 │     { crate = "r-efi", reason = "Depend on `patina` and use `patina::standard::efi` instead of r-efi directly.", wrappers = ["patina"] }
    │                ━━━━━             ───────────────────────────────────────────────────────────────────────────── reason
    │                │
    │                banned here
    │
    ├ r-efi v6.0.0
      └── getrandom v0.4.3
          └── rand v0.10.2
              └── (dev) patina_internal_core v23.0.2
                  └── patina_dxe_core v23.0.2
                      ├── (dev) patina_adv_logger v23.0.2
```

A small touchup to arch.rs is made to use the re-exported path.

---

Also includes a commit to remove a file referencing `r_efi` that is not included in the crate:

**patina_internal_cpu: Delete orphaned file (cpu.rs)**

This file is an orphaned version of:

`patina_dxe_core\src\cpu\efi_cpu\x64\cpu.rs`

That was accidentally left in place during earlier refactors of the
CPU code. It also follows some practices like directly referencing
`r-efi` that we want to avoid in favor of using `patina::standard::efi`
so it is being removed to avoid confusion.

Refactor references:

- Moved to patina_dxe_core in 3d598615fe0334d59e04f17b2320b8a56fc7bdb6
- Accidentally added back in ac11069630ece945a4a3368e842fff7fb59e39e3

---

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- `cargo make deny` before and after the deny.toml change

## Integration Instructions

- N/A


vineelko (approved):

apop5 (approved):

os-d (approved):

makubacki:
Pushed https://github.com/OpenDevicePartnership/patina/pull/1745/changes/4891cac91f1e13efe806f2bb70687f16bbe238bb to remove a file that is not caught by cargo-deny because it is orphaned. Included in this PR because it imports `r_efi`.

PR #1747, opened 2026-08-15
fv_read_section_efiapi: adjust data pointer

### LINKED ISSUE #1746: [Bug]: fv_read_section_efiapi returns wrong data when called for EFI_SECTION_FREEFORM_SUBTYPE_GUID (0x18) section

### Current Behavior

Function returns the subtype GUID header bytes followed by the payload, instead of returning only the payload.

### Expected Behavior

PI spec says on exit the buffer is a "Pointer to a pointer to a buffer in which the section contents are returned, not including the section header.". This type of section data starts with the GUID, we expect it to point to GUID, not passed it.

### Steps To Reproduce

1. Have a EFI_SECTION_FREEFORM_SUBTYPE_GUID section in the FV.
2. Call
            Status = FwVolumeProtocol->ReadSection(
                    FwVolumeProtocol,
                    FileGuid,
                    SectionType,
                    Instance,
                    &Data,
                    &DataSize,
                    &Authentication
            );
3. Status returns EFI_SUCCESS, DataSize is correct
4. Data points to the section_data + sizeof(EFI_GUID)

### Build Environment

```markdown
- OS(s): N/A
- Targets Impacted: an FV with EFI_SECTION_FREEFORM_SUBTYPE_GUID section.
```

### Version Information

```text
Commit: current
```

### Urgency

Medium

I will fix it

Maintainer feedback requested

### Anything else?

_No response_

olegilyasov:
## Description
The fix changes fv_read_section_efiapi so FREEFORM_SUBTYPE_GUID returns the subtype GUID header bytes followed by the payload, instead of returning only the payload.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested
Added test_fv_read_section_freeform_subtype_returns_guid_plus_payload

## Integration Instructions
N/A

### DIFF
--- a/patina_dxe_core/src/pi_dispatcher/fv.rs
+++ b/patina_dxe_core/src/pi_dispatcher/fv.rs
@@ -8,7 +8,7 @@
 //!
 use core::{ffi::c_void, mem::size_of, num::NonZeroUsize, ptr::NonNull, slice};
 
-use alloc::{boxed::Box, collections::BTreeMap};
+use alloc::{boxed::Box, collections::BTreeMap, vec::Vec};
 use patina::{
     pi::{
         self,
@@ -749,9 +749,23 @@ impl<P: PlatformInfo> FvProtocolData<P> {
             Ok(section) => section,
         };
 
-        let section_data = match section.try_content_as_slice() {
+        let section_content = match section.try_content_as_slice() {
             Ok(data) => data,
             Err(err) => return err.into(),
+        };
+        let owned_section_data: Vec<u8>;
+        let section_data = match section.header() {
+            patina_ffs::section::SectionHeader::FreeFormSubtypeGuid(header, _) => {
+                // EFI_SECTION_FREEFORM_SUBTYPE_GUID data returned by ReadSection includes the
+                // subtype GUID header followed by the payload, not just the payload bytes.
+                let header_bytes =
+                    // SAFETY: header is a valid reference to a SectionHeader::FreeFormSubtypeGuid,
+                    // and we are creating a slice of its bytes.
+                    unsafe { slice::from_raw_parts(core::ptr::from_ref(header).cast::<u8>(), core::mem::size_of_val(header)) };
+                owned_section_data = header_bytes.iter().chain(section_content.iter()).copied().collect();
+                owned_section_data.as_slice()
+            }
+            _ => section_content,
         };
 
         // get the buffer_size and buffer parameters from caller.
@@ -902,8 +916,10 @@ mod tests {
     use crate::{MockComponentInfo, MockCpuInfo, MockMemoryInfo, test_support};
     use patina::pi::{
         BootMode,
+        fw_fs::{fv::BlockMapEntry, fvb, ffs::section::header::FreeformSubtypeGuid},
         hob::{self, Hob, HobList},
     };
+    use patina_ffs::{file::File as FfsFile, section::{Section, SectionHeader}, volume::Volume};
     use patina_ffs_extractors::CompositeSectionExtractor;
     extern crate alloc;
     use crate::test_collateral;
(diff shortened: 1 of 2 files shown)


os-d:
I accidentally clicked the copilot review button and now I'm not sure how to get it to stop, my apologies.

patina_dxe_core/src/pi_dispatcher/fv.rs
```diff
@@ -749,10 +749,24 @@ impl<P: PlatformInfo> FvProtocolData<P> {
...
         };
 
-        let section_data = match section.try_content_as_slice() {
+        let section_content = match section.try_content_as_slice() {
             Ok(data) => data,
             Err(err) => return err.into(),
         };
+        let owned_section_data: Vec<u8>;
+        let section_data = match section.header() {
+            patina_ffs::section::SectionHeader::FreeFormSubtypeGuid(header, _) => {
```
joschock: I think this same issue applies to `Compression`, `GuidDefined`, `Version` as well. It might be better to add a new API on `Section` that returns the full content needed here and keeps the byte slice construction located inside the existing section implementation that is doing this sort of thing already.

joschock (changes requested):

patina_dxe_core/src/pi_dispatcher/fv.rs
```diff
@@ -749,10 +749,24 @@ impl<P: PlatformInfo> FvProtocolData<P> {
...
+                // EFI_SECTION_FREEFORM_SUBTYPE_GUID data returned by ReadSection includes the
+                // subtype GUID header followed by the payload, not just the payload bytes.
+                let header_bytes =
+                    // SAFETY: header is a valid reference to a SectionHeader::FreeFormSubtypeGuid,
+                    // and we are creating a slice of its bytes.
+                    unsafe { slice::from_raw_parts(core::ptr::from_ref(header).cast::<u8>(), core::mem::size_of_val(header)) };
```
makubacki: Extracting the headers through an unsafe `slice::from_raw_parts()` call is already done for `SectionHeader::FreeFormSubtypeGuid` here:

https://github.com/OpenDevicePartnership/patina/blob/f097fb1701393acbe6f42c2ab5d333ed4610b6d0/sdk/patina_ffs/src/section.rs#L179-L185

We could consider adding a helper method onto the `SectionHeader` type (e.g. `type_specific_header_bytes()`) that could be used in both places then we don't need multiple unsafe calls to do the same thing.

Also, feel to ignore if you'd like since this involves refactoring outside the main change here.

patina_dxe_core/src/pi_dispatcher/fv.rs
```diff
@@ -1889,4 +1905,111 @@ mod tests {
...
+                0,
+                &raw mut returned_ptr,
+                &raw mut returned_size,
+                &raw mut auth_status,
+            );
+
+            assert_eq!(status, efi::Status::SUCCESS);
+            assert_eq!(returned_size, expected_size);
+            assert_eq!(&returned[..16], subtype_guid.into_inner().as_bytes());
+            assert_eq!(&returned[16..returned_size], payload.as_slice());
```
makubacki: nit: It would be cleaner if a named constant was used for the GUID size instead of `16`.

patina_dxe_core/src/pi_dispatcher/fv.rs
makubacki: I was going to leave a comment on Line 764, but I'll just add it here since it is related to this point.

---

Extracting the headers through an unsafe `slice::from_raw_parts()` call is already done for `SectionHeader::FreeFormSubtypeGuid` here:

https://github.com/OpenDevicePartnership/patina/blob/f097fb1701393acbe6f42c2ab5d333ed4610b6d0/sdk/patina_ffs/src/section.rs#L179-L185

We could consider adding a helper method onto the `SectionHeader` type (e.g. `type_specific_header_bytes()`) that could be used in both places then we don't need multiple unsafe calls to do the same thing.

joschock (approved):

patina_dxe_core/src/pi_dispatcher/fv.rs
olegilyasov: Done.

makubacki (approved):

patina_dxe_core/src/pi_dispatcher/fv.rs
makubacki: Closing. Thanks for the update.

makubacki:
@olegilyasov, are you okay if this PR is squashed as a single commit? We can merge either way, but if you'd like separate commits, they'll need to be refactored so later commits are additive, and the changes are more logically grouped.

olegilyasov:
> @olegilyasov, are you okay if this PR is squashed as a single commit? We can merge either way, but if you'd like separate commits, they'll need to be refactored so later commits are additive, and the changes are more logically grouped.

Single commit is totally fine, thank you.

PR #1749, opened 2026-08-19
[feature/supv] Populate processor ID instead of CPU index into `cpu_manager`

kuqin12:
## Description

This change updates the hob parser to populate the processor ID into the `cpu_manager` static. This will allow the save_state module to query processor ID without going through the access into HOB buffer, which is marked as user RO during runtime.

This change also promotes some critical prints to info level.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This was tested on QEMU Q35.

## Integration Instructions

N/A

### DIFF
--- a/patina_mm_supervisor/src/cpu.rs
+++ b/patina_mm_supervisor/src/cpu.rs
@@ -15,7 +15,7 @@
 //! SPDX-License-Identifier: Apache-2.0
 //!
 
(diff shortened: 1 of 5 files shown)


vineelko (approved):
:shipit:

PR #1751, opened 2026-08-20
GCD: Add Better Logs on InvalidStateTransition

os-d:
## Description

Currently, invalid state transitions in the memory block layer all return the same error code and the GCD just prints/asserts that an invalid state transition occurred, which is not easily debuggable.

This adds prints to the memory block layer to identify why an invalid state transition occurred. Allocate is expected to fail when testing regions to allocate from, so a print is not added there.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Printing on a system where an invalid state transition was occurring.

## Integration Instructions

N/A.

### DIFF
--- a/patina_dxe_core/src/gcd/memory_block.rs
+++ b/patina_dxe_core/src/gcd/memory_block.rs
@@ -219,7 +219,13 @@ impl MemoryBlock {
                 md.attributes = attributes;
                 Ok(())
             }
-            _ => Err(Error::InvalidStateTransition),
+            _ => {
+                log::error!(
+                    "Invalid GCD state transition: Add({memory_type:?}) requires an unallocated NonExistent block and \
+                     a non-NonExistent target type. Block: {self:#x?}"
+                );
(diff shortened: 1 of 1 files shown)


patina_dxe_core/src/gcd/memory_block.rs
```diff
@@ -273,7 +285,12 @@ impl MemoryBlock {
                 *self = Self::Unallocated(*md);
                 Ok(())
             }
-            _ => Err(Error::InvalidStateTransition),
+            _ => {
+                log::error!(
+                    "Invalid GCD state transition: Free requires an allocated, non-NonExistent block. Block: {self:#x?}"
```
joschock: super nit: double negatives are weird. Consider `Existent`?

joschock (approved):

vineelko (approved):

makubacki (approved):

PR #1753, opened 2026-08-21
Patina: Gate architecture specific code on UEFI targets

vineelko:
## Description

Patina: Gate architecture specific code on UEFI targets

Following the `patina-paging` [PR #223](https://github.com/OpenDevicePartnership/patina-paging/pull/223), apply a similar change to the
Patina repo. This ensures architecture specific code is properly
guarded and only compiled for the appropriate UEFI targets.

Note that some external crates intentionally use inline assembly in host
builds:

- `corosensei` includes `global_asm!` in `x86_64_windows.rs` conditioned
  on `cfg(windows and x86_64)`.
- `uart_16550` includes a PIO backend that uses inline assembly
  conditioned on `cfg(x86_64)`.

So this commit is a best effort to keep Patina code free of inline
assembly when building for host targets.

NOTE: This PR only addresses the conversion of `cfg(not(test))` to
`cfg(target_os = "uefi")`. There are still many raw asm! blocks that
are not flagged by the internal toolchain, so they are being left as-is.

NIT: Update to the latest MTRR version due to coverage attribute
changes.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

`cargo make all` with public and internal toolchains(excluding above exceptions).

## Integration Instructions

NA

### DIFF
--- a/core/patina_internal_cpu/src/interrupts/aarch64/interrupt_manager.rs
+++ b/core/patina_internal_cpu/src/interrupts/aarch64/interrupt_manager.rs
@@ -18,7 +18,7 @@ use crate::interrupts::{
 };
 
 cfg_if::cfg_if! {
-    if #[cfg(not(test))] {
+    if #[cfg(target_os = "uefi")] {
         use core::arch::global_asm;
         use patina::{read_sysreg, write_sysreg, arch::aarch64::{AArch64El, get_current_el}};
 
@@ -64,7 +64,7 @@ impl InterruptManager for InterruptsAarch64 {}
 #[cfg_attr(coverage, coverage(off))]
 fn enable_fiq() {
     cfg_if::cfg_if! {
-        if #[cfg(not(test))]  {
+        if #[cfg(target_os = "uefi")]  {
             write_sysreg!(reg daifclr, imm 0x01, "isb sy");
         } else {
             unimplemented!()
@@ -75,7 +75,7 @@ fn enable_fiq() {
 #[cfg_attr(coverage, coverage(off))]
 fn disable_fiq() {
     cfg_if::cfg_if! {
-        if #[cfg(not(test))]  {
+        if #[cfg(target_os = "uefi")]  {
             write_sysreg!(reg daifset, imm 0x01, "isb sy");
         } else {
             unimplemented!()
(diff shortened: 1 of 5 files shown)


os-d (approved):

core/patina_stacktrace/src/stacktrace.rs
```diff
@@ -142,50 +139,49 @@ impl StackTrace {
...
-                    asm!(
+                    core::arch::asm!(
                         "adr {pc}, .",   // Get current PC (program counter)
                         "mov {sp}, sp",  // Get current SP (stack pointer)
                         "mov {fp}, x29", // Get current FP (frame pointer)
                         pc = out(reg) stack_frame.pc,
                         sp = out(reg) stack_frame.sp,
                         fp = out(reg) stack_frame.fp,
                     );
+                    StackTrace::dump_with(stack_frame)
```
makubacki, line 162: Can you update the safety comment to include the `dump_with()` call or move it outside the `unsafe` block?

kuqin12 (approved):

core/patina_internal_cpu/src/interrupts/aarch64/interrupt_manager.rs
makubacki, line 1: Is `initialize_fpu()` intentionally excluded?

https://github.com/OpenDevicePartnership/patina/blob/c3ed504f72b0058712b2bd1888b641e7c16b21d2/patina_dxe_core/src/cpu/efi_cpu/x64/cpu.rs#L44-L45

makubacki (approved):

core/patina_internal_cpu/src/interrupts/aarch64/interrupt_manager.rs
vineelko: I wonder why it didn't catch during my testing. Let me double check and fix it.

vineelko: I still couldn't find out why this and bunch of other asm! blocks are being not getting triggered with internal toolchain. May be looks like they are getting optimized away or something. Let me update the PR description to call this out. But will mark this block with `cfg(target_os="uefi")`.

core/patina_stacktrace/src/stacktrace.rs
vineelko: Fixed it.

PR #1756, opened 2026-08-25
Events with EVT_NOTIFY_SIGNAL type may fail

### LINKED ISSUE #1755: [Bug]: Events with EVT_NOTIFY_SIGNAL type processing may fail

### Current Behavior

When creating an event with EVT_NOTIFY_SIGNAL type, the code requires notify_function to be Some. But according to UEFI spec 2.10 section 7.1.2:

- Events with event groups CAN have notify functions (optional)
- Events without event groups that are notifiable (EVT_NOTIFY_*) MUST have notify functions
- TPL validation only applies when notify_function is provided

Patina doesn't handle this special case.

### Expected Behavior

Current event implementation (event_db.rs) returns an error when:
- event is notifiable && notify_function.is_none()

Expected return on error when:
- event is notifiable && event_group.is_none() && notify_function.is_none()

Validation of TPL is only applicable when notify function is provided.

### Steps To Reproduce

Event group member WITH or WITHOUT notify function should succeed, it does not currently succeed if notify function is not provided.

### Build Environment

```markdown
Any
```

### Version Information

```text
Current commit.
```

### Urgency

Medium

I will fix it

Maintainer feedback requested

### Anything else?

proposed fix in event_db.rs:
```
-        if notifiable {
-            if notify_function.is_none() {
-                return Err(EfiError::InvalidParameter);
-            }
+        // For notifiable events without event groups, notify_function is required
+        // For events with event groups, notify_function is optional per UEFI spec 2.10 Section 7.1.2
+        if notifiable && event_group.is_none() && notify_function.is_none() {
+        }

+        // Validate notify_tpl if notify_function is provided
+        if notify_function.is_some() {

```

olegilyasov:
## Description
- Events with event groups CAN have notify functions (optional)
- Events without event groups that are notifiable (EVT_NOTIFY_*) MUST have notify functions
- TPL validation only applies when notify_function is provided

Modify event handling for this special case.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Added test function `create_event_group_members_with_optional_notify_functions`

## Integration Instructions

N/A

### DIFF
--- a/patina_dxe_core/src/event_db.rs
+++ b/patina_dxe_core/src/event_db.rs
@@ -255,11 +255,14 @@ impl Event {
         let notifiable = (event_type & (efi::EVT_NOTIFY_SIGNAL | efi::EVT_NOTIFY_WAIT)) != 0;
         let event_type: EventType = event_type.try_into()?;
 
-        if notifiable {
-            if notify_function.is_none() {
-                return Err(EfiError::InvalidParameter);
-            }
+        // For notifiable events without event groups, notify_function is required
+        // For events with event groups, notify_function is optional per UEFI spec 2.10 Section 7.1.2
+        if notifiable && event_group.is_none() && notify_function.is_none() {
+            return Err(EfiError::InvalidParameter);
+        }
 
+        // Validate notify_tpl if notify_function is provided
+        if notify_function.is_some() {
             // Pedantic check; this will probably not work with "real firmware", so
             // loosen up a bit.
             // match notify_tpl {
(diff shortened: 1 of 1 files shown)


os-d:
@olegilyasov can you share the spec language directly that you are referring to? In Uefi Spec 2.10A section 7.1.2 I am only seeing this as relevant:

image

Which is what the code is currently doing. I'm not seeing the language about event groups with optional notify functions.

It looks like edk2 has the same behavior: https://github.com/tianocore/edk2/blob/7735ed4f8eb3afaf1a97d27a8862f382c387f0bc/MdeModulePkg/Core/Dxe/Event/Event.c#L429-L438

olegilyasov:
@os-d you are correct, I incorrectly assumed UEFI spec allowed event group members to have **optional** notify functions. UEFI spec requires NotifyFunction to be non-NULL for notifiable events, even for event groups.
The real bug: Patina's pi_dispatcher.rs is creating the event with `notify_function: None`, which violates the UEFI spec.
EDK2 provides EfiEventEmptyFunction - an empty stub function that does nothing. It uses this for events that only need to signal event groups but don't need custom notification logic.

Undoing the change and adding a notification function stub to pi_dispatcher....

joschock:
> @os-d you are correct, I incorrectly assumed UEFI spec allowed event group members to have **optional** notify functions. UEFI spec requires NotifyFunction to be non-NULL for notifiable events, even for event groups. The real bug: Patina's pi_dispatcher.rs is creating the event with `notify_function: None`, which violates the UEFI spec. EDK2 provides EfiEventEmptyFunction - an empty stub function that does nothing. It uses this for events that only need to signal event groups but don't need custom notification logic.
> 
> Undoing the change and adding a notification function stub to pi_dispatcher....

I looked at pi_dispatcher; both create_event calls I see in there provide `Some(notificaton_fn)` - can you point more specifically to what you are concerned about in pi_dispatcher?

the two event creations I see in pi_dispatcher are: 
https://github.com/OpenDevicePartnership/patina/blob/a334b36586da55b499a1f7be2738c69413e58f59/patina_dxe_core/src/pi_dispatcher.rs#L190
and 
https://github.com/OpenDevicePartnership/patina/blob/a334b36586da55b499a1f7be2738c69413e58f59/patina_dxe_core/src/pi_dispatcher.rs#L176

olegilyasov:
Apparently the problem is in my code. I added signaling EVENT_GROUP_DXE_DISPATCH after each DXE dispatch round where at least one driver was dispatched, allowing SMM dispatcher and other components to coordinate with DXE driver dispatch.
This allows SMM drivers to get loaded.
And of course I messed up the event preparation. No intention to push that change since Patina does not support SMI handlers.
Sorry about this and thank you for your time.

makubacki:
> Apparently the problem is in my code. I added signaling EVENT_GROUP_DXE_DISPATCH after each DXE dispatch round where at least one driver was dispatched, allowing SMM dispatcher and other components to coordinate with DXE driver dispatch. This allows SMM drivers to get loaded. And of course I messed up the event preparation. No intention to push that change since Patina does not support SMI handlers. Sorry about this and thank you for your time.

For clarity to others that might come across this, Patina supports the CPU operating mode called System Management Mode (SMM) and the ability to run what is colloquially referred to as "SMI handlers". At the Platform Initialization (PI) spec (software) level, there are two modes defined - Traditional SMM and Standalone MM, which leads to terminology across architectures and modes simplifying to "MM" and "MMI". Patina intentionally does not support Traditional SMM for the reasons noted in [Patina DXE Core Requirements - No Traditional SMM](https://opendevicepartnership.github.io/patina/integrate/patina_dxe_core_requirements.html#11-no-traditional-smm). However, it does support MMI handlers, just running in the Standalone MM model (in which, the actual handler code is usually the same).

PR #1758, opened 2026-08-28
Dxe Memory Bin: Fix Peak Usage Tracking

### LINKED ISSUE #1757: [Bug]: Memory bucket "Current Pages" don't reflect when less than the reserved bucket size is used

### Current Behavior

I'm not sure if this is a bug necessarily, but I wanted to highlight this behavior that's different from the EDK2 DXE core.

On my platform with a Patina build, the memory type information as retrieved via `EfiGetSystemConfigurationTable` with `gEfiMemoryTypeInformationGuid` in [BmMisc.c](https://github.com/microsoft/mu_basecore/blob/f44f77a3fee3a0e56f98ae864701b74b3dcf3eb1/MdeModulePkg/Library/UefiBootManagerLib/BmMisc.c#L276) shows the following - note that "Previous" and "Current" are equal for all entries:
```
15:26:25.0062429 - Memory  Previous  Current   Minimum    Next 
15:26:25.0063337 -  Type    Pages     Pages     Pages     Pages 
15:26:25.0064245 - ======  ========  ========  ========  ======= 
15:26:25.0065296 -   09    0000035B  0000035B  00000000  0000035B 
15:26:25.0066138 -   0A    00000090  00000090  00000000  00000090 
15:26:25.0067036 -   00    00000514  00000514  00000000  00000514 
15:26:25.0068198 -   06    00000540  00000540  00000000  00000540 
15:26:25.0069156 -   05    000004C8  000004C8  00000000  000004C8 
```

### Expected Behavior

When swapping the same platform build to the EDK2 DXE core, the memory type information prints as:
```
15:33:10.6784043 - Memory  Previous  Current   Minimum    Next 
15:33:10.6784917 -  Type    Pages     Pages     Pages     Pages 
15:33:10.6786380 - ======  ========  ========  ========  ======= 
15:33:10.6787304 -   09    0000035B  00000071  00000000  0000035B 
15:33:10.6886002 -   0A    00000090  00000050  00000000  00000090 
15:33:10.6887409 -   00    00000514  00000000  00000000  00000514 
15:33:10.6888831 -   06    00000540  00000380  00000000  00000540 
15:33:10.6890891 -   05    000004C8  00000340  00000000  000004C8 
```

This is helpful to the platform so we know when we can shrink the memory buckets if the used page count is well below the number reserved.

### Steps To Reproduce

Set one of the types in a memory type information hob to a large size (e.g. 100MB), observe that the "current" value reported from the system configuration table matches the size in the HOB.

### Build Environment

```markdown
- OS(s):
- Targets Impacted:
```

### Version Information

```text
23.1.0
```

### Urgency

Low

Someone else needs to fix it

Maintainer feedback requested

### Anything else?

_No response_

os-d:
## Description

Currently, patina is copying the memory type information HOB values in as the peak usage values in the config table. As a result, the config table is not tracking real usage, only configured usage.

This commit fixes that by dropping the copy. This requires initializing the table separately, since it was relying on getting a copy from the HOB. This approach follows edk2 and has the stats table and the info table match each other, indexable by memory type in both. The stats table only has its information_index set if the given type has a memory bin associated with it.

Closes #1757 

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on Q35. Prior to this change, BdsDxe always reported the HOB values as the current pages. After this change, it reports the accurate peak page usage.

## Integration Instructions

This is not a breaking change, but it may allow platforms to lower the bin sizes currently configured. A reboot may occur if the PCD is set to reboot on memory type changes.

### DIFF
--- a/patina_dxe_core/src/memory_bin.rs
+++ b/patina_dxe_core/src/memory_bin.rs
@@ -131,7 +131,7 @@ impl MemoryBinStatistics {
 /// Default `MemoryBinStatistics` initialization for all memory types.
 ///
 /// Indexed by `efi::MemoryType` value. Matches edk2's `mMemoryTypeStatistics` initialization.
-const DEFAULT_STATISTICS: [MemoryBinStatistics; EFI_MAX_MEMORY_TYPE + 1] = [
+const DEFAULT_STATISTICS: [MemoryBinStatistics; EFI_MAX_MEMORY_TYPE] = [
     MemoryBinStatistics::new(true, false),  // EfiReservedMemoryType (0)
     MemoryBinStatistics::new(false, false), // EfiLoaderCode (1)
     MemoryBinStatistics::new(false, false), // EfiLoaderData (2)
@@ -148,7 +148,6 @@ const DEFAULT_STATISTICS: [MemoryBinStatistics; EFI_MAX_MEMORY_TYPE + 1] = [
     MemoryBinStatistics::new(true, true),   // EfiPalCode (13)
     MemoryBinStatistics::new(false, false), // EfiPersistentMemory (14)
     MemoryBinStatistics::new(true, false),  // EfiUnacceptedMemoryType (15)
-    MemoryBinStatistics::new(false, false), // EfiMaxMemoryType sentinel (16)
 ];
 
 /// Manages memory bins for hibernate resume stability.
@@ -156,25 +155,33 @@ const DEFAULT_STATISTICS: [MemoryBinStatistics; EFI_MAX_MEMORY_TYPE + 1] = [
 /// The `MemoryBinManager` tracks per-memory-type bin regions and allocation statistics.
 pub(crate) struct MemoryBinManager {
     /// Per-memory-type bin statistics, indexed by `efi::MemoryType`.
-    statistics: [MemoryBinStatistics; EFI_MAX_MEMORY_TYPE + 1],
+    statistics: [MemoryBinStatistics; EFI_MAX_MEMORY_TYPE],
     /// Current memory type information with peak usage tracking for the BDS config table.
     /// This is a fixed-size array so that raw pointers to it remain valid for the
     /// lifetime of the static `MEMORY_BIN_MANAGER`.
     memory_type_information: [EFiMemoryTypeInformation; MAX_MEMORY_TYPE_INFO_ENTRIES],
-    /// Number of valid entries in `memory_type_information`.
-    memory_type_information_count: usize,
     /// Whether bins have been initialized.
     initialized: bool,
 }
 
 impl MemoryBinManager {
+    // get_mut is not const, so it can't be used here. This is safe because we are
+    // initializing the array and only traversing the length it returns.
+    #[allow(clippy::indexing_slicing)]
     /// Creates a new uninitialized `MemoryBinManager`.
     pub(crate) const fn new() -> Self {
         Self {
             statistics: DEFAULT_STATISTICS,
-            memory_type_information: [EFiMemoryTypeInformation { memory_type: 0, number_of_pages: 0 };
-                MAX_MEMORY_TYPE_INFO_ENTRIES],
-            memory_type_information_count: 0,
+            memory_type_information: {
+                let mut entries =
+                    [EFiMemoryTypeInformation { memory_type: 0, number_of_pages: 0 }; MAX_MEMORY_TYPE_INFO_ENTRIES];
+                let mut index = 0;
+                while index < entries.len() {
+                    entries[index].memory_type = index as efi::MemoryType;
+                    index += 1;
+                }
+                entries
+            },
             initialized: false,
         }
     }
@@ -309,7 +316,7 @@ impl MemoryBinManager {
 
         let mut top = end;
 
-        for (index, entry) in memory_type_info.iter().enumerate() {
+        for entry in memory_type_info {
             let mem_type = entry.memory_type;
             if mem_type as usize >= EFI_MAX_MEMORY_TYPE {
                 break;
@@ -330,7 +337,6 @@ impl MemoryBinManager {
 
             stats.base_address = top;
             stats.number_of_pages = u64::from(entry.number_of_pages);
-            stats.information_index = index;
 
             log::info!(
                 target: LOG_TARGET,
@@ -345,7 +351,6 @@ impl MemoryBinManager {
         }
 
         self.finalize_information_index(memory_type_info);
-        self.copy_memory_type_info(memory_type_info);
         self.initialized = true;
 
         log::info!(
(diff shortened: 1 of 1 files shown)


os-d:
@thomashinds I was able to reproduce the issue you described. Can you please test with this change and confirm it matches your expectation?

patina_dxe_core/src/memory_bin.rs
```diff
@@ -161,20 +161,28 @@ pub(crate) struct MemoryBinManager {
...
         Self {
             statistics: DEFAULT_STATISTICS,
-            memory_type_information: [EFiMemoryTypeInformation { memory_type: 0, number_of_pages: 0 };
-                MAX_MEMORY_TYPE_INFO_ENTRIES],
-            memory_type_information_count: 0,
+            memory_type_information: {
+                let mut entries =
+                    [EFiMemoryTypeInformation { memory_type: 0, number_of_pages: 0 }; MAX_MEMORY_TYPE_INFO_ENTRIES];
+                let mut index = 0;
+                while index < entries.len() {
```
cfernald, line 179: Might as well do `for index in 0..entries.len() {`

patina_dxe_core/src/memory_bin.rs
```diff
@@ -161,20 +161,28 @@ pub(crate) struct MemoryBinManager {
...
     /// lifetime of the static `MEMORY_BIN_MANAGER`.
     memory_type_information: [EFiMemoryTypeInformation; MAX_MEMORY_TYPE_INFO_ENTRIES],
-    /// Number of valid entries in `memory_type_information`.
-    memory_type_information_count: usize,
     /// Whether bins have been initialized.
     initialized: bool,
 }
 
 impl MemoryBinManager {
+    #[allow(clippy::indexing_slicing)]
```
cfernald, line 170: nit, I would add the explanation comment here instead of below.

patina_dxe_core/src/memory_bin.rs
```diff
@@ -358,44 +365,16 @@ impl MemoryBinManager {
...
-                        target: LOG_TARGET,
-                        "  Bin table: {} pages={}",
-                        memory_type_name(entry.memory_type),
-                        entry.number_of_pages
-                    );
-                }
-            }
+            stats.information_index = mem_type;
         }
+        log::trace!(target: LOG_TARGET, "Bin stats: finalized information indices, reset current_number_of_pages to 0 for all types");
```
joschock: I don't see logic to "reset current_number_of_pages to 0" - am I missing that?

os-d: No, that was there, but isn't needed. We initialize with 0 and then build from there. I'll update the log.

patina_dxe_core/src/memory_bin.rs
cfernald: just something i noticed while reviewing this, but is this sentinel actually needed? It seems in multiple places we check against max memory type, but since this is a rust type we might as well just use size and drop this.

cfernald (approved):

patina_dxe_core/src/memory_bin.rs
os-d: The stats doesn't need it, the info table does. At one point I had a new config table based on the stats structure to be able to share more information, but at this point probably not going to do down that path. But we can add it back if we do. I'll drop it.

patina_dxe_core/src/memory_bin.rs
os-d: Actually no:

image

patina_dxe_core/src/memory_bin.rs
makubacki, line 1: I don't have time to review not, but please leave open until I do.

patina_dxe_core/src/memory_bin.rs
```diff
@@ -358,43 +364,14 @@ impl MemoryBinManager {
...
-                        target: LOG_TARGET,
-                        "  Bin table: {} pages={}",
-                        memory_type_name(entry.memory_type),
-                        entry.number_of_pages
-                    );
-                }
-            }
+            stats.information_index = mem_type;
         }
     }
```
makubacki, line 374: `take_while()` is going to fail the predicate and bail early (e.g. pages for a bin is zero). I think this should follow existing logic and use simple continues and breaks.

```suggestion
    fn finalize_information_index(&mut self, memory_type_info: &[EFiMemoryTypeInformation]) {
        for info in memory_type_info {
            if info.memory_type as usize >= EFI_MAX_MEMORY_TYPE {
                break;
            }
            if info.number_of_pages == 0 {
                continue;
            }
            let mem_type = info.memory_type as usize;
            let stats = self.statistics.get_mut(mem_type).expect("All defined memory types should be in statistics");
            stats.information_index = mem_type;
        }
    }
```

patina_dxe_core/src/memory_bin.rs
makubacki: It looks like `stats.information_index` does not need to be set here if it is set in `finalize_information_index()` now.

makubacki, line 381: Since the HOB data is not copied as before, I think `seed_statistics_from_hobs()` needs to update `memory_type_information` to account for the PEI allocations.

makubacki (approved):

patina_dxe_core/src/memory_bin.rs
makubacki: Diff looks weird. Basically, just saying to drop the `take_while()` with the following as a suggestion:

```rust
    fn finalize_information_index(&mut self, memory_type_info: &[EFiMemoryTypeInformation]) {
        for info in memory_type_info {
            if info.memory_type as usize >= EFI_MAX_MEMORY_TYPE {
                break;
            }
            if info.number_of_pages == 0 {
                continue;
            }
            let mem_type = info.memory_type as usize;
            let stats = self.statistics.get_mut(mem_type).expect("All defined memory types should be in statistics");
            stats.information_index = mem_type;
        }
    }
````

File: patina_dxe_core/src/memory_bin.rs
Comment (makubacki) [reply patina_dxe_core/src/memory_bin.rs:1]: Done

Comment (makubacki) [reply patina_dxe_core/src/memory_bin.rs]: Yeah, it was in stats to avoid discrepancies when required in the info table. It's not a big deal either way to me.

File: patina_dxe_core/src/memory_bin.rs (hunk 4 above)
Comment (os-d) [reply patina_dxe_core/src/memory_bin.rs:374]: Agreed. This works as is because memory_type_info is not sparse, but that isn't an assumption to build off of. But I think `filter()` works here instead

File: patina_dxe_core/src/memory_bin.rs
Comment (os-d) [reply patina_dxe_core/src/memory_bin.rs:381]: Good catch. When I traced this through I thought the original core_allocate_pages was going to do that, but because the bins aren't initialized then, it won't. I'll add the update here.

**Comment by os-d** [comment]
I've addressed the current feedback, but I want to confirm with @thomashinds that this resolves the reported issue before I merge.

**Comment by thomashinds** [comment]
> I've addressed the current feedback, but I want to confirm with @thomashinds that this resolves the reported issue before I merge.

Yes, I can confirm that this resolves the issue. Thanks!

Memory type information prints with released 23.1.0:
```
15:26:25.0062429 - Memory  Previous  Current   Minimum    Next 
15:26:25.0063337 -  Type    Pages     Pages     Pages     Pages 
15:26:25.0064245 - ======  ========  ========  ========  ======= 
15:26:25.0065296 -   09    0000035B  0000035B  00000000  0000035B 
15:26:25.0066138 -   0A    00000090  00000090  00000000  00000090 
15:26:25.0067036 -   00    00000514  00000514  00000000  00000514 
15:26:25.0068198 -   06    00000540  00000540  00000000  00000540 
15:26:25.0069156 -   05    000004C8  000004C8  00000000  000004C8 
```

and with this PR (commit 7408b79153e2a460e4ef4dc0069e20da4eb77888):
```
10:03:03.4808935 - Memory  Previous  Current   Minimum    Next 
10:03:03.4809503 -  Type    Pages     Pages     Pages     Pages 
10:03:03.4809503 - ======  ========  ========  ========  ======= 
10:03:03.4810203 -   09    0000035B  00000000  00000000  0000035B 
10:03:03.4810596 -   0A    00000090  00000050  00000000  00000090 
10:03:03.4811039 -   00    00000514  00000010  00000000  00000514 
10:03:03.4811397 -   06    00000540  00000050  00000000  00000540 
10:03:03.4811645 -   05    000004C8  00000340  00000000  000004C8 
```

PR #1759, opened 2026-08-28
[major] adv_logger: Create adv_logger HW Port Trait

os-d:
## Description

The C based adv logger supports a platform provided hardware port library to let platforms control what kind of hardware port should be logged to.

This functionality is added to patina in the form of a trait. Now Patina doesn't rely directly on a serial port being instantiated and instead abstracts this through the HW port trait. A default serial port trait is provided for platforms that do not wish to use a different serial port or do runtime log filtering.

- [x] Impacts functionality?
- [ ] Impacts security?
- [x] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on a platform that uses the C AdvancedLoggerHdwPortLib. This interface allowed the same functionality.

## Integration Instructions

Platforms must update their bin wrapper repos. To get the same behavior as today, they need to use the `SerialHardwarePort` trait.

e.g.

```rust
// old

static LOGGER: AdvancedLogger<UartPl011> = AdvancedLogger::new(
    Format::Standard,
    &[
        TargetFilter { target: "goblin", log_level: log::LevelFilter::Off, hw_filter_override: None },
        TargetFilter { target: "gcd_measure", log_level: log::LevelFilter::Off, hw_filter_override: None },
        TargetFilter { target: "allocations", log_level: log::LevelFilter::Off, hw_filter_override: None },
        TargetFilter { target: "efi_memory_map", log_level: log::LevelFilter::Off, hw_filter_override: None },
    ],
    LOG_LEVEL,
    // SAFETY: 0xF0000 is the base address of the PL011 MMIO register block used for logging
    unsafe { UartPl011::new(0xF0000) },
);

add.component(AdvancedLoggerComponent::<UartPl011>::new(&LOGGER));

// new

static LOGGER: AdvancedLogger<SerialHardwarePort<UartPl011>> = AdvancedLogger::new(
    Format::Standard,
    &[
        TargetFilter { target: "goblin", log_level: log::LevelFilter::Off, hw_filter_override: None },
        TargetFilter { target: "gcd_measure", log_level: log::LevelFilter::Off, hw_filter_override: None },
        TargetFilter { target: "allocations", log_level: log::LevelFilter::Off, hw_filter_override: None },
        TargetFilter { target: "efi_memory_map", log_level: log::LevelFilter::Off, hw_filter_override: None },
    ],
    LOG_LEVEL,
    // SAFETY: 0xF0000 is the base address of the PL011 MMIO register block used for logging
    SerialHardwarePort::new(unsafe { UartPl011::new(0xF0000) }),
);

add.component(AdvancedLoggerComponent::<SerialHardwarePort<UartPl011>>::new(&LOGGER));
```
Platforms may now implement their own hardware port trait to do custom behavior on hardware port writes.

### DIFF
--- a/components/patina_acpi/src/lib.rs
+++ b/components/patina_acpi/src/lib.rs
@@ -16,7 +16,7 @@
 //!
 //! impl ComponentInfo for Intel {
 //! fn components(mut add: Add<Component>) {
-//!         add.component(AdvancedLoggerComponent::<Uart16550>::new(&LOGGER));
+//!         add.component(AdvancedLoggerComponent::<SerialHardwarePort<Uart16550>>::new(&LOGGER));
 //!         // Other platform components...
 //!         add.component(patina_acpi::component::AcpiProviderManager::new(oem_id, ... /* Other platform init. */));
 //!         add.component(patina_acpi::component::AcpiSystemProtocolManager::default());
--- a/components/patina_adv_logger/src/component.rs
+++ b/components/patina_adv_logger/src/component.rs
@@ -17,40 +17,39 @@ use patina::{
         service::{Service, perf_timer::ArchTimerFunctionality},
     },
     error::{EfiError, Result},
-    peripheral::serial::SerialIO,
     uefi::boot_services::{BootServices, StandardBootServices},
 };
 
-use crate::{logger::AdvancedLogger, protocol::AdvancedLoggerProtocol};
+use crate::{hardware_port::AdvancedLoggerHardwarePort, logger::AdvancedLogger, protocol::AdvancedLoggerProtocol};
 
 /// C struct for the internal Advanced Logger protocol for the component.
 #[repr(C)]
-struct AdvancedLoggerProtocolInternal<S>
+struct AdvancedLoggerProtocolInternal<P>
 where
-    S: SerialIO + Send + 'static,
+    P: AdvancedLoggerHardwarePort + 'static,
 {
     // The public protocol that external callers will depend on.
     protocol: AdvancedLoggerProtocol,
 
     // Internal component access only! Does not exist in C definition.
-    adv_logger: &'static AdvancedLogger<'static, S>,
+    adv_logger: &'static AdvancedLogger<'static, P>,
 }
 
 /// The component that will install the Advanced Logger protocol.
-pub struct AdvancedLoggerComponent<S>
+pub struct AdvancedLoggerComponent<P>
 where
-    S: SerialIO + Send + 'static,
+    P: AdvancedLoggerHardwarePort + 'static,
 {
-    adv_logger: &'static AdvancedLogger<'static, S>,
+    adv_logger: &'static AdvancedLogger<'static, P>,
 }
 
 #[component]
-impl<S> AdvancedLoggerComponent<S>
+impl<P> AdvancedLoggerComponent<P>
 where
-    S: SerialIO + Send + 'static,
+    P: AdvancedLoggerHardwarePort + 'static,
 {
     /// Creates a new `AdvancedLoggerComponent`.
-    pub const fn new(adv_logger: &'static AdvancedLogger<S>) -> Self {
+    pub const fn new(adv_logger: &'static AdvancedLogger<P>) -> Self {
         Self { adv_logger }
     }
 
@@ -72,7 +71,7 @@ where
 
         // SAFETY: `this` is null-checked above. The protocol struct is installed by Patina with
         //         `Box::leak`, so its alignment and validity are guaranteed.
-        let internal = unsafe { &*this.cast::<AdvancedLoggerProtocolInternal<S>>() };
+        let internal = unsafe { &*this.cast::<AdvancedLoggerProtocolInternal<P>>() };
 
         internal.adv_logger.log_write(error_level, None, data);
         efi::Status::SUCCESS
@@ -111,6 +110,7 @@ where
 mod tests {
     use super::*;
     use crate::{
+        hardware_port::SerialHardwarePort,
         logger::{AdvancedLogger, TargetFilter},
         writer::AdvancedLogWriter,
     };
@@ -134,11 +134,11 @@ mod tests {
         }
     }
 
-    static TEST_LOGGER: AdvancedLogger<UartNull> = AdvancedLogger::new(
+    static TEST_LOGGER: AdvancedLogger<SerialHardwarePort<UartNull>> = AdvancedLogger::new(
         Format::Standard,
         &[TargetFilter { target: "", log_level: log::LevelFilter::Trace, hw_filter_override: None }],
         log::LevelFilter::Trace,
-        UartNull {},
+        SerialHardwarePort::new(UartNull {}),
     );
 
     /// Initializes `TEST_LOGGER` with a newly allocated memory log and a mock timer.
@@ -155,9 +155,9 @@ mod tests {
     /// Builds a leaked `AdvancedLoggerProtocolInternal` structure and returns a `*const AdvancedLoggerProtocol`
     /// that can be passed as `this` to `adv_log_write`.
     fn leak_protocol_this() -> *const AdvancedLoggerProtocol {
-        let internal = Box::leak(Box::new(AdvancedLoggerProtocolInternal::<UartNull> {
+        let internal = Box::leak(Box::new(AdvancedLoggerProtocolInternal::<SerialHardwarePort<UartNull>> {
             protocol: AdvancedLoggerProtocol::new(
-                AdvancedLoggerComponent::<UartNull>::adv_log_write,
+                AdvancedLoggerComponent::<SerialHardwarePort<UartNull>>::adv_log_write,
                 TEST_LOGGER.get_log_address().unwrap_or(0),
             ),
             adv_logger: &TEST_LOGGER,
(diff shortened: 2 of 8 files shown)


components/patina_adv_logger/src/component.rs
```diff
@@ -17,40 +17,39 @@ use patina::{
...
     protocol: AdvancedLoggerProtocol,
 
     // Internal component access only! Does not exist in C definition.
-    adv_logger: &'static AdvancedLogger<'static, S>,
+    adv_logger: &'static AdvancedLogger<'static, P>,
 }
 
 /// The component that will install the Advanced Logger protocol.
-pub struct AdvancedLoggerComponent<S>
+pub struct AdvancedLoggerComponent<P>
```
joschock, line 39: As a potential ergonomic enhancement, consider providing a default generic for consumers that don't want a port at all? e.g.

```rust
struct NoHdwPort {}
impl AdvancedLoggerHardwarePort for NoHdwPort {
  fn write(&self, error_level: u32, buffer: &[u8]) -> Result<(), EfiError> {
    Ok(())
  }
}

pub struct AdvancedLoggerComponent<P = NoHdwPort>
```
Not sure this is worth the squeeze tho.

joschock (approved):

components/patina_acpi/src/lib.rs
makubacki, line 1: I don't have time to review at the moment, but please leave it open until I review.

components/patina_adv_logger/src/hardware_port.rs
```diff
@@ -0,0 +1,88 @@
...
+/// Implementations must be usable from a `static` and are expected to provide their own
+/// synchronization, as the advanced logger holds the port behind a shared reference.
+///
+/// ## Errors
+///
+/// Returns an error if the message could not be emitted. Implementations that intentionally
+/// suppress output should return `Ok(())` rather than an error, as a suppressed message is not a
+/// failure.
+#[cfg_attr(test, mockall::automock)]
+pub trait AdvancedLoggerHardwarePort: Send + Sync {
```
makubacki: Postfixing type names in this module (`hardware_port`) with `HardwarePort` is verbose and the long names are a little distracting to me. Using as `hardware_port::AdvancedLogger` where needed is easier to read.

components/patina_adv_logger/src/hardware_port.rs
```diff
@@ -0,0 +1,88 @@
...
+///
+/// Implementations must be usable from a `static` and are expected to provide their own
+/// synchronization, as the advanced logger holds the port behind a shared reference.
+///
+/// ## Errors
+///
+/// Returns an error if the message could not be emitted. Implementations that intentionally
+/// suppress output should return `Ok(())` rather than an error, as a suppressed message is not a
+/// failure.
+#[cfg_attr(test, mockall::automock)]
```
makubacki, line 38: We usually use:

```suggestion
#[cfg_attr(any(test, feature = "mockall"), automock)]
```

Was there a particular reason not to here?

components/patina_adv_logger/src/hardware_port.rs
makubacki, line 39: Have you considered making this:

```suggestion
pub trait AdvancedLoggerHardwarePort: Send + Sync + 'static
```

That would cleanup the where clauses. For example, `P: AdvancedLoggerHardwarePort + 'static,` to `P: AdvancedLoggerHardwarePort,`.

components/patina_adv_logger/src/hardware_port.rs
```diff
@@ -0,0 +1,88 @@
...
+pub trait AdvancedLoggerHardwarePort: Send + Sync {
+    /// Writes the provided message bytes to the hardware port.
+    fn write(&self, error_level: u32, buffer: &[u8]) -> Result<(), EfiError>;
```
makubacki, line 41: If the expectation is that code outside the crate will filter with the constants here:

https://github.com/OpenDevicePartnership/patina/blob/a334b36586da55b499a1f7be2738c69413e58f59/components/patina_adv_logger/src/memory_log.rs#L34-L42

Like:

```rust
impl AdvancedLoggerHardwarePort for MyGatedPort {
    fn write(&self, error_level: u32, buffer: &[u8]) -> Result<(), EfiError> {
        if error_level & patina_adv_logger::memory_log::DEBUG_LEVEL_ERROR...
}
```

I think that will be a problem because the `memory_log` module is not public:

https://github.com/OpenDevicePartnership/patina/blob/a334b36586da55b499a1f7be2738c69413e58f59/components/patina_adv_logger/src/lib.rs#L14

components/patina_adv_logger/src/hardware_port.rs
```diff
@@ -0,0 +1,88 @@
+//! UEFI Advanced Logger Hardware Port Support
+//!
+//! This module provides the [`AdvancedLoggerHardwarePort`] abstraction used by the advanced logger
+//! to emit log messages to a hardware port, along with [`SerialHardwarePort`], the default
+//! implementation that writes to a [`SerialIO`] instance.
+//!
+//! Platforms that need to gate, redirect, or otherwise customize hardware port output can provide
+//! their own implementation of [`AdvancedLoggerHardwarePort`] instead of using
+//! [`SerialHardwarePort`]. This is the Patina equivalent of the EDK II `AdvancedLoggerHdwPortLib`
```
makubacki, line 9: I'm not sure it's worth including "This is the equivalent of the EDK II `AdvancedLoggerHdwPort` library class. (1) that's not in EDK II (in mu_plus) (2) that could change and this becomes stale and (3) I don't think it means much - someone either knows that already or it doesn't matter and they just implement the trait as needed.

makubacki (approved):

vineelko:
I like the idea of having a new trait to address the current shortcomings of `SerialIO` and the issues of `AdvancedLogger` directly depending on `SerialIO`. However, I think the trait's definition, its name might need some minor fixes.

Nothing in the definition below is really hardware specific as side from its name:
```rust
pub trait AdvancedLoggerHardwarePort: Send + Sync {
    /// Writes the provided message bytes to the hardware port.
    fn write(&self, error_level: u32, buffer: &[u8]) -> Result<(), EfiError>;
}
```

The above definition can be refined to something like the following where SerialIO is the supertrait:

```rust
// components\patina_adv_logger\src\advanced_logger_serial_io.rs
pub trait AdvancedLoggerSerialIO: SerialIO {
    fn write(&mut self, _error_level: u32, buffer: &[u8]) -> Result<(), EfiError> {
        SerialIO::write(self, buffer);
        Ok(())
    }
}
```

and the `AdvancedLogger` will use it as `let result = self.hardware_port.write(error_level, data);` (exactly like what you are doing). 

The advanced logger could then be bounded by `AdvancedLoggerSerialIO` as `AdvancedLogger<S: AdvancedLoggerSerialIO>`, (exactly like what you are doing).

`AdvancedLoggerSerialIO` feels more closer to `AdvancedLogger` in my opinion. 

Minor update: May be we should actually name the trait as `AdvancedLoggerIO`

components/patina_adv_logger/src/component.rs
os-d: I don't really have a preference, don't mind adding. The only thing that may be odd about it is that we have the hdw port configuration knob and having a no-port hdw port would be the same as having the hdw port disabled. I agree it is odd to force a hw port if there isn't one, but I also worry about the conflicting settings.

components/patina_adv_logger/src/hardware_port.rs
os-d: Good catch, these should be moved to the SDK, they are standard error levels

components/patina_adv_logger/src/hardware_port.rs
os-d: Ugh, I had an llm check that I wasn't missing any locations and I thought I went and cleaned up the garbage it generated, but I missed this.

os-d:
> I like the idea of having a new trait to address the current shortcomings of `SerialIO` and the issues of `AdvancedLogger` directly depending on `SerialIO`. However, I think the trait's definition, its name might need some minor fixes.
> 
> Nothing in the definition below is really hardware specific as side from its name:
> 
> ```rust
> pub trait AdvancedLoggerHardwarePort: Send + Sync {
>     /// Writes the provided message bytes to the hardware port.
>     fn write(&self, error_level: u32, buffer: &[u8]) -> Result<(), EfiError>;
> }
> ```
> 
> The above definition can be refined to something like the following where SerialIO is the supertrait:
> 
> ```rust
> // components\patina_adv_logger\src\advanced_logger_serial_io.rs
> pub trait AdvancedLoggerSerialIO: SerialIO {
>     fn write(&mut self, _error_level: u32, buffer: &[u8]) -> Result<(), EfiError> {
>         SerialIO::write(self, buffer);
>         Ok(())
>     }
> }
> ```
> 
> and the `AdvancedLogger` will use it as `let result = self.hardware_port.write(error_level, data);` (exactly like what you are doing).
> 
> The advanced logger could then be bounded by `AdvancedLoggerSerialIO` as `AdvancedLogger<S: AdvancedLoggerSerialIO>`, (exactly like what you are doing).
> 
> `AdvancedLoggerSerialIO` feels more closer to `AdvancedLogger` in my opinion.
> 
> Minor update: May be we should actually name the trait as `AdvancedLoggerIO`

The general idea is that this is abstracting the hardware port from advanced logger. The platform may choose to log to serial, or to a SEL log, network, whatever. So that's where the name hardware port came from. It's not my name, its the existing name: https://github.com/microsoft/mu_plus/tree/release/202511/AdvLoggerPkg/Library/AdvancedLoggerHdwPortLib. My preference would be to keep the naming consistent with the C implementation unless we are changing the layering, etc. But since we will have mixed C and Rust use for a long time, I think it makes sense to be consistent here.

I'll have to play with the supertrait thing, but as long as that will allow other hdw port impls that aren't SerialIO, that's fine with me.

os-d:
@Javagedes just explained a similar SerialIO idea to me and that makes sense and also avoids the breaking change. I'm going to close this PR and open a new one targeting main.

components/patina_adv_logger/src/hardware_port.rs
os-d: Updated

components/patina_adv_logger/src/hardware_port.rs
os-d: Updated.

PR #1760, opened 2026-08-29
RFC: Patina UEFI Variables

makubacki:
## Description

Initial RFC for UEFI variables in Patina.

Still a work in progress. PR is in draft while implementation is completed so any changes can be captured in the RFC.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [x] Includes documentation?

## How This Was Tested

- `cargo make all`

## Integration Instructions

- N/A

### DIFF
--- a/docs/src/uefi_variables/uefi_variables_design.md
+++ b/docs/src/uefi_variables/uefi_variables_design.md
@@ -1,0 +1,560 @@
+# Patina UEFI Variable Design
+
+## Introduction
+
+This document describes the Patina UEFI variable design. This design is meant to be extensible and easy to test.
+
+As a result, the design involves a stronger use of interfaces and other software constructs than typically defined
+in firmware drivers (and today's EDK II UEFI variable stack). Such separation supports:
+
+- Better portability and cooperation of separate design parts across language boundaries
+- Increased extensibility to introduce new design elements and extend instances of presently defined elements
+- Isolated testing of individual parts of the design
+
+To establish a lexicon used to identify individual entities that compose the overall variable design, a set of primary
+design pieces are defined below and referenced throughout the remainder of the document and source repository.
+
+## Zones
+
+Zones are regions of the overall design in which residing elements share a common high-level responsibility and are
+completely decoupled from other design areas. Control only transfers between zones across well-defined APIs. All data
+transferred across zones is considered external and untrusted.
+
+- `External` - Code outside the front-end and back-end.
+  - All interaction with variable services occurs through externally defined interfaces.
+- `Front-end` - Performs data validation and business logic. When data leaves the front-end, it has been validated
+  and transformed to its final state for transfer to the back-end.
+- `Back-end` - Handles moving data between the front-end and data destinations.
+
+Every zone depends on a `Zone Gasket` to receive data from the zone before it, with the exception of the `External`
+zone, which is the first zone data enters, has no zone before it, and therefore has no `Zone Gasket`
+(see [Gasket](#gasket)).
+
+## Flows
+
+The variable software stack moves data from an external caller to a store and from a store back to an external caller.
+
+The direction of data movement from caller to store is called `input flow` and the movement back to the caller is
+called `output flow`.
+
+Some elements may perform the same actions regardless of flow. In other cases, flow will completely change how the
+component behaves.
+
+## Stages
+
+As data progress within a zone, it is passed between `stages`. A stage is a container of `elements`. The duration of
+execution for each element within the stage is called the element's `time`.
+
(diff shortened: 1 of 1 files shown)


docs/src/uefi_variables/uefi_variables_design.md
```diff
@@ -0,0 +1,560 @@
...
+A store with an associated cache should prime that cache (populate it from the store) during initialization, so
+reads can be served from the cache without relying upon the underlying store. In case of a cache miss, a read policy
+may be configured to either fall through to the store via a read request or return a not found error. In most cases,
+it is recommended to prime the cache to avoid potential performance penalties resulting from cache misses. In any case,
+an associate cache must be coherent with its store, so a write to the store must also update the cache, and the cache
+must accurately reflect the store's contents at all times.
+
+### Gasket
+
+A `gasket` is a data bridge that connects one design piece to another. A gasket does not modify the actual variable
```
cfernald, line 77: Why *Gasket* instead of just *Bridge*? It seems if we are going to define a gasket as a bridge, then the more standard term bridge may just be easier.

docs/src/uefi_variables/uefi_variables_design.md
```diff
@@ -0,0 +1,560 @@
...
+        ST["Storage Stage"]
+        D[("Store")]
+        Ca[("Cache")]
+        G --> ST --> D
+        D -.->|"associated cache"| Ca
+    end
+
+    Ext ==>|"input flow (Zone Gasket)"| A
+    R ==>|"routed data"| G
+    R ==>|"cache check"| Ca
```
cfernald, line 347: It seems odd that the router, part of the front-end, would explicitly check the cache in the back end. In my mind either the cache is part of the front-end and the router maintains it, or the cache is part of the backend, and the router is agnostic to its existence.

docs/src/uefi_variables/uefi_variables_design.md
```diff
@@ -0,0 +1,560 @@
...
+To establish a lexicon used to identify individual entities that compose the overall variable design, a set of primary
+design pieces are defined below and referenced throughout the remainder of the document and source repository.
+
+## Zones
+
+Zones are regions of the overall design in which residing elements share a common high-level responsibility and are
+completely decoupled from other design areas. Control only transfers between zones across well-defined APIs. All data
+transferred across zones is considered external and untrusted.
+
+- `External` - Code outside the front-end and back-end.
```
cfernald, line 23: Concrete examples of the zones would be helpful in digesting this. E.g. Front end: EDK2' Variable DXE driver, Back End: Standalone MM's variable driver, etc.

docs/src/uefi_variables/uefi_variables_design.md
```diff
@@ -0,0 +1,560 @@
...
+
+The variable software stack moves data from an external caller to a store and from a store back to an external caller.
+
+The direction of data movement from caller to store is called `input flow` and the movement back to the caller is
+called `output flow`.
+
+Some elements may perform the same actions regardless of flow. In other cases, flow will completely change how the
+component behaves.
+
+## Stages
```
cfernald, line 43: Perhaps I'm missing the point, but as a note. I think the transition between implementation boundary definitions and implementation specific definitions is a bit confusing. Zones, Gaskets, etc. all relate to well defines ABIs and connecting physically different entities where Stages gets into the specific implementation within a single binary. I wonder if it would be helpful to define this in layers such that you talk about the high-level connection between Zones and then the layering within the Zone. At least in my readthrough that would have helped me get a mental model faster.

docs/src/uefi_variables/uefi_variables_design.md
makubacki: There was an intention to make the term less ambiguous so it's clear it refers to this specific definition. For example, as you've pointed out, in normal conversation, "bridge" is generic and will likely not be interpreted as this specific design piece whereas "gasket" would.

I'm open to considering "bridge" but I'm also concerned about that washing away the unique identification of this piece.

docs/src/uefi_variables/uefi_variables_design.md
makubacki: The main reason is that the backend is expected to maintain its cache outside of high-cost entry/exit environments like MM while the main I/O operations (e.g., read and write) are expected/allowed to be within those environments. Does that make sense? If so, I can clarify this in the RFC.

docs/src/uefi_variables/uefi_variables_design.md
cfernald: I think it's reasonable to keep gasket if it adds value.

PR #1762, opened 2026-09-01
patina_dxe_core: pecoff: allow malformed debug directory

### LINKED ISSUE #1761: [Bug]: Treat PE/COFF debug directory as optional metadata during image load

A related failure was reported in OpenDevicePartnership/patina-dxe-core-qemu#134 and addressed by OpenDevicePartnership/patina#1378 for parsing an already-mapped DXE Core. This report concerns ordinary on-disk PE/COFF images passed to `LoadImage()`.

### Current Behavior

Patina can reject an otherwise valid and executable PE/COFF image when Goblin cannot parse or resolve its debug directory.

`UefiPeInfo::from_pe()` parses the complete image using `goblin::pe::PE::parse_with_opts()`. The default Goblin options use strict parsing and resolve data-directory RVAs through the section table. An error while parsing optional debug metadata therefore propagates out of `UefiPeInfo::parse()`.

`core_load_pe_image()` converts that parse error into `EfiError::Unsupported`, preventing the image from loading:

```rust
let pe_info = pecoff::UefiPeInfo::parse(image).map_err(|err| {
    log::error!("core_load_pe_image failed: UefiPeInfo::parse returned {err:?}");
    EfiError::Unsupported
})?;
```

A real x64 EFI application exposes the problem with this layout:

```text
Machine:                 IMAGE_FILE_MACHINE_AMD64
Subsystem:               IMAGE_SUBSYSTEM_EFI_APPLICATION
SectionAlignment:        0x1000
SizeOfHeaders:           0x400
Debug directory RVA:     0x3ac
Debug directory size:    0x1c
Debug type:              IMAGE_DEBUG_TYPE_CODEVIEW
CodeView data RVA:       0x3c8
CodeView file offset:    0x3c8
CodeView data size:      0x38
CodeView signature:      RSDS
```

The debug directory and CodeView record are intentionally located in the PE headers. Their ranges are within `SizeOfHeaders` and within the file.

Goblin's `find_offset()` resolves an RVA by looking only for a containing section when `resolve_rva` is enabled. Since `0x3ac` is in the headers and not in a section, Goblin returns an error equivalent to:

```text
Goblin(Malformed(
    "Cannot map ImageDebugDirectory rva 0x3ac into offset"
))
```

Patina propagates this optional-metadata error and rejects the entire image before loading its executable sections.

The affected image otherwise passes the relevant UEFI image checks:

- Valid PE32+ x64 EFI application
- 4 KiB section alignment
- No writable and executable section
- `IMAGE_DLLCHARACTERISTICS_NX_COMPAT` set
- Valid section ranges
- Valid relocation directory
- Valid Authenticode certificate table
- Valid entry point and image size

EDK2 successfully treats the debug information as ancillary metadata for CodeView/PDB discovery and does not require it to determine whether this image's executable sections can be loaded, relocated, and started.

### Expected Behavior

Failure to parse, resolve, or interpret the PE debug directory should not prevent an otherwise valid PE/COFF image from loading.

Patina should:

1. Continue validating all loader-critical PE/COFF structures strictly.
2. Treat the debug directory as optional metadata.
3. Use debug metadata when it is valid and supported.
4. Log or record that debug metadata was unavailable when it cannot be parsed.
5. Continue loading the image without a derived debug filename.
6. Preserve strict bounds checking and never dereference an invalid debug-directory range.

For this example, `UefiPeInfo::filename` should be `None` or use the caller's default name, while the image continues through section loading, relocation, memory protection, and execution.

### Steps To Reproduce

1. Create or obtain a valid x64 EFI application.
2. Set the PE debug data-directory entry to a valid range within `SizeOfHeaders`, rather than placing it in a section.
3. Add one `IMAGE_DEBUG_TYPE_CODEVIEW` entry and an RSDS record within the headers.
4. Call Patina `EFI_BOOT_SERVICES.LoadImage()` with the image.
5. Observe that `UefiPeInfo::parse()` fails while mapping the debug-directory RVA through the section table.
6. Remove or zero only the debug data-directory entry.
7. Observe that the same executable image loads successfully.

Example image identifier:

```text
SHA-256:
2B93EDDA7BA390ECFD2C38CAB7D593AD3B26E1D170B8D0D6D0E217BCDA1D25B3
```

The original binary should only be attached if its redistribution is permitted. A synthetic PE fixture with the same header layout should be sufficient for a regression test.

### Suggested Direction

Avoid making all PE parsing permissive, since that could unintentionally weaken handling of loader-critical structures such as sections, relocations, image sizes, or resource data.

Instead, isolate optional debug-directory extraction from the parse required to load the image. Possible approaches include:

- Add a Goblin option to disable debug-data parsing.
- Update Goblin RVA resolution to recognize valid RVAs within `SizeOfHeaders`.
- Parse loader-critical PE information first, then attempt debug metadata extraction as a best-effort operation.
- Catch only debug-metadata parsing failures and leave `UefiPeInfo::filename` unset.

The selected implementation should not hide errors in executable sections, relocations, architecture, subsystem, alignment, image size, or other loader-critical data.

### Acceptance Criteria

- [ ] A valid debug directory located within `SizeOfHeaders` does not cause `LoadImage()` to fail.
- [ ] Missing debug metadata does not cause `LoadImage()` to fail.
- [ ] An unsupported CodeView type does not cause `LoadImage()` to fail.
- [ ] An unresolvable or truncated debug record is safely ignored or reported without accessing memory outside the input buffer.
- [ ] Loader-critical PE/COFF validation remains strict.
- [ ] Valid CodeView metadata continues to populate `UefiPeInfo::filename`.
- [ ] Unit tests cover debug metadata in a section, in the headers, absent, unsupported, truncated, and out of bounds.
- [ ] The affected x64 EFI application reaches `StartImage()`.

### Build Environment

```text
Host OS: Windows
Target architecture: x64
Image type: EFI application
PE parser: goblin 0.10.7
Targets potentially impacted: Any target loading PE/COFF images whose debug directory is stored in the PE headers
```

### Version Information

```text
Current main inspected:
a334b36586da55b499a1f7be2738c69413e58f59

Affected platform version/tag:
TBD
```

### Urgency

Medium

This prevents standards-compatible third-party EFI applications from loading despite their executable content, section layout, relocations, and memory-mitigation metadata being valid.

No

Yes

Feedback is requested on whether the preferred fix should be implemented within Patina's PE/COFF abstraction or contributed upstream to Goblin, particularly for debug-directory RVAs located within `SizeOfHeaders`.

### Anything else?

Relevant references:

- Patina PE/COFF parsing: https://github.com/OpenDevicePartnership/patina/blob/main/patina_dxe_core/src/pecoff.rs
- Patina image-loading path: https://github.com/OpenDevicePartnership/patina/blob/main/patina_dxe_core/src/pi_dispatcher/image.rs
- Related mapped-image parsing fix: https://github.com/OpenDevicePartnership/patina/pull/1378
- Related SBSA failure: https://github.com/OpenDevicePartnership/patina-dxe-core-qemu/issues/134
- EDK2 BasePeCoffLib: https://github.com/tianocore/edk2/blob/master/MdePkg/Library/BasePeCoffLib/BasePeCoff.c
- Microsoft PE/COFF format: https://learn.microsoft.com/en-us/windows/win32/debug/pe-format

Javagedes:
## Description

The debug directory is an optional directory not required to load and dispatch an image. This commit updates the pecoff parsing such that a malformed debug directory will be skipped instead of outright rejecting the image.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Regression tests on an EFI image that has a malformed debug directory.

## Integration Instructions

N/A


os-d (approved):

cfernald (approved):
If we have never returned a `None` before from this code, might be worth a quick check that there are not expects on that data if you haven't already.

Javagedes:
> If we have never returned a `None` before from this code, might be worth a quick check that there are not expects on that data if you haven't already.

@cfernald the debug directory has always been optional, and the code supports the fact that the filename may or may not exist. This change simply updates the logic such that a malformed data directory is treated as if we don't have a data directory whereas before it was treated as an error that we would reject the image over. 

This change brings us in line with the C implementation regarding this.

PR #1763, opened 2026-09-01
[REBASE & FF]AdvLogger: Add Dynamic HW Port Error Level Filtering

os-d:
## Description

This adds partial responsibility of the C AdvLoggerHdwPortLib to Patina. That lib in the C code allows for abstracting the hw port logged to, which is already provided by the SerialIO interface in Patina. However, that lib also allows for dynamic hw port error level filtering, which Patina does not currently provide.

This PR adds an optional callback for platforms to provide an opportunity to dynamically set the hw port error level.

This originally targeted the major branch in: #1759 but feedback there gave an approach to avoid a breaking change.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

On a platform with the existing SerialIO initialization and another platform with a custom HW port trait.

## Integration Instructions

This is not a breaking change, but platforms may now introduce a custom callback, e.g.

```rust
fn platform_hw_print_level(hw_print_level: u32) -> u32 {
   if platform_debug_port_enabled() {
      hw_print_level
   } else {
      0
   }
}

static LOGGER: AdvancedLogger<UartNull> = AdvancedLogger::new(
   Format::Standard,
   &[],
   log::LevelFilter::Info,
   UartNull {},
).with_hw_print_level_override_callback(platform_hw_print_level);
```


components/patina_adv_logger/src/logger.rs
```diff
@@ -539,4 +535,83 @@ mod tests {
...
+
+    // === Hardware port dispatch ===
+
+    /// Creates a memory log with the requested global hardware print level and returns its address.
+    fn create_memory_log(hw_print_level: u32) -> efi::PhysicalAddress {
+        const LOG_LEN: usize = 0x2000;
+        let log_buff = Box::into_raw(Box::new([0_u8; LOG_LEN]));
+        let log_address = log_buff as *const u8 as efi::PhysicalAddress;
+
+        // SAFETY: We just allocated this memory so it is valid for the header.
```
Javagedes, line 549: nit:

Safety comment should address all safety requires specified in the unsafe function being called: https://doc.rust-lang.org/std/ptr/fn.write.html#safety

Javagedes: Oh this is in testing code. closing.

joschock (approved):

components/patina_adv_logger/src/hardware_port.rs
```diff
@@ -0,0 +1,57 @@
...
+/// and do not need to provide interior mutability. Ports must be valid for the static lifetime so
+/// they can be stored by a static logger.
+///
+/// ## Errors
+///
+/// Returns an error if the message could not be emitted. Implementations that intentionally
+/// suppress output should return `Ok(())` rather than an error, as a suppressed message is not a
+/// failure.
+#[cfg_attr(any(test, feature = "mockall"), mockall::automock)]
+pub trait AdvancedLoggerHardwarePort: Send + 'static {
```
Javagedes, line 34: This write function is doing two things at once - writing and filtering.

I'm wondering if we should follow in the footsteps of the actual `log` crate's `Log` trait by splitting that into two  two functions - writing and filtering.

```rust
pub trait AdvancedLoggerHardwarePort: Send + 'static {
  fn write(&mut self, buffer: &[u8]) -> Result<(), EfiError>;

  fn enabled(&self, error_level: u32) -> bool;
}

impl<S: SerialIO + 'static> AdvancedLoggerHardwarePort for S {
    fn write(&mut self, buffer: &[u8]) -> Result<(), EfiError> {
        <Self as SerialIO>::write(self, buffer);
        Ok(())
    }

    #[inline]
    fn enabled(&self, _error_level: u32) -> bool { true }
}


// elsewhere

if self.enabled() { self.write(bytes) }
```

Javagedes: I also question if "AdvancedLoggerHardwarePort" is the right name for the trait as all it does is provide an abstraction to write bytes to something, and to filter incoming messages to something. But that is just a thought because I'm pretty sure I don't fully understand the scope of this.

os-d: Think of this interface more as a gasket (yes I've been reading Michael's RFC :). Adv Logger maintains a memory log and then passes the responsibility of writing to a hardware port to another component. This is the gasket between the memory log writing that adv logger does and the hardware port, which it has no insight into. We don't want to tie adv logger into anything about the hardware port, because that is very platform specific.

So this interface is, okay, great, I've written this to the memory log, and now I'm throwing it over the fence to the hw port and that component can do whatever it wants with it.

Why does adv logger then care about the hw port error level in addition to having this interface? It shouldn't but we are constrained by history here. The hardcoded error level came first (in the C code) and then this interface (also in the C code). In order to not break existing consumers, both were kept. I think it would be valid to evaluate a breaking change to drop the hw port error level and only rely on this interface. It does require custom logic for every platform though, which many don't need this level of control.

As for the name, that was a comment on the last PR too, because this is an [existing interface in the C code](https://github.com/microsoft/mu_plus/tree/release/202511/AdvLoggerPkg/Library/AdvancedLoggerHdwPortLib) I wanted to preserve the name to make it easier to understand for platforms that have to work with both. If we were fundamentally changing the design, I would be fine with a new name, but in lieu of a strong reason, I'd prefer to keep the cohesion.

Javagedes (approved):

makubacki (approved):

sdk/patina/src/debug/log.rs
```diff
@@ -43,6 +43,51 @@ pub use serial_logger::Logger as SerialLogger;
 
 use crate::writelncrlf;
 
+/// Initialization messages.
```
cfernald, line 46: probably worth mentioning in a comment these mirror the EDK2 implementaiton.

sdk/patina/src/peripheral/serial/shared.rs
```diff
@@ -11,27 +11,27 @@ use spin::Mutex;
...
+/// A wrapper that provides guaranteed single threaded exclusive access to a hardware port **in the patina
+/// environment**. This is a special implementation because of the unique use-case of hardware ports that sit outside
+/// of the general Patina TPL based locking model.
 ///
-/// This structure makes the assumption that except for error scenarios, that serial port implementations will not be
+/// This structure makes the assumption that except for error scenarios, hardware port implementations will not be
 /// reentrant. Otherwise, spurious errors may be observed.
-pub struct SharedSerial<T: SerialIO> {
-    serial: Mutex<T>,
+pub struct SharedPort<T: Send> {
```
cfernald, line 20: I'm not convinced this is necessary. The ShareSerial was supposed to be a special-purpose lock where we could break the normal expectations because serial I/O is abnormally low-level and can't use our usual interrupt management methods (TPL). However, expanding this to be open means we now expose that functionality more broadly.

For the purposes of this PR. it seems it's only needed because `AdvancedLoggerHardwarePort` expects a mutable reference. That is, the layering is `SharedPort -> mut AdvancedLoggerHardwarePort -> mut SerialIO`. I wonder if instead AdvancedLoggerHardwarePort should take an immutable ref and provide any needed interior mutability itself. E.g. `AdvancedLoggerHardwarePort -> SharedSerial -> mut SerialIO`

components/patina_adv_logger/src/hardware_port.rs
```diff
@@ -0,0 +1,57 @@
...
+    /// Writes the provided message bytes to the hardware port.
+    fn write(&mut self, error_level: u32, buffer: &[u8]) -> Result<(), EfiError>;
```
cfernald, line 36: do we expect the implementation to do anything with the error level other than filter log messages? Instead of creating a wrapper trait, should we instead provide an optional filter the platform could provide?

sdk/patina/src/peripheral/serial/shared.rs
Javagedes: Making `AdvancedLoggerHardwarePort` be `&self` would turn this back into a breaking change because all consumers would have to pass `SharedSerial<S: SerialIO>` to the `AdvancedLogger::new`. This is because we will no longer be able to auto derive `AdvancedLoggerHardwarePort` for `SerialIO` due to needing interior mutability because `AdvancedLoggerHardwarePort` only takes a reference.

We could auto derive `AdvancedLoggerHardwarePort` for `SharedSerial<S: SerialIO>`, but that would still require all existing call sites be updated to pass in `SharedSerial::new(S: SerialIO)`

os-d: As Joey said, I had it without this, but it was a breaking change and he recommended to move to this for a non-breaking change. I don’t have too big of a preference. I’d rather the interface be the right one than worry about a breaking change.

components/patina_adv_logger/src/hardware_port.rs
os-d: This also abstracts the hardware port from adv logger. Now, you could say it will always be some form of SerialIO, that could be legit, but we’d want to ensure we aren’t coupling SerialIO to mean a UART if so. And I don’t know if we want to do that globally or not.

I was thinking of moving the hw port error level check into the trait itself instead of outside. Then the default SerialIO one would do the check against just the static level, like today. Makes a bit more sense.

sdk/patina/src/peripheral/serial/shared.rs
cfernald: Unfortunately, short of creating a separate top level struct on the adv logger side that takes the hardware port, I don't have any ideas on making this non-breaking.

But I'd hesitate to add an interface from the SDK that we would otherwise avoid for the sake of making it non-breaking. We could play games such as making it behind a "temporary" feature flag with plans to deprecate along with the serial instantiation, but not sure it's worth it. I'll think on it a bit though.

components/patina_adv_logger/src/hardware_port.rs
cfernald: Well part of the reason I chose `SerialIO` in the first place was that it describes characteristic more then a device. That is, bytes and written/readable one byte at a time, in series. In my mind this trait is basically a slimmed down rust `Read` + `Write` traits. We could consider even renaming it (with the alias to not break users) to something like `ByteIO` if we wanted.

To me the question is, is there value in wrapping the output other than filtering. Additionally when that filtering already intersects with the existing level controls in the advanced logger itself (`hw_print_level`). Perhaps it's better to have callbacks that are more explicitly coupled with those controls.

sdk/patina/src/peripheral/serial/shared.rs
Javagedes: A non-breaking change is to add a filter trait to the AdvancedLogger struct (in addition to the existing SerialIO trait) with a default trait implementation that does not filter anything. But that is conically different than the C implementation even though it does effectively the same thing.

cfernald: I think I generally prefer that approach, just because it allows us to align to other methods of filtering instead introducing a new layer. We already have `hw_print_level` so providing a `.with_hw_print_level_override_callback` seems better aligned to the design anyways.

os-d: Yeah, I've been thinking on this since yesterday and I agree, I think that we can get the same behavior with splitting HdwPort lib into two pieces: One is the filter, the other is the hdw port abstraction, which we already have with the SerialIO abstraction.

joschock: `SerialIo` requires implementors to supply read routines; so it's overspecified for AdvLogger which only wants to write. If I had some write-only transport it would annoying and counter-intuitive to have to implement read stub routines that neither side wants just to satisify the contract; so I liked the approach of a separate HdwPort trait that only stipulated `write` with a blanket implementation for SerialIo (which is clearly what _most_ users are gonna do since it also supports the debugger).  I also think that filtering and output are separate concerns and I support splitting those into two separate traits. We already have a fair degree of filtering available with the present logger, so it's not clear to me what an additional trait buys you.

os-d: We have existing platforms that do runtime filtering of the serial output. An unfortunate legacy of the adv logger design is that the static hw port error level came first, then the hdw port lib that allowed dynamic filtering (and a separate hw port). Both were kept for compat. In Patina, we probably could get rid of the static level, though that is a common use case.

But we have platforms that do things like read a GPIO and if it is asserted, they will emit serial logs. The way they do that today is through the hw port lib. In Patina, they don't have a way to do this. This support was for that specific scenario, but others can be envisioned.

I'm not personally tied to this abstraction point or keeping the fixed hw port level. If we want to always have a dynamic level I think that could work too and clean up a bit.

cfernald: > `SerialIo` requires implementors to supply read routines; so it's overspecified for AdvLogger which only wants to write. If I had some write-only transport it would annoying and counter-intuitive to have to implement read stub routines that neither side wants just to satisify the contract; so I liked the approach of a separate HdwPort trait that only stipulated `write` with a blanket implementation for SerialIo (which is clearly what _most_ users are gonna do since it also supports the debugger). I also think that filtering and output are separate concerns and I support splitting those into two separate traits. We already have a fair degree of filtering available with the present logger, so it's not clear to me what an additional trait buys you.

I was thinking about this yesterday as well. We could create a new `SerialWrite`/`SerialOut` and have an auto-implementation for implementors of `SerialIO` so it's not breaking. This can be done separate from the purpose of this PR though.

os-d:
@cfernald @joschock @Javagedes I have updated this to just be the callback and leaving the SerialIO trait to be the hw port abstraction. If we want to revisit a write only trait for that, we can do that in a separate PR. Let me know your thoughts.

components/patina_adv_logger/src/logger.rs
```diff
@@ -79,12 +80,23 @@ where
...
             memory_log: RwLock::new(None),
             timer: Service::new_uninit(),
         }
     }
 
+    /// Sets a callback that can override the effective hardware print level before each hardware port write.
+    ///
+    /// The callback receives the hardware print level selected from the memory log header or matching target filter
+    /// and returns the level to use. This can be used for dynamic platform filtering to the hw port.
+    #[must_use]
```
Javagedes, line 116: nit: I'd add an example here

components/patina_adv_logger/src/logger.rs
```diff
@@ -79,12 +80,23 @@ where
...
             format,
             memory_log: RwLock::new(None),
             timer: Service::new_uninit(),
         }
     }
 
+    /// Sets a callback that can override the effective hardware print level before each hardware port write.
+    ///
+    /// The callback receives the hardware print level selected from the memory log header or matching target filter
+    /// and returns the level to use. This can be used for dynamic platform filtering to the hw port.
```
cfernald: I would clarify in the comment when this gets called. I.e. gets called at the start of each log message processing.

components/patina_adv_logger/README.md
```diff
@@ -13,6 +13,43 @@ The crate stores records in a shared memory buffer that begins with an `ADVANCED
...
+      0
+   }
+}
+
+static LOGGER: AdvancedLogger<UartNull> = AdvancedLogger::new(
+   Format::Standard,
+   &[],
+   log::LevelFilter::Info,
+   UartNull {},
+).with_hw_print_level_override_callback(platform_hw_print_level);
```
Javagedes: I'd consider showing this example (or maybe the example in the function itself) with a lambda. that seems like the more likely use-case

`.with_hw_print_level_override_callback(|prev| prev & SOME_VALUE)`

components/patina_adv_logger/src/logger.rs
```diff
@@ -79,12 +80,23 @@ where
...
+    #[must_use]
+    pub const fn with_hw_print_level_override_callback(mut self, callback: fn(u32) -> u32) -> Self {
```
Javagedes: nit#2: I'm a bit hesitant to have the word "callback" in the function name as the interface is self explanatory. But I'll leave it up to you if its worth the change or not.

components/patina_adv_logger/src/logger.rs
```diff
@@ -138,9 +150,13 @@ where
...
         let log_guard = self.memory_log.read();
         if let Some(memory_log) = log_guard.as_ref() {
-            hw_write = match hw_print_mask_override {
-                Some(mask) => memory_log.hardware_write_enabled_with_mask(error_level, mask),
-                None => memory_log.hardware_write_enabled(error_level),
+            hw_write = match (hw_print_mask_override, self.hw_print_level_override_callback) {
+                (None, None) => memory_log.hardware_write_enabled(error_level),
+                (mask_override, callback) => {
+                    let hw_print_level = mask_override.unwrap_or_else(|| memory_log.hw_print_level());
+                    let hw_print_level = callback.map_or(hw_print_level, |callback| callback(hw_print_level));
```
cfernald: I think calling at the lowest level might be a bit spammy to the callback since it may be called many times for a single message. I would instead call it once when creating `BufferedWriter` and use the combined override as the passed override level.

Javagedes (approved):

cfernald (approved):

components/patina_adv_logger/src/logger.rs
os-d: Updated and put the same thing in the C entry

PR #1767, opened 2026-09-04
[major] Do Not Allow Non-Page Aligned GCD Allocations

### LINKED ISSUE #1614: [Bug]: Patina Needs to Consider Non-Page Aligned GCD Allocations

### Current Behavior

The GCD accepts non-page aligned allocations (for both address and size). The EFI_MEMORY_MAP requires the size to be page aligned. The allocators ensure page aligned allocations.

However, direct GCD allocations, such as for FV HOBs (which may represent actually non-page aligned regions) may break this expectation. Currently, Patina will not report these in the EFI_MEMORY_MAP.

### Expected Behavior

All available memory should be reported in the EFI_MEMORY_MAP. Patina should have a consistent story for GCD allocations and in particular FV allocations w.r.t. page alignment.

### Steps To Reproduce

n/a

### Build Environment

```markdown
- OS(s):
- Targets Impacted:
```

### Version Information

```text
top of tree
```

### Urgency

Medium

I will fix it

No maintainer feedback needed

### Anything else?

_No response_

os-d:
## Description

This PR fixes #1614 and has three commits. This PR is targeting the major branch because platforms should re-run the readiness tool to confirm FV regions are covered by memory allocation HOBs.

GCD: Do Not Allocate FV HOB Regions
--
Currently, Patina took the stance that it should allocate FVs to make sure memory allocations did not stomp on them. However, this led to an issue where FVs are not page aligned and so non-page aligned allocations were made in the GCD. The EFI_MEMORY_MAP must be page aligned, so these regions could not be added and had to be silently dropped. Patina does not have the information required to know if it is safe to extend these allocations to page aligned lengths.

Originally, Patina inherited behavior from EDK II to only allocate FV regions if they were MMIO. However, MMIO allocations are not used in the core.

It is the domain of the pre-DXE phase to properly describe allocations to DXE so that existing allocations are not stomped on. The PI spec defined way is to do this through memory allocation HOBs. As such, Patina should only rely on memory allocation HOBs to build the GCD, not FV HOBs. It is then a reasonable expectation to say that memory allocation HOBs must be page aligned.

This expectation is also testable and added to the patina-readiness-tool.

GCD: Enforce Page Aligned Allocations
--
The GCD must have page aligned allocations because the EFI_MEMORY_MAP is built from it, which must be page aligned per spec (and the structure only describes ranges in page count length). In addition, memory protections rely on page aligned regions.

This commit enforces this expectation.

Revert "dxe: GCD: Don't Assert on Non-Page Aligned GCD Allocations"
--
This reverts commit 7d883949cb5d6e4c2e6aa66faa6283af9618c001. This commit allowed for non-page aligned GCD allocations, but that allowance is no longer made because of the two preceding commits.

- [x] Impacts functionality?
- [ ] Impacts security?
- [x] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Through updating the patina-readiness-tool and confirming these assumptions are valid.

## Integration Instructions

Platforms should re-run the patina-readiness-tool v0.4.0 or greater when about to ingest this change. They should confirm that FV HOBs are covered by memory allocation HOBs. If not, page aligned memory allocation HOBs must be produced by the platform for the FV regions.

Separately, the platform should identify if any non-page aligned GCD allocations are being made (debug_asserts will catch this). If so, they must convert them to page aligned allocations. This does not affect AllocatePool allocations, only direct GCD allocations through AllocateMemorySpace().

### DIFF
--- a/patina_dxe_core/src/allocator.rs
+++ b/patina_dxe_core/src/allocator.rs
@@ -1173,52 +1173,6 @@ fn process_hob_allocations(hob_list: &HobList) {
                     continue;
                 }
             }
-            Hob::FirmwareVolume(hob::FirmwareVolume { header: _, base_address, length })
-            | Hob::FirmwareVolume2(hob::FirmwareVolume2 {
-                header: _,
-                base_address,
-                length,
-                fv_name: _,
-                file_name: _,
-            })
-            | Hob::FirmwareVolume3(hob::FirmwareVolume3 {
-                header: _,
-                base_address,
-                length,
-                authentication_status: _,
-                extracted_fv: _,
-                fv_name: _,
-                file_name: _,
-            }) => {
-                log::trace!("[{}] Processing Firmware Volume HOB:\n{:#x?}\n\n", function!(), hob);
-
-                // EDK II assumes all FVs are MMIO, however many platforms do not report the
-                // backing address space as an MMIO resource, causing the allocation to silently fail.
-                // Instead of assuming MMIO here, retrieve the backing memory type from the GCD if it
-                // exists and allocate the FV region using that type. Note: there are often multiple
-                // FV HOBs for the same FV. This will only allocate for the first HOB since
-                // get_memory_descriptor_for_address() is filtering for free descriptors.
-
-                //The 4K granularity rule does not apply to FV hobs, so allocate_pages cannot be used.
-                //This means they must be direct-allocated in the GCD, and no stats will be tracked for them.
-                match GCD.get_memory_descriptor_for_address(*base_address, |_, allocated| !allocated) {
-                    Ok(desc) => {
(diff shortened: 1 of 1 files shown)


os-d:
I need to come back and update the docs as well

joschock (approved):

makubacki (approved):
Looks good pending the doc update.

os-d:
@makubacki @joschock I added the doc updates and had to fix a bunch of unit tests that were doing non-page aligned GCD allocations.

cfernald (approved):

os-d:
I added one commit to fix a bug in is_efi_memory_map_descriptor that I noticed when adding tests to get codecov to pass.

PR #1768, opened 2026-09-05
[REBASE & FF] [feature/supv] Fix Ring 3 stack indexing, sizing, and user mapping

vineelko:
## Description

-  **Pass the dense CPU index into the runtime request path**: `enter_runtime`
   received the APIC ID and forwarded it to `bsp_request_loop` as a CPU index.
   On platforms whose APIC IDs are not `0..N-1` (observed: 0, 2, 4 … 46, BSP =
   16) this selected a Ring 3 stack outside the array the MM IPL provisioned, so
   the demoted routine faulted writing to a supervisor-owned page. The dense
   index the C core hands to `entry_point` is now used for the request path,
   while the APIC ID is still used for the `CpuManager` APIs keyed by it.

---

-  **Size the syscall interface from the MP Information HOB CPU count**:
   `SyscallInterface::init` was given the `MAX_CPUS` const generic capacity (32)
   rather than the number of CPUs the IPL provisioned stacks for (16), so
   `get_cpl3_stack` returned pointers past the end of that array, into
   supervisor memory, for any index in between. The MP Information HOB is now
   parsed before the PassDown HOB and its count is passed in, so out of range
   indices are rejected with `InvalidCpuIndex`.

---

-  **Map the Ring 3 stacks user-accessible during BSP init**: the supervisor
   relied on the MM IPL to leave the per-CPU CPL3 stacks user-accessible. The
   provisioned range is now mapped read/write, non-executable and
   user-accessible during BSP init, after confirming it lies inside MMRAM so
   nothing outside MMRAM is exposed to Ring 3, and every page is verified
   afterwards so a misconfiguration fails at init instead of at the first
   demotion.

---

- [x] Impacts functionality?
- [x] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- Validated on internal hardware platform.

## Integration Instructions

N/A

### DIFF
--- a/patina_mm_supervisor/src/init.rs
+++ b/patina_mm_supervisor/src/init.rs
@@ -651,17 +651,19 @@ impl<P: PlatformInfo, const MAX_CPUS: usize> MmSupervisorCore<P, MAX_CPUS> {
         let hob_list_info =
             unsafe { (hob_list as *const PhaseHandoffInformationTable).as_ref().ok_or(PolicyInitError::NullHobList)? };
 
-        // 1. Process the PassDown HOB (policy, syscall, memory policy)
+        // 1. Process the MP Information HOB (`gMpInformationHobGuid`) for the CPU count. It sizes
+        //    the Ring 3 stack array the PassDown HOB describes, so it is needed first.
+        let mp_information =
+            find_guid_hob(hob_list_info, crate::MP_INFORMATION_HOB_GUID).ok_or(PolicyInitError::HobNotFound)?;
+        let number_of_cpus = self.parse_mp_information_hob(mp_information).ok_or(PolicyInitError::InvalidPolicyData)?;
+
+        // 1b. Process the PassDown HOB (policy, syscall, memory policy)
         let pass_down_data =
             find_guid_hob(hob_list_info, crate::MM_SUPV_PASS_DOWN_HOB_GUID).ok_or(PolicyInitError::HobNotFound)?;
         // SAFETY: `pass_down_data` is a slice into the validated HOB list, so the buffer pointers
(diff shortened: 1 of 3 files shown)


kuqin12 (approved):

patina_mm_supervisor/src/init.rs
```diff
@@ -439,6 +439,55 @@ impl<P: PlatformInfo, const MAX_CPUS: usize> MmSupervisorCore<P, MAX_CPUS> {
...
+    ///
+    /// A demoted routine faults on its first push if its stack is supervisor-owned. The range
+    /// covers the `num_cpus` stacks the MM IPL provisioned, which is also the range
+    /// [`SyscallInterface::get_cpl3_stack`](crate::privilege_mgmt::syscall_setup::SyscallInterface::get_cpl3_stack)
+    /// hands out.
+    fn map_cpl3_stacks_to_user(&self, base: u64, per_core_size: u64, num_cpus: u64) {
+        if base == 0 || per_core_size == 0 {
+            log::error!("PassDown HOB does not describe a CPL3 stack region, skipping user remap");
+            return;
+        }
```
makubacki, line 454: In the bigger picture, is this expected? Is there a reason to not treat this as an actual error at this point (instead of just logging and moving on)?

vineelko: Good catch. Yaa this should be panic. Will fix it.

vineelko: Resolved the issue in the latest iteration.

makubacki (approved):

PR #1776, opened 2026-09-11
patina_dxe_core/misc_boot_services: Treat protocol pointer as static in tests

makubacki:
## Description

Right now, `metronome` is a stack-local variable that the test points `METRONOME_ARCH_PTR` at. Because `METRONOME_ARCH_PTR` is a process-global (in the unit test process) with no reset, the stack-local `metronome` will dangle for every later test once this function returns.

This commit makes it `static` so the underlying protocol buffer remains valid for the rest of the the test process' execution.

The same is also done for `WatchdogProtocol`.

This is not a recent regression and has been an issue since the code was added in c3fb83c.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- `cargo make all`

## Integration Instructions

- N/A - Only modifies unit tests


os-d (approved):

patina_dxe_core/src/misc_boot_services.rs
```diff
@@ -489,11 +489,11 @@ mod tests {
...
                 unimplemented!()
             }
-            let watchdog =
+            static WATCHDOG: protocol::watchdog::WatchdogProtocol =
                 protocol::watchdog::WatchdogProtocol { register_handler, set_timer_period, get_timer_period };
-            // SAFETY: The mock protocol lives for the duration of the test and the pointer is only used by the test.
+            // SAFETY: WATCHDOG is a 'static, so the pointer remains valid for the rest of the process.
             unsafe {
-                WATCHDOG_ARCH_PTR.init(&raw const watchdog as *mut c_void);
+                WATCHDOG_ARCH_PTR.init(&raw const WATCHDOG as *mut c_void);
```
vineelko, line 496: NIT: `(&WATCHDOG).cast()` should be good enough. We can drop the raw borrow and the cast to a mutable pointer(which I believe is UB with static)

vineelko (approved):

patina_dxe_core/src/misc_boot_services.rs
makubacki: Let me know if I misunderstood, but `cast()` is a pointer method and `WATCHDOG` is `WatchdogProtocol` type. So, `&WATCHDOG` does not become a pointer unless that's explicitly done.

That could be done with `core::ptr::from_ref()` like:

`WATCHDOG_ARCH_PTR.init(core::ptr::from_ref(&WATCHDOG).cast_mut().cast());`

But that's essentially a more verbose form what's there now. I tried it and got:

```
error[E0599]: no method named 'cast' found for reference '&WatchdogProtocol'
```

vineelko: My bad I should have tested it, the intent is not create a mutable pointer. Yes raw borrow is still required. Instead of casting it to mutable pointer, const pointer is sufficient. `&raw const WATCHDOG as *const c_void`

makubacki: The issue there is that `ArchProtocolPtr<T>.init()` is currently defined as:

```rust
unsafe fn init(&self, ptr: *mut c_void) 
```

So, I'd need to change that to `ptr: *const c_void` which starts impacting non-test code which I'm hesitant to do as it expands beyond the scope of this original change.

vineelko: Makes sense. Resolving the comment.

PR #1777, opened 2026-09-13
Add support for MP Services [FF & REBASE]

cfernald:
## Description

Adds UEFI MP Services support to DXE Core and the architecture-independent processor-control interface in `patina_internal_cpu`.

This implementation consumes the PEI MP handoff, brings inherited APs into a Patina-managed dispatch environment, exposes the UEFI MP Services Protocol, synchronizes architectural state, recovers APs whose procedures time out, and moves APs into an isolated parking environment for `ExitBootServices`.

For details on the implementation design, see the added `mp_services.md` file.

This implementation only supports x64 for the initial pass, but is intended for AArch64 to be added soon after.

### Key High-Level Structures:

- DXE Core:
    - `services.rs` - This is the primary implemented for the DXE core layer.
    - `dispatch.rs` - Controls the tracking information for a AP dispatch.
- CPU Internal:
    - `control.rs` - The implementation for the producer and consumer AP state machine.
    - `work.rs` - Implementation for the packetized function/context for the AP.
    - `mp.rs` - Generic trait and interface definitions for architecture abstraction.

### Differences from EDK2

This implementation is completely different in design, but also intentionally differs from EDK2 behavior in a few ways.

1. X64 implementation does not support INIT-SIPI-SIPI start of APs. This is more efficient but has the following effects.
    - X64 APs are only supported when described fully through HOBs and parked in a long-mode busy/mwait loop.
    - AP reset due to abort uses an NMI and re-entry instead of a INIT-SIPI-SIPI.
2. BSP switching is not supported. This could be added in the future.
3. On x64, uses a bounded amount of park pages by overloading the stack. This can be done by never actually reading pushed contents.

## PR Tags

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [x] Includes documentation?

## How This Was Tested

- [x] Unit tests
- [x] Integration tests on Q35 
- [x] Integration tests on Intel device
- [x] Boot to Windows on Q35 
- [x] Boot to Windows on Intel device

## Integration Instructions

Consumers who wish to use this implementation must drop their EDK2-style MP driver if using one, and override `enable_mp_services` in the platform `CpuInfo` to return `true`.


patina_dxe_core/src/cpu/mp_services.rs
```diff
@@ -0,0 +1,255 @@
+//! MP Services Protocol installation.
```
Javagedes, line 1: I don't like that this is enabled via a feature flag.

Is it possible to:

1. Pull this out into a fully standalone component (in this repo) that a platform can elect to register if they prefer?
2. Have some type of in-code configuration to enable or disable this rather then a feature flag?

cfernald: I debated pulling this out into its own crate and component, but the main problem is that there are some couplings between the core and MP implementation, around things like GDT, MTRRs, page table management, etc. that I think makes it more sustainable to keep coupled with the core. Additionally, this allows us to divide the architectural layer into `patina_internal_cpu`.

As for the platform controlling when the core installs the component, I'm happy to move this to something like `enable_mp_services` in `CpuInfo`. Does that work for you?

Javagedes: yes, that works for me personally :)

cfernald: Switched to `CpuInfo` config

core/patina_internal_cpu/src/interrupts/x64/idt.rs
```diff
@@ -30,50 +36,71 @@ struct IdtEntry {
...
 /// The 256-entry x86-64 Interrupt Descriptor Table.
 #[repr(C, align(16))]
-struct Idt {
+pub(crate) struct Idt {
     entries: [IdtEntry; 256],
 }
 
+impl Idt {
+    pub(crate) const fn filled(entry: IdtEntry) -> Self {
+        Self { entries: [entry; 256] }
```
os-d, line 66: nit: use a const for 256 everywhere

core/patina_internal_cpu/src/mp/x64/park.rs
```diff
@@ -0,0 +1,288 @@
...
+/// Reserved entry address consumed by the handoff assembly when no AP context exists.
+#[unsafe(no_mangle)]
+pub(super) static AP_PARK_ENTRY: AtomicU64 = AtomicU64::new(0);
+
+/// Reserved configuration address consumed by the handoff assembly.
+#[unsafe(no_mangle)]
+pub(super) static AP_PARK_CONFIG: AtomicU64 = AtomicU64::new(0);
+
+fn page_address(base: usize, index: usize) -> usize {
+    base + index * UEFI_PAGE_SIZE
```
os-d, line 75: nit uefi_pages_to_size! here and elsewhere, though you also have an inconsistency with the mut version where you do checked_mul

core/patina_internal_cpu/src/mp/x64/park.rs
```diff
@@ -0,0 +1,288 @@
...
+    let park_data = ParkData {
+        cr3: page_tables as u64,
+        gdtr: DescriptorTablePointer {
+            limit: (core::mem::size_of::<[u64; 8]>() - 1) as u16,
+            base: (data + core::mem::offset_of!(ParkData, gdt)) as u64,
+        },
+        idtr: DescriptorTablePointer { limit: (core::mem::size_of::<Idt>() - 1) as u16, base: idt as u64 },
+        stack_top: (data + UEFI_PAGE_SIZE) as u64,
+        parked_count: AtomicU32::new(0),
+        gdt: crate::gdt::minimal_long_mode_entries(),
```
os-d, line 141: Discussed offline, but dropping here, consider having a separate TSS descriptor per AP for the future case so we don't hit the issue of installing the GDT and tripping on the TSS busy bit being set.

core/patina_internal_cpu/src/mp/x64/park.rs
```diff
@@ -0,0 +1,288 @@
...
+            }
+
+            let page = self.next;
+            self.next += UEFI_PAGE_SIZE;
+            self.remaining_pages -= 1;
+            Ok(page as u64)
+        }
+    }
+
+    // Create an allocator patina-paging to use for the remaining non-root tables.
```
os-d: for patina-paging

os-d (commented):
In general, this approach seems reasonable to me. Obviously we need physical platform testing. I think we also need documentation (and please add a nice mermaid chart showing the state machine :), which also should include descriptions on the expected pre-DXE handoff state (and how to achieve it in edk2) and what the OS handoff state is. Also, we should update the readiness tool to check all the validation you do at the beginning here. This would be a case where we can only do that in the DXE core version of the readiness tool

core/patina_internal_cpu/src/mp/x64/park.rs
cfernald: Updated with unique TSS per AP, pointing back to their original stack

os-d (commented):
As I said, I think these changes are reasonable, but I'd like to see physical platform results before we merge them in.

For the docs, I think we need to make sure MpDxe isn't referenced anywhere and in the main dxe_core integration doc, point to this doc for MP requirements, as that is our one stop shop for integration requirements. Similarly getting the additional checks up in the readiness tool would be good.

cfernald:
> As I said, I think these changes are reasonable, but I'd like to see physical platform results before we merge them in.
> 
> For the docs, I think we need to make sure MpDxe isn't referenced anywhere and in the main dxe_core integration doc, point to this doc for MP requirements, as that is our one stop shop for integration requirements. Similarly getting the additional checks up in the readiness tool would be good.

This has been validated on a physical Intel platform, both with the patina_tests and booting to Windows.

PR #1778, opened 2026-09-14
[major] Add and integrate ServiceCell<T> [Rebase & FF]

makubacki:
## Description

A series of commits for adding a new type `ServiceCell<T>` in place of `Service::new_unit()`. This was already part of approved changes in https://github.com/OpenDevicePartnership/patina/pull/1743, but broken out here for clarity.

---

**sdk: Add ServiceCell<T> for bridging services to context-free callbacks**

extern "efiapi" functions backing EDK II protocols carry no context
parameter, so a component can't hand them a `Service<T>` the way it
can a closure-based callback. `ServiceCell<T>` gives components a
wait-free, write-once cell for this. The `publish()` function never
blocks even under contention, so it can't deadlock under preemption
like other mechanisms.

---

**sdk: Impl Debug for Service<T>**

Implements the `Debug` trait for `Service<T>` to allow the type to be
used in contexts where `Debug` is required.

Since the `value` field in `Service<T>` is `OnceCell<&'static dyn Any>`
and `dyn Any` does not implement `Debug`, the implementation here
is custom (instead of deriving `Debug`) and just prints the type
name of the service interface `T`.

---

**Swap Service::new_unit() usage to ServiceCell<T>**

Updates code using `Service::new_unit()` to use `ServiceCell<T>` to
align on a single approach for safely storing services accessed in
different contexts.

---

**sdk: Drop Service::new_uninit()**

There is nothing wrong with `Service::new_uninit()` for the use case
it directly supports, that allows a `Service<T>` to be created before
the underlying service is available.

For example, it allows `AdvancedLogger` to be const-constructible
before the actual component dispatch system is available. Then the
component entry point can call `.replace()` to safely fill in the
actual service implementation once the service is ready.

However, there is also a need to more generically support service
storage for cases when an `extern "efiapi"` protocol function produced
by a component and the protocol interface does not have a `this`
context pointer. Using `Service<T>` for that case is certainly
possible, but it starts to mix two different use cases in a single
type.

`Service<T>` relies heavily on `Deref` to provide a convenient
interface to the underlying service, but that convenience can be
dangerous when the service is not yet available. It is easy to
forget to check for availability before dereferencing, which can
lead to panics. `Deref::deref()` has to return `&Target`, it cannot
return `Option<&Target>`, so there is no way to signal that the
service is not yet available.

Of course, `is_init()` or `map_or_else()` could be used, but it
is safer and makes `Service<T>` easier to reason about if it
simply represents an available service. Then, dereferencing a
service is always safe (the common case).

This aligns with official `Deref` implementation guidance
(https://doc.rust-lang.org/std/ops/trait.Deref.html#when-to-implement-deref-or-derefmut):

```
In general, deref traits should not be implemented if:

1. the deref implementations could fail unexpectedly; or
2. the type has methods that are likely to collide with methods on the target type; or
3. committing to deref coercion as part of the public API is not desirable.
```

For these cases, having a dedicated type that does not implement
`Deref` allows the type to clearly distinguish between an
uninitialized service and an initialized one forcing users to
handle the uninitialized case explicitly while also making the
usage pattern more clear.

The `ServiceCell<T>` moves these semantics into a separate type
that allows the value to set once (wrapping `OnceCell<T>`) and then
read as an `Option<&T>` (using `get()`). Existing callers that want
to panic can simply call `.get().expect("service not published yet")`
to get the same behavior and provide a custom message at the call
site. The `ServiceCell<T>` type is also `Sync` and compare-exchange
guarded, so a second call to `publish()` will return an error instead
of racing the first.

---

**Remove clone calls on Service<T> instances**

Now that `Service<T>` implements `Copy`, explicit calls to `clone()`
are no longer necessary and can be removed.

Also prevents the `clippy::clone_on_copy` lint from warning about
unnecessary clones.

---

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- `cargo make all`
- Boot to EFI shell on QEMU ArmVirt & SBSA with all components enabled

## Integration Instructions

- N/A


makubacki:
Note: CI is currently broken by an unrelated change picked up from patina-paging in https://github.com/OpenDevicePartnership/patina-paging/commit/16ac9d0d6d5cfc198b984132edf535e09e6fc94d

os-d:
> Note: CI is currently broken by an unrelated change picked up from patina-paging in [OpenDevicePartnership/patina-paging@16ac9d0](https://github.com/OpenDevicePartnership/patina-paging/commit/16ac9d0d6d5cfc198b984132edf535e09e6fc94d)

I will have a PR up for that shortly, just merged an automock

Javagedes (approved):

os-d (approved):

makubacki:
Note: Waiting on https://github.com/OpenDevicePartnership/patina/pull/1781 to be completed before merging.

sdk/patina/src/component/service/cell.rs
```diff
@@ -0,0 +1,270 @@
...
+//!
+//! struct MyServiceImpl;
+//!
+//! impl MyService for MyServiceImpl {
+//!     fn value(&self) -> u32 {
+//!         42
+//!     }
+//! }
+//!
+//! static CURRENT: ServiceCell<dyn MyService> = ServiceCell::new();
```
cfernald, line 54: While I like the ServiceCell and I don't want to block the PR with a comment discussion, I do wonder if we should consider a different mechanisms for things like efiapi calls which call back for a static instance.

Why have every component/service/callback have to store their own copy of a service when that information is already readily available in the core. Should we instead provide a method (either static or auto-implemented for components to be mockable) by which they can dynamically request a service?

I realize this side-steps the purpose of dep-inj, but for instances such as `CURRENT` here where we are already going back to a static, it seems reasonable that we instead use the ones we already have. Now these things are not mutually exclusive, so I don't want to block the PR. We can take this discussion elsewhere as needed.

@Javagedes for thoughts.

makubacki: It would be a useful discussion to have, please include me in it if it happens. But it'd help to better identify the underlying problem being solved. To some extent, I think adhering to dependency injection more strictly prevents some of the loosely ordered dispatch designs that led to using a priori as a crutch.

sdk/patina/src/component/service/cell.rs
```diff
@@ -0,0 +1,270 @@
...
+    sync::atomic::{AtomicBool, Ordering},
+};
+
+use super::Service;
+
+/// A wait-free, write-once-then-read-many cell for a [`Service<T>`](super::Service).
+///
+/// See the [module documentation](self) for the motivation and the reasoning behind its design.
+pub struct ServiceCell<T: ?Sized + 'static> {
+    value: OnceCell<Service<T>>,
```
cfernald, line 88: Is there a reason to not just use `OnceLock` from the spin crate to avoid duplicating the protection logic?

sdk/patina/src/component/service/cell.rs
```diff
@@ -0,0 +1,270 @@
...
+        }
+        Err(service)
+    }
+
+    /// Returns the published service, if any.
+    ///
+    /// Never blocks. Returns `None` both when nothing has been published yet and while a concurrent call to
+    /// `publish` is in progress.
+    pub fn get(&self) -> Option<Service<T>> {
+        if self.writing.load(Ordering::Acquire) { None } else { self.value.get().copied() }
```
cfernald: I'm not sure if this is strictly safe. In the worst-case scenario, if this value is larger then a single aligned u64, couldn't the copy get interrupted mid-read where this would be accessible, no? I would instead consider a 3-value state of `state_uninit`, `state_writing`, `state_done` with compare exchange transitions, where the access will only occur if state == `state_done` (Or using OnceLock as mentioned above)

sdk/patina/src/component/service/cell.rs
cfernald: Yeah, my concern would be folks using such a mechanism when they should instead be using the dependency injection. In the case of going to the static, they are already breaking that, or at least at least deferring that to the entity that setup the callback so getting it from the one-off static vs the core seems equivalent there.

I will resolve this for now. We can continue discussing, but is not blocking for this PR.

sdk/patina/src/component/service/cell.rs
makubacki: `OnceLock` itself is not available in `no_std`. The other reason is that `OnceLock::set()` docs say, "May block if another thread is currently attempting to initialize the cell." and the design here is wait-free (immediately returning an error) with the compare/exchange serializing writers.

sdk/patina/src/component/service/cell.rs
makubacki: On the first publish, I think that's a potential issue and I added the three-state solution. Can you please check it and resolve the conversation if it matches what you had in mind?

sdk/patina/src/component/service/cell.rs
cfernald: Sorry, I meant Spin's `Once`: https://docs.rs/spin/latest/spin/type.Once.html

My memory failed me on the name, though it will still block on set, so your point stands. Resolving this.

cfernald (approved):

sdk/patina/src/component/service/cell.rs
cfernald: Looks good, thanks!

PR #1779, opened 2026-09-14
Consume automocks from patina-paging and patina-mtrr

os-d:
## Description

The two traits from these crates now produce automocks, consume these instead of directly mocking, which makes patina more resilient to non-breaking changes to these traits.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Running unit tests.

## Integration Instructions

N/A.


cfernald (approved):

core/patina_internal_cpu/src/paging/x64.rs
```diff
@@ -292,4 +250,18 @@ mod tests {
         let paging = EfiCpuPagingX64 { paging: MockPageTable::new(), mtrr: MockMtrr::new() };
         paging.handle_cacheability_change(0x1000, 0x1000, MemoryAttributes::Writeback, MemoryAttributes::Uncached);
     }
+
+    #[test]
+    fn test_mock_mtrr_memory_ranges() {
```
makubacki: I think the test itself is fine, but it seems like it belongs more in patina-mtrr than here.

core/patina_internal_cpu/Cargo.toml
```diff
@@ -31,6 +31,8 @@ arm-gic = { workspace = true }
 
 [dev-dependencies]
 mockall = { workspace = true }
+patina_mtrr = { workspace = true, features = ["mockall"] }
```
makubacki: Could this be gated with `[target.'cfg(all(target_arch="x86_64"))'.dependencies]` like patina_mtrr is above? It's not an issue now, just helps keep dependencies tracked a little more closely.

makubacki (approved):

core/patina_internal_cpu/src/paging/x64.rs
os-d: Yeah, dropped.

PR #1780, opened 2026-09-15
Paging: Handle MTRRs Unsupported

os-d:
## Description

MTRRs may be unsupported on a platform. This is the case for Intel TDX confidential VMs, for instance. In this case, the paging attributes should still be programmed when calling map_memory_region, even if the caching attributes can't be changed.

Query is the only other API that touches MTRRs and it already handles unsupported (it will report as uncacheable).

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on an enlightened TDX VM.

## Integration Instructions

N/A.


core/patina_internal_cpu/src/paging/x64.rs
```diff
@@ -52,7 +52,9 @@ where
         }
 
         match apply_caching_attributes(address, size, cache_attributes, &mut self.mtrr) {
-            Ok(()) => self.paging.map_memory_region(address, size, attributes & MemoryAttributes::AccessAttributesMask),
+            Ok(()) | Err(EfiError::Unsupported) => {
```
cfernald, line 55: Will this not cause functional problems, say for direct assigned devices, if the memory is not actually being made uncached?

os-d: I believe that is the realm of the hypervisor to maintain the cacheability as part of the direct assignment.

image

Where the Shared EPT is the page table managed by the VMM.

At any rate, on the TDX VMs I am working on, MTRRs return unsupported. There is a way to emulate them, but that is not enabled.

cfernald (approved):
Hopefully this is just a temporary change and eventually this falls back to PAT, but seems reasonable for now if unsupported by the hardware.

makubacki (approved):

PR #1781, opened 2026-09-15
[major] Defer event and protocol closures in self and nested close scenarios

makubacki:
## Description

While in practice it is difficult to cancel your own event or protocol install notification since the closures passed to functions that return the event handle or protocol notification registration include the handle in the return, it is possible.

These commits explicitly prevent an active event or protocol notification context from being cancelled/closed while it is active.

TLDR: These changes have the trampoline functions that wrap actual callbacks free things when control returns to them if something was cancelled/closed during the notify call stack that was still active.

---

**patina_dxe_core: Defer event closures**

Note: This can be squashed into commit 98e6dc3 when the major branch
is merged into main.

Currently, events are dispatched in `restore_tpl()` by calling
`EVENT_DB.consume_next_event_notify()`. Note that the event is only
actually removed from `EVENT_DB` by calling `close_event()`.

When `EventServices` are used, the actual "raw event notification"
is a wrapper (`notify_trampoline()`) that creates a `ClosureHolder`
that contains the actual notification function registered.

Once the closure is executing, if it (or a nested call within the stack),
calls `close_event()` on its own event, the `close_event()` will remove
the event from `EVENT_DB` immediately, which frees the `ClosureHolder`
(and boxed closure). The issue is that the closure might still be
executing earlier in the call stack.

Control will return to the closure body (after the `close_event()`
call) and any use of captured state in the closure is no longer valid.

This change shares a small set of state between `notify_trampoline()`
and `close_event()` to track whether the closure is still active,
allowing `close_event()` to determine if it is safe to free the event.

Active notify contexts are stored in `ACTIVE_NOTIFY_CONTEXTS`. Now,
`notify_trampoline()` will push its context before invoking the closure
and pop it after. `close_event()` can mark a given notify context for
closure and `notify_trampoline()` will free the `ClosureHolder` if it
sees that request.

To support this, `close_event()` now searches active notify contexts
in `ACTIVE_NOTIFY_CONTEXTS` to determine if a closure is still active
(includiing ancestors). If it is, it defers the free until execution
gets back to `notify_trampoline()` which is after the closure has
finished executing.

---

**patina_dxe_core: Defer protocol notification closures**

Note: This can be squashed into commit 98e6dc3 when the major branch
is merged into main.

A protocol `NotifyHolder` could be freed while the callback is still
executing if the notify function cancels its own registration.

In this case, `notify_install_trampoline()` is what's actually called
when the watched protocol is installed. This change updates it to
push its context before starting to call notification functions and
then to pop its context after all notification functions have been
called to determine if its registration was requested to be closed.

`cancel_install_notify()` is updated to be able to set the
`close_requested` flag on the active notification context and defer
dropping the `NotifyHolder` if the notification is still active.

---

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- `cargo make all`
- Updated patina_samples to reproduce self-cancel scenarios. When reproduced, boot continued as normal but that was because nothing occurred to immediately trigger a safety violation like a bad pointer dereference, the logical issue still exists.

## Integration Instructions

- N/A


patina_dxe_core/src/uefi_services/event.rs
```diff
@@ -23,24 +24,64 @@ use patina::error::EfiError;
...
+}
+
+// SAFETY: `ActiveNotify` values are only ever pushed, inspected, and popped while holding
+// `ACTIVE_NOTIFY_CONTEXTS`'s lock.
+unsafe impl Send for ActiveNotify {}
+
+/// Notification contexts currently executing, used to detect a callback closing its own event (or
+/// an ancestor event, if closed from a nested notification) instead of freeing the closure while
+/// it is still on the call stack. See [`notify_trampoline`] and [`CoreEventServices::close_event`].
+static ACTIVE_NOTIFY_CONTEXTS: TplMutex<Vec<ActiveNotify>> =
```
Javagedes: Can we track this information without these couple of statics? maybe as a field in the core event services structure or something?

os-d (approved):

patina_dxe_core/src/uefi_services/event.rs
makubacki: I agree it'd be nice to not grow the number of global statics. There are a few options, but I'm leaning toward a new, single static `UEFI_SERVICES_STATE` global as explained below.

- `CoreProtocolServices` is leaked and would technically have a `'static` address but the code that needs to use it is extern efiapi and doesn't have direct access to that state. That could be overcome (I think) by updating the buffer it gets (which is currently a boxed `NotifyHolder`) to have a reference/pointer, but the function that builds that box (`register_install_notify()`) would have to convert its `&Self` parameter `'static` which would require an unsafe cast and break some assumptions in tests.

- Other existing statics like `EVENT_DB` and `PROTOCOL_DB` are accessible but `PROTOCOL_DB` locks access to the protocol database at `TPL_NOTIFY` and I think we have situations where an access here might come from `TPL_HIGH` (for example, closing an event or cancelling a protocol notify from a `TPL_HIGH` function).

There is also `Core<P>`, but I don't think any of this needs generic platform info and it's kind of internal to "UEFI Services" as opposed to something that needs to be shared across the entire core.

I think the code in each file could be consolidated into a type like `ActiveDispatchStack` with methods that are generically used for events and protocols. Then, a `UefiServicesState` struct could be used to hold all of this, including the config table map:

```rust
pub(crate) struct UefiServicesState {
    pub(crate) event_notify: ActiveDispatchStack,
    pub(crate) protocol_install_notify: ActiveDispatchStack,
    pub(crate) config_table_types: TplMutex<BTreeMap<BinaryGuid, TypeId>>,
}
```

And that could be a single static:

```rust
pub(crate) static UEFI_SERVICES_STATE: UefiServicesState = UefiServicesState::new();
```

Let me know your thoughts.

makubacki: To reduce turnaround time for these changes, I've started implementing this. I think it's cleaner than what's currently in the PR and it's not exposed outside the crate, so we can refactor further in the future internal to the crate if needed.

makubacki: Update pushed.

Javagedes: I think the unification to reduce the number of statics is good and we can merge it as you've already done a lot of work that we should get in.

But the end state is to get rid of all statics. There is a distinct difference between leaking something as static, and having a static that ends up in the .data / .bss section - specifically around unit testing.

With the introduction of next-test maybe we are deciding statics are fine and we should roll back some of the work I've done to reduce statics in the core... but in my eyes its an all or nothing type thing regarding global statics.

makubacki: We talked about this offline a bit and the conclusion is to leave this code as-is and revisit the static reduction effort later. There are tradeoffs throughout the code and DXE core subsystems where way may complicate type management (carrying generics forward) without much practical gain (given nextest) and we may want to utilize context in some cases to reduce the need to access a global. Resolving this converstation thread.

makubacki:
Pushed a small change to bump patch code coverage from 78% to 100%.

makubacki:
I have another change dependent on this one, that I would like to get into `major` before continuing on with overall UEFI Services changes. Another review here when someone has time would be appreciated.

cfernald (approved):

PR #1784, opened 2026-09-15
Add Aliased Mapping Protocol

os-d:
## Description

This adds a protocol exposing the map_aliased_memory_region API in patina-paging. This is gated behind a
confidential_compute feature flag as only these platforms are expected to use this protocol; Patina should not expose this globally.

This also adds integration documentation for confidential VMs, which is currently incomplete, but subsequent PRs will update.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

On an enlightened TDX guest.

## Integration Instructions

Follow the instructions in the documentation.


cfernald (approved):

kuqin12 (approved):

patina_dxe_core/src/gcd/spin_locked_gcd.rs
```diff
@@ -2200,6 +2213,67 @@ impl SpinLockedGcd {
...
+        virtual_address: u64,
+        physical_address: u64,
+        len: u64,
+        attributes: u64,
+    ) -> Result<(), EfiError> {
+        let paging_attrs = MemoryAttributes::from_bits_truncate(attributes)
+            & (MemoryAttributes::AccessAttributesMask | MemoryAttributes::CacheAttributesMask);
+        self.aliased_mappings.lock().try_reserve(1).map_err(|_| EfiError::OutOfResources)?;
+        let mut page_table_guard = self.page_table.lock();
+        let page_table = page_table_guard.as_mut().ok_or(EfiError::NotReady)?;
```
makubacki: If I understand correctly, I think you should hold `self.page_table.lock()` first with its guard and then `self.aliased_mappings.lock()`.

patina_dxe_core/src/gcd/spin_locked_gcd.rs
```diff
@@ -2200,6 +2213,67 @@ impl SpinLockedGcd {
...
+
+        page_table.map_aliased_memory_region(virtual_address, physical_address, len, paging_attrs).map_err(|err| {
+            match err {
+                PtError::OutOfResources => EfiError::OutOfResources,
+                PtError::NoMapping => EfiError::NoMapping,
```
makubacki: In `efierror_to_pterror()` `PtError::NoMapping` is mapped to `EfiError::NotFound`. Is it intentional to map to different EFI status codes for the same case?

patina_dxe_core/src/gcd/spin_locked_gcd.rs
makubacki: If I understand correctly, I think you should hold `self.page_table.lock()` first and then `self.aliased_mappings.lock()`.

patina_dxe_core/src/cpu.rs
```diff
@@ -8,6 +8,8 @@
 //!
 //! SPDX-License-Identifier: Apache-2.0
 //!
+#[cfg(feature = "confidential_compute")]
```
makubacki: Should this be `#[cfg(any(test, feature = "confidential_compute"))]`?

I'm not sure this feature will always be a part of test runs and there are a few unit tests in `aliased_memory_map_protocol`.

patina_dxe_core/src/gcd/spin_locked_gcd.rs
makubacki, line 1: spin_locked_gcd.rs is around 7,500 lines long now. Do you think the aliased mapping code could be moved to its own file?

patina_dxe_core/src/gcd/spin_locked_gcd.rs
```diff
@@ -2200,6 +2213,67 @@ impl SpinLockedGcd {
...
+                _ => EfiError::InvalidParameter,
+            }
```
makubacki, line 2236: In `efierror_to_pterror()` `PtError::NoMapping` is mapped to `EfiError::NotFound`. Is it intentional to map to different EFI status codes for the same case?

Also, is there a reason to have a single `impl From<PtError> for EfiError`?

makubacki (approved):

patina_dxe_core/src/gcd/spin_locked_gcd.rs
os-d: I agree this should be split up, but I don't think splitting out the aliased mapping code on its own is the best, I think maybe the entire paging code. We have a task filed to reorganize this, I'd prefer to merge this and then look at splitting this up.

patina_dxe_core/src/gcd/spin_locked_gcd.rs
os-d: `efierror_to_pterror()` is the opposite way, it maps `EfiError::NotFound` to `PtError::NoMapping`. It's a 1:N case, where `PtError::NoMapping` has multiple possible `EfiError` values. In the `efierror_to_pterror()` case that is trying to make sure we return a `PtError` in the fn that is expecting it, whereas this is the reverse.

A single `Impl From<PtError> for EfiError` does make sense and I will add that.

patina_dxe_core/src/gcd/spin_locked_gcd.rs
makubacki: That sounds fine.

patina_dxe_core/src/gcd/spin_locked_gcd.rs
makubacki: Yeah, what I was getting at is whether it is 1:1 or 1:N and, if not 1:1 does it really matter based on the EFI error context. If it does fine, but generally `EfiError::NoMapping` is more precise.

os-d: The `efierror_to_pterror()` case is just trying to handle that the `apply_caching_attributes()` may return `EFI_NOT_FOUND`, but the `map_memory_region()` API needs to return a `PtError`, so when it encounters `EFI_NOT_FOUND`, it uses the closest relevant `PtError`. In this case, where we are interpreting the `PtError` and converting to an `EfiError`, I think `EfiError::NoMapping` is the right case. We could go change `apply_caching_attributes()` to return `EfiError::NoMapping` instead.

os-d: Actually, one issue with this is that patina-paging doesn't rely on the SDK, so it can't `impl From<EfiError> for PtError` and `EfiError` itself is defined in the SDK, so we'd have to add a patina-paging dependency in the SDK to do `impl From<PtError> for EfiError`, because we can only define that where the type is defined. I don't want to add the patina-paging dependency there because I want to keep it scoped to the core.

makubacki: I'm assuming a wrapper type could be used in the core, but it's not necessary.

PR #1786, opened 2026-09-16
Confidential Compute: Don't Log Between GDT Install and IDT Install

os-d:
## Description

Confidential VMs cannot log between the GDT being installed and the IDT being installed because MMIO access triggers #VE/#VC exceptions which cannot be handled until we've installed the IDT with the handlers.

This commit disables logging through the log crate globally during this transition. It is only done for the confidential VM scenario (as gated by the feature flag) because other platforms can get valuable logs in the event of failure here.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

On an Intel TDX VM.

## Integration Instructions

Confidential VMs set the feature flag. Note that this is being added here but is described in https://github.com/OpenDevicePartnership/patina/pull/1784 as that was originally adding it.


patina_dxe_core/src/cpu.rs
```diff
@@ -118,15 +118,25 @@ pub trait CpuInfo {
...
     })?;
 
     let mut interrupt_manager = Interrupts::new();
     interrupt_manager.initialize().inspect_err(|err| {
-        log::error!("Failed to initialize Interrupt Manager: {err}");
+        log::error!("Failed to initialize Interrupt Manager: {}", err);
     })?;
 
+    // Restore logging
+    log::set_max_level(log_level);
```
vineelko, line 138: This call may not restore the logging if the above calls fail and perform an early return due to `?` muting any sort of logging afterwards. Possibly we might need scoped restore of the logging.

os-d: True, though we are dead in the water if we fail here :). In this case, I actually think it is safer if we leave the logging disabled because if we fail in either of these two steps, we will triple fault and die trying to log. Hitting the assert on this status would be better so a debugger could connect and look. Also safer to prevent triple fault.

patina_dxe_core/src/cpu.rs
```diff
@@ -118,6 +118,13 @@ pub trait CpuInfo {
 #[cfg_attr(coverage, coverage(off))]
 pub fn initialize_cpu_subsystem() -> crate::error::Result<Interrupts> {
     let mut cpu = EfiCpu::default();
+    let log_level = log::max_level();
+
+    // Confidential VMs cannot log between GDT install and IDT install because this creates a #VE/#VC exception
+    // which cannot be handled until the IDT is installed. Disable logging globally through this transition.
+    #[cfg(feature = "confidential_compute")]
+    log::set_max_level(log::LevelFilter::Off);
```
cfernald, line 126: Should we just be installing the GDT & IDT in asm block universally? Seems reasonable and would prevent this and other potential issues if other exception occur during this time.

os-d: That was the direction I originally went, but it required more changes, especially since from the IDT code we are returning the interrupt manager. It wasn't super ergonomic with the current organization of the crate. But I'm not opposed to it. I can go back and show what that looks like.

cfernald (approved):

makubacki (approved):

vineelko (approved):

PR #1787, opened 2026-09-17
[major] Prevent protocol notification at TPL_HIGH

makubacki:
## Description

Note: This can be squashed into earlier commits when major is being prepared to merge into main.

TPLs allowed as arguments to `register_install_notify()` are defined in `patina::component::service::uefi_services::tpl::Tpl`.

This change prevents registering protocol install notifications at `TPL_HIGH`, ensuring that callbacks are only invoked at allowed TPL levels.

Matches the behavior defined in [`CoreCreateEventEx()` in edk2](https://github.com/tianocore/edk2/blob/96341752cef63f3626d64d0b19f60202367627cf/MdeModulePkg/Core/Dxe/Event/Event.c#L332-L339) for `EVT_NOTIFY_SIGNAL` events.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- `cargo make all`

## Integration Instructions

- N/A


patina_dxe_core/src/uefi_services/protocol.rs
```diff
@@ -108,6 +108,11 @@ impl ProtocolServices for CoreProtocolServices {
         notify_tpl: Tpl,
         callback: NotifyCallback,
     ) -> Result<NotifyRegistration, ProtocolError> {
+        // `efi::EVT_NOTIFY_SIGNAL` requires the notification TPL to be <= Tpl::Notify, so Tpl::HighLevel is invalid.
+        if notify_tpl == Tpl::HighLevel {
```
os-d: Should the check be exactly that, `if notify_tpl > Tpl::Notify`?

makubacki: This accepts an enum type where the allowed variants are type enforced, so practically it doesn't matter right now. I'm fine to do that though and will go ahead and implement `PartialOrd` to allow it.

makubacki: Updated

os-d (approved):

joschock (approved):

joschock:
This could go into main....

makubacki:
> This could go into main....

I might bring it in early.

PR #1790, opened 2026-09-17
Support Arm64 Windows host development [Rebase & FF]

### LINKED ISSUE #875: [Bug]: Patina fails to build on Windows-On-Arm host.

### Current Behavior

Patina fails to build on windows on ARM host. There are a number of modules that assume that if the `target_os` is not `uefi`, then it must be x86. This results in e.g. pulling in x86 assembly modules that don't compile for windows-on-arm. In addition, some crates (e.g. uefi-corosensei) explicitly don't support windows-on-ARM and fail to compile with "unsupported target" messages. 

The failing behavior can be reproduced by simply attempting `cargo make build` on a Windows-on-ARM host. 

### Expected Behavior

Patina should be able to compile on Windows-On-ARM hosts. 

### Steps To Reproduce

1. Obtain an ARM host machine running windows.
2. Install pre-reqs for building patina (note: installing `cargo-make` requires additional work for WOA, as some of its dependencies don't compile without `clang` in the path)
3. Attempt `cargo make build` and observe failures related to attempting to compile x86 assembly or unsupported targets.

### Build Environment

```markdown
- OS(s): Windows 11 on ARM
- Tool Chain(s): Standard rustc toolchain for windows on ARM + clang (required by cargo make).
- Targets Impacted: aarch64-pc-windows-msvc
```

### Version Information

```text
Tag: 11.3.3
```

### Urgency

Medium

Someone else needs to fix it

No maintainer feedback needed

### Anything else?

_No response_

makubacki:
## Description

Today, Patina cannot be used on `aarch64-pc-windows-msvc` hosts for two main reasons:

1. Corosensei does to support the target as tracked in: https://github.com/Amanieu/corosensei/issues/29
2. The `aarch64-pc-windows-msvc` produces malformed profile data as tracked in:
   - rust-lang: https://github.com/rust-lang/rust/issues/150123
   - Called out in the cargo-llvm-cov readme: https://github.com/taiki-e/cargo-llvm-cov#platform-support

Out of ~2,000 unit tests only a handful in the dispatcher are actually impacted by (1). So, it is simply stubbed out for that case in image.rs and the code is grouped into `mod`'s to better organize it. We, or someone else, can upstream proper Arm64 Windows Corosensei support in the future, but in the bigger Patina picture, that's not a blocker and has relatively very little impact.

(2) means coverage doesn't work but everything else does (linting, unit tests, etc.). Makefile.toml is updated to skip cargo-llvm-cov and go straight to cargo-nextest on Arm64 Windows hosts.

---

**Makefile.toml: Skip cargo-llvm-cov on aarch64-pc-windows-msvc**

Code coverage is currently not supported in the
aarch64-pc-windows-msvc target as tracked in:

https://github.com/rust-lang/rust/issues/150123

This means it is also not supported by cargo-llvm-cov
as originally tracked in: https://github.com/taiki-e/cargo-llvm-cov/issues/436

Given unit tests otherwise work in Patina on Arm64 hosts,
this just skips code coverage on that host for now.

---

**patina_dxe_core: Support AArch64 Windows hosts**

The main issue with developing Patina on AArch64 Windows hosts right
now is that corosensei does not support the target.

The actual dependencies in test code on corosensei is very minimal
and comes down to a few functions in pi_dispatcher.

This stubs out the coroutine-based dispatch mechanism for AArch64
Windows hosts and consolidates some code in the image module to
group related image handling functionality together.

---

**.github: Run CI on Windows Arm64**

Unit tests now run successfully and code coverage is skipped on
Arm64 hosts (due to Rust language support), so CI can now run
on Arm64 hosts.

The default runners in CiWorkflow.yml included are brought here
in addition to "windows-11-arm". The upstream workflow could be
updated in the future to include `windows-11-arm`, but that would
need further changes in other repos.

---

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- `cargo make all`
- Verify CI results (and tests run) on existing supported hosts and Arm64 (with the [arm-windows-11](https://github.com/actions/runner-images/blob/main/images/windows/Windows11-Arm64-Readme.md) image).

## Integration Instructions

- N/A


makubacki:
Note: It looks like there might be a lot of changes in image.rs, but most of it is just copy/paste of code laying around the file into the new `mod` blocks.

os-d (approved):
Does it make sense to run the build-x64 task on either of the ARM64 environments? I think that use case is pretty limited. Many many more people who would compile natively or cross compile ARM64 from x64. Especially as Windows is the slowest CI environment, may be worthwhile to reduce the tasks it is running.

makubacki:
> Does it make sense to run the build-x64 task on either of the ARM64 environments? I think that use case is pretty limited. Many many more people who would compile natively or cross compile ARM64 from x64. Especially as Windows is the slowest CI environment, may be worthwhile to reduce the tasks it is running.

This is purely a time saving question about essentially gating `build-x64` on Arm64 hosts, right?

os-d:
> > Does it make sense to run the build-x64 task on either of the ARM64 environments? I think that use case is pretty limited. Many many more people who would compile natively or cross compile ARM64 from x64. Especially as Windows is the slowest CI environment, may be worthwhile to reduce the tasks it is running.
> 
> This is purely a time saving question about essentially gating `build-x64` on Arm64 hosts, right?

Correct. I don't think there is much value in crosscompiling for x64 from arm64. I don't think that use case really exists.

makubacki:
> > > Does it make sense to run the build-x64 task on either of the ARM64 environments?
> > 
> > 
> > This is purely a time saving question about essentially gating `build-x64` on Arm64 hosts, right?
> 
> Correct. I don't think there is much value in crosscompiling for x64 from arm64. I don't think that use case really exists.

I'm fine to do that if no one else is opposed. I'll push it in a few minutes.

makubacki:
> I'm fine to do that if no one else is opposed. I'll push it in a few minutes.

It's in ed28ff4d9e7c5aaf18ef9609f7a00030c41e7629

kuqin12 (approved):

cfernald (approved):

makubacki:
> It's in [ed28ff4](https://github.com/OpenDevicePartnership/patina/commit/ed28ff4d9e7c5aaf18ef9609f7a00030c41e7629)

Confirmed it is skipped on the Windows Arm64 host as expected in the CI run:
image

PR #1796, opened 2026-09-21
[major] patina_dxe_core: Tighten down config table type tracking

makubacki:
## Description

Note: This can be squashed into an earlier commit when `major` is merged to `main`.

Makes the type tracking for configuration tables more strict and consistent between configuration table interfaces.

There are two general paths for working with config tables:

- Untyped: C FFI for UEFI Boot Services
  - Compatibility with existing C drivers using Boot Services
- Typed: Safe Rust API available through UEFI Services
  - Used by Patina Rust components

These have to be supported simultaneously while keeping the safe API as safe as possible.

The untyped path (C EFIAPI) is:
  - `unsafe extern "efiapi" fn install_configuration_table`
    - Which calls: `core_install_configuration_table()`
      - Calls: `install_configuration_table_entry()`

The typed path (Safe Rust API) is:
  - `ConfigurationTableServicesExt::install<T>()`
    - Calls: `ConfigurationTableServicesExt::install_typed_table()`
      - Calls: `core_install_typed_configuration_table()`
        - Calls: `install_configuration_table_entry()`

On the typed path, `core_install_typed_configuration_table()` is what checks if an update is allowed (overwrite is allowed) and record the type information for the configuration table entry.

Below that, the same `install_configuration_table_entry()` is used for both paths.

`CoreConfigurationTableServices::get_typed_table()` checks the type tracked map for a table and then gets the pointer from the actual configuration table in memory (shared with the untyped path).

To ensure that the "typed" API path is safe, an update through the untyped path for any configuration table already under a GUID will be unconditionally clear that table entry. This provides compatibility for the C API while ensuring any table considered "typed" is the correct type. This design is about ensuring the safety assertions of a typed table in safe code, not denying updates through the untyped path.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- `cargo make all`
- Boot QEMU Q35 with `patina_samples` components included

## Integration Instructions

- N/A


patina_dxe_core/src/uefi_services/state.rs
```diff
@@ -81,17 +78,13 @@ pub(crate) struct UefiServicesState {
     pub(crate) event_notify: ActiveDispatchStack,
     /// Tracks reentrant dispatch of `CoreProtocolServices` install-notify callbacks.
     pub(crate) protocol_install_notify: ActiveDispatchStack,
-    /// Records the Rust type installed under each GUID via `ConfigurationTableServices::install_typed_table`,
```
makubacki, line 84: @Javagedes, I added `config_table_types` to this `UefiServicesState` struct in a previous PR to attempt to reduce the total number of global statics. I can keep it in the struct (and not add `CONFIG_TABLE_TYPES` as a standalone static in config_tables.rs) in this PR if we really want, but I'd have to change `mod state;` to `pub(crate) mod state;` in uefi_services.rs and I thought it was neater to keep the UEFI services state isolated to that module.

cfernald (approved):

os-d (approved):

patina_dxe_core/src/uefi_services/state.rs
makubacki: Will continue with this for the reason stated.

